PR #2, opened 2016-06-24
[Docs] Source repository has been moved out of Phabricator and onto GitHub

nicolan:
* Development processes have changed (no longer using Arcanist for patch management)
* Repository URLs have changed.
* Bug reporting has changed (Bugzilla and Maniphest will be migrated into a single location: GitHub issues)


docsrc/conf.py
```diff
@@ -1183,14 +1183,14 @@
...
-    #Please see https://docs.cyrus.foundation/ for better maintained
+    #Please see https://www.cyrusimap.org/ for better maintained
     #documentation.
 #"""
 
 # Use this as :task:`18`
 extlinks = {
         'rfc':('http://tools.ietf.org/html/rfc%s', 'RFC '),
-        'task':('https://git.cyrus.foundation/T%s', 'Task #'),
+        'task':('https://github.com/cyrusimap/cyrus-imapd/issues/%s', 'Task #'),
```
elliefm: This will cause problems with all the old links to tasks.

Our release notes up until now are full of references to phabricator tasks (like, "Fixed :task:\`123\`: no longer wurgle the blurgle" etc)

If we replace the 'task' thing, these are all going to turn into links to github issues that don't exist

I would propose that instead of replacing the url in 'task', we add a new extlink, 'issue' for the github issues (with suitable link), and then either
a) leave 'task' as-is, so it points into the old phabricator stuff (which will be a dead link when phab dies), or
b) change 'task' in some way so that it leaves it as just "Task #123" without any link at all

And then our existing references to old phabricator content will continue to work as well as they can without conflicting with github issues, and we have an unambigious way of referring to github issues specifically, going forward.

nicolan: Ah, true. I was hoping that Brong's magic autoimporter would let us identify phab/bugzilla issues in github, but perhaps that was somewhat... wishful.

I'll leave it pointing to Phab for now and make a new :issue: extlink. When everything gets merged, we can revisit this.

PR #13, opened 2016-08-01
t114 dont replicate compiled sieve bytecode

elliefm:
(From an old phabricator task: https://git.cyrus.foundation/T114)

This stops replication of binary sieve bytecode files.  Instead, bytecode files are compiled by the replica on upload of the plain text script, or on activation via sync protocol.  Attempts to replicate bytecode files are silently ignored.

A new sync protocol error is introduced, IMAP_SYNC_BADSIEVE, which will result if a replicated sieve script fails to parse/compile.  This makes it important to ensure your master and replica servers have the same sieve extensions enabled.

Also, lmtpd now recompiles sieve bytecode files (if missing or stale) on demand.  So a temporary parse/compile failure during replication is somewhat self-healing if the sieve script needs to actually be invoked later.

If an old master (without this change) syncs to a new replica (with this change):
* master will send script files => replica will compile them to bytecode
* master will send bytecode files => replica will ignore them
* master will activate bytecode file => replica will recompile it (if missing or stale) and activate it

If a new master (with this change) syncs to an old replica (without this change):
* master will send script files => replica will accept them
* master will not send bytecode files => bytecode files on replica will be missing or stale
* master will activate bytecode file => replica will create symlink to missing or stale bytecode file
* if the replica needs to be brought into active service, something like tools/masssievec should be run first to ensure required bytecode files are present and current

If a new master (with this change) syncs to a new replica (with this change):
* master will send script files => replica will compile them to bytecode
* master will not send bytecode files => replica doesn't need them anyway
* master will activate bytecode file => replica will recompile it (if missing or stale) and activate it

This is somewhat invasive, so making it a pull request rather than just pushing it up.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -835,14 +835,14 @@ backup_backupd_SOURCES = \
     imap/sync_support.h \
     master/service.c \
     backup/backupd.c
-backup_backupd_LDADD = backup/libcyrus_backup.la $(LD_SERVER_ADD)
+backup_backupd_LDADD = backup/libcyrus_backup.la $(LD_SIEVE_ADD) $(LD_SERVER_ADD)
 
 backup_ctl_backups_SOURCES = \
     imap/mutex_fake.c \
     imap/sync_support.c \
     imap/sync_support.h \
     backup/ctl_backups.c
-backup_ctl_backups_LDADD = backup/libcyrus_backup.la $(LD_UTILITY_ADD)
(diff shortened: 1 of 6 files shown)


elliefm:
@brong I'm starting to get blocked on this -- I've got a bunch of other changes ready to go (and more in progress) that will be conflict hell if I push them first.

PR #18, opened 2016-08-22
Doc restructure: a simpler hierarchy

nicolan:
Making the initial left hand navigation simpler, easier to follow and faster to locate information as it's less cluttered.

Splitting up the old concepts guide into Concepts and Admin, and removing leftover html cruft that was from a time long ago.

Restructuring some of the content to get the hierarchy easier to read and navigate.

### DIFF
--- a/docsrc/concepts.rst
+++ b/docsrc/concepts.rst
@@ -1,0 +1,12 @@
+========
+Concepts
+========
+
+
+.. toctree::
+    :maxdepth: 2
+    :glob:
+
+    imap/features
+    imap/deployment
+    imap/admin
--- a/docsrc/concepts.rst
+++ b/docsrc/concepts.rst
@@ -5,7 +5,7 @@ Concepts
 
 .. toctree::
     :maxdepth: 2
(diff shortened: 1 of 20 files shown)


elliefm:
Looks good at a quick look through now, I'll pull it down and try it out properly tomorrow

nicolan:
The file structure on disk does not match the new hierarchy visually. And I think it should correlate for ease of management, but I'm holding off until I've finished further restructure and subdivision so the physical shuffle can be done in one lot.

elliefm:
Just pulled it onto my local repository, built docs, and it's looking good, no warnings or errors or anything.

Do you want this merged as is (and then presumably the rest of the restructure becomes a separate pull request, later), or do you want to leave it hanging for a bit and keep working further into it?

nicolan:
Merge as-is, I reckon. It looks shiny enough that it's worth pushing live. If I wait until it's "all done", this won't get merged until 2018.,

elliefm:
Done :)

PR #22, opened 2016-08-26
Doc directories

nicolan:
Physical restructure to match the virtual hierarchy. Lots of file moves. 

@elliefm Please check the toplevel build - not sure it's right!


Makefile.am
```diff
@@ -1748,15 +1748,15 @@ snapshot::
 	tar -C snapshot -c -f - $(PACKAGE_NAME)-$(GIT_VERSION) | gzip -9 > $(PACKAGE_NAME)-$(GIT_VERSION).tar.gz
 	@rm -rf snapshot
 
-docsrc/imap/admin/systemcommands/cyradm.rst: $(top_srcdir)/perl/imap/IMAP/Shell.pm
+docsrc/imap/reference/manpages/systemcommands/cyradm.rst: $(top_srcdir)/perl/imap/IMAP/Shell.pm
```
elliefm: These three filenames also need to be changed a bit further down in the "GENERATED_RSTS" variable, around line 1787

Also, around line 55 of .gitignore

elliefm:
The top level build fails on the man pages:

```
Exception occurred:
  File "/usr/lib/python2.7/dist-packages/sphinx/environment.py", line 1259, in get_doctree
    f = open(doctree_filename, 'rb')
IOError: [Errno 2] No such file or directory: u'/home/ellie/fastmail/cyrus-imapd/docsrc/build/doctrees/imap/admin/systemcommands/arbitron.doctree'
```

... and also fails the same way if you try to "make man" in the docsrc directory.  I can see a couple of things that might be causing this:

* The `man_pages =` section of docsrc/conf.py still lists all the old paths
* The arbitron.rst contains `.. _imap-admin-systemcommands-arbitron:` near the top, which presumably needs to be changed to match the new path? (Or maybe just deleted? I don't know for certain, but I don't think it's necessary.)  Most (all?) of the man page rst's seem to have this sort of thing.  A bunch of the other rst's seem to have it too, a label at or near the top of the page based on the (old) filename

elliefm:
The three generated rst files have been added to git, when they shouldn't be:

* docsrc/imap/reference/manpages/configs/imapd.conf.rst
* docsrc/imap/reference/manpages/systemcommands/cyradm.rst
* docsrc/imap/reference/manpages/usercommands/sieveshell.rst

All three were added by 182c739 [Docs] Physical restructure of files to match new hierarchy.

elliefm:
`tools/config2rst` around line 194 adds a section label based on the old path (when generating imapd.conf.rst)

nicolan:
Commits 2e5f8fa and b836d54 should fix all the issues you've mentioned.

I can't test the toplevel build as I don't have a full build environment, so if it's still barfing, send me more of the output.

Thanks so much for the detailed feedback. I'll know what to watch for next time, now!

elliefm:
It builds now! :)

Poking around in the generated docs, the front page near the top has a link to "Cyrus's full list of \<features\>" which is a dead link now?

elliefm:
On the subject of link checkers, I've discovered that sphinx-build has a -n ("nitpicky") option.

Running with that, I get a bunch of warnings like:

```
WARNING: u'cyrusman' reference target not found: cyradm(1)
```

I've had a look at a few of them and they seem to be references to man pages from the wrong section (or maybe: stale references to man pages that have since moved to a different section).  (Per example, cyradm is in section 8, not section 1.)

There might be a few that are referencing docs that are in the right spot but which are missing the `.. cyrusman:: foo(n)` label at/near the top, but I haven't looked too deeply for this case.  ... Actually, `mknewsgroups` seems to fit this case

And there might be a few that are referencing docs that don't actually exist?

To be clear, I don't think these are problems new to this pull request, they're just stuff that's turned up now because I've discovered the nitpicky option now. :)

On the bright side: it doesn't whinge about anything else, so maybe that means the rest of the links are all good :)

elliefm:
Are "genindex.html" and "sitemap.html" linked to from anywhere?

elliefm:
I've made sphinx-build nitpicky on master :)

nicolan:
Neither of those are linked to from anywhere. I was toying with the idea
of including them but I think (?) they are flagged :orphan: so the build
doesn't complain.  Maybe.


----- Original message -----

elliefm:
Ah yeah, I found them with a home grown link checker I've been toying with, which doesn't know about :orphan: because it doesn't read the rst, only the resulting html...

flrl:
I think this is okay to merge now?

elliefm:
I think this is okay to merge now?

Issue #38, opened 2016-10-03
Sieve vacation does not support UTF-8 in :subject


alecpl:
As mentioned in https://tools.ietf.org/html/rfc5230#section-4.3 in "subject" UTF-8 characters are allowed. The implementation must encode subject strings according to https://tools.ietf.org/html/rfc2047.

So, if you do:

```
if address :matches "from" "*@example.com"
{
     vacation :subject "ä,ö,ü"
            "Having lots of fun! Back in a day or two!";
}
```

Subject of the reply will be malformed.
Tested with cyrus 2.5.8.

elliefm:
Fixed in 0338a62

PR #58, opened 2016-10-14
Encode the subject line of an automatic response.

kanarip:
Resolve cyrusimap/cyrus-imapd#38 on cyrus-imapd-2.5.


elliefm:
Cherry-picked 0338a62 from master instead (which avoids leaking the encoded string)

PR #67, opened 2016-10-27
Correctly translate dotted userids between internal and external

kanarip:
mboxname_user_mbox() and mboxname_to_userid() did not correctly take in to account that a userid may contain dots in the local part (i.e. 'john.doe@example.org').


brong:
I'm fine with this for 2.5.  It's solved much more comprehensively on master, but the changes are far too invasive to merge back.

PR #70, opened 2016-11-03
Preserve the original partition of a (sub-)mailbox when using XFER

kanarip:
Preserve the partition of a (sub-)mailbox when using XFER, rather than use the highest-level mailbox of the hierarchy transferred as the target partition for all mailboxes in the hierarchy.

The use-case is mailbox hierarchies split over multiple partitions.

Example scenario;

  * `user.john` resides on the default partition, and `user.john.Archive` resides on a partition archive.

  * While 'default' is fast, expensive storage, we have only 1GB of it.

  * While 'archive' is slow, cheap storage, we have a few TBs of it.

  * The size of `user.john.Archive` is > 1 GB.

With this patch, unless a partition is explicitly specified to the issuing of the XFER command, any mailbox residing on;

  * the local default partition will be replicated to the target default partition,

  * a local non-default partition will be replicated to the target partition of the same name.

I do not believe this patch makes any sense for master, as I understand XFER has transitioned to the sync protocol, but please do correct me if a similar scenario may exist there, that deserves attention.

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -10106,7 +10106,11 @@ static int xfer_init(const char *toserver, const char *topart,
     xfer->remoteversion = backend_version(xfer->be);
 
     xfer->toserver = xstrdup(toserver);
-    xfer->topart = xstrdup(topart);
+
+    // Only use a global target partition if it has been explicitly specified.
+    if (topart)
+	xfer->topart = xstrdup(topart);
+
     xfer->seendb = NULL;
 
     /* connect to mupdate server if configured */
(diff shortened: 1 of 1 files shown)


kanarip:
Please note that the change to RFC 4466 style inclusion of the target partition in the changed LOCALCREATE command issuance has to do with a later change I'm yet to create a pull request for.

elliefm:
What happens if the remote server has the higher level partition, but not the submailbox partition(s)?

Will it:

a) fall back to the original behaviour of just putting the whole hierarchy on the one partition?
b) fail to copy the submailboxes, causing the whole XFER to fail?
c) fail to copy the submailboxes, but the XFER still reports "success"?

I don't really care much one way or other between a or b; but c would be unacceptable

kanarip:
The target backend for the transfer should respond as if an invalid partition was specified, meaning that it should bail out on the XFER -- so option b).

rsto:
The XFER command is deprecated.

PR #71, opened 2016-11-03
Preserve the existing unique ID of a mailbox renamed / transferred

kanarip:
A renamed mailbox doesn't really create a new mailbox, but more importantly, a transferred mailbox doesn't really create a new mailbox -- it's just moved.
    
The use-case in which this would otherwise be problematic is in transferring a mailbox between replicated backends. The replica master on the receiving end would (prematurely) create a mailbox, with a new uniqueid, provided the command issued is a `LC1 LOCALCREATE user.john (PARTITION default)`. The dumping and undumping of the mailbox's `cyrus.header` file would continue to refer to the original uniqueid, however, and thus so would `GETMETADATA`.
    
This patch ensures that the command is, instead;
    
 `LC1 LOCALCREATE user.john (PARTITION default UNIQUEID <uniqueid>)`.
    
It is not allowed for non-admins and non-proxyservers.

Also note that the 'active' database entry for a mailbox moved between partitions was "lost" temporarily, as `mboxlist_renamemailbox()` did not care to preserve the original in the new record.

**IMPORTANT:** This pull request includes #70, upon which 2eca709432e87fbbc85c7a3ff8693933497627b8 depends.

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -10106,7 +10106,11 @@ static int xfer_init(const char *toserver, const char *topart,
     xfer->remoteversion = backend_version(xfer->be);
 
     xfer->toserver = xstrdup(toserver);
-    xfer->topart = xstrdup(topart);
+
+    // Only use a global target partition if it has been explicitly specified.
+    if (topart)
+	xfer->topart = xstrdup(topart);
+
     xfer->seendb = NULL;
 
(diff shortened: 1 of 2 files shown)


rsto:
This already is fixed with the mailbox-by-id changes.

PR #73, opened 2016-11-13
Several changes.master

dilyanpalauzov:
I have done these changes locally over the last year and would be glad, if they are integrated on the master branch.

PR #74, opened 2016-11-13
Several changes.cyrus25

dilyanpalauzov:
I have done these changes locally over the last year and would be glad, if they are integrated on the cyrus-imapd-2.5 branch.


elliefm:
Please see my comment on #73 about splitting this up into separate pull requests per topic.

There's a bunch of good commits in here that I would happily merge now, except I can't because there's also a bunch of bigger stuff that warrants more extensive review.

PR #79, opened 2016-12-02
Change to build in-place libcrc32 library with libtool

obache:
libtool sometimes be confused to link shared library when linking with
static in-place library, and only generate static library.


elliefm:
Thanks -- I've been thinking about this change too.  Looks like the right thing to do.

PR #80, opened 2016-12-07
Updated git clone command

mellertson:
Updated git clone command.  The git clone command doesn't work (at least not for me) in its current form.
Updated markup on all other commands meant to be issued in *nix shell so they are formatted correctly.

### DIFF
--- a/docsrc/imap/installation/diy.rst
+++ b/docsrc/imap/installation/diy.rst
@@ -21,19 +21,19 @@ Clone the GIT repository:
 
 .. parsed-literal::
 
-    $ `git clone git@github.com:cyrusimap/cyrus-imapd.git`
+    $ :command: `git clone git@github.com:cyrusimap/cyrus-imapd.git`
 
 Check out the desired branch or revision:
 
 .. parsed-literal::
 
-    $ `git branch -la`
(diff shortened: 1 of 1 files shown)


mellertson:
I noticed some notation in the markdown like the following
:command:

I Google'd it, but it seems to not be part of the current mark down syntax on GitHub.  Maybe it's older syntax?  I'm not sure.  I replaced it by surrounding the command using the back-tick character.  It looks correct to me, but please let me know if I misinterpreted something.

elliefm:
The reason you can't find a reference to the markdown notation `:command:` is because these files are not markdown, they are restructured text (which is similar, but not the same) -- see http://www.sphinx-doc.org/ and http://docutils.sourceforge.net/rst.html.

The end result of this particular rst file can be viewed here: http://www.cyrusimap.org/imap/installation/diy.html

The correction to the git clone command itself looks good, but could you please amend your pull request to not alter the `:command:` notation?

mellertson:
Thanks for the feedback,  that makes sense.   I'll change it back to include the :command:  notation as per your suggestion.

elliefm:
Thanks :)

PR #1753, opened 2016-12-20
perlmake.m4: change to use perl instead of sed for portability

### LINKED ISSUE #78: SED portability issue in perlmake.m4
In perlmake.m4, `sed` is used, but not portable,

1. `-i` option

It may be in GNU and some BSD implementations, but not in the POSIX standards.
Additionally, some implementations require EXTENSION parameter for backup file name.

2. starting brank after `a` command

Some BSD origin SED strip initial blanks after 'a' command.
It result in creating broken Makefile, with no tab prefixed shell commands.

Maybe, it should be rewriten with perl, because it should available there.

obache:
fixes #78

### DIFF
--- a/cmulocal/perlmake.m4
+++ b/cmulocal/perlmake.m4
@@ -47,14 +47,15 @@ AC_CONFIG_FILES([$1/Makefile.PL])
 AC_CONFIG_COMMANDS($1/Makefile,[
     ( cd $1;
       $PERL Makefile.PL $MAKE_MAKER_ARGS;
-      sed -i -e'/^uninstall_from_sitedirs ::/{
-		n
-		s/\$(SITEARCHEXP)/\$(DESTINSTALLSITEARCH)/
-		a \
-	$(RM_F) "$(DESTINSTALLSITEARCH)/auto/$(FULLEXT)/.packlist"
-		a \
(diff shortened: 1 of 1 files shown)


cmulocal/perlmake.m4
```diff
@@ -47,14 +47,15 @@ AC_CONFIG_FILES([$1/Makefile.PL])
...
-	$(RM_F) "$(DESTINSTALLSITEARCH)/auto/$(FULLEXT)/.packlist"
-		a \
-	$(RM_F) "$(DESTINSTALLSITEARCH)/perllocal.pod"
-	  }' Makefile
+      $PERL -i -pe'next unless /^uninstall_from_sitedirs ::/;
+		print $_;
+		$_ = <>;
+		s/\$\(SITEARCHEXP\)/\$\(DESTINSTALLSITEARCH\)/;
+		$_ .= <<'END';
+	\$(RM_F) \"\$(DESTINSTALLSITEARCH)/auto/$(FULLEXT)/.packlist\"
```
elliefm: Otherwise, looks good, thanks!  I've fixed up the missing `\` myself

obache: Thanks for your review.
It might be lost between my test and commit :(

Issue #1721, opened 2015-04-24
build broken on BSD (__compar_fn_t)


brong:
From: [Antoine Jacoutot](ajacoutot@bsdfrog.org)
Bugzilla-Id: 3892
Version: 2.5.x (next)
Owner: [Ken Murchison](murch@andrew.cmu.edu)

elliefm:
Fixed by #1755

PR #1755, opened 2016-12-22
Fix condition for missing __compare_fn_t

### LINKED ISSUE #64: Wrong condition for own __compare_fn_t typedef
In lib/util.h:

```
#ifndef __GNUC__
typedef int (*__compar_fn_t)(const void *, const void *);
#endif 
```

But I can only find `__compar_fn_t` in glibc internal usage.
I don't know where it came from, but It may be mistake, `__GLIBC__` instead of `__GNUC__`?
Or configure should detect and use  glibc exported `comparison_fn_t` type instead.

obache:
"__compare_fn_t" for qsort(3) compare function prototype is not a
GCC feature, but a internal private type for glibc and some other
libc implementations.

Change to detect the type with configure script.

fixes #64 #1721

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -207,6 +207,9 @@ AC_STRUCT_TM
 AC_STRUCT_TIMEZONE
 AC_CHECK_FUNCS(timegm)
 
+dnl glibc internal qsort(3) compar function prototype
+AC_CHECK_TYPE(__compar_fn_t, AC_DEFINE(HAVE_GLIBC_COMPAR_FN_T,[],[Do we have glibc __compar_fn_t?]))
+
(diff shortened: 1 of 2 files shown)


elliefm:
Looks good, thanks :)

elliefm:
Does ec2e5f3ba4ef637403058bfa89eef0e754ebdcf0 introduce any problems for you?  It removes the need for the __compar_fn_t typedef entirely, by no longer casting the last argument in qsort() calls.

Seems like it should be fine, but maybe there's a portability concern here?  I don't know why we were casting like this when the function signatures are all consistent already.

PR #1756, opened 2016-12-26
Revert getmetadata list changes

javitonino:
I have a Kolab installation and I have recently update my cyrus installation (from 2.57 to 2.5.10). After the upgrade, I started having problems listing mailboxes in roundcube. The listing is done using the following IMAP command: `GETMETADATA * /shared/vendor/kolab/folder-type`. The problem is that cyrus-imapd 2.5.10 returns `NO Mailbox does not exist`, but in 2.5.7 it works.

After looking at the changelog, I tracked the problem to a set of changes by @kanarip, that change the way GETMETADATA interprets lists vs patterns. This PR reverts those changes and fixes the problem I'm seeing.

Here's a gist comparing the outputs pre and post patch: https://gist.github.com/javitonino/cb3d83dd40b5ab351c0297c527252034. The output is the same for mailbox lists, but the current version fail for mailbox patterns, and the patched version works.

Based on my observations, I think reverting this patch is correct, but I may be missing some key point on why this was initially applied. Maybe @kanarip can shed some light into the original patch rationale.

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -8156,7 +8156,7 @@ static int parse_metadata_string_or_list(const char *tag,
 
     // Assume by default the arguments are a list of entries,
     // until proven otherwise.
-    *is_list = 0;
+    *is_list = 1;
 
     c = prot_getc(imapd_in);
     if (c == EOF) {
@@ -8204,9 +8204,9 @@ static int parse_metadata_string_or_list(const char *tag,
 	strarray_append(entries, arg.s);
 
 	// It is a list if there are wildcards
-	if (!strchr(arg.s, '*') && !strchr(arg.s, '%')) {
+	if (!strchr(arg.s, '*') || !strchr(arg.s, '%')) {
 	    // No wildcards; Not a list
-	    *is_list = 1;
+	    *is_list = 0;
 	}
     }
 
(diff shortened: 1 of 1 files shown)


brong:
When the two of you have figured out what's correct (and convinced @elliefm) I'd love to see a Cassandane test to make sure that it's correct for 3.0 as well.

kanarip:
Does this additional commit help your case?

https://github.com/kanarip/cyrus-imapd/commit/56941ac425b5ebd18dbf2c798209447751bcc448#diff-8cdff4c243667531f3986d6afc08e0d0

javitonino:
I applied that commit on top of 2.5.10 and it works correctly for all my tests cases ✨

kanarip:
@javitonino -- I would rather see the new commit accepted, than the former commits reverted.

@brong -- I think this ugliness may no longer apply to master/3.0, because IIRC, the GETMETADATA options sorting got cleaned up to be solely RFC 5464 compliant (spec. vs. examples, IIRC, were the cause for confusion originally) -- but of course I'm not entirely sure this is the case.

javitonino:
Applying your commit looks good to me. The only difference with this reversion is that your commit always sets single-argument lists as patterns. and I don't think that's a problem, as the pattern matcher works ok for strings without wildcards.

javitonino:
I forgot about this for enough time to become irrelevant. Closing it.

PR #1771, opened 2017-01-03
Adding -w option to man pages

zdohnal:
Hi,
I found out option '-w passwd' is missing from manual pages, but it is in help. This commit will fix it.


nicolan:
Looks great, thanks for contributing this! Is the -w option also valid on other commands?

ksmurchison:
-w is valid for any command that is a variant of imtest, which means pop3test, nntptest, lmtptest, smtptest, mupdatetest, sivtest, synctest

zdohnal:
ksmurchison:
Sorry, I did not know that. I will spread it to all that *test man pages.

PR #1784, opened 2017-01-06
[Docs] 3.0 upgrade guide.

nicolan:
@elliefm @brong ( @rsto ) Can you please eyeball this guide?

This is a first cut of the 3.0 upgrade guide. It assumes (correctly? incorrectly?) that people will maintain two physical installations (one of existing, one of new 3.0) until they're ready to actually cut over. Is that assumption correct? (It does mention you can replicate to a 3.0 install, but for more info I might need ellie to write that bit)

It assumes that someone doing this will know where certain files (config/sieve/databases/spool/etc) live and being told explicitly is overkill. Is that reasonable or should I spell out the default locations?

### DIFF
--- a/docsrc/imap/concepts/deployment/databases.rst
+++ b/docsrc/imap/concepts/deployment/databases.rst
@@ -1,3 +1,5 @@
+.. _databases:
+
 =========
 Databases
 =========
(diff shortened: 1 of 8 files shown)


elliefm:
Seems alright to me?

postilion:
Nicola,
I've done some updates to this in https://github.com/cyrusimap/cyrus-imapd/pull/1789/commits

Please review & comment.
Cheers,
-nic

PR #1792, opened 2017-01-15
Rewrote translatesieve to work with 3.0

postilion:
The tools/translatesieve utility is quite needed for the 2.X > 3.X
transition, given that unixhierarchysep and altnamespace now default
to "on."  However, the old utility didn't handle all scenarios.

This version is more robust and has useful help, etc.

Also updated the manpage for same.


postilion:
Abandoning this pull request due to extraneous cyr_info cruft.

PR #1794, opened 2017-01-16
Fixed cyr_info.rst to match actual command names

postilion:
The man page for cyr_info(8) listed commnd names which didn't match
the actual program.

Also, cyr_info.c had a typo for the "conf-default" command.

This is a second attempt, with a clean git tree.


elliefm:
Thanks :)

PR #1796, opened 2017-01-16
Update man pages for the PTS group of commands.

postilion:
Fleshed out details for ptexpire(8), ptdump(8) and ptloader(8)
to replace the skeleton files which were there.

Moved these man pages from "work in progress".


elliefm:
Awesome, thanks, now I feel like I have some idea what these things actually do :)

PR #1797, opened 2017-01-17
Cyr info2

postilion:
Add "reid" command information to cyr_info

The "reid" command for cyr_info program was not documented.  This
has been added both to cyr_info(8) man page and to the usage.


postilion:
Re-basing, will re-issue.

PR #1803, opened 2017-01-20
Add supports to build on missing glibc extension `qsort_r`

### LINKED ISSUE #63: Using glibc extension qsort_r(3)
In imap/http_admin.c, qsort_r(3) is used, but it is glibc extension, not available for others.
Worse yet, FreeBSD (and having base system from it, macOS, DragonFly BSD,...) also has qsort_r(3), but different signature than glibc's one, so it may cause runtime crash even if compiled fotunately.

C11 standard define qsort_s(3), same functional with yet another signature, but it is not widely supported yet.

See also [Portable qsort_r / qsort_s](https://github.com/noporpoise/sort_r)

obache:
Some *BSD systems has `qsort_r` function with different signature
than glibc's one.  Add macros resolving such differences.

Additionally, add glibc `qsort_r` compatible function for others
with GCC compiler.

partially resolve #63

### DIFF
--- a/imap/http_admin.c
+++ b/imap/http_admin.c
@@ -478,10 +478,56 @@ static int add_procinfo(pid_t pid,
     return 0;
 }
 
-static int sort_procinfo(const struct proc_info **a, const struct proc_info **b,
-                         const char *key)
+#if defined(_GNU_SOURCE) && defined (__GLIBC__) && \
+	((__GLIBC__ > 2) || ((__GLIBC__ == 2) && (__GLIBC_MINOR__ >=0)))
+#define HAVE_GLIBC_QSORT_R
+#endif
+
+#if defined(__NEWLIB__) && \
+	((__NEWLIB__ > 2) || ((__NEWLIB__ == 2) && (__NEWLIB_MINOR__ >= 2)))
+#if defined(_GNU_SOURCE)
+#define HAVE_GLIBC_QSORT_R
+#else
+#define HAVE_BSD_QSORT_R
+#endif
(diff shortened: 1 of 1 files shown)


elliefm:
This looks okay to my eyes, but I don't have a BSD system to test it on.

Though, could the implementation detection be handled by configure.ac instead of by http_admin.c?  That way people whose platform has no qsort_r at all get an error at configure time rather than sometime during the build.  It'll probably also make it easier to drop in an alternative implementation for such platforms -- #1808 suggests Solaris has no implementation of qsort_r at all, for example.

obache:
Now 3.0.0 is RC1, so this patch is minimalized (not affect to currently buildable platforms), so "partially".

It is nasty to detect `qsort_r` with configure script and danger to believe it at compile time,
because same name but different behavior, and may be selectable them with visibility macros.

elliefm:
Thinking about it more, I agree, doing this in configure.ac would be nasty.

I wonder if it's possible to move it to it's own module, like say `lib/xqsort_r.c` -- similarly to how `lib/xsha1.c` handles abstraction of `SHA1()`.  Though in the qsort_r case, the need to parameterise the arguments to `sort_procinfo()` complicates things

obache:
I thought too past, and if you can provide glibc qsort_r compatible function,
you can use it unconditionally for any platforms.

elliefm:
I think I'm happy with what you've got here, but let me sit with it over the weekend, and if I'm still happy on Monday I'll merge it in :)

PR #1828, opened 2017-02-08
Complete the mailspool location doc, etc.

postilion:
Complete the mailspool location doc and add a new locations doc
for Xapian search tiers.

### DIFF
--- a/docsrc/imap/reference/admin/locations/mailspool.rst
+++ b/docsrc/imap/reference/admin/locations/mailspool.rst
@@ -7,25 +7,86 @@ Default Spool location
 ----------------------
 
 The main Cyrus IMAP mail data directory structure is the spool, and its
-default location is defined by the ``defaultpartition`` and
-``partition-default`` entries in :cyrusman:`imapd.conf(5)`.
+location is defined by ``partition-name`` entries in
+:cyrusman:`imapd.conf(5)`.
 
-Despite their similar names, these two configuration values have
-distinct meanings.  The first is the name of the default spool
-partition, the second is the path to it:
-
-.. include:: /imap/reference/manpages/configs/imapd.conf.rst
-        :start-after: startblob defaultpartition
-        :end-before: endblob defaultpartition
+Additionally, on servers with more than one spool partition, that
+partition in which new user mailboxes are to be created is specified by
+the ``defaultpartition`` directive.
 
 .. include:: /imap/reference/manpages/configs/imapd.conf.rst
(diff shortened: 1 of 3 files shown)


elliefm:
The use of a partition named "archive" is going to be really confusing, because of the also existing **archivepartition-_name_** group of settings

In my understanding the main benefits of dividing storage into **partition-_name_**'s are distributing IO load around (more disks = more r/w heads), cheaper capacity (at a certain point, multiple smaller disks are cheaper than one gigantic one), and reduced impact of disk issues (one partition fails, users on other partitions unaffected).

[I would also expect that most deployments would actually have RAID sets underlying each partition (which obviously provides its own redundancy/IO throughput improvements depending on its configuration) or SAN or whatever, but the point still stands about one failing partition (of whatever underlying implementation) not taking out users on other partitions :)]

I guess partitions could also be used if you have a bunch of logical partitions (in the fdisk sense) on a single physical disk, but that seems to defeat all the purposes of Cyrus partitioning

The fast/slow storage in your example is kind of more like what the archivepartition stuff is meant for.  I'm not sure if we have documentation about how that works but I think we do (because otherwise I wouldn't know?)

postilion:
Okay, I'll change the example names.  I understand all of the partition-related points you've made, and that stuff is covered in other documents, such as concepts/deployment/storage.rst or concepts/features/mail-spool-partitions.rst.  

What I was trying to achieve with the "locations" set of documents was to address the issue, raised by Nicola in this message:

> On 01/22/2017 11:20 PM, Nicola N wrote:
> >
> > The databases doc covers the databases, but not the process files, log files, config files and the care and feeding.

so it's really mostly about how to determine where stuff is or will be in your Cyrus installation.  I wrote this thinking of an administrator taking over an existing installation, and trying to figure out where stuff is, but also thinking of a careful planner, who wants to understand that topic before committing to a configuration.

Does that make sense?  In other words, to me this should become a treatise on why one partitions, but rather an explanation of how to figure out partitions, and how to change them, etc.

As for the fast/slow vs. archive; the archive is not available until v3.0, and this document is meant to be more general.  I have recently upgraded a recalcitrant client from 2.3.16 to 2.5.10, and have them using just this sort of fast/slow scheme, since they wouldn't agree to beta test 3.0, but needed an improvement. Still, your point is valid.  I'll try to come up with something clearer.

Thanks for the input! :-)

elliefm:
Looks good, thanks! :)

Issue #1826, opened 2017-02-08
squatter's usage() and manual page don't list all allowed options


rsto:
squatter accepts options ````C:I:N:RXT:S:Fc:de:f:mn:rsiavz:t:ou```` but both it's
usage() message as well as the manual page are missing a couple of these

postilion:
@rsto, I'll be happy to take care of the manpage updates, as long as the usage message is corrected.

rsto:
I've completely overhauled the usage message in https://github.com/cyrusimap/cyrus-imapd/commit/4643639d6fc9637b964fa1cc6090ec40cfbc099b

Note that getopt was instructed to accept option 's' but squatter rejected it as unknown. I've removed it. I also started to cross-check the usage string with the manual page and it seems as if we need to iterate: Some options in the manual page are not supported by squatter (-A comes to mind), others aren't documented.

Is the new usage() message OK for you to read? It's a bit tricky to show on the command line which options are OK to combine with which one. Thanks.

rsto:
Yay, one step closer to 3.0 :)

PR #1833, opened 2017-02-10
Update squatter man page to match actual usage.

postilion:
Man page for squatter updated to accord with proper usage, as per @rsto update in 4643639.  Further to #1826

### DIFF
--- a/docsrc/imap/reference/manpages/systemcommands/squatter.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/squatter.rst
@@ -13,10 +13,18 @@ Synopsis
 
 .. parsed-literal::
 
-    **squatter** [ **-C** *config-file* ] [ **-r** ] [ **-s** ] [ **-i** ] [ **-a** ] [ **-v** ] *mailbox*...
-    **squatter** [ **-C** *config-file* ] [ **-r** ] [ **-s** ] [ **-i** ] [ **-a** ] [ **-v** ]  **-u** *user*...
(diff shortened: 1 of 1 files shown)


hagedose:
I'm surprised that Sphinx is mentioned explicitly and Xapian not at all. It is my understanding that Sphinx support is basically deprecated and Xapian is expected to be the main search engine going forward, but I can't find a source offhand. Regardless, Xapian should probably be mentioned alongside Sphinx in the "NOTE:" section.

postilion:
Not sure what to say about this.  Sphinx support was added in v3.0 and is still there.  The notes in the man page were put there by whomever handled the old squatter.8 man page from before March of 2015 (which is when it was first converted to rst).  As for deprecating Sphinx, while Bron's post to the Xapian Discuss mailing list from 10/2014 says:

> at first we were looking at embedding Sphinx, but the single server model worried us.

but nobody has ever ripped out the code for Sphinx, which lives on to this day.

Oh, wait, I've just found this from Bron from cyrus-info, 9/2014:

>  So Sphinx was just too IO intensive, we had to ditch it entirely, but we didn't kill the code. It's probably stale though - I wouldn't use it without doing a ton of testing.

So it looks like we could remove the Sphinx notes, but unfortunately, I'm not sure what the nature of any "notes" for Xapian would be, as I haven't yet used it, and there aren't really any notes-worthy items in the few sources I can find.

If anyone can clue me into notes-worthy content, I'll be happy to add them.  In the meantime, I'll look at excising the Sphinx stuff.

hagedose:
Thanks. I can't really help with advice on Xapian, because I'm a novice myself, but I'm pretty sure that there _should_ be more information ...

Issue #1837, opened 2017-02-14
cyr_virusscan does not have a man page


hagedose:
That's all.

postilion:
@elliefm I assume this will need to be added to the top-level Makefile, but I leave that to you.

postilion:
@elliefm, my latest commit includes the Makefile change.  Also added examples to man page.

PR #1838, opened 2017-02-14
Docedit cyr virusscan

postilion:
Address lack of man page for cyr_virusscan, as identified by @hagedose in #1837

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1500,6 +1500,7 @@ dist_man8_MANS = \
 	man/cyr_expire.8 \
 	man/cyr_info.8 \
 	man/cyr_synclog.8 \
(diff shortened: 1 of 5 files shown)


elliefm:
That looks great!  Thanks :)

Though, "was introduced in 3.0":  I think cyr_virusscan has existed for a long time (the segfault I fixed the other day appeared to have been introduced in _2012_) -- but I have no idea if it has ever worked.  

It's in 2.5, but probably doesn't work (but I expect I'll backport the recent fixes before the next release from that series).  

So I'm not really sure what to say on when it was introduced.  Will merge as is, and we can refine it later I guess :)

PR #1842, opened 2017-02-15
Move the REID operation from cyr_info to mbtool

postilion:
It's always struck me as odd that re-IDing a mailbox is a job for
cyr_info, rather than for mbtool.  From their respective man pages:

- **cyr_info** is a tool for getting information from Cyrus.
- **mbtool** is a tool for performing various actions on the indexes of a list of mailboxes.

Wouldn't the "reid" action then fall under the purview of mbtool?  Granted it's not an "index" operation, however, but that's just semantics.

It looked to me as though it should be pretty easy to move it, so I've tried just that.  Man pages are updated, too.

**Note:** I am by no means a C programmer, so please treat as a proof of concept, not final goods.  However, I have tested it and it does work as expected.

### DIFF
--- a/docsrc/imap/reference/manpages/systemcommands/cyr_info.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/cyr_info.rst
@@ -18,7 +18,6 @@ Synopsis
     **cyr_info** [OPTIONS] conf-all
     **cyr_info** [OPTIONS] conf-lint
     **cyr_info** [OPTIONS] proc
-    **cyr_info** [OPTIONS] reid *mailbox*
 
 Description
 ===========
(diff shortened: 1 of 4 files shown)


elliefm:
This looks okay to me, am happy to merge it.  @brong, @ksmurchison, any objection to this conceptually?  Is there underlying reasoning behind the way things currently are that I'm missing?

ksmurchison:
No objection here.  I don't recall how this functionality ended up in cyr_info.c

brong:
Probably just so I could dump it somewhere because I needed it stat.  Yes, mbtool is a much saner place.  Thanks,

PR #1856, opened 2017-03-07
Update cyrus.conf(5) man page and sample

postilion:
Add information about new DAEMON section to man page and sample
for cyrus.conf(5) file.


elliefm:
Awesome, thanks! :)

PR #1863, opened 2017-03-14
Docedit/dav

postilion:
This is a first pass at documenting CardDAV/CalDAV support in Cyrus.  This stuff all falls under the Concepts section, in Features.  It's a start, but a bare one.  Some pages are mere place-holders.


elliefm:
Thanks! :D

PR #1864, opened 2017-03-14
Replace empty man page for cyr_userseen

postilion:
Create a man page for the cyr_userseen command.

### DIFF
--- a/docsrc/conf.py
+++ b/docsrc/conf.py
@@ -439,6 +439,17 @@ man_pages = [
     ),
 
         (
+            'imap/reference/manpages/systemcommands/cyr_userseen',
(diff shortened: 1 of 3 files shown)


elliefm:
I feel bad being so pedantic but there's a typo that really jumped out at me!

>    This command *does not* take a **constraining** argument but acts on the

You should be able to amend the existing commit on your end, and then push it up with the -f argument, to keep the history nice.  Happy to help in more detail if you need it -- hit me up on IRC if I'm online when you are, or ask here. :)

postilion:
@elliefm here it is.  Thanks for the proof reading! :-)

PR #1884, opened 2017-04-05
Update packagers.rst

postilion:
Updated the packagers.rst file with sample configurations, new features
and other guidance to packagers.


elliefm:
Awesome, thanks!

One teeny correction:

> +A new section, DAEMON, was added to :cyrusman:`cyrus.conf(5)` in
> +version 3.0.  Please consult :cyrusman:`cyrus.conf(5)` for details.

This was added in 2.5(!) :)

postilion:
@elliefm I've got a bunch more changes to this, so please don't confirm it until later today.  Just wanted to get it started.  I'll fix the 2.5-ism

elliefm:
Sure, no worries, let me know when you're ready for it to merge

postilion:
@elliefm I'm going to bundle the new cyrus.conf samples in with this stuff.  Here's a question on that; should we continue to carry along the CMU examples?  Do they pertain any more?

elliefm:
@ksmurchison do CMU need us to keep bundling `master/conf/cmu-backend.conf` and `master/conf/cmu-frontend.conf`?

@postilion I guess leave those two files where they are for now, and we can pull them out later once we know the story

postilion:
@elliefm Just so I'm sure about this; when do programs listed in DAEMON get started?  Will any files in tmpfs, such as deliver.db, have been created already?  I ask because this is not the case with START.

elliefm:
Entries in the START section are run sequentially, more or less as the cyrus.conf is being parsed.  Nothing continues until the START section has finished running.  (But: idled, traditionally invoked from START, will fork and background itself, allowing master to finishing starting.  There might be some other things that behave similarly, I'm not sure.)

Entries in the SERVICES, DAEMON and EVENTS sections are all run from master's main spawn/babysit loop.

I think deliver.db et al get initialised when the first cyrus process that needs databases starts up (so, something from the SERVICES section generally).  Which I guess is why they don't yet necessarily exist during START.  (It's the `cyrus_init(...)` call that provokes this initialisation, which all the "real" services do, but I dunno what other things are out there.)

postilion:
Great, I'll make note of that in the docs.
Just pushed more updates to packagers.rst, and the cyrus.conf files, plus several new ones.  

@elliefm @nicolan Now we need to update makefiles to get these samples put into the right place, or should we just move them in the source?

elliefm:
Just move them in the source, that way they'll be in the same location whether your source tree is from a release tarball or from git.  Everything under `doc/` gets automatically included in release tarballs, so it shouldn't require any Makefile edits at all (but once it's moved, I'll test that to confirm, and tweak if necessary)

postilion:
@elliefm Okay, my latest push moves these files.  Since, as you've noted, there's a need to distinguish what these are samples for, I've created doc/examples/cyrus_conf and doc/examples/imapd_conf directories, and moved these samples into the former.  Next up I'll whip up some sample imapd.conf files, and try to think of examples which show things like service-specific option prefixes, etc.

postilion:
@elliefm Let's go ahead and merge this as it sits.  I'd like to hear some feedback from packagers and admins before going much further.

PR #1888, opened 2017-04-06
Update admin page, new port & socket info, etc.

postilion:
Updated admin page with new section for ports & sockets.  Added a page
for each of those, intended to be "incuded" into other pages as needed.
Updated configs/cyrus.conf.rst to correct typos.

Update basicserver.rst to include info from other pages.


elliefm:
Are you able to squash out the merge commit?

For what it's worth, generally when master has moved forward underneath your feature branch, it's a bit cleaner to rebase your branch on top of master, rather than merging master back into it.  It makes the history a bit easier to follow later on if archaeological digs are needed

postilion:
Yeah, I got caught up in a git-stuidity loop on that one.  My tree is off in a different direction right now, but I'll check back on that later today.

PR #1890, opened 2017-04-06
Correctly calculate INBOX for user names containing a dot character

### LINKED ISSUE #1875: listmailbox returns empty list for users with dot in their name
In cyrus-imapd 3.0, listmailbox does not show any mailboxes for users with dot in their name.

mmatuska:
Fixes #1875


brong:
This is looking good per the 4 new Cassandane tests I've added to test it with all combinations altnamespace on and off, and virtdomains on and off.

PR #1917, opened 2017-04-11
Doc/sample imapd conf

postilion:
Add a bunch of new configuration samples as well as some real content for the Quickstart guide.


docsrc/quickstart.rst
```diff
@@ -20,69 +20,189 @@ your OS distribution of choice and their packaging, where available.  If
...
-
-    * squat:              yes
-    * sphinx:             no
-    * xapian:             yes
-    * xapian_flavor:      vanilla
-
-Installation directories:
-    * prefix:             /usr
-    * sysconfdir:         /etc
++---------------------+-------------+
```
elliefm: The plain text formatter isn't able to handle tables with spanned cells, so `make doc` fails like this:

```
Exception occurred:
  File "/usr/lib/python2.7/dist-packages/sphinx/writers/text.py", line 505, in visit_entry
    raise NotImplementedError('Column or row spanning cells are '
NotImplementedError: Column or row spanning cells are not implemented.
```

I'm not sure if we've tripped over this before or not, maybe @nicolan knows a workaround?

doc/examples/imapd_conf/.gitignore
```diff
@@ -0,0 +1 @@
+imapd.conf.sample
```
elliefm: I'd prefer if this was in the top-level .gitignore rather than buried down in a subdirectory -- makes it easier to see all the ignored files in one place -- but it's not a big deal.  Looking around, I see there's a few other directory-specific .gitignore files lying around, so I might just do a big cleanup of all of them sometime.

docsrc/quickstart.rst
postilion: @elliefm I'll change this to a list instead.  It works fine on my laptop, so maybe I've got a different version of Sphinx or something like that.

doc/examples/imapd_conf/.gitignore
postilion: @elliefm I'll leave this be and let you effect some master plan.

elliefm:
FYI, I added 062e9c5 to ensure imapd.conf.sample gets included in releases: even though the whole `doc/` tree is automatically included, its contents are calculated quite early, before this file is generated.  The solution: hook onto the `doc:` rule to ensure that it gets generated first. :)

postilion:
@elliefm Thanks! :-)

PR #1923, opened 2017-04-13
Change int with uint64_t to avoid overflow in print_…

landgraf:
…stats

If there're mailboxes >2GB int it leads int to overflow.

Bug-Url: https://bugzilla.redhat.com/1212525
Signed-off-by: Pavel Zhukov <pzhukov@redhat.com>


brong:
clearly fine :)

brong:
Ok, not QUITE fine... I had to add explicit casts, because uint64_t and (long long unsigned) aren't necessarily considered identical by compilers.

Issue #1937, opened 2017-04-24
Docs: Add note to upgrade docs about increased memory usage for upgrades from 2.3


nicolan:
For those upgrading from 2.3, it is worth noting that by design, cyrus uses more memory.

Selected mailboxes use a lot more memory intentionally to get faster performance and more accurate standards compliance

postilion:
@nicolan I'm not quite sure where to put this.  Sounds like something that should be in a release note, but at what point was this change introduced?  Given what @brong said on conference call, this is a delta-v2.3 issue, so should be covered in 2.4 or 2.5 docs.

I can craft language to explain it, but I just don't know where that language should go.

nicolan:
Rats. I was hoping you'd have a magical cunning plan because I had the same set of thoughts with no clear conclusion. 

I guess it needs to go in the 2.4, 2.5 docs. And in the 3.0 release notes and in the upgrade guide as  a few folks will want to just go straight to the latest. (which says "asset" to me)

elliefm:
I'm reluctant to modify release notes after the fact too much, because it feels like the sort of thing people expect to not change over time...

That said, I think the 2.5.0 (specifically) release notes had an upgrade guide of sorts (all the subsequent 2.5.x release notes refer back to it), so maybe it could be mentioned in there.  And also in the 3.0 upgrade guide.

We can't really put this into 2.4 docs without doing a new 2.4.x release, since we don't publish 2.4 docs from sphinx.  I don't really expect this to be a concern, because in 2017 I don't really expect people to be upgrading _to_ 2.4.x...

PR #1939, opened 2017-04-27
Add note explaining increased memory footprint.

postilion:
This is added to 2.5.0 release notes and 3.0 upgrade docs.  A new
"asset" is added.  This will require that 2.5.0 have new asset
directory, too.

Addresses issue #1937

### DIFF
--- a/docsrc/assets/cyrus-more-memory-post23.rst
+++ b/docsrc/assets/cyrus-more-memory-post23.rst
@@ -1,0 +1,5 @@
+For those upgrading from 2.3.X; newer releases of Cyrus IMAP will use
+significantly more memory per selected mailbox.  This is not an error
+or bug; it's a feature.  The newer code is holding more data and
+metadata in memory for purposes of faster access to more of the
+mailbox.  This is not a memory leak.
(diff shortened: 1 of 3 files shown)


postilion:
@elliefm Even though I now have the power, I'd prefer if you review this, as I am unsure what ramifications are for the 2.5.0 stuff.  Will this require back-porting to 2.5, or just what?  Enquiring minds want to know... :-)

elliefm:
This will need to be backported to both the cyrus-imapd-3.0 and cyrus-imapd-2.5 branches (since, at least at the moment, all versions have all versions' release notes).

The backport to cyrus-imapd-3.0 should be fairly trivial, since everything's basically the same in the docsrc tree already anyway (and the entire sphinx-documentation concept has been 3+ all along)

Offhand, I'm not sure what it will be like to backport to 2.5, since the sphinx-documentation was kind of wedged into there backwards and the world is a bit different there.  I'm sure it won't be too hard, but it might not be seamless :)

elliefm:
Cherry-picked to cyrus-imapd-3.0 as 838a0e3, and to cyrus-imapd-2.5 with massaging as 589f3c6

Issue #1928, opened 2017-04-20
Improve replication docs: periodic vs rolling


nicolan:
Useful information came in over #cyrus in IRC.

Review current documentation on failover and improve based on the info below!

<elegast> just to be clear though: the differences between 'rolling' and 'periodic' replication is that in both instances *periodically* the master syncs, except with 'periodic', it's cron that triggers sync_client and otherwise imapd does every few intervals?
<elliefm_> with periodic replication, you schedule sync_client to run a single sync regularly from cron, and it replicates whatever you tell it to (which might be some specific mailboxes or might be the world, see options to sync_client)
<elliefm_> with rolling replication, all of the daemons that make changes to state (imapd, lmtpd, etc) will append a line to a channel's sync log for each action they perform
<elegast> oh ok, theres a clear difference there.
<elliefm_> sync_client wakes up, reads the current sync_log, syncs anything that the log says needs syncing, and then goes to sleep for a bit, forever (or until you shut it down)
<elliefm_> (there's a bunch of smarts in there for atomically handling the sync log files that i'm glossing over for brevity's sake)
<elegast> hmm, yeh, now im confused again
<elegast> I still get the impression the only difference is actually how sync_client is invocated?
<elegast> (and when)
<elliefm_> note very carefully that the sync log doesn't contain *details* of what changed (like, it's not "add this message with these properties to this mailbox), it's just a list of things that have changed (like "this mailbox"), and then sync_client works out the differences in the same way as if you'd done a one shot "sync this mailbox"
<elliefm_> more or less
<elliefm_> rolling replication requires you to configure sync logging and at least one channel (see the sync_ options in imapd.conf)
<elegast> ohh..
<elliefm_> whereas periodic replication you can provide every option it needs with command line switches
<elliefm_> and then, with period replication, you either schedule sync_client with cron, or stick it in the events section of cyrus.conf
<elegast> I thought it was exactly that, the actual data to replicate
<elegast> ok now the difference makes sense.
<elliefm_> in a way, the sync log is just a "now run sync client on this mailbox", "now run sync client on this user", "now run sync client on this other mailbox", with the difference that the sync_client process keeps running rather than stopping and starting 
<elliefm_> if you were a masochist you could do the same thing by hand by watching your imap logs and running sync client by hand every time you see something do something that would need syncing :P
<elegast> Mhmm. I can be! But I don't like to do things when there's processes that do it for me. Thanks ellie!
[...]

PR #1945, opened 2017-04-28
Updated replication.rst for more completeness.

postilion:
This is to address issue #1928.

Note: This shouldn't be in SOP.  There must be a better place...


elliefm:
Cherry-picked to 3.0 too :)

PR #1963, opened 2017-05-09
Fix crash with cyrus-sasl-2.1.26

felfert:
See my mail at cyrus-devel ML:
https://lists.andrew.cmu.edu/pipermail/cyrus-devel/2017-May/004063.html


ksmurchison:
cyrus-imap master (3.0.x) works fine with cyrus-sasl master (soon-to-be 2.1.27)

The issue with the digest reauth cache had already been fixed in git.
I still don't see how/why setting maxbufsize to zero causes an issue.  Perhaps this also was fixed in SASL previously

PR #1969, opened 2017-05-14
Makefile changes

jsoref:
This is split from #73.
Please don't rely too much on the actual changes, as they're my sloppy attempt at splitting things out/rebasing. I can't explain any of the changes, as they aren't mine (I'm not claiming authorship).


rsto:
The conflicting C file changes all are addressed upstream already. Rejecting the Makefile changes since they seem to not fix actual build issues. Otherwise, please document the environment on which build issues still occur.

PR #1973, opened 2017-05-17
Bug 1972: mboxevents feature cannot be disabled.

### LINKED ISSUE #1972: make mboxevents mandatory to compile in
At the moment you can compile without mboxevents, which means massive ifdef statements and we don't really test without them compiled in.

We should just strip all that so that we always compile mboxevents in and simplify the configuration job.

ajaysusarla:
This commit makes the mboxevents feature mandatory for Cyrus, thus
removing the ability to disable it when building(and configuring).
Also makes the necessary documentation updates.

Fixes #1972.


nicolan:
Can you also add a commit to update the release notes in docsrc/imap/download/release-notes/3.0/x/3.0.2.rst ?  (Follow the convention in the 2.5.n release notes for incremental releases)

Thanks!

ajaysusarla:
@nicolan Done!

PR #1974, opened 2017-05-18
Remove references to `librt`.

ajaysusarla:
We don't seem to use `librt` anymore, so removing references to it from
the repo.


elliefm:
The main references to LIB_RT were removed by 780d6c878c525b1a355c7ee24d5320a1cc202efc.  We still need to bring in librt for some platforms (as per the issue referenced in that commit).

Looks like I missed tiying the perl module makefiles, doh.  I'm not sure if they actually need librt or not

Issue #1965, opened 2017-05-09
man cyradm: document getmd and setmd


dilyanpalauzov:
perl/imap/cyradm.sh does not document setmetadata and getmetadata in 2.5.  Hence it does not state in particular, that special-use flags have to be set with two leading backslashes.

elliefm:
@rsto do you want to tackle this while you're working on the annotations code here anyway?

rsto:
sorry for late reply: i'm not sure what to make of this. we'll deprecate ANNOTATEMORE in 3.1 but i had no plan to backport this to 2.5.

is this purely a matter of documenting setmetadata/getmetadata in cyradm for 2.5 or is there code that needs fixing?

rsto:
Following up an email from @dilyanpalauzov: the Perl doc in cyradm.sh does not document the getmd/setmd commands, especially not the format of special-use flags. That's across all Cyrus branches.

While documenting that we might want to check if there are any other discrepancies between the interactive help of cyradm and it's Perl doc:

    > help
    authenticate, login, auth         authenticate to server
    chdir, cd                         change current directory
    createmailbox, create, cm         create mailbox
    deleteaclmailbox, dam, deleteacl  remove ACLs from mailbox
    deletemailbox, dm, delete         delete mailbox
    disconnect, disc                  disconnect from current server
    exit, quit                        exit cyradm
    getmetadata, getmd                display mailbox/server metadata
    help, ?                           show commands
    info                              display mailbox/server annotations
    listacl, lam, listaclmailbox      list ACLs on mailbox
    listmailbox, lm                   list mailboxes
    listquota, lq                     list quotas on specified root
    listquotaroot, lqr, lqm           show quota roots and quotas for mailbox
    mboxcfg, mboxconfig               configure mailbox
    reconstruct                       reconstruct mailbox (if supported)
    renamemailbox, rename, renm       rename (and optionally relocate) mailbox
    server, connect, servername       show current server or connect to server
    setaclmailbox, setacl, sam        set ACLs on mailbox
    setinfo                           set server metadata
    setmetadata, setmd                set metadata to mailbox
    setquota, sq                      set quota on mailbox or resource
    subscribe, sub                    subscribe to a mailbox
    unsubscribe, unsub                unsubscribe from a mailbox
    version, ver                      display version info of current server
    xfermailbox, xfer                 transfer (relocate) a mailbox to a different server

rsto:
[...]

PR #1977, opened 2017-05-18
Updated cyradm and other Perl docs for get/setmd

postilion:
Added examples and notes on use of getmetadata and setmetadata
commands in cyradm.


postilion:
Closes #1965

PR #1982, opened 2017-05-22
Option to build Cyrus statically. 

ajaysusarla:
This patch provides configure option to enable static builds. (Actually
it just uses the `--enable-static` and `--enable-shared` options
provided by `autoconf`). When configured statically, we ensure that we
link with `libstdc++` which is a dependency since, we link with
`libxapian`.

What does it actually mean to build Cyrus statically?
Only the binaries and libraries we build(`libcyrus_*) as a part of
`cyus-imapd` are built and linked statically. Only static archives(`.a`)
are built. All the external libraries are still dynamically linked.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -55,6 +55,13 @@ AC_CONFIG_HEADERS(config.h)
 AC_CONFIG_MACRO_DIR([cmulocal])
 AC_CANONICAL_SYSTEM
 
+dnl shared or static
+ENABLE_SHARED="$enable_shared"
+AC_SUBST(ENABLE_SHARED)
+
+ENABLE_STATIC="$enable_static"
+AC_SUBST(ENABLE_STATIC)
+
 dnl carve the version string (defined in AC_INIT above) into useful chunks
 dnl e.g. for version 3.0.0-beta:
 dnl   CYRUS_IMAPD_SERIES   => 3.0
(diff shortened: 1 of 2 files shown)


elliefm:
I feel like there should be a way for this to be taken care of automatically, rather than crudely adding stuff direct to $LIBS...

I just had a look at the xapian-config-1.5 on my system, and it takes an optional --static argument to make the other arguments return the correct invocation for static linking rather than dynamic linking, so this might be the way to go.

I would suggest doing something in the `if test "x$enable_xapian" != xno ; then` block of configure.ac. Maybe set up a $XAPIAN_LINKAGE variable that's just an empty string for normal shared builds, but is set to "--static" if it's doing a static build.  And then make all of the calls to $XAPIAN_CONFIG pass $XAPIAN_LINKAGE as an argument.

No idea if this will actually work or not (it assumes that xapian-config-1.5 reports the correct linkages, and I don't know it well enough to trust it yet), but if it does work, this will be a lot cleaner. :)

ajaysusarla:
Thanks for the review @elliefm. I agree, that manipulating `LIBS` is ugly. Although, what you are suggesting is to be able to link `xapian` statically to Cyrus, which isn't what we want to do. We want to continue to link to dynamically to all the external libs and only statically link to all the `libcyrus` stuff.

Since the linker/loader need to resolve symbols(cpp source pertaining to libxapian we have in our repo), i have to explicitly link to `libstdc++` now. Earlier, when linked dynamically, they used to get resolved during loading time.

elliefm:
Hmm, good point

PR #1998, opened 2017-05-31
[Bug 1971] New and Improved(tm) "tell exists count".

### LINKED ISSUE #1971: Rewrite the "tell EXISTS count earlier" code to do single-pass inline unsolicited FETCHes for index_fetch
The work I did on Friday(In [Cyrus](https://github.com/cyrusimap/cyrus-imapd/commit/c3141cc3b7b7c33596827a1ec1156663212376e6) & [Cassandane](https://github.com/cyrusimap/cassandane/commit/f1a2f49776c6b9d0c05ed69aa1f0796240d4609e) ) spits out the EXISTS count early enough, so that's all good - but it could be made slightly more efficient/simple.

index_fetch has some efficiency hacks that if a range of UIDs or MSGNOs is specified, it will find the start and end recno and only iterate over that range checking for set membership before fetching.  Good efficiency, but if the modseq has changed, we then do index_tellchanges(), which will iterate the entire mailbox anyway!

When we do any fetch on a record, we check the told_modseq and if it's older than the record modseq, we also tell FLAGS and update the told_modseq, so that index_tellchanges doesn't tell it again.

An alternative, more efficient way to do this for a FETCH would be to check if the modseq has changed since last fetch (we know that anyway with index_refresh_locked having been called), and if so then scan the entire mailbox rather than limiting the range, and check every record in order.

And always do the EXISTS up-front if it has changed, we can use the test from index_tellchanges still.

Test case:

create 3 messages
fetch in second session
mark second message \Seen
append a new message
fetch 3:* (UID FLAGS) in second session.

Current behaviour would be:

TAG FETCH 3:* (UID FLAGS)
* 3 FETCH (UID 3 FLAGS ())
* 4 EXISTS
* 1 RECENT
* 4 FETCH (UID 4 FLAGS (\Recent))
* 2 FETCH (FLAGS (\Seen))
TAG OK Completed

Wanted behaviour is:

TAG FETCH 3:* (UID FLAGS)
* 4 EXISTS
* 1 RECENT
* 2 FETCH (FLAGS (\Seen))
* 3 FETCH (UID 3 FLAGS ())
* 4 FETCH (UID 4 FLAGS (\Recent))
TAG OK Completed

NOTE IMPORTANTLY: the unsolicited fetch of msgno 2 does NOT include a UID unless QRESYNC is enabled, this is for annoying legacy broken client reasons :(

[@ajaysusarla update the comment with commit id's]

ajaysusarla:
This patch improves the `index_fetch` code to do single-pass inline
unsolicited `FETCH`es.

Fixes #1971.

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -786,6 +786,7 @@ static void index_refresh_locked(struct index_state *state)
     state->oldexists = state->exists; /* we last knew about this many */
     state->exists = msgno - 1; /* we actually got this many */
     state->delayed_modseq = delayed_modseq;
+    state->oldhighestmodseq = state->highestmodseq;
     state->highestmodseq = mailbox->i.highestmodseq;
     state->generation = mailbox->i.generation_no;
     state->uidvalidity = mailbox->i.uidvalidity;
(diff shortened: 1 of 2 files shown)


ajaysusarla:
With the changes I have made, what looked like this **earlier**(on **master**):
```
C: 9 fetch 3:* (uid flags)
S: * 3 FETCH (FLAGS () UID 3)
S: * 4 EXISTS
S: * 1 RECENT
S: * 4 FETCH (FLAGS (\Recent) UID 4)
S: * 2 FETCH (FLAGS (\Flagged))
```
will look like this;
```
C: 9 fetch 3:* (uid flags)
S: * 4 EXISTS
S: * 1 RECENT
S: * 2 FETCH (FLAGS (\Flagged))
S: * 3 FETCH (FLAGS () UID 3)
S: * 4 FETCH (FLAGS (\Recent) UID 4)
```

@brong this might not be the right way to fix this issue. Please do let me know if I'm heading the right direction with this patch.

The test is available in this pull request: https://github.com/cyrusimap/cassandane/pull/22

PR #2000, opened 2017-06-01
jmap_mail.c: Add missing format specifier.

ajaysusarla:
The `fprintf()` is missing a format specifier. This patch makes sure we
add one with CRLF.

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -4728,8 +4728,10 @@ static int writeattach(jmap_req_t *req, json_t *att, const char *boundary, FILE 
     else {
         content_type = xstrdup("message/rfc822");
     }
-    fprintf(out, "Content-Type: ");
-    fprintf(out, content_type);
+
+    fwrite("Content-Type: ", 1, strlen("Content-Type: "), out);
+    fwrite(content_type, 1, strlen(content_type), out);
+
     if (name) {
         /* RFC 2045 dropped the "name" parameter value for Content-Type,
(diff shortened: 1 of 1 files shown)


rsto:
The reason the fprintf statement omits CRLF is because we might need to write additional parameters in the following ~20 lines. The final CRLF sequence to end this header is written out unconditionally at https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/jmap_mail.c#L4729

I don't see how the current code would break things, did you run into an issue with the current code?

ajaysusarla:
@rsto That makes sense. So the CRLF can go. But I still think, we should have the format specifier in there, like so:
```
fprintf(out, "%s", content_type);
```
. Nothing broke. I was just going through the code, and thought it made sense. I will close this pull request if you think it isn't necessary.

rsto:
@ajaysusarla ah, I totally missed that! Yes, that's silly. Thanks.

Would you care for changing this to a fwrite then? Otherwise, I'll upload a patch in the course of today.

ajaysusarla:
@rsto updated the pull request. Thank you.

rsto:
Thanks. Could you please change that to `fwrite` rather than `fprintf`? It's a minor thing, but we try to avoid printf/buf_printf when there's no real need for it.

ajaysusarla:
@rsto update the patch, based on your recommendation.

rsto:
LGTM, thanks

PR #2022, opened 2017-06-15
cyr_expire: Cleanup.

ajaysusarla:
This patch does a fair bit of tidying up `cyr_expire`:
* Add a wrapper functino for verbose printing.
* Cleanup in signal handler (avoid checking if singal has been handled)
* Refactor the argument parsing, initialisation and cleanup code.

### DIFF
--- a/imap/cyr_expire.c
+++ b/imap/cyr_expire.c
@@ -52,6 +52,7 @@
 #include <sys/stat.h>
 #include <syslog.h>
 #include <signal.h>
+#include <errno.h>
 
 #include <sasl/sasl.h>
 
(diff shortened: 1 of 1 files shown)


elliefm:
I'm not comfortable with the change to calling `cyr_expire_cleanup()` from the signal handler -- I don't know how much of the `free_hash_table()` and `foo_done()` calls are signal safe.  Probably not much, since everything else just uses the signal handler to set a flag and handles its shutdown process outside the signal handler, so they're probably written with this assumption in mind.

I like the `fprint_verbose` implementation (we should use something like this much more widely), but one little nit about the name: the first f in "fprintf" refers to the first argument being a `FILE *`, and the second f is what refers to the format string, so for strict consistency, this would be more correctly called `printf_verbose` (since there is a format string, but no `FILE *`).  But tbh, I'd probably just call this `verbosef` instead -- for both brevity, and clarity that it takes a format string argument.  Brevity is good here: format string arguments often get pretty long, and our line lengths are nominally < 80 characters.  This would be a good function to move into `lib/util.h` down the track.

And a style nit: the switch statement in the `parse_args` has the `case '...'` lines indented one level deeper than the `switch (opt) {` line.  For whatever historical reason, our code style places these at the same level of indentation. :)  There's some info on our code style in `doc/internal/hacking` in the repo (though it's probably out of date... worth a read, but use your best judgement) :)

ajaysusarla:
@elliefm Thanks for your review and feedback. I've made the changes you've suggested. 
With regards to the `fprint_verbose`/`print_verbose`/`verbosep` function, I personally prefer the `print_verbose`. I've changed it to `verbosep` though, to signify verbose printing, and the function could print to `FILE *` or use `printf` as we chose(currently printing to `FILE *`).
I agree with you about moving it to `/lib`. I shall make it generic and move it eventually.  I've updated the patch (waiting for the `green tick`), please review and merge when you deem appropriate.

PR #2030, opened 2017-06-19
[Bug 1988] Add per folder annotation for cyr_expire

ajaysusarla:
`cyr_expire` already has annotation for `expire`, this patch adds support for
`archive` and `delete` operations.

### DIFF
--- a/imap/cyr_expire.c
+++ b/imap/cyr_expire.c
@@ -40,6 +40,14 @@
  * OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
+/*
+  NOTES:
+  * Precedence of configuration in `cyr_expire`:
+     Highest                       Lowest
+     |----------------------------------|
+     Annotation -> Command-line -> Config
+ */
+
 #include <config.h>
 
 #ifdef HAVE_UNISTD_H
@@ -53,6 +61,7 @@
 #include <syslog.h>
 #include <signal.h>
 #include <errno.h>
+#include <stdbool.h>
 
 #include <sasl/sasl.h>
 
@@ -89,15 +98,22 @@ struct arguments {
     int expire_seconds;
     int expunge_seconds;
 
-    /* bools */
-    int do_expunge;
     int do_cid_expire;
-    int do_userflags;
-    int skip_annotate;
+
+    /* bools */
+    bool do_expunge;
+    bool do_userflags;
+    bool skip_annotate;
 
     const char *altconfig;
     const char *mbox_prefix;
     const char *userid;
+};
+
+struct archive_rock {
+    time_t archive_mark;
+    unsigned long messages_archived;
+    bool skip_annotate;
 };
 
 struct expire_rock {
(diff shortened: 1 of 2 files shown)


imap/cyr_expire.c
```diff
@@ -273,6 +287,44 @@ static int parse_duration(const char *s, int *secondsp)
...
+    do {
+        buf_free(&attrib);
+        ret = annotatemore_lookup(buf, annot_entry, "", &attrib);
+        if (ret ||              /* error */
+            attrib.s)           /* found an entry */
+            break;
+    } while (mboxname_make_parent(buf) && iterate);
+
+    if (attrib.s && parse_duration(attrib.s, secondsp))
+        ret = 0;
```
elliefm: the truthiness of this function seems to be the inverse of `parse_duration()`?  `parse_duration()` returns 1 on success, 0 on failure, but this returns 0 on success, 1 on failure?

imap/cyr_expire.c
```diff
@@ -326,9 +380,16 @@ static int archive(const mbentry_t *mbentry, void *rock)
     if (mailbox_open_iwl(mbentry->name, &mailbox))
         goto done;
 
-    verbosep("archiving mailbox %s\n", mbentry->name);
+    if (!arock->skip_annotate &&
+        get_annotation_value(mbentry->name, IMAP_ANNOT_NS "archive",
```
elliefm: ... but then this usage of `get_annotation_value()` seems to expect non-zero on success, the same as `parse_duration()`

imap/cyr_expire.c
```diff
@@ -650,22 +849,20 @@ static int parse_args(int argc, char *argv[], struct arguments *args)
 int main(int argc, char *argv[])
 {
     int r = 0;
-    struct arguments args;
+    struct cyr_expire_ctx ctx;
```
elliefm: i would pre-initialise this like `struct cyr_expire_ctx ctx = {0};` rather than memsetting it later (though it probably makes no difference with optimisations)

ajaysusarla: I did want to do that (assigning to `{0}` is my preferred method to initialise, but didn't see it being used in cyrus, so chose the `memset` way :). More than happy to change!

imap/cyr_expire.c
ajaysusarla: ack! I seem to have messed it up in translation. Let me double check.

imap/cyr_expire.c
ajaysusarla: @elliefm `gcc` on travis doesn't accept struct intialisations with `{0}`, so I'm continuing to use `memset` for now.

ajaysusarla:
@elliefm I think I've made all the changes and I think is complete. This is ready for review again. Please review when you can.

elliefm:
The `Metadata.private` and `Metadata.shared` cassandane tests fail with this patch applied, looks like they're now finding the "delete" annotation where they previously didn't expect to?

The way to handle this will probably be to do a version check within these two tests, and if it's >= 3.1.0, then add the "delete" annotation to the expected results hash.  Have a look for code using `get_version()` for examples of tests that do this sort of thing :)

Looks okay otherwise!

PR #2046, opened 2017-07-03
[Docs] Generate google sitemap.xml

nicolan:
Add new extension sphinxlocal.sitemap to conf.py
Add new config value sitemap_website to conf.py

Produces sitemap.xml in build/<target>/
where <target> is either man or html, depending on output format
being generated

[Docs] Sitemap generator

Only generate if output is html.
Support multiple webroots.

### DIFF
--- a/docsrc/conf.py
+++ b/docsrc/conf.py
@@ -46,6 +46,13 @@ extensions.append('sphinxlocal.builders.gitstamp')
 
 gitstamp_fmt = "%b %d %Y"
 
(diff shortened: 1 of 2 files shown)


nicolan:
If this passes, it should end up in the right spot on the website automatically.

I made some changes on cyrusimap/cyrusimap.org for a robots.txt and sitemapindex.xml which should support things once they're at cyrusimap.org/ web root.

PR #2051, opened 2017-07-06
[lib] RFC5322 date time parser.

ajaysusarla:
This commit adds support for RFC5322 date time parsing as per this[1].

We should eventually nuke the `time_from_rfc822()` and `time_from_rfc3501()`
functions.

[1] - https://tools.ietf.org/html/rfc5322#section-3.3

### DIFF
--- a/cunit/times.testc
+++ b/cunit/times.testc
@@ -442,4 +442,248 @@ test_leapyear_rfc822(void)
     CU_ASSERT_EQUAL(r, 31);
     CU_ASSERT_EQUAL(t, FEB2004_TIMET);
 }
+
+static void
+test_rfc5322(void)
+{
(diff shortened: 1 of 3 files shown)


brong:
I don't see any reason to do times-private.h - static definitions at the top of a C file when they are only used in that one C file are standard throughout the codebase.  That also mollifies me somewhat about the use of very generic names for the lookup tables, though I'd still prefer them to be a little more grep-friendly!

ajaysusarla:
@brong culled the `times-private.h` file and moved everything to times.c - and renamed lookup tables to, what I think are saner names(prefixed them with `rfc5322_`, so they are now `rfc5322_special`, `rfc5322_usascii_charset` etc.).

Issue #1938, opened 2017-04-27
Fix use of sidebar when pulling in imapd.conf blobs


nicolan:
We use a sphinx sidebar to show imapd.conf entries in line where they're referenced in config/install guides (see jmap).

Except that what we pull in to show a user is the config defaults, which almost always isn't what we need a user to set. This is confusing!

Options:

- Put a note in, at the top of each affected sidebar saying "these are the defaults, you need to adjust based on what's in the text"
- Stop using the sidebar and make it into a deep link into the man page instead, so it's obvious we are showing the config file defaults.
- Change the manpage source, so the format is different, with the default not listed against the config name directly.
- Other?

postilion:
@nicolan Now you're just being petty & mean. :-)
I don't know anything about the use of sidebar, or how it gets setup.  I can see your point, but since the text we're using derives from /lib/imapoptions, and that is _supposed_ to cite the default setting, we cannot change the source.  I think your first option is the best, but I don't know what to change to effect that.  Can you shed some light on "sidebar"?

nicolan:
Fixed in above mentioned commits.

PR #2054, opened 2017-07-07
[Docs] Show that included snippets are just defaults.

nicolan:
GitHub issue #1938


nicolan:
The formatting isn't great, but all the whitespace changes I tried made too much or not enough. This is the best I could manage without it taking another 2 hours of fiddling.

PR #2058, opened 2017-07-10
[Bug 2044] Date header parsing compliance with RFC5322

### LINKED ISSUE #2044: Date header parsing needs to accept everything that RFC5322 does
From a FastMail internal bug

In order for us to store messages in Cyrus, we require that the Date: header be in the form: "Day, d Mon Y H:M:S TZ" (Tue, 27 Jun 2017 14:58:39 +1000). In order to ensure it's in this form, we MAY mangle the email header after we do DKIM verification (and add a verification header), by renaming Date: to X-OriginalDate:, and inserting a new Date: that Cyrus deems valid. This usually just seems to involve adding the day of week to the front. Mangling this header breaks subsequent attempts at DKIM verification.

As per https://tools.ietf.org/html/rfc5322#section-3.3, The day of week is not actually a required component of an RFC5322 Date: header. Ideally Cyrus should support *all* valid Date: header forms, including this one, so we don't need to mangle the header for storage and thus break DKIM verification.

This bug was highlighted by the customer on PTN352545.

ajaysusarla:
Use `time_from_rfc5322()` which complies with the RFC5322.

Fixes #2044

### DIFF
--- a/cunit/times.testc
+++ b/cunit/times.testc
@@ -451,127 +451,127 @@ test_rfc5322(void)
 
     /* 1 Jan 1970 */
     t = UNINIT_TIMET;
-    r = time_from_rfc5322("Thu, 01 Jan 1970 00:00:00  ", &t);
+    r = time_from_rfc5322("Thu, 01 Jan 1970 00:00:00  ", &t, DATETIME_FULL);
     CU_ASSERT_EQUAL(r, 27);
     CU_ASSERT_EQUAL(t, 0);
 
     t = UNINIT_TIMET;
-    r = time_from_rfc5322("Thu, 01 Jan 1970 01:00:00 +0100", &t);
+    r = time_from_rfc5322("Thu, 01 Jan 1970 01:00:00 +0100", &t, DATETIME_FULL);
     CU_ASSERT_EQUAL(r, 31);
     CU_ASSERT_EQUAL(t, 0);
 
     /* Zero Hour */
     t = UNINIT_TIMET;
-    r = time_from_rfc5322("WED, 31 DEC 1969 19:36:29 -0500", &t); /* NYC */
+    r = time_from_rfc5322("WED, 31 DEC 1969 19:36:29 -0500", &t, DATETIME_FULL); /* NYC */
     CU_ASSERT_EQUAL(r, 31);
     CU_ASSERT_EQUAL(t, 2189);
 
     t = UNINIT_TIMET;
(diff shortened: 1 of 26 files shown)


elliefm:
Looking good :)

Something I'm trying to get us doing though, is rather than using boolean arguments (whether `int` or `bool` by type), using small enums with meaningful names.

For example, if you look through the mailbox and mboxname code especially, there's heaps and heaps of this sort of thing:

`some_function(mailbox, /*dothing*/ 0, /*otherthing*/ 1, /*foo*/ 0);` where a bunch of booleans have gotten hard enough to keep track of at some point that we've started making comments explaining what the damn things mean... and if we're typing all that out anyway, might as well make it mean something (and then you don't run into the problem where someone changes the value but not the comment).

So I'd suggest doing something like, instead of:

```
static int tokenise_str_and_create_tm(struct rfc5322dtbuf *buf, struct tm *tm,
                                      int *tz_offset, bool usetime)
```

do:

```
enum datetime_parse_option {
    DATETIME_DATE_ONLY = 0,
    DATETIME_FULL,
};

static int tokenise_str_and_create_tm(struct rfc5322dtbuf *buf, struct tm *tm,
                                      int *tz_offset, enum datetime_parse_option usetime)
```

`usetime` ends up having the same boolean semantics within tokenise_str_and_create_tm (i.e., it's true if we're using the time/tz, false if we're using just the date), but it means we can change all the calls from being like:

```
    /* reader has to look up function definition to figure out what "true" does */
    r = time_from_rfc5322("Thu, 01 Jan 1970 00:00:00  ", &t, true);
    [...] 
    if (time_from_rfc5322(body->date, &record->sentdate, false) < 0)
```

to:

```
    r = time_from_rfc5322("Thu, 01 Jan 1970 00:00:00  ", &t, DATETIME_FULL);
    [...]
    if (time_from_rfc5322(body->date, &record->sentdate, DATETIME_DATE_ONLY) < 0)
```

Which means one less bit of mental context to hold onto while reading :)

ajaysusarla:
@elliefm Thanks for the review and feedback. I absolutely agree with you regarding the `enum`. I was debating in my head if it is better of to `#define` the function to something. But I personally don't prefer `#defines` as it hinders with debugging. 
I do love the `enum` idea. I shall update the patch. I also will use the `enum` names as you've suggested.

ajaysusarla:
@elliefm made the changes, let me know if you are ok with the changes.

PR #2061, opened 2017-07-11
[lib] Remove unused strcasecmp.c

ajaysusarla:
The functions in this file aren't used anywhere in cyrus-imapd.


elliefm:
https://github.com/cyrusimap/cyrus-imapd/blob/c7f1a96bf04244e9ad5e6dd95748ccfa395b9ab4/lib/util.c#L277-L281 uses `strcasecmp()`

ajaysusarla:
@elliefm That `strcasecmp()` is from `libc` (strings.h) and not that one in `cyrus-imapd/lib`.

elliefm:
The strcasecmp.c in lib is a portability version for platforms whose libc doesn't provide those functions.  This is why the `AC_REPLACE_FUNCS(memmove strcasecmp ftruncate strerror posix_fadvise strsep memmem)` macro in configure.ac references it, and the same should apply for the rest of those functions.

ajaysusarla:
I thought so too initially. But we don't even seem to compile `strcasecmp.c` anywhere. (No mention of it in `Makefile.am`).

elliefm:
Yeah, it'll be pulled in automatically by configure (theoretically) if it's needed -- but since our platform doesn't need it, it doesn't happen for us. (Which means, I also have no idea if it works in practice, cause I have no access to a platform that needs it, so I can never see whether it works or breaks! :P)

I'm not sure what platforms in 2017 still don't provide strcasecmp() or strncasecmp(), but I'd say you can safely treat anything in AC_REPLACE_FUNCS as "maybe compiled, even if my platform ignores it, and probably there cause something needed it at some point".

ajaysusarla:
@elliefm in which case, I shall send another patch, which requires header guards for `strcasecmp.c`. Will close this for now. Thanks for your review!

PR #2062, opened 2017-07-13
[imap] Use an enum in jmapmsg_from_body().

ajaysusarla:
`jmapmsg_from_body()` uses an integer for `is_embeded` variable, changing that
to an enum for clarity.

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -119,6 +119,11 @@ jmap_msg_t jmap_mail_messages[] = {
 #define JMAP_INREPLYTO_HEADER "X-JMAP-In-Reply-To"
 #define JMAP_HAS_ATTACHMENT_FLAG "$HasAttachment"
 
+typedef enum MsgType {
+        MSG_IS_ROOT = 0,
+        MSG_IS_ATTACHED = 1,
+} MsgType;
+
 static int _wantprop(hash_table *props, const char *name)
 {
     if (!props) return 1;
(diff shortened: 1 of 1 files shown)


ajaysusarla:
@rsto I might not have named the `enum` appropriately. Please do suggest a different name if you think it isn't appropriate.

rsto:
@ajaysusarla I think the enums are a good idea, but the naming is currently not correct. I don't expect there to be ever more than two possible values: either the message is the top message, or it's an attached message (a mail body of type  `message/rfc822`).

I suggest `MSG_IS_TOP=0` and `MSG_IS_ATTACHED=1`?

There's a bunch of alternatives: a "top" message could also be named the "root" message, whereas an "attached" message is also referred to as "embedded" or "encapsulated" (the latter term is used in the [definition](https://www.w3.org/Protocols/rfc1341/rfc1341.html) of the `message/rfc822` subtype). Do you have a preference?

ajaysusarla:
@rsto Thanks for your feedback. I think `MSG_IS_ROOT = 0` and `MSG_IS_ATTACHED = 1` make sense according to what you say.  Accordingly, our `enum` will be:
```
typedef enum MsgBodyType {
        MSG_IS_ROOT = 0,
        MSG_IS_ATTACHED = 1,
} MsgBodyType;
```
Do you think the `enum` itself should be named as `MsgType` instead of `MsgBodyType`? According to the definition you linked to, it seems it should be `MsgContentType`.

rsto:
I'd prefer `MsgType`, content type could be a lot of things.

ajaysusarla:
@rsto Done! I've updated the patch, please review when possible. Thanks again!

rsto:
thanks! lgtm

PR #2063, opened 2017-07-14
add prometheus stats tracking

elliefm:
This adds prometheus stats tracking.  

Each process that wants to record stats needs to `#include "imap/prometheus.h"`, and it can then use `prometheus_increment(stat)` and `prometheus_decrement(stat)` on statistics.  When a process that counts statistics runs, a proc-like file will be created for it in `config_dir/stats/` (override with `prometheus_stats_dir` in imapd.conf), into which counts for this process will be recorded.  Statistics are defined in `imap/promdata.p`, which is converted into C header/source by `imap/promdatagen` during build.

Initially, this is collecting the same set of imapd and lmtpd statistics that our SNMP implementation collects.

The promstatsd process is responsible for periodically collating these files into `config_dir/stats/report.txt`, and a new http module serves up this file in response to GET requests for "/prometheus".

To set it up:

* set `prometheus_enabled` to "yes" in imapd.conf
* add `prometheus` to your `httpmodules` in imapd.conf
* [optionally] set `prometheus_need_auth`, `prometheus_update_freq` and `prometheus_stats_dir` in imapd.conf
* add a job to run `promstatsd -c` to the START section of cyrus.conf (this cleans up the stats files from the previous run)
* add a job to run `promstatsd` to the DAEMON section of cyrus.conf (the actual daemon process)


elliefm:
I've added a prometheus_enabled config switch, default off, so that installs that don't want or haven't considered prometheus don't get a directory full of stats files they don't care about.

elliefm:
I've got some basic cassandane tests for this now, at https://github.com/cyrusimap/cassandane/compare/master...elliefm:v31/prometheus

robn:
Can we change `/prometheus` to `/metrics`? `/metrics` is a strong convention for Prometheus metrics endpoints, and its not easy (possible?) to change it in all Prometheus configs (in our case, when the Consul SD backend is being used to discover services).

elliefm:
Sure, easy enough

elliefm:
One thing I know still needs to be done is cleaning up per-process stats files when they've exited, but while Cyrus is still running -- without losing the stats the contain (because those counters aren't allowed to go backwards.  I envision accumulating them into a "finished processes" file and then removing the per-process file (with careful locking to avoid double-counting in between).

Question though, what's more useful from a service point of view:  get rid of them as soon as they're no longer needed, or hang onto them for a while in case they're useful?  And in the latter case, how long is "a while"?

The 'promstatsd -c' invocation at startup does a good job of ensuring we start up with a fresh slate, but doesn't do anything to prevent a long running cyrus from potentially filling up its stats partition, or reversing counters when a reused pid gets its stats trashed back to zero.

robn:
I'm pretty sure they're never likely to be useful. Clean them up whenever you're ready.

robn:
I think some of the metrics might do well to be renamed.

I'm following the guidance in:
* [Metric and label naming](https://prometheus.io/docs/practices/naming/), official Prometheus guidelines
* [On the naming of things](https://www.robustperception.io/on-the-naming-of-things/), which seems to be referenced by a number of third-party Prometheus exporters.

* Prefix them all with `cyrus_`, so we can search for them easily and distinguish them from similarly-named metrics from other systems.
* Don't suffix counters with `_count`. `_total` is right.
* I think all counters get `_total`, eg in things like `lmtp_received_messages`, the "unit" would be `message`?

I'll chip in with a patch for these shortly. Sorry - I know you asked me about naming before, but I don't think I fully understood until I really starting looking at these (and I now have to go and redo some of my own exporters).

robn:
I notice there's no metrics for eg http, pop3, sync, etc daemons, and no metrics for mailbox-level stuff (where we'd typically see auditlog). I expect that this is just a "we can add more later" kind situation? I certainly don't think they're necessary to start; just making sure there wasn't some fundamental issue that makes them impossible?

elliefm:
httpd stats are on the way (already sitting in a branch, waiting to see that the general approach checks out).  Will have a PR for them up not long after this one lands.

The initial set of stats is just "what SNMP already (didn't) collect (but wanted to)", hence just imapd/lmtpd

Will review stat names, thanks!

robn:
My attempt at metric name updates: robn/cyrus-imapd@e53a00a. I'm still testing, I'll let you know soon if I think they're close.

PR #2066, opened 2017-07-17
[ischedule] Use a valid struct stat buf.

ajaysusarla:
In `meth_get_ischedule()` a `struct stat` buffer is used to generate the ETag.
But the stat buffer used is never initialised. This patch fixes that.

Possibly broken because of changes in 7ae6fee12e09bb171438f3dd7624b1c0b9511e41

### DIFF
--- a/imap/http_ischedule.c
+++ b/imap/http_ischedule.c
@@ -175,7 +175,7 @@ struct namespace_t namespace_domainkey = {
 };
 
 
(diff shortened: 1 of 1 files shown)


elliefm:
Does this fix need to be backported to 3.0 too?

ajaysusarla:
@elliefm yes. I just checked the `3.0` branch. I shall send a pull request for the `3.0` branch as well.
Thanks for the review.

PR #2068, opened 2017-07-18
GCC 7 compilation fixes.

ajaysusarla:
This patch fixes the warnings when compiling `cyrus-imapd` with gcc 7.
Primarily address thing fallthrough cases in `switch` statements[1].

[1] https://gcc.gnu.org/onlinedocs/gcc-7.1.0/gcc/Statement-Attributes.html

### DIFF
--- a/imap/append.c
+++ b/imap/append.c
@@ -1257,7 +1257,7 @@ HIDDEN int append_run_annotator(struct appendstate *as,
 
     /* Reset user flags */
     uint32_t user_flags[MAX_USER_FLAGS/32];
-    memset(user_flags, 0, sizeof(user_flags)/sizeof(user_flags[0]));
+    memset(user_flags, 0, sizeof(user_flags));
     r = msgrecord_set_userflags(msgrec, user_flags);
     if (r) goto out;
 
--- a/imap/ctl_mboxlist.c
+++ b/imap/ctl_mboxlist.c
@@ -1007,6 +1007,7 @@ int main(int argc, char *argv[])
 
     case M_POPULATE:
         syslog(LOG_NOTICE, "%spopulating mupdate", warn_only ? "test " : "");
+        GCC_FALLTHROUGH
 
     case DUMP:
         mboxlist_init(0);
--- a/configure.ac
+++ b/configure.ac
@@ -154,6 +154,13 @@ AH_BOTTOM([#if HAVE_VISIBILITY
 #define HIDDEN
 #endif])
 
+AH_BOTTOM([#if defined __GNUC__ &&  __GNUC__ > 6
+    #define GCC_FALLTHROUGH __attribute__((fallthrough));
+#else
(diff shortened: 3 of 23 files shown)


imap/append.c
```diff
@@ -1257,7 +1257,8 @@ HIDDEN int append_run_annotator(struct appendstate *as,
 
     /* Reset user flags */
     uint32_t user_flags[MAX_USER_FLAGS/32];
-    memset(user_flags, 0, sizeof(user_flags)/sizeof(user_flags[0]));
```
elliefm: Both the original and the fixed version seem bogus here.  It will only zero out the first quarter of the array?

The `sizeof(foo) / sizeof(foo[0])` idiom is for calculating the number of array elements.  But memset's third argument is a size in bytes, not a number of elements.

It should be correct to do `memset(user_flags, 0, sizeof(user_flags));` (and, this should also stop gcc7 from whinging about it).

imap/ctl_mboxlist.c
```diff
@@ -1007,6 +1007,9 @@ int main(int argc, char *argv[])
 
     case M_POPULATE:
         syslog(LOG_NOTICE, "%spopulating mupdate", warn_only ? "test " : "");
+#if defined __GNUC__ &&  __GNUC__ > 6
```
elliefm: Ironically, all the #ifdef noise makes it less clear that this is a falling through...

I would maybe put something like this into lib/util.h:

```
#if defined __GNUC__ &&  __GNUC__ > 6
    #define GCC_FALLTHROUGH __attribute__((fallthrough));
#else
    #define GCC_FALLTHROUGH /* fall through */
#endif
```

and then the switches can do something like:

```
case M_POPULATE:
    syslog(blah);
    GCC_FALLTHROUGH

case DUMP:
    ...
```

An even better place than lib/util.c would be if it wound up in config.h, since everything includes that.  I don't remember off the top of my head how stuff gets into there (maybe config.h.in?) but I'm sure it's not hard to work out

imap/append.c
ajaysusarla: Aha! Absolutely! :) WIll correct that!

imap/ctl_mboxlist.c
ajaysusarla: I did implement it as a `#define`  initially (you can see that I've left the `GCC_VERSION` macro in `util.h` that I've still left there). But I wasn't very comfortable with the macro way, since they are an absolute pain when debugging.
Then again, if you think that it is better, I shall go ahead an update the patch.

elliefm: Yeah, I don't really think one-line macros like this make debugging any harder (especially when it's a no-op line like this anyway), and encapsulating the nastiness makes the switch cases much more readable

ajaysusarla: Ack!

ajaysusarla:
@elliefm Updated the patch based on your review.

PR #2075, opened 2017-08-02
[idle] Fix conflicting variable names for idle socket.

ajaysusarla:


brong:
Clearly better than reusing the same name for totally different datatypes 5 lines apart!

Issue #1155, opened 2010-05-27
PATCH TLS doesn't check CRLs


brong:
From: stacy@millions.ca
Bugzilla-Id: 3229
Version: 2.3.x
Owner: [Ken Murchison](murch@andrew.cmu.edu)

ajaysusarla:
@ksmurchison Do you think this bug is still relevant? If yes, how can I test this in Cyrus?

ksmurchison:
I don't know for sure if its still relevant, but I expect that it might be.  I'm neither a TLS nor a security expert.  It might be worth doing a little OpenSSL research on certificate revocation lists.

I do know that the patch is flawed in that it should be adding the new option to lib/imapoptions and not the derived imapopts.[ch] files

ajaysusarla:
@ksmurchison Thank you! I shall dig into this. Cheers!

PR #2078, opened 2017-08-07
[Bug 1155] Check CRL's in TLS.

ajaysusarla:
We don't check CRL's in TLS and this patch adds support for that. An
imap option called `IMAPOPT_TLS_CRL_FILE` is available to specify the
path of the CRL PEM file.

This patch is based of the work done by `stacy@millions.ca`.

### DIFF
--- a/imap/tls.c
+++ b/imap/tls.c
@@ -336,6 +336,8 @@ static int verify_callback(int ok, X509_STORE_CTX * ctx)
     case X509_V_ERR_ERROR_IN_CERT_NOT_AFTER_FIELD:
         syslog(LOG_NOTICE, "cert has expired");
         break;
+    case X509_V_ERR_CERT_REVOKED:
+        syslog(LOG_NOTICE, "cert has been revoked");
     }
 
     return (ok);
@@ -694,6 +696,7 @@ EXPORTED int     tls_init_serverengine(const char *ident,
     const char   *server_ca_file;
     const char   *server_cert_file;
     const char   *server_key_file;
+    const char   *crl_file_path;
     enum enum_value tls_client_certs;
     int server_cipher_order;
     int timeout;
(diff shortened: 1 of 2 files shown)


imap/tls.c
```diff
@@ -943,6 +946,27 @@ EXPORTED int     tls_init_serverengine(const char *ident,
...
+
+        if (crl_file) {
+            while ((crl = PEM_read_bio_X509_CRL(crl_file, NULL, NULL, NULL))) {
+                X509_STORE_add_crl(store, crl);
+                X509_CRL_free(crl);
+            }
+
+            X509_STORE_set_flags(store,
+                                 X509_V_FLAG_CRL_CHECK |
+                                 X509_V_FLAG_CRL_CHECK_ALL);
```
elliefm: I think we're leaking `crl_file` here?

ajaysusarla: That is true. Will update the patch! Thanks @elliefm

elliefm:
by the way, if you refer to the issue as #1155 then github creates cross references between the issue and the pull request, like I've just done now :)

elliefm:
(you can also use the same shorthand in commit messages and github will do the right thing with them)

imap/tls.c
```diff
@@ -943,6 +946,29 @@ EXPORTED int     tls_init_serverengine(const char *ident,
...
+        } else {
+            syslog(LOG_ERR, "Can't load CRL file %s.", crl_file_path);
+        }
+
+        BIO_free(crl_file);
```
ajaysusarla: Freeing the file handle now!

ajaysusarla:
@elliefm Updated the commit message to reference #1155

ajaysusarla:
@elliefm I'm getting my head around the CRL's. This seems sane as per openssl API documentation. I intend to write a cassandane test for this.

PR #2083, opened 2017-08-09
[mailbox] Cached header lookup improvements.

ajaysusarla:
Use binary search instead of looping through the `mailbox_header_cache`
list. This could be further optimised, by using a hash table.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -143,7 +143,8 @@ BUILT_SOURCES += \
 	lib/htmlchar.c \
 	lib/htmlchar.h \
 	imap/rfc822_header.c \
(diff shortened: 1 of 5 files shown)


brong:
Hmm... yeah, interesting.  In fact, since you suggest a hash... we already use gperf for hashing headers in another part of the code - it's probably worth looking at gperf for this!

ajaysusarla:
Will do. I think it does make sense for optimising this, since it is called pretty frequently.

ajaysusarla:
Closing this pull request, to prevent noise. I'm creating another one.

PR #2100, opened 2017-08-15
remove all trace of SNMP

### LINKED ISSUE #1765: Move SNMP out from master into a separate daemon
Greg explained this to me before he left FastMail.

Basically, every process should have its own file (like it already does with proc / lock / idle, etc) in which it stores stats in a simple format (possibly just binary, mmaped, write out numbers to locations - could be on tmpfs).

And then the snmp or other stats daemons just open the directory, read each file, increment counters.  This is blindingly fast in practice, and it means there's no messy threaded snmp junk sitting in the master process.  You'd just run the statsd as a separate daemon.

elliefm:
SNMP support has been flaky and largely broken for a long time, it's time to send it on its way.

All the stuff that our SNMP monitoring purported to collect is now available via Prometheus monitoring instead, which actually works. :)

Merging this closes #1765


elliefm:
@brong @ksmurchison @rsto I'd like to get this merged onto master now that 3.2 has been released.  Can I get some eyes on it please?

Should be a pretty quick/easy review: there's lots of files touched, but the logic is trivial

Issue #1827, opened 2017-02-08
Report maxchild value for each service via SNMP


postilion:
Add a new serviceEntry to CYRUS-MASTER.mib with maxchild for each service, so when reading table one may alert on serviceActive as a percentage of serviceMaxchild.

i.e. 
CYRUS-MASTER-MIB::serviceForks.1 = Counter32: 27
CYRUS-MASTER-MIB::serviceActive.1 = Gauge32: 27
CYRUS-MASTER-MIB::serviceName.1 = STRING: imap
CYRUS-MASTER-MIB::serviceId.1 = INTEGER: 1
CYRUS-MASTER-MIB::serviceConnections.1 = Counter32: 31
CYRUS-MASTER-MIB::serviceMaxchild.1 = Counter32: 250

"Service imap 11%; using 27 of 250"

elliefm:
maybe do this at the same time as #1765

elliefm:
@postilion we don't seem to have anything useful in the way of SNMP docs.  All I can find is `doc/legacy/install-snmpmon.html` which I think is both ancient and totally wrong -- it references something called `tugowar`, which looks like it was removed as of _2.1.12(!)_

It sounds like you have it running to some degree or another -- are you able to put together some documentation for it?  Even rough notes would be a help for now -- I'm trying to wrap my head around it but I've never touched it before

postilion:
@elliefm Yeah, one of the things for the To Do list.  I sussed out the proper use of this a couple of months ago, mostly by reading the code, and have been using it in process monitoring since then for a few clients.  One of the things I would like to tackle whilst there, if FastMail even uses it, that is.

elliefm:
We don't at the moment, but would like to once it's in better shape

elliefm:
I think I'll add maxchild to 3.0 too (but the eventual refactor will only be on master)

But in the meantime, I'm putting together some cassandane tests so I can decipher how this thing works, and then verify that I haven't broken it when I start making changes! :P

elliefm:
I've got this working on a side branch, and have the guts of cassandane tests going too (just need some actual tests now):

```
ellie@debian-testing::~$snmpwalk -v1 -cpublic localhost .1.3.6.1.4.1.3.6
CYRUS-MASTER-MIB::cyrusMasterInfoDescr.0 = STRING: Cyrus IMAP server master process
[...]

PR #2102, opened 2017-08-16
master: report service metrics for prometheus

elliefm:
This adds a mechanism by which master can include its service metrics into the prometheus metrics, while only depending on code already in libcyrus_min.

I'm particularly interested in suggestions for what these metrics should be named, their help strings, and most importantly, whether they're actually useful.  I've also left out the "nconnections" stat for now cause I'm not really clear what it's actually counting.

I consider this pull request to be a work in progress at the moment.

### DIFF
--- a/imap/prometheus.c
+++ b/imap/prometheus.c
@@ -311,5 +311,25 @@ EXPORTED int prometheus_text_report(struct buf *buf, const char **mimetype)
     mappedfile_unlock(mf);
     mappedfile_close(&mf);
     free(report_fname);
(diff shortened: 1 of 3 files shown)


elliefm:
#1827 asked for the (SNMP) stats to include the maxchild value for each service.  This could be added in here easily enough.

elliefm:
If we're going to report maxchild for comparing against nactive, it would make sense to also report maxforkrate, for comparing against forkrate.

PR #2103, opened 2017-08-17
[sieve] Cleanup clutter in autocreate_sieve().

ajaysusarla:
The autocreate_sieve() function needed a little bit of love.

### DIFF
--- a/imap/autocreate.c
+++ b/imap/autocreate.c
@@ -93,88 +93,167 @@ static const char *get_script_name(const char *filename)
         return p + 1;
 }
 
+enum SieveFileType {
+    SIEVE_TMP_1 = 1,
+    SIEVE_TMP_2,
+    SIEVE_BC_TMP,
+    SIEVE_SCRIPT,
+    SIEVE_BC_SCRIPT,
+    SIEVE_DEFAULT,
+    SIEVE_BC_LINK,
+    SIEVE_NUM_FILE_TYPES
+};
+static struct sieve_scripts_info {
+    enum SieveFileType stype;
+    char *ext;
+} sieve_names[] = {
+    {SIEVE_TMP_1, ".script.NEW"},
+    {SIEVE_TMP_2, ".NEW"},
+    {SIEVE_BC_TMP, ".bc.NEW"},
+    {SIEVE_SCRIPT, ".script"},
+    {SIEVE_BC_SCRIPT, ".bc"},
+    {SIEVE_DEFAULT, ""},
+    {SIEVE_BC_LINK, ".bc"}
+};
+
+struct sieve_scripts {
+    char tmpname1[MAX_FILENAME];
+    char tmpname2[MAX_FILENAME];
+    char bctmpname[MAX_FILENAME];
+    char scriptname[MAX_FILENAME];
+    char bcscriptname[MAX_FILENAME];
+    char defaultname[MAX_FILENAME];
+    char bclinkname[MAX_FILENAME];
+};
+
+/*
+ * Given the path to the sieve script directory `script_dir` and the sievename,
+ * generate filenames and return it in the `fnames` structure.
+ *
+ * On Success, returns 0, else returns 1
+ */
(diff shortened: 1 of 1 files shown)


imap/autocreate.c
```diff
@@ -93,330 +93,398 @@ static const char *get_script_name(const char *filename)
...
+        syslog(LOG_ERR, "autocreate_sieve: Invalid sieve path %s, %s, %s",
+               sieve_script_dir, sievename, userid);
+        goto failed_start;
     }
 
     /* Check if a default sieve filter already exists */
-    if(!stat(sieve_default,&statbuf)) {
-        syslog(LOG_WARNING,"autocreate_sieve: Default sieve script already exists");
-        return 1;
+    if(!stat(script_names.defaultname, &statbuf)) {
```
elliefm: while you're in here, you could fix all the `if(...)` to `if (...)`  (with a space after "if") :)

elliefm:
It's worth considering whether you can use the `sieve_rebuild()` function here -- it deduplicates a lot of the parse/compile/output.  But there are a few cases where it's unsuitable, I can't remember if this is one of those cases.

imap/autocreate.c
ajaysusarla: ack! I missed a few of these. :-/ I wanted to rewrite the whole thing, but it gave me a headache trying to figure out the code flow, so thought I would first bring some semblance of sanity and go from there.

ajaysusarla:
@elliefm This sure needs a second look. I wanted to cleanup the most obvious of these. I will do a second parse of this soon.

imap/autocreate.c
```diff
@@ -93,330 +93,399 @@ static const char *get_script_name(const char *filename)
...
+            case SIEVE_BC_SCRIPT:
+                ret = snprintf(fnames->bcscriptname, MAX_FILENAME, "%s/%s%s",
+                               script_dir, sievename, info->ext);
+                break;
+            case SIEVE_DEFAULT:
+                ret = snprintf(fnames->defaultname, MAX_FILENAME, "%s/%s%s",
+                               script_dir, sievename, info->ext);
+                break;
+            case SIEVE_BC_LINK:
+                /* For some reason bclink name has no directory. */
```
elliefm: This is because a relative symlink target is relative to the location of the symlink, and since the defaultname symlink is being created in the appropriate directory, its target can't also specify the directory (otherwise you'd get like "user/f/foo/default" pointing to "[user/f/foo/]user/f/foo/somescript.bc" and things would fall apart :) )

imap/autocreate.c
```diff
@@ -93,330 +93,399 @@ static const char *get_script_name(const char *filename)
...
-        return 1;
+        umask(oldmask);
+        goto failed2;
     }
     umask(oldmask);
 
-    while((r = fread(buf,sizeof(char), sizeof(buf), in_stream)) > 0) {
-        if( fwrite(buf,sizeof(char), r, out_fp) != (unsigned)r) {
-            syslog(LOG_WARNING,"autocreate_sieve: Problem writing to sieve script file %s: %m",sieve_tmpname);
+    while ((r = fread(buf, sizeof(char), sizeof(buf), in_stream)) > 0) {
```
elliefm: actually, if you haven't already, I'd say it's worth spending some time just reading through lib/util.c from top to bottom, there's a lot of stuff in there that comes in handy if you know it's available

ajaysusarla: There indeed seems to be a lot of stuff in `lib/util.c`, sorry for not noticing it earlier. I shall use a few of them (here and in other places as well.)

imap/autocreate.c
ajaysusarla: Aha! Thanks for that. I wasn't sure as to why this was being done. I shall add your comment on the code for future reference.

ajaysusarla:
@elliefm Thanks for the review. I agree with you that this needs a test. I shall write one.

PR #2123, opened 2017-08-25
Avoid returning -1 as 255

### LINKED ISSUE #2122: [2.5/ppc64le]: cunit/parse.testc failures
wrap_getint32() and wrap_getsint32() fail, as they appear to return 255 rather than -1.

kanarip:
Resolves #2122

### DIFF
--- a/cunit/parse.testc
+++ b/cunit/parse.testc
@@ -23,7 +23,7 @@ int wrap_getint32(const char *s, int32_t *valp)
 {
     struct protstream *prot;
     char *b;
-    int c;
+    int8_t c;
 
     b = xstrdup(s);	/* work around bug in prot_ungetc */
     prot = prot_readmap(b, strlen(b));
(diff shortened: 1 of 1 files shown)


elliefm:
This doesn't seem right, these are `int`s all the way down, so no part of them should be converting -1 to 255 unless the platform is only 8bit (which, I'm guessing "ppc64le" is 64-bit).  Converting down to int8_t seems like it might _cause_ `-1==255`, not avoid it??

These functions are explicitly dealing in 32 bit integers so if anything, the correct explicit type should be int32_t?  And if there's an endianness issue that causes some weird `-1=255` conversion then this converting down to an 8bit type won't it, just maybe hide it...

kanarip:
I believe it is an issue related to older versions of the compiler as well, if that helps. GCC 4.7.3 (iirc) on ppc64le does show this weird behaviour, but more up-to-date 6+ and 7+ series (iirc) do not.

elliefm:
Do you still have need to support the older compilers, or can we just close this off?

kanarip:
I can conditionally patch it in my packaging, if it makes no sense to include upstream...

FWIW, this would be RHEL 7 on PPC and ARM -- old, but not too old ;-)

PR #2136, opened 2017-09-05
[Docs] Odd-even release cycle versioning.

nicolan:
Some documentation outlining the odd and even approach
in our release versioning. Odd = development only, even= stable.

Flagging the odd releases with big warning messages so
nobody can be surprised at what they're seeing.


nicolan:
@ajaysusarla Does this look suitable for you per our conversations a week ago or so? If so, go ahead and merge on master and 3.0 branch.

PR #2147, opened 2017-09-11
[Annotator] Control use of annotator during append

### LINKED ISSUE #2109: Add separate config variable to control use of annotator during append
Right now annotator_callout controls both the xrunannotator command and the append_fromstage callout.  It would be good to have a switch so annotations aren't run on append, but xrunannotator still works for testing purposes.

The option would be on by default to keep existing behaviour.

ajaysusarla:
Add separate config variable to control use of annotator during append.

Fixes #2109


ajaysusarla:
Cassandane test is available here: https://github.com/cyrusimap/cassandane/pull/39

PR #2158, opened 2017-09-20
sync_support: downgrade recoverable SYNCERRORs to SYNCNOTICE

robn:
Assuming that SYNCERROR means what I think it (should mean), namely "I
can't possible continue and will need an operator to intervene", all the
cases that are recoverable by doing a full resync should only warn.

The only thing I'm not sure about is if a full resync can ever fail for
one of these reasons. If it can, then there's now nothing to alert the
operator, and this patch probably should be applied as-is (at the cost
of someone getting woken up :/ )

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -2442,7 +2442,7 @@ static int sync_mailbox_compare_update(struct mailbox *mailbox,
             /* GUID mismatch is an error straight away, it only ever happens if we
              * had a split brain - and it will take a full sync to sort out the mess */
             if (!message_guid_equal(&mrecord.guid, &rrecord->guid)) {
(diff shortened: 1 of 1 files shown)


brong:
Yep, looks good to me


On Wed, 20 Sep 2017, at 13:48, Partha Susarla wrote:
> *@ajaysusarla* approved this pull request.


> This sounds reasonable.


> — You are receiving this because your review was requested. Reply to
> this email directly, view it on GitHub[1], or mute the thread[2].> 

--
  brong@fastmailteam.com



Links:

  1. https://github.com/cyrusimap/cyrus-imapd/pull/2158#pullrequestreview-63860894

PR #2159, opened 2017-09-20
sync_support: emit a SYNCERROR on unexpected mailbox uniqueid change

robn:
I saw this today while recovering after a disk crash. It did not recover
by itself. I don't know if it should be recoverable by a full sync (and
this code should return IMAP_SYNC_CHECKSUM), but as long as requires
operator interventions, is should log a SYNCERROR.

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -2681,7 +2681,7 @@ int sync_apply_mailbox(struct dlist *kin,
             mailbox->header_dirty = 1;
         }
         else {
-            syslog(LOG_ERR, "Mailbox uniqueid changed %s (%s => %s) - retry",
+            syslog(LOG_ERR, "SYNCERROR: Mailbox uniqueid changed %s (%s => %s) - retry",
                    mboxname, mailbox->uniqueid, uniqueid);
             r = IMAP_MAILBOX_MOVED;
             goto done;


brong:
Unexpected UniqueID change means that the mailbox was created twice in two different locations - so it's strictly two different mailboxes with the same name!

This will probably be fixed as part of the replication redesign work.

robn:
Oh crap, I meant to comment further on this. I ran it at FM for a day while rebuilding a failed server, and I can't be sure, but I think it was flagging things that it was actually able to recover from. I was unsure enough that I reverted it, because I don't want to get paged for non-problems!

But you tell me - is that actually possible? I don't grok this case very well.

brong:
It should only happen if a mailbox got created on two different servers
with different uniqueids and the same name.
I'm hoping that we fix it properly when we revisit the sync protocol.

Bron.


On Tue, 26 Sep 2017, at 22:25, Rob N ★ wrote:
> Oh crap, I meant to comment further on this. I ran it at FM for a day
> while rebuilding a failed server, and I can't be sure, but I think it
> was flagging things that it was actually able to recover from. I was
> unsure enough that I reverted it, because I don't want to get paged
> for non-problems!> But you tell me - is that actually possible? I don't grok this case
> very well.> — You are receiving this because you modified the open/close state.
> Reply to this email directly, view it on GitHub[1], or mute the
> thread[2].> 

--
  brong@fastmailteam.com



Links:

  1. https://github.com/cyrusimap/cyrus-imapd/pull/2159#issuecomment-332181590

Issue #855, opened 2006-06-28
timsieved SIGSEGV


brong:
From: [Kwiatek](kwiatek@kwiatek.eu.org)
Bugzilla-Id: 2856
Version: 2.3.x
Owner: [Ken Murchison](murch@andrew.cmu.edu)

chenz-svsarrazin:
Backtrace from 2.4.18 (Ubuntu 16.04 LTS)

Invoked (as root): `sieveshell --authname cyrus localhost`

```
#0  strlen () at ../sysdeps/x86_64/strlen.S:106
#1  0x00007f92aafbed19 in xstrdup (str=0x2 <error: Cannot access memory at address 0x2>) at xmalloc.c:92
#2  0x00007f92aaf701e2 in cmd_authenticate (errmsg=<synthetic pointer>, initial_challenge=<optimized out>, mechanism_name=0x7f92ac0aa1e0, sieved_in=0x7f92ac0a7bf0, 
    sieved_out=0x7f92ac0a8f00) at parser.c:830
#3  parser (sieved_out=0x7f92ac0a8f00, sieved_in=0x7f92ac0a7bf0) at parser.c:230
#4  0x00007f92aaf6eac3 in cmdloop () at timsieved.c:203
#5  0x00007f92aaf6ed9f in service_main (argc=<optimized out>, argv=<optimized out>, envp=<optimized out>) at timsieved.c:341
#6  0x00007f92aaf6bf80 in main (argc=<optimized out>, argv=<optimized out>, envp=0x7ffc2acf5638) at service.c:586
```

PR #2172, opened 2017-10-04
Fix segfault due to unitialized local variable in timsieved.

chenz-svsarrazin:
This happens when a connection is made as an admin-user without a
mailbox (`sieveshell --authname cyrus localhost`)

See issue #855.


elliefm:
Thanks! :)

PR #2176, opened 2017-10-17
getMessageListUpdates: Fix uptoMessageId parsing

wolfsage:
readprop returns < 0 if there is an error, not greater than


elliefm:
Thanks for the pull request, sorry it's taken so long to get to it.  Looks like this was already manually applied as 4809be7a4.

Issue #1699, opened 2014-06-02
cyr_expire only accepts internal mailbox naming


brong:
From: [Rudy Gevaert](rudy.gevaert@ugent.be)
Bugzilla-Id: 3854
Version: 2.4.17
Owner: [Bron Gondwana](brong@fastmail.fm)

karagian:
Not sure if I should open a new bug, but I noticed the same problem with "quota" command in the 3.0.x version.

elliefm:
I don't think we want to change behaviour on 3.0, but this makes sense for master.

FastMail infrastructure will need coordination because we have scripts that run off the back of these commands.

I'm 80% sure there's another open issue about the same thing, opened by @brong I believe?

karagian:
Hi Ellie! Thank you for taking a look into this. The behavior of these command line tools changes in versions 2.5.x and 3.0.x. Administrators will only bump into this bug only if **unixhierarchysep** is set and will not be able to use these commands properly, unless they perform a conversion in their scripts of the mailbox names coming from other command line tools... The bug fix will save them from this trouble.
If unixhierarchysep is not set, the bug is never noticed and scripts using these commands do not need to change anything, since they already handle mailbox names in the internal format.

elliefm:
Ah thanks, that's very helpful!

PR #2204, opened 2017-11-29
cyr_expire, quota : Fix for issue #1699. Accept parameters based on current namespace settings

karagian:
Fix for issue #1699. Convert command line mailbox parameters to internal namespace based on current settings. This applies to cyr_expire and quota commands

### DIFF
--- a/imap/cyr_expire.c
+++ b/imap/cyr_expire.c
@@ -593,6 +593,17 @@ int main(int argc, char *argv[])
 
     mboxevent_setnamespace(&expire_namespace);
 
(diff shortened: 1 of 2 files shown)


elliefm:
I like this PR for master, but I'm not sure for 3.0.  Changing interpretation of command line parameters in a stable release feels like a bad idea -- if admins have stuff scripted off this, it will break on them.

rsto:
@elliefm do you still think it's a good idea?

elliefm:
Looks like I fixed this on master myself in https://github.com/cyrusimap/cyrus-imapd/pull/2923, which was released in 3.4.  I still agree with my past self that applying it to 3.0 (or 3.2 for that matter) isn't really appropriate

PR #2206, opened 2017-11-30
Fix imapd.conf example files for murder

### LINKED ISSUE #2099: doc example murder-frontend.conf with wrong proxyservers parameter
The example murder-frontend.conf contains a comment about proxyservers and a default/example for this parameter.

Poorly this is completely wrong in this configuration. Compared to the manpage...

"A  list  of  users  and  groups that are allowed to proxy for other users, separated by spaces.  Any user               listed in this will be allowed to login for any other user: use with caution.  In a standard murder  this               option should ONLY be set on backends.  DO NOT SET on frontends or things won't work properly."

The most important sentence is the last one... the last should be added to the example for the backend and the whole part of proxyservers removed from murder-frontend.conf .

See also: https://lists.andrew.cmu.edu/pipermail/cyrus-devel/2017-August/004127.html

postilion:
There was an error regarding the use of "proxyservers:" in the example
configuration files for murder frontends.  This commit fixes it.

Closes #2099

### DIFF
--- a/doc/examples/imapd_conf/murder-backend.conf
+++ b/doc/examples/imapd_conf/murder-backend.conf
@@ -22,7 +22,8 @@ mupdate_password: <secret>
 # The proxyservers parameter is a list of users or groups who are
 # allowed to proxy for any other users, and are necessary for backend
 # servers to support user and mailbox moves between backends.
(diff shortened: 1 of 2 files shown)


elliefm:
Thanks :)

PR #2216, opened 2017-12-14
Fix error message about failing to install SIGUSR2 handler

jikamens:
The error message when the SIGUSR2 handler fails to be installed said
SIGTERM when it should have said SIGUSR2.


elliefm:
Thanks! :)

elliefm:
Also cherry-picked this to the 3.0 branch

PR #2219, opened 2017-12-18
promstatsd: report usage information

elliefm:
This PR builds upon #2102 and must be merged after it.

This adds usage information (per partition) to the prometheus report:

* number of user accounts
* number of deleted mailboxes
* number of shared mailboxes
* quota commitments

The quota commitments are tricky, because there is no hard relationship between quota roots and disk partitions.  If a quota root crosses partition boundaries, its quota commitments will be counted towards each partition it affects, making it hard to infer anything useful from the data.  (Quota roots that cross partition boundaries are hard to work with in plenty of other ways, so my advice here is "just don't do that then".)

This does not count actual on-disk usage or free space; get those from your operating system metrics if you want to calculate percentages or alarm at risky thresholds etc.

Also, note that quota commitments can be infinite.  This is represented correctly per the Prometheus spec, but is worth a heads up. :)

This is an ad-hoc implementation within promstatsd.c, similar to the one in master.c (#2102).  These stats aren't suitable for counting from a service process, as they're global to the Cyrus instance, so promstatsd just calculates them itself at report time, bypassing the service API from prometheus.h.  I expect to generalise a bunch of this once the gremlins are shook out.

I have Cassandane tests for the tricky bits at  https://github.com/elliefm/cassandane/tree/v31/prometheus, but I won't merge them until the functionality is in Cyrus master.

Sample report contents:

```
# HELP cyrus_usage_deleted_mailboxes The number of deleted Cyrus mailboxes
# TYPE cyrus_usage_deleted_mailboxes gauge
cyrus_usage_deleted_mailboxes{partition="default"} 0 1513575267167
cyrus_usage_deleted_mailboxes{partition="p2"} 0 1513575267166
# HELP cyrus_usage_users The number of Cyrus user Inboxes
# TYPE cyrus_usage_users gauge
cyrus_usage_users{partition="default"} 1 1513575267167
cyrus_usage_users{partition="p2"} 0 1513575267166
# HELP cyrus_usage_mailboxes The number of Cyrus mailboxes
# TYPE cyrus_usage_mailboxes gauge
cyrus_usage_mailboxes{partition="default"} 4 1513575267167
cyrus_usage_mailboxes{partition="p2"} 2 1513575267166
# HELP cyrus_usage_shared_mailboxes The number of shared Cyrus mailboxes
# TYPE cyrus_usage_shared_mailboxes gauge
cyrus_usage_shared_mailboxes{partition="default",namespace="bar"} 3 1513575491793
cyrus_usage_shared_mailboxes{partition="default",namespace="foo"} 3 1513575491793
# HELP cyrus_usage_quota_commitment The amount of quota committed
# TYPE cyrus_usage_quota_commitment gauge
cyrus_usage_quota_commitment{partition="default",resource="STORAGE"} 12000 1513575267167
cyrus_usage_quota_commitment{partition="default",resource="MESSAGE"} inf 1513575267167
cyrus_usage_quota_commitment{partition="default",resource="X-ANNOTATION-STORAGE"} inf 1513575267167
cyrus_usage_quota_commitment{partition="default",resource="X-NUM-FOLDERS"} inf 1513575267167
cyrus_usage_quota_commitment{partition="p2",resource="STORAGE"} 10000 1513575267166
cyrus_usage_quota_commitment{partition="p2",resource="MESSAGE"} inf 1513575267166
cyrus_usage_quota_commitment{partition="p2",resource="X-ANNOTATION-STORAGE"} inf 1513575267166
cyrus_usage_quota_commitment{partition="p2",resource="X-NUM-FOLDERS"} inf 1513575267166
```

### DIFF
--- a/imap/promstatsd.c
+++ b/imap/promstatsd.c
@@ -307,7 +307,7 @@ struct partition_data {
     int64_t n_deleted;
     int64_t n_shared; /* XXX ??? */
 
-    double quota_commitment;
+    double quota_commitment[QUOTA_NUMRESOURCES];
 
     int64_t timestamp;
 };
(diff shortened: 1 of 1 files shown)


elliefm:
This has been reverted, the calculation was too expensive

elliefm:
bron says on chat:

> yeah, but check this bit out
> 
> ```
> num_misses++;
> if (num_misses > FOREACH_LOCK_RELEASE) {
>     /* take a copy of they key - just in case cb does actions on this database
>      * and clobbers loc */
>     buf_copy(&keybuf, &db->loc.keybuf);
> ```
> 
> in cyrusdb_twoskip.c
> if you're guaranteed not to modify the DB at all, you can put the counting in find_p and return 0

Try this optimisation out too if bringing this PR back

A read-only foreach would be a nice addition to the cyrusdb api, but in the meantime....

PR #2224, opened 2017-12-29
Try to clarify documentation in re: Partitions

postilion:
This is a first attempt to clear up some confusion in regards to the
nameing of partitions and the use of the various "partition-name"
configuration directives in imapd.conf(5).

It's my goal to eliminate all use of common words like "default" or
"archive" in the names of partitions in samples and descriptive
language, lest people perceive such use to mean that those names
*must* be used.

For example, rather than using the following:
    defaultpartition: default
    partition-default: /var/spool/cyrus

we can use:
    defaultpartition: part1
    partition-part1: /var/spool/cyrus

The file /lib/imapoptions has been lightly modified in service to this
cause, and to correct typos in the definition of searchpartition-name
settings.

Note: There had been some trailing white space in /lib/imapoptions which my editor removed, which accounts for the last couple of diffs in that file.

### DIFF
--- a/docsrc/imap/reference/admin/locations.rst
+++ b/docsrc/imap/reference/admin/locations.rst
@@ -35,6 +35,55 @@ In general, Cyrus allows one to maintain several separate partitions for
 each of these data types, and to establish rules governing distribution
 of data within each pool.
 
+Types of Partitions
+-------------------
+
+Cyrus supports several different types of partitions:
+
+*   Mail Spool Partitions
+*   Metadata Partitions
+*   Backup Partitions
+*   Archive Partitions
+*   Search Partitions
+
+Each of these are discussed in their own sections of the documentation.
+All share in common a few aspects in how they are configured.  For each
+partition defined, you must tell Cyrus where the partition is rooted in
+the filesystem.  This is accomplished via a "*partition*-*name*:" directive
+for each partition, where "*partition*" specifies the partition type and
+"*name*" is the actual name of the partition.
+
+Here are some sample declarations of each different type of partition
+supported within Cyrus.  For the purposes of this example, we'll
+stipulate the following:
+
+*   There are two main storage pools, "fast" and "slow," which are on
+    SSDs and traditional disks, respectively.
+*   The mailboxes are divided alphabetically, with A to M in one group
+    and N to Z in the other.
+
+.. parsed-literal::
+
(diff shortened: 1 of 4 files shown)


docsrc/imap/reference/admin/locations.rst
```diff
@@ -35,6 +35,55 @@ In general, Cyrus allows one to maintain several separate partitions for
...
+
+    # The Mail Spool Partitions
+    partition-atom: /var/spool/cyrus/fast/mail/atom/
+    partition-ntoz: /var/spool/cyrus/fast/mail/ntoz/
+
+    # The Metadata Partitions
+    metapartition-atom: /var/spool/cyrus/fast/meta/atom/
+    metapartition-ntoz: /var/spool/cyrus/fast/meta/ntoz/
+
+    # Backup Partition
```
elliefm: I think having a backup partition entry specified here is misleading.  This is for the backup service, which should generally not be on the same Cyrus instance as your regular mail partitions.

Also, the backup service is experimental in 3.0.x anyway, so I don't think we want to encourage cargo-culting of configuration for it.

nicolan:
Looks ok to me. I'll leave it to @elliefm to provide final approval on her requested changes.

elliefm:
Thanks!

Any thoughts on whether this update is suitable for back porting to the stable branch?  It seems reasonable to do so, but I can't remember if there's a good reason not to.  Maybe it could be confusing to people who set up their 3.0 system before it changed?  @postilion your input would be particularly appreciated here since of the three of us, you're the one who actually uses these docs as a Cyrus admin!

hagedose:
FWIW, I'm in favor of back porting these changes. I don't think they will cause confusion, and they will help sysadmins who are yet to make the switch to 3.0.

postilion:
@elliefm Yes, please do backport this to 3.0  I would do so, but given that my Git skills present a clear and present danger to the codebase, I think this is better handled by you. :-)

Seriously, I was totally confounded by the documentation on these subjects, which is particularly embarrassing given that I wrote (or at least massaged) much of it.  The language was just too confusing.  There's still more work to be done, I will note, to remove the use of "default" as an example default, or of "archive" as an example archive partition name.
-nic

elliefm:
Doc changes have been back ported to 3.0 :)

Issue #2182, opened 2017-10-23
Merge all the different things that call 'sendmail' to a common API


brong:
e.g. sieve notify, JMAP, calendar notifies.

rsto:
I've spent some time looking at the current codebase (I'm writing this up here since we'll need it during development):

Currently, Cyrus supports two main flows how to send mail: either by executing sendmail in "normal" mode (other than -bs) or by using the smptclient API that I introduced for JMAP mail. The latter is a thin wrapper over either a TCP-based or sendmail-backed SMTP connection.

The configuration by the former approach is via the cyrus.imapd `sendmail` option, where as the latter is configured by `jmap_smtp_backend`. Ideally, we'd just have one option to configure how Cyrus sends mail and just one way to do so.

That requires us to make use of SMTP backed email sending across all places in Cyrus, be that CalDAV scheduling, Sieve, NNTP or JMAP. But the current smptclient API as used in JMAP is too low level for this!

Especially the detection and handling of SMTP extensions is crude (if not to say: broken. Note how e.g. the hardcoded check for the AUTH extension via the string `250-AUTH` [here](https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/jmap_mail.c#L8377)).
We'll need better SMTP extension support for DSNs, at least.

Another data point: Note how currently the JMAP mail module takes care of properly dot-escaping mail messages ([here](https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/jmap_mail.c#L8432)).

TL;DR: We shouldn't just refactor the current sendmail-backed mail sending modules to make use of SMTP, but rework the smtpclient API into a nicer, high-level API while doing so.

rsto:
This randomly-assigned diceroll is highly relevant: https://github.com/cyrusimap/cyrus-imapd/issues/1637

rsto:
This will land as https://github.com/cyrusimap/cyrus-imapd/pull/2225

rsto:
Fixed with https://github.com/cyrusimap/cyrus-imapd/commit/e0d4a90c1787709578a04f4c236463bc41c82858 and https://github.com/cyrusimap/cassandane/commit/2519d4abbc618610576d016a98005586076f5ffc

PR #2225, opened 2018-01-02
Use SMTP client to send mails

rsto:
This patch rewrites the smtpclient API to use a generic SMTP backend for sending mail.

Up until now, Cyrus forked in the sendmail binary as specified in the `sendmail` imapd.conf option to send email. The new implementation is agnostic of the underlying SMTP backend, using Cyrus protocol streams (lib/prot.h) to abstract the transport.

The new implementation provides TCP-based and sendmail-process based SMTP backends.

Also see: https://github.com/cyrusimap/cyrus-imapd/issues/2182


rsto:
@ksmurchison May I ask you to review at least the refactored CalDAV, NNTP and Sieve code? I had no chance to test them properly, do you? I might need to set up Cassandane tests for them.

One notable difference is that previously these modules used files to read/compose the email. I have rewritten the code to keep the composed emails in memory until they are sent by the SMTP client. In my understanding the message templates are small, so this shouldn't be a memory issue?

rsto:
> Big question: I wonder if it makes sense for the "host" backend to leverage code that already exists in backend.c ?
> Smaller question: Do we want to standardize the "host" backend on port 587 (SMTP Submission) rather than port 25?

I think the answers to these questions are related. Port 587 -> absolutely! But this brings up the question: do we want to support STARTTLS if the host supports it?

Regarding backend.c: What I like about the current design of smtpclient is that all it reads from and writes to is a lib/prot byte stream. That should make it very simple to plug in any other SMTP backends, and it already came handy with the sendmail and host backends. I don't know enough yet about the backend module if it will allow to keep this level of abstraction.

On the other hand, the STARTTLS support in protocol.h looks promising and will dig deeper in the code to understand I we can reuse it.

Pinging @brong in case he finds time to chime in.

ksmurchison:
backend.c is a generic framework for the client-side of all of the protocols used in Cyrus.  It can parse capabilities and do STARTTLS and authentication.  We already have a profile for LMTP, so I'm sure adding SMTP would be trivial.

imap/http_caldav_sched.c
```diff
@@ -322,23 +321,6 @@ static int imip_send_sendmail(icalcomponent *ical,
         }
     }
 
-    argv[0] = "sendmail";
```
ajaysusarla: Thanks for remove this piece of code.

rsto:
> backend.c is a generic framework for the client-side of all of the protocols used in Cyrus. It can parse capabilities and do STARTTLS and authentication.

The proposed smtpclient uses prot.h readers/writers to abstract away the underlying backend, be that a forked sendmail process or a TCP host. The backend.h module seems to go the other way round: it sits on top of the prot streams. Would be able to plug in a forked sendmail process under it? It's what most Cyrus installations will use.

This task could actually be a great opportunity for our pair programming session!

ksmurchison:
Yes, we could certainly look into this in a pairing session.

rsto:
@ksmurchison could you please have a look at https://github.com/cyrusimap/cyrus-imapd/pull/2225/commits/e07ac938dfe350b2183b5d54dbbd4ea851f31bb4 ?

Is that all there is to add SASL auth to backend connections? I don't have a setup to test this locally.

ksmurchison:
@rsto That looks about right to me.  That's what mupdate-client.c does.

Issue #2120, opened 2017-08-25
syncserver error handling should be better


hagedose:
I have a production server running 2.4.20 and a testing server with 3.0.3. The production server has many partitions, whereas the testing server currently only has the default partition.

I created my account on the testing server and appended a few messages to the INBOX. Later I wanted to experiment with replication and first tried to manually sync my inbox like this:

```
sync_client -vm user/a0620
MAILBOXES user.a0620
Error from do_mailboxes(): bailing out!
```

On the testing server I saw this in the log:

```
Aug 25 00:39:56 cyrus3-test cyrus/syncserver[17457]: conversations_rename_folder: deleted user.a0620
Aug 25 00:39:56 cyrus3-test cyrus/syncserver[17457]: Deleted mailbox user.a0620
Aug 25 00:40:19 cyrus3-test cyrus/syncserver[17457]: Fatal error: Internal error: assertion failed: imap/dlist.c: 156: base != NULL

```
First, I found it a little surprising that the existing mailbox is simply deleted. Then the error message isn't very helpful unless you’re familiar with server internals. What's more, afterwards there was a broken entry in mailboxes.db:

```
# ctl_mboxlist -d
user.a0620	16 (null)
```

I figured out that the underlying problem is that my mailbox is on a partition that doesn't exist on the testing server. Could the error handling for this scenario be improved? Is the requirement for partitions to be identical on all replica servers even documented? I might have overlooked it, but I don't recall seeing it.

hagedose:
FWIW, I just reproduced the issue with 3.0.5:

```
Jan 04 15:32:41 cyrus3-test.rrz.uni-koeln.de cyrus/syncserver[3265]: conversations_rename_folder: deleted user.a0620
Jan 04 15:32:41 cyrus3-test.rrz.uni-koeln.de cyrus/syncserver[3265]: Deleted mailbox user.a0620
Jan 04 15:32:41 cyrus3-test.rrz.uni-koeln.de cyrus/imap[3169]: graceful shutdown initiated by unexpected process 3265 (sync_server)
Jan 04 15:32:41 cyrus3-test.rrz.uni-koeln.de cyrus/master[3146]: process type:SERVICE name:imap path:/usr/lib/cyrus-imapd/imapd age:768.769s pid:3169 exited, status 75
[...]

PR #2230, opened 2018-01-05
Add note about partition requirements to SOP/replication

postilion:
Added verbiage to documentation noting the requirement that all partitions from master instance must also exist on replica instances.  Addresses #2120


nicolan:
Looks good to me!

PR #2223, opened 2017-12-22
Remove JMAP support from Cyrus IMAPd 3.0 release

rsto:
As discussed we want to get rid of the experimental JMAP implementation in 3.0. It's doing installations no favour building on a JMAP codebase that's way outdated in contrast to the current JMAP spec implementation on the master branch.

And it makes our Cassandane and Jenkins tests unnecessarily complex and fiddle.

Please check the commit message if the wording is OK with you?

@nicolan I removed JMAP from all code, but left documentation untouched. E.g. here is a list of affected files that still mention JMAP:

    doc/legacy/changes.html
    docsrc/quickstart.rst
    docsrc/overview/cyrus_roadmap.rst
    docsrc/index.rst
    docsrc/imap/developer/jmap.rst
    docsrc/imap/developer/developer-testing.rst
    docsrc/imap/developer/compiling.rst
    docsrc/imap/reference/faqs/interop-8bit.rst
    docsrc/imap/developer.rst
    docsrc/imap/concepts/features.rst
    docsrc/imap/download/upgrade.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.0-beta6.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.0-beta1.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.1.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.0-beta3.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.2.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.0-beta4.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.3.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.0.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.0-rc4.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.0-rc3.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.0-rc2.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.0-beta5.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.0-beta2.rst
    docsrc/imap/download/release-notes/3.0/x/3.0.0-rc1.rst
    docsrc/imap/download/installation/manage-dav.rst
    docsrc/imap/download/installation/http/jmap.rst
    docsrc/imap/download/packagers.rst

The only other places are:

    imap/annotate.c

which is better left untouched and

[...]

PR #2233, opened 2018-01-16
Clarify status of JMAP support

marclaporte:
https://github.com/cyrusimap/cyrus-imapd/pull/2223
https://datatracker.ietf.org/wg/jmap/documents/

### DIFF
--- a/docsrc/overview/cyrus_roadmap.rst
+++ b/docsrc/overview/cyrus_roadmap.rst
@@ -14,7 +14,7 @@ High Level Roadmap
 Future
 ------
 
-* Improved JMAP support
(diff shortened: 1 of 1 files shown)


rsto:
LGTM, thanks! @nicolan wording is OK with you?

elliefm:
I think it's effectively "not in" 3.0.x anymore.  Even if you have a version that includes it (3.0.0 - 3.0.5, plus the betas and release candidates), the implementation included in those releases is not standard, and compliant JMAP clients won't be able to talk to it.

rsto:
Thanks!

Issue #2234, opened 2018-01-22
squatter and cyrus.conf: START | EVENTS


dilyanpalauzov:
https://cyrusimap.org/imap/developer/install-xapian.html suggests putting squatter in cyrus.conf:START.

https://cyrusimap.org/imap/reference/manpages/systemcommands/squatter.html suggests putting squatter in cyrus.conf:EVENTS.

The latter should state when to put squatter in START and when in EVENTS.

elliefm:
Oh, well spotted, thanks :)

My understanding here is that you put `squatter -R` in START when you want new messages to be indexed (almost) immediately; or you put a non-`-R` invocation into EVENTS when you want indexing to occur on some schedule.

But...  https://cyrusimap.org/imap/reference/manpages/systemcommands/squatter.html also says,

> Note
> 
> Incremental updates are very inefficient with the SQUAT search engine. If using SQUAT for large and active mailboxes, you should run squatter periodically as an EVENT in cyrus.conf(5).

@rsto Does this inefficiency affect the Xapian search engine, or only the Squat engine?

And https://cyrusimap.org/imap/developer/install-xapian.html says,

> Xapian requires a running squatter(8) instance:

Does this simply mean that you need to arrange for squatter to run (either by START or EVENTS)?  Or does it mean more specifically that Xapian indexes can only be built in `-R` mode?

elliefm:
(I'll update the docs once I'm clear on my understanding.)

elliefm:
If you're using multiple search tiers, you'll also need to periodically run squatter with the -t and -z arguments to compact them, so even if you were primarily indexing in rolling mode, you would also want some EVENT entries for this.  I think multiple search tiers are only supported for the Xapian engine.

I've just re-read the old email that the install-xapian doc references, and it says:

> We built a C API wrapper around the Xapian features we needed, and tried to run in production.  Unfortunately, the IO hit from indexing new messages in real time was too much - our server couldn't cope.

So it sounds like the rolling mode performance issue is a problem with the Xapian engine too, not just Squat.  So I guess our recommendation would be something like:

* if you want immediate index updates: use `squatter -R` in START, but make sure your `defaultsearchtier` is on fast storage, and then use compact jobs running from EVENTS (or cron, etc) to periodically move indexes to larger/slower storage
* if you don't need immediate index updates: just run `squatter` from EVENTS

[...]

PR #2235, opened 2018-01-23
Update squatter.8 manpage for clarity

postilion:
The squatter.8 man page was vague about differences between the
squatter command and the old SQUAT search engine.  Comments have
been added to clarify this.

Notes have been added or strengthened in regards to SQUAT vs.
Xapian indexing; requirements for rolling operation; where in
cyrus.conf to invoke squatter; etc.

The example for incremental operation used the wrong flag, -s
rather than -i.

Addresses #2234


elliefm:
Great, thanks! Will cherry-pick this to 3.0 as well

elliefm:
Hmm, the pre-existing documentation for -Z says it's new in v3, but the implementation only exists on master.  Trying to decide whether I should treat the docs as correct in intent, and backport the patch, or treat the code as correct, and fix the docs.  (Noticed this because the difference makes it not cherry-pick cleanly.)

I'm currently leaning towards adding -Z to 3.0, but will sit with it a bit longer I think.

elliefm:
I've done a bit of experimenting, and backporting -Z (and -U) to 3.0 are not trivial, so I think I'm going to skip them and just fix the doc conflicts instead. :)

PR #2241, opened 2018-02-06
imapoptions: improve docs for sync_log_channels

elliefm:
How's this look, wrt discussion in #2234?

### DIFF
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -2255,11 +2255,12 @@ product version in the capabilities
 
 { "sync_log_channels", NULL, STRING }
 /* If specified, log all events to multiple log files in directories
-   specified by each "channel".  To run these log files, you need to pass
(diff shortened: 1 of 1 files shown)


postilion:
LGTM.  In earlier discussion you mentioned backups.  Does the backup scheme also consume sync_logs to figure out what to do?  If so, does it bear mentioning yet, or is that a future thing?  Just curious.

Thanks!
-nic

elliefm:
The backup server implements the csync protocol, so you send it stuff to be backed up using sync_client.  The anticipated way to configure it would be to set up a dedicated sync log channel, whose sync_client talks to the backup server rather than to a conventional replica.  I'm pretty sure this is covered in depth in the backup documentation, but development of it has stalled out pending updates to the csync protocol.  In the meantime I'd prefer to avoid confusion. :)

PR #2243, opened 2018-02-06
squatter: get rid of undocumented '-c' argument

### LINKED ISSUE #2242: squatter: get rid of undocumented -c option
As far as I can tell, this existed mainly so that the Sphinx search engine could be handheld through its boot process, and also so that squatter would shut down when master does.  

Since the Sphinx engine is gone on master, and the shut down problem has been solved by the DAEMON section of cyrus.conf, we should be able to throw away the code supporting the `-c` option as well?

elliefm:
This was needed when the Sphinx engine had its own separate
daemon that needed hand-holding.  The Sphinx engine is no more.

Closes #2242

### DIFF
--- a/imap/search_engines.c
+++ b/imap/search_engines.c
@@ -81,8 +81,6 @@ static const struct search_engine default_search_engine = {
     NULL,
     NULL,
     NULL,
-    NULL,
-    NULL,
 };
 
 static const struct search_engine *engine(void)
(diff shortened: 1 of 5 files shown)


elliefm:
(This cannot be backported to 3.0 because 3.0 still contains the Sphinx engine.)

elliefm:
(rebased on master, conflicts resolved)

PR #2244, opened 2018-02-07
prototype mappedfile_discard() api

elliefm:
A possibility for dealing with IOERROR during mappedfile operations.  If it's not an abortable file format, and fails mid-write, it's probably corrupt.  This function is for the "just throw it away then" solution.

### DIFF
--- a/imap/prometheus.c
+++ b/imap/prometheus.c
@@ -136,8 +136,8 @@ static void prometheus_init(void)
 
     r = mappedfile_pwrite(handle->mf, &stats, sizeof(stats), 0);
     if (r != sizeof(stats)) {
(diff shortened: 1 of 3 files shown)


elliefm:
I am tempted to add an `assert(mf->lock_status != MF_READLOCKED)` near the top of `mappedfile_discard()` too, but I'm on the fence about it.  We need to support mappedfile's that don't get locked at all (no prob), but if you're holding a read lock that implies this file needs locking, and you're not holding a strong enough lock to permit unlinking it.

On the other hand, the sky is probably falling, and we want to clean up after ourselves without leaving corruption behind, not sook about the situation being insufficiently perfect.  Insights appreciated.

Issue #2248, opened 2018-02-14
Generalized use of sync_log produces unintended consequences


postilion:
From a report to cyrus-info by Per olof Ljungmark <lists@nethead.se>, in a discussion about Xapian and "rolling indexing" mode (-R) of squatter:

> 
> One more thing though, if replication is involved, it appears one needs one log for squatter and another for replication, am I correct? I got replication to work only after I added another log, like,
> 
> sync_log_channels: squatter replog
> and
> syncclient       cmd="/usr/local/cyrus/sbin/sync_client -r -n replog"
> 
> Not sure if this is mentioned in the docs somewhere.

If the presence of squatter as a sync_log_channel effectively deletes the "dafault" sync_log channel (specified as "" in imapd.conf(5)) then we must be clearer about this.

elliefm:
We discussed this on today's meeting and the verdict was to get rid of the magical "" channel for replication, and make all replication users configure sync_log_channels in some way (possibly with a named default channel).  I'll put together a PR for this.  It won't be backported to 3.0.

rsto:
> I'll put together a PR for this. It won't be backported to 3.0.

@elliefm has the PR been merged and just this issue not closed?

elliefm:
I can't find a PR for it, and can't even find a branch for it.  Maybe I got distracted and never started work on it?

lib/imapoptions still describes the magic `""` behaviour so I guess that part hasn't changed, but maybe it did change and the documentation wasn't updated?  Actually, looks like I updated that documentation in 70ca011fbe6f, which was part of the work for #2234, which was also just closed as stale.

Gonna close this one, since I think the documentation changes we intended have been done, and I no longer remember anything about my earlier intention to get rid of `""`.

PR #2249, opened 2018-02-14
Clarify documentation regarding default sync_log

postilion:
The Cyrus IMAP replication subsystem assumes a default sync_log
channel named "" (the two-character string U+22 U+22).  If one modifies
the sync_log_channels: directive with a name for another sync_log
channel, such as "squatter" then the default "" channel vanishes, unless
it is also specified explicitly. [issue #2248]

This commit is a first draft at making this clear.  Changes are made to
both the manpage for squatter(8) and to the Admin Guide, Replication
pages.

I feel that we actually need to produce more generalized documentation
of the sync_log mechanism, and then link to that from all points
which deal with sync_log usage, as these will no doubt grow due to the
usefullness of the sync_log functionality.

Addresses #2248.

### DIFF
--- a/docsrc/imap/reference/admin/sop/replication.rst
+++ b/docsrc/imap/reference/admin/sop/replication.rst
@@ -226,6 +226,12 @@ it in :cyrusman:`imapd.conf(5)` using these options:
     See :ref:`replication-channels`, below, for details on how to use
     these settings to control syncing to multiple replicas.
 
+.. Important::
+    If using sync_log_channels for any other purpose, such as
+    specifying the sync_log used by :cyrusman:`squatter(8)` command,
+    you must *also* either specify a sync_log channel for replication,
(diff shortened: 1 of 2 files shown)


elliefm:
> This commit is a first draft

Did you want me to merge this one as it is now, or leave it open for now for further iteration?

postilion:
@elliefm Yes, please go ahead with this as it sits.  I plan to circle back around on this once I have a better sense of how to handle this topic, these topics, in a more holistic fashion.  I think I may need to parlay with @nicolan on the topic, and maybe @brong, too.

elliefm:
I've cherry-picked the commit to the 3.0 branch too :)

postilion:
@elliefm Thanks for inclusion in 3.0.  I really should have mentioned that I think this is of utmost importance, as it's a potential show-stopper for anyone trying to implement both replication & rolling replication at the same time.

Issue #2250, opened 2018-02-16
squatter manpage synopsis is incomplete


hagedose:
In version 3.0.5 the manage for squatter misses at least one optional flag for the seventh synopsis (compacting). You can compact just one user by specifying "-u",

PR #2252, opened 2018-02-19
More updates to squatter.rst to clarify operations.

postilion:
Add documentation for -Z to usage message in squatter.c
In squatter.rst:
 * Update synopsis to include all instances where -u, -S and -Z may be
   used.
 * clarify freshness of indexes for different scenarios.
 * clear up some language
 * detail which options are specific to Xapian
 * add a note to the HTML version referring users to install-xapian
   documentation.

Generalize description of "/vendor/cmu/cyrus-imapd/squat" annotation in
the FAQ.
Updated install-xapian.rst to clear up some "default"-isms.
Update cyrus.conf.rst to clarify the nanny aspect of master(8) for
those items listed in the DAEMONS section of cyrus.conf(5).

Addresses #2234 & #2250

### DIFF
--- a/docsrc/imap/developer/install-xapian.rst
+++ b/docsrc/imap/developer/install-xapian.rst
@@ -61,10 +61,10 @@ You also need (at least one) search tier. Add this to :cyrusman:`imapd.conf(5)`.
     search_engine: xapian
     search_index_headers: no
     search_batchsize: 8192
-    defaultpartition: default
+    defaultpartition: base
     defaultsearchtier: t1
-    partition-default: /var/cyrus/spool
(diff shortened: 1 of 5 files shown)


elliefm:
Oh, might be worth noting that the `-S` option accepts non-integer seconds (e.g. "1.5" etc).

elliefm:
(The duration is converted down to microseconds internally, but its granularity is further subject to the peculiarities of floating point representation and your system's clock resolution.  So, "0.000001" probably _won't_ wait exactly one microsecond, but say "0.25" will almost certainly wait about a quarter-second.  This is probably too much info for docs, but maybe gives you background context to help choose a phrasing?)

PR #2254, opened 2018-02-20
More doc edits: squatter.rst & cyrus.conf.rst

postilion:
Add missing option -S to synopsis 1 in squatter.rst
Add note on effect of -D option for master(8) to cyrus.conf.rst.

Addresses further issues in issue #2234

### DIFF
--- a/docsrc/imap/reference/manpages/configs/cyrus.conf.rst
+++ b/docsrc/imap/reference/manpages/configs/cyrus.conf.rst
@@ -55,6 +55,13 @@ processes listed in **START** are expected to end and clean up after
 themselves, while those in **DAEMON** will be shutdown by
 :cyrusman:`master(8)`.
 
+.. Note::
+
+    If :cyrusman:`master(8)` is started in debugging mode (**-D**) the
+    behavior of **DAEMON** will be altered, as :cyrusman:`master(8)`
+    will no longer be backgrounded.  Thus, processes started under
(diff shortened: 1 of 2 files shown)


docsrc/imap/reference/manpages/configs/cyrus.conf.rst
```diff
@@ -55,6 +55,13 @@ processes listed in **START** are expected to end and clean up after
 themselves, while those in **DAEMON** will be shutdown by
```
elliefm: I think I'd like to see this part say something like,

> Similarly, processes listed in START will be run to completion before any SERVICES are started, while those in DAEMON will be managed by master.

(n.b. one of the things DAEMON does that START does not is restart the process if it exits early or crashes, just like a SERVICE minus the socket.  It's not just about shutdowns.)

elliefm: Similarly, down under the START heading (github won't let me add a comment there because it's too far away from the actual diff), maybe get rid of the "or backgrounded themselves"?  And then add some sort of "historical note" that prior to v3, non-service daemons like idled were started from START but would background themselves, but now they are better managed through the DAEMON section?

postilion: @elliefm This has been fixed in d7ff32f

Thanks!
-nic

elliefm:
I've cherry-picked the cyrus.conf.rst changes to 3.0.

PR #2268, opened 2018-02-27
Check if mechlist buffer is not empty

landgraf:
Check if mechlist is null is redundant because it's initialized in
ask_capabilities flow. As the result some capabilities may be lost and
reported as not advertised
Keeping (now reduntant) null check to avoid regressions in the future
This is regression in 9fd201ba2b4ab58eda3372fb6765e1d5d8f027b4

Bug-Url: https://bugzilla.redhat.com/1543481


elliefm:
Thanks! :)

elliefm:
This has been cherry-picked to the 3.0 branch as e6bf267dd

PR #2277, opened 2018-03-02
sync improvements

elliefm:
There's more to do here still, but I'll open this PR now so it can be reviewed iteratively.

This first stage bundles MAILBOX updates by user, instead of processing them purely in the order they were seen.  Within each user, their MAILBOX order is preserved (though this doesn't make much difference in practice, I don't think: the whole list ends up as a single GET MAILBOXES call further down the stack).  However, the user list is processed in no particular order (i.e. hash table enumeration order -- I don't think this is a problem: we're already making user-ordering fairly meaningless by clustering mailboxes by user).

Instead of splitting operations into batches of exactly 1000 mailboxes, we collect user/mailbox operations one-whole-user at a time until we've accumulated 1000+ mailboxes, and then send that batch (rinse/repeat until the user/mailbox list is exhausted).  This means that when a user's mailbox is renamed, the updates to both the oldname and newname are always performed in the same batch.

Non-user mailboxes are batched as if they all belong to a single user called "".

So far, this PR only affects the batching process, the actual replication is unchanged.

There's still work to be done to better cope with mailbox renames between users, to clarify the handling of UNMAILBOX, and to automatically promote failed MAILBOX updates into USER updates.

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -3034,7 +3034,7 @@ index_copy(struct index_state *state,
 
     /* we log the first name to get GUID-copy magic */
     if (!r)
-        sync_log_mailbox_double(index_mboxname(state), name);
+        sync_log_rename(index_mboxname(state), name);
 
 done:
     free(copyargs.records);
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -7425,7 +7425,7 @@ static int _copy_msgrecords(struct auth_state *authstate,
     r = append_commit(&as);
     if (r) goto done;
 
-    sync_log_mailbox_double(src->name, dst->name);
+    sync_log_rename(src->name, dst->name);
 done:
     return r;
 }
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -2456,7 +2456,7 @@ EXPORTED int mboxlist_renamemailbox(const mbentry_t *mbentry,
              * we never nuke the mailbox from the replica before realising
              * that it has been renamed.  This can be moved later again when
              * we sync mailboxes by uniqueid rather than name... */
-            sync_log_mailbox_double(oldname, newname);
+            sync_log_rename(oldname, newname);
 
             mailbox_rename_cleanup(&oldmailbox, isusermbox);
 
--- a/imap/sync_client.c
+++ b/imap/sync_client.c
@@ -379,9 +379,8 @@ static int do_sync(sync_log_reader_t *slr, const char **channelp)
             report_verbose("  Deferred: QUOTA %s\n", action->name);
         }
         else if (r) {
-            /* XXX - bogus handling, should be user */
-            sync_action_list_add(mailbox_list, action->name, NULL);
-            report_verbose("  Promoting: QUOTA %s -> MAILBOX %s\n",
+            sync_action_list_add(user_list, action->name, NULL);
+            report_verbose("  Promoting: QUOTA %s -> USER %s\n",
                            action->name, action->name);
         }
     }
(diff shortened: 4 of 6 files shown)


elliefm:
I've rebased onto current master, and added one new commit, which fixes renames of mailboxes between users to not be split up by the batching process.

It does this by changing the sync log entry for a rename from being two separate MAILBOX records (one for oldname, one for newname) into a single MAILBOX record with both the oldname and newname.

When a MAILBOX record is encountered with only a single mboxname, it is treated normally, and will be batched based on the user who owns it.  When a MAILBOX record containing two mboxnames is encountered, they will both be batched according to the owner of the first user.

(Note that this still only affects batching of requests, there's no changes to the actual replication, so there's no "it's the wrong user" issue here.)

elliefm:
Latest commit replaces the short-lived "MAILBOX old new" syntax with "RENAME old new", restoring MAILBOX to only taking one argument.

In the case where the userids for the old and new names differ, the new RENAME syntax batches both names against both users.   Whichever one hash_enumerate() pulls out first will prompt the rename to occur, and whichever one comes out second will see an up to date state and do nothing.

elliefm:
Latest commit (from late yesterday) makes sure `UNMAILBOX mboxname` is only sent to the replica if the master has a tombstone for it.  But I'm not sure what to do when there's no tombstone?

Looks like tombstones get cleared after, not-configurably, 7 days (by `cyr_expire`).  So you'd expect there to be a tombstone most of the time, but if your replication is stale for some reason it might be already gone.

brong:
On Tue, 6 Mar 2018, at 11:19, elliefm wrote:
> Latest commit (from late yesterday) makes sure UNMAILBOX mboxname is
> only sent to the replica if the master has a tombstone for it.  But
> I'm not sure what to do when there's no tombstone?
Yep, so if there's no tombstone, what we should be doing is creating the
mailbox locally and copying the messages back from the replica into it!
> Looks like tombstones get cleared after, not-configurably, 7 days (by
> cyr_expire).  So you'd expect there to be a tombstone most of the
> time, but if your replication is stale for some reason it might be
> already gone.

That's a risk.  It's hopefully pretty rare that replication is that far
behind - but maybe we should at least make it configurable!
Bron.

--
  brong@fastmailteam.com

elliefm:
> Yep, so if there's no tombstone, what we should be doing is creating the
> mailbox locally and copying the messages back from the replica into it!

That makes sense, but I don't think this code path is the place for that?  We only end up here if we saw an `UNMAILBOX foo` in the sync log, and we _only_ get such a line from `mboxlist_deletemailbox()`.

Not sure where the right spot would be though, all our replication code currently expects to advance the replica's state to match master, not so much the other way round.

> That's a risk.  It's hopefully pretty rare that replication is that far
> behind - but maybe we should at least make it configurable!

Yeah, and also make it a distinct argument to cyr_expire I think, or at least document it properly.  Currently it's just baked into a code path that iirc the manpage claims is for cleaning the duplicate db. :/

elliefm:
Ahhh, `do_folders()` in sync_support.c feels like it might be the place for it (that's the function that compares master_folders against replica_folders).  Here: https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/sync_support.c#L5665-L5676

At the very least, I think this bit should also be looking for looking for tombstones too, not just wildly deleting anything it couldn't find locally.  Not sure if that should be done here, or up in reserve_messages where the master_folders/replica_folders lists are populated.

elliefm:
rebased on current master, fixed conflict

elliefm:
rebased again on current master (no conflicts)

PR #2307, opened 2018-04-05
Update http_dav.c

karagian:
irock->owner can be null in the case of public collections. This causes httpd to segfault.
problem can be reproduced when using a shared calendar url as the account url from apple devices

### DIFF
--- a/imap/http_dav.c
+++ b/imap/http_dav.c
@@ -10039,7 +10039,7 @@ static void xml_add_sharee(const char *userid, void *data, void *rock)
     int isadmin;
 
     if ((rights & DACL_SHARE) != DACL_SHARE) return;  /* not shared */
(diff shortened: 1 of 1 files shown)


karagian:
I think this should also be on the upcoming 3.0 stable release... @elliefm

elliefm:
This looks good to me, but I don't know the DAV codebase. @ksmurchison, anything we need to watch out for here?

dilyanpalauzov:
Does the problem still exist?

dilyanpalauzov:
I mean, does the problem exist in the 3.2 codebase.

rsto:
This looks to be fixed in

d680f10b2c9 (Ken Murchison   2019-02-08 17:08:12 -0500 1585)         irock.owner = "";

PR #2318, opened 2018-04-16
Include squatter.8 if SQUATTER, not if USE_SQUAT

uutari:
The man page was (incorrectly) only included when squat was compiled in, not if squatter was being used e.g. with another search engine.


elliefm:
Looks good, thanks! :)

PR #2349, opened 2018-05-09
Split the system_flags field in struct index_record.

### LINKED ISSUE #2331: Split the system_flags field in struct index_record
Right now, the system_flags field is a 32 bit field as follows:

typedef enum _MsgFlags {
    FLAG_ANSWERED           = (1<<0),
    FLAG_FLAGGED            = (1<<1),
    FLAG_DELETED            = (1<<2),
    FLAG_DRAFT              = (1<<3),
    FLAG_SEEN               = (1<<4),
    FLAG_SPLITCONVERSATION  = (1<<27),
    FLAG_NEEDS_CLEANUP      = (1<<28),
    FLAG_ARCHIVED           = (1<<29),
    FLAG_UNLINKED           = (1<<30),
    FLAG_EXPUNGED           = (1U<<31),
} MsgFlags;

The first 5 are the real system flags, the other 5 are internal bookkeeping.  In particular, FLAG_ARCHIVED and FLAG_UNLINKED have semantics about how the message is handled on disk.

Lots of code that manipulates system_flags has to remember about these different meanings.  It would make more sense to keep system_flags for just those first 5 again, and for the rest, create unsigned int foo:1 for each.

Those fields could be serialised and deserialised then in mailbox_buf_to_index_record and mailbox_index_record_to_buf without needing to handle the fact that they're stored in the same 32 bit field ANYWHERE ELSE in the code any more!

ajaysusarla:
The `system_flags` field in `struct index_record` is used to store the
following flags on disk:

-- System flags:
FLAG_ANSWERED
FLAG_FLAGGED
FLAG_DELETED
FLAG_DRAFT
FLAG_SEEN

-- Internal flags for bookeeping:
FLAG_SPLITCONVERSATION
FLAG_NEEDS_CLEANUP
FLAG_ARCHIVED
FLAG_UNLINKED
FLAG_EXPUNGED

This patch, removes the use of `system_flags` to store the internal
flags in memory. Instead, we now have several bit fields in `struct
index_record` representing the internal flags.

The on-disk representation continues to be the same.

The methods `mailbox_buf_to_index_record()` and
`mailbox_index_record_to_buf()` serialise and deserialise these fields.

Fixes #2331

### DIFF
--- a/imap/index.h
+++ b/imap/index.h
@@ -96,11 +96,12 @@ struct index_init {
 struct index_map {
     modseq_t modseq;
     modseq_t told_modseq;
+    uint64_t cache_offset;
+    uint32_t user_flags[MAX_USER_FLAGS/32];
     uint32_t uid;
     uint32_t recno;
     uint32_t system_flags;
-    uint64_t cache_offset;
-    uint32_t user_flags[MAX_USER_FLAGS/32];
+    uint32_t internal_flags;
     unsigned int isseen:1;
     unsigned int isrecent:1;
 };
@@ -159,6 +160,7 @@ typedef struct msgdata {
     bit32 hasflag;              /* hasflag values (up to 32 of them) */
     struct message_guid guid;   /* message guid */
     uint32_t system_flags;      /* system flags */
+    uint32_t internal_flags;    /* internal flags */
 
     /* items from the conversations database */
     modseq_t convmodseq;        /* modseq of conversation */
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -3442,7 +3442,7 @@ static int _email_mailboxes_cb(const conv_guidrec_t *rec, void *rock)
     jmap_req_t *req = data->req;
     struct mailbox *mbox = NULL;
     msgrecord_t *mr = NULL;
-    uint32_t flags;
+    uint32_t system_flags, internal_flags;
     int r;
 
     if (rec->part) return 0;
@@ -3453,10 +3453,13 @@ static int _email_mailboxes_cb(const conv_guidrec_t *rec, void *rock)
     r = msgrecord_find(mbox, rec->uid, &mr);
     if (r) goto done;
 
-    r = msgrecord_get_systemflags(mr, &flags);
+    r = msgrecord_get_systemflags(mr, &system_flags);
     if (r) goto done;
 
-    if (!r && !(flags & (FLAG_EXPUNGED|FLAG_DELETED))) {
+    r = msgrecord_get_internalflags(mr, &internal_flags);
+    if (r) goto done;
+
+    if (!r && !(system_flags & FLAG_DELETED || internal_flags & FLAG_INTERNAL_EXPUNGED)) {
         json_object_set_new(mboxs, mbox->uniqueid, json_string(mbox->name));
     }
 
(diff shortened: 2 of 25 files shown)


imap/index.h
```diff
@@ -96,13 +96,19 @@ struct index_init {
...
+    uint32_t user_flags[MAX_USER_FLAGS/32];
     uint32_t uid;
     uint32_t recno;
     uint32_t system_flags;
-    uint64_t cache_offset;
-    uint32_t user_flags[MAX_USER_FLAGS/32];
     unsigned int isseen:1;
     unsigned int isrecent:1;
+    /* Internal flags */
+    uint32_t flag_splitconversation:1;
```
rsto: Why use bitfields here rather than keeping it a bit mask? Personally, I vouch for using a single uint32_t. But probably I'm not seeing the use case here?
E.g. looking at index_reload_record it looks more cumbersome to copy one flag after the other and easy to miss when adding a new flag?

imap/jmap_mail.c
```diff
@@ -9400,15 +9417,17 @@ static int _email_flagupdate_cb(const conv_guidrec_t *rec, void *rock)
...
     r = msgrecord_find(mbox, rec->uid, &mrw);
     if (r) goto done;
     r = msgrecord_get_systemflags(mrw, &system_flags);
     if (r) goto done;
-    if (system_flags & (FLAG_EXPUNGED|FLAG_DELETED)) goto done;
+    r = msgrecord_get_expungedflag(mrw, &expunged);
+    if (r) goto done;
+    if ((system_flags & FLAG_DELETED) || expunged) goto done;
 
     /* Determine if to patch or reset flags */
```
rsto: I wonder if there's any unwanted side-effects here? Before, the JMAP code copied all system flags, including the internal ones. Now, it looks as if it's just copying the non-internal flags. Which _could_ be the right thing to do. But I'm not entirely sure. Did the JMAP Mail tests all pass in Cassandane with this PR?

imap/index.h
ajaysusarla: @rsto I really don't have a preference for one over the other. The intent was to make the internal flags obvious. My quick test interms of the number of instructions generated by the compiler in both cases are similar(without optimisation even)

imap/jmap_mail.c
ajaysusarla: There are a few failures that seem unrelated, so I'm working on them. I wanted your review on the patch while I figure out the issues with Cassandane.

ajaysusarla:
@rsto I've changed the patch to use a `uint32_t` for storing the internal flags, instead of a bit fields. Also, the JMAPMail  and ImapTest suites in Cassandane pass. Please review when possible. 
Thank you!

imap/mailbox.h
```diff
@@ -383,18 +384,18 @@ typedef enum _MsgFlags {
     FLAG_DELETED            = (1<<2),
     FLAG_DRAFT              = (1<<3),
     FLAG_SEEN               = (1<<4),
+} MsgFlags;
+
+typedef enum _MsgInternalFlags {
```
rsto: I wonder if these should be called FLAG_INTERNAL_XXX to make clear which flag goes to which bitmask. I know understand that for this particular concern, a bitfield would indeed have been less error-prone. Still, I think it's worth keeping the bitmask, but probably making clear which FLAG_ needs to go to which flag mask would bring value?

ajaysusarla: I do agree with that. I shall update the patch to reflect that.

ajaysusarla:
@rsto updated the patch with the `INTERNAL` in the naming of the internal fields. Please review when you can!

rsto:
LGTM

ajaysusarla:
Thanks @rsto - I will merge it onto master

PR #2432, opened 2018-07-10
cyrusdb_lmdb.c: Fixes a repeated typo cryusdb

simonamor:
Mentioned by kworr on freenode #cyrus

Most of the syslog lines in the file were already cyrusdb but a few said cryusdb. This PR makes them all the same.


rsto:
Thanks. I thought we nuked experimental LMDB support, though? @elliefm I guess we kept lmdb in the 3.0 branch for backwards compatibility?

elliefm:
Yep, LMDB support is totally gone from master.

It hasn't been removed from 3.0 because I didn't want to rip it out from under anybody, but it still might be removed in the future (since it's unstable and unmaintained).

I'll merge this for now, so that anyone experimenting with it can at least find their log lines with grep :)

PR #2438, opened 2018-07-14
fix performance issue in charset header parser

cjnewman:
fix performance issue in charset header parser when header is long and contains many encoded words, also fix non-portable use of ulong and declaration of variable in conversions.h that should be in conversions.c

### DIFF
--- a/imap/conversations.c
+++ b/imap/conversations.c
@@ -100,6 +100,8 @@
 
 #define CONVERSATIONS_VERSION 0
 
+struct conversations_open *open_conversations;
+
 static conv_status_t NULLSTATUS = { 0, 0, 0};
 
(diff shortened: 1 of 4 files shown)


cjnewman:
Here's the automated robot test I used in our test framework; I wasn't able to get Cyrus test framework working in a reasonable amount of time, so I can't contribute a test at this time.

I18N005 append long MIME encoded header
    ${mimehdr} =    Repeat String     ${SPACE}=?utf-8?Q?=F0=9F=98=B3pointlessly_long_Subject_header_repeating_and_repeating?=\n    185000
    IMAP Append    INBOX    Subject: ${mimehdr}\nbody

PR #2482, opened 2018-08-21
[zeroskip] Support for Zeroskip DB.

ajaysusarla:
This commit adds support for Zeroskip to Cyrus DB. When Cyrus is build with Zeroskip, all the unit tests pass. There are issues with Cassandane, which I'm going to fix soon.

### DIFF
--- a/lib/bsearch.c
+++ b/lib/bsearch.c
@@ -214,6 +214,40 @@ HIDDEN int bsearch_ncompare_mbox(const char *s1, int l1, const char *s2, int l2)
     }
 }
 
+HIDDEN int bsearch_uncompare_mbox(const unsigned char *s1, size_t l1,
+                                  const unsigned char *s2, size_t l2)
+{
+    ssize_t min = l1 < l2 ? l1 : l2;
+    int cmp = 0;
+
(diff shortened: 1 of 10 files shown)


lib/bsearch.c
```diff
@@ -214,6 +214,40 @@ HIDDEN int bsearch_ncompare_mbox(const char *s1, int l1, const char *s2, int l2)
...
+    size_t min = l1 < l2 ? l1 : l2;
+    int cmp = 0;
+
+    while ((min-- >0) &&
+           (cmp = (TOCOMPARE(*s1) - TOCOMPARE(*s2))) == 0) {
+        s1++;
+        s2++;
+    }
+
+    return cmp;
```
elliefm: Does this suffer from the same bug as was fixed in cmpstringp_mbox by aff48467b2?

ajaysusarla: Good find! It probably is. Will fix that.

ajaysusarla:
Merged to master, with the following commit id's:
b3596cfe09a2f02383a411ea09b516c5167b58a9 , 6ab78038cf598605f23f6b5532254c9f5c39d22b and a18a1da6ad38321f3a073f0fade19e23ae66ce14

Issue #2525, opened 2018-09-16
Re-add annotatemore support for 3.0 and (tbd 3.2) protected by a config option


brong:
There's still client software in the wild which uses it, and we'd prefer to give those sites a chance to upgrade while still running modern Cyrus.  While it's not massively inconvenient to support the old annotatemore codepaths, let's add it back!

(besides I promised it on the mailing list)

elliefm:
Done ^^

PR #2540, opened 2018-09-27
bring back SETANNOTATION/GETANNOTATION, behind config

elliefm:
Brings back the ANNOTATEMORE capability and SETANNOTATION and GETANNOTATION commands, hidden behind a new config item `annotate_enable_legacy_commands` (default: off).  This is for #2525.

Requesting code review as Cassandane doesn't exercise these commands directly, so we have no real tests for it, and I dunno how useful it is writing a bunch of brand new tests for a feature we keep trying to get rid of...  

(It looks like maybe Cassandane used to exercise these commands implicitly through calls to cyradm, but cyradm has since been updated to use the METADATA extension.)


brong:
Looks reasonable as a readthrough.  I'd be happy to just apply this and fix things if people report them breaking.

elliefm:
Cool, merging :)

PR #2541, opened 2018-10-01
Replace simple return with cleanup flow

landgraf:
Make cleanup more consistence to prevent leaks of memory pointed by
filter/base/res


elliefm:
Thanks!

PR #2542, opened 2018-10-01
Close file descriptior in case of error

landgraf:
Make static code analizers happy.
If stat() failed for some reason it may lead backup fd unclosed.


elliefm:
Thanks!

Issue #2564, opened 2018-10-22
Making idle socket addresses should not be wrapped in assert()


zlynx:
In imap/idle.c the following two lines are dangerously wrong:

    122     assert(idle_make_client_address(&local));
    123     assert(idle_make_server_address(&idle_remote));

Asserts should never be used for necessary program functions.

Even if arguing that idle.c uses the lib/assert.h which is not affected by `#define NDEBUG` instead of the C Library assert.h is risky because the cyrus source code is not consistent on this point: some source files use `#include <assert.h>` and sometimes `#include "assert.h"`. If relying on particular behavior from a function named `assert()` it should use a different name than the standard library version of it. Maybe `xassert()`.

A quick check with grep seems to show that these two lines are the only assert calls with necessary side effects.

They should be changed to assign to a result variable, then assert that result variable is true.

elliefm:
Thanks for the PR, I've also cherry-picked this to the 3.0 branch :)

PR #2566, opened 2018-10-23
Take idle socket setup out of assert wrapper.

zlynx:
This fixes having side effects inside of assert calls. Issue #2564 

Nothing critical about this, just seeing this in the code bugged me.


elliefm:
Agreed, thanks!

PR #2575, opened 2018-10-31
Update documentation of cyrusdb_sync function

### LINKED ISSUE #2573: Review API documentation of cyrusdb_sync function
doc/internal/cyrusdb-api.html states:

    <h3>cyrusdb_sync(const char *backend)</h3>

    <p>Perform a checkpoint of the database environment.  Used by berkeley
    backend.  Is called by <tt>ctl_cyrusdb -c</tt> on a regular basis</p>

It is presumably not correct to say this is used by the berkeley backend as that has been removed. Is this function still used at all? If so, how should the documentation be updated?

rbricheno:
As per suggestion from elliefm. Should close #2573


elliefm:
Thanks! :)

Issue #2547, opened 2018-10-08
Labels via IMAP


brong:
It would be great to add a fetch item for labels via IMAP and maybe even allow setting it a-la-gmail, matching how mailboxIds works via JMAP.

brong:
There's two parts here: fetch (aka: follow GUID => Folder ID mappings) and set (AKA: move / copy / delete)

elliefm:
Note to self: for FETCH, `_email_mailboxes()` in jmap_mail.c contains logic for finding which mailboxes a given message is in (taking into account acls and flags)

brong:
This was done

PR #2581, opened 2018-11-13
Labels via IMAP

elliefm:
Implements the first stage (FETCH) of #2547 

```
    C: 5 fetch 1 (X-MAILBOXID)
    S: * 1 FETCH (X-MAILBOXID (INBOX))
    S: 5 OK Completed (0.000 sec)
    C: 6 create INBOX.target
    S: 6 OK [MAILBOXID (b265a82e-9a73-4aa5-873a-8e7a42dcea3b)] Completed
    C: 7 copy 1 INBOX.target
    S: 7 OK [COPYUID 1542073619 1 1] Completed
    C: 8 fetch 1 (X-MAILBOXID)
    S: * 1 FETCH (X-MAILBOXID (INBOX INBOX.target))
    S: 8 OK Completed (0.000 sec)
```

Cassandane tests here: https://github.com/cyrusimap/cassandane/compare/master...elliefm:v31/2547-imap-labels

I'm not wedded to the spelling `X-MAILBOXID` -- in fact, the more I see it, the more I dislike it.  I initially used the `X-` prefix because gmail's similar `X-GM-LABELS` does, but none of our other non-standard FETCH extensions use an `X-` prefix, so it's a bit out of place in a Cyrus context (the IMAP spec only specifies the `X-` prefix for commands, not other words).  Also, in hindsight I think it should be plural, and there's now a certain overlap with the nonstandard `FOLDER` fetch item.

So, I'm thinking to rename it to just `MAILBOXIDS` (which conveniently mirrors jmap lingo too).  Thoughts?

I'd appreciate a close review of the conversations db interactions. It's not clear to me when `conversations_state`s are suitably precached (e.g. in `mailbox->local_cstate`) or not, so I'm not certain whether just going ahead and opening a new state like I'm doing is safe or not.

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -3813,6 +3813,96 @@ static int index_fetchannotations(struct index_state *state,
     return r;
 }
 
+struct mailboxid_rock {
+    struct index_state *state;
+    int sep;
+    const struct fetchargs *fetchargs;
+};
+
+static int _mailboxid_cb(const conv_guidrec_t *rec, void *rock)
+{
+    static const int needrights = ACL_READ|ACL_LOOKUP;
+    struct mailboxid_rock *mbid_rock = (struct mailboxid_rock *) rock;
+    mbentry_t *mbentry = NULL;
+    int myrights = 0;
+    struct mailbox *mailbox = NULL;
+    msgrecord_t *msgrecord = NULL;
+    uint32_t system_flags, internal_flags;
+    char *extname = NULL;
+    int r = 0;
+
+    assert(mbid_rock->state != NULL);
+    assert(mbid_rock->fetchargs != NULL);
+
+    /* make sure we have appropriate rights */
+    r = mboxlist_lookup(rec->mboxname, &mbentry, NULL);
+    if (r) goto done;
+    myrights = cyrus_acl_myrights(mbid_rock->state->authstate, mbentry->acl);
+    if ((myrights & needrights) != needrights)
+        goto done;
+
+    /* open rec->mboxname */
+    r = mailbox_open_irl(rec->mboxname, &mailbox);
+    if (r) goto done;
+
+    /* find record for rec->uid */
+    r = msgrecord_find(mailbox, rec->uid, &msgrecord);
+    if (r) goto done;
+
+    /* skip if flag deleted or flag internal expunged */
+    r = msgrecord_get_systemflags(msgrecord, &system_flags);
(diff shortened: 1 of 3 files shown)


elliefm:
@rsto I'm using your new cached flags from convdb :)  FYI `_email_mailboxes()` in jmap_mail.c uses the same logic to do the same thing, so a) it could benefit from using that cache too, and b) maybe there's some scope to deduplicate these two implementations

elliefm:
`MAILBOXIDS` now returns the mailbox ids ;) and `MAILBOXES` returns mboxnames.  Cassandane tests updated to check both.

imap/index.c
```diff
@@ -3813,6 +3813,124 @@ static int index_fetchannotations(struct index_state *state,
...
+    assert(fmb_rock->fetchargs != NULL);
+
+    /* make sure we have appropriate rights */
+    r = mboxlist_lookup(rec->mboxname, &mbentry, NULL);
+    if (r) goto done;
+    myrights = cyrus_acl_myrights(fmb_rock->state->authstate, mbentry->acl);
+    if ((myrights & needrights) != needrights)
+        goto done;
+
+    if (rec->version >= 1) {
```
rsto: Nit: Checking version >= 1 and expunged/deleted flags at the start of this function would allow to skip a mboxlist and ACL lookup call for expunged messages. At least for FastMail or other recent installations that upgrade their conversations.db, the version will always be >= 1 soon.

imap/index.c
```diff
@@ -3813,6 +3813,124 @@ static int index_fetchannotations(struct index_state *state,
...
+        /* grab flags from convdb, if it has them */
+        system_flags = rec->system_flags;
+        internal_flags = rec->internal_flags;
+    }
+    else {
+        /* grab flags from message record */
```
rsto: Opening the mailbox in the GUID callback will be expensive if clients fetch a sequence set of non-trivial length. I don't think this will be an issue, given that opening the mailbox here is just a fallback for old conversation.dbs. But generally we suffered in the JMAP implementation at quite some places and turned to handling all messages in the same mailbox in bulk.

elliefm: I think I agree that this won't be an issue --  it will only affect people who are running the bleeding edge code (and get this feature) but haven't updated their convdb yet.

Once this becomes general release, presumably people will update their convdb as part of their upgrade process, and will just get the cheap version immediately.

Issue #2449, opened 2018-07-16
lmtpd segfault in make_sieve_db with null userid [Cyrus 3.0 head]


postilion:
Friends,
We're witnessing quite regular segfaults in lmtpd with some spam messages (we think it's just spam).  Here's the stack trace from gdb:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  strlcat (dst=dst@entry=0x55ab576d3520 <buf> ".", src=src@entry=0x0, len=4095, len@entry=4097) at lib/xstrlcat.c:55
55    lib/xstrlcat.c: No such file or directory.
(gdb) bt
#0  strlcat (dst=dst@entry=0x55ab576d3520 <buf> ".", src=src@entry=0x0, len=4095, len@entry=4097) at lib/xstrlcat.c:55
#1  0x000055ab574caab6 in make_sieve_db (user=user@entry=0x0) at imap/lmtp_sieve.c:101
#2  0x000055ab574cb79b in send_rejection (origid=0x55ab58a90360 "<20180713213031.CB91F63271@smtp.onlight.com>", rejto=0x55ab58a6a980 "myjmwcjs@dygu.com",
    origreceip=origreceip@entry=0x0, mailreceip=0x0, reason=reason@entry=0x7f36763d311c <error: Cannot access memory at address 0x7f36763d311c>, file=file@entry=0x55ab58a903d0)
    at imap/lmtp_sieve.c:282
#3  0x000055ab574cbecc in sieve_reject (ac=<optimized out>, ic=<optimized out>, sc=0x7ffeb99ec3c0, mc=<optimized out>, errmsg=0x7ffeb99eb318) at imap/lmtp_sieve.c:498
#4  0x00007f3675fa2de8 in do_action_list (errmsg=<optimized out>, actions_string=0x7ffeb99eb350 "Action(s) taken:\n", notify_list=0x55ab58a92cd0, actions=0x55ab58a92fb0,
    imapflags=0x7ffeb99eb320, message_context=0x7ffeb99ed510, script_context=0x7ffeb99ec3c0, interp=0x55ab58a91af0) at sieve/script.c:774
#5  sieve_execute_bytecode (exe=<optimized out>, interp=interp@entry=0x55ab58a91af0, script_context=script_context@entry=0x7ffeb99ec3c0,
    message_context=message_context@entry=0x7ffeb99ed510) at sieve/script.c:964
[...]

PR #2597, opened 2018-12-03
LMTP segmentation fault on mailshare with sieve redirections

toony:
Create a mailshare with _email mailshare@domain.tld_ and set sieve script with redirection like:
`if allof ( address :contains ["to", "cc"] "sender") {
        redirect :copy "target@domain.tld";
        stop;
}`

Got a LMTP segmentation fault when receiving mail matching sieve rule:
`Nov 28 12:53:46 cyrus-vm postfix/qmgr[10701]: CDDAD800E1: from=<sender@domain.tld>, size=74989, nrcpt=1 (queue active)
Nov 28 12:53:46 cyrus-vm cyrus/master[11265]: process type:SERVICE name:lmtp path:/usr/lib/cyrus/lmtpd age:9.817s pid:11291 signaled to death by signal 11 (Segmentation fault)
Nov 28 12:53:46 cyrus-vm postfix/lmtp[11344]: CDDAD800E1: to=<+mailshare@domain.tld>, relay=172.16.20.51[172.16.20.51]:24, delay=165587, delays=165587/0.02/0/0.06, dsn=4.4.2, status=deferred (lost connection with 172.16.20.51[172.16.20.51] while sending end of data -- message may be sent more than once)`

### DIFF
--- a/imap/lmtp_sieve.c
+++ b/imap/lmtp_sieve.c
@@ -414,7 +414,7 @@ static int sieve_redirect(void *ac,
     /* if we have a msgid, we can track our redirects */
     if (m->id) {
         snprintf(buf, sizeof(buf), "%s-%s", m->id, rc->addr);
-        sievedb = make_sieve_db(mbname_userid(sd->mbname));
+        sievedb = make_sieve_db(mbname_recipient(sd->mbname, ((deliver_data_t *) mc)->ns));
 
         dkey.id = buf;
         dkey.to = sievedb;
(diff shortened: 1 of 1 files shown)


elliefm:
This sounds similar to the crash in #2449.  @postilion, does this patch resolve the crash for you?

elliefm:
On further examination, I don't think this particular patch solves #2449, but I think the same approach in the affected area would

elliefm:
Thanks very much for this, I've also cherry-picked it to the master branch :)

toony:
You're welcome. Pleased to help!

PR #2605, opened 2018-12-05
Fix: sieve runtime error on fileinto for a mailshare

toony:
Create shared folder: _mailshare@domain.tld_ with subfolder _Sent_.
Put a sieve script on _mailshare@domain.tld_  shared folder with a _fileinto_ instruction like:
`
if allof ( header :is "Subject" "fileinto") {
        fileinto "Shared Folders/mailshare/Sent";
        stop;
}
`

Sending email with subject _fileinto_ to this shared folder using LMTP and get:
`
Dec  5 11:55:18 bluemind-40 cyrus/lmtp[14072]: sieve runtime error for (null) id <e6a78d5b724c21fd4fc8071c1803fa1b@blue-mind.net>: Fileinto (Shared Folders/mailshare/Sent): Mailbox does not exist
`
Mail is delivered into _mailshare@domain.tld_ shared folder instead of its _Sent_ subfolder.

### DIFF
--- a/imap/lmtp_sieve.c
+++ b/imap/lmtp_sieve.c
@@ -529,7 +529,10 @@ static int sieve_fileinto(void *ac,
     char namebuf[MAX_MAILBOX_BUFFER];
     int ret;
 
-    char *intname = mboxname_from_external(fc->mailbox, sd->ns, mbname_userid(sd->mbname));
+    char *intname = mboxname_from_external(fc->mailbox, sd->ns,
+                                           mbname_recipient(sd->mbname,
+                                                            sd->ns));
+
     strncpy(namebuf, intname, sizeof(namebuf));
     free(intname);
 


elliefm:
We need to do something here, but I don't think this patch is correct.  Note that the third argument to mboxname_from_external is explicitly a "userid", not a recipient like it has been in the other cases:

```
EXPORTED char *mboxname_from_external(const char *extname, const struct namespace *ns, const char *userid)
```

I think it is intended that this argument be passed a NULL pointer in the case of shared mailboxes that do not belong to a user -- if so, then passing it the result of `mbname_userid(sd->mbname)` here is correct.  But the internals might be buggy!

`mboxname_from_external` is a small wrapper around `mbname_from_extname`, so if `mbname_from_extname` doesn't handle this case correctly that could be the source of the problem here?

Our cunit/mboxname.testc unit tests don't really exercise mbname_from_extname unfortunately :/

elliefm:
Actually, I think I'm wrong about the intent of the `userid`, but I'm still not sure it should be a recipient in this case?  This is complicated...

elliefm:
So, I think the `userid` is the user in whose context the `extname` is to be understood. -- usually this would be the logged-in user.  But we're not in an imapd here, we're processing a sieve script, and it's a shared-folder sieve script, not a user sieve script, so I'm not sure what the `mbname_from_extname` API needs us to provide in this case.

It looks like there might be differences in this code path from 3.0 to master already, maybe this case has already been fixed there?  Are you able to test that, by any chance?

rsto:
I can not replicate this issue. Could you please validate if this still is an issue on the current main branch? Please document the exact commands that you took that lead to this issue, preferably by adding a Cassandane test to your PR.

The tests https://github.com/cyrusimap/cyrus-imapd/blob/master/cassandane/tiny-tests/Sieve/deliver and https://github.com/cyrusimap/cyrus-imapd/blob/e269746ab0b626990f4d7757c6204e7dc0ac1eff/cassandane/Cassandane/Cyrus/Delivery.pm#L350-L379 should be a good starting point.

rsto:
Please reopen this PR or an issue if this still is relevant.

PR #2609, opened 2018-12-10
remove chdir for do_examine and do_quota

### LINKED ISSUE #2606: mbexamine can breake if archive- and meta-partition are used
I the following conditions are all true:
- archive- and meta-partition are enabled, 
- all cyrus meta files reside on the meta-partition
- a folder has no subfolders 
- all mails in the folder are archived.

the "base" directory of the folder on the partition can become empty. 
In some case, depending on the code path the last file from this directory 
was removed, cyrus will remove this directory from the filesystem. 

If this directory gets removed mbexamine will fail to chdir to this directory 

The following codeblock is included in the function ``o_examine``, ``do_quota`` and ``do_compare`` which relate to the 3 operation modi of mbexamine: 

```
   if (chdir(mailbox_datapath(mailbox, 0)) == -1) {
        r = IMAP_IOERROR;
        goto done;
    }
```
as far as i can see, only ``do_compare`` uses the fact the directory was changed by 
by calling ``opendir(".")`` and scanning the filesystem 

This was tested with cyrus 3.0.8

MichaelMenge:
Partial Fixes #2606 by removing the chdir from do_examine and do_quota.
The the function do_compare still needs a fix.

### DIFF
--- a/imap/mbexamine.c
+++ b/imap/mbexamine.c
@@ -216,11 +216,6 @@ static int do_examine(struct findall_data *data, void *rock __attribute__((unuse
     r = mailbox_open_irl(name, &mailbox);
     if (r) return r;
 
-    if (chdir(mailbox_datapath(mailbox, 0)) == -1) {
-        r = IMAP_IOERROR;
-        goto done;
-    }
-
     printf(" Mailbox Header Info:\n");
     printf("  Path to mailbox: %s\n", mailbox_datapath(mailbox, 0));
     printf("  Mailbox ACL: %s\n", mailbox->acl); /* xxx parse */
@@ -352,7 +347,6 @@ static int do_examine(struct findall_data *data, void *rock __attribute__((unuse
         printf("Desired message not found\n");
     }
 
- done:
     mailbox_close(&mailbox);
 
     return r;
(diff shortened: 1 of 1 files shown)


elliefm:
I wonder if the chdir check is useful in some cases?  In the case where the base directory being missing is not a problem, the chdir check could be skipped.  But if we expect there to be subfolders, mail, or metadata in there, then a missing basedir is a real problem that mbexamine should report?  I'm not sure how plausible it is to distinguish these cases though, any thoughts?

MichaelMenge:
> I wonder if the chdir check is useful in some cases?
@elliefm in general or only for the do_compare function?

In do_examine and do_quota the files are accessed by mailbox_open_advanced and mailbox_record_fname, which will use the correct path for the files and will handle archive-,
meta-partitions and object storage. As far as i can see these functions also should report 
any errors if they are unable to access the files they need to access. 

I don't know if do_compare mode is needed any more, as recontruct will inspect the correct paths in the filesystem. But I didn't want to remove a function/option in a patch for a minor update.

elliefm:
also cherry-picked this to master :)

PR #2613, opened 2018-12-13
support clamav-0.101.0 and apply CLAMAV_CFLAGS

adamgolebiowski:
minor change is required to imap/cyr_virusscan.c to support clamav-0.101.0.

Also pass CLAMAV_CFLAGS when building imap/cyr_viursscan.c


elliefm:
Thanks! :)

PR #2637, opened 2019-01-31
Tiny improvements to docs

FSchumacher:
Corrected a typo in the ldap attribute example and used a bit of markup to
highlight the code examples in text a bit more. Finally use mdash instead of
ndash and simple dashes, when they are used to split subsentences.


elliefm:
Thanks!

PR #2640, opened 2019-02-04
Calalarm additional properties

ajaysusarla:
I couldn't get any clear documentation about this. But the assumption is that the property `ICAL_CALID_PROPERTY` gives the if of the owner of the calendar, based on the tests that I did. I might be way off here.


elliefm:
>the assumption is that the property `ICAL_CALID_PROPERTY` gives the id of the owner of the calendar

@ksmurchison is this correct?

The pr looks fine syntactically, but I don't know enough to know if it's semantically correct!

ksmurchison:
Actually this property is defined by the defunct Calendar Access Protocol (RFC 4324).  Its the ID of the calendar, but we would use the UID property for that.

What problem are we trying to solve?

ajaysusarla:
@ksmurchison @elliefm , Thanks for looking into this. I want to be able to send the id of the calendar owner. I need that information to be able to make a JMAP `CalendarEvent/get` on an event id from the even listener end(a.k.a pusher), for which i need the `accountId`.
 Based on what @ksmurchison has said, I'm guessing this is the wrong field. 
The alternative is to probably use `mboxname_to_userid` or `mboxname_to_external` functions and pass that along.

ksmurchison:
If you need this info in the actual iCalendar data, you probably can use an X- property.  Something like X-JMAP-ACCOUNTID

ajaysusarla:
How would I set that?

brong:
@ajaysusarla I'm digging through really old PRs that have a review request for me.  Is this still something we need?

rsto:
This is already handled now in mboxevent.c.

PR #2643, opened 2019-02-06
Return correct group names when groups are resolved by filter

FSchumacher:
This fixes an regression, that was introduced with 61f5296c0d727faee4726525a6812b200d946d83.
Back then the logic was changed from `ldap_count_values(vals) != 1` to two `if` clauses
`ldap_count_values(vals) < 1` and `ldap_count_values(vals) > 1`
which logged errors and an else clause (which matched the correct number
of `ldap_count_values(val) == 1`) that got the old error handling code.


elliefm:
Nice catch, really appreciate all these ldap patches :)

elliefm:
Accidentally reverted this while intending to revert something else (doh!), so I've now reverted the bad revert, and therefore this patch is back in place.

PR #2669, opened 2019-03-13
Set a Mime-Version header on created emails

### LINKED ISSUE #2668: JMAP email creation doesn’t set MIME-Version header
[SpamAssassin’s MIME_HEADER_CTYPE_ONLY rule](https://wiki.apache.org/spamassassin/Rules/MIME_HEADER_CTYPE_ONLY) slaps a fairly large penalty on messages missing `MIME-Version: 1.0`. The JMAP email crafter should add this header.

chris-morgan:
Fixes #2668.

Tests doubtless need to be updated for this, but I’m not sure where they are. The update should be straightforward.


rsto:
Thanks. I'll add a regression test to Cassandane.

PR #2672, opened 2019-03-15
rely on CL_SCAN_STDOPT instead of LIBCLAMAV_MAJORVER

philfry:
This PR uses a different approach for 09b5bf11e12185727d53a6ae06d71a767952a537.

As `LIBCLAMAV_MAJORVER` is only set in `clamav-config.h` which is not being installed by `make install` the devel-packages of clamav don't contain that macro. So the majorver switch point doesn't work – at least with RHEL/CentOS/SL using EPEL's clamav or Fedora. Not sure regarding other distributions, though.
IMHO it is quite easier checking for the existence of `CL_SCAN_STDOPT`.

### DIFF
--- a/imap/cyr_virusscan.c
+++ b/imap/cyr_virusscan.c
@@ -192,7 +192,7 @@ int clamav_scanfile(void *state, const char *fname,
     int r;
 
     /* scan file */
-#if LIBCLAMAV_MAJORVER < 9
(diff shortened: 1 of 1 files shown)


elliefm:
Also cherry-picked to the 3.0 branch.

Thanks! :)

PR #2678, opened 2019-03-21
imap/conversations.h: C++ interface

dilyanpalauzov:
Permit `#include <imap/conversations.h>` in C++ programs.


elliefm:
Thanks! Also cherry-picked to master as f850b316e

PR #2684, opened 2019-03-21
notify/nodifyd.c: update synopsis

dilyanpalauzov:


dilyanpalauzov:
Can this make progress, please?

rsto:
@dilyanpalauzov I do not understand the suggested change. The updated usage synopsis documents argument options that are not handled by notifyd's getopt argument parsing. Am I missing something here?

dilyanpalauzov:
The parameters are documented at https://www.cyrusimap.org/imap/reference/manpages/systemcommands/notifyd.html . In addition `getopt()` handles the `m:`ethod parameter.

rsto:
Right, my bad. These arguments are processed by `master` before it calls `service_init`. Would you please change `EC_USAGE` to `EX_USAGE` as return status?

rsto (approved):

rsto:
Thanks, sorry that this got not handled earlier.

PR #2685, opened 2019-03-21
imap/annotate.c:annotate_state_writemask: put shared attributes on shared mailboxes

dilyanpalauzov:
The mailbox #addressbook/shared does not belong to anybody, but a user with write-privileges to it shall be able to set global properties over WebDAV like CARD:addressbook-description or DAV:displayname.

### DIFF
--- a/imap/annotate.c
+++ b/imap/annotate.c
@@ -2643,7 +2643,7 @@ EXPORTED int annotate_state_writemask(annotate_state_t *state,
                                       const struct buf *value)
 {
     /* if the user is the owner, then write to the shared namespace */
-    if (mboxname_userownsmailbox(userid, state->mailbox->name))
+    if (mboxname_userownsmailbox(userid, state->mailbox->name) || !mboxname_isusermailbox(state->mailbox->name, 1))
         return annotate_state_write(state, entry, "", value);
     else
         return annotate_state_write(state, entry, userid, value);


elliefm:
@ksmurchison This seems sane to me, but I don't know DAV well, what do you think? This is for 3.0... will we need this on master too, or has the world changed?

ksmurchison:
My opinion is that DAV properties such as displayname, description, color, etc should be per-user and therefore stored in the private namespace.  That said, I will listen to arguments to the contrary.

dilyanpalauzov:
Shared collections share the content between the users, I do not see why shall the properties not be shared.  The idea is to grant some users the right to update the properties (description) and these changes are propagated to all users.  The users which change the properties do not have to be administrators, this is the whole point.

ksmurchison:
You may want to name a shared calendar 'foo' and I may want to name it 'bar'.  You may want it to appear blue and I may want red.  I shouldn't be forced to use your preferred name, nor should I be able to overwrite it.  I could see perhaps allowing the creator of the calendar to write a shared annotation which becomes the default for others until they write their own private one.

dilyanpalauzov:
As far as I remember without this change only the administrator could set the properties.  You cannot have it red, because you lack the rights to set properties.  It is primary about setting a long description, where users will be able to read what is that particular calendar about.

dilyanpalauzov:
Whatever.  My use case would be to allow very few users to change the properties and it shall not be the administrator doing the changes.  But if the workflow is that the administrator has to change the initial value and the others can customize it, then this is also acceptable.

dilyanpalauzov:
I agree with Ken and withdraw my proposal.

PR #2686, opened 2019-03-21
imap/http_caldav_sched.c:imip_send_sendmail: print less redundant information in the text/plain part about attendees

dilyanpalauzov:
And in general, when the CUA uploads ORGINIZER/ATTENDEE with CN that is the same as its email address, do not render the CN.


dilyanpalauzov:
Duplicate of https://github.com/cyrusimap/cyrus-imapd/pull/3721.

PR #2687, opened 2019-03-21
configure.ac: add AM_SILENT_RULES([yes])

dilyanpalauzov:


elliefm:
Rather not change this default on stable branch.  You can already get this effect by passing --enable-silent-rules to configure

PR #2690, opened 2019-03-21
GH#2394: Send Content-Disposition: "inline;" and not "attachment;"

dilyanpalauzov:


dilyanpalauzov:
This was addressed in cfbc403645e11.

PR #2693, opened 2019-03-21
imap/http_caldav_sched.c:imip_send_sendmail() quote the display-addr part of From:/To: addresses, when necessary

dilyanpalauzov:


dilyanpalauzov:
* change imap/http_caldav_sched.c:imip_send_sendmail not to include display-part in the To:/From: addresses, unless the text is different from the email address
* introduce lib/charset.c:mime_quote_string() which returns a qouted string, if the input must be quoted for inclusion in headers, or NULL otherwise
* If the display part of From:/To: addresses in imap/http_caldav_sched.c:imip_send_sendmail() must be quoted, quote it.

dilyanpalauzov:
My understanding is that the underlaying problem, solved here, was reported in https://github.com/cyrusimap/cyrus-imapd/issues/4955 and corrected in https://github.com/cyrusimap/cyrus-imapd/pull/4956, five years after I provided a fix for the same problem.

It would have been better, if any concerns on the current change were raised here earlier, in case anything was unclear.

PR #2129, opened 2017-08-26
timsieved, user: fix script path for users with dots in their names

maciejsszmigiero:
For an user "foo.bar" with a sieve directory of /var/spool/sieve lmtp
searches for his scripts in a /var/spool/sieve/f/foo^bar/ directory but
timsieved creates and manages them in a /var/spool/sieve/f/foo.bar/
directory.

This is caused by an inconsistency in an user name -> path translation
between sieve_find_script() in imap/lmtp_sieve.c and actions_setuser() in
timsieved/actions.c.

Fix this by adding dot-translating code from imap/lmtp_sieve.c also to
timsieved/actions.c.

This change will need to be applied to 3.0 branch, too.

PR #2694, opened 2019-03-21
Fix #2129 timsieved sieve_path

dilyanpalauzov:
-- imap/autocreate.c: IMAPOPT_SIEVEDIR has default value, so that it cannot be undefined.

Use a single implementation of sieve_find_script() instead of five.

### DIFF
--- a/imap/autocreate.c
+++ b/imap/autocreate.c
@@ -104,7 +104,6 @@ static int autocreate_sieve(const char *userid, const char *source_script)
     int do_compile = 0;
     const char *compiled_source_script = NULL;
     const char *sievename = get_script_name(source_script);
-    const char *sieve_script_dir = NULL;
     char sieve_script_name[MAX_FILENAME];
     char sieve_bcscript_name[MAX_FILENAME];
     char sieve_default[MAX_FILENAME];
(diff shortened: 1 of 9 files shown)


imap/autocreate.c
```diff
@@ -121,49 +120,51 @@ static int autocreate_sieve(const char *userid, const char *source_script)
...
-        return 1;
-    }
-
     /* Check if autocreate_sieve_compiledscript is defined in imapd.conf */
     if(!(compiled_source_script = config_getstring(IMAPOPT_AUTOCREATE_SIEVE_SCRIPT_COMPILED))) {
         syslog(LOG_WARNING, "autocreate_sieve: autocreate_sieve_compiledscript option is not defined. Compiling it");
         do_compile = 1;
     }
 
+    const char *sieve_script_dir;
```
elliefm: I still need to look at the rest of this properly, but in the meantime the Travis build is failing because this needs to not be `const`.

I guess it was never a problem before because we didn't bother freeing it?

ksmurchison:
This appears to be outdated.  Is this still an issue?

dilyanpalauzov:
At the time I upload this, it was a very good idea!  Now it is outdated.

Issue #2124, opened 2017-08-25
link libcyrus with -lm


dilyanpalauzov:
because of ceil() and log() in lib/bloom.c

```diff
--- a/Makefile.am
+++ b/Makefile.am
@@ -1405,7 +1405,7 @@ lib_libcyrus_la_SOURCES += lib/nonblock_fcntl.c
 else
 lib_libcyrus_la_SOURCES += lib/nonblock_ioctl.c
 endif
-lib_libcyrus_la_LIBADD = ${LIB_SASL} $(SSL_LIBS) $(GCOV_LIBS)
+lib_libcyrus_la_LIBADD = ${LIB_SASL} $(SSL_LIBS) $(GCOV_LIBS) -lm
 lib_libcyrus_la_CFLAGS = $(AM_CFLAGS) $(CFLAG_VISIBILITY)
 
 nodist_lib_libcyrus_min_la_SOURCES = \

```

dilyanpalauzov:
Doing `./configure` prints
```
Cyrus Server configured components

   gssapi:             yes
   autocreate:         no
   idled:              no
   httpd:              no
   kerberos V4:        no
   murder:             no
   nntpd:              no
   replication:        no
   sieve:              yes
   srs:                no
   calalarmd:          no
   jmap:               no
   objectstore:        no
   backup:             no

External dependencies:
   ldap:               no
   openssl:            yes
   zlib:               yes
   pcre:               yes
   clamav:             no
   snmp:               no
   -----------------------
   caringo:            no
   openio:             no
   -----------------------
   nghttp2:            no
   brotli:             no
   xml2:               no
   ical:               no
   icu4c:              no
   shapelib:           no

Database support:
   mysql:              no
   postgresql:         no
   sqlite:             no

Search engine:
   squat:              yes
   xapian:             no
   xapian_flavor:      none

Hardware support:
   SSE4.2:             yes

Installation directories:
   prefix:             /usr/local
   sysconfdir:         /etc

Build info:
   shared:             yes
   static:             no
   cflags:             -Wall -Wextra -pipe -O3 -fno-fat-lto-objects -flto
   cxxflags:           -pipe -Wall -Wextra -O3 -fno-fat-lto-objects -flto
   libs:                -lgssapi_krb5 -lkrb5 -lk5crypto -lcom_err -lkrb5support    -lpcre -lpcreposix -lz
   ldflags:            -Wl,-O1,-s -fuse-ld=gold -flto=4
```

[...]

PR #2695, opened 2019-03-21
Don't link everything with libbrotlienc

dilyanpalauzov:


dilyanpalauzov:
`-lm` is included in `imap_httpd_LDADD`.

configure.ac
```diff
@@ -1575,21 +1573,14 @@ dnl                AC_MSG_WARN([Your version of OpenDKIM can not support iSchedu
         dnl httpd needs libmath in a few places
```
elliefm: can you also move this block of comments to the new `-lm` location please. i don't wanna have to dig through history later to figure out wtf they mean, nor do i want their purpose to be lost when reading code at the new location. thanks

dilyanpalauzov: I do not understand what 
```
dnl XXX really should check for this properly but AC_SEARCH_LIBS/AC_CHECK_LIB
dnl XXX break under -Werror(!)
```
is supposed to mean.  Copying these lines to Makefile.am will make the interpretetation harder.

elliefm: It means we should not be hardcoding `-lm` at all: we should be letting autoconf determine whether it's needed, and if so what it's called.  But we couldn't just do that at the time the comment was written, because the relevant autoconf macros were broken.  I dunno if they're fixed now -- if they are, we can fix `-lm` properly, but that's a separate job.  In the meantime, keep the comment with the `-lm` so the context is preserved.

dilyanpalauzov: Where does the requirement for libm come from?  libbrotlidec.so has as undefiend symbol log2.  Both versions 1.0.3 and 1.0.7 have libm.so as DT_NEEDED.  sqlite3.pc has in its .pc file `Libs.private: -lz -lm -ldl -lpthread`.

dilyanpalauzov: lib/bloom.c calls log() and ceil(), so it is a matter passing -lm when linking libcyrus, see https://github.com/cyrusimap/cyrus-imapd/issues/2124.

ksmurchison:
Do we still want to do this?  It needs updating.

elliefm:
I don't think we do.  If the general shape of the idea is good, we should consider doing it on master (and perhaps backporting it).  But this PR is for 3.0.  There's no sense landing something on 3.0 that doesn't land anywhere else, it'll just be confusing later.

dilyanpalauzov:
When ./configure prints 
```
checking for libbrotlienc... yes
   brotli:             yes
…
$ make V=1 master/master
/bin/sh ./libtool  --tag=CC   --mode=link gcc  -Wwrite-strings -Wbad-function-cast -Wdiscarded-qualifiers -O3  -Wdeclaration-after-statement -DPCRE2POSIX_SHARED  -licui18n -licuuc -licudata -lxml2  -L/usr/local/lib64 -lical -licalss -licalvcal  -ljansson  -L/usr/local/lib -lwslay -lbrotlienc   -lm -L/usr/local/lib -lchardet -ljansson  -flto=4 -o master/master master/master.o master/masterconf.o lib/libcyrus_min.la -lgssapi_krb5 -lkrb5 -lk5crypto -lcom_err -lkrb5support -lresolv -lresolv  -lresolv   -lpcre2-posix -lpcre2-8 -lz -lxml2 -L/usr/local/lib64 -lical -licalss -licalvcal -licalvcard  -lsqlite3  -lm -lgssapi_krb5 -lkrb5 -lk5crypto -lcom_err -lkrb5support -lresolv -lresolv  -lresolv   -lpcre2-posix -lpcre2-8 -lz -lxml2 -L/usr/local/lib64 -lical -licalss -licalvcal -licalvcard  -lsqlite3
libtool: link: gcc -Wwrite-strings -Wbad-function-cast -Wdiscarded-qualifiers -O3 -Wdeclaration-after-statement -DPCRE2POSIX_SHARED -flto=4 -o master/.libs/master master/master.o master/masterconf.o  -L/usr/local/lib64 -L/usr/local/lib lib/.libs/libcyrus_min.so -licui18n -licuuc -licudata /usr/local/lib/libwslay.so -lbrotlienc /usr/local/lib/libchardet.so -ljansson -luuid -lm -lgssapi_krb5 -lkrb5 -lk5crypto -lcom_err -lkrb5support -lresolv -lpcre2-posix -lpcre2-8 -lz -lxml2 -lical -licalss -licalvcal -licalvcard -lsqlite3

$ readelf -d master/.libs/master

 0x0000000000000001 (NEEDED)             Shared library: [libbrotlienc.so.1]
```
 so the master process is linked with libbrotlienc, which is not necessary.
 
Irrespecitve of whether this change targeted the development or stable branch, at the time it was submitted, five years later, when it came in the queue, the change would not apply straight-forward any more. Neither do I remember the details.

PR #2697, opened 2019-03-21
configure.ac: search for GitPython in the current python

dilyanpalauzov:


elliefm:
I agree with this PR in principle, but I _think_ the explicit `python2` might be required by the build script on the server that cyrusimap.org is currently hosted on.

I'll leave this PR open for now, and merge it once we finish the migration to github.io

J5lx:
Any news on this?

elliefm:
Not yet, we're still hosted on the same server.  I see elsewhere that that conversation has been bumped again (maybe in response to your comment), so we'll see if anything comes of it this time

J5lx:
Sorry for bugging you about this again, but as far as I can tell the site is now being served from github.io, so are there any news about this? I’d kinda like to get rid of that downstream patch for 3.2, even though it’s only small ^^

On another note, the DNS A entry for cyrusimap.org (without www) points to 127.0.0.17, which… doesn’t seem quite right :)

elliefm:
Thanks for bumping it up, that's a good thought.

Yeah, we are no longer building the website using the old server, so in theory we no longer have the hard dependency on the older version of sphinx-build.

I'm not really sure where our generate-the-website-script is currently running from. When the old server dropped off the planet, I was expecting to have to run it by hand from my laptop, but I checked and it was still updating every hour... so, somewhere in FM?  But I'm not sure if it's on a virtual machine or somebody's desktop or what.  With the current situation, it'll probably be low priority to properly sort it out for a while.  But in the meantime I'm the only person touching docs, and I'm checking they rebuild properly each time, and if they don't I can do it manually from my laptop, so it's at least as under control as anything is at the moment... 🤷‍♀️

But I'd prefer not to fiddle with it too much until it's definitely running from a new long-term home, cause I don't want to destabilise things any further.

> On another note, the DNS A entry for cyrusimap.org (without www) points to 127.0.0.17, which… doesn’t seem quite right :)

Yep, that's 100% wrong, is not something we asked them to do, and is it's own whole drama.  Hopefully it can be fixed soon. /sigh

dilyanpalauzov:
This is addressed in 3de55bcdf2c.

PR #2703, opened 2019-03-21
Remove double included entries in TOC: GH1996

dilyanpalauzov:


ksmurchison:
This appears to be out of date.  Do we want to do anything with it?

elliefm:
These two files are used to generate:
* http://www.cyrusimap.org/imap/developer.html
* https://www.cyrusimap.org/imap/download/release-notes/index.html

Looking at them on the web, I don't see anything that particularly needs changing in the content.

I think the main issue is that in the left-side sitewide navigation, the "compiling" page is linked from both Developers and Setup.  I think that's fine, actually.

It might be nice to have separate compiling instructions for "developers setting up a development environment" vs "admins wanting to compile their own package", but that would require someone to write (and maintain) the two pages.  For now we only have one set of compiling instructions, and since two kinds of people might need to read it, it makes sense to me for it to be linked from two places in the TOC.

dilyanpalauzov:
> I think the main issue is that in the left-side sitewide navigation, the "compiling" page is linked from both Developers and Setup.

I think the same and newer Sphinx complain about this.  When do you plan to update Sphinx?

elliefm:
Ahh interesting.

We're currently using Sphinx 3 to build the website, on Debian Bullseye I think.  On my laptop I have Sphinx 5 from Debian Bookworm, which generates very different html from what Sphinx 3 generated, and a lot of CSS rules in our stylesheet no longer match up with the html.  One annoying consequence of this is that I can no longer run ad-hoc website updates from my laptop, because it will make a mess of it.  To start using Sphinx 5 or newer, we'll need to rewrite the stylesheet at the same time -- and when we do, Sphinx 3 will no longer work correctly, so we need to be careful about when and how we do this.

Our current stylesheet looks like it was just copied from an old default, and lightly modified.  I hope that when the time comes, we can just copy a newer default version and lightly modify that, but I haven't looked into it any deeper.  If we have to actually rewrite the stylesheet ourselves, then it'll probably never happen, because I'm not a web developer and nobody else cares.

Looks like Sphinx 8 is currently in development, so even Sphinx 5 is pretty old at this point. :/

dilyanpalauzov:
Why is Sphinx not upgraded to version 8 with some standard stylesheet? I mean, first upgrade to Sphinx 8 and then if somebody wants, the stylesheet can be upgraded.

PR #2706, opened 2019-03-21
Good bye cyrusdb_sync

dilyanpalauzov:
It was used only by BDB.


ksmurchison:
@brong Is this something we are interested in doing?  I don't see any harm in leaving the stub functions in place in case we ever support a new backend that needs a sync

ksmurchison (approved):

PR #2708, opened 2019-03-21
Compiler warnings and other fixups

dilyanpalauzov:


elliefm:
@ksmurchison: the rest of this looks fine, but I'm not sure of the ramifications (if any) of the change to the success return in `report_card_query()` (from `0` to `HTTP_MULTI_STATUS`)?

ksmurchison:
The change to report_card_query() is wrong, as it will result in a second 207 response.
The other changes appear to be just cosmetic.

dilyanpalauzov:
I created this changeset back in 2017, when it contained real compiler warning. These were fixed upstream, so that the changeset got smaller.  I do not remember why I have modified the return code.   In any case I have removed the snippet now.

PR #2709, opened 2019-03-21
timsieved/actions.c:scriptname_valid: ensure script name is short enough

dilyanpalauzov:


elliefm:
What is that number 1015 based on?

dilyanpalauzov:
timsieve/actions.c:putscript() contains 
```c
char path[1024];
result = scriptname_valid(name);
snprintf(path, 1023, "%s.script.NEW", name->s);
```
So name->s can only be 1012 bytes long.

elliefm:
I just had a look through `timsieved/actions.c` and the whole damn thing is full of magic numbers (many wrong), so in the meantime one more magic number can't hurt.

PR #2711, opened 2019-03-21
Makefile.am: play with EXTRA_DIST

dilyanpalauzov:
Instead of keeping source files in EXTRA_DIST, that are build using nodist_ prefix, mention the files in (dist_).._SOURCES.

imap/lmtpstats.c, imap/pushstatc.s and imap/rfc822_header.c were already in imap_(imapd,lmtpd,libcyrus_imap)_SOURCES, so there is no point to repeat them in EXTRA_DIST.

Include imap/imap_err.[ch] in the distribution, as it is done for (mupdate,tz,nttp)_err.[ch] .


elliefm:
This is an interesting one...

Any generated source files that are included in the distribution end up, in practice, being built by me, in my environment, when I do a release.  In general, I prefer generated files to _not_ be included in the distribution, unless their build dependencies are a nuisance (like `xxd`).

In the case of the foo_err.[ch] files, I agree that either:

- they should ALL be distributed prebuilt (in which case com_err is no longer a build dependency), OR
- NONE should be distributed prebuilt (in which case com_err is a build dependency, but we already bundle it!)

I'm leaning towards NONE (i.e. like imap_err, not like mupdate_err/etc)

In the case of the other generated files, I don't have an opinion right now, I would need to look at their build dependencies cause I can't remember them offhand.

Generally, I would take the `nodist_` prefixed stuff as canonical/deliberate, and if these things are also EXTRA_DIST'd then that should be removed (the bad EXTRA_DIST's are mostly bits I failed to clean up, I think).

Makefile.am
```diff
@@ -68,6 +68,7 @@ AM_LDFLAGS = $(COV_LDFLAGS) $(ICU_LIBS)
 
 BUILT_SOURCES = \
 	imap/imap_err.c \
+	imap/jmap_err.c \
```
elliefm: I assume this is meant to be `.h`? :)

Makefile.am
```diff
@@ -115,15 +116,12 @@ SUBDIRS = .
 DIST_SUBDIRS = . perl/annotator perl/imap perl/sieve/managesieve
 dist_sysconf_DATA =
 lib_LTLIBRARIES = lib/libcyrus_min.la lib/libcyrus.la
-EXTRA_PROGRAMS =
```
elliefm: Generally I'm keeping all these initialisations here at the top (even if they're blank) so that conditional paths later can always-append without needing to chase the original declaration around.

Makefile.am
```diff
@@ -115,15 +116,12 @@ SUBDIRS = .
...
 check_PROGRAMS =
 libexec_PROGRAMS =
 sbin_PROGRAMS =
 noinst_HEADERS =
 noinst_LTLIBRARIES =
 noinst_PROGRAMS =
-EXTRA_DIST = \
```
elliefm: Same comment as EXTRA_PROGRAMS

Makefile.am
```diff
@@ -323,9 +297,10 @@ endif # SERVER
 
 endif # SIEVE
 
-EXTRA_DIST += \
+EXTRA_DIST = \
```
elliefm: Leave this a `+=` with the initialisation back at the top, then if we need to add back more conditional stuff we don't have to chase the declaration around.

Makefile.am
```diff
@@ -960,15 +934,6 @@ imap_imapd_LDADD = $(LD_SIEVE_ADD) $(LD_SERVER_ADD)
 imap_ipurge_SOURCES = imap/cli_fatal.c imap/ipurge.c imap/mutex_fake.c
 imap_ipurge_LDADD = $(LD_UTILITY_ADD)
 
-nodist_imap_libcyrus_imap_la_SOURCES = \
```
elliefm: Leave these as `nodist_` I think, these files don't need to be distributed.  If they were in EXTRA_DIST, that's the bug, not this

Makefile.am
```diff
@@ -1219,10 +1190,12 @@ imap_httpd_SOURCES = \
 	imap/http_applepush.c \
 	imap/http_caldav.c \
 	imap/http_caldav.h \
+	imap/http_caldav_js.h \
```
elliefm: Yep, the foo_js.h files depend on `xxd`, so these ones definitely need to be in the distribution.

elliefm:
(Looks like we're also unnecessarily distributing the foostats.[ch] files (their only build dependency is perl) but I've already got a PR in to get rid of SNMP entirely, so they'll disappear when that's merged.)

Makefile.am
dilyanpalauzov: No.  Since BUILT_SOURCES contains imap/imap_err.c I see no reason why it should contain imap/jmap_err.h instead of imap/jmap_err.h.

Makefile.am
dilyanpalauzov: EXTRA_PROGAMS is not utilized anyhow in Makefile.am , so there is no need for this variable.

Makefile.am
dilyanpalauzov: There is no reason to include files in EXTRA_DIST under conditional paths: either the file is included by “make dist” or it is not included.  Hence a single EXTRA_DIST in Makefile.am for all relevant files is sufficient.

Makefile.am
elliefm: Oh wait, that's my bad, I misread "imap" and "jmap" as being the same, whoops!

elliefm:
I managed to use the github "resolve conflicts" web tool (!) to fix up a conflict with 320aef9e34, so this might be ready to merge, just waiting for CI to finish on the new head...

PR #2718, opened 2019-03-27
GH#2665 imap/http_caldav_sched.c:propcmp() compare in the DATE-TIME properties the absolute value

dilyanpalauzov:
Comparing only the value with icalproperty_get_value_as_string() ignores the TZ information and does not trigger a change, when the timezone changes, the event is uploaded, the relative value remains the same, but the value converted to UTC does change.


dilyanpalauzov:
icalproperty_get_datetime_with_component() is int the yet released libical 3.0.5.

dilyanpalauzov:
Time to insist on libical 3.0.5 for compiling Cyrus?

PR #2723, opened 2019-03-30
imap/http_caldav.c:caldav_put() compare the ORGANIZERs case-insensitive

dilyanpalauzov:
PUTting a
```
BEGIN:VCALENDAR
BEGIN:VEVENT
ORGANIZER;CN=A B:mailto:a.b@example.org
END:VEVENT
BEGIN:VEVENT
ORGANIZER:MAILTO:a.b@example.org
END:VEVENT
END:VCALENDAR
```
shall not trigger the CalDAV:same-organizer-in-all-components precondition.

So “mailto:a.b@example.org” and “MAILTO:a.b@example.org” must be compared
with strcasecmp().

### DIFF
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -3992,7 +3992,7 @@ static int caldav_put(struct transaction_t *txn, void *obj,
         if (prop) nextorg = icalproperty_get_organizer(prop);
         /* if no toplevel organizer, use the one from here */
         if (!organizer && nextorg) organizer = nextorg;
-        if (nextorg && strcmp(organizer, nextorg)) {
+        if (nextorg && strcasecmp(organizer, nextorg)) {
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
I want to remind, that at least the `mailto:` part of the ORGANIZER must be compared case-insensitively.

[Evolution used to](https://gitlab.gnome.org/GNOME/evolution/-/issues/382) upload iCalendar objects, where each component had same ORGANIZER, where the mailto: part was spelled differently in the different components, but otherwise the ORGANIZER was the same.

Comparing email addresses in ORGANIZER (that could in theory be PRINCIPAL URI of any form), shall be done also case-insesitively, as email addresses are case-insesitive.

ksmurchison:
In theory, the local-part of an email address CAN be case-sensitive, so comparing the entire address in a case-insensitive manner would be incorrect.

PR #2733, opened 2019-04-06
Makefile.am: make sure the output directory for fig2dev exists

dilyanpalauzov:
This is essential for VPATH builds.


elliefm:
Thanks :)

PR #2734, opened 2019-04-06
configure.ac: enable implicitly the unit tests, if cunit is found

dilyanpalauzov:
Otherwise cunit/registers.h is not built during “make distcheck” and the latter fails.


elliefm:
Thanks :)

PR #2740, opened 2019-04-17
sieve: add x-cyrus-log capability and log action

### LINKED ISSUE #2281: Sieve: log action
I would like a Sieve action that will log something to syslog. Ideally, it would also log the message id of the message triggering it, the username of the script, etc (context), and the provided text.

The idea here is that we would include these log actions in our generated Sieve scripts so we could determine which UI rule a message hit. This helps our support staff assist users with their Sieve scripts.

Presumably this is gated behind a "cyrus-log" require feature or something like that.

A simple limiter might be necessary, like, no more than N log actions per script run, just to stop user-provided scripts doing something stupid.

robn:
Adds a `x-cyrus-log` capability, which adds a simple `log STRING` action. When encountered, sends the text along with user and message info to the server log.

This script:

```sieve
require "x-cyrus-log";
log "hello";
```

yields:

```
2019-04-13T20:59:36.695849-04:00 fastmail1 slotf1t01/lmtp[1369640]: sieve log: userid=robn@fastmaildev.com messageid=<ad564e7f-24ba-41c3-8968-a8efaae7e18d@www.fastmaildev.com> text=hello
```

I am not committed to naming of either the capability or the action.

There's no rate limiting or other controls, as it wasn't really obvious what any given deployment would want. Obviously there's little value in allowing users to write their own log actions. The value comes from script generators, which can now arrange for additional information to be logged so they can be correlated with rules from a rule builder, for example.

Closes #2281.

### DIFF
--- a/sieve/interp.c
+++ b/sieve/interp.c
@@ -181,6 +181,9 @@ EXPORTED const strarray_t *sieve_listextensions(sieve_interp_t *i)
 
         if (config_sieve_extensions & IMAP_ENUM_SIEVE_EXTENSIONS_MAILBOXID)
             buf_appendcstr(&buf, " mailboxid");
+
+        if (config_sieve_extensions & IMAP_ENUM_SIEVE_EXTENSIONS_X_CYRUS_LOG)
+            buf_appendcstr(&buf, " x-cyrus-log");
     }
 
     return i->extensions;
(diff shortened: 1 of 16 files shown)


sieve/interp.c
```diff
@@ -181,6 +181,9 @@ EXPORTED const strarray_t *sieve_listextensions(sieve_interp_t *i)
 
         if (config_sieve_extensions & IMAP_ENUM_SIEVE_EXTENSIONS_MAILBOXID)
             buf_appendcstr(&buf, " mailboxid");
+
+        if (config_sieve_extensions & IMAP_ENUM_SIEVE_EXTENSIONS_MAILBOXID)
+            buf_appendcstr(&buf, " x-cyrus-log");
     }
```
ksmurchison: @robn  You missed changing the config option name after the cut-n-paste

Issue #2729, opened 2019-04-03
conversations: consider more forms of whitespace when normalising subjects


robn:
I have an email here with a U+00A0 (non-breaking space) in the subject (from a UTF-8 MIME-encoded header).

(this, basically: `Subject: =?UTF-8?Q?=C2=A0?=`).

It then has a reply, but the sending client has flattened that into a regular ASCII space.

`conversation_normalise_subject` flattens whitespace, but only a very limited amount of ASCII whitespace.

As a result, these two subjects don't end up matching, and so become two separate conversations.

The cheap fix is to just add U+00A0 to `whitespace_re`, resulting in something like: `([ \t\r\n]|\xC2\xA0)+`. And, perhaps adding other UTF-8 whitespace. Assuming the internal format is UTF-8.

But maybe something more comprehensive is in order? As you like!

elliefm:
There's also a test for this function, `static void test_subject_normalise(void)` in `cunit/conversations.testc` (https://github.com/cyrusimap/cyrus-imapd/blob/master/cunit/conversations.testc#L1314-L1342).  But of course, it's all in the internal format (whatever that is)...

We're using the POSIX `regcomp()` API here, but we are also generally using the libpcre implementation of it, which we try to build with utf-8 support, sooooo, I'm not really sure what the ramifications of utf-8 support are here.  At a guess, if we wanted to do utf-8 aware character class matching (vs the "cheap fix") we may need to use libpcre's native API rather than its POSIX wrapper.  (But, see also #1731 for discussion on ongoing libpcre woes.  And #2399 re switching to re2 as first-preference regex engine.)

> Assuming the internal format is UTF-8.

I don't actually know what the internal format is here... can someone who knows please clarify? (@rsto, @brong?)  I would also assume UTF-8, but since the tests are already in the internal format I can't just drop the encoded subject in and see what comes out...

rsto:
The normalise_subject function assumes the subject to already be decoded according to RFC 2047. It's safe to assume that it is UTF-8 because that's the internal format returned by our charset library.

elliefm:
Thanks for the clarification -- I knew it expected it to be already decoded, but didn't know our internal format :)

elliefm:
Quick fix is now on master and 3.0, with cunit and cassandane tests :)

PR #2745, opened 2019-04-22
whitespace_re: take as much "simple" whitespace as possible before recursing

robn:
Every alternation adds a couple of stack frames for backtracking, and since we're only taking one space at a time, a long run of spaces can easily blow the stack.

Its still possible to break this by doing a long run of non-breaking spaces, or alternating simple whitespace & nbsp, and there's not an easy way to fix that because you can't include the \xC2\xA0 sequence in a character class. Improvements might include use of `[:space:]` or `\p{Z}` if PCRE's POSIX compat mode supports, or use of things like atomic groups if we were willing to switch to using PCRE proper.

Related: #2729.

### DIFF
--- a/imap/conversations.c
+++ b/imap/conversations.c
@@ -727,7 +727,7 @@ EXPORTED void conversation_normalise_subject(struct buf *s)
     int r;
 
     if (!initialised_res) {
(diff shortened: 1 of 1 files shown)


elliefm:
cherry-picked to 3.0 (and also Bron's later fix-fix)

PR #2768, opened 2019-05-25
Update documentation for squatter

dilyanpalauzov:
• State, that -v can be used in any mode
• State that in squatter -u *users* and squatter *mailboxes* the items are space separated.
• In the synopsis, consistently surround options with spaces.
• Remove references of “squatter -I"
• Mention -A and -P options.


elliefm:
Thanks :)

PR #2770, opened 2019-05-27
allow building against clamav >= 0.101

moench-tegeder:
https://github.com/Cisco-Talos/clamav-devel/commit/048a88e changed
the handling of scan options, removing the CL_SCAN_STDOPT macro.
This change is already in the cyrus-imapd-3.0 branch as
09b5bf11e12185727d53a6ae06d71a767952a537 and fd72fcef79a09cc3e349ee36bef1904e6e8e1030
but missing from the still maintained 2.5 branch.


elliefm:
Thanks :)

PR #2771, opened 2019-05-27
Fix README links

moench-tegeder:
Some links where pointing to stale locations. While here, have links
point to https sites (the http server redirects to https) and document
the current "current branch".


elliefm:
Nice, thanks! I'll cherry pick this to the other branches too

PR #2776, opened 2019-05-28
docsrc/murder-installation: mention that mupdate does not work with STARTTLS

dilyanpalauzov:


elliefm:
Thanks, will cherry-pick to master too

dilyanpalauzov:
In 3.0 and master the notes are at different places.

PR #2788, opened 2019-06-07
_email_query: emailsearch_getcachepath() takes no arguments

wolfsage:
Why does gcc allow this?!

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -2730,7 +2730,7 @@ static void _email_query(jmap_req_t *req, struct jmap_query *query,
     query->query_state = _email_make_querystate(current_modseq, 0);
 
     /* Open cache */
-    cache_fname = emailsearch_getcachepath(req->accountid);
+    cache_fname = emailsearch_getcachepath();
     if (cache_fname) {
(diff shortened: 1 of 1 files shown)


elliefm:
The prototype has empty parens, which according to C rules it's making no statement about what parameters are valid, and so it accepts any parameters...

elliefm:
I edited the function to specify `void` in the parameter list, and now the build fails correctly! :joy:

```
imap/jmap_mail.c: In function ‘_email_query’:
imap/jmap_mail.c:2733:19: error: too many arguments to function ‘emailsearch_getcachepath’
     cache_fname = emailsearch_getcachepath(req->accountid);
                   ^~~~~~~~~~~~~~~~~~~~~~~~
imap/jmap_mail.c:2473:14: note: declared here
 static char *emailsearch_getcachepath(void)
              ^~~~~~~~~~~~~~~~~~~~~~~~
```

elliefm:
b4d91d8c4 fwiw

elliefm:
("empty parens == no arguments" is the C++ behaviour, but this is a C file)

wolfsage:
Ahhhh interesting. Yeah I was compiling with clang to see what kind of issues popped up and this was one

PR #2789, opened 2019-06-07
_email_import: Initialize variables before we hit a goto

wolfsage:
...so we don't try to free them later without initializing them.


elliefm:
Thanks!

PR #2796, opened 2019-06-15
imap/message.[ch]: remove unused message_ref()

dilyanpalauzov:


elliefm:
I don't think we should remove this.  Even though it's not currently used, it forms part of the message API, and the ref/unref pair makes it clear to future developers how the pattern works, even if they got their ref indirectly through one of the other functions.

dilyanpalauzov:
`message_ref()` was never ever used since its birth in 2012.  The current code is obviously bloated and deserves some attention.

ksmurchison:
I agree with @elliefm.  Just because we aren't currently using it doesn't mean we won't in the future.  And perhaps we SHOULD be using it somewhere now.  And the amount of bloat here is negligible.

PR #2798, opened 2019-06-15
imap/message.[ch]: remove unused msg_ functions

dilyanpalauzov:


elliefm:
These are all accessors with assertions in them... very useful!  I would argue that we should use them, rather than remove them.

ksmurchison:
I agree with @elliefm.  These should stay, and possibly be used rather than directly accessing the fields.

PR #2802, opened 2019-06-21
imap/spool.c: less xstrdup() calls

dilyanpalauzov:
* introduce xstrduplcase(), which is faster than lcase(xstrdup())
* make sure the header names passed to spool_getheader(), spool_append_header(), spool_prepend_header(), spool_replace_header() are lower cased, so that the function does not have to make a copy of the parameter, lower-case it, use it, and free() the copy
* make sure the header names passed do spool_remove_header() are lower-cased and spool_remove_header() does not free() the header name.  In turn, the caller does not have to duplicatethe header name, before passing it to the function.

### DIFF
--- a/cassandane/tiny-tests/Sieve/editheader_basic
+++ b/cassandane/tiny-tests/Sieve/editheader_basic
@@ -44,8 +44,8 @@ EOF
 
     $msg1 = $res->{1}->{rfc822};
 
-    $self->assert_matches(qr/^X-Cassandane-Test: prepend1\r\n/, $msg1);
-    $self->assert_matches(qr/X-Cassandane-Test: append1\r\nX-Cassandane-Test: append3\r\nX-Cassandane-Test: append5\r\n\r\n/, $msg1);
+    $self->assert_matches(qr/^x-cassandane-test: prepend1\r\n/, $msg1);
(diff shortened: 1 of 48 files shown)


ksmurchison:
Removing xstrdup() calls seems like a good idea, but unfortunately this PR is out of date.  If you want to update it in the next couple of weeks, we can take another look.

Also, how is xstrduplcase() faster than lcase(xstrdup)?

dilyanpalauzov:
> how is xstrduplcase() faster than lcase(xstrdup)?

xstrdup iterates twice over the string - to calculate the length and to copy the bytes.  lcase iterates one more time, so three in total.  Moreover, each byte is written twice.

xstrduplcase() iterates twice in total and writes each bytes once.

dilyanpalauzov:
Is in http_h2.c:header_cb()  `my_name` always lower-cased string?

ksmurchison:
> Is in http_h2.c:header_cb() `my_name` always lower-cased string?

Yes, since HTTP/2 sends lowercase headers over the wire.

ksmurchison:
After discussing this with the Cyrus dev team, we have decided that we don't want the callers to have to worry about lowercasing header names.  Also, we will use a struct buf to do the lowercasing which will reduce the number of xstrdup() and free().

This is done in #5045

PR #2805, opened 2019-06-23
imap/spool.c: reduce amount of xstrdup() calls (3.0)

dilyanpalauzov:
* Introduce xstrduplcase(), which is faster than lcase(xstrdup())

* Make sure the header names passed to spool_getheader()  are lower cased, so that the function does not have to make a copy of the parameter, lower-case it, use it, and free() the copy

* Make sure the header names passed do spool_cache_header(), spool_remove_header() and spool_replace_header() are lower-cased and spool_remove_header() does not free() the header name. In turn, the caller does not have to duplicate the header name, before passing it to the functions.


ksmurchison:
No longer maintaining 3.0.x

PR #2809, opened 2019-06-25
imap/icu_wrap.cpp: C++ tweaks

dilyanpalauzov:
The result of string.c_str() cannot be NULL, so
  tzid = str.c_str()
  if (tzid)
does nothing.

Remove variable char *tzid, as it is not really needed.

Call only one constructor for uWinID.

There is no point to assign NULL to utf8cnv, as the value is never read.


rsto:
Thanks. I also moved the variable declarations close to their first use.

PR #2810, opened 2019-06-25
imap/xapian_wrap.cpp: C++ tweaks

dilyanpalauzov:
For iterators, prefer ++it over it++.  The former creates a copy of the iterator, modifies the original and returns the copy.  The latter modifies the original and returns it, without creating copies.

In the constructor of CyrusSearchStemmer pass "en" directly to the member stem of type Xapian::Stem, which will call the Xapian::Stem constructor, instead of creating a Xapian::Stem("en") object, and copying its result in the member stem.

In xapian_snipgen_add_match: abolish variable r.

Change the type of make_lang_count_key to accept as second parameter const std::string&.  This avoids converting std::string to char* and constructing from it a temporary std::string, instead of using all the time the original string.

In xapian_snipgen_begin_doc: avoid calling two constructors for snippet.

### DIFF
--- a/imap/xapian_wrap.cpp
+++ b/imap/xapian_wrap.cpp
@@ -1,4 +1,3 @@
-
 #include <errno.h>
 #include <config.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


rsto:
Thanks!

PR #2812, opened 2019-06-26
imap/xapian.cpp: further C++ tweaks

dilyanpalauzov:


dilyanpalauzov:
At line 96 `stoppers.at(iso).get();` I missed a `return`.  Correct is
```C
     // Lookup cached entry.
     try {
        return stoppers.at(iso).get();
     } catch (const std::out_of_range&) {};
```

PR #2818, opened 2019-06-30
imap/xapian.cpp: more C++ tweaks

dilyanpalauzov:
* xapian_dbw’s members stemmer and stopper are removed.  The former was not used.  The value of the latter was always nullptr and this value is passed to dbw→term_generator.set_stopper().

* set the visibility to "hidden" to internal symbols

* remove the word “struct” for defining a type, as it is not necessary in C++.

### DIFF
--- a/imap/xapian_wrap.cpp
+++ b/imap/xapian_wrap.cpp
@@ -43,7 +43,7 @@ extern int charset_flags;
 
 /* ====================================================================== */
 
-static void make_cyrusid(struct buf *dst, const struct message_guid *guid, char doctype)
+static void make_cyrusid(buf *dst, const message_guid *guid, char doctype)
 {
     buf_reset(dst);
     buf_putc(dst, '*');
@@ -66,7 +66,7 @@ static const Xapian::Stopper* get_stopper(const std::string& iso)
 
     // Lookup cached entry.
     try {
-        stoppers.at(iso).get();
+        return stoppers.at(iso).get();
     } catch (const std::out_of_range&) {};
 
     // Lookup language name by ISO code.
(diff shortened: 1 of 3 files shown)


imap/xapian_wrap.cpp
```diff
@@ -417,82 +415,73 @@ void xapian_check_if_needs_reindex(const strarray_t *sources, strarray_t *torein
 
 struct xapian_dbw
 {
-    Xapian::WritableDatabase *database;
-    Xapian::Stem *stemmer;
-    Xapian::TermGenerator *term_generator;
-    Xapian::Document *document;
+    Xapian::WritableDatabase database;
```
rsto: I remember we ran into trouble passing C++ objects in this struct to the C code. That's why we used pointers. But I can't replicate that issue with my current setup, so probably it's due to and old compiler version we are using. I'll check.

elliefm:
@rsto You reviewed this in 2019 and requested changes.  Looks like Dilyan made some changes, but you didn't come back and re-review.  Can you have a look again please?

@dilyanpalauzov Can you rebase this on current master and fix the conflicts?  Or just close it if it's no longer useful.  Thanks

dilyanpalauzov:
I think these changes are good, they switch from pointers to C++ objects; utilize C++ syntax for methods with `try {…} catch {…}` being the sole function body, without surrounding `{}`.

If I update the changes now, so that these can be applied clearly towards the current master branch, it is in no way evident, whether in five years I will be asked again to refresh the snippets, so that they can be applied without conflicts to the master branch.

So I think it would be good if somebody looks at the proposed changes, and adjusts them to the current tip of git, but I am personally not going to do this.

rsto:
> So I think it would be good if somebody looks at the proposed changes, and adjusts them to the current tip of git, but I am personally not going to do this.

@dilyanpalauzov I am sorry that I had first asked for changes, then didn't follow up. I should have asked from the start to dissect this commit into smaller units of change, e.g. one for each stylistic or idiomatic update, which would make it simpler to reason about. I won't ask you do that now but will rather look at the current PR as-is and then decide how to proceed with it.

rsto:
I'm inclined to close this pull request. Again, my apologies for having it linger around so long. I think we should approach this with more structure and smaller changes and I have started this effort with https://github.com/cyrusimap/cyrus-imapd/pull/5017 . I think that we should start using clang-tidy as a common base to define what "sane C++" looks like and I appreciate any additional PRs or contributions that follow this direction.

PR #2819, opened 2019-07-02
mboxname: uniquer mboxname_todeleted suffixes

elliefm:
If the same mboxname gets deleted twice in the same second (i.e. delete-create-delete) the latter would get the same new name and be refused (NO).

This is not a real problem, but imaptest runs fast enough to tickle it and the false positives are annoying.

Mixing the timestamp's microseconds into the existing suffix (rather than appending or something else) to avoid breaking middleware that wants to pattern match these filenames (FM has some of these)


elliefm:
from bron: breaks cyr_expire, doh. can't  mix the microseconds into the blat because it's actually not just an opaque blat

PR #2831, opened 2019-07-22
httpd: print slash after LibXML in the version info footer (3.0)

dilyanpalauzov:
All other software uses slashes, like: ` Cyrus-HTTP/3.0.10 Cyrus-SASL/2.1.27 OpenSSL/1.1 Zlib/1.2.11 Brotli/1.0.7 LibXML2.9.9 SQLite/3.29.0 LibiCal/3.0 ICU4C/63.1 Jansson/2.12 Server at …`

### DIFF
--- a/imap/httpd.c
+++ b/imap/httpd.c
@@ -953,7 +953,7 @@ int service_init(int argc __attribute__((unused)),
     buf_printf(&serverinfo, " Brotli/%u.%u.%u",
                (version >> 24) & 0xfff, (version >> 12) & 0xfff, version & 0xfff);
 #endif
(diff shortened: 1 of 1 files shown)


elliefm:
Seems reasonable.  

The / used to be there, but was removed by @ksmurchison in 8bb22524f8, not sure why?  Haven't made the corresponding change on master yet, but it would go here: https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/httpd.c#L744

ksmurchison:
Looks like removing the slash was accidental or stupid.  Either way it was wrong

PR #2835, opened 2019-07-24
mboxlist: add 'p' callback to mboxlist_find*() API

### LINKED ISSUE #2834: expose cyrusdb `find_p` callback in mboxlist_find* APIs
This will allow some additional context-aware optimisation in big mboxlist_find* lookups

elliefm:
Exposes the 'p' callback from the cyrusdb_foreach API via the mboxlist_find* APIs, so that if you need to do a cheap custom check on matches before calling the expensive callback, you can.

Nothing uses this at the moment, so all the interesting changes in this PR are the ones in mboxlist.[ch], everything else is just noise due to the number of params changing.

It'll be useful for #2833 and closes #2834

### DIFF
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -2759,7 +2759,8 @@ struct find_rock {
     mbname_t *mbname;
     mbentry_t *mbentry;
     int matchlen;
-    findall_cb *proc;
+    findall_p *p;
+    findall_cb *cb;
     void *procrock;
 };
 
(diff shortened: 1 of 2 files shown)


elliefm:
If/when this is merged, it'll probably conflict with everyone else's WIP branches, just because it needs to change every call that uses this API!

ksmurchison:
This looks sane to me

brong:
A much less invasive way to do this is the trick I decided to use in a recent commit:

https://github.com/brong/cyrus-imapd/commit/361cddb6be0fa5fd5d37afaf6829b48614d49653

Basically: create a more advanced API with a new name, wrap it with the old API, and only use the new API from the one place.

brong:
(The reason I'd recommend that trick now is precisely because there's some huge other branches out there... so it's nice not to make invasive changes now.  We can rebase this all out later when we want to rationalise the APIs)

elliefm:
Yeah wow that's really obvious in hindsight! 😂  Is there any particular reason you kept the old-name-wrapper as a proper function, rather than just a macro?

brong:
Nah, not really.  Often because there's more smarts than just a macro-worth in wrappers (like the various shapes of mailbox create), but a macro would work fine in this case.

PR #2838, opened 2019-07-26
Initial implementation of JMAP delayed send (future release)

ksmurchison:


imap/caldav_alarm.c
```diff
@@ -865,20 +838,169 @@ static void process_one_record(struct mailbox *mailbox, uint32_t imap_uid,
...
 }
 
+static void process_futurerelease(struct mailbox *mailbox,
+                                    struct index_record *record)
+{
+    struct buf buf = BUF_INITIALIZER;
+    json_t *submission = NULL, *identity, *envelope;
+    int r = 0;
+
+    syslog(LOG_DEBUG, "processing future delivery");
```
brong: This log message should probably contain the mailbox name and record->uid.

imap/caldav_alarm.c
```diff
@@ -865,20 +838,169 @@ static void process_one_record(struct mailbox *mailbox, uint32_t imap_uid,
...
+    close(msgfd);
+    if (!buf_len(&buf)) {
+        syslog(LOG_ERR, "process_futurerelease: can't mmap %s: %m", fname);
+        goto done;
+    }
+
+    /* Parse the submission object from the first header field */
+    json_error_t jerr;
+    size_t hdrlen = strlen(JMAP_SUBMISSION_HDR) + 1; /* +1 for ':' */
+    submission = json_loadb(buf_base(&buf) + hdrlen, buf_len(&buf) - hdrlen,
```
brong: We should check that this actually matches the JMAP Submission Header as well...

imap/caldav_alarm.c
```diff
@@ -865,20 +838,169 @@ static void process_one_record(struct mailbox *mailbox, uint32_t imap_uid,
...
+    if (!submission) {
+        syslog(LOG_ERR,
+               "process_futurerelease: failed to parse submission obj");
+        goto done;
+    }
+    envelope = json_object_get(submission, "envelope");
+    identity = json_object_get(submission, "identityId");
+
+    /* Trim submission header field */
+    const char *p = strstr(buf_base(&buf), "}\r\n") + 3; /* +3 for '}\r\n' */
```
brong: This feels dangerous in conjunction with arbitrary content and presumably header wrapping - we could wind up not removing the entire header

imap/caldav_alarm.c
```diff
@@ -865,20 +838,169 @@ static void process_one_record(struct mailbox *mailbox, uint32_t imap_uid,
...
+static void process_futurerelease(struct mailbox *mailbox,
+                                    struct index_record *record)
+{
+    struct buf buf = BUF_INITIALIZER;
+    json_t *submission = NULL, *identity, *envelope;
+    int r = 0;
+
+    syslog(LOG_DEBUG, "processing future delivery");
+
+    if ((record->internal_flags & FLAG_INTERNAL_EXPUNGED) ||
```
brong: This DELETED test could probably be lifted up into the upper loop - we're not going to want to process calendar events if the record has been marked DELETED either really, though it will probably never happen.

imap/caldav_alarm.c
```diff
@@ -865,20 +838,169 @@ static void process_one_record(struct mailbox *mailbox, uint32_t imap_uid,
...
+    r = mailbox_rewrite_index_record(mailbox, record);
+    if (r) {
+        syslog(LOG_ERR, "expunging record (%s:%u) failed: %s",
+               mailbox->name, record->uid, error_message(r));
+    }
+
+  done:
+    if (r) {
+        /* XXX  Error handling: retry, DSN, both, neither? */
+        caldav_alarm_delete_record(mailbox->name, record->uid);
```
brong: Yeah, this is a bit annoying.  A dav_reconstruct or similar would presumably re-add this record, because we haven't updated the record at all.  We do want to think about this some more.

imap/caldav_alarm.c
```diff
@@ -865,20 +838,169 @@ static void process_one_record(struct mailbox *mailbox, uint32_t imap_uid,
...
+
+    if (mailbox->mbtype == MBTYPE_CALENDAR) {
+        process_valarms(mailbox, &record, floatingtz, runtime);
+    }
+#ifdef WITH_JMAP
+    else if (mailbox->mbtype == MBTYPE_SUBMISSION) {
+        process_futurerelease(mailbox, &record);
+    }
+    else if (mailbox->i.options & OPT_IMAP_HAS_ALARMS) {
+        /* XXX  Check special-use flag on mailbox */
```
brong: We should probably at least syslog here "something odd happening!"

Also if we don't clear the alarm, we're going to infinite loop on this I guess?  We need to make sure we always make some progress.

imap/jmap_support.c
```diff
@@ -0,0 +1,86 @@
+/* jmap_mail_submission.c -- Routines for handling JMAP mail submission
```
brong: This doesn't match the filename

PR #2852, opened 2019-08-19
Emailsubmission useflags

brong:
Hi Ken, popping this one up for your review before it goes live.  We discovered that cancel was storing UNLINKED which breaks replication and made things sad - so I reviewed and came up with this proposal.  Also talked to Neil about whether we wanted EmailSubmission objects to persist past the submission and he said yes.  We'll probably have to make cyr_expire clean them up eventually, but that's future-us' problem.

### DIFF
--- a/imap/jmap_mail_submission.c
+++ b/imap/jmap_mail_submission.c
@@ -359,7 +359,7 @@ static int store_submission(struct mailbox *mailbox,
         /* Already sent */
         msglen = 0;
         holduntil = time(0);
-        strarray_append(&flags, "\\Expunged");
+        strarray_append(&flags, "\\Answered");
     }
 
     /* Prepare to stage the message */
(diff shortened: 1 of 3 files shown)


imap/jmap_mail_submission.c
```diff
@@ -909,15 +909,11 @@ static void _emailsubmission_update(struct mailbox *submbox,
...
-                        if (!(record->internal_flags & FLAG_INTERNAL_UNLINKED)) {
-                            /* Already sent */
-                            *set_err = json_pack("{s:s}", "type", "cannotUnsend");
-                        }
-                        else if (!strcmp(strval, "canceled")) {
-                            /* Already canceled */
-                            continue;
-                        }
+                    if (record->system_flags & FLAG_ANSWERED) {
+                        if (!strcmp(strval, "final")) continue;
```
ksmurchison: The cannotUnsend error is missing for the "final" case

ksmurchison:
OK, so after testing I found one potential problem:
mailbox_update_email_alarms() is used for both futuresend and snooze.  And since a client can sent \flagged and \answered on emails prior to snoozing, the necessary alarms won't be set due to the new flag checks.

I see two ways around this:
1. Use new internal flags for undoStatus (which don't break replication)
2. Split mailbox_update_email_alarms() into separate cases/functions depending on which mbtype/special-use we are dealing with

imap/jmap_mail_submission.c
brong: hmm yeah good point.

ksmurchison: Already fixed after merge.

ksmurchison:
I chose option #2 after merging

PR #2854, opened 2019-08-28
squatter.rst: write down that -F is broken for 3.0

dilyanpalauzov:
See https://lists.andrew.cmu.edu/pipermail/cyrus-devel/2019-June/004487.html


elliefm (approved):
Thanks

PR #2857, opened 2019-08-28
Access deleted data

brong:
This branch adds the ability to:

* list deleted mailboxes using VENDOR.FM-INCLUDE-DELETED
* select deleted mailboxes using VENDOR.FM-ALLOW-DELETED
* see \Expunged messages by selecting mailboxes using VENDOR.FM-INCLUDE-EXPUNGED

There isn't a place (yet) in Alt Namespace for these DELETED folders to appear, but maybe one day we will create one.

The intent of this is a framework on which to build restore tools.

### DIFF
--- a/imap/arbitron.c
+++ b/imap/arbitron.c
@@ -242,6 +242,7 @@ static void usage(void)
 static int do_mailbox(struct findall_data *data, void *rock __attribute__((unused)))
 {
     if (!data) return 0;
+    if (!data->is_exactmatch) return 0;
     int r;
     struct mailbox *mailbox = NULL;
     const char *name = mbname_intname(data->mbname);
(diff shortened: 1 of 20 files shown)


elliefm:
I just tried running Cassandane across this branch and it's barfing hard??

So far both of the erroring tests I've looked at are failing to create inbox (here's one):

1) test_append(Cassandane::Cyrus::Conversations)
 Perl exception: Cannot create folder "inbox": IMAP Command : 'create' failed. Response was : no - Mailbox already exists
 at Cassandane/IMAPMessageStore.pm line 159.

brong:
Yeah!  I gotta push the fix.  Fricking reversed logic on the LIST.  Thanks for checking :)

elliefm:
This is conflicting, but it looks like it's conflicting because a slightly modified version has already been merged to master?  So I think we can just close this (and delete the branch)?

Issue #1845, opened 2017-02-20
Make notification from cyr_virusscan localizable and/or customizable


hagedose:
Currently the notification that cyr_virusscan generates when an infected message has been deleted (and notifications are enabled) is hard-coded in English. I could live with an option for a site to customize that notification, but other sites might want to have localized notifications. @elliefm suggested that template support from CALDAV might be reused here.

elliefm:
https://github.com/cyrusimap/cyrus-imapd/issues/2516#issuecomment-529267173

PR #2861, opened 2019-09-06
configurable cyr_virusscan notification text

### LINKED ISSUE #2516: Message text for cyr_virusscan should be controllable
The current message template for reporting viruses, in cyr_virusscan, is hard coded in append_notifications().  This should be controllable by the admin, to allow for local site preference, or language needs.

elliefm:
Adds two new imapd.conf options, for the notification subject and a template digest chunk respectively.  The defaults are the same as the old version, but converted to use the template style message construction.

Fixes the "From:" field of the digest chunks to look like a normal from header, instead of a dlist straight out of mailbox cache.

Fixes how the "mailbox" is reported to calculate an extname correctly

Failure to append a notification is now logged(!), and the template is checked early so if it looks un-appendable, we can bail out before deleting messages, instead of deleting them then being unable to notify that we did so

Notification messages are now properly MIME-encoded (I think!), so non-ascii mailbox names won't produce an invalid notification, and the custom template is expected to be in utf-8 so non-English notifications should Just Work.

Cassandane tests in https://github.com/cyrusimap/cassandane/compare/master...elliefm:v31/2516-virusscan-custom-notification-text

Closes #2516 and #1845

### DIFF
--- a/cunit/buf.testc
+++ b/cunit/buf.testc
@@ -1592,6 +1592,51 @@ static void test_appendoverlap(void)
     buf_free(&buf);
 }
 
+static void test_tocrlf(void)
+{
+    struct testcase {
+        char *input;
+        char *expect;
+    };
+
+    struct testcase testcases[] = {
+        { "\rleading bare cr", "\r\nleading bare cr" },
+        { "\nleading bare lf", "\r\nleading bare lf" },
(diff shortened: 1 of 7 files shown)


ksmurchison:
Looks sane to me.

PR #2868, opened 2019-09-10
Add compatibility for Sphinx 2 (fixes #2732)

### LINKED ISSUE #2732: sphinx-doc 2.0 cannot be used for man pages, but 1.8.5 works
In Sphinx-doc 2.0, thanks https://github.com/sphinx-doc/sphinx/commit/98a9b50c52577eb08e6a29ecfcf0662ca81babba, MACRO_DEF is not imported anymore in sphinx/writers/manpage.py, so since sphinx-doc 2.0.0b1:
```
docsrc/exts/sphinxlocal/writers/manpage.py-16-from sphinx.writers.manpage import (
docsrc/exts/sphinxlocal/writers/manpage.py:17:    MACRO_DEF,
docsrc/exts/sphinxlocal/writers/manpage.py-18-    ManualPageWriter,
docsrc/exts/sphinxlocal/writers/manpage.py-19-    ManualPageTranslator as BaseTranslator
docsrc/exts/sphinxlocal/writers/manpage.py-20-)
```
does not work any more.

J5lx:
This allows the documentation to be generated using version 2 of Sphinx – which is particularly handy on rolling release distros like Arch where version 2 is the only version conveniently available from the repositories. Nearly all of the changes are directly based on changes in upstream Sphinx, however they have been adapted to also keep compatibility with version 1.3.2 (which is the minimum version required by the current configuration and not much older than 1.3.6 – the version used to generate the website. On that note, the conf.py in master incorrectly states a minimum version of 1.2, so I updated that while I was at it). Should that much backwards compatibility not actually be necessary, I could also simplify the changes quite a bit. I *think* with a minimum version of 1.6 none of the BC mechanisms should be needed.

If these changes are okay, it would be nice to have them in 3.0 as well.

### DIFF
--- a/docsrc/conf.py
+++ b/docsrc/conf.py
@@ -25,7 +25,7 @@ sys.path.insert(0, os.path.abspath('exts'))
 # -- General configuration ------------------------------------------------
 
 # If your documentation needs a minimal Sphinx version, state it here.
-needs_sphinx = '1.2'
+needs_sphinx = '1.3.2'
 
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
--- a/docsrc/conf.py
+++ b/docsrc/conf.py
@@ -301,7 +301,7 @@ current = os.path.abspath(os.getcwd())
 for tuple in pathset:
     os.chdir(tuple[0])
     for rstfile in glob.glob("*.rst"):
-        author = [("The Cyrus Team")]
+        authors = [("The Cyrus Team")]
         orphan = 'False';
         with io.open(rstfile,'r',encoding="utf8") as f:
             for line in f:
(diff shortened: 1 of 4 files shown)


docsrc/conf.py
```diff
@@ -301,22 +301,22 @@
 for tuple in pathset:
     os.chdir(tuple[0])
     for rstfile in glob.glob("*.rst"):
-        author = [("The Cyrus Team")]
```
J5lx: Brief explanation because the need for this change might not be obvious: In Python, for loops do not affect variable scoping. This causes the author variable to bleed into the global scope, where it becomes the author configuration option. However that option needs to be a string while this variable is a list, and this causes Sphinx 2 to fail because it tries to perform a character translation on the list.

elliefm:
Just tested and it works fine on 1.4.9 (debian stretch), so I'll merge this to master and then see what the web server (1.3.6) thinks.

docsrc/conf.py
elliefm: Thanks!

elliefm:
This is amazing:

```
Running Sphinx v1.3.6
Initializing cyrusman plugin

Sphinx version error:
This project needs at least Sphinx v1.3.2 and therefore cannot be built with this version.
```

This might be why the "min version" was still set at 1.2?  I'm not gonna pretend to have any idea where/how to fix this properly, instead I've just reverted the version bump commit instead (thanks for making that a separate commit!) and that's worked.

elliefm:
The compatibility commit cherry-picked cleanly to 3.0, so I've done that too: 237b86297

J5lx:
Yeah, that is pretty amazing indeed. I just looked it up in the documentation and apparently support for the micro version part was only added in 1.4. That explains why I didn’t catch this by testing with 1.3 and 2.2. Now I feel kinda stupid lol. Anyways, thanks for merging!

elliefm:
I've added a comment to this file in both branches, so next time we scratch our heads about it still saying "1.2" we know why

PR #2874, opened 2019-09-16
imap/mboxevent.c: wrap macro FILL_STRING_PARAM in {…} (3.0)

dilyanpalauzov:
as the macro expands to several statements and calling it immediately after *if* executes only the first statement of `FILL_STRING_PARAM` conditionally and the remaining statements uncondionaly.

### DIFF
--- a/imap/mboxevent.c
+++ b/imap/mboxevent.c
@@ -319,14 +319,17 @@ EXPORTED struct mboxevent *mboxevent_new(enum event_type type)
 
     FILL_UNSIGNED_PARAM(mboxevent, EVENT_PID, getpid());
 
-    if (mboxevent_expected_param(type, EVENT_SESSIONID))
+    if (mboxevent_expected_param(type, EVENT_SESSIONID)) {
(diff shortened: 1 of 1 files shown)


elliefm:
Good catch!  Once this was applied, it exposed a bug that was depending on the dodgy unconditionality, so I cherry-picked 331991df4 from master (where it had already been fixed)

I also just went ahead fixed up the bad macros in mboxevent.h and some other ones too, so this won't be a problem in future -- 9cad6eb43 on 3.0, 8e20982d0 on master

dilyanpalauzov:
For the record, gcc 9.1.1 20190804 catched this, by emitting a warning.

dilyanpalauzov:
This is needed also for PARSE_PARTIAL in imapd.c (or at least the compiler say that after *if* there is a macor expanding to several statements and only the first statemen is executed conditionally).

elliefm:
Ahh yep, good find.  My search missed that one cause there isn't a `{` until the third line!  67c52ef86 on master, 49662f990 on 3.0

PR #2883, opened 2019-09-25
Modification of the year of the launch of Cyrus' development

lyabs243:
In order not to confuse the beginning of Cyrus it is important to note that Cyrus has been under development since late 1993 and early 1994 as reported in the code history https://www.openhub.net/p/cyrus-imap/commits/summary

### DIFF
--- a/docsrc/index.rst
+++ b/docsrc/index.rst
@@ -32,7 +32,7 @@ Features
 
 Read more in our full :ref:`list of features <imap-features>`.
 
-Cyrus has been under active development since the year 1993 when the project was launch at Carnegie Mellon University. It's used in production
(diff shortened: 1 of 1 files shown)


elliefm:
Wikipedia says the project was started in 1994 (but citing no source), so it's interesting that there's commits back to 1993.  I have no way of knowing which is correct...  The "2000" date might have come from the creation of the `master` process (a84b4249e)?

> when the project was launch 

Could you make that "launched" please :)

lyabs243:
@elliefm Thank you, i made this change.

elliefm:
Thanks! I've also cherry-picked these to the 3.0 branch :)

PR #2888, opened 2019-10-02
convert imapd.conf duration options to new DURATION type 

### LINKED ISSUE #26: timeouts with inconsistent units
We have a bunch of timeouts.  Some are specified in minutes, some are specified in seconds, one is specified in seconds but its default value is 3 _hours_, one is specified in minutes but its default value is 24 hours...

Here's the list, including default value and units:
- [x] client_timeout: 10 (seconds)
- [x] failedloginpause: 3 (seconds)
- [x] httpkeepalive: 20 (seconds)
- [x] httptimeout: 5 (minutes)
- [x] imapidlepoll: 60 (seconds)
- [x] ldap_time_limit: 5 (seconds)
- [x] ldap_timeout: 5 (minutes)
- [x] lmtptxn_timeout: 300 (seconds)
- [x] nntptimeout: 3 (minutes)
- [x] plaintextloginpause: 0 (seconds)
- [x] popminpoll: 0 (minutes)
- [x] poptimeout: 10 (minutes)
- [x] ptscache_timeout: 10800 (seconds)
- [x] rss_maxage: 0 (days)
- [x] sieve_vacation_min_response: 259200 (seconds)
- [x] sieve_vacation_max_response: 7776000 (seconds)
- [x] sync_repeat_interval: 1 (seconds)
- [x] sync_timeout: 1800 (seconds)
- [x] tcp_keepalive_idle: 0 (seconds)
- [x] tcp_keepalive_intvl: 0 (seconds)
- [x] timeout: 32 (minutes)
- [x] tls_session_timeout: 1440 (minutes!)
- [x] openio_rawx_timeout: 30 (seconds)
- [x] openio_proxy_timeout: 5 (seconds)

(There might be some I missed, I'll keep an eye out and add them here if I find them.)

Minutes are useful, human readable units for many of these -- but are a pain for testing (a test script would need to hang for a whole minute even at the lowest setting if it wanted to make sure the timeout worked)

Seconds are great for testing, are human readable for some settings (like client_timeout), but are unwieldy when dealing with long timeouts.

None of these are in hour units, even though two of them have default values that are of multiple hours duration.

These are all "INT"s in the config parser.  Seems like we could resolve all this incongruity by adding a new config value type (call it, say, "DURATION") which allows values like 10s, 30m, 24h etc.  

For backward compatibility with old config files, the `config_getduration()` (or whatever) function should take an option to specify the default unit interpretation if none is provided.  That way an old config file containing, say, "timeout: 30" doesn't accidentally get interpreted as 30 seconds when its intention was 30 minutes.

elliefm:
This adds a new config option type, DURATION, along with corresponding config_getduration() and config_parseduration() accessors, and converts all our existing relevant config options to use them.  Our existing option deprecation handling, and the default unit handling in this PR, mean this update should be seamless without requiring changes to imapd.conf.

DURATIONs are strings specifying an amount of time, such as "30s" or "15m" or "24h" or "90d".  They are parsed for validity-only at `config_read()`, and parsed for their actual value-in-seconds at the time a value is retrieved (with `config_getduration()` or `config_parseduration()`).  This is because, at the time of `config_read()`, we don't have enough context to know how to interpret numbers without units, which we need to allow for backward compatibility to older config files!  This means they are stored internally as malloced strings, like STRING and STRINGLIST, even though they act like a number.

config_getduration() and config_parseduration() both calculate and provide a value in seconds from the value specified in the config file, which works out well because we're generally using time_t for most things anyway.

All existing relevant config options have been converted to use the new DURATION type, using their historical unit as the default if none is provided.  Some options (specifically: `archive_days`, `backup_retention_days` and `conversations_expire_days`) needed a name change as part of this, so the old name has been marked as deprecated and the new name set up.

There's cunit tests for everything I could think of, and every Cassandane test that currently passes on master also passes on this branch.

Reviewers, I suggest stepping through this commit by commit rather than trying to read it as a single diff 😅

Merging this PR closes #26


lib/imapoptions
```diff
@@ -83,6 +83,13 @@ Blank lines and lines beginning with ``#'' are ignored.
 For boolean and enumerated options, the values ``yes'', ``on'', ``t'',
 ``true'' and ``1'' turn the option on, the values ``no'', ``off'',
 ``f'', ``false'' and ``0'' turn the option off.
+.PP
+Duration options take the form of a number followed by a unit, for example
+\fB32m\fR (32 minutes).  Units are \fBd\fR (days), \fBh\fR (hours), \fBm\fR
+(minutes) and \fBs\fR (seconds).  Multiple units can be combined and will
```
rsto: Nit: Days not necessarily have a duration of 24 hours, but are hard-coded as such in the current parser implementations. It would be unreasonable to support day light saving time for evaluation of the configured items, but let's just be explicit here and state "Day durations are multiples of 24 hours"?

imap/http_caldav_sched.c
```diff
@@ -2474,14 +2474,14 @@ static void update_attendee_status(icalcomponent *ical, strarray_t *onrecurids,
...
-    int age = config_getint(IMAPOPT_CALDAV_HISTORICAL_AGE);
+    int age = config_getduration(IMAPOPT_CALDAV_HISTORICAL_AGE, 'd');
     icaltimetype cutoff;
 
     if (age < 0) return icaltime_null_time();
 
     /* Set cutoff to current time -age days */
     cutoff = icaltime_current_time_with_zone(icaltimezone_get_utc_timezone());
-    icaltime_adjust(&cutoff, -age, 0, 0, 0);
+    icaltime_adjust(&cutoff, 0, 0, 0, -age);
```
rsto: libical is aware of the day of month, so adjusting with seconds here isn't them same as adjusting for days. I don't think it makes much of a difference for the use case, but you might want to check with @murch if we want to keep this option a days-bound integer value.

lib/imapoptions
```diff
@@ -1735,15 +1735,20 @@ If all partitions are over that limit, this feature is not used anymore.
 { "plaintextloginalert", NULL, STRING }
 /* Message to send to client after a successful plaintext login. */
 
-{ "popexpiretime", -1, INT }
-/* The number of days advertised as being the minimum a message may be
+{ "popexpiretime", "-1", DURATION }
```
rsto: Why isn't this '-1d' rather than '-1'?

imap/http_caldav_sched.c
ksmurchison: I don't think this is a problem, but we need to verify that icaltime_adjust() can properly handle a large number of seconds (e.g. 365 days)

ksmurchison: I just ran a quick test and icaltime_adjust() works fine with -365 days worth of seconds.

elliefm: Thanks for checking that, I meant to ask about it but then forgot!

lib/imapoptions
elliefm: It seems more natural, when using a negative value as a behaviour flag, to leave off the units? (As in: that's what I would do if I was setting up an imapd.conf for a server.)  And that being the case, I think it makes the docs a bit more clear that the default is NEVER, not a number of days, if the unit is left off the default in this case.

elliefm: Like, the imapd.conf man page ends up looking like this, which I think makes the most sense in this context:

>           popexpiretime: -1
>               The duration advertised as being the minimum a message may be left on the
>               POP  server  before  it  is deleted (via the CAPA command, defined in the
>               POP3 Extension Mechanism, which some clients may support).  This duration
>               [...]

imap/http_caldav_sched.c
elliefm: I'm really not sure whether to change this one back to simple int days.  All the other DURATIONs are things that, one way or another, were being converted down to simple seconds and then compared to time_t's.  But this bit is actually capable of doing real d/h/m/s calculations, so it seems a bit weird to squish it down to seconds the simple way and fudge it through like that.

ksmurchison: I tend to agree with your sentiment, but I'm fine with it either way.

PR #2890, opened 2019-10-06
libcyrus(_min).so: move *_method_desc to read-only segments

dilyanpalauzov:
So that the users of the libraries do not get their own copy for the lock_method_desc, map_method_desc and nonblock_method_desc symbols.

### DIFF
--- a/lib/cyr_lock.h
+++ b/lib/cyr_lock.h
@@ -53,7 +53,7 @@
 
 #include <sys/stat.h>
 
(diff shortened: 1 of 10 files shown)


elliefm:
I agree with this in principle, and I think I can see how it works, but how can I test that it has the desired effect in practice?

dilyanpalauzov:
If `nm -D libcyrus(_min).so` shows the symbol as R or r, then it is in the readonly section.

PR #2897, opened 2019-10-23
lib/times: consolidate month and weekday names into one variable

dilyanpalauzov:
and put that variable (wday/monthname) into read-only segment.


elliefm:
Looks good, thanks :)

PR #2900, opened 2019-10-27
imap/tls.c: add support for two certificates, e.g. RSA and EC, when Cyrus IMAP acts as server (master)

dilyanpalauzov:


elliefm:
Thanks!

Issue #2902, opened 2019-10-29
master: exit with error if ports cannot be bound


robn:
If master cannot bind a port during startup, it will log:

```
2019-10-28T18:48:28.930907-04:00 imap2 sloti2d3t17/master[3095803]: unable to bind to httpd/ipv4 socket: Address already in use
```

and then continue with that port unbound. That can give the appearance the Cyrus is "up", but certain services may not be available.

Better would be to just exit with error if any port can't be bound, so we can always be sure that either all services are available or none of them are.

elliefm:
It does this inside a loop over the `getaddrinfo()` results, and when it hits this condition it `continue`s with the next result, I guess supposing that one of the other results might be appropriate.

If, after everything, it hasn't got any proper socket for the service, it'll disable it (but in such a way that it'll try to start it up again on SIGHUP):

```
    if (nsocket <= 0) {
        syslog(LOG_ERR, "unable to create %s listener socket: %m", s->name);
        service_forget_exec(s);
        return;
    }
```

... though this sounds like it might be able to result in a situation where the service is successfully "up" on IPv6 but not on IPv4 (or v/v), when you expected both.  I was hoping this would be a simple patch, but this complicate things: if a service is up on one port, but not up on all the ports we expect it to be up on, I think this should also be "exit with error", yeah?  But, off the top of my head, I'm not sure how to accurately determine "all the ports we expect it to be up on" to make that check.

[The behaviour's the same for `open()` and `listen()` failures: log about them individually, then disable the service if we reach the end of the addrinfo loop with no good sockets.]

I'd probably wanna put the "exit on bind failure" behaviour behind an imapd.conf setting, cause it's quite a bit different from the historical default, not sure what to call it though?

robn:
Hmm, interesting. Can you bind a service to multiple ports explicitly (some list incantation like `listen="foo,bar"` or even `listen="foo" listen="bar"` or is it only one, but that might be `*:1234` for all interfaces?

Either way, I would argue that any bind failure means everything should fail, and if you specify `*` and the OS gives you something you didn't ask for, then you get to keep it. But then I think `*` was designed by someone having a laugh. Ambiguity should be met by "please be less ambiguous", not "I do what I want lol".

[...]

PR #2908, opened 2019-11-05
master: option to treat bind errors as fatal

elliefm:
This is the least-overthinky implementation of #2902 

Questions presently unanswered:

* should errors from the other socket-setup calls (socket(2), listen(2)) on this code path also be fatal with this option, or only bind(2)?
* does a bind on one getaddrinfo result ever fail when a bind on a subsequent result would have succeeded? (I don't think so)
* if the bind fails while setting up new services in an existing master process (i.e. after SIGHUP to provoke a cyrus.conf reread), should this situation be treated differently somehow?
* if there are already service processes running, should these be shut down?

### DIFF
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -1468,6 +1468,11 @@ Blank lines and lines beginning with ``#'' are ignored.
 /* Notifyd(8) method to use for "MAIL" notifications.  If not set, "MAIL"
    notifications are disabled. */
 
+{ "master_bind_errors_fatal", 0, SWITCH, "3.1.10" }
+/* If enabled, failure to bind a port is treated as a fatal error, causing
+   master to shut down immediately.  The default is to keep running, with
+   the affected service disabled until the next SIGHUP causes it to retry. */
+
 { "maxheaderlines", 1000, INT, "2.3.17" }
 /* Maximum number of lines of header that will be processed into cache
    records.  Default 1000.  If set to zero, it is unlimited.
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -1469,9 +1469,14 @@ Blank lines and lines beginning with ``#'' are ignored.
    notifications are disabled. */
 
 { "master_bind_errors_fatal", 0, SWITCH, "3.1.10" }
-/* If enabled, failure to bind a port is treated as a fatal error, causing
-   master to shut down immediately.  The default is to keep running, with
-   the affected service disabled until the next SIGHUP causes it to retry. */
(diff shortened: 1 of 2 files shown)


robn:
> should errors from the other socket-setup calls (socket(2), listen(2)) on this code path also be fatal with this option, or only bind(2)?

I'd say everything should be fatal. As the admin, I want either system up, or system not up. I'm also just handwave dreaming because I get to pretend that I don't know how software is made.

As for HUP, I don't know, mostly because I don't use it at all. I suppose, keeping in the same spirit, it might be nice to try and spin up the new state and if we can't, go back to the old state, but I know that's a total minefield (because I do actually know how software is made).

I wonder how many people actually use a master HUP to create/remove whole services. Like, how often are you doing that such that you can't afford an actual restart. I know not everyone has FM-like sharding/balancing stuff in front but still.

I suppose, if pressed, I might say just make the "die if can't bind" behaviour only work during first start, and after that its just best effort. I could argue that they're very different operational situations (ie transition from "off" to "on" is far more controlled and less dangerous than transition from "on" to "different on", and your monitoring should account for that) but like I say, I don't actually use this thing so I'm likely dead wrong.

elliefm:
I did some more reading and some more thinking, and I think socket() and listen() are mostly only gonna fail when something is pretty badly wrong (like, it's not just gonna be "this is the wrong getaddrinfo result, next one should work"), so those are now fatal too.

I've also made it so the fatal behaviour only happens when master is starting up; on a SIGHUP reload, if a socket/bind/listen fails, "you get to keep the pieces" (where did that phrase come from? I've seen it before somewhere and it's excellent)

But yeah -- I don't really think either of us is positioned to really say what the right behaviour is because we don't really use it in the way it expects to be used.  I think I might end up emailing the list for input on this one, see what people who use it in the world are doing and how their mental models work.

robn:
(https://english.stackexchange.com/a/250982 says its from a 1994 BSD manpage. I wondered if it has non-software history, but I couldn't find it quickly.)

That all sounds quite cromulent. Thanks. I have still not had chance to give it a try, but it's shaping it pretty close to what I think I want!

elliefm:
I've been wracking my brain on and off for weeks trying to work out how I want to frame a question to the mailing list about this, and I think I've concluded that I actually don't care what colour the bikeshed used to be?

The old apachectl-graceful-ish behaviour hasn't changed; and the new behaviour is small, sensible, and gated behind an off-by-default config option, so it won't hurt anyone.

So, I think it's probably fine to just merge this once _we're_ happy with it!

PR #2909, opened 2019-11-06
Make sure socket paths in example configuration match

J5lx:
Currently the example configuration files for cyrus.conf and imapd.conf use different socket paths, even though the imapd.conf examples themselves explicitly state that they should match. This PR adjusts the paths in cyrus.conf examples (/var/imap/socket) to match those in imapd.conf examples (/run/cyrus/socket), since the latter seem to be a bit more commonly referred to in the documentation.


elliefm:
Yeah nice, thanks :)

Issue #2629, opened 2019-01-15
Terrible breakage with -Wl,--as-needed in LDFLAGS


jasontibbitts:
Back in August, Fedora rawhide started using `-Wl,--as-needed` in LDFLAGS by default.  It turns out that this breaks Cyrus horribly.

Fedora runs Cassandane as part of the package build process, and with that flag enabled more than 50% of the tests fail.  Things randomly die with SIGILL, SIGABRT and SIGSEGV.  For me, the Cassandane itself just hangs at the end with a few tests pending and I have to kill the testrunner.pl stragglers.  We thought it might just be Cassandane so we disabled it for a bit, but the software itself is still terribly unstable.  One user provided some backtraces in https://bugzilla.redhat.com/show_bug.cgi?id=1659828 but they don't point to anything obvious (to me, at least).

I flipped an option to disable that flag and everything is back to normal, but I'd really like to understand what is going on since these failures may point to real issues somewhere.  It's tough to get good cores and backtraces out of our build process but another maintainer of our Cyrus package (@landgraf) thought it might have something to do with calling into Xapian.  That's been problematic in the past (see #2267) so... maybe.  But maybe you would see the same issue if you added `-Wl,--as-needed` to LDFLAGS and ran your normal tests.

elliefm:
_LDFLAGS_?

I've had `-Wl,--as-needed` in my _CFLAGS_ for ages, because things broke back when Ubuntu (11.10) made this the default, and I wanted to catch any further regressions.  I don't have it in my _CPPFLAGS_ though, so maybe something in Xapian (which is mostly C++) links badly and I don't see it because I don't give it that flag?

I'm not sure why I have this in my _CFLAGS_ specifically, I guess an Ubuntu user told me that was how to reproduce it at the time, and I thought no more of it.  I believe CFLAGS is included at both compile and link phases (for C), whereas LDFLAGS is only used at link phase.  But I don't remember enough about how linking works at a low level to really reason about whether there's a meaningful difference there.

Though... looking at those backtraces in the redhat tracker, both of them are in `glob_free()` -- and based on that alone I'd be willing to bet that this turns out to be Yet Another Conflict between libpcre's POSIX regex functions and glibc's POSIX regex functions (which provide the same symbols, but an incompatible implementation).  This wouldn't be the first time we've seen crashes out of this code when it has compiled against libpcre's headers but at runtime ends up in libc's functions (I don't think we've seen the inverse though)

[...]

PR #2910, opened 2019-11-06
Fix missing libcap in Perl module

J5lx:
Right now, the IMAP perl module is linked with libcyrus_min, which, if enabled, uses libcap, however the Perl module is _never_ linked to libcap. This leads to an error when loading the module:

```console
$ cyradm
Can't load '/usr/lib/perl5/5.30/vendor_perl/auto/Cyrus/IMAP/IMAP.so' for module Cyrus::IMAP: /usr/lib/perl5/5.30/vendor_perl/auto/Cyrus/IMAP/IMAP.so: undefined symbol: cap_set_proc at /usr/lib/perl5/5.30/core_perl/DynaLoader.pm line 193.
 at /usr/lib/perl5/5.30/vendor_perl/Cyrus/IMAP/Admin.pm line 43.
Compilation failed in require at /usr/lib/perl5/5.30/vendor_perl/Cyrus/IMAP/Admin.pm line 43.
BEGIN failed--compilation aborted at /usr/lib/perl5/5.30/vendor_perl/Cyrus/IMAP/Admin.pm line 43.
Compilation failed in require at /usr/lib/perl5/5.30/vendor_perl/Cyrus/IMAP/Shell.pm line 59.
BEGIN failed--compilation aborted at /usr/lib/perl5/5.30/vendor_perl/Cyrus/IMAP/Shell.pm line 59.
Compilation failed in require at /usr/bin/vendor_perl/cyradm line 66.
BEGIN failed--compilation aborted at /usr/bin/vendor_perl/cyradm line 66.
```

To fix this issue, this PR splits the libcap library out of LIBS (to avoid bloat), then adds it to the Perl module.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1525,7 +1525,7 @@ else
 lib_libcyrus_min_la_SOURCES += lib/map_nommap.c
 endif
 endif
-lib_libcyrus_min_la_LIBADD = $(LTLIBOBJS) $(LIB_UUID) $(GCOV_LIBS)
+lib_libcyrus_min_la_LIBADD = $(LTLIBOBJS) $(LIB_UUID) $(GCOV_LIBS) $(LIBCAP_LIBS)
 lib_libcyrus_min_la_CFLAGS = $(AM_CFLAGS) $(CFLAG_VISIBILITY)
 
 lib/chartable.c: lib/mkchartable.pl lib/charset/unifix.txt \
--- a/configure.ac
+++ b/configure.ac
@@ -1384,6 +1384,7 @@ AC_ARG_WITH([libcap],
         with_libcap="${withval}")
 
 have_libcap=no
+LIBCAP_LIBS=""
 case $host_os in
 linux*)
         if test "x$with_libcap" = "xyes"; then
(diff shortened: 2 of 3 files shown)


J5lx:
One thing I realised while making this patch: Cyrus has a number of libraries linked into binaries that don’t actually use them (see [here](https://gist.github.com/J5lx/3aee111f2ab1ff44c8329e8776a414ea)). Would you accept a patch to clean this up a bit like I’ve done for libcap in this PR?

elliefm:
> One thing I realised while making this patch: Cyrus has a number of libraries linked into binaries that don’t actually use them (see [here](https://gist.github.com/J5lx/3aee111f2ab1ff44c8329e8776a414ea)). Would you accept a patch to clean this up a bit like I’ve done for libcap in this PR?

I think this is because of how we handle dependencies hierarchically? Like, everything that needs anything in libcyrus_min.so gets linked against all of libcyrus_min.so's dependencies; anything that needs libcyrus_imap.so gets linked against all of _its_ dependencies, and so on.  Do you/your platform build with `-Wl,--as-needed`?

I'm not sure how I feel about it, and I'd be happy to review a patch to clean it up similarly. But, without having yet seen such a patch, I worry that individually maintaining every tool's specific list of dependencies might become its own headache?

J5lx:
Ah yes, that makes sense. The problem with -Wl,--as-needed is that I can’t use it because of #2629. I’ll think a bit more about my proposal.

PR #2923, opened 2019-11-25
treat cyr_expire and quota cmdline mboxprefix as admin namespace, not internal

### LINKED ISSUE #1699: cyr_expire only accepts internal mailbox naming

elliefm:
Affects `cyr_expire [...] -p mboxprefix` and `quota [...] mboxprefix` (no `-u`).

Mostly only noticeable when unixhierarchysep and/or virtdomains is in use?

Fixes #1699

### DIFF
--- a/imap/cyr_expire.c
+++ b/imap/cyr_expire.c
@@ -109,6 +109,8 @@ struct arguments {
     const char *altconfig;
     const char *mbox_prefix;
     const char *userid;
+
+    char *freeme; /* for mbox_prefix */
 };
 
 struct archive_rock {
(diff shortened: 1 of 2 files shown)


elliefm:
rebased on current master, fixed conflicts

imap/cyr_expire.c
```diff
@@ -210,6 +212,8 @@ static void cyr_expire_init(const char *progname, struct cyr_expire_ctx *ctx)
 
 static void cyr_expire_cleanup(struct cyr_expire_ctx *ctx)
 {
+    if (ctx->args.freeme) free(ctx->args.freeme);
+
```
brong, line 216: Should we be zeroing out more stuff here?

brong (approved):

imap/cyr_expire.c
elliefm: Just checked and yeah it's fine -- `ctx->args.freeme` is the only thing that was malloced here.  And we only call `cyr_expire_cleanup()` right before `exit()`, so we don't need to zero out the structures for later reuse or anything. :)

PR #2931, opened 2019-12-04
Make build reproducible

guimard:
Hi,

here is a request from [Reproducible Team](https://tests.reproducible-builds.org), please consider it:

> The differences in /usr/lib/x86_64-linux-gnu/libcyrus.so.0.0.0 are 100% ordering differences: each hexdump line (four 32-bit words) appears on both the left and right side of the diff, but the lines are in a different order.
> The dbgsym differences suggest that difference is caused by different ordering of chartables:
> (first build)
```
00000000000342a0 chartables_iso_8859_3
0000000000034aa0 chartables_iso_8859_2
00000000000352a0 chartables_iso_8859_16
0000000000035aa0 chartables_iso_8859_15
00000000000362a0 chartables_iso_8859_14
0000000000036aa0 chartables_iso_8859_13
00000000000372a0 chartables_iso_8859_11
0000000000037aa0 chartables_iso_8859_10
00000000000382a0 chartables_iso_8859_1
0000000000038aa0 chartables_iso_2022_kr
00000000000792a0 chartables_iso_2022_jp
```
> (second build)
```
00000000000342a0 chartables_iso_8859_3
0000000000034aa0 chartables_iso_8859_2
00000000000352a0 chartables_iso_8859_1
0000000000035aa0 chartables_iso_8859_16
00000000000362a0 chartables_iso_8859_15
0000000000036aa0 chartables_iso_8859_14
00000000000372a0 chartables_iso_8859_13
0000000000037aa0 chartables_iso_8859_11
00000000000382a0 chartables_iso_8859_10
0000000000038aa0 chartables_iso_2022_kr
00000000000792a0 chartables_iso_2022_jp
```
> The only difference between these two sets is the location of chartables_iso_8859_1.
> The static arrays are written by lib/mkchartable.pl in argv order [1], where argv is the shell glob «lib/charset/*.t» [2], which expands differently in the "C" v. "fr_CH.UTF-8" locales.
> Hence, sort()ing the arguments in the Perl script should fix this issue

### DIFF
--- a/lib/mkchartable.pl
+++ b/lib/mkchartable.pl
@@ -56,7 +56,6 @@ GetOptions( 'map|m=s'    => \@maps,      #strings
             'aliases|a=s' => \$aliasfile, #string
             'output|o=s' => \$output);    #string
 open (OUTPUT, ">$output");
-@maps = sort @maps;
 @ARGV = sort @ARGV;
 printheader(\@maps, \@ARGV);
 
--- a/lib/mkchartable.pl
+++ b/lib/mkchartable.pl
@@ -56,6 +56,8 @@ GetOptions( 'map|m=s'    => \@maps,      #strings
             'aliases|a=s' => \$aliasfile, #string
             'output|o=s' => \$output);    #string
 open (OUTPUT, ">$output");
+@maps = sort @maps;
+@ARGV = sort @ARGV;
 printheader(\@maps, \@ARGV);
 
 # first we parse the chartable unicode mappings and the fixes


elliefm:
I'm not sure why the CI build failed, but it failed while running mkchartable.pl, so it seems significant.

guimard:
@elliefm: thanks for considering this. Note that build doesn't fail but generates a diff that makes build not reproducible

elliefm:
Reproduced locally, oom killer got it:

```
Dec  4 17:12:33 debian kernel: [56824.127040] Out of memory: Kill process 27740 (perl) score 672 or sacrifice child
Dec  4 17:12:33 debian kernel: [56824.127044] Killed process 27740 (perl) total-vm:8344164kB, anon-rss:8306180kB, file-rss:68kB, shmem-rss:0kB
```

That's a _really_ interesting thing to shake out of a two-line change, I wonder what's going on.

guimard:
@elliefm : may be better with last commit (only sort \@ARGV, which is enough for reproducible build). Compilation OK here

guimard:
NB: patch tested with last 3.0 release

elliefm:
Yeah, that does seem to fix it, I guess the order of the mappings is significant.

Makes me wonder if it would have produced the right output eventually (if it hadn't run out of memory) or if it would've just spun forever...

guimard:
Order is probably not important since it is different between architectures _(and chosen by the shell)_

elliefm:
I'm going to merge this manually later today and squash it together with some comments and stuff, but I'll leave your authorship intact when I do.  Thanks :)

elliefm:
(I haven't been able to reproduce the original problem -- I spent some time stuffing with locales on my dev box but couldn't persuade it to glob differently.  But I've verified from perl documentation that the sort function's default order is stable, as long as `use locales` is *not* used, so this fix should definitely fix it even if I can't prove it myself.)

elliefm:
This is now on the 3.0 and master branches, thanks!

guimard:
Thanks !

PR #2935, opened 2019-12-05
imap/ Remove useless assignments (master)

dilyanpalauzov:


elliefm:
Sorry this took so long to look at.  For what it does, I don't think it's worth the trouble to evaluate and resolve all the conflicts that now exist

PR #2937, opened 2019-12-06
include file name (where known) in errors from message_parse functions

### LINKED ISSUE #2782: squatter: on reporting “more than 1000 header lines”, reference that message
When squatter prints “ERROR: message has more than 1000 header lines, not caching any more”, it shall say which message (mail) is meant.

elliefm:
If you've got opinions on the text of the error messages, sing out

I haven't yet tried to backport this for 3.0, but once it lands on master I'll look at that (and if it's easy and unintrusive to backport, I'll do so).

This closes #2782

### DIFF
--- a/cunit/message.testc
+++ b/cunit/message.testc
@@ -22,7 +22,7 @@ static void test_parse_trivial(void)
     struct body body;
 
     memset(&body, 0x45, sizeof(body));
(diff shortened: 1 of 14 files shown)


elliefm:
(I've called the new argument to these functions "efname" rather than "fname" to try to indicate that it's only for error reporting, not for any other purpose.)

PR #2938, opened 2019-12-06
Userlock

brong:
As mentioned during the week Ken - this is the "wrap all user renames with a lock" stuff.

In some cases I've gone direct to the lock, but in others where it's called often, I wait for a "doesn't exist" and then take a lock, check if it's been created in the race, then do the create.  This removes some race conditions, which is nice!

Hopefully this won't screw with your uuid-mailbox stuff too much!  I'm hoping to deploy it next week :)

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -6877,6 +6877,8 @@ static void cmd_create(char *tag, char *name, struct dlist *extargs, int localon
 
     mbname_t *mbname = mbname_from_extname(name, &imapd_namespace, imapd_userid);
 
+    struct mboxlock *namespacelock = user_namespacelock(mbname_userid(mbname));
+
     const char *type = NULL;
 
     dlist_getatom(extargs, "PARTITION", &partition);
@@ -7236,6 +7238,7 @@ localcreate:
     imapd_check(NULL, 0);
 
 done:
+    mboxname_release(&namespacelock);
     mailbox_close(&mailbox);
     buf_free(&specialuse);
     mbname_free(&mbname);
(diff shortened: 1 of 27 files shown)


ksmurchison:
This looks sane to me.  I assume all existing Cass test pass, yes?
I don't think (fingers crossed) this will cause a ton of problems with uuid-mailboxes.

brong:
well... there's a couple of memory leaks actually :p  I'll add the fixes for them to this branch before I merge it

imap/imapd.c
```diff
@@ -6875,7 +6875,9 @@ static void cmd_create(char *tag, char *name, struct dlist *extargs, int localon
         name[strlen(name)-1] = '\0';
     }
 
-    mbname_t *mbname = mbname_from_extname(name, &imapd_namespace, imapd_userid);
+    char *intname = mboxname_from_external(name, &imapd_namespace, imapd_userid);
+    mbname_t *mbname = mbname_from_intname(intname);
+    free(intname);
```
elliefm: wait, how can this help, when mboxname_from_external is just a wrapper around mbname_from_extname?

brong: hah, yeah.  So here's the deal:

extname: domain.com!user.foo parses to this:

```
username: NULL
domain: NULL
boxes: ["domain", "com!user", "foo"]
```

which converts back to the intname: "domain.com!user.foo".  If you round trip this a second time you get:

```
username: foo
domain: domain.com
boxes: []
```

BUT if you don't roundtrip through the intname, you wind up treating it as share namespace in the cmd_create code, but a user in the mboxlist_createmailbox code.

So this commit reverts to existing behaviour by taking the namespacelock with the second case as well, hence locking `foo@domain.com` rather than `""` (the shared namespace).

This is still somewhat bogus - we should just reject the bad extname, but I wanted to not change any existing behaviour with this commit.

elliefm: I see, so it relies on the behaviour that "make intname from an mbname with only boxes, no localpart or domain" is just concatenating the boxes back together with hiersep, but which accidentally produces a correct intname here, which can then be safely constructed into the mbname we actually want to use?

Yeah, that is still fairly bogus! But I agree, this PR isn't the place to fix it

PR #2939, opened 2019-12-06
imap/mailbox.c: remove unused mailbox_cid_rename()

dilyanpalauzov:


elliefm:
Looks like the ability to rename conversations was removed in fb5f7a48c, but I guess this function was never cleaned up.  Makes sense to remove it.

elliefm:
I tried using the "resolve conflicts" thing in the GitHub web UI to resolve the conflict, but it's created a backwards merge commit to do it, which is nasty.

I'll let the CI job finish, but then I'll land this manually rather than through the UI.

elliefm (approved):
Thanks

elliefm:
This was merged manually as e5b73a56d

PR #2940, opened 2019-12-07
imap/mailbox.c: abolish macro zeromailbox (master)

dilyanpalauzov:
Substitute the only occurrence of the macro with macro’s content.


elliefm:
Sorry about the long wait.  Nowadays the zeromailbox macro is used in more than one place, so it was useful after all

PR #2964, opened 2020-01-05
Removing dead links

marclaporte:
Not clear what to do with "Anonymous IMAP" label...


elliefm:
Does the anonymous imap access even work these days?  I think I remember hearing that it was dead, but I don't have a client handy to check with

elliefm:
Looks like the anon imap doesn't work anymore.  I'll merge this, then nuke the anon imap links myself. :)

PR #2969, opened 2020-01-10
Update documentation for user_deny.db

dilyanpalauzov:
Recommend cyr_deny instead of cyr_dbtool for managing user_deny.db.

Do not use the online documentation for cyrus imap 2.4.17 to reference the user_deny database format.


elliefm:
Thanks!

PR #2980, opened 2020-01-24
make libical 3.0.0+ and libicu mandatory

elliefm:
This bumps our minimum supported libical up to 3.0.0+ (removing a lot of feature-detection and workarounds going back to pre-1.0), and consolidates our icu usage and making it explicitly mandatory at configure time.

It also fixes some stuff that didn't build right with `--disable-jmap`, because that was on the circuitous path I needed for testing this.

This works great on my machine (tested with both the cyruslibs libical, and also a vanilla libical-3.0.0 from the release tarball), but we're all running different compilers and different library environments, so I'd appreciate if y'all could try this branch out in your own environment(s) before I merge it down, in case there's some platform-specific detail I've missed.

Thanks!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1791,7 +1791,10 @@ sieve_libcyrus_sieve_la_CFLAGS = $(AM_CFLAGS) $(CFLAG_VISIBILITY)
 sieve_sievec_LDADD = $(LD_SIEVE_ADD)
 sieve_sieved_LDADD = $(LD_SIEVE_ADD)
 
-sieve_test_SOURCES = sieve/test.c imap/mutex_fake.c imap/jmap_util.c imap/jmap_mail_query.c
+sieve_test_SOURCES = sieve/test.c imap/mutex_fake.c
+if JMAP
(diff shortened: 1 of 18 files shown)


elliefm:
(If you click "view command line instructions", below the "Merge pull request" button, it will show you how you can pull the branch to test it.  Just skip the "step 2: merge changes and update github" steps!)

ksmurchison:
@ellie This builds fine on my machine

dilyanpalauzov:
docsrc/imap/reference/faqs/install-linkerwarnings.rst still contains:
```
Libical v1.0.1 or v2.0
----------------------

You may see warnings regarding libical v2.0 being recommended to support certain
functionality. Currently libical v1.0.1 is sufficient, unless you need/want
RSCALE (non-gregorian recurrences), VPOLL (consensus scheduling), or
VAVAILABILITY (specifying availability over time) functionality. If v2 is
required, it will need to be installed from `github
<https://github.com/libical/libical>`_.


```

PR #2986, opened 2020-02-04
retry.c: rewrite retry_writev() to properly handle an initial failure

ksmurchison:
So, I think any issues that we may have seen with this function is if the initial writev() failed and returned -1.  We would then use that value in adjusting the iov which would result in an invalid iov_base pointer

### DIFF
--- a/lib/retry.c
+++ b/lib/retry.c
@@ -136,33 +136,9 @@ EXPORTED ssize_t retry_writev(int fd, const struct iovec *srciov, int iovcnt)
         len += srciov[i].iov_len;
     }
 
-    n = written = writev(fd, srciov, iovcnt > iov_max ? iov_max : iovcnt);
-
-    /* did we get lucky and write it all? */
-    if (written == len)
-        return written;
-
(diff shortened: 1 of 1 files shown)


lib/retry.c
```diff
@@ -176,6 +152,29 @@ EXPORTED ssize_t retry_writev(int fd, const struct iovec *srciov, int iovcnt)
...
+            iov++;
+            iovcnt--;
+            if (!iovcnt) fatal("ran out of iov", EX_SOFTWARE);
+        }
+
+        /* Skip whatever portion of the current iov that has been written */
+        iov->iov_base += n;
+        iov->iov_len -= n;
+
+        srciov = iov;
```
brong: I feel like we have too many different versions of iov here...

ksmurchison: How so?  iov is our malloc'd copy which we can edit.  srciov always points to what we are trying to output.
I have another version which keeps the initial write separate from the rest, and simply resets n to zero on failure

Issue #2975, opened 2020-01-17
IMAP: ESORT (RFC 5267)


kael:
An IMAP extension request.

The [ESORT extension](https://tools.ietf.org/html/rfc5267#section-3) would be quite useful as a complement to ESEARCH.

Since Cyrus supports SORT and ESEARCH, it wouldn't be too complicated to implement ESORT, would it ?

ksmurchison:
ESORT will appear in Cyrus 3.3+

kael:
Thanks a lot. :pray:  Cheers !

dilyanpalauzov:
Clarification: Cyrus supports the ESEARCH capability, this is imap response (RFC 4731) announced with the ESEARCH capability.

The ESEARCH command, as introduced by RFC 7377 (IMAP4 Multimailbox SEARCH Extension), “MULTISEARCH” capability string, is not supported.

PR #2987, opened 2020-02-05
Imap esort

ksmurchison:
To close issue #2975

### DIFF
--- a/imap/sequence.c
+++ b/imap/sequence.c
@@ -98,20 +98,19 @@ EXPORTED void seqset_add(struct seqset *seq, unsigned num, int ismember)
     if (ismember && num == seq->prev && seq->len && seq->set[seq->len-1].high == num)
         return;
 
-    if (num <= seq->prev)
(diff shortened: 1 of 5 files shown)


imap/sequence.c
```diff
@@ -98,20 +98,19 @@ EXPORTED void seqset_add(struct seqset *seq, unsigned num, int ismember)
...
+            seq->prev = num;
+            return;
+        }
 
-    if (!ismember) {
-        seq->prev = num;
-        return;
+        /* as if the previous number was given to us */
+        if (seq->flags & SEQ_SPARSE)
+            seq->prev = num - 1;
```
brong: I feel like we need to probably make seqset handle out of order sequences as a real thing rather than pulling magic tricks like this - but it's not like you've made it more evil than it already was!

ksmurchison: Yeah, the only evil that I added was to not fatal() if num < prev and to force a new set (RFC 5267 prefers 10,9,8 rather than 10:8).  Otherwise, the existing logic was left as-is.

PR #2991, opened 2020-02-15
Add compatibility with recent versions of libcap

### LINKED ISSUE #2974: 3.0.13 Build fails with libcap 2.29+
```
lib/util.c:582:12: error: conflicting types for 'cap_setuid'
  582 | static int cap_setuid(int uid, int is_master)
      |            ^~~~~~~~~~
In file included from lib/util.c:57:
/usr/include/sys/capability.h:138:12: note: previous declaration of 'cap_setuid' was here
  138 | extern int cap_setuid(uid_t uid);
      |            ^~~~~~~~~~
make[2]: *** [Makefile:5747: lib/lib_libcyrus_min_la-util.lo] Error 1
```

https://github.com/cyrusimap/cyrus-imapd/blob/95bfb40c041704158be48758d8ab3782caf1f980/lib/util.c#L582

J5lx:
Fixes #2974. I’ve tested it only on 3.0.13, but I see no reason why it wouldn’t also work on newer versions.

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -666,7 +666,7 @@ EXPORTED int set_caps(int stage __attribute__((unused)),
 }
 #endif
 
-static int cap_setuid(int uid, int is_master)
+static int cyrus_cap_setuid(int uid, int is_master)
 {
     int r;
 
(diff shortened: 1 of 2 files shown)


elliefm:
Great, thanks for that!  I've cherry-picked it to the 3.2 branch as well.

PR #3013, opened 2020-04-01
Synclock

brong:
I've noticed that if you have two sync_clients running concurrently (e.g. one by hand and one rolling) and they race, the second will get a bail out due to the remote changing between first and second query.

To work around that, this takes a named lock on the userid and channel any time it's replicating something for a user, and hence the second will wait and discover nothing to do.  This is more efficient and all around better :)


brong:
(Ken reviewed it separately and messaged me)

Issue #2973, opened 2020-01-16
master: start and stop daemons in predictable order


brong:
We added a "DAEMON" section a while back, which allows you to run things that need to run to support the services (e.g. idled, sync_server, squatter, notifyd).

We want to make sure we don't shut down these support service until we're done with them.  Further, they may have interdependencies, so it would be best to start each one in order and have a way to know that it's ready before starting the next - and likewise shut them down in reverse order once all the services are done.

elliefm:
Haven't started this yet but my branch will be here: https://github.com/cyrusimap/cyrus-imapd/compare/master...elliefm:v31/2973-master-daemon-start-stop-order

elliefm:
Sketching out a plan by starting with the docs.  How's this sound? https://github.com/cyrusimap/cyrus-imapd/commit/2ba40b074ae32e8ecba8d6f12cf610379b4b4ec5

elliefm:
The functionality aspect of this is almost all implemented now (I think)

Detouring to write some Cassandane tests to make sure it at least works, before I focus back in to polish up the corners.  Tests will be here (but there's nothing there yet): https://github.com/cyrusimap/cassandane/compare/master...elliefm:waitdaemons

elliefm:
I think we probably want to update `idled` to support being started as a waitdaemon, it would make some sense to be able to guarantee it's ready before `imapd` starts accepting connections

PR #3018, opened 2020-04-22
V31/2973 master daemon start stop order

elliefm:
(#2973)

This is almost done, so I'm putting it up for code review now

It adds a "wait" switch option (default: off) to entries in the DAEMON section of cyrus.conf, which toggles the new behaviour when on.  See the changes to docsrc/.../cyrus.conf.rst for how it works.

I have two remaining open questions:

1. If a wait=y daemon takes "too long" to report its readiness, should master complain and bail out?  At the moment, if a wait=y daemon hangs at startup, master will also hang at startup.  This makes sense, because "wait=y" means "refuse to start anything else until this thing is successfully started".  It might be useful for master to detect a hang and fatal out?  But I'm not currently convinced the added complexity would be worth it.

2. If a wait=y daemon successfully starts up at startup, and later exits for any reason (maybe it crashed, maybe Cyrus was SIGHUP'd by the operator), master will restart it automatically, like a normal daemon (i.e. without the ordering constraint, or the wait-for-readiness behaviour).  If it fails to restart for some reason, or restarts and crashes again soon after, etc, what should we do?  For normal DAEMON and SERVICE entries, if it fails more than MAX_READY_FAILS(5) times within MAX_READY_FAIL_INTERVAL(10) seconds, then it will disable that entry, and not try to start it again until the next SIGHUP.  But "wait=y" means "this needs to always be running as long as Cyrus is", and disabling it would invalidate that.  At the moment, master simply never gives up trying to restart it.  Should it fatal out instead?

Cassandane tests are here:
https://github.com/cyrusimap/cassandane/compare/master...elliefm:waitdaemons

`utils/waitdaemon.pl` in the Cassandane branch demonstrates how to make a perl script that conforms to the wait=y api

### DIFF
--- a/docsrc/imap/reference/manpages/configs/cyrus.conf.rst
+++ b/docsrc/imap/reference/manpages/configs/cyrus.conf.rst
@@ -257,6 +257,48 @@ is exiting.
     The command (with options) to spawn as a child process.  This
     string argument is required.
 
+.. parsed-literal::
+
+    **wait=**\ 0
+
+..
+
+    Switch: whether or not :cyrusman:`master(8)` should wait for this
+    daemon to successfully start before continuing to load.
+
+    If *wait=n* (the default), the daemon will be started asynchronously
+    along with the service processes.  The daemon process will not have
+    file descriptor 3 open, and does not need to indicate its readiness.
+
+    If *wait=y*, the daemon MUST write "ok\\r\\n" to file descriptor 3
+    to indicate its readiness; if it does not do this, and master has
+    been told to wait, master will continue to wait.... If it writes
+    anything else to this descriptor, or closes it before writing
+    "ok\\r\\n", master will exit with an error.
+
+    Daemons with *wait=y* will be started sequentially in the order
+    they are listed in cyrus.conf, waiting for each to report readiness
+    before the next is started.
+
+    Service processes, and *wait=n* daemons, are not started until after
+    the *wait=y* daemons are all started and ready.
+
+    At shutdown, *wait=y* daemons will be terminated sequentially in the
+    reverse order they were started, commencing after all other services
+    and *wait=n* daemons have finished.
+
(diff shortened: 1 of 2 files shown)


elliefm:
from chat with @brong:

1) current behaviour is good, not worth the extra complexity to make it smarter

2) just keep trying to restart it, and YELL LOUDLY

lib/retry.c
```diff
@@ -62,7 +63,9 @@ EXPORTED ssize_t retry_read(int fd, void *vbuf, size_t nbyte)
     char *buf = vbuf;
 
     for (nread = 0; nread < nbyte; ) {
+        syslog(LOG_DEBUG, "%s: top of loop, nread=%lu", __func__, (uint64_t) nread);
         ssize_t n = read(fd, buf + nread, nbyte - nread);
+        syslog(LOG_DEBUG, "%s: read returned %ld (%m)", __func__, (int64_t) n);
```
brong, line 68: For sure these should be removed before we ship this - or at least hidden behind an `#ifdef` so we don't have to call out to any additional syscalls during such a tight loop!  retry_read is used a lot.

elliefm:
Thanks!  I'll give this another look over myself, and add whatever I need to add to achieve the "YELL LOUDLY" we discussed a couple of weeks ago, fix up the history as the remaining "wips" are now resolved, and discard the "junk debug" commit since I already resolved whatever bug I was investigating with it :)

(Or in other words, for outside observers, please don't merge this yet: the core is good, but it still needs a final polish.)

elliefm:
The updates we discussed are now done, and as far as I can tell, this branch is looking pretty solid.

I think we should NOT merge this to master until it's been exercised at FM to work out any kinks that might arise.  It's only a few months until we start gearing up for 3.4 (!), and I don't want this in it unless we already know it's good.

Issue #3023, opened 2020-04-27
Deduplicate implicit keep against fileinto in Sieve


neilj:
If a sieve script does multiple `fileinto` calls that resolve to the same mailbox, Cyrus will deduplicate and only do the *last* one (the order is important, because the keywords may be different at the time of the different `fileinto`s). This is perfect. However, it does not do the same with the implicit keep mailbox. If you have `fileinto :copy INBOX` for example, and the implicit keep is not cancelled later on, Cyrus will attempt to deliver two copies to that mailbox. Delivery deduplication on the Message-Id, if turned on, will cause the latter copy to be discarded, but this is different behaviour to duplicate fileintos (because you get the keywords of the earlier copy). Instead, the implicit keep should be treated like a `fileinto` and replace any previous `fileinto` to the same mailbox (so you get the keywords at the time of the implicit keep, not the earlier fileinto).

ksmurchison:
Just so I'm clear, you always want the final keep/fileinto "INBOX" action to win, regardless of whether its implicit or explicit, correct?

neilj:
Correct.

ksmurchison:
I assume that we want to take into account fileinto :specialuse and :maillboxid as well?  What about fileinto :specialuse :create?

ksmurchison:
@brong  I'm thinking that the fileinto and keep callbacks each need to be split into a "what mailbox does the action resolve to and can we deliver to it piece" and a "go ahead and deliver to the mailbox piece".  By doing this, we will have the final mailbox name to use for comparison when creating the action list and we will know that 
fileinto "INBOX.foo"
fileinto :mailboxid "id-of-foo"
fileinto :specialuse "\\foo"
keep (user+foo)
all resolve to user.user.foo

similarly
fileinto "INBOX"
keep (user+nonexistent_mailbox)
both resolve to user.user

otherwise, I don't see how we can catch all instances of duplicate delivery before we do the deliver.db checks

ksmurchison:
Should be fixed by https://github.com/cyrusimap/cyrus-imapd/pull/3024

PR #3024, opened 2020-04-27
Sieve: the last keep / fileinto "INBOX" should always win

ksmurchison:
Possible solution to issue #3023

### DIFF
--- a/sieve/bc_eval.c
+++ b/sieve/bc_eval.c
@@ -1698,11 +1698,11 @@ int sieve_eval_bc(sieve_execute_t *exe, int is_incl, sieve_interp_t *i,
 
             if (i->edited_headers) i->getheadersection(m, &headers);
 
-            res = do_keep(actions, actionflags, headers);
+            res = do_keep(i, sc, actions, actionflags, headers);
             if (res == SIEVE_RUN_ERROR)
                 *errmsg = "Keep can not be used with Reject";
-
-            implicit_keep = 0;
+            else
+                implicit_keep = 0;
 
             actionflags = NULL;
             break;
@@ -1712,7 +1712,8 @@ int sieve_eval_bc(sieve_execute_t *exe, int is_incl, sieve_interp_t *i,
         case B_DISCARD:
             res = do_discard(actions);
 
-            implicit_keep = 0;
+            if (res == SIEVE_OK)
+                implicit_keep = 0;
             break;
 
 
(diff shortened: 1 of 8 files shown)


ksmurchison:
@brong  Care to review / sanity check this?

neilj:
I haven't studied it closely, but from a cursory glance of the code this looks like it will only work if the implicit keep mailbox is the inbox? But if you use plus addressing, the implicit keep mailbox could be initialised as something else, right?

sieve/bc_eval.c
```diff
@@ -1764,6 +1771,8 @@ int sieve_eval_bc(sieve_execute_t *exe, int is_incl, sieve_interp_t *i,
             if (res == SIEVE_RUN_ERROR)
                 *errmsg = "Fileinto can not be used with Reject";
 
+            implicit_keep &= cmd.u.f.copy;
```
brong, line 1774: Bitwise logic is a bit confusing here, I'd do:

if (!cmd.u.f.copy) implicit_keep = 0;

sieve/bc_eval.c
```diff
@@ -1870,6 +1881,8 @@ int sieve_eval_bc(sieve_execute_t *exe, int is_incl, sieve_interp_t *i,
             if (res == SIEVE_RUN_ERROR)
                 *errmsg = "Redirect can not be used with Reject";
 
+            implicit_keep &= cmd.u.r.copy;
+
```
brong, line 1885: as above

sieve/message.c
```diff
@@ -172,7 +172,8 @@ int do_fileinto(action_list_t *a, const char *mbox, const char *specialuse,
             return SIEVE_RUN_ERROR;
         }
 
-        if (a->a == ACTION_FILEINTO && !strcmp(a->u.fil.mailbox, mbox)) {
+        if ((a->a == ACTION_FILEINTO && !strcmp(a->u.fil.mailbox, mbox)) ||
+            ((a->a == ACTION_KEEP && !strcmp("INBOX", mbox)))) {
```
brong, line 176: As Neil commented - shouldn't we be implicit keeping into the potentially plus-addressed destination mailbox rather than "INBOX" specifically?

ksmurchison:
The problem is that given the current API, the guts of Sieve doesn't have the destination mailbox, only lmtpd does.  We would have to pass the destination mailbox into the Sieve engine in order to make this work.
If we want to make the first keep/fileinto be the winner across the board, it would be simpler to code.
https://tools.ietf.org/html/rfc5228#section-2.10.3 is silent on what is the correct solution, so we have some leeway

neilj:
No, it has to be the last sorry; I can't make it work with the first.

PR #3032, opened 2020-05-04
sync_log: bring back sync_log_mailbox_double for copy/move

elliefm:
886ee1537c1ef4527b573dd0917661e4260972a0 replaced all calls to `sync_log_mailbox_double()` with a new macro, `sync_log_rename()`, and added improved atomicity of replication of mailbox renames.

But, not all occurrences of `sync_log_mailbox_double()` were mailbox renames! Some were copies or moves of messages between folders -- which got confusing when reviewing sync_log, because it looked like lots of bad renames were happening when it was just people moving messages between folders.

This PR brings back `sync_log_mailbox_double()`, for use when two mailboxes need to be replicated together (but not due to a rename).

`sync_log_rename()` remains for actual renames

### DIFF
--- a/imap/sync_log.h
+++ b/imap/sync_log.h
@@ -78,7 +78,7 @@ void sync_log_reset();
     sync_log("UNMAILBOX %s\n", name)
 
 #define sync_log_mailbox_double(name1, name2) \
-    sync_log("MAILBOX %s %s\n", name1, name2)
+    sync_log("DOUBLEMAILBOX %s %s\n", name1, name2)
 
 #define sync_log_rename(name1, name2) \
     sync_log("RENAME %s %s\nMAILBOX %s\nMAILBOX %s\n", name1, name2, name1, name2)
@@ -114,7 +114,7 @@ void sync_log_reset();
     sync_log_channel(channel, "UNMAILBOX %s\n", name)
 
 #define sync_log_channel_mailbox_double(channel, name1, name2) \
-    sync_log_channel(channel, "MAILBOX %s %s\n", name1, name2)
+    sync_log_channel(channel, "DOUBLEMAILBOX %s %s\n", name1, name2)
 
 #define sync_log_channel_rename(channel, name1, name2) \
     sync_log_channel(channel, "RENAME %s %s\nMAILBOX %s\nMAILBOX %s\n", \
--- a/imap/sync_log.h
+++ b/imap/sync_log.h
@@ -77,6 +77,9 @@ void sync_log_reset();
 #define sync_log_unmailbox(name) \
     sync_log("UNMAILBOX %s\n", name)
 
+#define sync_log_mailbox_double(name1, name2) \
+    sync_log("MAILBOX %s %s\n", name1, name2)
+
 #define sync_log_rename(name1, name2) \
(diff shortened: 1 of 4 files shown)


imap/sync_log.h
```diff
@@ -73,6 +73,9 @@ void sync_log_channel(const char *channel, const char *fmt, ...);
 #define sync_log_unmailbox(name) \
     sync_log("UNMAILBOX %s\n", name)
 
+#define sync_log_mailbox_double(name1, name2) \
+    sync_log("MAILBOX %s %s\n", name1, name2)
```
brong, line 77: This is supposed to be "MAILBOX %s\nMAILBOX %s\n" - but since it's a single write under lock, they're guaranteed to be picked up by the same run.

I don't THINK the reader handling multiple mailboxes is safe like this because it will parse as a single name in the reader - whitespace isn't special!

elliefm: This is handled by the change in sync_client.c, also in this PR -- it explicitly looks for a second MAILBOX argument now.

It needs to be like this now, and not two distinct MAILBOX lines like it once was, since we started batching -- we need to know that in this instance, these two mailboxes are related, even if they don't belong to the same user, so that they can be batched and ordered correctly.

This originally happened at 1fa18fb128, but then when I later added the distinct RENAME command I mistakenly removed it (breaking copy/move).  The fix here is bringing back the updated mailbox double for copy/move, while keeping RENAME for renames.

The parser in `sync_log_reader_getitem()` already handles two arguments properly (and looks like it has done so forever).  It looks for a word, an astring, and then an optional second astring...  though this now has me wondering whether `sync_log()` itself _writes_ the correct thing?  I'm skimming quickly, but I don't see it doing any quoting -- looks like it just uses printf formatting?  So I wonder what happens if a mailbox name contains a space??

elliefm: Okay, I've checked out `sync_log()` and it's quoting properly after all, I just missed the salient detail while skimming.

So I think this is all fine?

elliefm:
@brong I've just rebased this on current master, cause it was conflicting...

Turns out the conflict was due to b388693482, which removed a seemingly redundant `sync_log_append()`.

If you look at my original commits in this PR, the first commit replaced that `sync_log_append()` with the `sync_log_mailbox_double()` that it originally was (before I accidentally broke it by long ago, and then iirc rsto "fixed" it by changing it to a sync_log_append).  The second commit realised both were needed together: `sync_log_mailbox_double()` (for replication) and a `sync_log_append()` (for squatter), and added comments to that effect.

This PR is pretty old, and I'm not sure whether the `sync_log_append()` call was needed back then, but is no longer needed now because append.c does it.  Or whether it's still needed, and b388693482 was mistaken?

Anyway, while rebasing, I've kept both calls with their new comments, which means adding that `sync_log_append()` call back, which might nowadays be redundant, I'm not sure.  What do you think?

It would be nice to get this fix approved and merged, so it stops getting in the way of progress! :)

elliefm:
Rebased to fix conflicts (a big chunk of sync_client.c had been moved to sync_support.c).

Also made it sync log mailbox doubles as DOUBLEMAILBOX rather than just MAILBOX, to make the sync logs a little easier to read, as suggested on slack.

PR #3045, opened 2020-05-15
Matchmime tempdir

rsto:
We came to realize that the in-memory Xapian database is way less efficient than the on-disk backends. Let's switch to using the default backend when evaluating a message against a JMAP Sieve rule. Cyrus installations should point their imapd.conf temp_path option to a tmpfs filesystem.

### DIFF
--- a/imap/jmap_mail_query.c
+++ b/imap/jmap_mail_query.c
@@ -53,6 +53,8 @@
 #include "json_support.h"
 #include "search_engines.h"
 
+#include "imap/imap_err.h"
+
 #ifndef JMAP_URN_MAIL
 #define JMAP_URN_MAIL                "urn:ietf:params:jmap:mail"
 #endif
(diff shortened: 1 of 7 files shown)


elliefm:
If we do it this way, we also need to update `lib/imapoptions` to document that `temp_path` ought to be on tmpfs, and we need to bump the last modified version field, so that administrators upgrading know they need to review their existing config.

Alternatively, what if we made a dedicated option for a path for this (rather than requiring temp_path to be on tmpfs), and fall back to the in-memory implementation if the option is unset?  Then hypothetically, when a future xapian no longer needs this workaround, we could simply deprecate the option, and temp_path need not be touched at all.

I think I saw in slack that lmtp uses `temp_path`.  Does putting `temp_path` on tmpfs introduce a risk of data loss in the event of a power outage during delivery?  I suppose, if the system loses power, the upstream sender won't get a "success" response, and will simply redeliver later, so maybe not.

rsto:
Thanks for the feedback @elliefm . From our Cyrus conf call I understood we'll go with temp_path. I've updated the imaptions entry accordingly and bumped the version. If you are OK with the updated temp_path option it I'll push this to master.

imap/jmap_mail_query.c
```diff
@@ -1025,6 +1035,8 @@ HIDDEN int jmap_email_matchmime(struct buf *mime,
     xapian_db_close(db);
 
 done:
+    if (dbpath) removedir(dbpath);
+    free(dbpath);
```
brong, line 1039: should we close dbw before we nuke the path out from under it?

lib/util.c
```diff
@@ -480,6 +481,38 @@ EXPORTED int create_tempfile(const char *path)
...
+                        const struct stat *sb __attribute__((unused)),
+                        int typeflag __attribute__((unused)),
+                        struct FTW *ftwbuf __attribute__((unused)))
+{
+    return remove(fpath);
+}
+
+EXPORTED int removedir(const char *dbpath)
+{
+    return nftw(dbpath, removedir_cb, 128, FTW_DEPTH|FTW_PHYS);
```
brong, line 513: this could probably be used in a few more places, in particular some of the other xapian stuff is shelling out to `rm` I think!

rsto:
Thanks @brong and @elliefm

PR #3046, opened 2020-05-16
imap/http_cgi.c: exactly declare `extern char **environment`

obache:
From the standard specifications, it must be declared by the user if
it is to be used directly.


elliefm:
Thanks :)

elliefm:
Also cherry-picked to cyrus-imapd-3.2 as f9177c4dce

PR #3050, opened 2020-05-17
Update docsrc/imap/rfc-support.rst

dilyanpalauzov:
I did `grep` for all mentioned RFCs in the source code and then updated docsrc/imap/rfc-support.rst as accurate as I could.

I reordered `imap/imapd.c:capa_struct` to match the order at https://www.iana.org/assignments/imap-capabilities/imap-capabilities.xhtml.  Then found out that [RFC 5524](https://tools.ietf.org/html/rfc5524) talks about `URLAUTH=BINARY` capability, but means the `URLFETCH=BINARY` capability and vise-versa. Then I opened a discussion at https://mailarchive.ietf.org/arch/browse/imapext/ .

RFCs from the IANA registy, that are not implemented, went to the wishlist section.  Likewise for recent RFCs that obsolete IMAP RFC, I guessed that the implementation in Cyrus IMAP is for the old RFCs.  I wrote a comment on this in imapd.c and put the new RFCs in the wishlist.

Summarizing the supported RFCs has been huge effort, so I hope that it will be integrated soon, as I will likely forget the details why did I something in a particular ways, and will not be able to answer questions, if there are any questions.

I do not know about JMAP, so these RFCs are not covered. 

In the comments, I switched to consistent use of space after “RFC ” and before the number.

RFC 7377 is nearly the same as RFC 6237.  Cyrus does not announce the MULTISEARCH capability.  So I have not mentioned RFC 6237.

http_err.et defines HTTP_DUPLICATE: “208 Already Reported”, HTTP_LOOP: “508 Loop Detected" from RFC 5842, HTTP_IM_USED "226 IM Used" from RFC 3229, HTTP_TOO_EARLY “425 Too Early” from RFC 8470, and HTTP_CENSORED "451 Unavailable For Legal
Reasons" from RFC 7725,  HTTP_VARIANT_NEGOTIATES “506 Variant Also Negotiates” says but these are unused.  I have not read the RFCs, so I cannot say if they are useful for Cyrus IMAP.
I put those RFCs in the wishlist.

The RFC 6901, 8288 and 8621 are mentioned in the code. but I cannot say if they are implemented, so I put them in the wishlist.

imap/jmap_mail.c:843 says that RFC 5322 is not fully mapped to JMAP.

RFC 5235, 5703 and 6558 are moved from the Wishlist to the “Ready” section.

I have not investigated the SMTP related RFCs like 8098, 2640, or 2852.


elliefm:
This looks good, thanks! :)

For what it's worth, "LITERAL-" is sent, but it's just sent automatically if "LITERAL+" is:

https://github.com/cyrusimap/cyrus-imapd/blob/81630a996ebface4cd2aee9dd575c60a81bf8e25/imap/imapd.c#L3503-L3505

I'm happy to merge this as is, and then update both the LITERAL- comments (yours, and the one in imapd.c to include the RFC number so it's searchable in the future) myself.

I see you've force-pushed it a few times since submitting it.  Do you have more you expect to do with it, or is it fine for me to go ahead and merge it as-is?

dilyanpalauzov:
I have no plans to change this further, unless something more comes to my mind.

PR #3054, opened 2020-05-25
http_ws: really do telemetry log each request separately!

brong:


ksmurchison:
Why does each JMAP request have to be logged separately?  Why can't this be treated like an IMAP connection with multiple commands?

brong:
@ksmurchison we log them all separately so we can easily replay a single one using the replayjmap tool.

rjbs:
@brong Are you looking to get this merged to master, or to keep it as a downstream customization?

brong:
Yes, merge to master.  IMAP is stateful, so you need to keep the connection together.  For websockets you don't, and each request is independent.

PR #3058, opened 2020-06-02
Jmap blob handling

ksmurchison:
Start of uniform handling of blobs.

Now need to move handling of "smart" blobIds into jmap_findblob() and out of Blob/get

### DIFF
--- a/imap/jmap_api.c
+++ b/imap/jmap_api.c
@@ -1248,6 +1248,15 @@ static int _jmap_findblob(jmap_req_t *req, const char *from_accountid,
 
     syslog(LOG_DEBUG, "jmap_findblob (%s, %s)", from_accountid, blobid);
 
+    if (blob && req->inmemory_blobs) {
+        const struct buf *inmem = hash_lookup(blobid, req->inmemory_blobs);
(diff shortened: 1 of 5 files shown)


imap/jmap_api.c
```diff
@@ -1228,6 +1235,15 @@ HIDDEN int jmap_findblob(jmap_req_t *req, const char *from_accountid,
...
 
+    if (blob && req->inmemory_blobs) {
+        const struct buf *inmem = hash_lookup(blobid, req->inmemory_blobs);
+        if (inmem) {
+            buf_init_ro(blob, buf_base(inmem), buf_len(inmem));
+            r = 0;
+            goto done;
+        }
+    }
+
```
elliefm, line 1246: Do I remember correctly from discussion the other day that the point of "findblob_exact" is for limiting the results to things that can be hardlinked?  If so, do we need to exclude in-memory blobs when exact is true?

brong: Yes, we do want to exclude if `exact`

ksmurchison:
I think this is ready to go to master

Issue #3056, opened 2020-06-01
Blob/get and Email/matchMime are non-standard methods


elliefm:
They're currently behaving as standard methods in 3.2.  They need to be guarded behind the jmap_nonstandard_extensions config and use the correct capabilities, and anything else I haven't thought of yet

* **Email/matchMime** -- I think af76e5d222 on master fixes this one completely.  Note that it's already part of the JMAP_MAIL_EXTENSION capability.  So I think it just needs to be cherry-picked to 3.2?

* **Blob/get** -- 1fc65566ba on master moves it from `jmap_core_methods_standard[]` to `jmap_core_methods_nonstandard[]`.  But its capability is still JMAP_URN_CORE, with an XXX comment to change it to JMAP_BLOB_EXTENSION.  Not yet sure how to do that properly.

Do we need to audit all our jmap_method_t arrays to find if there are any other extensions that snuck in as "standard"?

ksmurchison:
@elliefm I have already audited the method arrays and I think these are the only 2 methods that were mis-characterized.

The only reason that I didn't change the capability got Blob/get is that we're using that method in production and we need to get the middleware to include this capability in "using" first

elliefm:
@ksmurchison That's good intel, thanks.

To get this onto 3.2, it would be good to already have the capability tidy up on master.  To solve the problem of "FM still needs the current behaviour", I guess we could make a PR containing reverts of the new commit(s), and label the PR so that it's auto-included in FM builds.  @rjbs does this sound right?

How complicated is it to add the capability and hook it behind the jmap_nonstandard_extensions config?  I guess it's probably not difficult, but I simply haven't looked yet, so I have no insight

elliefm:
Ah, the plot thickens.  

On master, I think all that's needed is to replace JMAP_URN_CORE with JMAP_BLOB_EXTENSION like the comment says, because JMAP_BLOB_EXTENSION already exists for Blob/set.

3.2 doesn't have Blob/set, so it doesn't have JMAP_BLOB_EXTENSION set up yet.

The commit (f8f7d1b5f0c) that adds JMAP_BLOB_EXTENSION on master is also the commit that adds and hooks up all of Blob/set, so to get JMAP_BLOB_EXTENSION onto 3.2 will require gently teasing the two apart, not just a simple cherry-pick.

I think, in the meantime, I'll drop a version-hack into the Cassandane test to get it passing on 3.2, and then at least CI won't be broken in the meantime.

[...]

PR #3059, opened 2020-06-03
3.2: protect Blob/get and Email/matchMime behind jmap_nonstandard_extensions

elliefm:
This is the full 3.2 fix for #3056 

@ksmurchison can you please cast an eye across this?  I think it's complete, but I'd like a second opinion :)


ksmurchison:
This looks correct to me.

elliefm:
Thanks :)

PR #3061, opened 2020-06-03
httpd.c: zlib_compress() rewrite to handle lack of deflatePending()

ksmurchison:
This is a proposed fix for platforms that don't yet have a version of zlib that has deflatePending() (e.g. OpenBSD).
In my brief testing, none of the paths to zlib_compress() (length-delimited response, chunked response, WebSockets PMCE) triggered avail_out == 0, but I didn't want to remove the condition entirely.  I chose a fallback of pending=7 because its >6 per the comment, but we could bump this to 16/32/1024/etc.
I'd like to deploy this on a FM test server for a while to see if I broke anything.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1340,6 +1340,9 @@ if test "$with_zlib" != "no"; then
     AC_DEFINE(HAVE_ZLIB,[],[Do we have zlib?])
     ZLIB="-lz"
     HAVE_ZLIB=1
+
+    AC_CHECK_LIB(z, deflatePending,
+            AC_DEFINE(HAVE_DEFLATE_PENDING,[], [Do we have deflatePending()?]))
 else
     CPPFLAGS=$save_CPPFLAGS
     LDFLAGS=$save_LDFLAGS
--- a/imap/httpd.c
+++ b/imap/httpd.c
@@ -135,7 +135,7 @@ HIDDEN int zlib_compress(struct transaction_t *txn, unsigned flags,
                          const char *buf, unsigned len)
 {
     z_stream *zstrm = txn->zstrm;
-    unsigned flush;
+    unsigned flush, pending;
 
     if (flags & COMPRESS_START) deflateReset(zstrm);
 
@@ -155,19 +155,6 @@ HIDDEN int zlib_compress(struct transaction_t *txn, unsigned flags,
     do {
         int zr;
 
-        if (!zstrm->avail_out) {
-            unsigned pending;
-
-            zr = deflatePending(zstrm, &pending, Z_NULL);
-            if (zr != Z_OK) {
-                /* something went wrong */
-                syslog(LOG_ERR, "zlib deflate error: %d %s", zr, zstrm->msg);
-                return -1;
-            }
-
-            buf_ensure(&txn->zbuf, pending);
-        }
-
         zstrm->next_out = (Bytef *) txn->zbuf.s + txn->zbuf.len;
         zstrm->avail_out = txn->zbuf.alloc - txn->zbuf.len;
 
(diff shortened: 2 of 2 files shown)


anatoli26:
Ken, I didn't understand why you chose 7 for pending. It means the bytes count to increase the buffer for a second try to `deflate()`, right? Why to chose so small value if we can use say 1K without any considerable impact on RAM usage, but at the same time the chance that the buffer increase would be enough to finally complete the operation in a single (second) `deflate()` run would be near 100%?

On the other hand, I guess the condition when `deflateBound()` would not be enough to complete `deflate()` in one call is only when the data to compress in uncompressible, e.g. encrypted or already compressed (images, video) data, that's why you can't see it under normal use where all the data is structured text.

ksmurchison:
Yes, a larger value would make more sense, but how large does it have to be?  I chose 7 mainly to see if A) its ever used, and B) it has any impact on performance.  As I said above, in my testing I wasn't able to trigger a case where avail_out == 0 after a deflate() call, so the entire block of code could be useless.  This certainly needs more testing, both with and without deflatePending()

anatoli26:
`buf_ensure()` rounds the size of the requested buffer with `roundup()`, the minimum rounding with non-zero full buff is 64 and most probably it would be rounded to 1024, so 7 won't be used anyway.

> I wasn't able to trigger a case where avail_out == 0 after a deflate() call, so the entire block of code could be useless

I guess you won't be able to trigger it under normal circumstances. The data to be compressed should be uncompressible, otherwise the compression algo overhead would fit in the allocated buffer.

But IMO the checking/repeating block should NOT be removed, in the future things (and the data) may change and there could happen a situation when this effect will be triggered. Your current implementation is efficient, the checks just add a couple of asm instructions, IMO no performance impact.

elliefm:
I think, when you're not flushing, deflate will accumulate data internally until it feels like writing it, and there's sufficient output space for it.  As long as there's input still to process, and as long as its internal buffer has space, and as long as it hasn't been asked to flush, then the output buffer not being big enough is ignorable.  Like, I think if it can't output during this call, but it can process more input into its internal buffer, it just does that, and doesn't complain about the output buffer.  (The longer it goes without committing output, the better the final compression ratio.)

Which, I think means you'd only see `avail_out == 0` if a) it coincidentally decides to write (maybe its internal buffer is full), and there's _exactly_ enough space in the output buffer, or b) there is no further input available to process, and the available output buffer space insufficient to make progress (not necessarily zero), or c) you ask it you flush.

The case is definitely important to handle, but deliberately exercising it from a test would, I think, require knowing zlib's internal buffer size(s), giving it crafted inputs, and a carefully chosen output buffer size.  And then the test probably wouldn't work on some other platform where zlib was built with a different-sized internal buffer anyway!

No strong opinion on 7 vs any other number... though, given that buf_ensure() always (if increasing the buffer size) rounds upwards to at least 32, it'll never increase the buffer size by exactly 7 anyway.  But also, given that buf_ensure() always rounds to a sensible size, and given that we can't simply ask zlib how much space it actually wants in this case, then any random number >6 is fine here: buf_ensure() will do the best thing it possibly can under the circumstances.  So actually, I think 7 is technically correct, the best kind of correct.

elliefm:
While I'm in here anyway, this looks like a bug:

```
        if (!(zr == Z_OK || zr == Z_STREAM_END || zr == Z_BUF_ERROR)) {
            /* something went wrong */
            syslog(LOG_ERR, "zlib deflate error: %d %s", zr, zstrm->msg);
            return -1;
        }
```

By my understanding, Z_BUF_ERROR just means that there's not enough space in the output buffer to make output progress, and there was also no new input, so no progress can be made on either front.  You might legitimately see this on the final call to zlib_compress (no new input) if the output buffer is too small.  So Z_BUF_ERROR shouldn't `return -1`, it should just fall through to the same grow-the-output-buffer code and loop around again with more space.

ksmurchison:
Z_BUF_ERROR doesn't result in returning -1.  We only return -1 if the response code is NOT one of those 3

elliefm:
oh yeah, my bad!

anatoli26:
@ksmurchison, is there anything pending to merge this change? Is @brong's approval needed to proceed?

brong:
I approve - gonna merge this :)

brong:
@elliefm you might want to backport this one to 3.2 and 3.0 :)

elliefm:
@anatoli26 have you been able to test this version of the patch on OpenBSD?  I saw in the mailing list thread that you'd successfully tested the earlier version (without the configure.ac changes etc), but it'd be good to have the same confirmation of this version before I put it in a stable release :)

If you only have the released sources (i.e. not a git checkout) you can get this one as a patch here: https://github.com/cyrusimap/cyrus-imapd/commit/3248a4481ca5d26149f7e072677b6cea2f7ed272.patch

It should apply cleanly to 3.2 sources, cause it cherry-picks cleanly within git.

It doesn't pick cleanly to 3.0 (at a glance, looks like the libbrottli changes get in the way, and maybe others) but I'll worry about that once 3.2 is confirmed okay

anatoli26:
@elliefm, I can confirm that everything builds successfully with this change on OpenBSD-current with `--enable-http`.

I haven't tried to run it yet as I need to setup an environment for that, but I guess this changes, once it compiles successfully, should be ok at runtime. Anyway, I'll try to run it next week and if something is not working I'll let you know.

BTW, I guess it's not necessary to backport it to 3.0, as OpenBSD follows the latest stable sources and 3.2.2 is the current cyrus-imapd version in -current... and the 6.7-release won't get this change anyway (it has 3.0.13 at this moment, but it's compiled without http and this won't change).

PR #3063, opened 2020-06-09
Xapian great reindex

brong:


brong:
I've now tagged this "include-in-fastmail" since it's on all servers.  @rsto I'm good to merge it to master when you're ready.

PR #3064, opened 2020-06-10
Revert "jmap_core: move Blob/get to JMAP_BLOB_EXTENSION capability"

elliefm:
The master branch now has 5f5909037d, forward-ported from 3.2 as part of #3056.

This PR reverts it for FM, until our middleware is ready for it.


elliefm:
middleware has been updated, so this is no longer required

PR #3066, opened 2020-06-15
jmap_mail: guidsearch only for filter criteria that are indexed in Xapian

brong:
(=require no post-processing which seems to cause issues).


brong:
Will close this later once we're sure it's not needed.

PR #3076, opened 2020-06-24
lib/cyrusdb_twoskip.c: simplify

dilyanpalauzov:


lib/cyrusdb_twoskip.c
```diff
@@ -976,9 +973,7 @@ static int relocate(struct dbengine *db)
             loc->forwardloc[i] = _getloc(db, &loc->record, i);
 
         /* make sure this record is complete */
-        r = check_tailcrc(db, &loc->record);
-
-        if (r) return r;
+        return check_tailcrc(db, &loc->record);
     }
```
brong, line 977: Same complaint I had on the skiplist one... I don't like this style of cleanup generally, because it makes the final if() block behave specially with regards to returning early on non-failure cases, and this handling of 'r' will need to be re-added later if another block gets placed afterwards.

lib/cyrusdb_twoskip.c
```diff
@@ -1548,7 +1540,7 @@ static int myfetch(struct dbengine *db,
     if (foundkey) *foundkey = db->loc.keybuf.s;
     if (foundkeylen) *foundkeylen = db->loc.keybuf.len;
 
-    if (!r && db->loc.is_exactmatch) {
```
brong, line 1551: I mean, this is clearly correct, 'r' can't be non-zero here.  It's a pointless test.

lib/cyrusdb_twoskip.c
```diff
@@ -1510,7 +1502,7 @@ static int myfetch(struct dbengine *db,
             const char **data, size_t *datalen,
             struct txn **tidptr, int fetchnext)
 {
-    int r = 0;
```
brong, line 1513: Even though it's impossible to get to the end without initialising 'r', I'd still set it as a habit because it makes a class of mistakes less likely.

brong:
Having said that, this is fine and I will merge it - though I might change those bits back next time I touch them...

PR #3077, opened 2020-06-24
lib/cyrusdb_skiplist.c: simplify

dilyanpalauzov:


lib/cyrusdb_skiplist.c
```diff
@@ -1387,8 +1359,7 @@ static int mystore(struct dbengine *db,
 
     if (localtid) {
         /* commit the store, which releases the write lock */
-        r = mycommit(db, tid);
-        if (r) return r;
```
brong, line 1391: I like many of the simplifications in here, but I don't like this one because it's a flow break.  It changes the space after this and before the final 'return 0' from "runs if there is no error" to "never runs if there's a localtid", which makes it more complex to add additional code there.

Returning early on non-error cases makes it more complex to extend code and isn't an automatic good.

dilyanpalauzov:
I do not think that the code shall be kept bloated for the hypothetical case that there might be changes on the code one day and these changes would be easier.  More code → harder to read, slower to execute.

This pattern is also very common:
```C
int a = 0;
a = b
```
and is avoided by declaring the variables on the place of their first usage.

brong:
We have had multiple bugs due to that second approach.  Don't be smarter than the compiler - it can perfectly well exclude that first assignment if it's never read before another assignment, but it reduces the amount of context you need in order to make changes.

Ditto the "happy path reaches the end of function".

My argument for consistency of blocks is about it being easier to read.  Again - it's not going to change the size of the executable - the compiler can do the right thing for lots of these.

PR #3078, opened 2020-06-25
Sync intermediates

brong:
Hi Ken, bloody intermediates again.  This one goes along with the sync-intermediates branch on cassandane, which tests the case.

CREATE INBOX.folderA.folderB.folderC
RENAME INBOX.folderA INBOX.folderZ

And ensures that replication is clean and the uniqueid remains with it.


elliefm:
I've just fast-forwarded cassandane master up to the sync-intermediates branch, so we have that new test now :)

brong:
Thanks!  Awesome.

PR #3079, opened 2020-06-25
Readonlyimap

brong:
Here's another review :)  This one creates a "readonly" mode.  I haven't added it to nntpd, because... meh.

It also doesn't do CalDAV, CardDAV, etc.  We could probably do those just by banning meth_post, meth_delete, meth_put, etc.

Can you please especially review the cmdloop entries and see if I got all the things that try to write.  I'm tempted to add a syslog if it tries to take a writelock on any database!


brong:
I've applied the IMAP diff - thanks - that's a good point.

Regarding the HTTP one... METH_SAFE isn't going to allow JMAP post at all, which is a problem because JMAP has read-only stuff happen with POST.  It's going to need a little more care.

PR #3081, opened 2020-06-27
imap: URLAUTH=BINARY clarifications

dilyanpalauzov:
The IMAP Capabilities Registry was updated to state that the capability in RFC 5524 is not URLFETCH=BINARY but URLAUTH=BINARY.


elliefm:
Looks good, thanks

PR #3085, opened 2020-06-29
byteorder: rename from byteorder64 and optimise to use GCC builtins

brong:


lib/byteorder.c
```diff
@@ -96,16 +96,16 @@ EXPORTED unsigned long long _ntohll(unsigned long long x)
 
 #include <string.h>
 
-EXPORTED void *align_htonll(void *dst, unsigned long long src)
+EXPORTED void *align_htonll(void *dst, uint64_t src)
 {
-    unsigned long long tmp = htonll(src);
-    return memcpy(dst, (void *) &tmp, sizeof(unsigned long long));
+    uint64_t tmp = htonll(src);
+    return memcpy(dst, (void *) &tmp, sizeof(uint64_t));
```
elliefm, line 102: This (and similar cases elsewhere in this commit) are safe because it's a big find-and-replace that gets all occurrences, BUT, we've been bitten in the past by variables whose type has been changed without the corresponding `sizeof(type)` lines being updated to match.  

We can trivially prevent this problem in the future by using `sizeof(variable_name)` instead of `sizeof(type)` -- so while we're in this file anyway, seems like a good idea to make this fix now too?

brong: Yeah.  It does have issues where the variable is a pointer to something with bigger size, but it's a reasonable thing usually.

PR #3086, opened 2020-06-30
crc32: merge "unroll by 16" and "unroll by 8" logic into a single fun…

brong:
…ction for speed

The "unroll by 8" case is no slower, because we've just moved the "length > 64" test
into the for loop inside `crc32` rather than before hand.

The "unroll by 16" case is faster, because it does the "unroll by 8" if there's more
than 7 extra characters


brong:
From Rob's testing:

I created a harness that fills 100,000 * 108 bytes of memory (the length of each cyrus.index record that's crc32'd is 108 bytes, a largish mailbox is 100,000 records) with repeatable data. I then ran crc32 on each 108 byte block on all 100,000 records, and then repeated the whole thing 1000 times to use a measurable amount of time. I printed the crc32 at the end to ensure both versions gave the same result.

Original code:

```[fminabox root(robm)@fastmail1 ~]# time ./a.out
2608708667

real	0m14.310s
user	0m14.164s
sys	0m0.052s
```

Updated code:

```[fminabox root(robm)@fastmail1 ~]# time ./a.out
2608708667

real	0m5.709s
user	0m5.664s
sys	0m0.032s```

elliefm:
I was going to comment that our existing tests for this module don't exercise the large-buffer codepath, but I see you've added one that does. Nice 👍

PR #3098, opened 2020-07-07
jmap_calendar: ignore ACL when updating newly created calendar

rsto:
Recently introduced https://github.com/cyrusimap/cyrus-imapd/commit/9aeb3d9794c4be941c5359c514c3ff9fdd9af89d only sets the ACL for owner and admin on newly created users. This causes the following case the fail: user A shares their account with B, B creates a calendar in account A.

We probably don't even want to support that use case, but we currently have breaking Cassandane tests which assume we can!

### DIFF
--- a/imap/jmap_calendar.c
+++ b/imap/jmap_calendar.c
@@ -779,14 +779,15 @@ struct setcalendar_props {
  * NULL values and negative integers are ignored. Return 0 on success. */
 static int setcalendars_update(jmap_req_t *req,
                                const char *mboxname,
-                               struct setcalendar_props *props)
+                               struct setcalendar_props *props,
+                               int ignore_acl)
 {
     struct mailbox *mbox = NULL;
     annotate_state_t *astate = NULL;
     struct buf val = BUF_INITIALIZER;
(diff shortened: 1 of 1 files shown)


ksmurchison:
I think we want to fix the test, since Bron's ACL fix is designed to prevent exactly what the test is taking advantage of.

rsto:
So we outright refuse users creating calendars in other accounts, even if the calendar home ACL allows them do just that? I'm fine with either, but want to make sure we agree on that.

ksmurchison:
I think we need to review the JMAP/sharing/ACLs stuff and work through what makes sense, and also separate what the specs allow vs what FM wants to allow.

rsto:
Agreed. Let's wait until @brong chimes in.

brong:
Ooh... I think maybe we should allow the authenticated user to be copied into the ACL as well.  If you have an ACL which allows you to create the calendar then we should allow creating the calendar.

brong:
My commit was about not putting ACLs on the calendar which would later be removed by applying the actual share.With.

PR #3099, opened 2020-07-07
jmap_calendar: don't special-case deletion of Default calendar

rsto:
We hard-coded not to the Cyrus-created 'Default' calendar. But neither the CalDAV clients nor the JMAP clients or spec require the Default calendar to exist. So let's get rid of that.

### DIFF
--- a/imap/jmap_calendar.c
+++ b/imap/jmap_calendar.c
@@ -1388,29 +1388,8 @@ static int jmap_calendar_set(struct jmap_req *req)
             uid = newuid;
         }
 
-        /* Do not allow to remove the default calendar. */
-        char *mboxname = caldav_mboxname(req->accountid, NULL);
-        static const char *defaultcal_annot =
-            DAV_ANNOT_NS "<" XML_NS_CALDAV ">schedule-default-calendar";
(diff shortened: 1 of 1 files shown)


ksmurchison:
CalDAV Scheduling (RFC 6638) requires a default calendar for new invites, otherwise they will get summarily rejected.

I think we want to have a default calendar, just not one with a hard-coded name/id.  I think which calendar is the default should be settable, and probably flagged with a special-use flag or annotation.

rsto:
At Fastmail, we handle default calendars in CalendarPreferences (in the middleware). We could check for the {CALDAV}:schedule-default-calendar property in JMAP and reject deletion. But AFAIK we also don't check that for CalDAV currently.

ksmurchison:
We don't check that CALDAV prop because its hard-coded to always be #calendars/Default

rsto:
OK. So we need to change that in the DAV code. And at Fastmail, the middleware should also make sure to set this DAV property when changing the default calendar in the preferences.

rsto:
This was fixed a while ago with 9542837361a469a0271134dbe5b4b33d36ee883c and f0d2e3558356954eb0fc4a460dfbcc09083b2934

PR #3104, opened 2020-07-13
carddav: leniently accept whitespace in VCARD VERSION property

rsto:
As seen in the wild: a VERSION property ` 3.0` (note that leading whitespace). This patch silently accepts this bogus value. It does not rewrite the property value. We could make it so if `carddav_repair_vcard` is enabled, but this option is not enabled currently and we should first agree on if this is dead code, actually.

### DIFF
--- a/cunit/vparse.testc
+++ b/cunit/vparse.testc
@@ -38,6 +38,28 @@ static void test_wrap_onechar(void)
     // XXX test value
 }
 
+static void test_repair_version(void)
+{
+    char card[] = "BEGIN:VCARD\n"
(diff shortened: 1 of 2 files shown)


rsto:
It's tested here: https://github.com/cyrusimap/cassandane/commit/dc712b7cc6a2194d95cc1ecb02de7234d3b1cbd4

rsto:
We agreed in today's Cyrus call to rewrite the VERSION value, so I've updated this branch at https://github.com/cyrusimap/cyrus-imapd/commit/f824c1f20001333a7e6e04ff25ee5ae086b604a5

PR #3105, opened 2020-07-13
vparse: unconditionally repair control characters in VCARD

rsto:
We introduced VCARD reparation about 3 years as optional, but today decided to make Cyrus unconditionally repair bogus input in VCARD (as far as we can). This patch updates the VCARD parser to leniently skip control characters in VCARD input.

The unit tests have been updated to reflect this change. For a list of test cases see https://github.com/cyrusimap/cyrus-imapd/blob/d69a61c946f2d8c719061a12bbed5929c2d69cd6/cunit/vparse.testc#L63

### DIFF
--- a/cunit/vparse.testc
+++ b/cunit/vparse.testc
@@ -60,280 +60,258 @@ static void test_repair_version(void)
     vparse_free(&vparser);
 }
 
-static void test_control_chars(void)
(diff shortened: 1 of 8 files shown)


rsto:
@robmueller tagging you mainly to review the test cases. But of course feel free to review the rest of the code, too!

PR #3106, opened 2020-07-14
ctl_conversationsdb: show or fix thread mismatches

brong:
This will probably need some cleanup, but it's a fixer tool :)


brong:
I think this could still be useful, I'll rebase it and tidy it up.

brong:
I finally dug this out of the trash and fed it into Claude for a big tidyup.  It now makes '-R' fix everything '-A' notices, and it looks for all the mismatched basecid behavior that we didn't have good tests for before.

rjbs:
@brong This is marked both review-requested and draft.  Confirm status?

PR #3107, opened 2020-07-14
xapian: lowercase query text using Unicode-aware function

rsto:


brong:
This isn't relevant any more because we merged the xapian_great_reindex branch which already contained this fix :)

PR #3110, opened 2020-07-15
conversations: make subject normalisation regex not require PCRE

rsto:
We reworked conversation subject normalisation recently (4c643deb98e5cbcf47371eb1ee0ab74dbc4ba987) but the new regex did not work for builds with `--disable-pcre`. This patch rewrites the regex to be POSIX-compliant.

### DIFF
--- a/imap/conversations.c
+++ b/imap/conversations.c
@@ -709,9 +709,9 @@ EXPORTED void conversation_normalise_subject(struct buf *s)
     if (!initialised_res) {
         r = regcomp(&whitespace_re, "([ \t\r\n]+|\xC2\xA0)", REG_EXTENDED);
         assert(r == 0);
-        r = regcomp(&bracket_re, "(\\[[^\\[\\]]*\\])|(^[^\\[]*\\])|(\\[[^\\]]*$)", REG_EXTENDED);
+        r = regcomp(&bracket_re, "(\\[[^]\\[]*])|(^[^\\[]*])|(\\[[^]]*$)", REG_EXTENDED);
         assert(r == 0);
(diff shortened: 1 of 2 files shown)


imap/conversations.c
```diff
@@ -709,9 +709,9 @@ EXPORTED void conversation_normalise_subject(struct buf *s)
     if (!initialised_res) {
         r = regcomp(&whitespace_re, "([ \t\r\n]+|\xC2\xA0)", REG_EXTENDED);
         assert(r == 0);
-        r = regcomp(&bracket_re, "(\\[[^\\[\\]]*\\])|(^[^\\[]*\\])|(\\[[^\\]]*$)", REG_EXTENDED);
+        r = regcomp(&bracket_re, "(\\[[^]\\[]*])|(^[^\\[]*])|(\\[[^]]*$)", REG_EXTENDED);
         assert(r == 0);
-        r = regcomp(&relike_token_re, "^[ \\t]*[^\\s]+:", REG_EXTENDED);
+        r = regcomp(&relike_token_re, "^[ \\t]*[^\\t\\r\\n\\f]+:", REG_EXTENDED);
         assert(r == 0);
```
ksmurchison, line 715: Doesn't \s include SP as well, or don't we want to exclude SP here?

ksmurchison: Otherwise, this appears to solve the problem

rsto: Thanks. I pushed a fix in https://github.com/cyrusimap/cyrus-imapd/pull/3110/commits/2df28618c871d17eea3c7dcf37f74dcdc2f03f73

PR #3112, opened 2020-07-16
Remove cld2 dependency checks

### LINKED ISSUE #3102: Remove cld2 and SSE4.2 from configure.ac
@elliefm,

I see that the cld2 detection code is still in configure.ac. Shouldn't it be removed based on what we've discussed recently on cyrus-devel@ (i.e. cld2 is not used anywhere and was included by mistake)?

And related, recently there was a thread ("cyrus-imap buildin: sse extention") in info-cyrus@ about SSE4.2 compilation for CRC32c algo. I've commented that maybe the detection and informational code should also be removed from configure.ac as it is not used in any way at this moment and there are no plans to make it work, so to remove the confusion and simplify the code.

I can submit a PR if you prefer (without the merge mess this time : ).

anatoli26:
Closes #3102 (1/2)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1145,11 +1145,6 @@ imap_libcyrus_imap_la_SOURCES += \
 imap_libcyrus_imap_la_LIBADD += $(XAPIAN_LIBS)
 imap_libcyrus_imap_la_CXXFLAGS += $(XAPIAN_CXXFLAGS)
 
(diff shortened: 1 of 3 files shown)


elliefm:
Thanks! I've cherry-picked this to the 3.2 branch, so it'll be in the next release from that too

PR #3113, opened 2020-07-17
Remove SSE4.2 dependency checks and related code

### LINKED ISSUE #3102: Remove cld2 and SSE4.2 from configure.ac

anatoli26:
SSE4.2 was checked to use a hardware native x86 processors instruction
for CRC32c calculation. Nevertheless, its result was not compatible with
the existing CRC32 algorithm, so it was never actually used anywhere in
the project, but the code was still getting compiled introducing
distribution issues for those processors that had no such instruction.

Actually, there probably was no such issue in practice as there was a
check at runtime in crc32c_init() for availability of the instruction,
though it might have had some issues in some corner cases on some
platforms and it was confusing for the maintainers.

In the future if a need arises to revive this functionality, just undo
this commit.

Closes #3102 (2/2)


elliefm:
Thanks, this looks pretty tidy!

I've added it to the agenda for the cyrus-devel meeting next week, to figure out whether we'll remove it, ignore it, make it configurable, or something else

anatoli26:
@elliefm, could you please cherry-pick this commit (+ the one from Ken) to the 3.2 branch?

elliefm:
Oh whoops, thanks for the reminder.  I think I meant to do this, but since Bron did the merge rather than me, the next step slipped my mind.  Doh!

I've just cherry-picked both commits locally, and will push them upstream once the tests finish.

elliefm:
Done :)

PR #3114, opened 2020-07-17
Sieve script parse string

ksmurchison:
New function to parse a script from a script rather than from a file


imap/jmap_sieve.c
```diff
@@ -456,8 +456,37 @@ static int jmap_sieve_get(jmap_req_t *req)
...
+    }
+
+    /* open a new file for the script */
+    snprintf(new_path, sizeof(new_path),
+             "%s/%s%s.NEW", sievedir, name, SCRIPT_SUFFIX);
+
+    f = fopen(new_path, "w+");
+
+    if (f == NULL) {
+        syslog(LOG_ERR, "fopen(%s): %m", new_path);
```
elliefm, line 483: Does this log line want an IOERROR label on it?  I notice that we didn't previously have one in the old location, but I'm not sure if that was deliberate or oversight.

PR #3117, opened 2020-07-23
compiling.rst: add explanation for CLD2 dependency

anatoli26:
Closes 3102 (1 pending)

### DIFF
--- a/docsrc/imap/developer/compiling.rst
+++ b/docsrc/imap/developer/compiling.rst
@@ -178,6 +178,10 @@ Other
     installing SSL certificates."
     `ClamAV`_, libclamav-dev, clamav-devel, "no", "It is used by
     **cyr_virusscan**, otherwise not needed."
+    `CLD2`_, libcld2-dev, cld2-devel, "yes/no", "Compact Language Detector 2
+    (probabilistically detects over 80 languages in Unicode UTF-8 text, either
+    plain text or HTML/XML). Required for **Xapian** (``--enable-xapian``),
+    otherwise not needed."
     `openldap`_, libldap2-dev, openldap-devel, "no", "Development headers
     to enable **ptloader** to interface with LDAP directly, for canonification
     of login usernames to mailbox names, and verification of login usernames,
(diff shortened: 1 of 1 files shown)


docsrc/imap/developer/compiling.rst
```diff
@@ -178,6 +178,10 @@ Other
     installing SSL certificates."
     `ClamAV`_, libclamav-dev, clamav-devel, "no", "It is used by
     **cyr_virusscan**, otherwise not needed."
+    `CLD2`_, libcld2-dev, cld2-devel, "yes/no", "Compact Language Detector 2
+    (probabilistically detects over 80 languages in Unicode UTF-8 text, either
+    plain text or HTML/XML). Required for **Xapian** (``--enable-xapian``),
```
rsto, line 183: It is required for Xapian if the search_index_language option in imapd.conf is set. Thanks!

anatoli26: I just added the clarification about `search_index_language`.

elliefm: @rsto What happens if `search_index_language` is enabled on an installation that was compiled without CLD2?

anatoli26: I guess we get the error @brong mentioned in the related issue: "IOERROR: Xapian: language indexing requires CLD2 library"

anatoli26:
@elliefm, any feedback on this PR?

elliefm:
I think I'd prefer to leave the detail about `search_index_language` out, and leave it so that we say it's needed for any Xapian (regardless of that setting).

My general preference is that changing a setting in imapd.conf shouldn't suddenly require a recompile with new dependencies -- you either have the feature built in and can enable/disable/customise it from imapd.conf, or you don't have the feature built in, and any settings for it are ignored.

I think you could probably reset your branch to just the first commit (discarding the ", p2" one), force push, and we'd be in a good place.

anatoli26:
@elliefm, I just force-pushed the undo of the last commit as you suggested.

Now, what do you think about adding a check for cld2 as a required dependency in configure when Xapian is enabled?

elliefm:
I think I previously didn't want to do that, but now that I understand that this dependency's presence determines whether an imapd.conf setting works or not, I've changed my mind and I think that's the right way to go.  I'll add a similar note to the main issue.

PR #3119, opened 2020-07-23
Refactor guidsearch to postprocess non-Xapian filter criteria

rsto:
This PR fixes a bug in guidsearch using twoskip as a backend for conversations.db. If one of the xxxInThreadKeywords filter criteria was used, then the original code loaded conversation records from within the callback of guid_foreach. This worked fine for skiplist, but messed up the database-internal iterator state for twoskip.

The patch refactors guidsearch to first obtain all results from Xapian by use of the guid_foreach callback. Only after that, it reduces the result set by evaluating non-Xapian filter expression. This fixes the issue of interleaved conversation.db calls. However, where the original code could reuse one instance of a guidsearch_msg, the patched code now has to maintain a guidsearch_msg for each G record returned by Xapian.

To not get penalized for probably thousands of malloc() calls, the code:

- Allocates at once enough memory for all guidsearch_msg entries
- Applies all filtering and collapsing in-place
- Refactors bitvector_t to not heap-allocate memory if the highest bit value is less than 64. For inMailbox and inMailboxOtherThan filter criteria, this allows to not allocate memory if the inspected account does not have more than 64 mailboxes. An initial analysis showed this to be a reasonable amount, but we can still change that.

Furtehr, the guidsearch code was reorganized to help with future maintenance.

### DIFF
--- a/cunit/bitvector.testc
+++ b/cunit/bitvector.testc
@@ -29,46 +29,46 @@ static void test_basic(void)
     /* can set bit0 and get it back */
     bv_set(&bv, 0);
     CU_ASSERT_EQUAL(1, bv.length);
-    CU_ASSERT_NOT_EQUAL(0, bv.alloc);
+    CU_ASSERT_EQUAL(0, bv.alloc);
     CU_ASSERT_EQUAL(1, bv_isset(&bv, 0));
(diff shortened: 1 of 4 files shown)


brong:
This looks great.  I guess we'll deploy it on Monday :)

Bron.

PR #3127, opened 2020-07-27
xapian: don't index stemmed terms larger than max term size

rsto:
We limit the indexed term size to be at most 200 bytes (in UTF-8), but it turns out that this check only is applied by Xapian to the unstemmed term. For a couple codepoints the stem in Cyrus search form can be substantially longer, and we return an empty string as a stem in this case to make Xapian ignore that stem.

There's a test for that at https://github.com/cyrusimap/cassandane/tree/xapian_unicode_fdbx


brong:
Yep, that looks great :+1:

PR #3135, opened 2020-08-03
index: try three time to POST to the extractor

brong:
As promised Robert, the indexer "retry 3 times" logic.  I'll put some comments in to describe why I did things.

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -5472,7 +5472,7 @@ static int extract_attachment(const char *type, const char *subtype,
             sleep(retry);
             extractor_disconnect(ext);
             r = extractor_connect(ext);
-            if (r) goto done;
+            if (r) continue;
             be = ext->be;
         }
 
--- a/imap/index.c
+++ b/imap/index.c
@@ -5598,7 +5598,8 @@ EXPORTED void index_text_extractor_destroy(void)
 
     extractor_disconnect(ext);
     free(ext->be);
-    ext->be = NULL;
+    free(ext->hostname);
+    free(ext->path);
     free(ext);
 
     index_text_extractor = NULL;
--- a/imap/index.c
+++ b/imap/index.c
@@ -5026,9 +5026,10 @@ EXPORTED int index_search_evaluate(struct index_state *state,
 }
 
 struct extractor_ctx {
-    struct backend *be;
     struct protstream *clientin;
-    unsigned failed;
+    char *hostname;
+    char *path;
+    struct backend *be;
 };
 
 struct getsearchtext_rock
@@ -5274,11 +5275,6 @@ done:
 
 #define IDLE_TIMEOUT (5 * 60)  /* 5 min */
 
-struct backend_ctx {
-    char *path;
-    int timeout;
-};
-
 static int login(struct backend *s __attribute__((unused)),
                  const char *userid __attribute__((unused)),
                  sasl_callback_t *cb __attribute__((unused)),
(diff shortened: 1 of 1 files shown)


imap/index.c
```diff
@@ -5026,9 +5026,10 @@ EXPORTED int index_search_evaluate(struct index_state *state,
 }
 
 struct extractor_ctx {
-    struct backend *be;
     struct protstream *clientin;
-    unsigned failed;
+    char *hostname;
+    char *path;
+    struct backend *be;
```
brong, line 5032: I've moved these into the context object, of which there's only one, and they get initialised at creation time

imap/index.c
```diff
@@ -5274,11 +5275,6 @@ static int extract_vcardbuf(struct buf *raw, charset_t charset, int encoding,
 
 #define IDLE_TIMEOUT (5 * 60)  /* 5 min */
 
-struct backend_ctx {
-    char *path;
-    int timeout;
-};
-
```
brong, line 5281: and then we don't need a backend context at all, because we already have the path in the extractor context - and I looked at the timeout logic and it seemed un-necessary.

imap/index.c
```diff
@@ -5288,60 +5284,20 @@ static int login(struct backend *s __attribute__((unused)),
...
-    }
-    else {
-        syslog(LOG_DEBUG, "extractor_ping: HEAD %s: got status %u",
-               ctx->path, statuscode);
-    }
-
-    if (resp_hdrs) spool_free_hdrcache(resp_hdrs);
-
-    return r;
+    return 0;
```
brong, line 5290: I decided the ping was pointless - it doesn't tell us anything about the actual backend, so it's just a callout that always works.

imap/index.c
```diff
@@ -5288,60 +5284,20 @@ static int login(struct backend *s __attribute__((unused)),
...
 }
 
 static int logout(struct backend *s __attribute__((unused)))
 {
     return 0;
 }
 
-static void extractor_disconnect(struct backend *be)
+static void extractor_disconnect(struct extractor_ctx *ext)
```
brong, line 5298: API normalisation, this says "extractor" so it takes an extractor

imap/index.c
```diff
@@ -5363,161 +5319,96 @@ extractor_timeout(struct protstream *s __attribute__((unused)),
...
+    extractor_disconnect(ext);
 
     return NULL;
 }
 
 static struct protocol_t http =
 { "http", "HTTP", TYPE_SPEC, { .spec = { &login, &ping, &logout } } };
 
-static int extractor_connect(const char *exturl, struct extractor_ctx *ext)
+static int extractor_connect(struct extractor_ctx *ext)
```
brong, line 5335: we don't need exturl here because the hostname is stored in the extractor_ctx

imap/index.c
```diff
@@ -5363,161 +5319,96 @@ extractor_timeout(struct protstream *s __attribute__((unused)),
...
-    char path[256];
     time_t now = time(NULL);
-    int r = 0;
 
-    syslog(LOG_DEBUG, "extractor_connect(%s)", exturl);
-
-    if (ext->failed > 1) {
-        syslog(LOG_INFO, "extractor failed for url: %s", exturl);
-        return IMAP_INTERNAL;
-    }
```
brong, line 5394: We don't track failures any more - we'll try three times for every single attachment!

imap/index.c
```diff
@@ -5363,161 +5319,96 @@ extractor_timeout(struct protstream *s __attribute__((unused)),
...
-         * if the connection succeeds */
-        r = extractor_connect(exturl, index_text_extractor);
-        if (r) return r;
-        str->ext = index_text_extractor;
+    if (!index_text_extractor) {
+        /* This is a legitimate case for sieve and lmtpd (so we don't need
+         * to spam the logs! */
+        syslog(LOG_DEBUG, "%s: ignoring uninitialized extractor",
+                __func__);
+        return 0;
```
brong, line 5392: Since all the exturl checking is up-front, we just check if there's an extractor configured at all by checking for the global.

imap/index.c
```diff
@@ -5363,161 +5319,96 @@ extractor_timeout(struct protstream *s __attribute__((unused)),
...
-                type ? type : "<null>", subtype ? subtype : "<null>");
+               type ? type : "<null>", subtype ? subtype : "<null>");
         return 0;
     }
 
-    /* Fetch previously extracted text */
-    unsigned statuscode = 0;
+    struct extractor_ctx *ext = str->ext = index_text_extractor;
+
+    r = extractor_connect(ext);
```
brong, line 5403: This is still fine to call if we're already connected - it won't even ping any more, just check for a socket fd.

imap/index.c
```diff
@@ -5538,50 +5429,50 @@ static int extract_attachment(const char *type, const char *subtype,
         if (r) {
             syslog(LOG_ERR,
                    "extract_attachment: failed to read response for GET %s/%s",
-                   ctx->path, guidstr);
-            r = IMAP_IOERROR;
-            goto done;
+                   ext->path, guidstr);
+            statuscode = 599;
```
brong, line 5433: Rather than bailing on pretty much any error, we create a synthetic failure code!

imap/index.c
```diff
@@ -5538,50 +5429,50 @@ static int extract_attachment(const char *type, const char *subtype,
...
-            if (param->value) {
-                buf_putc(&buf, '=');
-                buf_appendcstr(&buf, param->value);
-            }
-            param = param->next;
+    int retry;
+    for (retry = 0; retry < 3; retry++) {
+        if (retry) {
+            // second and third time around, sleep and reconnect
+            sleep(retry);
```
brong, line 5471: This means 3 seconds total sleeping, 1 then 2.  We disconnect each time, which means that the following reconnect will be a real reconnect!

imap/index.c
```diff
@@ -5647,10 +5550,42 @@ static int extract_attachment(const char *type, const char *subtype,
...
+    else port = https ? 443 : 80;
+
+    /* Build servername, port, and options */
+    struct buf buf = BUF_INITIALIZER;
+    buf_printf(&buf, "%s:%u%s/noauth", server, port, https ? "/tls" : "");
+
     index_text_extractor = xzmalloc(sizeof(struct extractor_ctx));
     index_text_extractor->clientin = clientin;
+    index_text_extractor->path = xstrdup(path);
+    index_text_extractor->hostname = buf_release(&buf);
```
brong, line 5588: So all the parsing code got moved into here, and this is the only place exturl is referenced.  We just cache those two things.

Hmm, I realise I probably didn't create a cleanup for those - incoming commit in a sec.

imap/index.c
rsto: We could also not bail out of this loop early if extractor_connect failed, e.g. if the network was down for a second.

rsto:
I've updated the Cassandane test on a feature branch: https://github.com/cyrusimap/cassandane/tree/reindexer-retry

PR #3139, opened 2020-08-10
Sieve: update snooze to draft-ietf-extra-sieve-snooze-00

ksmurchison:
Whichever one of you feels most comfortable reviewing Sieve code


ksmurchison:
The failed Cass tests are fixed here: 
https://github.com/cyrusimap/cassandane/commit/112394a0f3eb066f623ec64b10320abd0345f4a9

Is there a way to have Travis use a Cass PR/branch when compiling a Cyrus PR/branch?

elliefm:
> Is there a way to have Travis use a Cass PR/branch when compiling a Cyrus PR/branch?

That's an interesting idea.  @ajaysusarla, any insight as to whether this seems plausible?

Though, if CI fails for a PR, and it only fails for tests that we know would need updates due to the PR, then I think it's reasonable to just say that.  It's also not hard for a reviewer to pull both branches and test it locally, if they wanna be careful.

So if we can automate this relatively trivially, it would be pretty nice to automate, but at this point I'm not sure it's worth putting a lot of effort into?

ajaysusarla:
I could give it a shot. For starters though, We would have to decide on a naming convention between the Cyrus branches and Cassandane branches, for that to work. Or some other way for us to figure out which Cassandane branch should be used if it is not the `master` branch on Cyrus.

lib/imapoptions
```diff
@@ -2375,7 +2375,7 @@ product version in the capabilities
...
+{ "sieve_extensions", "fileinto reject vacation vacation-seconds notify include envelope environment body relational regex subaddress copy date index imap4flags mailbox mboxmetadata servermetadata variables editheader extlists duplicate ihave fcc special-use redirect-dsn redirect-deliverby mailboxid vnd.cyrus.log vnd.cyrus.jmapquery snooze", BITFIELD("fileinto", "reject", "vacation", "vacation-seconds", "notify", "include", "envelope", "environment", "body", "relational", "regex", "subaddress", "copy", "date", "index", "imap4flags=imapflags", "mailbox", "mboxmetadata", "servermetadata", "variables", "editheader", "extlists", "duplicate", "ihave", "fcc", "special-use", "redirect-dsn", "redirect-deliverby", "mailboxid", "vnd.cyrus.log=x-cyrus-log", "vnd.cyrus.jmapquery=x-cyrus-jmapquery", "snooze=vnd.cyrus.snooze=x-cyrus-snooze"), "3.3.0" }
```
elliefm, line 2378: Can you please also bump the version at the end of this line to 3.3.1, since that's the "version" that will be the version that this change appears in.  Thanks

elliefm:
Looks good to me other than the (trivial) comment I left.  I thought it was going to let me "approve changes", but I think when commenting on the commit I lost that ability somehow.

PR #3144, opened 2020-08-15
Sync rightnow

brong:
This branch is now hopefully ready to run.  It does real time replication of many things, and still works with fallback as well.  It won't do anything until we deploy the config changes to turn it on.

### DIFF
--- a/imap/http_dav.c
+++ b/imap/http_dav.c
@@ -78,6 +78,7 @@
 #include "times.h"
 #include "syslog.h"
 #include "strhash.h"
+#include "sync_log.h"
 #include "tok.h"
 #include "user.h"
 #include "util.h"
@@ -4420,9 +4421,10 @@ static int dav_move_collection(struct transaction_t *txn,
         buf_free(&mrock.newname);
 
         if (mrock.root) {
+            mboxname_release(&namespacelock);
+            sync_log_checkpoint();
             xml_response(HTTP_MULTI_STATUS, txn, mrock.root->doc);
             xmlFreeDoc(mrock.root->doc);
-            mboxname_release(&namespacelock);
             return 0;
         }
     }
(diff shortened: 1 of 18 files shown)


imap/http_dav.c
```diff
@@ -6409,6 +6421,9 @@ EXPORTED int meth_propfind(struct transaction_t *txn, void *params)
     xml_partial_response(txn, fctx.root->doc, NULL /* end */, 0, &fctx.xmlbuf);
     xmlBufferFree(fctx.xmlbuf);
 
+    // might have made a change!
+    sync_checkpoint(txn->conn->pin);
+
```
ksmurchison, line 6426: @brong maybe its just too early for me, but what could PROPFIND actually change.  Its a safe (read-only) method

imap/http_dav.c
```diff
@@ -8244,6 +8265,9 @@ int meth_report(struct transaction_t *txn, void *params)
     /* Process the requested report */
     if (!ret) ret = (*report->proc)(txn, rparams, inroot, &fctx);
 
+    // might have made a change!
+    sync_checkpoint(txn->conn->pin);
+
```
ksmurchison, line 8270: @brong Same question as for PROPFIND

ksmurchison (approved):
Other than my 2 inline questions, this looks good to me.

imap/http_dav.c
brong: It can create mailboxes if they don't exist yet!

imap/http_dav.c
brong: same answer :P

brong:
@ksmurchison if no changes are made, sync_checkpoint is a noop anyway, because the log is empty :)

PR #3145, opened 2020-08-17
xapian: look for starting list-id angle bracket from end of string

rsto:
Some mail services use HTML tags in the phrase part of RFC2919
List-Id header values (and tend to forget the closing angle
bracket on the actual list-id). Treating the last '<' character
as starting angle bracket allows us to propertly extract the
list-id value from such headers.

This also works nicely with variants where clients cannot make
up their mind if their list-ids should be RFC2919 or Groups-style.

### DIFF
--- a/imap/xapian_wrap.cpp
+++ b/imap/xapian_wrap.cpp
@@ -930,14 +930,14 @@ static std::string parse_listid(const char *str)
     std::string val;
 
     /* Extract list-id */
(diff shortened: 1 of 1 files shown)


rsto:
Also updated the Cassandane tests at https://github.com/cyrusimap/cassandane/commit/8d5f259b14f710bbc2b1a3460df4272fe48142a6

PR #3147, opened 2020-08-18
Matchmime cache

brong:
Only parse the message once, no matter how many JMAP queries there are to run!

### DIFF
--- a/imap/jmap_mail_query.c
+++ b/imap/jmap_mail_query.c
@@ -104,7 +104,6 @@ static int _email_threadkeyword_is_valid(const char *keyword)
 #include "index.h"
 #include "search_query.h"
 #include "times.h"
-#include "xapian_wrap.h"
 
 HIDDEN void jmap_email_contactfilter_init(const char *accountid,
                                           const char *addressbookid,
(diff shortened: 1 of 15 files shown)


imap/jmap_mail_query.c
```diff
@@ -929,17 +928,100 @@ HIDDEN void jmap_email_filtercondition_validate(const char *field, json_t *arg,
     }
 }
 
-HIDDEN int jmap_email_matchmime(struct buf *mime,
+HIDDEN struct matchmime_t *jmap_email_matchmime_init(struct buf *mime, json_t **err)
+{
+    struct matchmime_t *matchmime = xzmalloc(sizeof(struct matchmime_t));
+    int r = 0;
+
+    matchmime->mime = mime;
```
rsto, line 936: I get why we want to pass in the mime buffer as function argument. But the transfer of buffer ownership is confusing and just asking for a double free corruption. I'd say we either

- do not free matchmime->mime in matchmime_free and keep memory management of the buffer to caller
or
- do not pass the buf* as argument to init but allow callers to set up the buffer after initialization

brong: The problem here is that we need the buffer for the queries, so we can't free it until we're finished.  The other option would be to keep the buffer separately in the message_data_t in lmtp_sieve, then we wouldn't need to do the lifetime management inside the struct.  I'll write that way up.

brong: Fixed in the new commits.

PR #3149, opened 2020-08-21
WIP syslog wrapper

elliefm:
This is a prototype of the syslog wrapper we've been talking about for standardising Cyrus's log output.

We've talked about this a few times recently, and each time the details of what we want have been slightly different.  I've thrown something together that broadly seems to do what we want, expecting that some of the specifics will be subject to further discussion and refinement.

I've called it `xsyslog()`, because I want to be able to naively search sources for "syslog" and find legacy and new kinds of log output, and because I want the casual reader to understand that this macro is roughly analogous to `syslog()`.

Some points for consideration:

* Automatic inclusion of the file and line number, but not the function name.  I'm reluctant to clutter it up with too much noise, and the full trifecta seems over the top.  Which are most useful?  (I would usually just include the function name myself, but someone else asked for file/line, so I went with that for now.)
* Automatic inclusion of the "%m" error string, if errno is non-zero.  A bunch of our existing syslogging only needs printf-style formatting to include this, so I just made it a freebie.  The implementation is careful to not alter errno.
* We've talked about "category" a bunch, and it's not clear to me whether that means the "IOERROR/DBERROR/SYNCERROR/etc" part, or the "easily matchable constant string description of the error with no embedded variables" part, or some combination of both.  I've guessed the latter, and called it "description", cause that's what made sense to me at the time (but on reflection I'd probably prefer to call it "message" instead...).  Do we want "IOERROR/etc" to be a separate argument, or to keep baking it in like we currently do?
* Some of the discussion maybe implies an expectation of automatically doing "foo=&lt;foo&gt;" formatting of arbitrary variables?  But C lacks the introspective capacity to do this any better than printf already does, so I'm assuming this part of the discussion was about code style rather than implementation detail, and that the distinction was lost in the meeting minutes.
* ~~The `struct buf buf` could potentially be declared `static`, which might save us some malloc/free churn on its internals.  But we don't generally log in hot loops, so maybe it doesn't matter?  I don't have a strong opinion either way.~~

There's currently two commits here:

* the first one adds the `xsyslog()` macro and the `xsyslog_fn()` function that implements it.
* the second one updates a small but hopefully representative set of old `syslog()` lines to use the new macro, to test/demonstrate how it's used and how it differs from what we had.

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -2078,7 +2078,7 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
         buf_putc(&buf, ' ');
     }
     if (saved_errno) {
-        buf_appendmap(&buf, "error=<", 7);
+        buf_appendmap(&buf, "syserror=<", 7);
         buf_appendcstr(&buf, strerror(saved_errno));
         buf_appendmap(&buf, "> ", 2);
     }
--- a/lib/util.c
+++ b/lib/util.c
@@ -2060,8 +2060,7 @@ EXPORTED void tcp_disable_nagle(int fd)
 }
 
 EXPORTED void xsyslog_fn(int priority, const char *description,
-                         const char *file_loc, int line_loc,
-                         const char *extra_fmt, ...)
+                         const char *func, const char *extra_fmt, ...)
 {
     static struct buf buf = BUF_INITIALIZER;
     int saved_errno = errno;
@@ -2081,7 +2080,7 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
     if (saved_errno) {
         buf_printf(&buf, "error=<%s> ", strerror(saved_errno));
     }
-    buf_printf(&buf, "srcfile=<%s> srcline=<%d>", file_loc, line_loc);
+    buf_printf(&buf, "func=<%s>", func);
 
     syslog(priority, "%s", buf_cstring(&buf));
     buf_free(&buf);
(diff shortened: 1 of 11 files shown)


lib/util.c
```diff
@@ -2058,3 +2058,32 @@ EXPORTED void tcp_disable_nagle(int fd)
...
+        va_start(args, extra_fmt);
+        buf_vprintf(&buf, extra_fmt, args);
+        va_end(args);
+
+        buf_putc(&buf, ' ');
+    }
+    if (saved_errno) {
+        buf_printf(&buf, "error=<%s> ", strerror(saved_errno));
+    }
+    buf_printf(&buf, "srcfile=<%s> srcline=<%d>", file_loc, line_loc);
```
rsto, line 2084: I ran a few callgrind tests and buf_printf for simple `<%s>` string substitutions performs worse by factor 4 than using a combo of buf_putc, buf_appendcstr. If we log lots of stuff in a loop that cost could show up. If we want to omit that cost, we'd need a buf helper function to also encode the line number, though.

rsto (commented):
Neat! Personally I'd also prefer function names rather than file names. But I can work with both.

elliefm:
@rsto what do you think of 9c0c3f38b6 specifically?  I'm not sure whether the JMAP method name should be part of the static description section, or whether it should be a `foo=<%s>` detail.

rsto:
@elliefm I think we don't need the JMAP method at all in the log message. The new xsyslog logger will provide us with the function name anyway. And we have the session id propagating through our logs which will allow us to identify the JMAP request.

elliefm:
Great, that'll neaten things up considerable, thanks.

PR #3150, opened 2020-08-21
Fix to declare missing MAXLOGNAME properly

obache:
Use LOGIN_NAME_MAX if it is available for the condition.

### DIFF
--- a/imap/fud.c
+++ b/imap/fud.c
@@ -60,6 +60,7 @@
 #include <netdb.h>
 #include <errno.h>
 #include <pwd.h>
+#include <limits.h>
 
 #include "acl.h"
 #include "mboxlist.h"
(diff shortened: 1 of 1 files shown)


imap/fud.c
```diff
@@ -97,7 +98,11 @@ static void send_reply(struct sockaddr *sfrom, socklen_t sfromsiz, int status,
 static int soc = 0; /* inetd (master) has handed us the port as stdin */
 
 #ifndef MAXLOGNAME
+#  ifdef LOGIN_NAME_MAX
+#define MAXLOGNAME (LOGIN_NAME_MAX - 1)
```
elliefm, line 102: I went looking for the significance of this being `LOGIN_NAME_MAX - 1` and not just `LOGIN_NAME_MAX`.  Apparently `LOGIN_NAME_MAX` includes the terminating nil character.  It's not immediately clear to me whether `MAXLOGNAME` does too, or not.  But it doesn't matter  anyway, because the only place this is used is here:

```
    char buf[MAXLOGNAME + MAXDOMNAME + MAX_MAILBOX_BUFFER];
```

which is `MAXLOGNAME + 20 + 1024`.  But a few lines later:

```
        r = recvfrom(soc, buf, 511, 0, sfrom, &sfromsiz);
```

We only ever use 511 bytes of this buffer!  Which is more than accommodated by the 1024 from `MAX_MAILBOX_BUFFER`, so unless `MAXLOGNAME` is negative its value is irrelevant.

Perhaps we should amend the recvfrom line to actually use the whole buffer, something like:

```
        r = recvfrom(soc, buf, sizeof(buf) - 1, 0, sfrom, &sfromsiz);
```

Do you want to do this too while you're in there?  I'm not familiar with the fud protocol, so I don't know whether this would be a useful improvement in practice, but it's better code.

elliefm (approved):
Sorry this took so long to look at.  It looks good, thanks.  I've identified one other change you might want to make while you're in there.

rjbs:

PR #3155, opened 2020-08-24
parseaddr: ignore control characters in quoted names

rsto:
This showed up while debugging a failing JMAP query sieve script: a message with a stray CR in the (non-ASCII) email address name part was indexed properly in Xapian. But when matching the same email address in a Sieve JMAP query the filter failed. This is because the message.c code handles lone CR and LFs, but our email address parser in Xapian did just pass the bogus header value on to parseaddr.

I suggest we make parseaddr lenient with regards to control characters: it preserves TAB, replaces lone CR and LR with space and ignores all other control characters.


rsto:
There is an issue with my local Cassandane setup and it seems unrelated to this patch. Still, I won't merge it until I can verify that all regression tests pass.

rsto:
Now passes all tests. There was a glitch in my Cassandane setup, unrelated to this patch.

PR #3156, opened 2020-08-24
xapian: do not log Xapian error context

rsto:
It is almost always empty for our use cases anyway.


brong (approved):
Yep, that all looks fine :)

PR #3158, opened 2020-08-25
Usercounters

brong:
As requested by Rob M - a way to get at usercounters via JMAP to avoid having to make an IMAP connection for session setup :)

### DIFF
--- a/imap/jmap_core.c
+++ b/imap/jmap_core.c
@@ -67,6 +67,7 @@ static int jmap_core_echo(jmap_req_t *req);
 static int jmap_blob_get(jmap_req_t *req);
 static int jmap_blob_set(jmap_req_t *req);
 static int jmap_quota_get(jmap_req_t *req);
+static int jmap_usercounters_get(jmap_req_t *req);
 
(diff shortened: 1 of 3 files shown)


imap/jmap_core.c
```diff
@@ -747,3 +759,248 @@ static int jmap_quota_get(jmap_req_t *req)
...
+
+    /* Does the client request specific responses? */
+    if (JNOTNULL(get.ids)) {
+        json_t *jval;
+        size_t i;
+
+        json_array_foreach(get.ids, i, jval) {
+            const char *id = json_string_value(jval);
+
+            if (!strcmp(id, "singleton"))
```
rsto, line 986: Wow, that's even more hackish than the "mail" quota id. I guess both quotas and usercounters should really go into an Account object. But let's just go with that singleton-id idiom for now.

rsto (approved):
Looks good to me. Great that you found that copy-paste error in mboxname counters. Is there a Cassandane test for the UserCounters object?

brong:
@rsto there's the start of a Cassandane test - I'll finish it up tomorrow, getting late now!

PR #3160, opened 2020-08-30
squatter: add search_index_skip_domains and search_index_skip_users

brong:
This is related to the 'pendingdelete' users that we create at Fastmail to rename users to when we delete them and support undo for a week.

Right now, we wind up reindexing the entire user when we rename it, even though we'll reindex it AGAIN when we rename back - this is wasteful, so let's just stop it.  I added the users option as well just because it parallels nicely.

this will be paired with a 'search_index_skip_domains: pendingdelete' imapd.conf setting.

### DIFF
--- a/imap/squatter.c
+++ b/imap/squatter.c
@@ -108,6 +108,9 @@ static int reindex_minlevel = 0;
 static const char *temp_root_dir = NULL;
 static search_text_receiver_t *rx = NULL;
 
+static strarray_t *skip_domains = NULL;
+static strarray_t *skip_users = NULL;
+
 static const char *name_starts_from = NULL;
 
 static void shut_down(int code) __attribute__((noreturn));
@@ -192,7 +195,13 @@ static void become_daemon(void)
 
 static int should_index(const char *name)
 {
+    // skip early users
+    if (strcmpsafe(name, name_starts_from) < 0)
+        return 0;
+
+    int ret = 1;
     mbentry_t *mbentry = NULL;
+    mbname_t *mbname = mbname_from_intname(name);
     /* Skip remote mailboxes */
     int r = mboxlist_lookup(name, &mbentry, NULL);
     if (r) {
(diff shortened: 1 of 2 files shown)


lib/imapoptions
```diff
@@ -2204,6 +2204,18 @@ If all partitions are over that limit, this feature is not used anymore.
   Deprecated. No longer used.
  */
 
+{ "search_index_skip_domains", NULL, STRING, "3.3.0" }
+/*
+ A space separated list of domains - if set, any users in the listed domains
+ will be skipped when indexing.
+ */
+
+{ "search_index_skip_users", NULL, STRING, "3.3.0" }
```
elliefm, line 2213: Can you please make these "3.3.1", as that will be the first "release" they'll eventually appear in?  Thanks

imap/squatter.c
```diff
@@ -206,29 +215,46 @@ static int should_index(const char *name)
...
     if (mbentry->mbtype & MBTYPE_COLLECTION) {
-        mboxlist_entry_free(&mbentry);
-        return 0;
+        ret = 0;
+        goto done;
+    }
+
+    // skip listed domains
+    if (mbname_domain(mbname) && skip_domains &&
+        strarray_find(skip_domains, mbname_domain(mbname), 0) > 0) {
```
rsto, line 242: 0 is a legit index, so this should read `>= 0`

imap/squatter.c
```diff
@@ -206,29 +215,46 @@ static int should_index(const char *name)
...
+        ret = 0;
+        goto done;
+    }
+
+    // skip listed users
+    if (mbname_userid(mbname) && skip_users &&
+        strarray_find(skip_users, mbname_userid(mbname), 0) > 0) {
```
rsto, line 249: 0 is a legit index, so this should read `>= 0`

rsto (changes requested):
Definitely a useful feature! The strarray_find function calls need an update. I added a Cassandane test to validate this review: https://github.com/cyrusimap/cassandane/commit/0794e6ad8045bb21cf5d2256f776249eae1d4ac0

lib/imapoptions
brong: Oh yeah - thanks.  I realised that as I did them and then forgot to come back to it.

imap/squatter.c
brong: yes, you're right.  Good point!

brong:
Fixes pushed for both the above :)

rsto (approved):
Looks good to me. Passed all SearchFuzzy tests on https://github.com/cyrusimap/cassandane/tree/squatter-skip

PR #3163, opened 2020-08-31
imap/caldav_db.c: change to use unitptr_t for hash variable

obache:
Hash variables willl be stored as void*, but It is unsure that
sizeof(unsigned long long) == sizeof(void*).


elliefm:
This seem correct to me, but I'm not real familiar with CalDAV specifically.  @ksmurchison can you please cast an eye over this and make sure I haven't missed some subtlety?

PR #3165, opened 2020-08-31
JMAP: handle role updates in Mailbox/set

rsto:
This patch makes mailbox role updates in Mailbox/set more sophisticated:

- Similar to changes to the mailbox tree, it predicts the final state of Mailbox roles after all operations in the Mailbox/set method were applied.
If the final state is valid, it suppresses Cyrus regular specialuse checks during mailbox renames and deletes, so that all operations can be applied.
If the final state is invalid, it rejects all operations that would alter a mailbox with specialuse, even if some of these operations would be valid.

IMAP specialuse rules and JMAP roles semantically differ and they impose different rules. This patch only accepts a final mailbox state to be valid if the restrictions of both IMAP and JMAP are satisfied. As a consequence a single role update might require the caller to first "clean up" the whole mailbox tree role assignments to qualify as a valid final mailbox state. This patch attempts to provide as much error description as possible to aid clients with this.

Cassandane tests have been updated at https://github.com/cyrusimap/cassandane/tree/mboxset_role

### DIFF
--- a/imap/annotate.c
+++ b/imap/annotate.c
@@ -3280,7 +3280,7 @@ static int annotation_set_pop3showafter(annotate_state_t *state,
 }
 
 EXPORTED int specialuse_validate(const char *mboxname, const char *userid,
-                                 const char *src, struct buf *dest)
+                                 const char *src, struct buf *dest, int allow_dups)
 {
     const char *specialuse_extra_opt = config_getstring(IMAPOPT_SPECIALUSE_EXTRA);
     char *strval = NULL;
@@ -3324,7 +3324,7 @@ EXPORTED int specialuse_validate(const char *mboxname, const char *userid,
     new_attribs = strarray_split(src, NULL, 0);
 
     for (i = 0; i < new_attribs->count; i++) {
-        int skip_mbcheck = 0;
+        int skip_mbcheck = allow_dups;
         const char *item = strarray_nth(new_attribs, i);
 
         for (j = 0; j < valid->count; j++) { /* can't use find here */
@@ -3390,7 +3390,7 @@ static int annotation_set_specialuse(annotate_state_t *state,
     }
 
     r = specialuse_validate(state->mailbox->name, state->userid,
-                            buf_cstring(&entry->priv), &res);
+                            buf_cstring(&entry->priv), &res, 0);
     if (r) goto done;
 
     r = write_entry(state->mailbox, state->uid, entry->name, state->userid,
--- a/imap/jmap_mailbox.c
+++ b/imap/jmap_mailbox.c
@@ -2012,9 +2012,7 @@ static void _mboxset_args_parse(json_t *jargs,
         } else {
             char *specialuse = _mbox_role_to_specialuse(role);
             struct buf buf = BUF_INITIALIZER;
-            // XXX: we should be passing the mailbox name to specialuse_validate to allow
-            // setting the current role back on a mailbox...
-            int r = specialuse_validate(NULL, req->userid, specialuse, &buf);
+            int r = specialuse_validate(NULL, req->userid, specialuse, &buf, 1);
             if (r) is_valid = 0;
             else args->specialuse = buf_release(&buf);
             free(specialuse);
@@ -2214,7 +2212,7 @@ done:
     return r;
 }
 
-enum mboxset_runmode { _MBOXSET_FAIL, _MBOXSET_SKIP, _MBOXSET_TMPNAME };
+enum mboxset_runmode { _MBOXSET_FAIL, _MBOXSET_SKIP, _MBOXSET_INTERIM };
 
 struct mboxset_result {
     json_t *err;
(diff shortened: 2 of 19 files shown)


rsto:
Tagging @ksmurchison as the mboxlist changes might impact the mailbox-by-id branches.

ksmurchison (approved):
I think the only conflicts I will have are with the annotation lookup calls, but I should be able to resolve these fairly easily, especially since you have Cass tests to verify correctness

imap/mboxlist.c
```diff
@@ -1789,9 +1791,7 @@ EXPORTED int mboxlist_deletemailbox_full(const char *name, int isadmin,
                                          const char *userid,
                                          const struct auth_state *auth_state,
                                          struct mboxevent *mboxevent,
-                                         int checkacl,
-                                         int local_only, int force,
-                                         int keep_intermediaries, int silent)
+                                         int flags, int silent)
```
brong, line 1795: We could probably add MBOXLIST_DELETE_SILENT to flags and then ditch the _full API and just pass that flag through from sync_support.

imap/mboxlist.h
```diff
@@ -217,40 +217,35 @@ int mboxlist_createsync(const char *name, int mbtype, const char *partition,
...
 /* same but with silent */
 int mboxlist_deletemailbox_full(const char *name, int isadmin, const char *userid,
                            const struct auth_state *auth_state,
-                           struct mboxevent *mboxevent,
-                           int checkacl,
-                           int local_only, int force,
-                           int keep_intermediaries, int silent);
+                           struct mboxevent *mboxevent, int flags, int silent);
 /* same but wrap with a namespacelock */
 int mboxlist_deletemailboxlock(const char *name, int isadmin, const char *userid,
```
brong, line 248: I thought about suggesting a _LOCK flag as well, but I think the separate function that handles lock setup and cleanup regardless of the error case is good - it saves adding complexity to the internal _deletemailbox function.

imap/annotate.c
```diff
@@ -3280,7 +3280,7 @@ static int annotation_set_pop3showafter(annotate_state_t *state,
 }
 
 EXPORTED int specialuse_validate(const char *mboxname, const char *userid,
-                                 const char *src, struct buf *dest)
+                                 const char *src, struct buf *dest, int allow_dups)
```
brong, line 3283: Oh man - what a pain.  I can see why we need it though.

imap/jmap_mailbox.c
```diff
@@ -2538,6 +2544,28 @@ static void _mbox_update(jmap_req_t *req, struct mboxset_args *args,
...
+        if (args->specialuse && *args->specialuse) {
+            result->skipped = 1;
+        }
+        if (!result->skipped) {
+            struct buf val = BUF_INITIALIZER;
+            annotatemore_lookup(mbentry->name, "/specialuse", req->accountid, &val);
+            if (buf_len(&val)) {
+                buf_free(&val);
+                result->skipped = 1;
+            }
```
brong, line 2564: I'd probably write this as:

```
    if (buf_len(&val))
        result->skipped = 1;
    buf_free(&val);
```

The assumption is that if the length is value, nothing was malloced, but that may not always be true!  There's probably heaps of other code that would need reviewing, but it's nice not to add this assumption in more places.

imap/jmap_mailbox.c
```diff
@@ -2931,6 +2961,17 @@ static void _mbox_destroy(jmap_req_t *req, const char *mboxid,
...
 
+    /* Skip role updates in first iteration */
+    if (mode == _MBOXSET_SKIP) {
+        struct buf val = BUF_INITIALIZER;
+        annotatemore_lookup(mbentry->name, "/specialuse", req->accountid, &val);
+        if (buf_len(&val)) {
+            buf_free(&val);
+            result->skipped = 1;
+            goto done;
+        }
```
brong, line 2972: as above, with of course `if (result->skipped) goto done;`

imap/jmap_mailbox.c
```diff
@@ -3395,82 +3441,343 @@ static void _mboxset_state_conflict_cb(const char *id, void *data, void *rock)
...
-            struct mboxset_entry *entry = hash_lookup(args->mbox_id, entry_by_id);
+            /* Update entry in in-memory mailbox tree */
+            struct mboxset_entry *entry = hash_lookup(args->mbox_id, state->entry_by_id);
             if (!entry) goto done;
             if (args->name) {
                 free(entry->name);
                 entry->name = xstrdup(args->name);
             }
             if (args->parent_id) {
-                free(entry->parent_id);
```
brong, line 3452: can this ever be called twice times?  If so we want `free(entry->old_parent_id)` before setting it.

brong (commented):
Yep, looks good!  I made some comments about buf freeing only on the "has length" path which I think is bad hygiene, but otherwise really good.

It's also a really complex algorithm doing all this, but I don't know how to make it simpler!

brong (approved):

imap/mboxlist.c
rsto: done

imap/jmap_mailbox.c
rsto: done

imap/jmap_mailbox.c
rsto: it should not, but it doesn't hurt to free old_parent_id

Issue #3154, opened 2020-08-23
httpd/3.2.2 segfaults when accessing /admin/proc


felixjogris:
Hi,

httpd of a non-murder 3.2.2 setup segfaults when accessing /admin/proc with a webbrowser to get a list of currently running services. Syslog on the server:

```
process type:SERVICE name:http path:/usr/local/cyrus/libexec/httpd age:0.586s pid:9828 signaled to death by signal 11 (Segmentation fault, core dumped)
```

lldb reveals:

```
$ lldb -c ./httpd.core /usr/local/cyrus/libexec/httpd                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
(lldb) target create "/usr/local/cyrus/libexec/httpd" --core "./httpd.core"
Core file '/root/httpd.core' (x86_64) was loaded.
(lldb) bt
* thread #1, name = 'httpd', stop reason = signal SIGSEGV
  * frame #0: 0x000000000022eb7a httpd`sort_procinfo(k=0x0000000802243700, pa=0x0000000802243708, pb=0x00007fffffff9ff7) at http_admin.c:505:9
    frame #1: 0x0000000800c5b193 libcyrus.so.0`qsort_r_compar_func(a=0x0000000802243700, b=0x0000000802243708) at cyr_qsort_r.c:24:12
    frame #2: 0x0000000801731a70 libc.so.7`qsort + 4256
    frame #3: 0x0000000800c5b13a libcyrus.so.0`qsort_r(base=0x0000000802243700, nmemb=5, size=8, thunk=0x00007fffffff9ff7, compar=(httpd`sort_procinfo at http_admin.c:494)) at cyr_qsort_r.c:33:5
    frame #4: 0x000000000022c16a httpd`action_proc(txn=0x00007fffffffa718) at http_admin.c:615:5
    frame #5: 0x000000000022b8c6 httpd`meth_get(txn=0x00007fffffffa718, params=0x0000000000000000) at http_admin.c:212:17
    frame #6: 0x000000000029a763 httpd`process_request(txn=0x00007fffffffa718) at httpd.c:1799:9
    frame #7: 0x00000000002a3220 httpd`http1_input(txn=0x00007fffffffa718) at httpd.c:1871:29
    frame #8: 0x000000000029826f httpd`cmdloop(conn=0x00007fffffffd9a8) at httpd.c:2015:13
[...]

PR #3129, opened 2020-07-28
OpenBSD: upstreaming patches

anatoli26:

PR #3166, opened 2020-09-01
Freebsd admin proc

felixjogris:
On FreeBSD, /admin/proc displays no values for state, start, and vmsize:

Bildschirmfoto_2020-08-31_23-47-32

The /proc filesystem is not mounted on a default system, and if so, there's no /proc/stat, and also no /proc/$pid/stat. There are /proc/$pid/status entries, but these do not offer a vmsize nor a full process state.
This patch makes use of the libkvm library (see `man 3 kvm`) to gather process details like top(1) or procstat(1) do. kvm has been available since FreeBSD 9. Sample output:

Bildschirmfoto_2020-09-01_01-43-43

I never used autoconf/automake before, and i'm not happy with all the `#ifdef HAVE_LIBKVM` in imap/http_admin.c. Also, both `imap/index.h` and `/usr/include/sys/proc.h` define a `struct thread`. So, what would be the best way to get this right? Two OS specific imap/http_admin_linux.c and http_admin_freebsd.c, each with its own add_procinfo() and output routines?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -650,6 +650,7 @@ cunit_TESTS = \
 	cunit/msgid.testc \
 	cunit/parseaddr.testc \
 	cunit/parse.testc \
+	cunit/procinfo.testc \
 	cunit/prot.testc \
 	cunit/ptrarray.testc \
 	cunit/quota.testc \
@@ -790,6 +791,7 @@ include_HEADERS = \
 	lib/murmurhash2.h \
 	lib/nonblock.h \
 	lib/parseaddr.h \
+	lib/procinfo.h \
 	lib/retry.h \
 	lib/rfc822tok.h \
 	lib/signals.h \
@@ -1470,6 +1472,7 @@ lib_libcyrus_la_SOURCES = \
 	lib/murmurhash.c \
 	lib/mkgmtime.c \
 	lib/parseaddr.c \
+	lib/procinfo.c \
 	lib/prot.c \
 	lib/ptrarray.c \
 	lib/rfc822tok.c \
--- a/cunit/procinfo.testc
+++ b/cunit/procinfo.testc
@@ -1,0 +1,163 @@
+/* unit test for lib/procinfo.c */
+#include <unistd.h> /* for getpid() */
+#include "config.h"
+#include "cunit/cyrunit.h"
+#include "lib/procinfo.h"
+
+static void test_init_piarray(void)
+{
+	piarray_t piarray;
+
+	init_piarray(&piarray);
+
+	CU_ASSERT_EQUAL(piarray.count, 0);
+	CU_ASSERT_EQUAL(piarray.alloc, 0);
+	CU_ASSERT_PTR_NULL(piarray.data);
+}
+
+#if 0
+static void not_test_add_procinfo_generic(void)
+{
+	piarray_t piarray;
+	struct proc_info *pinfo;
+
+	init_piarray(&piarray);
+
+	pinfo = add_procinfo_generic(&piarray, 2342, "service", "host",
+	                             "user", "mailbox", "command");
+
+	CU_ASSERT_PTR_NOT_NULL(pinfo);
+
+	CU_ASSERT_EQUAL(piarray.count, 1);
+	CU_ASSERT_TRUE((piarray.alloc >= piarray.count));
+	CU_ASSERT_PTR_NOT_NULL(piarray.data);
+
+	CU_ASSERT_EQUAL(pinfo->pid, 2342);
+	CU_ASSERT_STRING_EQUAL(pinfo->servicename, "service");
+	CU_ASSERT_STRING_EQUAL(pinfo->user, "user");
+	CU_ASSERT_STRING_EQUAL(pinfo->host, "host");
+	CU_ASSERT_STRING_EQUAL(pinfo->mailbox, "mailbox");
+	CU_ASSERT_STRING_EQUAL(pinfo->cmdname, "command");
+}
+#endif
+
+static void test_add_procinfo(void)
+{
+	piarray_t piarray;
+	struct proc_info *pinfo;
+	int res;
+
+	init_piarray(&piarray);
+
+	res = add_procinfo(getpid(), "service", "host", "user",
+	                   "mailbox", "command", &piarray);
+
+	CU_ASSERT_EQUAL(res, 0);
+
+	CU_ASSERT_EQUAL(piarray.count, 1);
+	CU_ASSERT_TRUE((piarray.alloc >= piarray.count));
+	CU_ASSERT_PTR_NOT_NULL(piarray.data);
+
+	pinfo = piarray.data[0];
+
+	CU_ASSERT_PTR_NOT_NULL(pinfo);
+	CU_ASSERT_EQUAL(pinfo->pid, getpid());
+	CU_ASSERT_STRING_EQUAL(pinfo->servicename, "service");
+	CU_ASSERT_STRING_EQUAL(pinfo->user, "user");
+	CU_ASSERT_STRING_EQUAL(pinfo->host, "host");
+	CU_ASSERT_STRING_EQUAL(pinfo->mailbox, "mailbox");
+	CU_ASSERT_STRING_EQUAL(pinfo->cmdname, "command");
+	CU_ASSERT_STRING_NOT_EQUAL(pinfo->state, "");
+	CU_ASSERT_NOT_EQUAL(pinfo->start, 0);
+	CU_ASSERT_TRUE((pinfo->start <= time(NULL)));
+	CU_ASSERT_TRUE((pinfo->vmsize > 0));
+}
+
+static void test_sort_procinfo(void)
+{
+	piarray_t piarray;
+	struct proc_info *pinfo1, *pinfo2, *pinfo3, *pinfo4;
+	int res;
+
+	init_piarray(&piarray);
+
+	res = add_procinfo(getpid(), "service1", "host1", "user1",
+	                   "mailbox1", "command1", &piarray);
+	CU_ASSERT_EQUAL(res, 0);
+	CU_ASSERT_EQUAL(piarray.count, 1);
+
+	res = add_procinfo(1, "service2", "host2", "user2",
+	                   "mailbox2", "command2", &piarray);
+	CU_ASSERT_EQUAL(res, 0);
+	CU_ASSERT_EQUAL(piarray.count, 2);
+
+	res = add_procinfo(getpid(), "service1", "host1", "user1",
+	                   "mailbox1", "command1", &piarray);
+	CU_ASSERT_EQUAL(res, 0);
+	CU_ASSERT_EQUAL(piarray.count, 3);
+
+	res = add_procinfo(1, "service2", "host2", "user2",
+	                   "mailbox2", "command2", &piarray);
+	CU_ASSERT_EQUAL(res, 0);
+	CU_ASSERT_EQUAL(piarray.count, 4);
+
+	pinfo1 = piarray.data[0];
+	pinfo2 = piarray.data[1];
+
+	snprintf(pinfo1->state, sizeof(pinfo1->state), "%s", "running1");
+	snprintf(pinfo2->state, sizeof(pinfo2->state), "%s", "running2");
+	pinfo1->start = 1;
+	pinfo2->start = 2;
+	pinfo1->vmsize = 1;
+	pinfo2->vmsize = 2;
+
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "p") < 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "s") > 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "q") > 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "t") > 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "v") > 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "h") > 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "u") > 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "r") > 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "c") > 0));
+
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "P") > 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "S") < 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "Q") < 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "T") < 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "V") < 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "H") < 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "U") < 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "R") < 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo2, "C") < 0));
+
+	pinfo3 = piarray.data[2];
+	pinfo4 = piarray.data[3];
+
+	snprintf(pinfo3->state, sizeof(pinfo3->state), "%s", "running1");
+	snprintf(pinfo4->state, sizeof(pinfo4->state), "%s", "running2");
+	pinfo3->start = 1;
+	pinfo4->start = 2;
+	pinfo3->vmsize = 1;
+	pinfo4->vmsize = 2;
+
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo3, "p") == 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo3, "s") == 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo3, "q") == 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo3, "t") == 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo3, "v") == 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo3, "h") == 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo3, "u") == 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo3, "r") == 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo1, &pinfo3, "c") == 0));
+
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo2, &pinfo4, "P") == 0));
+	CU_ASSERT_TRUE((sort_procinfo QSORT_R_COMPAR_ARGS(&pinfo2, &pinfo4, "S") == 0));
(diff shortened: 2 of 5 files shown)


elliefm:
This is hard to review/advise on because we don't really have much BSD-specific expertise.

I just quickly checked and it looks like OpenBSD also has a libkvm that looks (at a very casual glance) like it  provides similar interfaces, which seem to be inherited from SunOS.  So this implementation might be useful for BSDs generally and not just FreeBSD, but I'm not certain.

> I never used autoconf/automake before

Your Makefile.am and configure.ac changes look good to me, by eye. :)  I have not yet compile-tested against breakages on Linux, and would definitely do so before merging, but nothing looks out of place so far.

> and i'm not happy with all the #ifdef HAVE_LIBKVM in imap/http_admin.c.

Some amount of this stuff is usually a necessary evil, but it's definitely nice to minimise and encapsulate it where possible...

> Also, both imap/index.h and /usr/include/sys/proc.h define a struct thread.

Ouch.  `sys/proc.h` looks to be a BSD-specific header, so I have no specific insight to offer here.  If I ran into a problem like this on Linux I guess would look into whether I could trim back to a point where I didn't need sys/proc.h, but maybe it's fundamental to your implementation, I'm not sure.  Or I would try to encapsulate the use of the sys/proc.h APIs within a module that didn't also need imap/index.h, such that nothing ever sees both `struct thread`s at the same time.

> So, what would be the best way to get this right? Two OS specific imap/http_admin_linux.c and http_admin_freebsd.c, each with its own add_procinfo() and output routines?

If anything, I might propose splitting the procinfo bits out of imap/http_admin.c and into a new lib/procinfo.c (which would be included in libcyrus.so via lib_libcyrus_la_SOURCES in Makefile.am), and encapsulating the platform-specific bits behind some platform-agnostic interface there.  http_admin.c (and anything else that might eventually need it) can then just call the libcyrus interfaces, without needing to care about the underlying implementation.

I believe @obache and @anatoli26 have OpenBSD experience, having contributed a number of compatibility fixes, so their input/advice/testing would be very welcome here.

anatoli26:
Hi, in OpenBSD there's no `/proc` or `/admin` filesystems, I guess this feature won't work there. Though there is libkvm, so maybe there could be other uses apart from `/proc` entries?

felixjogris:
@anatoli26 You're right, but this patch is about the URL `/admin/proc` on the webserver which is part of Cyrus mailserver. On Linux, Cyrus uses information from the /proc filesystem, which just NetBSD has by default, but neither do Free-, Open-, nor DragonFlyBSD.
@elliefm I followed your advice and put OS specific routines into lib/procinfo.{c,h}. Also added support for NetBSD, OpenBSD, and DragonFlyBSD as well. For this, I examined the source code of `top` on all 4 BSD systems. The patch compiles and runs fine on FreeBSD. I will test it on the other 3 *BSDs next couple of days. If everything is fine, I will also add a unit tests.

anatoli26:
@felixjogris ahh ok, great! I will try to test it on OpenBSD this week.

felixjogris:
Patch applied to yesterday's checkout builds and runs on NetBSD 9.0, applied to vanilla Cyrus 3.2.3 builds and runs on DragonFlyBSD 5.8.1.
On DragonFlyBSD, even for an unpatched 3.2.3, I had to modify lib/cyr_qsort_r.c and imap/tls.c, because
- configure detected that DragonFly's gcc 8.3 has function nesting, but complains during compilation:
```
In file included from lib/cyr_qsort_r.c:1:
lib/cyr_qsort_r.h:42:55: error: conflicting types for 'qsort_r'
 #define cyr_qsort_r(base, nmemb, size, compar, thunk) qsort_r(base, nmemb, size, thunk, compar)
                                                       ^~~~~~~
lib/cyr_qsort_r.c:5:15: note: in expansion of macro 'cyr_qsort_r'
 EXPORTED void cyr_qsort_r(void *base, size_t nmemb, size_t size,
               ^~~~~~~~~~~
In file included from lib/cyr_qsort_r.h:19,
                 from lib/cyr_qsort_r.c:1:
/usr/include/stdlib.h:298:7: note: previous declaration of 'qsort_r' was here
 void  qsort_r(void *, size_t, size_t, void *,
       ^~~~~~~
```
- LibreSSL 2.9.1 seems not to define SSL_OP_NO_TLSv1_3:
```
imap/tls.c: In function 'tls_init_serverengine':
imap/tls.c:803:16: error: 'SSL_OP_NO_TLSv1_3' undeclared (first use in this function); did you mean 'SSL_OP_NO_TLSv1_1'?
         off |= SSL_OP_NO_TLSv1_3;
                ^~~~~~~~~~~~~~~~~
                SSL_OP_NO_TLSv1_1
imap/tls.c:803:16: note: each undeclared identifier is reported only once for each function it appears in
gmake[2]: *** [Makefile:5190: imap/imap_libcyrus_imap_la-tls.lo] Error 1
```

On OpenBSD 6.7, I'm stuck with some c++ and/or icu related issues. Vanilla 3.2.3 compiles until this:
```
mv -f imap/.deps/imap_libcyrus_imap_la-ical_support.Tpo imap/.deps/imap_libcyrus_imap_la-ical_support.Plo
/bin/sh ./libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I.  -I/root/cyrus-imapd-3.2.3/com_err/et -I. -I./lib -I. -I./lib -DLIBEXEC_DIR=\"/usr/local/libexec\" -DSBIN_DIR=\"/usr/local/sbin\" -DSYSCONF_DIR=\"/etc\" -DHAVE_CONFIG_H    -I/usr/local/include  -I/usr/local/include/libxml2 -I/usr/local/include  -I/usr/local/include  -I/usr/local/include -I/usr/local/include      -I/usr/local/include  -I/usr/local/include    -g -O2 -MT imap/imap_libcyrus_imap_la-icu_wrap.lo -MD -MP -MF imap/.deps/imap_libcyrus_imap_la-icu_wrap.Tpo -c -o imap/imap_libcyrus_imap_la-icu_wrap.lo `test -f 'imap/icu_wrap.cpp' || echo './'`imap/icu_wrap.cpp
libtool: compile:  g++ -DHAVE_CONFIG_H -I. -I/root/cyrus-imapd-3.2.3/com_err/et -I. -I./lib -I. -I./lib -DLIBEXEC_DIR=\"/usr/local/libexec\" -DSBIN_DIR=\"/usr/local/sbin\" -DSYSCONF_DIR=\"/etc\" -DHAVE_CONFIG_H -I/usr/local/include -I/usr/local/include/libxml2 -I/usr/local/include -I/usr/local/include -I/usr/local/include -I/usr/local/include -I/usr/local/include -I/usr/local/include -g -O2 -MT imap/imap_libcyrus_imap_la-icu_wrap.lo -MD -MP -MF imap/.deps/imap_libcyrus_imap_la-icu_wrap.Tpo -c imap/icu_wrap.cpp  -fPIC -DPIC -o imap/.libs/imap_libcyrus_imap_la-icu_wrap.o
In file included from /usr/local/include/unicode/uenum.h:23,
                 from /usr/local/include/unicode/ucnv.h:53,
                 from imap/icu_wrap.cpp:50:
/usr/local/include/unicode/localpointer.h:71: error: invalid pure specifier (only `= 0' is allowed) before ';' token
/usr/local/include/unicode/localpointer.h:71: error: initializer specified for static member function 'static void* icu::LocalPointerBase<T>::operator new(size_t)'
/usr/local/include/unicode/localpointer.h:72: error: invalid pure specifier (only `= 0' is allowed) before ';' token
/usr/local/include/unicode/localpointer.h:72: error: initializer specified for static member function 'static void* icu::LocalPointerBase<T>::operator new [](size_t)'
/usr/local/include/unicode/localpointer.h:74: error: invalid pure specifier (only `= 0' is allowed) before ';' token
/usr/local/include/unicode/localpointer.h:74: error: initializer specified for static member function 'static void* icu::LocalPointerBase<T>::operator new(size_t, void*)'
/usr/local/include/unicode/localpointer.h:224: error: expected ',' or '...' before '&&' token
/usr/local/include/unicode/localpointer.h:224: error: invalid constructor; you probably meant 'icu::LocalPointer<T> (const icu::LocalPointer<T>&)'
/usr/local/include/unicode/localpointer.h:224: error: expected ';' before 'noexcept'
/usr/local/include/unicode/localpointer.h:238: error: expected `;' before 'explicit'
/usr/local/include/unicode/localpointer.h:238: error: expected `)' before '<' token
/usr/local/include/unicode/localpointer.h:255: error: expected ',' or '...' before '&&' token
/usr/local/include/unicode/localpointer.h:255: error: expected ';' before 'noexcept'
/usr/local/include/unicode/localpointer.h:270: error: expected `;' before 'LocalPointer'
/usr/local/include/unicode/localpointer.h:270: error: declaration of 'operator=' as non-function
/usr/local/include/unicode/localpointer.h:270: error: expected ';' before '(' token
/usr/local/include/unicode/localpointer.h:280: error: expected `;' before 'void'
/usr/local/include/unicode/localpointer.h:280: error: expected ';' before 'noexcept'
/usr/local/include/unicode/localpointer.h:291: error: expected `;' before 'friend'
/usr/local/include/unicode/localpointer.h:291: warning: friend declaration 'void icu::swap(icu::LocalPointer<T>&, icu::LocalPointer<T>&)' declares a non-template function
/usr/local/include/unicode/localpointer.h:291: warning: (if this is not what you intended, make sure the function template has already been declared and add <> after the function name here) -Wno-non-template-friend disables this warning
/usr/local/include/unicode/localpointer.h:291: error: expected ';' before 'noexcept'
/usr/local/include/unicode/localpointer.h:300: error: expected `;' before 'void'
/usr/local/include/unicode/localpointer.h:342: error: expected type-specifier
/usr/local/include/unicode/localpointer.h:399: error: expected ',' or '...' before '&&' token
/usr/local/include/unicode/localpointer.h:399: error: invalid constructor; you probably meant 'icu::LocalArray<T> (const icu::LocalArray<T>&)'
/usr/local/include/unicode/localpointer.h:399: error: expected ';' before 'noexcept'
/usr/local/include/unicode/localpointer.h:413: error: expected `;' before 'explicit'
/usr/local/include/unicode/localpointer.h:413: error: expected `)' before '<' token
/usr/local/include/unicode/localpointer.h:430: error: expected ',' or '...' before '&&' token
/usr/local/include/unicode/localpointer.h:430: error: expected ';' before 'noexcept'
/usr/local/include/unicode/localpointer.h:445: error: expected `;' before 'LocalArray'
/usr/local/include/unicode/localpointer.h:445: error: declaration of 'operator=' as non-function
/usr/local/include/unicode/localpointer.h:445: error: expected ';' before '(' token
/usr/local/include/unicode/localpointer.h:455: error: expected `;' before 'void'
/usr/local/include/unicode/localpointer.h:455: error: expected ';' before 'noexcept'
/usr/local/include/unicode/localpointer.h:466: error: expected `;' before 'friend'
/usr/local/include/unicode/localpointer.h:466: warning: friend declaration 'void icu::swap(icu::LocalArray<T>&, icu::LocalArray<T>&)' declares a non-template function
/usr/local/include/unicode/localpointer.h:466: error: expected ';' before 'noexcept'
/usr/local/include/unicode/localpointer.h:475: error: expected `;' before 'void'
/usr/local/include/unicode/localpointer.h:525: error: expected type-specifier
In file included from /usr/local/include/unicode/ucnv.h:53,
                 from imap/icu_wrap.cpp:50:
/usr/local/include/unicode/uenum.h:68: error: expected ',' or '...' before '&&' token
/usr/local/include/unicode/uenum.h:68: error: invalid constructor; you probably meant 'icu::LocalUEnumerationPointer (const icu::LocalUEnumerationPointer&)'
/usr/local/include/unicode/uenum.h:68: error: expected ';' before 'noexcept'
/usr/local/include/unicode/uenum.h:68: error: expected `;' before 'explicit'
/usr/local/include/unicode/uenum.h:68: error: expected `)' before '<' token
/usr/local/include/unicode/uenum.h:68: error: expected ',' or '...' before '&&' token
/usr/local/include/unicode/uenum.h:68: error: expected ';' before 'noexcept'
imap/icu_wrap.cpp:76: error: expected `;' at end of input
imap/icu_wrap.cpp:76: error: expected `}' at end of input
/usr/local/include/unicode/uenum.h:68: error: expected unqualified-id at end of input
/usr/local/include/unicode/uenum.h:68: error: expected `}' at end of input
gmake[2]: *** [Makefile:6215: imap/imap_libcyrus_imap_la-icu_wrap.lo] Error 1
gmake[2]: Leaving directory '/root/cyrus-imapd-3.2.3'
gmake[1]: *** [Makefile:6561: all-recursive] Error 1
gmake[1]: Leaving directory '/root/cyrus-imapd-3.2.3'
gmake: *** [Makefile:2898: all] Error 2
```
However, these issues are not related to this patch.

elliefm:
> On DragonFlyBSD, even for an unpatched 3.2.3, I had to modify lib/cyr_qsort_r.c and imap/tls.c, because
> 
> * configure detected that DragonFly's gcc 8.3 has function nesting, but complains during compilation:

I think this would be fixed by applying 4dd810dff1 from #3154 (or just building from the cyrus-imapd-3.2 git branch rather than the release tarball).  It'll be in 3.2.4 when it's released, but 3.2.3 doesn't have it.

> * LibreSSL 2.9.1 seems not to define SSL_OP_NO_TLSv1_3:

OpenBSD ports (or whatever their equivalent thing is called) has a patch to work around this, I believe.  We considered upstreaming it in #3129, but weren't sure exactly what it did, or whether any other BSD's needed it, and it looked like it might be soon obsoleted, so we left it out of upstream.  It remains an OpenBSD-only patch for now.  Maybe it's worth revisiting upstreaming it, maybe not, I'm still not sure -- the feedback from another BSD is definitely useful though!

> On OpenBSD 6.7, I'm stuck with some c++ and/or icu related issues. Vanilla 3.2.3 compiles until this:
> 
> ```
> [...]
> ```
> 
> However, these issues are not related to this patch.

I have no insight here.  Maybe @anatoli26 does?

elliefm:
> Merge remote-tracking branch 'upstream/master' into freebsd-admin-proc

I just noticed this commit in your branch, can you rebase it out please?  (I assume you know how to do this, but let me know if you need guidance!)  Having merges in feature branches turns the history graph into spaghetti when the feature branch is itself merged later on.  Generally, if you need to update your feature branch on top of the current upstream, please rebase it rather than merging it.  Thanks :)

elliefm:
I've just tried this out on Debian, and needed the following changes for it to build (seems like just little things overlooked during conversion):

```
diff --git a/lib/procinfo.c b/lib/procinfo.c
index 451c7f5114..647979746a 100644
--- a/lib/procinfo.c
+++ b/lib/procinfo.c
@@ -1,3 +1,6 @@
+#include <sys/types.h>
+#include <sys/stat.h>
+
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
@@ -6,6 +9,7 @@
 
 #include "procinfo.h"
 #include "xmalloc.h"
+#include "xstrlcat.h"
 
 static struct proc_info *
 add_procinfo_generic(piarray_t *piarray, pid_t pid, const char *servicename,
@@ -461,7 +465,7 @@ EXPORTED int add_procinfo(pid_t pid, const char *servicename, const ch
     free(s);
     fclose(f);
 
-    if (c == EOF) {
+    if (res == EOF) {
         return 0;
     }
 
```

felixjogris:
> > On DragonFlyBSD, even for an unpatched 3.2.3, I had to modify lib/cyr_qsort_r.c and imap/tls.c, because
> > 
> > * configure detected that DragonFly's gcc 8.3 has function nesting, but complains during compilation:
> 
> I think this would be fixed by applying [4dd810d](https://github.com/cyrusimap/cyrus-imapd/commit/4dd810dff14a0131d82b22cab4ceefe37a0dca56) from #3154 (or just building from the cyrus-imapd-3.2 git branch rather than the release tarball). It'll be in 3.2.4 when it's released, but 3.2.3 doesn't have it.

It does. I should have thought of that myself.

> 
> > * LibreSSL 2.9.1 seems not to define SSL_OP_NO_TLSv1_3:
> 
> OpenBSD ports (or whatever their equivalent thing is called) has a patch to work around this, I believe. We considered upstreaming it in #3129, but weren't sure exactly what it did, or whether any other BSD's needed it, and it looked like it might be soon obsoleted, so we left it out of upstream. It remains an OpenBSD-only patch for now. Maybe it's worth revisiting upstreaming it, maybe not, I'm still not sure -- the feedback from another BSD is definitely useful though!

It fixes this issue on DragonFlyBSD as well.

felixjogris:
> > Merge remote-tracking branch 'upstream/master' into freebsd-admin-proc
> 
> I just noticed this commit in your branch, can you rebase it out please? (I assume you know how to do this, but let me know if you need guidance!) Having merges in feature branches turns the history graph into spaghetti when the feature branch is itself merged later on. Generally, if you need to update your feature branch on top of the current upstream, please rebase it rather than merging it. Thanks :)

Thanks for your explanation. I did:

- git checkout master
- git pull upstream master
- git checkout freebsd-admin-proc
- git rebase master
- git push --all --force-with-lease origin

with origin being https://github.com/felixjogris/cyrus-imapd.git. Hope I didn't make it worse.

anatoli26:
@felixjogris, on OpenBSD I suggest you try to build it using the ports infrastructure.

Clone to `/usr/ports` the https://github.com/openbsd/ports repo, then go to `/usr/ports/mail/cyrus-imapd` and issue there:
```
make patch
make configure
make build
make package
```

(you can also do it in just one command `make package`, it would do all the previous stages/commands for you, but one command at a time you'll see the outputs of each logical stage separately).

To start it again issue: `make clean`.

The patches are in the `./patches` directory inside each port.

In order to create a new patch you can do the following from within the port's dir (e.g. `/usr/ports/mail/cyrus-imapd`):
```
cd `make show=WRKSRC` ; cp foo/bar.c{,.orig}
edit foo/bar.c to fix the error.
cd - ; make update-patches
```
The `./patches` dir would now have new `patch-xxx` files corresponding to each file you changed.

The ports infrastructure has everything needed to successfully build all ports in the official repo.

You can also check these docs:
 * https://www.openbsd.org/faq/ports/ports.html
 * https://www.openbsd.org/faq/ports/guide.html
 * https://www.openbsd.org/faq/ports/index.html
 * https://man.openbsd.org/bsd.port.mk.5

obache:
It seems that `libkvm` is not used on OpenBSD, using `sysctl` instead of `kvm_getprocs()` so no need to check HAVE_LIVKVM and no need to link against `libkvm`.
It is possible to apply similar replacement on NetBSD, using `KERN_PROC2` instead of `KERN_PROC` as `mib` with `struct kinfo_proc2 *kip`.
FreeBSD and DragonFly BSD might too.

felixjogris:
@obache Thanks for the hint. I just rewrote the patch with sysctl() and without libkvm at all.

felixjogris:
@anatoli26 Thank you for your detailed help. That's an elegant way btw. I was successful in building 3.2.3 on OpenBSD 5.8-current. I placed this patch and [4dd810d](url) from [3154](url) into patches/ as you described. But I also had to modify the Makefile:
```
diff --git a/mail/cyrus-imapd/Makefile b/mail/cyrus-imapd/Makefile
index 6b5072e0498..de23726f85f 100644
--- a/mail/cyrus-imapd/Makefile
+++ b/mail/cyrus-imapd/Makefile
@@ -37,7 +37,7 @@ LIB_DEPENDS=          databases/sqlite3 \
                        sysutils/e2fsprogs \
                        textproc/icu4c
 
-CONFIGURE_STYLE=       gnu
+CONFIGURE_STYLE=       autoreconf
 CONFIGURE_ENV=         CPPFLAGS="-I${LOCALBASE}/include" \
                        LDFLAGS="-L${LOCALBASE}/lib" \
                        cyrus_cv_sse42=no
@@ -62,7 +62,7 @@ CONFIGURE_ARGS +=     --without-snmp
 
 # XXX notyet; FLAVOR
 CONFIGURE_ARGS +=      --without-clamav \
-                       --disable-http
+                       --enable-http --without-zlib
 
 DEBUG_PACKAGES=                ${BUILD_PACKAGES}
 
diff --git a/mail/cyrus-imapd/pkg/PLIST b/mail/cyrus-imapd/pkg/PLIST
index 353fcf89e61..a0c6ac36929 100644
--- a/mail/cyrus-imapd/pkg/PLIST
+++ b/mail/cyrus-imapd/pkg/PLIST
@@ -24,6 +24,7 @@ cyrus/libexec/
 @bin cyrus/libexec/fud
 @bin cyrus/libexec/idled
 @bin cyrus/libexec/imapd
+@bin cyrus/libexec/httpd
 @bin cyrus/libexec/lmtpd
 cyrus/libexec/lmtpproxyd
 @bin cyrus/libexec/master
```
httpd does not get build normally, as zlib is too old and does not offer deflatePending(), I guess.
`CONFIGURE_STLYE=autoreconf` is only needed as my patch modifies Makefile.am in order to build lib/procfile.{c,h}

anatoli26:
@felixjogris, yes, your modifications to the `Makefile` are necessary for what you're testing.

By default, the port uses the pre-built `configure`, so it's necessary to add `autoreconf`, as you've done, to make the port to rebuild it every time (I was thinking about adding it permanently to the port to facilitate the process of testing new code). Alternatively, you could also do `make gen` to regenerate it once (you'd have to repeat it after `make clean`).

With respect to `--enable-http`, currently the port builds only the non-http (and non-other_components) version. I'm in the process of defining the port's flavors, i.e. when multiple variations of the port are built with different params, in particular a http-enabled flavor.

felixjogris:
@elliefm Any chance this will be in 3.2.5?

elliefm:
I think this is more of a new feature than a simple bug fix, and it's a decent sized set of changes, so I think I'd merge it to master for 3.4, but not backport it to 3.2.  We're on a yearly release cycle now, so 3.4 is not far away anyway: the master branch will go into feature freeze at the end of the year, and then there'll be some betas early in the new year and a stable release around feb/march or so.

How do you feel about rewriting the history, so that there isn't all the fixes-a-previous-commit commits in there?  I'm mostly bothered by "FIXME: el-cheapo fix for struct thread"; but, going by commit messages, "small fixes", "small cleanup" and "fixed unit test for..." look like they could probably also be easily squashed into the original commit that they're repairing.  I'm happy to guide you through this if you wanna try but need help.  It makes the history a bit easier to follow when the trivial mistakes "never even happened". :)

I'm not sure how to feel about the inclusion of code from each BSD's `top` tool -- our license is 4-clause BSD, which I think is incompatible with the actual BSDs' these days.  The included snippets look at a casual glance like "this is the only sensible way to do it", so I'm not sure it makes sense to "rewrite" them; but I'm also deeply uncomfortable making licensing decisions alone.  I've added it to the agenda for this week's Cyrus conference call to talk about it, and will let you know how that goes.

elliefm:
We talked about the code based on top last week, and consensus was it seems fine to include it, since the point is to describe the processes in the same way the system tools do, and that means doing it the same way.

felixjogris:
@elliefm That's great news. Shall I rebase -i all my intermediate commits and push --force them along with a more accurate message?

elliefm:
That would be great, please do!  Thanks :)

felixjogris:
@elliefm Done so.

elliefm:
I've merged this (well, rebase-and-ff'd, since it was only one commit) by hand so that I could sort out the conflict with @anatoli26's format-specifier changes, and to save @felixjogris yet another rebase. :)  It's now on master as 93061c5c8e

elliefm:
Thanks heaps for your work on this :)

PR #3167, opened 2020-09-02
cyr_qsort_r: only compile fallback function if we need it

elliefm:
This is the fix for #3154, in which we discovered that we were accidentally (by way of macro expansion in cyr_qsort_r.c) replacing the real `qsort_r()` with our last resort fallback function, instead of simply having it as a last resort fallback.

The patch is fairly trivial and I think obviously correct (and works fine for me), BUT it has the "side effect" that we're now definitely using the real `qsort_r()`, whereas before we might not have been (depending on link order, I guess).

So I'd appreciate if we could all give this a thorough test before we merge it, just in case some new edge case shakes out from switching underlying implementations.

### DIFF
--- a/lib/cyr_qsort_r.c
+++ b/lib/cyr_qsort_r.c
@@ -1,5 +1,7 @@
 #include "cyr_qsort_r.h"
 
+#ifndef cyr_qsort_r
+
 #ifdef HAVE_FUNCTION_NESTING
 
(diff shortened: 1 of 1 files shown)


rsto:
@elliefm thanks. I ran this branch on the JMAP test suites which have a couple of tests that assert the right sort order. It all looks good.

rsto (approved):

elliefm:
Thanks :)

PR #3171, opened 2020-09-04
Xapian compact nowrite

brong:
Yes, this is a faff - but I figured I'd have a look at whether it's possible.

I'd love to be able to pass the metadata overwrite information in as a hashtable or something rather than having to do this horror show of making a whole separate database.  I think I might have to make a PR against Xapian for it...

### DIFF
--- a/imap/xapian_wrap.cpp
+++ b/imap/xapian_wrap.cpp
@@ -172,7 +172,7 @@ static int calculate_language_counts(const Xapian::Database& db,
     return 0;
 }
 
-static void remove_legacy_metadata(Xapian::WritableDatabase& db)
+static void remove_legacy_metadata(Xapian::WritableDatabase& metadb, Xapian::Database& db)
 {
     const std::string prefix{XAPIAN_LANG_DOC_KEYPREFIX "."};
     for (Xapian::TermIterator key = db.metadata_keys_begin(prefix);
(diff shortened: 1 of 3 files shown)


imap/xapian_wrap.cpp
```diff
@@ -181,22 +181,10 @@ static void remove_legacy_metadata(Xapian::WritableDatabase& db)
...
-            if (!val.empty() && !isalpha(val[0])) {
-                doc.remove_value(SLOT_DOCLANGS);
-            }
-        }
-        catch (Xapian::DocNotFoundError e) {
-            // ignore
+            metadb.set_metadata(*key, "OVERWRITE:");
         }
     }
+    // NOTE: we're not removing DOCLANG slot because we can't tell the compactor to remove it
```
rsto, line 187: Yeah, that's too bad indeed. We could do that in a two-pass compact if we encounter a db with a legacy language count slot. Our reindexed dbs shouldn't have one anyway.

imap/xapian_wrap.cpp
```diff
@@ -518,20 +513,57 @@ int xapian_compact_dbs(const char *dest, const char **sources)
...
+            if (it != db_versions.begin()) val << ",";
+            val << *it;
+        }
+        if (db.get_metadata("cyrus.db_version").empty()) {
+            metadb.set_metadata("cyrus.db_version", val.str());
+        }
+        else {
+            metadb.set_metadata("cyrus.db_version", std::string("OVERWRITE:") + val.str());
+        }
+        if (db.get_metadata("cyrus.stem-version").empty()) {
```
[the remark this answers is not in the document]
brong: Yeah, that's not actually necessary - it's a noop, setting a metadata value to blank is the same as removing it, but since this is a new DB clearly it's not going to have that value!  The other path ("OVERWRITE:") is necessary, because that will cause it to replace any metadata value from the other databases in what gets written to the new one :)

imap/xapian_wrap.cpp
brong: I'm not too stressed about that then, because we'll be reindexing everything up to v15 over the next few weeks.

brong:
So ... I wrote about the Honey issue on the Xapian mailing list:

https://lists.xapian.org/pipermail/xapian-devel/2020-September/003384.html

And Olly replied - he's going to look put the "don't look if the DB is empty" logic into the Honey repacker so it doesn't crash, then we could use Honey format indexes for our compact targets :)

Bron.

brong:
I think this can probably just die.  Moving to honey isn't on the roadmap any time soon.

PR #3172, opened 2020-09-04
Not all systems have memrchr()

citrus-it:
I've just finished packaging Cyrus imapd for OmniOS which is an illumos (née OpenSolaris) distribution.

There were a few small compatibility changes needed, this is the first.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -193,7 +193,7 @@ dnl check for -R, etc. switch
 CMU_GUESS_RUNPATH_SWITCH
 
 AC_CHECK_HEADERS(unistd.h sys/select.h sys/param.h stdarg.h)
-AC_REPLACE_FUNCS(memmove strcasecmp ftruncate strerror posix_fadvise strsep memmem)
+AC_REPLACE_FUNCS(memmove strcasecmp ftruncate strerror posix_fadvise strsep memmem memrchr)
 AC_CHECK_FUNCS(strlcat strlcpy strnchr getgrouplist fmemopen pselect futimens futimes)
 AC_HEADER_DIRENT
 
@@ -2354,6 +2354,10 @@ extern char *strsep(char **, const char *);
 
 #ifndef HAVE_MEMMEM
 extern void *memmem(const void *, size_t, const void *, size_t);
+#endif
+
+#ifndef HAVE_MEMRCHR
+extern void *memrchr(const void *, int, size_t);
 #endif
 
 /* compile time options; think carefully before modifying */
--- a/lib/memrchr.c
+++ b/lib/memrchr.c
@@ -1,0 +1,64 @@
+/* memrchr.c -- replacement memrchr() routine
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in
+ *    the documentation and/or other materials provided with the
+ *    distribution.
+ *
+ * 3. The name "Carnegie Mellon University" must not be used to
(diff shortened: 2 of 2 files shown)


citrus-it:
Alternatively, the one place that uses `memrchr()` could be changed to not use it - let me know your preference.

elliefm:
Thanks.  I had a quick look to check the complexity of just not using `memrchr()`, and the one place it's "actually used" is here:

https://github.com/cyrusimap/cyrus-imapd/blob/4376a87773f709516105b84e3cf12b7d55775a55/imap/conversations.c#L528-L533

Which you can see has already been disabled by bd2a3787f9 on the master branch.  But I guess you're packaging 3.2, which doesn't have this commit, and therefore still needs `memrchr()`.

Rather than adding a memrchr implementation, maybe we just need to apply the same change to 3.2 as was done on master.  @brong @ksmurchison what's the production ramification of suddenly allowing multiple '@'s in msgids here?  Will admins need to rebuild their conversationsdb's?  Is it pretty seamless?

ksmurchison:
I'll defer the conv.db question to @brong, but if we keep this check I don't see why we couldn't use strchr() and strrchr() instead.

brong:
strrchr would be fine for this case.  There are some places where memchr is needed because the buffer could contain NULLs, but not here.

brong:
(also I don't really care that much about this test - we could drop it)

ksmurchison:
I don't see any harm in adding a replacement for memrchr() regardless of whether we keep the check in conversations.c.  The source won't be compiled if we don't need it.

citrus-it:
Let's go for this then and avoid the non-standard memrchr() entirely. I know it's disabled code in master - can this patch be back-ported to 3.2.<next>?

elliefm:
I applied the merge locally and then cherry-picked the commit to 3.2 to see what would happen.  Builds fine, but one of the conversations unit tests fails in three places:

```
Suite: conversations
  Test: open ...passed
  Test: getset ...passed
  Test: abort ...passed
  Test: prune ...FAILED
    1. ./cunit/conversations.testc:203  - nseen >= 3
    2. ./cunit/conversations.testc:204  - ndeleted >= 2
    3. ./cunit/conversations.testc:221  - CU_ASSERT_EQUAL(arrayu64_size(&cids)=1,0=0)
  Test: two ...passed
  Test: cid_encode ...passed
  Test: cid_decode ...passed
  Test: folder_rename ...passed
  Test: folders ...passed
  Test: folder_ordering ...passed
  Test: dump ...passed
  Test: subject_normalise ...passed
```

Guessing at the cause, the `msgid` buffer is probably not 0-terminated (which is unsurprising, since we're passing around an explicit length with it), so it's not safe to use the `str...` functions on it.  Which might be the real reason why we used `memchr` and `memrchr` to begin with.

The internet suggests `strnchr` and `strnrchr` (which each take a length parameter) might exist, except they also don't seem to consistently exist, so that's hardly a solution to "memrchr might not exist"...

At this point I wonder if maybe you should just patch this within your package -- either by grabbing bd2a3787f9 from the master branch (to disable this check altogether), or by including a memrchr replacement.  I'm not sure there's an option here that's sensible to include upstream.

citrus-it:
Thanks for checking - I'm surprised that msgids can contain embedded NULs, but fair enough.

Back to this then, including a memrchr() replacement, which is the local patch I'm already carrying.
I could rewrite the existing check to just use memchr() too but I don't think there is any harm in including the replacement when using non-standard functions. As @ksmurchison said, if it is not needed then it won't be compiled.

elliefm:
> I'm surprised that msgids can contain embedded NULs

They can't; but they can be stored in unterminated buffers (e.g. fixed-width field in an mmap'd file...), in which case the `str...` functions overrun the end.

> Back to this then,

Yeah, that looks alright to me.  I can't easily test the replacement function, but I suppose it won't break for anyone who it wasn't already broken for anyway...

elliefm:
This is on the 3.2 branch as 94d0c02555 and will be in the next release.  Thanks :)

PR #3174, opened 2020-09-05
Not all systems have futimes() or TIMESPEC_TO_TIMEVAL()

citrus-it:
The `futimes()` function is not standardised and only available in Linux and some BSDs.

Check if it exists before using it, and fall back to alternatives.

The TIMESPEC_TO_TIMEVAL() macro is also not available on all platforms.

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -604,10 +604,25 @@ static int _copyfile_helper(const char *from, const char *to, int flags)
     }
 
     if (keeptime) {
+        int ret;
+#if defined(HAVE_FUTIMENS)
+        struct timespec ts[2];
+
+        ts[0] = sbuf.st_atim;
+        ts[1] = sbuf.st_mtim;
+        ret = futimens(destfd, ts);
+#elif defined(HAVE_FUTIMES)
         struct timeval tv[2];
+
         TIMESPEC_TO_TIMEVAL(&tv[0], &sbuf.st_atim);
         TIMESPEC_TO_TIMEVAL(&tv[1], &sbuf.st_mtim);
-        if (futimes(destfd, tv)) {
+        ret = futimes(destfd, tv);
+#else
+        close(destfd);
+        destfd = -1;
+        ret = utimes(to, tv);
+#endif
+        if (ret) {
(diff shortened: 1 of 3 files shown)


lib/util.c
```diff
@@ -607,7 +607,13 @@ static int _copyfile_helper(const char *from, const char *to, int flags)
         struct timeval tv[2];
         TIMESPEC_TO_TIMEVAL(&tv[0], &sbuf.st_atim);
         TIMESPEC_TO_TIMEVAL(&tv[1], &sbuf.st_mtim);
+#if defined(HAVE_FUTIMES)
         if (futimes(destfd, tv)) {
+#elif defined(HAVE_FUTIMESAT)
+	if (futimesat(destfd, NULL, tv)) {
+#else
+	if (utimes(to, tv)) {
```
elliefm, line 615: These lines have tab characters in them, can you please convert those to 8-spaces to match the surrounding style?  Thanks.

elliefm (changes requested):
Thanks!  Visually, this looks good once the tab/space changes are made.

I can't test this on a system without these functions, but it doesn't break anything on my system with them.

citrus-it:
> Thanks! Visually, this looks good once the tab/space changes are made.
> 
> I can't test this on a system without these functions, but it doesn't break anything on my system with them.

I was thinking about this over the weekend, and I think `futimens()` should be preferred as that's standardised in POSIX, and also avoids the conversion between timeval and timespec.

I'll rework this later today (and fix the tabs!)

citrus-it:
Here's a new version which will use the POSIX `futimens()` function if it's available, then fallback to the non-standard `futimes()` if it is available, and finally just close the file and use `utimes()` as a last resort.

I confirmed that on OmniOS, `futimens()` is now being used, and that `config.h` looks right.

```
% grep -n UTIME config.h
147:#define HAVE_FUTIMENS 1
150:/* #undef HAVE_FUTIMES */
```

elliefm (approved):

elliefm:
Thanks :)

Also cherry-picked to cyrus-imapd-3.2 branch as a1f974f684, so the next stable release will include this fix.

PR #3176, opened 2020-09-09
user_sieve_path() needs to work around caret-escaped userids

elliefm:
This is the minimally-intrusive fix for #3169 that I plan to roll out for 3.2, and I would like to also add it to master so we don't get out of sync.  A real fix will probably take some surgery in the mbname API, which would be nice to do, but not for 3.2.

The problem hereby solved manifests when replicating from 2.4 with unixhierarchysep, which caret-escapes localpart dots in the sync USERID fields(!).  It's mostly quietly glossed over and not a problem; but if you're replicating sieve scripts specifically and you have fulldirhash enabled, then a modern-day replica hashes the caret-escaped version of the userid instead of the real userid, and puts the uploaded/activated scripts in the wrong on-disk location, where nothing else can find them.

My patch has been confirmed to fix the problem by the user that reported it.  But I would like some other eyes/brains on this before I merge it, in case there's a better unintrusive approach, or a subtlety I've missed and need to consider.

### DIFF
--- a/imap/user.c
+++ b/imap/user.c
@@ -133,10 +133,21 @@ static int user_deleteacl(char *name, int matchlen, int category, void* rock)
 }
 #endif
 
-EXPORTED const char *user_sieve_path(const char *user)
+EXPORTED const char *user_sieve_path(const char *inuser)
 {
     static char sieve_path[2048];
     char hash, *domain;
+    char *user = xstrdupnull(inuser);
+    char *p;
+
+    /* Make sure it's a real userid, with no ^ escaping!
+     * XXX It's kinda bogus to be handling this here; it should be fixed
+     * XXX much further up somewhere, but that may require deep surgery.
+     */
+    for (p = user; p && *p; p++) {
+        if (*p == '^')
+            *p = '.';
+    }
 
     if (config_virtdomains && (domain = strchr(user, '@'))) {
         char d = (char) dir_hash_c(domain+1, config_fulldirhash);
@@ -154,6 +165,7 @@ EXPORTED const char *user_sieve_path(const char *user)
                  config_getstring(IMAPOPT_SIEVEDIR), hash, user);
     }
 
+    free(user);
     return sieve_path;
 }
 


ksmurchison (approved):
If it works, it works

dilyanpalauzov:
Detecting the sieve path for a user is re-implemented in imap/user.c:user_rename_sieve().  I cannot say, if the parameters of user_rename_sieve() are passed, after the caret is replaced with dot before the call.  If they should be caret-escaped, but are not, replacing
```c
*oldpath = strdup(user_sieve_path(olduser));
*newpath = user_sieve_path(newuser);
```
with the path-detection code in user_rename_sieve(), might be a very good idea.

Is there a reason, why user_rename_sieve() does not call user_sieve_path()?  Or is this just consequence of the fact that user_rename_sieve() exists since [2003](8d6cf338198bae0), while user_sieve_path() was born in [2010](262ace2630d1)?

elliefm:
Good catch, I'll have a look and see if it also needs fixing.  Thanks

As to why it's different, I think you're probably right, and it's just that one existed before the other.

ksmurchison:
You could take a look at the usermeta-by-id-bis branch where I believe that I consolidated these.  Of course on that branch, renaming a user is essentially a no-op

elliefm:
`user_renamesieve()` is only called from `user_renamedata()`, which in turn is only called from `cmd_rename()`.  `cmd_rename()` uses the `mboxname_...` APIs to convert its external-namespaced arguments to internal namespace and userids, so this is all fine.  It's not going to be getting "internal" encoding across the wire (unlike the sieve replication code), and if it does, the mboxname APIs will throw it out.

So the only issue here is the duplicated-looking parser in `user_renamesieve()`.  I haven't looked closely to see whether it is exactly the same as user_sieve_path (and thus could be deduplicated) or whether it's bespoke for some reason.

But I checked the usermeta-by-id-bis branch, and this is indeed consolidated (or at least: `user_renamedata()` and `user_renamesieve()` are ifdef'd out and no longer used; I didn't look for their replacements, if any).  Anyway, given that we already have a refactor planned for this, I won't bother deduplicating `user_renamesieve()` at the moment.

PR #3177, opened 2020-09-10
jmap_mail: add non-mail folders to foldernum_by_mboxname

brong:
We need this for inMailboxOtherThan to work.

I added a second part to this to also label the mailbox as not existing!

There's a related test over in Cassandane: https://github.com/cyrusimap/cassandane/tree/notes-guidsearch


rsto (approved):
Looks good to me!

PR #3179, opened 2020-09-12
DEBUG: syslog lock ordering to track deadlocks

brong:
NOT FOR INCLUSION IN UPSTREAM.

This is a useful thing to add to debug builds for tracking the order in which locks are created.


elliefm:
fwiw I think you can use the "Still in progress? Convert to draft" button in the right hand sidebar to flag the PR such that it can't be accidentally merged

brong:
We've reworked so much, this is no longer useful

Issue #3043, opened 2020-05-12
SET METADATA error with user defined annotation


falon:
Hello,

  let suppose I define in imapd.conf
`annotation_definitions: /etc/annoIMAP.conf`

and `annoIMAP.conf` is:

```/vendor/EXAMPLE/partition/mypart,server,string,backend,value.priv,lrswipkxtea```

Then the SETMETADATA fails:
```
a SETMETADATA "" (/private/vendor/EXAMPLE/partition/mypart "prova1")
a NO Permission denied 

```
I have to write the name downcase in annotation_definitions:

```/vendor/example/partition/mypart,server,string,backend,value.priv,lrswipkxtea```
In this last case:
```
a SETMETADATA "" (/private/vendor/EXAMPLE/partition/mypart "prova1")
a OK Completed
```

The RFC5464 says:

> "Except as noted otherwise, all alphabetic characters are case-
>    insensitive.  The use of upper or lower case characters to define
>    token strings is for editorial clarity only.  Implementations MUST
>    accept these strings in a case-insensitive fashion."

this is true in IMAP query, but in the definition file the names must be downcase.

I don't know if you can consider this like a bug. At least the man of imapd.conf could remember to set the  annotation_definitions names in downcase fashion.

See also [here](https://lists.andrew.cmu.edu/pipermail/info-cyrus/2020-April/041217.html).
Thank you

**Comment by elliefm** [comment]
I think the fix for this might be as simple as adding an `lcase(ae->name);` after this line:
https://github.com/cyrusimap/cyrus-imapd/blob/94ecc2c44a252f76b0746cbbc3504a05d9265b92/imap/annotate.c#L4046

I haven't seen this option before.  We have cunit tests for the parser for it, but we don't seem to have cassandane tests for its behaviour, and maybe we should.

**Comment by elliefm** [comment]
I don't have a strong opinion on whether or not this is a bug, but it looks like the case-sensitivity of that file has existed for as long as the option to use it has, wow.

**Comment by falon** [comment]
I agree. The real behavior is quite confusing. I need to explain better, because the possible bug is not really related to the downcase in 3.x version, but is related to the different behavior from SETANNOTATION on 2.4 releases and SETMETADATA on 3.x releases.

Let suppose to provide:

```
/vendor/EXAMPLE/partition/comune.prova.it,server,string,backend,value.priv,lrswipkxtea
```
with the exact case shown above on 2.4 and 3.2 Cyrus IMAP servers.

[...]

PR #3181, opened 2020-09-14
load server-defined annotation definitions file case-insensitively

elliefm:
Per RFC, annotations are meant to be case-insensitive, but if the `annotation_definitions` option is set to a file of additional server-defined annotation definitions, the case in the file is currently preserved.  Our case insensitivity is implemented by expecting the definitions to be lowercase, but the examples in the RFC are in uppercase for some reason, which causes problems for admins who base their custom definitions on the examples (#3043).

It's trivially fixed by `lcase()`ing each annotation definition as it's loaded from the file, which is the primary purpose of this PR.  But our DAV annotations are case-sensitive.  Is this going to be a problem?  Is it useful for an admin to define custom DAV annotations, which need case sensitivity?

This PR also adds a unit test to prove the problem/solution; and fixes a bug in annotate.c (found by the test) whereby annotation definitions loaded by `annotate_init()` were not unloaded by `annotate_done()`, which led to state leaking between the annotate tests.

If we decide we don't want the `lcase()` fix for DAV reasons, I'll still merge the annotate.c fix.

### DIFF
--- a/cunit/annotate.testc
+++ b/cunit/annotate.testc
@@ -20,6 +20,7 @@
 #define PARTITION       "default"
 #define COMMENT         "/comment"
 #define EXENTRY         "/vendor/example.com/a-non-default-entry"
+#define UCEXENTRY       "/vendor/EXAMPLE.COM/a-non-default-entry"
 #define VALUE_SHARED    "value.shared"
 #define SIZE_SHARED     "size.shared"
 #define VALUE0          "Hello World"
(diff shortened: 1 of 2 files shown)


ksmurchison (commented):
So, DAV property names have to be universally unique per RFC 4918, Section 4.4, and therefore are case-sensitive.  Plus, our DAV server doesn't/shouldn't use the annotation definitions file because we allow clients to set any "dead" DAV properties that they want (also per the RFC IIRC).

So, I'm fine with the changes for IMAP purposes, as long as they don't impact the current state of our DAV code.

brong (approved):

PR #3188, opened 2020-09-22
Allow server set props

ksmurchison:
This reverts the two server-set property validation commits for Foo/set

[ not to be merged:  this is working around a bug to be fixed elsewhere ]

### DIFF
--- a/cassandane/Cassandane/Cyrus/JMAPContacts.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPContacts.pm
@@ -458,7 +458,7 @@ sub test_contact_set_nickname
     $self->assert_not_null($res);
 }
 
-sub test_contact_set_invalid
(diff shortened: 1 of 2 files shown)


elliefm:
Are we gonna remove these commits entirely, or is this a temporary FM-only thing while sorting out something in our client/middleware?

I ask because I've already backported these commits onto the 3.2 branch, but they haven't been included in a release yet.  Do you I need to revert them there as well?

rjbs:
@ksmurchison It would be useful to know the bug this was working around (or other specific rationale).

ksmurchison:
Have run FM testsuite against current master without this PR and it passes.  Client dev team is also fairly confident that we no longer include server-set props in Foo/set{create}

rsto:
Empirically tested to not be required anymore.

PR #3189, opened 2020-09-22
squatter: add experimental -M and -D options

rsto:
-M allows to define a seqset of UIDs to consider during indexing. Multiple -M options may be specified. The UID sequence set is evaluated across mailboxes.

-D allows to define an alternative directory to generate the activefile and search indexes in.

Both options are experimental and mainly useful for debugging.

Basic tests are at: https://github.com/cyrusimap/cassandane/tree/squatter_indexmsg

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -5247,7 +5247,8 @@ static int _snippet_get(jmap_req_t *req, json_t *filter,
             _snippet_tr_flush,
             _snippet_tr_audit_mailbox,
             _snippet_tr_index_charset_flags,
-            _snippet_tr_index_message_format
+            _snippet_tr_index_message_format,
+            NULL
         },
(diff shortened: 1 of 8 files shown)


brong:
Removing this from fastmail builds due to conflict with #3266

rsto:
Let's rebase and merge that into master, right?

brong:
@rsto - do we still want to go ahead with this?

rsto:
I'd say yes, but only as an undocumented feature. I just let it linger until the patch for long argument options landed.

elliefm:
> Both options are experimental and mainly useful for debugging.

> I'd say yes, but only as an undocumented feature.

I'd say document them anyway, and be explicit that they're experimental and mainly useful for debugging.  It'll save us questions about "what are these undocumented things for" later, and might even net us useful feedback from others experimenting with them.

PR #3192, opened 2020-09-25
Conversations read nocrc

brong:
This allows conversationsdb to skip the CRC32 checks when in shared mode (i.e. reading only).  This is because we have users with millions of emails where over 50% of the CPU usage is just crc32 despite us using a really efficient one - so this skips doing those checks when it's just a read, while still having the checks when we're doing writes.

### DIFF
--- a/lib/cyrusdb_twoskip.c
+++ b/lib/cyrusdb_twoskip.c
@@ -513,13 +513,16 @@ static int read_header(struct dbengine *db)
 
     crc = ntohl(*((uint32_t *)(BASE(db) + OFFSET_CRC32)));
 
+    db->end = db->header.current_size;
+
+    if ((db->open_flags & CYRUSDB_NOCRC))
+        return 0;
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):
Looks good to me

lib/cyrusdb_twoskip.c
```diff
@@ -513,14 +513,17 @@ static int read_header(struct dbengine *db)
 
     crc = ntohl(*((uint32_t *)(BASE(db) + OFFSET_CRC32)));
 
+    db->end = db->header.current_size;
```
wolfsage, line 516: Does it matter that we're now setting `db->end` above an error return but we don't do that for any other error return in this function?

brong: Pretty sure it's fine - if an error happens here we're going to be aborting the whole database open.

Issue #2843, opened 2019-08-06
Cyrus may not send iMIP message with STATUS:CANCELLED and Content-Type: text/calendar;  method=REQUEST


dilyanpalauzov:
When the orginizer PUTs a VEVENT with changed STATUS to CANCELLED, cyrus generates a message containing both `Content-Type: text/calendar; charset=utf-8; method=REQUEST; component=VEVENT` and `STATUS:CANCELLED`.

This violates RFC 5546, which permits STATUS:CANCELLED only with method=CANCEL and permits method=REQUEST only with STATUS: TENTATIVE/CONFIRMED/(absent status)

dilyanpalauzov:
When a VEVENT with STATUS:CONFIRMED is HTTP DELETED, cyrus sends iMIP message with `Content-Type: text/calendar; charset=utf-8; method=CANCEL; component=VEVENT` and `STATUS:CONFIRMED`.  If STATUS is present with method=CANCEL, then STATUS must be CANCELLED.

ksmurchison:
Fixed by https://github.com/cyrusimap/cyrus-imapd/commit/207b8b643f270e8dc97ade4a8c686c1bea772077

elliefm:
cherry-picked to 3.2 as faeb70665a

PR #3194, opened 2020-09-25
http_caldav_sched.c: Properly handle STATUS:CANCELLED

ksmurchison:
This PR does two things:

1. When STATUS:CANCELLED is set by the organizer, METHOD:CANCEL iTIP messages are sent to attendees rather than METHOD:REQUEST iTIP messages

2. iTIP replies to STATUS:CANCELLED instances are ignored

Both of these issues were discovered by @rjbs via some manual testing.

With this PR applied, all Cass tests still pass as usual.  I have a new test for #1.  Working on a test for 2

### DIFF
--- a/imap/http_caldav_sched.c
+++ b/imap/http_caldav_sched.c
@@ -1399,7 +1399,7 @@ static const char *deliver_merge_reply(icalcomponent *ical,
                                        icalcomponent *reply)
 {
     struct hash_table comp_table;
-    icalcomponent *comp, *itip;
+    icalcomponent *comp, *itip, *master = NULL;
     icalcomponent_kind kind;
     icalproperty *prop, *att;
     icalparameter *param;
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
See also https://github.com/cyrusimap/cyrus-imapd/issues/2843: when an event is HTTP DELETEd, the sent iTIP message has METHOD:CANCEL with the original STATUS (e.g. CONFIRMED).  Per RFC 5545, if STATUS is present with METHOD:CANCEL, then STATUS must be CANCEL.

brong (approved):
This all looks right, glad it's been found!  Tests would be excellent.  There's already a ton of tests in the Cassandane CalDAV module, but they didn't do overridden recurrences :(

elliefm:
Cherry-picked to 3.2 as 2fbb141f99

PR #3206, opened 2020-09-29
httpd.h: remove useless rfc3339date_gen declaration

dilyanpalauzov:


elliefm:
Ah, the actual rfc3339date_gen() function was removed by f9e97588e4.

Thanks :)

PR #3214, opened 2020-10-01
message.c: allow 8-bit chars in MIME parameter values

ksmurchison:
We have seen email messages with a Content-Type: name= parameter value which is unencoded UTF8(?)  This just blindly allows these chars.  We could also properly encode them as UTF8 as we build the MIME structure but then the IMAP envelope wouldn't match the raw message.

### DIFF
--- a/imap/message.c
+++ b/imap/message.c
@@ -1438,7 +1438,11 @@ static void message_parse_params(const char *hdr, struct param **paramp)
                     }
                     else break;
                 }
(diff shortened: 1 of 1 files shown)


rsto:
Did you test this in both IMAP and JMAP? I just looked at the Email/get code and it seems to handle unencoded parameters, but it doesn't seem to be covered yet in the JMAPEmail.test_email_get_attachment_name test.

PR #3219, opened 2020-10-02
Allow re-evaluating fuzzy matches against actual message data

ksmurchison:


imap/search_expr.c
```diff
@@ -2515,7 +2515,7 @@ EXPORTED void search_attr_init(void)
             (void *)message_get_subject
         },{
             "to",
-            SEA_FUZZABLE,
+            SEA_FUZZABLE|SEA_UNFUZZ,
```
brong, line 2518: Presumably we should be doing "From" here as well.

imap/jmap_mail.c
```diff
@@ -3278,6 +3297,40 @@ static int guidsearch_expr_eval(struct conversations_state *cstate,
...
+                char *mboxname = NULL;
+                msgrecord_t *mr = NULL;
+                message_t *msg = NULL;
+                uint32_t uid;
+                int ret = 0;
+
+                message_guid_decode(&guid, match->guidrep);
+                jmap_set_emailid(&guid, email_id);
+
+                if (!jmap_email_find(e->req, email_id, &mboxname, &uid) &&
```
brong, line 3315: This is a significant extra expense to these queries.

brong (commented):
Unfortunately this looks technically correct, but too expensive :(  The VIPs function uses exactly these fields, and requires that the searches are fast.

We're going to have to reindex again.  Hopefully we can do it more cheaply than reindexing everything with some kind of clever "parse the cache record and just index from that".

imap/jmap_mail.c
ksmurchison: Yes, but I see no way around it with the current indexing.  Maybe limiting this to only when the search string contains '@' but if we're going to re-index then it doesn't matter.

Is there anyway to implement prefix, suffix, and complete string matches via Xapian?  If not, we're stuck doing the above for those matches.

Issue #2410, opened 2018-06-25
specialusealways: 0? /  SPECIAL-USE with simple LIST


dilyanpalauzov:
[RFC 6415](https://tools.ietf.org/html/rfc6154) says: Section 5.1 Example of an IMAP LIST Command

> C: t1 LIST "" "%"
> S: * LIST (\Sent \HasNoChildren) "/" SentMail

Section 5.2 Example of an Extended IMAP LIST Command
> C: t2 LIST "" "%" RETURN (SPECIAL-USE)
> S: * LIST (\Sent) "/" SentMail

Note how \Sent is returned also in the non-extended LIST.  Cyrus 3.0 does:
> C: t1 LIST "" "%"
> S: * LIST (\HasNoChildren) "/" Sent
> S: * LIST (\HasNoChildren) "/" Spam

But
> C: t1 LIST "" "%" RETURN (SPECIAL-USE)
> S: * LIST (\HasNoChildren \Sent) "/" Sent
> S: * LIST (\HasNoChildren \Junk) "/" Spam

Cyrus IMAPD 3.0 should have returned \Sent and \Junk in the non-extended LIST.  Gnome Evolution is supposed to understand SPECIAL-USE, but does not with Cyrus by default, so probably it relies on the simple LIST.

Why is `specialusealways` by default `off` and hence by default does not implement the RFC correctly?

brong:
On Tue, Jun 26, 2018, at 08:14, Дилян Палаузов wrote:
> RFC 6415[1] says: Section 5.1 Example of an IMAP LIST Command



I assume you mean 6154.

>> C: t1 LIST "" "%"
>>  S: * LIST (\Sent \HasNoChildren) "/" SentMail


> Section 5.2 5.2.  Example of an Extended IMAP LIST Command


>> C: t2 LIST "" "%" RETURN (SPECIAL-USE)
>>  S: * LIST (\Sent) "/" SentMail


> Note how \Sent is returned also in the non-extended LIST.  



Yes, though that is optional.  The examples in the RFC show a server
which returns the special-use information without the RETURN option
being specified, but they don't require it.
> 
> Cyrus 3.0 does:
>> C: t1 LIST "" "%"
>>  S: * LIST (\HasNoChildren) "/" Sent
>>  S: * LIST (\HasNoChildren) "/" Spam


> But


>> C: t1 LIST "" "%" RETURN (SPECIAL-USE)


[...]

PR #3221, opened 2020-10-03
Clarifications specialusealways: 1

dilyanpalauzov:
As of RFC 6154 Section 2 an IMAP server that supports the SPECIAL-USE extension MAY include any or all of the following attributes in responses to the non-extended IMAP LIST command…

The option `specialusealways` alters exactly the behaviour of the unextended LIST command.  Describing the behaviour of `specialusealways` in terms of the undocumented XLIST command is suboptimal.

https://github.com/cyrusimap/cyrus-imapd/issues/2410

### DIFF
--- a/docsrc/imap/rfc-support.rst
+++ b/docsrc/imap/rfc-support.rst
@@ -608,8 +608,8 @@ The following is an inventory of RFCs supported by Cyrus IMAP.
 
     .. NOTE::
 
-        The LIST and LSUB commands return the special-use flags, unless the
-	``specialusealways`` configuration variable is explicitly turned off.
+        The unextended LIST and LSUB commands return the special-use flags, unless
+        the ``specialusealways`` configuration variable is explicitly turned off.
 
 :rfc:`6203`
 
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -2853,8 +2853,8 @@ product version in the capabilities
    or renamed to have a different parent. Default is the built-in list*/
 
 { "specialusealways", 1, SWITCH, "3.1.1" }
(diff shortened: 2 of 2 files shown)


docsrc/imap/rfc-support.rst
```diff
@@ -604,8 +604,8 @@ The following is an inventory of RFCs supported by Cyrus IMAP.
 
     .. NOTE::
 
-        The LIST and LSUB commands return the special-use flags, unless the
-	``specialusealways`` configuration variable is explicitly turned off.
+        The unextended LIST and LSUB commands return the special-use flags, unless
+        the `specialusealways`` configuration variable is explicitly turned off.
```
elliefm, line 608: Lost a backtick here:

```suggestion
        the ``specialusealways`` configuration variable is explicitly turned off.
```

lib/imapoptions
```diff
@@ -2542,8 +2542,8 @@ product version in the capabilities
    or renamed to have a different parent. Default is the built-in list*/
 
 { "specialusealways", 1, SWITCH, "3.1.1" }
-/* If enabled, this option causes LIST and LSUB output to always include
-   the XLIST "special-use" flags */
+/* If enabled, this option causes unextended LIST and LSUB output to
+   include the "special-use" flags */
```
elliefm, line 2546: This now conflicts because a "." was added at the end at some point

elliefm (changes requested):
Thanks.  This seems reasonable, but can you please rebase on current master to fix the imapoptions conflict, and fix the missing backtick in rfc-support.rst

elliefm (approved):

PR #3229, opened 2020-10-07
Sync cache

brong:
Cache mailbox state and use the cache where possible!


ksmurchison (approved):
This all looks sane to me

PR #3233, opened 2020-10-07
xapian_wrap.cpp: index full addr-spec as a single value and use it fo…

ksmurchison:
…r localpart@domain queries

### DIFF
--- a/imap/xapian_wrap.cpp
+++ b/imap/xapian_wrap.cpp
@@ -1729,11 +1729,12 @@ static Xapian::Query *query_new_email(const xapian_db_t *db,
     // query localpart@domain (ONLY if no wildcards)
     if ((atsign > str) && atsign[1]) {
         struct address *addr = NULL;
-        char *a = NULL;
 
         parseaddr_list(str, &addr);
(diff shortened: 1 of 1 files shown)


imap/xapian_wrap.cpp
```diff
@@ -1718,6 +1726,33 @@ static Xapian::Query *query_new_email(const xapian_db_t *db,
         q |= db->parser->parse_query(str, qpflags, prefix);
     }
 
+    // query localpart@domain
+    if ((atsign > str) && atsign[1]) {
+        struct address *addr = NULL;
+        char *a = NULL;
+
+        parseaddr_list(str, &addr);
+        if (addr && (a = address_get_all(addr, /*canon_domain*/1))) {
```
brong, line 1735: Per discussions on slack - we need to detect if there's a wildcard in the address, and if so fall back to the 'D' and 'L' keys .  Only exact matches should use the 'A' key.

brong (changes requested):

imap/xapian_wrap.cpp
ksmurchison: Resolved with the check below

brong (approved):
Yep, that looks good!

PR #3237, opened 2020-10-09
sieve: 'include :once' means 'include :optional' and vice versa

dilyanpalauzov:
as opposed to 'include :global', which is decompiled to ':personal', but interpreted as ':global'.

Does not need recompiling the sieve scripts, as only the interpretation of the bytecode is adjusted.


ksmurchison:
Thanks for spotting this bug.  I fixed it a little differently with
https://github.com/cyrusimap/cyrus-imapd/commit/d126b8e5a4e640cc22544d4d062f7217657ca923

Issue #3006, opened 2020-03-10
Is cyrus-timezones required for 3.2?


dilyanpalauzov:
configure.ac contains `${CYRUS_TIMEZONES_CFLAGS}`, but I do not see where the cyrus-timezones.pc file is read.

In docsrc/imap/developer/compiling.rst cyrus-timezones is also not mentioned.

So the question is, why does cyrus-timezones exist.

elliefm:
We recently patched the cyrus-timezones detection out of configure.ac (discussion was here: https://github.com/cyrusimap/cyrus-timezones/pull/6). Looks like I missed a line though, thanks for catching it, I'll remove that too.

It's not mentioned in cyrus-imapd docs because it's not required.

It still exists in case it's useful to people, but I don't know if it's useful to people.

dilyanpalauzov:
tools/vzic was removed from cyrus-imapd.git in 2017.  docsrc/imap/download/installation/http/caldav.rst says today:
```
Administration
--------------

This module is designed to use the IANA Time Zone Database data (a.k.a. Olson Database) converted to the iCalendar format.

Cyrus uses a modified `vzic <https://github.com/libical/vzic>`_ to convert IANA formatted data into iCalendar format. There is more information on Cyrus vzic in ``tools/vzic/README``.
```
The IANA Database, https://data.iana.org/time-zones/releases/tzdb-2019c.tar.lz, says in tz-link.html that vzic can be downloaded from https://sourceforge.net/projects/vzic/ .

So, is the modified vzic used the one offered by github.com/libical/vzic, that contains changes towards https://sourceforge.net/projects/vzic/ , or does Cyrus use a vzic, which is different from the one of github.com/libical/vzic ?

The documentation shall not reference tools/vzic/README, as this file does not exist.

elliefm:
Oh awesome, I didn't realise that stuff was even documented.  I've updated it to refer to the new zoneinfo_dir setting and to link to the cyrus-timezones package.

dilyanpalauzov:
Why does fastmail make changes to two different vzic: one within libical and one within cyrus-timezones?  The vzics are currently not identical.

dilyanpalauzov:
lib/imapoptions says tzdist and caldav are two distinct http modules.  It amends that zoneinfo_dir and zoneinfo_db_path are needed only by the tzdist module.

docsrc/imap/download/installation/caldav.rst says, that zoneinfo_dir is required by the caldav module.

[...]

PR #3238, opened 2020-10-11
Clarifications on tzdist setup: the vendor prefix in TZID must be the empty string

dilyanpalauzov:
This commit mentions copying of the leap-seconds.list file, that is needed at runtime and by ctl_zoneinfo.

Moreover it elaborates on the TZID vendor prefixes.  My experience is, that Cyrus recognizes only the timezones, if the TZID contains no vendor prefix.  To feed no vendor prefix to zoneinfo.db vzic must be compiled with TZID_PREFIX set to '' in Makefile.  This is by default done in cyrus-timezones.

Moreover, creating zoneinfo_dir/ with vendor prefixes and calling ctl_zoneinfo does grow zoneinfo.db . Doing afterwards the “corect actions” does not shrink zoneinfo.db.

I tried creating zoneinfo.db with and without vendor prefix, then uploading VEVENTS and checking if the timezone is stripped from the event.  It is stripped if the timezone is available without vendor prefix in the database.

I do not pretend to understand how vendor prefixes or libical works, these are empirical results.

I hope some experts on these topics will review this in reasonable time and clarify if cyrus does require stripped vendor prefixes.

See also https://github.com/cyrusimap/cyrus-imapd/issues/3006.


elliefm:
I thought there was a way to tell Cyrus which vendor prefix is in use in the timezone data, but I don't remember and am probably mistaken.  Last time I looked at this much was when refactoring the handling of `zoneinfo_dir` option.  I also don't know libical, and just figure things out empirically as needed.

That point about the leap-seconds.list file is interesting.  What's the behaviour if it's missing?

dilyanpalauzov:
Missing leap-seconds.list leads to a syslog() message at run time.  Does ctl_zoneinfo integrate the content of leap-seconds in zoneinfo.db, so that it is needed either at run-time by httpd or during the compilation by ctl_zoneinfo?

elliefm:
> My experience is, that Cyrus recognizes only the timezones, if the TZID contains no vendor prefix.

Oh, I think this might be because of this line:
https://github.com/cyrusimap/cyrus-imapd/blob/82b47f1533ceb3730e4d8c2c8562eb6ed270cbe1/imap/ical_support.c#L74

I guess we might make that prefix configurable at some point, but for now it's fine to just document that it must be empty.

> Missing leap-seconds.list leads to a syslog() message at run time.

I can't find a leap-seconds.list on my own system (using the cyrus-timezones package), but I'm not sure what syslog message to look for when it's missing?

Anyway I think these documentation updates are fine, at least as far as they're more specific than what was there, though I don't know enough to tell if they're correct.  @ksmurchison do you have any feedback?

dilyanpalauzov:
leap-seconds.list  is part of the Olson database.  The syslog message is:
```c
imap/zoneinfo_db.h:52:#define FNAME_LEAPSECFILE "/leap-seconds.list"
--
imap/ctl_zoneinfo.c:158:        snprintf(buf, sizeof(buf), "%s%s", zoneinfo_dir, FNAME_LEAPSECFILE);
imap/ctl_zoneinfo.c-159-        if (verbose) printf("Processing leap seconds file %s\n", buf);
imap/ctl_zoneinfo.c-160-        if (!(fp = fopen(buf, "r"))) {
imap/ctl_zoneinfo.c-161-            fprintf(stderr, "Could not open leap seconds file %s\n", buf);
--
imap/http_tzdist.c:603:    snprintf(buf, sizeof(buf), "%s%s", zoneinfo_dir, FNAME_LEAPSECFILE);
imap/http_tzdist.c-604-    if (!(fp = fopen(buf, "r"))) {
imap/http_tzdist.c-605-        syslog(LOG_ERR, "Failed to open file %s", buf);
imap/http_tzdist.c-606-        return;
```

elliefm:
Ahh I see -- the error doesn't show up for me on casual examination because I don't enable the tzdist module or run the ctl_zoneinfo command, but I think it would if I did because I don't have the file.

dilyanpalauzov:
When the tzdist module is enabled, Cyrus strips the VTIMEZONEs from the uploaded data, irrespective of the tzid-by-reference support in the CUA.  If another CUA downloads the data, it does not get the VTIMEZONEs included.  If this workflow works without any problems, it would essentially mean, that on upload any CUA can always refrain from sending the tz data, despite what rfc5545 says and irrespective of caldav-tz-by-reference support.

To verify the thesis, there shall be many servers stripping the timezone data on upload, which can be done, once the documentation on the tzdist setup process is improved.

See also https://forums.bitfire.at/topic/2401.

dilyanpalauzov:
I changed also
```diff
diff --git a/docsrc/imap/download/installation/http/caldav.rst b/docsrc/imap/download/installation/http/caldav.rst
index 91d6bc0ab..7fd7f05f2 100644
--- a/docsrc/imap/download/installation/http/caldav.rst
+++ b/docsrc/imap/download/installation/http/caldav.rst
@@ -329,9 +329,8 @@ the time zone information up to date then falls upon the Cyrus administrator.
 TZDist is optional: without Cyrus having TZDist enabled, calendar clients should
 still be able to get their timezone information from their client or their OS.
 
-TZDist is also required if you want the CalDAV server to strip known VTIMEZONEs
-from incoming iCalendar data (as advertised by the ``calendar-no-timezone`` DAV
-option from :rfc:`7809`).
+TZDist strips known VTIMEZONEs from incoming iCalendar data (as
+advertised by the ``calendar-no-timezone`` DAV option from :rfc:`7809`).
 
 Configuration
 -------------
```
to outline, that stripping known time zones is unconditional, once TZDist is enabled.

dilyanpalauzov:
Can this case make some progress, please?

rsto:
Thanks. This looks all sane to me, but @ksmurchison has a final say on that. There are no plans to make the timezone prefix configurable and it looks to me as if this does not any cause issues for you, @dilyanpalauzov? Otherwise, please let me know.

dilyanpalauzov:
Long time ago I had issued with the vendor prefix, with the /freeassociation.sourceforge.net/ prefix, with [icaltzutil_fetch_timezone() is always returning NULL, when libical is configured with -DUSE_BUILTIN_TZDATA=ON]( https://github.com/libical/libical/issues/466).  Since a lot of time passed, I do not remember the details.

ksmurchison (approved):
I think we always want the prefix to be empty for Cyrus

PR #3243, opened 2020-10-13
timsieved:actions_setuser(): sieved_userid can be leaked

dilyanpalauzov:
Upon `UNAUTHENTICATE` and another `AUTHENTICATE` the variable sieved_userid is set to a new value and the old value is leaked.

In timsieved/actions.c:actions_setuser() there is no need for local variable ‘user’, which does the same as the static variable ‘sieved_userid’.

### DIFF
--- a/timsieved/actions.c
+++ b/timsieved/actions.c
@@ -77,8 +77,7 @@ extern int sieved_userisadmin;
 extern sieve_interp_t *interp;
 
 static char *sieve_dir_config = NULL;
-
-static const char *sieved_userid = NULL;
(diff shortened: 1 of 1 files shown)


ksmurchison (changes requested):
I think sieved_userid should still be initialized to NULL, so the free() doesn't crash

dilyanpalauzov:
Aren’t `static` variables implicitly zeroed?

ksmurchison:
Probably, but the explicit assignment won't hurt

PR #3244, opened 2020-10-14
fix up locking for conversations DB

brong:
This puts a user namespace lock into either conversations.db or mailbox, whichever opens first!  Since that blocks either of the other from opening, it means we can't have lock ordering problems, I hope!

The most complex bit is in mailbox_lock_index(), where the lock could have been dropped, because mailbox_unlock_index() drops it to mean long running IMAP fetches don't block other actions.  This means that if there's no namespacelock, we need to drop everything and relock.  There's already logic for that when changing locktype, so I hooked this in.  Pretty sure it's correct.


ksmurchison (approved):
This all looks sane

PR #3245, opened 2020-10-15
docs: update links to CMU lists to point to Topicbox

rjbs:


docsrc/imap/support/feedback-mailing-lists.rst
```diff
@@ -4,80 +4,29 @@
 Mailing Lists
 =============
 
+Cyrus is discussed at a few official mailing lists hosted by
+`Topicbox <https://cyrus.topicbox.com/>` — which is, itself, powered by Cyrus!
```
elliefm, line 8: This link is missing the magical `_` after the closing "`", which means instead of a link to topicbox, we get an inline code block containing "Topicbox <https://cyrus.topicbox.com/>" in italicised monospace 😅 (RST!)

elliefm (changes requested):
Do you wanna amend and force push this so that the missing "_" never happened?

I would just do it myself locally, but I also wanted to ask: do we want to somehow include the email address for these lists on the page?  Dave included them in his emails to the lists, but someone who wasn't already on the list or didn't read Dave's emails closely might mistakenly think that Topicbox is some "web only" thing and get wound up about it, and I'd rather forestall that if possible

elliefm:
Decided to just make the one character fix myself, and merged it directly.

82b47f1533 on master
e584a185a8 on 3.2
98eca6b9c8 on 3.0
1a120841a3 on 2.5

PR #3249, opened 2020-10-16
configure.ac: permit explicitly turning off brotli and zstd

dilyanpalauzov:
* in the summary at the end print the wslay and nghttp2 status, even when --with-http was not passed to ./configure

* remove AC_MSG_NOTICEs.  The summary at the end has the purpose to substitute these.

* when zstd is disabled, do not cause any run-time overhead.  The rationale is that no http client (CUA) uses Zstandard currently.

Besides, when httpd runs behind a reverse proxy, the proxy server can take care of the compression, which it does any way for other stuff, so this change makes httpd lighter.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -355,7 +355,7 @@ dnl which means if your sqlite is in a non-default location and you need httpd
 dnl or object storage etc, you're suddenly also supporting it as a cyrusdb
 dnl backend, possibly unexpectedly.
 AC_ARG_WITH(sqlite,
-        [AS_HELP_STRING([--with-sqlite=DIR], [use SQLite (in DIR) [no]])],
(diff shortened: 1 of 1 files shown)


ksmurchison (changes requested):
We discussed this PR on our Cyrus call today and while we are fine with the configure changes, we prefer to not populate the code with a bunch of #ifdef , which is why we have stubs for the compression functions.  We probably should simply flag these stub functions as inline.

dilyanpalauzov:
Have you considered, that the ifdefs in fact reduce the total amount of lines?

ksmurchison:
Yes, but for a minimal gain at the expense of cluttering the code.

PR #3252, opened 2020-10-19
mailbox.c IOERROR xsyslog updates

elliefm:
I'm trying to keep these changes in small batches, cause they need very close review (we have no test coverage for most of our errors), and the changes are all very similar and eye-glazey.

Here's the updates for the IOERRORs in mailbox.c.  I also added a format attribute to the xsyslog definition, like our other printf-alike functions, so we get a bit of compiler noise for the class of mistakes the compiler can detect.

This would benefit from a compile test, in case your compiler has newer/better warnings than mine.

### DIFF
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -473,17 +473,22 @@ static int cache_parserecord(struct mappedfile *cachefile, uint64_t cache_offset
 
         /* bounds checking */
         if (offset >= buf_size) {
-            syslog(LOG_ERR, "IOERROR: offset greater than cache size "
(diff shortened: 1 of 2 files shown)


ksmurchison (commented):
Do we really want to lose the %m string?

Otherwise, this compiles cleanly on my system

elliefm:
We don't; xsyslog() includes it automatically if errno is non-zero, so we just don't need to ask for it explicitly anymore.

elliefm:
(fixed conflicts from the conversations locking changes)

PR #3253, opened 2020-10-19
Jmap name safety

brong:
I've got Cassandane tests for this over at:

https://github.com/cyrusimap/cassandane/pull/97


ksmurchison (approved):
This looks sane and still passes existing Cass tests

PR #3256, opened 2020-10-20
Cyrusdb lock

brong:
This isn't actually needed for anything right now, so I separated it out into its own PR.

I think it's a useful API to have still.

### DIFF
--- a/lib/cyrusdb.c
+++ b/lib/cyrusdb.c
@@ -335,6 +335,17 @@ EXPORTED int cyrusdb_delete(struct db *db,
     return db->backend->delete_(db->engine, key, keylen, tid, force);
 }
 
+EXPORTED int cyrusdb_lock(struct db *db, struct txn **tid, int flags)
+{
+    if (!db->backend->lock)
+        return CYRUSDB_NOTIMPLEMENTED;
+#ifdef DEBUGDB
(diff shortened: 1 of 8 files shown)


elliefm:
So if I'm reading this right, the point is to be able to explicitly request a lock/new transaction, whereas previously this detail was buried in how you interacted with the various functions' **tid parameter?  Yeah I like it.

I wonder if it's worthwhile eventually removing the implicit create-transaction behaviour from the other functions, so that all call sites have to explicitly lock?  Might make it a bit easier on the casual reader?

I spent a while wondering why there was not a corresponding unlock, until I read a bit wider than just the diff, and realised that commit/abort are already explicit about ending the transaction and invalidating tid, so "unlock" would just be a synonym for abort.

brong:
Yep - and yes, I like the idea of a separate lock request and being able to explicitly say "I only want a shared lock here, I promise not to write".

elliefm (commented):
Not sure why I only commented and didn't approve this at the time, but it seems fine on a re-read.  Possibly bitrotted by now though!  Also, CI has never run on it, and I can't seem to kick off a CI run manually.

If you still think this is useful, can you please rebase it on current master and force push.  The rebase will tell us if the patches still apply cleanly, and the push will trigger a CI run, and if that's all good, then we're good.

elliefm (approved):

PR #3257, opened 2020-10-20
charset.c: properly QP encode address-list headers

ksmurchison:
This resolves a customer issue where they edit a To: header which contains a character that needs encoding.  We can't just blindly QP encode the entire header without breaking it for clients.  This patch encodes only the display-name portion of name-addrs, and leaves addr-spec, angle-addr, the "," separating the addresses, and any surrounding whitespace alone.

### DIFF
--- a/cunit/charset.testc
+++ b/cunit/charset.testc
@@ -341,6 +341,9 @@ static void test_encode_mimeheader(void)
              "=?UTF-8?Q?01234567890123456789012345678901234567890123456789012345?="
              "\r\n ""=?UTF-8?Q?=E2=82=AC?=");
 
+    /* only encode display-name of address */
+    TESTCASE("\"abc\xe2\x88\x97xyz\" <foo@example.com>", "=?UTF-8?Q?\"abc=E2=88=97xyz\"?= <foo@example.com>");
+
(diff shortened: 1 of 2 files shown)


brong:
OMG that's a lot of magic.   Is there an associated set of testcases to prove that this works safely?

dilyanpalauzov:
See also https://github.com/cyrusimap/cyrus-imapd/pull/2693.

ksmurchison:
Tested with https://github.com/cyrusimap/cassandane/pull/100

rsto:
> Tested with [cyrusimap/cassandane#100](https://github.com/cyrusimap/cassandane/pull/100)

Couldn't this also be tested as a cunit test? Or is it necessary to invoke the test for qp_encode through the sieve paths?

ksmurchison:
It probably could, but the test that I wrote simulates what the use that reported the issue was attempting to do with Sieve.

ksmurchison:
@brong @rsto Can I merge this and the associated Cass fix?

rsto:
Looks good to me. I still would prefer testing it as a `charset.cunit` test, but the Cassandane tests seem to cover the issue.

PR #3260, opened 2020-10-21
jmap_mail: don't use seen_lockread

brong:
We're inside a userlock which will stop anything else writing to it,
but if we're holding the seen lock then the append will fail to addseen.

See test case: https://github.com/cyrusimap/cassandane/pull/98


rsto:
Nice and easy :+1:

Issue #964, opened 2007-09-26
imapd behaves as if normalizeuid is always set to 1 in /etc/imapd.conf


brong:
From: [Dwayne Litzenberger](dwayne@oscl.ca)
Bugzilla-Id: 2994
Version: 2.3.x
Owner: [Ken Murchison](murch@andrew.cmu.edu)

PR #3262, opened 2020-10-22
Format specifier fixes

anatoli26:
@elliefm, @obache, all,

I tried to pay a lot of attention to the details, but please check this PR exhaustively, each change separately, as there are many small details with a single char changing the entire meaning.

@elliefm, also please test-run this PR on as much platforms and OS as possible.

On the other hand, I have some observations:
 * please check the commit msg of the 1st commit in the PR. The `configure.am` checks don't make much sense for a _real_ detection of the specifier, but I proceeded as was suggested by @elliefm.
 * **imap/mailbox.c**, **imap/sync_support.c**, etc.: it was `%lu`, now it is `%ld`/`%lld`, i.e. signed/unsigned change. Not sure why it was unsigned as `time_t` is a signed data type.
 * **imap/mbexamine.c** line 259: why was that cast instead of `time_t`?
 * **master/master.c**: is `PRIu64` the correct specifier?

### DIFF
--- a/cmulocal/cyr_inttype.m4
+++ b/cmulocal/cyr_inttype.m4
@@ -5,20 +5,22 @@
 # DESCRIPTION
 #
 #   Figures out the underlying integer type of TYPE-NAME and the appropriate
-#   printf format string to use for it.
+#   printf format string and strtol-like parse function to use for it.
 #
 #   BUILD-PREREQS is whatever C code is required for this type to be defined.
 #   This is probably a #include statement!
 #
-#   Sets two cache variables:
+#   Sets three cache variables:
 #
 #   * $cyr_cv_type_foo    => the underlying integer type for the type "foo"
 #   * $cyr_cv_format_foo  => the format string to use for type "foo"
+#   * $cyr_cv_parse_foo   => the strtol-like parse function to use for type "foo"
 #
 #   Example:
 #
 #       CYR_INTTYPE([time_t], [#include <time.h>])
 #       AC_DEFINE_UNQUOTED([TIME_T_FMT], ["$cyr_cv_format_time_t"], [...])
+#       AC_DEFINE_UNQUOTED([strtotimet(a,b,c)], [$cyr_cv_parse_time_t(a,b,c)], [...])
 #
 AC_DEFUN([CYR_INTTYPE],[
     dnl First, figure out what type of integer it is, by exploiting the
@@ -68,4 +70,25 @@ AC_DEFUN([CYR_INTTYPE],[
     )
     AS_IF([test "x$AS_TR_SH([cyr_cv_format_$1])" = "xunknown"],
           [AC_MSG_ERROR([Unable to determine printf format string for `$1'])])
+
+    dnl And another quick table lookup to turn the known types into the
+    dnl appropriate strtol-like parse function
+    dnl Note that this cheats a little for int/unsigned int
+    AC_CACHE_CHECK(
+        [strtol-like parse function for `$1'],
+        [AS_TR_SH([cyr_cv_parse_$1])],
+        [
+            AS_CASE([$AS_TR_SH([cyr_cv_type_$1])],
+                ["int"], [eval AS_TR_SH([cyr_cv_parse_$1])=strtol],
+                ["long int"], [eval AS_TR_SH([cyr_cv_parse_$1])=strtol],
+                ["long long int"], [eval AS_TR_SH([cyr_cv_parse_$1])=strtoll],
+                ["unsigned int"], [eval AS_TR_SH([cyr_cv_parse_$1])=strtoul],
+                ["unsigned long int"], [eval AS_TR_SH([cyr_cv_parse_$1])=strtoul],
+                ["unsigned long long int"], [eval AS_TR_SH([cyr_cv_parse_$1])=strtoull],
+                [eval AS_TR_SH([cyr_cv_parse_$1])=unknown]
+            )
+        ]
+    )
+    AS_IF([test "x$AS_TR_SH([cyr_cv_parse_$1])" = "xunknown"],
+          [AC_MSG_ERROR([Unable to determine strtol-like parse function for `$1'])])
 ])
--- a/cmulocal/cyr_inttype.m4
+++ b/cmulocal/cyr_inttype.m4
@@ -1,0 +1,71 @@
+# SYNOPSIS
+#
+#   CYR_INTTYPE([TYPE-NAME], [BUILD-PREREQS])
+#
+# DESCRIPTION
+#
+#   Figures out the underlying integer type of TYPE-NAME and the appropriate
+#   printf format string to use for it.
+#
+#   BUILD-PREREQS is whatever C code is required for this type to be defined.
+#   This is probably a #include statement!
+#
+#   Sets two cache variables:
+#
+#   * $cyr_cv_type_foo    => the underlying integer type for the type "foo"
+#   * $cyr_cv_format_foo  => the format string to use for type "foo"
+#
+#   Example:
+#
+#       CYR_INTTYPE([time_t], [#include <time.h>])
+#       AC_DEFINE_UNQUOTED([TIME_T_FMT], ["$cyr_cv_format_time_t"], [...])
+#
+AC_DEFUN([CYR_INTTYPE],[
+    dnl First, figure out what type of integer it is, by exploiting the
+    dnl behaviour that redefining a variable name as the same type is only
+    dnl a warning, but redefining it as a different type is an error.
+    AC_CACHE_CHECK(
+        [underlying integer type of `$1'],
+        [AS_TR_SH([cyr_cv_type_$1])],
+        [
+            AC_LANG_PUSH([C])
+            saved_CFLAGS=$CFLAGS
+            CFLAGS=-Wno-error
+            found=no
+            for t in "int" "long int" "long long int" \
+                    "unsigned int" "unsigned long int" "unsigned long long int"
+            do
+                AC_COMPILE_IFELSE(
+                    [AC_LANG_PROGRAM([$2], [extern $1 foo; extern $t foo;])],
+                    [AS_TR_SH([cyr_cv_type_$1])=$t; found=yes; break]
+                )
+            done
+            AS_IF([test "x$found" != "xyes"],
+                  [eval AS_TR_SH([cyr_cv_type_$1])=unknown])
+            CFLAGS=$saved_CFLAGS
+            AC_LANG_POP([C])
+        ]
+    )
+    AS_IF([test "x$AS_TR_SH([cyr_cv_type_$1])" = "xunknown"],
+          [AC_MSG_ERROR([Unable to determine underlying integer type of `$1'])])
+
+    dnl Then, a quick table lookup to turn the known types into the
+    dnl appropriate format string.
+    AC_CACHE_CHECK(
+        [printf format string for `$1'],
+        [AS_TR_SH([cyr_cv_format_$1])],
+        [
+            AS_CASE([$AS_TR_SH([cyr_cv_type_$1])],
+                ["int"], [eval AS_TR_SH([cyr_cv_format_$1])=%d],
+                ["long int"], [eval AS_TR_SH([cyr_cv_format_$1])=%ld],
+                ["long long int"], [eval AS_TR_SH([cyr_cv_format_$1])=%lld],
+                ["unsigned int"], [eval AS_TR_SH([cyr_cv_format_$1])=%u],
+                ["unsigned long int"], [eval AS_TR_SH([cyr_cv_format_$1])=%lu],
+                ["unsigned long long int"], [eval AS_TR_SH([cyr_cv_format_$1])=%llu],
+                [eval AS_TR_SH([cyr_cv_format_$1])=unknown]
+            )
+        ]
+    )
+    AS_IF([test "x$AS_TR_SH([cyr_cv_format_$1])" = "xunknown"],
+          [AC_MSG_ERROR([Unable to determine printf format string for `$1'])])
+])
--- a/configure.ac
+++ b/configure.ac
@@ -247,14 +247,20 @@ AC_FUNC_VPRINTF
 CYR_INTTYPE([time_t], [#include <time.h>])
 AC_DEFINE_UNQUOTED([TIME_T_FMT], ["$cyr_cv_format_time_t"],
                    [printf formatter for time_t])
+AC_DEFINE_UNQUOTED([strtotimet(a,b,c)], [$cyr_cv_parse_time_t(a,b,c)],
+                   [strtol-like parse function for time_t])
 
 CYR_INTTYPE([size_t], [#include <stdlib.h>])
 AC_DEFINE_UNQUOTED([SIZE_T_FMT], ["$cyr_cv_format_size_t"],
                    [printf formatter for size_t])
+AC_DEFINE_UNQUOTED([strtosizet(a,b,c)], [$cyr_cv_parse_size_t(a,b,c)],
+                   [strtol-like parse function for size_t])
 
 CYR_INTTYPE([off_t], [#include <sys/types.h>])
 AC_DEFINE_UNQUOTED([OFF_T_FMT], ["$cyr_cv_format_off_t"],
                    [printf formatter for off_t])
+AC_DEFINE_UNQUOTED([strtoofft(a,b,c)], [$cyr_cv_parse_off_t(a,b,c)],
+                   [strtol-like parse function for off_t])
 
 dnl function for doing each of the database backends
 dnl parameters: backend name, variable to set, withval
--- a/configure.ac
+++ b/configure.ac
@@ -262,6 +262,12 @@ AC_DEFINE_UNQUOTED([OFF_T_FMT], ["$cyr_cv_format_off_t"],
 AC_DEFINE_UNQUOTED([strtoofft(a,b,c)], [$cyr_cv_parse_off_t(a,b,c)],
                    [strtol-like parse function for off_t])
 
+CYR_INTTYPE([rlim_t], [#include <sys/resource.h>])
+AC_DEFINE_UNQUOTED([RLIM_T_FMT], ["$cyr_cv_format_rlim_t"],
+                   [printf formatter for rlim_t])
+AC_DEFINE_UNQUOTED([strtorlimt(a,b,c)], [$cyr_cv_parse_rlim_t(a,b,c)],
+                   [strtol-like parse function for rlim_t])
+
 dnl function for doing each of the database backends
 dnl parameters: backend name, variable to set, withval
 
--- a/configure.ac
+++ b/configure.ac
@@ -244,6 +244,18 @@ AC_SUBST(CPPFLAGS)
 AC_SUBST(LOCALDEFS)
 AC_FUNC_VPRINTF
 
+CYR_INTTYPE([time_t], [#include <time.h>])
+AC_DEFINE_UNQUOTED([TIME_T_FMT], ["$cyr_cv_format_time_t"],
+                   [printf formatter for time_t])
+
+CYR_INTTYPE([size_t], [#include <stdlib.h>])
+AC_DEFINE_UNQUOTED([SIZE_T_FMT], ["$cyr_cv_format_size_t"],
+                   [printf formatter for size_t])
+
+CYR_INTTYPE([off_t], [#include <sys/types.h>])
+AC_DEFINE_UNQUOTED([OFF_T_FMT], ["$cyr_cv_format_off_t"],
+                   [printf formatter for off_t])
+
 dnl function for doing each of the database backends
 dnl parameters: backend name, variable to set, withval
 
@@ -2320,27 +2332,6 @@ struct sockaddr_storage {
 
 #ifndef HAVE_SHUTDOWN
 #define shutdown(fd, mode) 0
-#endif
-
-/* *printf() macros */
-#if (SIZEOF_SIZE_T == SIZEOF_INT)
-#define SIZE_T_FMT "%u"
-#elif (SIZEOF_SIZE_T == SIZEOF_LONG)
-#define SIZE_T_FMT "%lu"
-#elif (SIZEOF_SIZE_T == SIZEOF_LONG_LONG_INT)
-#define SIZE_T_FMT "%llu"
-#else
-#error dont know what to use for SIZE_T_FMT
-#endif
-
-#if (SIZEOF_OFF_T == SIZEOF_LONG)
-#define OFF_T_FMT "%ld"
-#define strtoofft(nptr, endptr, base) strtol(nptr, endptr, base)
-#elif (SIZEOF_OFF_T == SIZEOF_LONG_LONG_INT)
-#define OFF_T_FMT "%lld"
-#define strtoofft(nptr, endptr, base) strtoll(nptr, endptr, base)
-#else
-#error dont know what to use for OFF_T_FMT
 #endif
 
 #ifndef HAVE_POSIX_FADVISE
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -451,7 +451,7 @@ static int update_alarmdb(const char *mboxname,
     if (!alarmdb) return -1;
     int rc = SQLITE_OK;
 
-    syslog(LOG_DEBUG, "update_alarmdb(%s:%u, %ld)",
+    syslog(LOG_DEBUG, "update_alarmdb(%s:%u, " TIME_T_FMT ")",
            mboxname, imap_uid, nextcheck);
 
     if (nextcheck)
@@ -603,7 +603,7 @@ static int write_lastalarm(struct mailbox *mailbox,
            mailbox->name, record->uid);
 
     if (data) {
-        buf_printf(&annot_buf, "%ld %ld", data->lastrun, data->nextcheck);
+        buf_printf(&annot_buf, TIME_T_FMT " " TIME_T_FMT, data->lastrun, data->nextcheck);
     }
     syslog(LOG_DEBUG, "data: %s", buf_cstring(&annot_buf));
 
@@ -632,7 +632,7 @@ static int read_lastalarm(struct mailbox *mailbox,
                             annotname, "", &annot_buf);
 
     if (annot_buf.len &&
-        sscanf(buf_cstring(&annot_buf), "%ld %ld",
+        sscanf(buf_cstring(&annot_buf), TIME_T_FMT " " TIME_T_FMT,
                &data->lastrun, &data->nextcheck) == 2) {
         r = 0;
     }
--- a/imap/http_admin.c
+++ b/imap/http_admin.c
@@ -602,7 +602,7 @@ static int action_proc(struct transaction_t *txn)
         char buf[1024];
 
         while (fgets(buf, sizeof(buf), f)) {
-            if (sscanf(buf, "btime %ld\n", &boot_time) == 1) break;
+            if (sscanf(buf, "btime " TIME_T_FMT "\n", &boot_time) == 1) break;
             while (buf[strlen(buf)-1] != '\n' && fgets(buf, sizeof(buf), f)) {
             }
         }
--- a/imap/http_admin.c
+++ b/imap/http_admin.c
@@ -255,7 +255,7 @@ static int action_murder(struct transaction_t *txn)
        and the config file size/mtime */
     assert(!buf_len(&txn->buf));
     stat(config_filename, &sbuf);
-    buf_printf(&txn->buf, "%ld-%ld-%ld", (long) compile_time,
+    buf_printf(&txn->buf, TIME_T_FMT "-" TIME_T_FMT "-" OFF_T_FMT, compile_time,
                sbuf.st_mtime, sbuf.st_size);
 
     message_guid_generate(&guid, buf_cstring(&txn->buf), buf_len(&txn->buf));
@@ -341,7 +341,7 @@ static int action_menu(struct transaction_t *txn)
      * Extend this to include config file size/mtime if we add run-time options.
      */
     assert(!buf_len(&txn->buf));
-    buf_printf(&txn->buf, "%ld", (long) compile_time);
+    buf_printf(&txn->buf, TIME_T_FMT, compile_time);
     message_guid_generate(&guid, buf_cstring(&txn->buf), buf_len(&txn->buf));
     etag = message_guid_encode(&guid);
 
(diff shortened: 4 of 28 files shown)


configure.ac
```diff
@@ -2333,16 +2333,26 @@ struct sockaddr_storage {
...
+#elif (SIZEOF_OFF_T == SIZEOF_LONG)
+#define OFF_T_FMT "%ld"
+#define strtoofft(nptr, endptr, base) strtol(nptr, endptr, base)
 #else
 #error dont know what to use for OFF_T_FMT
 #endif
 
+#if (SIZEOF_TIME_T == SIZEOF_LONG_LONG_INT)
+#define TIME_T_FMT "%lld"
+#define strtoofft(nptr, endptr, base) strtoll(nptr, endptr, base)
```
obache, line 2348: should not be defined here `time_t` case , defined above `off_t` case.

obache: or define `strtotimet` and change to use it for scanning `time_t`, for example, `parse_zoneinfo`
 https://github.com/cyrusimap/cyrus-imapd/blob/95cc49f06b9794012ddba8879f1a842374281b9a/imap/zoneinfo_db.c#L140
(already fixed for `zoneinfo_store` case)

anatoli26: @obache thanks! Fixed by removing the duplicate lines (I've altered the 1st commit and force-pushed it).

lib/auth_pts.c
```diff
@@ -400,7 +401,7 @@ static int ptload(const char *identifier, struct auth_state **state)
 
         syslog(LOG_DEBUG,
                "ptload(): fetched cache record (%s)" \
-               "(mark %ld, current %ld, limit %ld)", identifier,
+               "(mark " TIME_T_FMT ", current %" PRIu64 ", limit %" PRIu64 ")", identifier,
```
elliefm, line 404: I think all three of these should be TIME_T_FMT, not PRIu64? Note that at line 398/399 `now` is declared as `time_t`, and the printout of "current" and "limit" are calculated from `now`.

imap/http_ischedule.c
```diff
@@ -222,7 +222,7 @@ static int meth_get_isched(struct transaction_t *txn,
     /* Generate ETag based on compile date/time of this source file,
        the number of available RSCALEs and the config file size/mtime */
     assert(!buf_len(&txn->buf));
-    buf_printf(&txn->buf, "%ld-%zu-%ld-%ld", (long) compile_time,
+    buf_printf(&txn->buf, TIME_T_FMT "-%zu-" TIME_T_FMT "-" OFF_T_FMT, compile_time,
```
elliefm, line 225: The %zu should probably also be changed to SIZE_T_FMT here

imap/caldav_alarm.c
```diff
@@ -632,7 +632,7 @@ static int read_lastalarm(struct mailbox *mailbox,
                             annotname, "", &annot_buf);
 
     if (annot_buf.len &&
-        sscanf(buf_cstring(&annot_buf), "%ld %ld",
+        sscanf(buf_cstring(&annot_buf), TIME_T_FMT " " TIME_T_FMT,
```
elliefm, line 635: Hmmm, does it make sense to use these format specifiers for `sscanf` conversions?  I'm not sure.  @ksmurchison, any thoughts?

imap/http_admin.c
```diff
@@ -602,7 +602,7 @@ static int action_proc(struct transaction_t *txn)
         char buf[1024];
 
         while (fgets(buf, sizeof(buf), f)) {
-            if (sscanf(buf, "btime %ld\n", &boot_time) == 1) break;
+            if (sscanf(buf, "btime " TIME_T_FMT "\n", &boot_time) == 1) break;
```
elliefm, line 605: (ditto re `sscanf`)

elliefm (commented):
This looks okay to me by eye.  I'm about to do a local build and test run and see how that goes, will report back in a little bit.

elliefm:
@guimard Is there any way we could try this out on the Debian build farm, to see what pops up on the various platforms it tests?

lib/auth_pts.c
elliefm: Yeah, they do need to be TIME_T_FMT.  Build fails on linux/gcc/x86_64 with PRIu64's here.

configure.ac
```diff
@@ -2333,16 +2333,24 @@ struct sockaddr_storage {
...
 #endif
 
+#if (SIZEOF_TIME_T == SIZEOF_LONG_LONG_INT)
+#define TIME_T_FMT "%lld"
+#elif (SIZEOF_TIME_T == SIZEOF_LONG)
+#define TIME_T_FMT "%ld"
+#else
+#error dont know what to use for TIME_T_FMT
+#endif
+
```
elliefm, line 2353: Annoyingly, changing the order here breaks the linux/gcc/x86_64 build!

I think because on this platform, SIZEOF_OFF_T == SIZEOF_LONG == SIZEOF_LONG_LONG_INT, so whichever of those conditions we put first is the one that matches, but it _really_ wants `%ld`, and complains if it gets `%lld`, even though they're the same size...

elliefm: (Same applies for TIME_T_FMT too: it really wants %ld, but all three types are the same size, so whichever is checked first wins.)

guimard:
@elliefm : of course we can

guimard:
@elliefm: test this patch or wait for some changes ?

elliefm:
Wait for some changes I think, but I thought I'd rope you into the discussion early. :)

configure.ac
anatoli26: In this case we need to properly test the format specifiers in `configure.am`. I guess what you were trying with the redefined-type test is the way to go.

anatoli26:
> Wait for some changes I think, but I thought I'd rope you into the discussion early. :)

Please let me know how would you like to proceed. The `PRIu64` changes were based on what we discussed in the related issue. I could change the things you've marked, rebase and force-push the PR. Or would you prefer some other way?

How your redefined-type test should be incorporated in all this? The 1st commit should be deleted all-together or somehow replaced with your code.

imap/http_admin.c
anatoli26: The clang on OpenBSD was complaining:

```
imap/http_admin.c:605:44: warning: format specifies type 'long *' but the argument has type
  'time_t *' (aka 'long long *') [-Wformat]
            if (sscanf(buf, "btime %ld\n", &boot_time) == 1) break;
                                   ~~~     ^~~~~~~~~~
                                   %lld
```

imap/caldav_alarm.c
anatoli26: Clang on OpenBSD was complaining:

```
imap/caldav_alarm.c:639:16: warning: format specifies type 'long *' but the argument has type
   'time_t *' (aka 'long long *') [-Wformat]
               &data->lastrun, &data->nextcheck) == 2) {
               ^~~~~~~~~~~~~~
imap/caldav_alarm.c:639:32: warning: format specifies type 'long *' but the argument has type
   'time_t *' (aka 'long long *') [-Wformat]
               &data->lastrun, &data->nextcheck) == 2) {                                                                                                                                                                                                                                                                
                               ^~~~~~~~~~~~~~~~
```

lib/auth_pts.c
anatoli26: Right, got symbols-resolved var `now` to function `now()` which is `int64_t`. Will fix this on the next fixing iteration.

imap/http_ischedule.c
anatoli26: OK

anatoli26:
I guess I could drop the 1st commit, incorporate the 2 changes (`PRIu64` & `%zu`) and it would be ready to incorporate your redefined-type check. OK?

guimard:
@anatoli26, @elliefm: to launch the cross platform test, I just need a patch for 3.2.4

elliefm:
> I guess I could drop the 1st commit, incorporate the 2 changes (`PRIu64` & `%zu`) and it would be ready to incorporate your redefined-type check. OK?

I need to fiddle more with the redefined-type check first, but I think this sounds like a plan.  Eventually, you should be able to add my fork as a remote, pull my branch, rebase your branch on top of it, and then force push it here, and my commit(s) with the redefined-type checks will end up as part of this PR too.

> to launch the cross platform test, I just need a patch for 3.2.4

Great, thanks.  Once we've got this working for at least both @anatoli26's system and mine, I'll make a branch/patch that is the same as this PR but based on 3.2 instead of master, and ping you to test it. :)

elliefm:
> I need to fiddle more with the redefined-type check first, but I think this sounds like a plan. Eventually, you should be able to add my fork as a remote, pull my branch, rebase your branch on top of it, and then force push it here, and my commit(s) with the redefined-type checks will end up as part of this PR too.

Okay, fiddling done, I think: https://github.com/cyrusimap/cyrus-imapd/compare/master...elliefm:v33/3128-time-offset-format-detection

So you should be able to rebase on top of this now, dropping your first commit in the process.  Please let me know if you need help. :)

Please check it on your system too -- it works fine for me, and I have no reason to expect it to not be portable, but let's not assume!

configure.ac
obache: How about `strtol` usage?

anatoli26:
@elliefm,

I've rebased my branch on yours, dropped the 1st commit and made the 2 fixes.

On my OpenBSD it builds correctly now (i.e. no format specifier warnings) and `./configure` shows the following:
```
checking underlying integer type of `time_t'... long long int                                                                                                                                                                                                                                                           
checking printf format string for `time_t'... %lld                                                                                                                                                                                                                                                                      
checking underlying integer type of `size_t'... unsigned long int                                                                                                                                                                                                                                                       
checking printf format string for `size_t'... %lu                                                                                                                                                                                                                                                                       
checking underlying integer type of `off_t'... long long int                                                                                                                                                                                                                                                            
checking printf format string for `off_t'... %lld          
```

I guess we're ready to test it on other platforms.

anatoli26:
@obache you mentioned something about `strtol` usage, but I can't find your comment here (got it via mail notifications).

Please let us know if you see something missing.

obache:
> @obache you mentioned something about `strtol` usage, but I can't find your comment here (got it via mail notifications).
> 
> Please let us know if you see something missing.

https://github.com/cyrusimap/cyrus-imapd/pull/3262#discussion_r510586635

configure.ac
anatoli26: @elliefm, what about the `strtoofft` function define that got removed with your 2 commits? Was it unused? And what about the `strtol` that @obache mentions?

elliefm: Oh, I missed that hiding in there!  But searching the source tree, looks like it's never used.  Checking history, it was added in 2008, only used in one place, and then stopped being used in 2010.  So it's just cruft, and doesn't hurt to remove it.

I'm not sure about that `strtol` yet, will think about it some more...

elliefm:
I've updated my branch to do a similar thing to figure out the correct strtol-like parse function for these types, and defines `strtotimet`, `strtosizet` and (brings back) `strtoofft` based on it.  Also converted the `strtol` in zoneinfo_db.c that @obache found to use `strtotimet` instead.

You should be able to pull and rebase on top of it again to get the changes.

anatoli26:
@elliefm, just rebased and fpushed everything. Should be good now, I've just compiled cyrus on OpenBSD with the composite patch and it looks ok, but please check everything again.

I had to change the way I merged the `%zu` fix before, it entered into a wrong commit (the http one), now it's in the correct one (the compile_time one), so all 4 format specifier changes of that line are in the same commit.

elliefm:
Great, thanks!

@guimard Here are the same commits as this PR, but on top of 3.2 instead of master: https://github.com/cyrusimap/cyrus-imapd/compare/cyrus-imapd-3.2...elliefm:v32/3262-patch

or as a single .patch file:
https://github.com/cyrusimap/cyrus-imapd/compare/cyrus-imapd-3.2...elliefm:v32/3262-patch.patch

Can you kick this off on the Debian farm please?  I'm particularly curious to see what happens on the platforms where `time_t` is only 32 bits, since that's been a nuisance in the past!

guimard:
Hi,

I launched the job, you'll see result in a few hours [here](https://buildd.debian.org/status/logs.php?pkg=cyrus-imapd&ver=3.2.4-5), and a few hours later, [test results on amd64, arm64, armhf and i386 only](https://release.debian.org/britney/pseudo-excuses-experimental.html#cyrus-imapd) _(wait for 3.2.4-5, nothing wrong means OK)_

guimard:
@elliefm: build succeeded for all architectures \o/

anatoli26:
> @elliefm: build succeeded for all architectures \o/

But... there are some warnings related to this PR on i386:
```
master/master.c:1990:27: warning: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘rlim_t’ {aka ‘long unsigned int’} [-Wformat=]
 1990 |         syslog(LOG_DEBUG, "set maximum file descriptors to %" PRIu64 "/%" PRIu64,
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1991 |                rl.rlim_cur, rl.rlim_max);
      |                ~~~~~~~~~~~ 
      |                  |
      |                  rlim_t {aka long unsigned int}
In file included from master/master.c:73:
/usr/include/inttypes.h:105:34: note: format string is defined here
  105 | # define PRIu64  __PRI64_PREFIX "u"
master/master.c:1990:27: warning: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 4 has type ‘rlim_t’ {aka ‘long unsigned int’} [-Wformat=]
 1990 |         syslog(LOG_DEBUG, "set maximum file descriptors to %" PRIu64 "/%" PRIu64,
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1991 |                rl.rlim_cur, rl.rlim_max);
      |                             ~~~~~~~~~~~
      |                               |
      |                               rlim_t {aka long unsigned int}
In file included from master/master.c:73:
/usr/include/inttypes.h:105:34: note: format string is defined here
  105 | # define PRIu64  __PRI64_PREFIX "u"
master/master.c:1996:16: warning: format ‘%llu’ expects argument of type ‘long long unsigned int’, but argument 3 has type ‘rlim_t’ {aka ‘long unsigned int’} [-Wformat=]
 1996 |                "setrlimit: Unable to set file descriptors limit to %" PRIu64 ": %m",
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1997 |                rl.rlim_cur);
      |                ~~~~~~~~~~~
      |                  |
      |                  rlim_t {aka long unsigned int}
In file included from master/master.c:73:
/usr/include/inttypes.h:105:34: note: format string is defined here
  105 | # define PRIu64  __PRI64_PREFIX "u"
```
Looks like we can't use `PRIu64` directly and `rlim_t` should also have its own format specifier.


```
imap/index.c: In function ‘extract_attachment’:
imap/index.c:5499:21: warning: format ‘%ld’ expects argument of type ‘long int’, but argument 13 has type ‘size_t’ {aka ‘unsigned int’} [-Wformat=]
 5499 |                     "PUT %s/%s %s\r\n"
      |                     ^~~~~~~~~~~~~~~~~~
......
 5510 |                     type, subtype, buf_cstring(&buf), buf_len(data));
      |                                                       ~~~~~~~~~~~~~
      |                                                       |
      |                                                       size_t {aka unsigned int}
imap/index.c:5506:40: note: format string is defined here
 5506 |                     "Content-Length: %ld\r\n"
      |                                      ~~^
      |                                        |
      |                                        long int
      |                                      %d
```

And it looks like we've missed some hardcoded format specifiers as `buf_len` resolves to `size_t`:
```
lib/util.h
307:size_t buf_len(const struct buf *buf);
```

```
imap/mailbox.c: In function ‘_commit_one’:
imap/mailbox.c:1878:69: warning: comparison of integer expressions of different signedness: ‘ssize_t’ {aka ‘int’} and ‘uint32_t’ {aka ‘unsigned int’} [-Wsign-compare]
 1878 |     if (retry_write(mailbox->index_fd, buf, mailbox->i.record_size) != mailbox->i.record_size) {
      |                                                                     ^~
```
And there are some other unrelated, though probably important warnings to consider. In total only in the i386 build there are more than 300 warnings. I believe we have to check them all to be sure everything works as expected. Different signedness sometimes may lead to security issues.

anatoli26:
> So it looks like we can't use `PRIu64` directly and `rlim_t` should also have its own format specifier.
> 
> And I guess we've missed some hardcoded format specifiers as `buf_len` resolves to `size_t`:

powerpc, mipsel, armhf and armel archs have the same issues.

anatoli26:
And just found in `sieve/sieved.c`:

`static void print_time(uint64_t t)` - looks like it's unused, but if not, is it correct to use a hardcoded 64bit type for time variables?

anatoli26:
@elliefm, I've just fpushed the fixes for `PRIu64` and hardcoded format specifier for `buf_len`, and added a new commit defining `RLIM_T_FMT` and `strtorlim` the same way you've done it for other types. Builds OK on OpenBSD.

@guimard, I guess Ellie is out-of-office now, could you please meanwhile test the new code of this PR just to be sure is has no known issues? A consolidated patch is available here: https://github.com/cyrusimap/cyrus-imapd/pull/3262.patch

guimard:
@anatoli26: I've this error:
```
configure: error: Unable to determine underlying integer type of `rlim_t'
```
Is there a missing patch ?

anatoli26:
> Is there a missing patch ?

Nop. I've just tried it on Linux and got the same error, works well on OpenBSD though.

It looks like there's some include missing on Linux for `rlim_t`, probably `resources.h` that should be explicitly included in the type tests, but I'll leave it for @elliefm.

anatoli26:
@guimard, I've just fpushed a fix. Now it compiles correctly on both OpenBSD and Linux (the correct include was `sys/resource.h`).

Please try to run it. A consolidated patch is available here: https://github.com/cyrusimap/cyrus-imapd/pull/3262.patch (which is the URL of this PR + `.patch` at the end)

guimard:
@anatoli26: build succeeded locally, I'm going to push to debian machines, result [here](https://buildd.debian.org/status/logs.php?pkg=cyrus-imapd&ver=3.2.4-6) in a few minutes (may be 1 hour)

anatoli26:
@guimard, thanks!

@elliefm, now there are no warnings related to this PR on any arch of the build farm, though I would suggest checking the other warnings. Most of them are nothing special like the ones we've seen in the other issue and things like (non-)reproducible builds, but some of them, like pointer casts to int of different size, signedness, etc. + 64bit types and data structures on 32bit archs IMO should be checked with more attention.

Also my latest fix for `buf_len()` implies that the max size it can handle is less than 4Gb on 32bit platforms (`buf_len()` returns `size_t` which is 32bit). Is that an expected behavior?

guimard:
@elliefm: the last build embeds:
 * Debian patches _(like all my uploads, no changes)_:
   * [Disable runpath checking](https://salsa.debian.org/debian/cyrus-imapd/-/blob/master/debian/patches/0001-Disable-runpath-checking.patch)
   * [Shutdown and close sockets cleanly](https://salsa.debian.org/debian/cyrus-imapd/-/blob/master/debian/patches/0002-Shutdown-and-close-sockets-cleanly.patch)
   * [Fix syslog prefix _(missing "cyrus/")_](https://salsa.debian.org/debian/cyrus-imapd/-/blob/master/debian/patches/0003-Fix-syslog-prefix.patch)
   * [silence erroneous RLIMIT_NUMFDS-related log messages](https://salsa.debian.org/debian/cyrus-imapd/-/blob/master/debian/patches/0004-silence-erroneous-RLIMIT_NUMFDS-related-log-messages.patch)
   * Updates calling of the perl interpreter to what we expect in Debian
   * Fix paths for Debian in tools/rehash
   * [Adds some enhancements to sieveshell and fixes some paths](https://salsa.debian.org/debian/cyrus-imapd/-/blob/master/debian/patches/0007-Adds-some-enhancements-to-sieveshell-and-fixes-some-.patch)
   * [Accept invalid From: header in email](https://salsa.debian.org/debian/cyrus-imapd/-/blob/master/debian/patches/0008-Accept-invalid-From-header-in-email.patch)
   * [Normalize the authentication ID](https://salsa.debian.org/debian/cyrus-imapd/-/blob/master/debian/patches/0009-Normalize-the-authentication-ID.patch)
   * [Fix PATH_MAX on GNU/Hurd](https://salsa.debian.org/debian/cyrus-imapd/-/blob/master/debian/patches/0010-Fix-PATH_MAX-on-GNU-Hurd.patch)
   * [Fix extra libpci in SNMP_LIBS](https://salsa.debian.org/debian/cyrus-imapd/-/blob/master/debian/patches/0011-Fix-extra-libpci-in-SNMP_LIBS.patch)
   * Use UnicodeData.txt from system
   * [conversations: update the G keys even if the message has no threadid](https://salsa.debian.org/debian/cyrus-imapd/-/blob/master/debian/patches/0016-conversations-actually-need-to-update-GUID-counts-after-calculating-CID.patch)
   * Increase test timeout
 * HTTP/2 fix (2696ed7)
 * #3262 last patch

I'm Cyrus-Imapd maintainer in Debian Project for 1 year, I don't know if the Debian patches have been reported (probably not). I added links to patches that may interest you. I can build some MR with some of them if you want.

Cheers,
Xavier

anatoli26:
@elliefm, there's small issue with the `rlim_t` size check. On all archs under Linux `./configure` shows: `checking size of rlim_t... 0` (and on OpenBSD it shows the correct size).

`config.log` shows:
```
677:configure:17531: checking size of rlim_t
680:conftest.c:69:57: error: 'rlim_t' undeclared (first use in this function)
```

The `AC_CHECK_SIZEOF(rlim_t)` should include `sys/resource.h` either explicitly in the 3rd param or implicitly via `AC_INCLUDES_DEFAULT`.

Anyway, it's a cosmetic issue at this stage.

elliefm:
> 64bit types and data structures on 32bit archs IMO should be checked with more attention.

These multi-platform tests are very useful for finding bad assumptions and stress-testing the configure script, but they also produce a LOT of noise.

For example, Cyrus is an email and calendar server, both of which features need precise timestamps into the future and past, and to store potentially a lot of data.  There is absolutely no reason for anyone to be running Cyrus in 2020 on a platform whose `time_t` or `size_t` is only 32 bits!  (We even warn about it from configure.)  So, it's useful to see what i386 et al complain about, but we don't really need to expend any effort making it portable to them if the only problem is portability.

Using the c99 fixed-size types (like uint64_t etc) is fine too -- it's completely reasonable to say that a particular variable needs to be a particular size and signedness (we often need this specificity when the data being represented is from one of our binary cyrusdb files).  If the platform doesn't support that size or signedness, we simply don't support the platform.

The cunit tests are all passing, so nothing looks too out of place in a way that matters (or at least, in way that matters and is tested for...)

> in the i386 build there are more than 300 warnings

How did you get this number?  I suspect it's polluted with the irrelevant library path warnings from libtool...

I checked the log by eye, and it looks like the interesting ones (`rlim_t`, `buf_len()`) you've already found and taken care of.  Great :)

elliefm:
> And just found in `sieve/sieved.c`:
> 
> `static void print_time(uint64_t t)` - looks like it's unused, 

It's a callback function for `print_vallist`.  It's used at line 828.

> but if not, is it correct to use a hardcoded 64bit type for time variables?

Yeah it's reasonable.  Any platform worth actually using has a 64-bit time_t these days (32 bit time_t overflows on dates beyond 2038...) .  

The fact that it's unsigned is strange (unsigned `time_t` can't represent dates before 1970), but I don't know this section of code, so I don't know if that's a real problem or an imaginary one in this context.  It's sieve though, which is mostly about email (not calendaring), and email didn't exist in 1970, so it probably doesn't matter...

elliefm:
@anatoli26 can you reorder your commits please?  At the moment the commit that uses RLIM_T_FMT occurs before the commit that detects and defines it, which will be a nuisance for anyone running a bisect past this code in the future.

@guimard That "fix path max on gnu/hurd" patch looks familiar... Ahh, f03b10ae1e.  We got a few patches from Debian around the same time, but I can't remember if they were from a GitHub issue/PR or from the mailing list, nor who submitted them.  Looks like the current version of the Debian patch is different from the one that's in our source, so I guess nobody removed the Debian patch after it was upstreamed, and it keeps getting local modifications...

"conversations: update the G keys..." looks like it was a Debian backport for 3.0.  It's 95d204ff12 and is already in 3.2, so you can just drop it.

"Normalize the authentication ID" is interesting.  It adds a new "normalizeuid" imapd.conf setting, and apparently it's been floating around a vendor-only patch since at least 2007, according to #964.  The apparent author, Jeroen, was a regular Cyrus contributor a few years ago, so it's strange that this is floating around as a vendor patch and not actually integrated??  Looks like there's a couple of other patches here from Jeroen that aren't upstream, very curious.

Some of the patches look like Debian-specific things, but some are probably useful more generally.  For the general stuff, I'm happy to shepherd through PRs to integrate them if you're happy to submit them. :)

elliefm:
> there's small issue with the `rlim_t` size check. On all archs under Linux `./configure` shows: `checking size of rlim_t... 0` (and on OpenBSD it shows the correct size).
> [...]
> Anyway, it's a cosmetic issue at this stage.

You can just remove that `AC_CHECK_SIZEOF(rlim_t)`: nothing uses the result, and it's not interesting enough to keep around for config.log

anatoli26:
@elliefm,

> How did you get this number? I suspect it's polluted with the irrelevant library path warnings from libtool...

On the build results page [1] there are 308 warnings, most of them noise warnings. As I've already mentioned and as you replied, there are some 32bit vs 64bit and some signedness issues, but if the policy is not to support i386 and other 32bit archs, maybe it would be clearer and safer to explicitly say that in the docs and other places, so nobody even tries to build and run it on those archs and we save on maintenance?

BTW, on all archs there are these 2 warnings that I'm not sure about:
```
imap/global.c:355:24: warning: implicit conversion from ‘enum cyrus_opt’ to ‘enum imapopt’ [-Wenum-conversion]
  355 |       config_getswitch(CYRUSOPT_NORMALIZEUID));
      |                        ^~~~~~~~~~~~~~~~~~~~~
```
and
```
imap/message.c:314:39: warning: array subscript has type ‘char’ [-Wchar-subscripts]
  314 |   if (is_from > 0 && *p != from_header[is_from]) {
      |                                       ^
```

> @anatoli26 can you reorder your commits please? At the moment the commit that uses RLIM_T_FMT occurs before the commit that detects and defines it, which will be a nuisance for anyone running a bisect past this code in the future.

Done!

> You can just remove that `AC_CHECK_SIZEOF(rlim_t)`: nothing uses the result, and it's not interesting enough to keep around for config.log

Done!

Is there anything else pending for this PR?

[1] https://buildd.debian.org/status/fetch.php?pkg=cyrus-imapd&arch=i386&ver=3.2.4-6&stamp=1604675816&raw=0

guimard:
> "conversations: update the G keys..." looks like it was a Debian backport for 3.0. It's [95d204f](https://github.com/cyrusimap/cyrus-imapd/commit/95d204ff12daa16b1987e2ad1495eb5e73556c49) and is already in 3.2, so you can just drop it.

@elliefm: removed, thanks!

> "Normalize the authentication ID" is interesting. It adds a new "normalizeuid" imapd.conf setting, and apparently it's been floating around a vendor-only patch since at least 2007, according to #964. The apparent author, Jeroen, was a regular Cyrus contributor a few years ago, so it's strange that this is floating around as a vendor patch and not actually integrated?? Looks like there's a couple of other patches here from Jeroen that aren't upstream, very curious.
> 
> Some of the patches look like Debian-specific things, but some are probably useful more generally. For the general stuff, I'm happy to shepherd through PRs to integrate them if you're happy to submit them. :)

Done also :wink:

elliefm:
> BTW, on all archs there are these 2 warnings that I'm not sure about:
> 
> ```
> imap/global.c:355:24: warning: implicit conversion from ‘enum cyrus_opt’ to ‘enum imapopt’ [-Wenum-conversion]
>   355 |       config_getswitch(CYRUSOPT_NORMALIZEUID));
>       |                        ^~~~~~~~~~~~~~~~~~~~~
> ```

This is from the "Normalize the authentication ID" Debian patch, which @guimard has now submitted as #3283 (but I have not yet looked at further).  We can ignore it here, and deal with it there instead; it's not related to this PR and is not currently upstream either.

> 
> and
> 
> ```
> imap/message.c:314:39: warning: array subscript has type ‘char’ [-Wchar-subscripts]
>   314 |   if (is_from > 0 && *p != from_header[is_from]) {
>       |                                       ^
> ```

I haven't seen this warning before, but looks like it's pointing out that `char`'s signedness is platform dependent and could potentially be signed and negative...

But now I've searched master and 3.2 for `from_header` and it's not in upstream code, so I guess this is from another one of the Debian patches -- which would explain why I don't see the warning on my own builds!  In either case, it's not related to this PR.

> Is there anything else pending for this PR?

I don't think so, but I want to check it over again just in case.  I'll do that today :)

elliefm:
I've also merged this onto the 3.2 branch

anatoli26:
Great! Thanks!

cmulocal/cyr_inttype.m4
```diff
@@ -0,0 +1,94 @@
...
+#
+#       CYR_INTTYPE([time_t], [#include <time.h>])
+#       AC_DEFINE_UNQUOTED([TIME_T_FMT], ["$cyr_cv_format_time_t"], [...])
+#       AC_DEFINE_UNQUOTED([strtotimet(a,b,c)], [$cyr_cv_parse_time_t(a,b,c)], [...])
+#
+AC_DEFUN([CYR_INTTYPE],[
+    dnl First, figure out what type of integer it is, by exploiting the
+    dnl behaviour that redefining a variable name as the same type is only
+    dnl a warning, but redefining it as a different type is an error.
+    AC_CACHE_CHECK(
```
ksmurchison, line 29: This method of forcing a warning fails on my dev box using:

gcc (GCC) 10.2.1 20201016 (Red Hat 10.2.1-6)
CPPFLAGS="-W -Wall -Werror -Wno-cpp"

I have to remove -Werror in order for configure to complete successfully.  Since I always use -Werror to check for questionable code in Cyrus, this change is a non-starter for me.  Not to mention that it seems somewhat fragile.

ksmurchison: I just committed https://github.com/cyrusimap/cyrus-imapd/commit/b3f30361953a915e40dab27689ff4daf6f35ae0b

Which works arounf the problem for me, pending a possible better fix

obache: Such flags should not be used for configure, or some kind of feature detections may fail silently and will be disabled,
OK to use for building, though, just to check codes.

elliefm: @ksmurchison Oh right, thanks. I already did this for CFLAGS (because I always have -Werror in CFLAGS), but forgot about CPPFLAGS.  I'll tidy this up to put the CPPFLAGS handling in the same place as the CFLAGS handling for consistency, but I'll also check docs and see if there's anything else that needs the same treatment.

PR #3266, opened 2020-10-27
xapian: skip indexing of already indexed body parts

rsto:
New squatter -D option allows to skip duplicate part check.

Tested in https://github.com/cyrusimap/cassandane/commit/c0c5d37ce3c9e8b6cc47f9a8e684214396f29c69

### DIFF
--- a/imap/search_xapian.c
+++ b/imap/search_xapian.c
@@ -2145,6 +2145,7 @@ struct xapian_update_receiver
     strarray_t *activetiers;
     hash_table cached_seqs;
     int mode;
+    int flags;
 };
 
 /* receiver used for extracting snippets after a search */
@@ -2159,6 +2160,13 @@ struct xapian_snippet_receiver
     struct xapiandb_lock lock;
     const search_snippet_markup_t *markup;
 };
+
+struct is_indexed_rock {
+    xapian_update_receiver_t *tr;
+    char doctype;
+};
+
+static int is_indexed_cb(const conv_guidrec_t *rec, void *rock);
 
 static const char *xapian_rootdir(const char *tier, const char *partition)
 {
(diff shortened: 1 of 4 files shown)


imap/xapian_wrap.cpp
```diff
@@ -780,16 +780,14 @@ int xapian_dbw_open(const char **paths, xapian_dbw_t **dbwp, int mode, int nosyn
     }
 
     /* open the read-only databases */
-    if (mode == XAPIAN_DBW_XAPINDEXED) {
```
brong, line 783: Unfortunately this adds a significant IO hit :(  every time a message gets indexed you need to open all the databases, which is the reason why XapIndexed isn't on by default already despite being clearly the better option.

I think we're going to need to use conversations.db to see if there part is mentioned in an already indexed email :(

brong (changes requested):

imap/xapian_wrap.cpp
rsto: Thanks for noticing this. OK, I'll rewrite it to use conversation.db.

rsto:
The updated patch now also checks conversation.db and cyrus.indexed.db for CONVINDEXED mode. To make this work I had to remove the logic that is_indexed for a message adds that message uid to the indexed sequence set. This seemed bogus anyways and we now defer addition to the seqset until the end of end_message_update.

imap/squatter.c
```diff
@@ -889,7 +892,7 @@ int main(int argc, char **argv)
 
     setbuf(stdout, NULL);
 
-    while ((opt = getopt(argc, argv, "C:N:RUBXZT:S:Fde:f:mn:riavpPL:Az:t:ouhl")) != EOF) {
+    while ((opt = getopt(argc, argv, "C:N:RUBXZDT:S:Fde:f:mn:riavpPL:Az:t:ouhl")) != EOF) {
```
brong, line 895: alphabet soup!  We must be nearly finished with the entire ASCII character set here

imap/search_xapian.c
```diff
@@ -2661,9 +2714,16 @@ static uint32_t first_unindexed_uid(search_text_receiver_t *rx)
...
-    xapian_update_receiver_t *tr = rock;
+    xapian_update_receiver_t *tr = ((struct is_indexed_rock*)rock)->tr;
+    char doctype = ((struct is_indexed_rock*)rock)->doctype;
 
-    if (rec->part) return 0;
+    if (doctype == XAPIAN_WRAP_DOCTYPE_MSG && rec->part) return 0;
+
+    /* Is this a part in the message we are just indexing? */
+    if (doctype == XAPIAN_WRAP_DOCTYPE_PART && rec->uid == tr->super.uid &&
+            !strcmp(rec->mboxname, tr->super.mailbox->name)) {
```
brong, line 2724: I guess this is OK... I'd prefer to number compare on rec->foldernum, but this is fine.

Question: is this supposed to stop the *P* key being created as well, or make the attachment text not be indexed along with the main message itself?

brong (commented):
Yes, this is the right path :)  I'm still not clear on what the is_indexed logic does for being the current message rather than checking if it's already indexed at all - since we could have other future messages not yet indexed - so we definitely need to do a lookup in cyrus.indexed.db, not just check current message.

imap/search_xapian.c
rsto: It's supposed to prevent a P-key being created more than once (along with its indexed terms). The attachment text will be indexed along with the G key multiple times (once for each message).

imap/search_xapian.c
```diff
@@ -2460,6 +2468,33 @@ static int compare_segs(const void **v1, const void **v2)
...
 
+static int is_indexed_part(xapian_update_receiver_t *tr, const struct message_guid *guid)
+{
+    if (tr->mode == XAPIAN_DBW_XAPINDEXED) {
+        return xapian_dbw_is_indexed(tr->dbw, guid, XAPIAN_WRAP_DOCTYPE_PART);
+    }
+
+    struct conversations_state *cstate = mailbox_get_cstate(tr->super.mailbox);
+    if (!cstate) {
+        xsyslog(LOG_INFO, "can't open conversations for %s",
```
wolfsage, line 2479: %s will come through in the error logs literally here. xsyslog requires a third argument that provides any printf-style arguments.

This also affects the two error messages in activefile_open()

elliefm: Annoyingly, gcc only catches this particular misuse with -Wformat=2 or -Wformat-nonliteral, but we we have cases elsewhere where we legitimately need to use a non-literal format string, so I guess none of us have this warning level enabled usually.

I fixed another case of this that already slipped through to master here: b86e0207d8

rsto: Fixed and force-pushed a branch update

PR #3269, opened 2020-11-02
jmap_contact: decode VCARD text values to UTF-8

rsto:
This patch changes Contact/get to handle invalid UTF-8 data in VCARD fields such as EMAIL, N, FN, etc.

However, in contrast to Email/get this does not allow for repairing bogus data: a couple of tests on libchardet reveal that it is likely to incorrectly guess the charset for very short text such as email addresses and other contact data. In contrast to Email/get the Contact/get method does not allow indicate to the client if any of the text values may have character encoding issues. To err on the safe side, this patch instructs the charset detector to always fall back to UTF-8 replacement characters for any invalid UTF-8 sequences. A Cassandane test highlights this issue https://github.com/cyrusimap/cassandane/commit/56ab4804233d560cbb7be793ae5a1b591c96727b (where we could repair the wrongly encoded data, but at the likely cost of returning bogus values for other cases).

This patch also refactors the charset detection helper used in Email/get for general use by the JMAP modules. The API allows to specify a minimum confidence [0.0,1.0] for the charset detection heuristic.

### DIFF
--- a/cunit/jmap_util.testc
+++ b/cunit/jmap_util.testc
@@ -134,4 +134,85 @@ static void test_patchobject_create(void)
 #undef TESTCASE
 }
 
+static void test_decode_to_utf8(void)
+{
+    struct testcase {
+        const char *data;
+        const char *charset;
+        int encoding;
+        float confidence;
+        const char *want_val;
+        int want_alloc;
+        int want_is_encoding_problem;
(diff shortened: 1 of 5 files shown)


ksmurchison (approved):
Looks good to me

PR #3270, opened 2020-11-03
Sievedir consolidation

ksmurchison:
This is a first pass at consolidating all of the code that touches a user's sieve directory into a single set of APIs


imap/sievedir.c
```diff
@@ -0,0 +1,426 @@
...
+}
+
+EXPORTED int sievedir_script_exists(const char *sievedir, const char *name)
+{
+    char path[PATH_MAX];
+    struct stat sbuf;
+
+    snprintf(path, sizeof(path), "%s/%s%s", sievedir, name, SCRIPT_SUFFIX);
+
+    return (stat(path, &sbuf) == 0);
```
elliefm, line 198: In other places we ignore files that are not regular files, do we need to do the same here?  I guess the question is what this API reports.  Does it report "something with this name exists", or does it report "a usable script with this name exists"?

imap/sievedir.c
```diff
@@ -0,0 +1,426 @@
...
+
+    snprintf(target, sizeof(target), "%s%s", name, BYTECODE_SUFFIX);
+    snprintf(active, sizeof(active), "%s/%s", sievedir, DEFAULTBC_NAME);
+    snprintf(tmp, sizeof(tmp), "%s.NEW", active);
+
+    /* N.B symlink() does NOT verify target for anything but string validity,
+     * so activation of a nonexistent script will report success.
+     */
+    if (symlink(target, tmp) < 0) {
+        syslog(LOG_ERR, "IOERROR: unable to symlink %s as %s: %m", target, tmp);
```
elliefm, line 248: Can you do the IOERRORs with xsyslog please?  They're just gonna be converted anyway, might as well start out right.

imap/sievedir.c
```diff
@@ -0,0 +1,426 @@
...
+
+EXPORTED int sievedir_delete_script(const char *sievedir, const char *name)
+{
+    char path[PATH_MAX];
+    int r;
+
+    snprintf(path, sizeof(path), "%s/%s%s", sievedir, name, SCRIPT_SUFFIX);
+    r = unlink(path);
+    if (r) {
+        if (errno == ENOENT) return SIEVEDIR_NOTFOUND;
```
elliefm, line 282: So, if there is a .bc file but the corresponding .script file is missing for some reason, this function will bail out here and not remove the .bc file?

imap/sievedir.c
```diff
@@ -0,0 +1,426 @@
...
+
+#include "assert.h"
+#include "map.h"
+#include "sievedir.h"
+#include "sieve/bc_parse.h"
+#include "sieve/sieve_interface.h"
+#include "util.h"
+#include "xstrlcpy.h"
+
+EXPORTED void sievedir_foreach(const char *sievedir,
```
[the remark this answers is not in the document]
ksmurchison: https://github.com/cyrusimap/cyrus-imapd/commit/7c9d1e6628463049767efd110340a7438110ed56

imap/sievedir.c
ksmurchison: https://github.com/cyrusimap/cyrus-imapd/commit/5044b936283aa8000a55f23077ca75a3d79d474a

imap/sievedir.c
ksmurchison: https://github.com/cyrusimap/cyrus-imapd/commit/534ad6f45f996076c98722207d62e9bf7f165fb8

imap/sievedir.c
ksmurchison: https://github.com/cyrusimap/cyrus-imapd/commit/4a3e19ae40feab8cbfe8a27f0685e23439a27280

imap/sievedir.c
```diff
@@ -0,0 +1,426 @@
...
+    /* must be at least one character long */
+    if (len < 1) return 0;
+
+    ptr = buf_base(name);
+
+    for (lup = 0; lup < len; lup++) {
+        if ((ptr[lup] == '/') || (ptr[lup] == '\0')) return 0;
+    }
+
+    return (lup < 1013);
```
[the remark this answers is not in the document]
ksmurchison: https://github.com/cyrusimap/cyrus-imapd/commit/a5c61e643d2900203197e7e186ee228f6cc50992

dilyanpalauzov:
It is enough to include imap/sievedir.h in Makefile.am a single time for the whole Makefile.  The purpose of mentioning the file is just to include it in the tarball on 'make dist'.  Since sievedir.c is used that ofthen it might be good to include it in a library (libimap).

ksmurchison:
The problem with including sievedir.c in libcyrus_imap is that we then have to pull in libcyrus_sieve for everything.
If you want to put together a PR that resolves this, I would gladly accept it.

dilyanpalauzov:
The problem with the PRs is, that it is not foreseeable what will be accepted and what not.  For instance, I stated above, that mentioning a .h file in Makefile.am more than once has no added value.  This was not tackled and it is not predictable, whether a PR on this will be considered, because I do not know the reasons why my comment was ignored.

Using the autocreate-on-first-login function, that shall be extended also on autocreate-on-first-login-over-webdav creates a circular dependency between libcyrus_imap and libcyrus_sieve (in the way I implemented it).  The concept of having separate libraries for the master process and the other processes, that comes from the times when the libraries were static (so essentially an archive of .o files that are included in the binaries) is outdated.  The separate libraries were from the beginning on just a collection of (no-pic) .o files, that were included in the binaries and once the libraries were converted into shared ones, it turned out, that the separations between the libraries were done wrong (as far as I remember it lead to circular dependencies, when the libraries were converted from static to shared).

The [Zygotte process](https://medium.com/@voodoomio/what-the-zygote-76f852d887d9) on Android has a different approach: it loads all dependencies in the master process and on each fork, these dependencies do not have to be loaded again.  The Zygotte-forking-optimization is for a JavaVM, I guess for pure Linux processes it will work the same (but I am not 100% sure).

Duplicating code in each binary in order do avoid loading a shared library, essentially makes shared libraries not so useful.

PR #3274, opened 2020-11-06
Fix SRS enable funcionality

Devstellar:
Currently although you enable SRS in configure, SRS is not enabled
There is because there is a missing AC_DEFINE in configure.ac not generating the "USE_SRS" define.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -698,16 +698,17 @@ AC_ARG_ENABLE(srs,
     [AS_HELP_STRING([--enable-srs], [enable Sender Rewriting Scheme support])],,
     [enable_srs=no])
 AC_MSG_RESULT($enable_srs)
+AM_CONDITIONAL([USE_SRS], [test "${enable_srs}" != "no"])
 if test "x$enable_srs" != xno; then
     AC_CHECK_HEADER([srs2.h],,
         [AC_MSG_ERROR([cannot find SRS headers])])
     AC_CHECK_LIB(srs2, srs_forward,,
         [AC_MSG_ERROR([cannot find SRS lib])])
     LIBS="$LIBS -lsrs2"
+    AC_DEFINE(USE_SRS,[],[Build with srs functionality])
 else
     AC_MSG_NOTICE([Outgoing messages will not support the Sender Rewriting Scheme.  Consider installing libsrs2 and --enable-srs])
 fi
-AM_CONDITIONAL([USE_SRS], [test "${enable_srs}" != "no"])
 
 dnl look for an option to disable sign-comparison warnings (needed for
 dnl flex-generated sieve sources when building with -Werror)


elliefm:
With this patch, and with libsrs2 installed, what happens when you:

```
./configure --enable-srs [...]
make
make check
```

For me, `make` finishes fine, but `make check` fails because cunit/unit has not been correctly linked, and cannot find libsrs2.so.0:

```
[...]
Running unit tests
/home/ellie/fastmail/cyrus-imapd/cunit/.libs/unit: error while loading shared libraries: libsrs2.so.0: cannot open shared object file: No such file or directory
```

Since the configure.ac was wrong, I guess this code has never been compiled and run, and probably has more issues...

Devstellar:
Honestly I don't run make check properly because I haven't cunit installed and I don't enable "--enable-unit-tests"

Code should be safe, it's only used in Sieve (mainly for "redirect"):

```
cyrus-imapd-3.2.4# grep -ri "use_srs" *
configure:USE_SRS_FALSE
configure:USE_SRS_TRUE
configure:  USE_SRS_TRUE=
configure:  USE_SRS_FALSE='#'
configure:  USE_SRS_TRUE='#'
configure:  USE_SRS_FALSE=
configure:if test -z "${USE_SRS_TRUE}" && test -z "${USE_SRS_FALSE}"; then
configure:  as_fn_error $? "conditional \"USE_SRS\" was never defined.
configure.ac:AM_CONDITIONAL([USE_SRS], [test "${enable_srs}" != "no"])
imap/lmtp_sieve.c:#ifdef USE_SRS
imap/lmtp_sieve.c:#else /* !USE_SRS */
imap/lmtp_sieve.c:#endif /* USE_SRS */
```

And you need to enable explicitly in imapd.conf:
```
srs_domain: <none>
    The domain to use in rewritten addresses. This must point only to machines which
    know the encoding secret used by this system. **When present, SRS is enabled**.
```
Actually you need to fill "srs_domain" and, at least, one secret in "srs_secrets", for it to work without complaining.

Full SRS funcionality is not as important as to have "redirect" really working in a full SPF/DMarc world like the actual one.
This is, "redirect" in actual real world will fail cause SPF checks on the destination MTAs (mails are redirected to different domains) and they will reject messages if "from address" is unaltered. So, it is critical to use SRS (or another similar code) to change "from address" and avoid this.
Ex. changed "from address" in SMTP transaction when SRS enabled: "Return-Path: SRS0=jJEn=EM=finaldestination.com=name@originaldestination.com".

I have been using SRS in cyrus-imapd for some days and it is working fine, no more rejects or mails going into spam folder on final destination domains.

elliefm:
> For me, `make` finishes fine, but `make check` fails because cunit/unit has not been correctly linked, and cannot find libsrs2.so.0:

Ah, my bad.  libsrs2 is not packaged by Debian, so I had to install it from source, and forgot to run `ldconfig` after doing so.  The tests work fine when I install the library properly! 😅

elliefm:
Cherry-picked to 3.2 as 22ecfe7d97, will be in the next 3.2 release

PR #3276, opened 2020-11-08
configure.ac: improve detection of optimisation support

elliefm:
[This fixes some annoying warnings from clang on OpenBSD.  We were trying to do the right thing, but weren't quite there.]

Missing support is only a compiler warning, so we need to force
-Werror on to be able to detect it.

Convert to AC_COMPILE_IFELSE and AC_LANG_PROGRAM so we can fiddle
with CFLAGS, and also convert to AC_CACHE_CHECK since we're
refactoring anyway...

Part of #3265


ksmurchison (approved):
This change works as expected on my Fedora 33 box

PR #3277, opened 2020-11-09
Fix PATH_MAX on GNU/Hurd

guimard:
This little patch is used in Debian for GNU/Hurd. See also f03b10ae

### DIFF
--- a/imap/pop3d.c
+++ b/imap/pop3d.c
@@ -99,6 +99,10 @@
 #define MAXHOSTNAMELEN 256
 #endif
 
+#ifndef MAXHOSTNAMELEN
+#define MAXHOSTNAMELEN
+#endif
+
 #ifdef HAVE_KRB
 /* kerberos des is purported to conflict with OpenSSL DES */
 #define DES_DEFS
--- a/master/master.c
+++ b/master/master.c
@@ -76,6 +76,10 @@
 #define PATH_MAX 4096
 #endif
 
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
+
 #ifndef INADDR_NONE
 #define INADDR_NONE 0xffffffff
 #endif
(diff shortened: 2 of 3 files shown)


imap/pop3d.c
```diff
@@ -99,6 +99,10 @@
 #define MAXHOSTNAMELEN 256
 #endif
 
+#ifndef MAXHOSTNAMELEN
+#define MAXHOSTNAMELEN
```
elliefm, line 103: In the original patch that we already have, this is defined to `256`, but in this version, it's defined without a value.  Do you know what the difference is?

I also just realised, the "256" definition is immediately above, so it will already be defined and not do anything.

master/master.c
```diff
@@ -76,6 +76,10 @@
 #define PATH_MAX 4096
 #endif
 
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
+
```
elliefm, line 82: This is also duplicated immediately above

master/service.c
```diff
@@ -77,6 +77,10 @@
 #define PATH_MAX 4096
 #endif
 
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
+
```
elliefm, line 83: As is this one!

elliefm (commented):
So it looks like we already have all the changes needed here!

The only question is, what's the difference between `#define MAXHOSTNAMELEN 256` vs `#define MAXHOSTNAMELEN (no value)`?

The no-value version seems nonsensical to me (since presumably anything depending on it expects a numeric length??).

Depending on the answer, maybe you can just drop this patch from Debian entirely, since all the changes it wants to make seem to be already in 3.2 :)

rsto:
These changes all are already merged.

PR #3278, opened 2020-11-09
Shutdown and close sockets cleanly

guimard:
Comment from Patch author _(Debian Dev)_:
> Cleanly shutdown and close sockets, this is supposed to allow for better TCP teardown on the remote end, and reduces CLOSE_WAIT time.
> 
> This patch was written 8 years ago, it is possible that nowadays nothing will benefit from a shutdown() right before close().  The commit log from eight years ago mentions that SHUT_RD should be upgraded to SHUT_RDWR where possible, but only after verification that this is not going to cause problems (e.g. by discarding data still on flight to the remote).
> 
> Also, it is possible that new daemons and utils in Cyrus 2.2 and 2.3 may need similar patches.

### DIFF
--- a/imap/fud.c
+++ b/imap/fud.c
@@ -145,6 +145,15 @@ void shut_down(int code)
     seen_done();
     closelog();
     cyrus_done();
+
+    /* be nice to remote */
+    shutdown(0, SHUT_RD);
+    shutdown(1, SHUT_RD);
+    shutdown(2, SHUT_RD);
+    close(0);
+    close(1);
+    close(2);
+
     exit(code);
 }
 
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -1136,6 +1136,10 @@ void shut_down(int code)
 #ifdef HAVE_SSL
     tls_shutdown_serverengine();
 #endif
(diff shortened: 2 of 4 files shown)


elliefm:
@brong @ksmurchison @rsto  What do you think of this?  It doesn't seem like it causes any harm, but it also doesn't seem like it gets us any gain (otherwise, I guess we would have spotted this while profiling and done it ourselves).  In all cases, we're about to call exit(), and all our sockets will be shut down and closed then anyway.

It's probably also very incomplete, since the description mentions needing to check and update the "new daemons and utils in 2.2 and 2.3"...

If this is useful, it seems like it wouldn't be hard to apply this consistently across all our services.  But I'm not convinced it is.  And if it's not useful, Debian should save themselves some maintenance work and just drop this patch.

rsto:
It's not clear if this change actually is useful. We haven't experienced any issues with the current implicit close/shutdown sequence of the socket. By the description of the Debian patch it also does not seem to be clear if this patch actually solves an issue or not.
I will close this PR due to inactivity. If there is an issue with the implicit socket close in the current code, please let us know.

PR #3279, opened 2020-11-09
Fix syslog prefix

guimard:
Make sure all programs log (to syslog) with "cyrus/<program>" as the log prefix.

### DIFF
--- a/imap/global.c
+++ b/imap/global.c
@@ -156,6 +156,9 @@ static int get_facility(const char *name)
     return SYSLOG_FACILITY;
 }
 
+/* syslog prefix tag */
+static char syslog_prefix[20];
+
 struct cyrus_module {
     void (*done)(void *rock);
     void *rock;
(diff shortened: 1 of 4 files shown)


elliefm:
I think this one is a Debianism, and it would be a surprising change to users of upstream Cyrus, so I won't merge it to 3.2.  

We can discuss changing it on master for 3.4 -- I don't have a strong opinion about it either way.  We would want to check carefully to make sure all programs are updated properly, since it's not clear from looking at the diff alone whether it's been kept up to date with the addition of new programs over time.

But it also seems fine to just leave this one as a Debian patch

guimard:
Thanks @elliefm, what surprised me is this patch: was needed only in a few places, so I pushed it to see if this was not a oversight. No problem to maintain this patch in Debian, that's the "Debian Developer" job :wink:

rsto:
We'll keep logging as-is, glad to hear it's not causing a hassle for Debian package maintenance, too.

PR #3280, opened 2020-11-09
silence erroneous RLIMIT_NUMFDS-related log messages

guimard:
Debian author comment:
> Fixes setrlimit(RLIMIT_NUMFDS) handling to be less obnoxious and not barf error messages to syslog incorrectly, nor log nonsense if getrlimit(RLIMIT_NUMFDS) failed.

### DIFF
--- a/master/master.c
+++ b/master/master.c
@@ -1968,7 +1968,7 @@ static void limit_fds(rlim_t x)
     struct rlimit rl;
 
 #ifdef HAVE_GETRLIMIT
-    if (!getrlimit(RLIMIT_NUMFDS, &rl)) {
+    if (getrlimit(RLIMIT_NUMFDS, &rl) >= 0) {
         if (x != RLIM_INFINITY && rl.rlim_max != RLIM_INFINITY && x > rl.rlim_max) {
             syslog(LOG_WARNING,
                    "limit_fds: requested %" PRIu64 ", but capped to %" PRIu64,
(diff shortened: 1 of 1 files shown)


master/master.c
```diff
@@ -1968,7 +1968,7 @@ static void limit_fds(rlim_t x)
     struct rlimit rl;
 
 #ifdef HAVE_GETRLIMIT
-    if (!getrlimit(RLIMIT_NUMFDS, &rl)) {
+    if (getrlimit(RLIMIT_NUMFDS, &rl) >= 0) {
```
elliefm, line 1971: `getrlimit` only returns `0` (on success) or `-1` (on error), so this is only a code style change -- and a misleading one at that, since it incorrectly implies that it might return a non-zero positive number sometimes.

master/master.c
```diff
@@ -1987,7 +1987,7 @@ static void limit_fds(rlim_t x)
                rl.rlim_cur, rl.rlim_max);
     }
 
-    if (setrlimit(RLIMIT_NUMFDS, &rl) < 0) {
+    if (setrlimit(RLIMIT_NUMFDS, &rl) < 0 && x != RLIM_INFINITY) {
```
elliefm, line 1990: And this change makes it so that it _won't_ log an error when it tries and fails to make the file descriptors unlimited?  Which seems wrong: if the admin has requested the process should have an unlimited number of fds, and yet the process does not, surely they wish to know about this?

elliefm (commented):
This patch smells like it was conceived to shut up some run time warnings by someone who didn't want or care to configure their setup better?  It doesn't really seem useful to me

guimard:
Thanks @elliefm, dropped : https://salsa.debian.org/debian/cyrus-imapd/-/commit/b22b81d

PR #3281, opened 2020-11-09
Add --password and --execfile options to sieveshell

guimard:


elliefm:
Thanks.  I cleaned this up a little and manually merged it as 8c15246156 on 3.2 and 4f3d911e51 on master.

You should be able to remove this patch from Debian once 3.2.5 is released :)

PR #3282, opened 2020-11-09
Accept invalid From: header in email

guimard:
Debian author comment:
> This stops Cyrus from ruining everything all by itself; Your MTA is the point where the validity of headers should be checked. For more information on such and so forth, see (amongst others):
> * http://www.postfix.org/header_checks.5.html
> 
> If one chooses to allow slightly invalid headers in the MTA, then Cyrus should not raise problems.

### DIFF
--- a/imap/message.c
+++ b/imap/message.c
@@ -190,6 +190,8 @@ EXPORTED int message_copy_strict(struct protstream *from, FILE *to,
     int munge8bit = config_getswitch(IMAPOPT_MUNGE8BIT);
     int inheader = 1, blankline = 1;
     struct buf tmp = BUF_INITIALIZER;
+    char is_from;
+    static const char * from_header = "From ";
 
     while (size) {
         n = prot_read(from, buf, size > 4096 ? 4096 : size);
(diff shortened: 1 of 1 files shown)


imap/message.c
```diff
@@ -190,6 +190,8 @@ EXPORTED int message_copy_strict(struct protstream *from, FILE *to,
     int munge8bit = config_getswitch(IMAPOPT_MUNGE8BIT);
     int inheader = 1, blankline = 1;
     struct buf tmp = BUF_INITIALIZER;
+    char is_from;
+    static const char * from_header = "From ";
```
dilyanpalauzov, line 194: This shall be changed from *char to char[], so that the string is moved in readonly section and shared by all libimap callers.

imap/message.c
```diff
@@ -305,13 +307,18 @@ EXPORTED int message_copy_strict(struct protstream *from, FILE *to,
...
+	    is_from = (*p == from_header[0])?0:1;
+            for (; *p != ':' && is_from < 5; p++) {
+		if (is_from > 0 && *p != from_header[is_from]) {
+                    is_from = -1;
+		}
+                if (is_from >= 0) {
+                    is_from++;
+                } else {
+                    if (*p <= ' ') return IMAP_MESSAGE_BADHEADER;
+		}
```
dilyanpalauzov, line 321: This leaks `tmp` and uses inconsistent indentation.

elliefm: Looks like this patch predates 3.0, and therefore uses mixed tabs/spaces.  It can probably be fixed up easily enough with tools/remove-tabs.pl, assuming it uses mixed tabs/spaces correctly.

But more to the point though, what is this logic doing?  

The description says we're trying to accept bad "From:" headers rather than reject them, but note that it's using "From " (space, no colon)... this is not "From:" headers as an MTA might see, but a "From " line as you would find in an mbox-format (single file) mail store.

It seems to set up `from_header` as an array containing "From " to compare to, so that it can skip the error message when processing a "From " but preserve the error message when parsing anything else.  But the extant `strstr(buf, "From ")` check already does this...

I think this patch is actually a really terrible implementation of what was better implemented by 1c918cbc40.  I don't think this patch is needed for anything anymore, and Debian should simply drop it.

If it turns out that there is a real problem that is somehow solved by this patch, I would like to see a proper description of the bad behaviour, so that we can develop a good fix (with tests!)

guimard: Thanks @elliefm, dropped : https://salsa.debian.org/debian/cyrus-imapd/-/commit/5fff1454b

PR #3283, opened 2020-11-09
Normalize the authentication ID

guimard:
Debian author comment:
> By normalize, it is intended that;
>  1. Authentication IDs all can be lowercased for more accurate comparison without being volatile to, say, user error, and
>  2. Any leading or trailing blank space can be stripped

### DIFF
--- a/imap/global.c
+++ b/imap/global.c
@@ -346,6 +346,8 @@ EXPORTED int cyrus_init(const char *alt_config, const char *ident, unsigned flag
                                   config_getswitch(IMAPOPT_UNIX_GROUP_ENABLE));
         libcyrus_config_setswitch(CYRUSOPT_USERNAME_TOLOWER,
                                   config_getswitch(IMAPOPT_USERNAME_TOLOWER));
+	libcyrus_config_setswitch(CYRUSOPT_NORMALIZEUID,
+				  config_getswitch(CYRUSOPT_NORMALIZEUID));
         libcyrus_config_setswitch(CYRUSOPT_SKIPLIST_UNSAFE,
                                   config_getswitch(IMAPOPT_SKIPLIST_UNSAFE));
         libcyrus_config_setstring(CYRUSOPT_TEMP_PATH,
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -2919,6 +2919,11 @@ of the incoming network interface, or if no record is found, the
 { "fastmailsharing", 0, SWITCH, "3.0.0" }
 /* If enabled, use FastMail style sharing (oldschool full server paths) */
 
+{ "normalizeuid", 0, SWITCH }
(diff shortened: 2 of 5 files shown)


imap/global.c
```diff
@@ -346,6 +346,8 @@ EXPORTED int cyrus_init(const char *alt_config, const char *ident, unsigned flag
                                   config_getswitch(IMAPOPT_UNIX_GROUP_ENABLE));
         libcyrus_config_setswitch(CYRUSOPT_USERNAME_TOLOWER,
                                   config_getswitch(IMAPOPT_USERNAME_TOLOWER));
+	libcyrus_config_setswitch(CYRUSOPT_NORMALIZEUID,
+				  config_getswitch(CYRUSOPT_NORMALIZEUID));
```
dilyanpalauzov, line 350: Bad indentation.

elliefm: The bad indentation can be fixed easily enough with tools/remove-tabs.pl...

The bigger problem here is calling `config_getswitch()` with a CYRUSOPT_ option rather than an IMAPOPT_ one.  It should probably be `config_getswitch(IMAPOPT_NORMALIZEUID)` like the other lines around it.

I think this is the source of one of the warnings @anatoli26 noticed.

lib/imapoptions
```diff
@@ -2919,6 +2919,11 @@ of the incoming network interface, or if no record is found, the
 { "fastmailsharing", 0, SWITCH, "3.0.0" }
 /* If enabled, use FastMail style sharing (oldschool full server paths) */
 
+{ "normalizeuid", 0, SWITCH }
+/* Lowercase uid and strip leading and trailing blanks. It is recommended
+   to set this to yes, especially if OpenLDAP is used as authentication
+   source. */
+
```
elliefm, line 2926: This "it is recommended" advice directly contradicts the comment around line 165 of auth_unix.c, where it explains that we used to do something naive like this, but then did it better instead...

Also, we already have a standard option that does the same thing, and is enabled by default:

>           username_tolower: 1
>               Convert  usernames to all lowercase before login/authentication.  This is
>               useful with authentication backends which  ignore  case  during  username
>               lookups (such as LDAP).

It doesn't trim leading and trailing whitespace... but if this is a real problem, we should apply a real fix (with tests).

elliefm (commented):
Upon deeper examination, I don't think this patch is useful, and Debian should just drop it (eventually).

I guess you'll probably need some sort of transition plan for users who are using it to change their configuration to use `username_tolower: yes` instead -- though it's on by default, so maybe it just works already and you can drop it immediately with just a note in the changes file.

rsto:
Closing this for the reasons outlined in https://github.com/cyrusimap/cyrus-imapd/pull/3283#discussion_r520961653

PR #3285, opened 2020-11-10
imapd: allow to override search_fuzzy_always by annotation

rsto:
Tested in https://github.com/cyrusimap/cassandane/commit/f23d7ad9b440530f445651b4d204488c8050bf80

### DIFF
--- a/imap/annotate.c
+++ b/imap/annotate.c
@@ -255,6 +255,9 @@ static int annotation_set_mailboxopt(annotate_state_t *state,
 static int annotation_set_pop3showafter(annotate_state_t *state,
                                         struct annotate_entry_list *entry,
                                         int maywrite);
+static int annotation_set_fuzzyalways(annotate_state_t *state,
+                                        struct annotate_entry_list *entry,
+                                        int maywrite);
 static int annotation_set_specialuse(annotate_state_t *state,
                                      struct annotate_entry_list *entry,
                                      int maywrite);
@@ -2248,6 +2251,16 @@ static const annotate_entrydesc_t mailbox_builtin_entries[] =
         NULL,
         NULL
     },{
+        IMAP_ANNOT_NS "search-fuzzy-always",
+        ATTRIB_TYPE_STRING,
+        BACKEND_ONLY,
+        ATTRIB_VALUE_SHARED,
+        0,
+        annotation_get_fromdb,
+        annotation_set_fuzzyalways,
+        NULL,
+        NULL
+    },{
         IMAP_ANNOT_NS "server",
         /* _get_server does its own access control check */
         ATTRIB_TYPE_STRING | ATTRIB_NO_FETCH_ACL_CHECK,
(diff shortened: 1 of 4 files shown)


imap/imapd.c
```diff
@@ -6091,11 +6091,25 @@ static void cmd_search(char *tag, int usinguid)
...
+    struct buf val = BUF_INITIALIZER;
+    if (imapd_id.quirks & QUIRK_SEARCHFUZZY) {
+        /* Quirks overrule anything */
         searchargs->fuzzy_depth++;
     }
-
+    else if (!annotatemore_lookup(inbox, annot, imapd_userid, &val) && val.len) {
+        /* User may override global config */
+        if (config_parse_switch(buf_cstring(&val))) {
+            searchargs->fuzzy_depth++;
```
elliefm, line 6104: -1 is a true value, so if the annotation is set to a junk value that the parser rejects, this `if()` will treat that the same as "yes"

If we want it to treat junk as "not-enabled", then this should be like `if(config_parse_switch(...) == 1)`, but I don't have an opinion about which behaviour we want.

elliefm: Though I see we validate attempts to set it, so I guess it shouldn't ever be junk at this point anyway

imap/annotate.c
```diff
@@ -2247,6 +2250,16 @@ static const annotate_entrydesc_t mailbox_builtin_entries[] =
         annotation_set_pop3showafter,
         NULL,
         NULL
+    },{
+        IMAP_ANNOT_NS "search-fuzzy-always",
+        ATTRIB_TYPE_STRING,
+        BACKEND_ONLY,
+        ATTRIB_VALUE_PRIV,
```
elliefm, line 2257: When we talked about this on the conf call, I thought we concluded that this needed to use a shared value, not a private one?  But I don't understand/remember why (I don't think about annotations much), and maybe I'm misremembering?

imap/imapd.c
rsto: Treating invalid values as true was on purpose. But now I come to realize that this was under the assumption, that search_fuzzy_always is enabled (as we do at Fastmail). I'll change this to use whatever is set in imapd.conf for invalid values.

imap/annotate.c
rsto: I don't remember that we concluded that, but I am fine with either. I just wouldn't see  a use case why one user would be able to toggle the setting for another user.

imap/imapd.c
rsto: Fixed this in latest commit

imap/annotate.c
rsto: Changed to shared annotation.

brong (commented):
Nice!  One tiny nit which I think is good for slightly better API safety, but not a big deal.

imap/annotate.c
```diff
@@ -3356,6 +3369,24 @@ static int annotation_set_pop3showafter(annotate_state_t *state,
...
+                                        int maywrite)
+{
+    struct mailbox *mailbox = state->mailbox;
+
+    assert(mailbox);
+
+    if (!mboxname_isusermailbox(mailbox->name, /*isinbox*/1)) {
+        return IMAP_PERMISSION_DENIED;
+    }
+    if (entry->shared.s && config_parse_switch(buf_cstring(&entry->shared)) < 0) {
```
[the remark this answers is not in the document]
rsto: done

imap/annotate.c
elliefm: I'm not sure either, but I guess it might be so that the admin process setting the flag from the web interface, and the user imapd process doing the search, both see the same annotation value, and not distinct ones private to each of them.

Issue #3272, opened 2020-11-04
JMAP incompatible with Meli because of missing eventSourceUrl


rudihorn:
I was wanting to test Cyrus' JMAP with meli (https://github.com/meli/meli/), but it fails with an error. I think it would be really great if the few existing clients out there could be tested/made to work to increase support for JMAP.

Essentially meli fails because Cyrus does not include the eventSourceUrl in the JMAP session resource. Looking at RF8620 this field seems to be non-optional. While I understand that this is not implemented in Cyrus yet, and that this would be a lot of work, I think it would still be good to publish an event source URL that leads to an HTTP 501 (not implemented) error.

### Justification
I think this is the right approach because meli fails with a parser error, and I don't think it would be right for meli to change its parsing to be compliant with something that diverges from the RFC. Instead it would be better for meli to detect a 501 error on the websocket URL and to then gracefully handle events not working, while still supporting as much functionality as it can.

### Reproduction
~Meli currently requires this patch to reproduce the issue: https://github.com/meli/meli/pull/1~ Connect to Cyrus using master branch of meli.

PR #3286, opened 2020-11-11
jmap_core: add http handler stub for eventSourceUrl

rsto:
JMAP Session object requires a `eventSourceUrl` property, but Cyrus does not implement push events. This patch adds a stub that always returns HTTP 501 (Not Implemented) for the advertised event source URL. Also see #3272

Tested in https://github.com/cyrusimap/cassandane/commit/de17790b8525a3cee0087a15d86f84b8c2463c2a

### DIFF
--- a/imap/http_jmap.c
+++ b/imap/http_jmap.c
@@ -72,6 +72,8 @@
 #define JMAP_UPLOAD_TPL    "{accountId}/"
 #define JMAP_DOWNLOAD_COL  "download/"
 #define JMAP_DOWNLOAD_TPL  "{accountId}/{blobId}/{name}?accept={type}"
+#define JMAP_EVENTSOURCE_COL  "eventsource/"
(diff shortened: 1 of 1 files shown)


ksmurchison (approved):
Haha!  I had started almost the exact same code.

I did find this text in https://www.w3.org/TR/eventsource/

"Clients will reconnect if the connection is closed; a client can be told to stop reconnecting using the HTTP 204 No Content response code."

"HTTP 500 Internal Server Error, 502 Bad Gateway, 503 Service Unavailable, and 504 Gateway Timeout responses, and any network error that prevents the connection from being established in the first place (e.g. DNS errors), must cause the user agent to asynchronously reestablish the connection.

Any other HTTP response code not listed here must cause the user agent to fail the connection."

So, it looks like both 204 and 501 will work, but I'm not sure if one if better than the other.

rsto:
Ha :) 204 seems like the better choice to me. I'll update the patch.

PR #3287, opened 2020-11-11
Ignore . and .. while iterating through the directories holding usernames

DanielO:
This is necessary on FreeBSD otherwise it goes back up a directory and gets lost.


elliefm:
This seems obviously correct, thanks.

elliefm:
Also cherry-picked to 3.2 (as 7e3d8ee1b0) and 3.0 (as 463aaf5482), and will be included in the next release from each

PR #3290, opened 2020-11-16
xapian_wrap: use xsyslog for IOERRORs

elliefm:
Nibbling away at converting existing IOERRORs to use xsyslog.  This time, xapian_wrap.cpp.

There's a few spots where I'm not quite sure what to do and have added XXX comments, feedback/discussion would be appreciated!

Please read closely, but also build/test, just in case your compiler finds problems that mine doesn't.

### DIFF
--- a/imap/xapian_wrap.cpp
+++ b/imap/xapian_wrap.cpp
@@ -514,7 +514,8 @@ int xapian_compact_dbs(const char *dest, const char **sources)
             // Aggregate language counts.
             r = calculate_language_counts(subdb, lang_counts);
             if (r) {
-                syslog(LOG_ERR, "IOERROR: Xapian: compact %s: corrupt language metadata", thispath);
+                xsyslog(LOG_ERR, "IOERROR: corrupt language metadata",
+                                 "path=<%s>", thispath);
                 return r;
             }
         }
@@ -535,8 +536,9 @@ int xapian_compact_dbs(const char *dest, const char **sources)
         write_language_counts(newdb, lang_counts);
     }
     catch (const Xapian::Error &err) {
-        syslog(LOG_ERR, "IOERROR: Xapian: caught exception compact_dbs: %s (%s)",
-                err.get_description().c_str(), thispath);
+        xsyslog(LOG_ERR, "IOERROR: caught exception",
+                         "exception=<%s> path=<%s>",
+                         err.get_description().c_str(), thispath);
         r = IMAP_IOERROR;
     }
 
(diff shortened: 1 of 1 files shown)


elliefm:
Note to future ellie: this is how I'm prioritising which files to target:

```
ack -cl ' syslog\(LOG_ERR, "IOERROR' | sort -nr -t: -k2 | less
```

rsto (commented):
Cool, thanks. I commented on a couple of the XXX open questions.

imap/xapian_wrap.cpp
```diff
@@ -1904,8 +1921,10 @@ xapian_query_new_match(const xapian_db_t *db, int partnum, const char *str)
 
     int min_version = *db->db_versions->begin();
     if (min_version < XAPIAN_DB_MIN_SUPPORTED_VERSION) {
-        syslog(LOG_ERR, "IOERROR: Xapian: db versions < %d are deprecated. Reindex your dbs.",
-                XAPIAN_DB_MIN_SUPPORTED_VERSION);
+        // XXX is this an IOERROR? seems like more of a warning, since we keep going?
```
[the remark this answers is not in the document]
elliefm: Great.  I'm changing this to a warning, and also adding `paths=<%s>`, so the admin noticing the warning knows which databases need reindexing...

imap/xapian_wrap.cpp
```diff
@@ -2038,14 +2061,18 @@ int xapian_query_run(const xapian_db_t *db, const xapian_query_t *qq, int is_leg
 
             /* ignore documents with no cyrusid.  Shouldn't happen, but has been seen */
             if (cyrusid.length() != 43) {
-                syslog(LOG_ERR, "IOERROR: Xapian: zero length cyrusid for document id %u in index files %s",
-                                d.get_docid(), db->paths->c_str());
+                // XXX paths seems to be plural here?
```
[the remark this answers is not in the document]
elliefm: Great, thanks for the clarification.  I won't get rid of it in this PR to keep the scope narrow (esp since I've needed to add it to another xsyslog myself), but we can always refactor it later.

imap/xapian_wrap.cpp
```diff
@@ -1110,6 +1119,7 @@ int add_text_part(xapian_dbw_t *dbw, const struct buf *part, int partnum)
         if (config_getswitch(IMAPOPT_SEARCH_INDEX_LANGUAGE)){
             // Index by language.
 #ifndef HAVE_CLD2
+            // XXX is this really an "IOERROR"?
             syslog(LOG_ERR, "IOERROR: Xapian: language indexing requires CLD2 library");
             return IMAP_IOERROR;
 #else
```
elliefm, line 1125: @rsto what do you think about this?

I feel like this sort of "we weren't compiled to support this config option" error would usually be a `fatal()` rather than a syslog-and-return-error.  But, we're in C++ code here, and maybe we deliberately don't `fatal()` because it's unsafe?  So I think syslog-and-return-error is correct, but thought I'd double-check...

rsto: We have a write-opened Xapian database open at this stage, so I chose to abort gracefully. Also: this codepath is called in two codepaths mostly:
1. squatter during indexing: in which case the return error should bubble up all to squatter and the process ends with an error code
2. sieve scripts which use the jmap extensions, in which case it should better be handled in sieve how to handle that message (e.g. just put in the inbox).

elliefm: Great, that makes sense, thanks for clarifying. :)

PR #3292, opened 2020-11-20
imapd: maxsize for appends

brong:


ksmurchison (commented):
This is similar but slightly different from #3288

brong:
g;dammit - sorry Ken.  I didn't notice that.  Sorry for duplicating your work!

I've done a bunch more on this one now which avoids spooling the message in the first place, so I think this is probably better - but I'll look into whether we should copy the existing option rather than making a new one.

ksmurchison:
In mine, I fetch the maxmessagesize in service_init() and make it global, so I don't have to pass it around.  I did miss checking the literal size early however

ksmurchison (commented):
I thought we were going to use the existing maxmessagesize option

imap/imapd.c
```diff
@@ -3828,6 +3828,8 @@ static void cmd_append(char *tag, char *name, const char *cur_name)
     const char *parseerr = NULL, *url = NULL;
     struct appendstage *curstage;
     mbentry_t *mbentry = NULL;
+    size_t maxsize = config_getint(IMAPOPT_APPEND_MAXSIZE) * 1024;
+    if (!maxsize) maxsize = UINT32_MAX;
```
dilyanpalauzov, line 3833: Would the call to config_getint be reduced to just one for the whole program lifetime, if maxsize is declared `static`? (as opposed to executing config_getint anytime cmd_append is called)
```c
static size_t maxsize;
if (!maxsize) {
  maxsize = config_getint(IMAPOPT_APPEND_MAXSIZE) * 1024;
  if (!maxsize) maxsize = UINT32_MAX;
}

brong: The cost of cmd_append is so high that this kind of optimisation isn't worth making the code more complex for.  Optimisations in tight loops are great, but optimisations at this level aren't worth it.

ksmurchison:
@brong Added APPENDLIMIT= capability.  Any reason we can't merge this?

dilyanpalauzov:
In docsrc/imap/rfc-support.rst RFC 7889 should be moved from Wishlist to Supported.

PR #3293, opened 2020-11-20
append.c: if we have an existing message with the same GUID, hardlink…

ksmurchison:
… to it rather than duplicating the content

### DIFF
--- a/imap/append.c
+++ b/imap/append.c
@@ -865,6 +865,33 @@ out:
     return r;
 }
 
+struct findstage_cb_rock {
+    const char *partition;
+    char *stagefile;
+    size_t size;
+};
+
+static int findstage_cb(const conv_guidrec_t *rec, void *vrock)
+{
+    struct findstage_cb_rock *rock = vrock;
+    mbentry_t *mbentry = NULL;
+
+    int r = mboxlist_lookup(rec->mboxname, &mbentry, NULL);
+    if (r) return 0;
+
+    if (!strcmp(rock->partition, mbentry->partition)) {
+        strlcpy(rock->stagefile,
+                mboxname_datapath(mbentry->partition, mbentry->name,
+                                  mbentry->uniqueid, rec->uid),
+                rock->size);
+        r = CYRUSDB_DONE;
+    }
+
+    mboxlist_entry_free(&mbentry);
+
+    return r;
+}
+
 /*
  * staging, to allow for single-instance store.  the complication here
  * is multiple partitions.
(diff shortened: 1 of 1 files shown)


imap/append.c
```diff
@@ -908,6 +935,30 @@ EXPORTED int append_fromstage_full(struct appendstate *as, struct body **body,
...
     }
 
+    if (!nolink) {
+        /* attempt to find an existing message with the same guid
+           and use it as the stagefile */
+        struct conversations_state *cstate, *mycstate = NULL;
+
+        cstate = conversations_get_mbox(mailbox->name);
+        if (!cstate) {
+            r = conversations_open_mbox(mailbox->name, 1/*shared*/, &mycstate);
```
brong, line 945: This will need to be 0/*shared*/ because we're planning to write to the mailbox.  I think we can just use:

```cstate = mailbox_get_cstate(mailbox);```

and avoid all the mycstate guff.  It will get the correct type anyway, given that the mailbox will be opened read-write.

File: imap/append.c [hunk 2]
```diff
@@ -865,6 +865,33 @@ static int append_apply_flags(struct appendstate *as,
...
+    int r = mboxlist_lookup(rec->mboxname, &mbentry, NULL);
+    if (r) return 0;
+
+    if (!strcmp(rock->partition, mbentry->partition)) {
+        strlcpy(rock->stagefile,
+                mboxname_datapath(mbentry->partition, mbentry->name,
+                                  mbentry->uniqueid, rec->uid),
+                rock->size);
+        r = CYRUSDB_DONE;
+    }
```
brong, line 888: I think we should be hard-linking the file into the spool path rather than setting the stagefile to be directly the path inside the mailbox.

imap/append.c
```diff
@@ -865,6 +865,33 @@ static int append_apply_flags(struct appendstate *as,
...
+    const char *partition;
+    char *stagefile;
+    size_t size;
+};
+
+static int findstage_cb(const conv_guidrec_t *rec, void *vrock)
+{
+    struct findstage_cb_rock *rock = vrock;
+    mbentry_t *mbentry = NULL;
+
```
brong, line 878: This needs a

```if (rec->part) return 0;```

Otherwise we could wind up pointing to an email which has this GUID as an attachment by mistake.

brong (changes requested):

brong (approved):
Yeah, looks good - I'll get this into FM builds

PR #3295, opened 2020-11-25
configurable special-use mailboxes that may not contain children

elliefm:
Adds a "specialuse_nochildren" imapd.conf option and a corresponding check in `mboxlist_create_namecheck()`.  If the immediate parent of the mailbox to be created has a special-use attribute listed in `specialuse_nochildren`, the create is rejected (IMAP_PERMISSION_DENIED).

This restriction does not apply to admins (so replication to a server with different configuration can still succeed), but does apply to ordinary users with implicit admin rights to their own mailboxes.

To-do: we talked about restricting attempts to setmetadata special-use attributes in this list on mailboxes that already have children.  I'm not yet sure where to hook this in; guidance appreciated!

Open questions:
* should we somehow prevent/complain/ignore if "\Inbox" is set in this list?
* any weird cases I haven't anticipated/tested for?

Corresponding Cassandane tests at https://github.com/cyrusimap/cassandane/compare/master...elliefm:no-junktrash-subfolders

### DIFF
--- a/imap/annotate.c
+++ b/imap/annotate.c
@@ -3466,6 +3466,23 @@ EXPORTED int specialuse_validate(const char *mboxname, const char *userid,
             }
         }
 
+        /* some attributes may not be set on mailboxes containing children */
+        if (mboxname && config_getstring(IMAPOPT_SPECIALUSE_NOCHILDREN)) {
+            strarray_t *forbidden = strarray_split(
+                config_getstring(IMAPOPT_SPECIALUSE_NOCHILDREN),
+                NULL,
+                STRARRAY_TRIM
+            );
+
+            if (strarray_find(forbidden, strarray_nth(valid, j), 0) != -1
+                && mboxlist_haschildren(mboxname))
+            {
+                r = IMAP_MAILBOX_HASCHILDREN;
+            }
+
+            strarray_free(forbidden);
+        }
+
         /* normalise the value */
         strarray_set(new_attribs, i, strarray_nth(valid, j));
     }
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -1058,6 +1058,38 @@ static int mboxlist_create_namecheck(const char *mboxname,
             r = quota_check_useds(root, qdiffs);
             if (r) goto done;
         }
+
+        /* make sure parent isn't forbidden from containing children */
+        if ((!isadmin || mboxname_userownsmailbox(userid, mboxname))
+            && config_getstring(IMAPOPT_SPECIALUSE_NOCHILDREN))
+        {
+            struct buf attrib = BUF_INITIALIZER;
+            mbname_t *mbname;
+
+            mbname = mbname_from_intname(mbentry->name);
+            annotatemore_lookup(mbentry->name, "/specialuse",
+                                mbname_userid(mbname), &attrib);
+            mbname_free(&mbname);
+
+            if (buf_len(&attrib)) {
+                strarray_t *uses = strarray_split(buf_cstring(&attrib), NULL, 0);
+
+                strarray_t *forbidden = strarray_split(
+                    config_getstring(IMAPOPT_SPECIALUSE_NOCHILDREN),
+                    NULL,
+                    STRARRAY_TRIM
+                );
+
+                if (strarray_intersect(uses, forbidden))
+                    r = IMAP_PERMISSION_DENIED;
+
+                strarray_free(forbidden);
+                strarray_free(uses);
+            }
+
+            buf_free(&attrib);
+            if (r) goto done;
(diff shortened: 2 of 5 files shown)


lib/strarray.c
```diff
@@ -371,15 +371,8 @@ EXPORTED strarray_t *strarray_splitm(char *buf, const char *sep, int flags)
...
-            if (flags & STRARRAY_UNIQ) {
-                int found = -1;
-                if ((flags & STRARRAY_LCASE))
-                    found = strarray_find_case(sa, p, 0);
-                else
-                    found = strarray_find(sa, p, 0);
-
-                if (found != -1) continue;
-            }
+            if ((flags & STRARRAY_UNIQ) && strarray_find(sa, p, 0) != -1)
```
rsto, line 374: nit: it did not occur to me that splitm with UNIQ is O(n^2) in disguise. But I guess since this hasn't shown to be a problem so far, it is fine.

imap/mboxlist.c
```diff
@@ -1063,8 +1063,12 @@ static int mboxlist_create_namecheck(const char *mboxname,
         /* XXX should isadmin override this? */
         if (!config_getswitch(IMAPOPT_ALLOWJUNKTRASHSUBFOLDERS)) {
             struct buf attrib = BUF_INITIALIZER;
+            mbname_t *mbname;
 
-            annotatemore_lookup(mbentry->name, "/specialuse", userid, &attrib);
+            mbname = mbname_from_intname(mbentry->name);
```
rsto, line 1069: nit: why not declare mbname_t *mbname here, rather than two lines above?

rsto (approved):
Looks good to me, only found two nits (of which the strarray UNIQ flag nit most likely will be a no-fix). The only thing I found confusing is that multiple commits changing the same code aren't squashed :)

lib/strarray.c
elliefm: I would like to have a `strset` type (probably built on `hash_table`) and convert all the existing code that uses strarrays as a set to use that instead.  But that use case clearly hasn't been a problem so far (same algorithm as `strarray_intersect()`), so I've consciously just gone with the simplest thing for now, rather than waste time trying to bolt an awkward optimisation onto what is already the wrong data type.

If we find this is a problem in practice, that'll be a good prompt to do sets properly...

imap/mboxlist.c
elliefm: Because I like to keep my variable declarations at the top of the block, so I know where to look for them, and I have a nice neat list of things to make sure are cleaned up at the end of the scope.  The C++/C99 style of scattering them around just wherever is very annoying and adds a lot of mental effort to memory management.

elliefm:
I think I've found where to hook in the prevention of adding these attributes to mailboxes that already have children (`specialuse_validate()`), and have hooked it in.  I'm calling it a "mailbox already exists" error, but if anyone has a better idea, sing out.

Updated tests to check this case too.  It seems to do the right thing under replication (i.e. the annotations are replicated, despite them being forbidden on the replica).  But I didn't specifically handle this case in the implementation, so I guess `specialuse_validate()` is bypassed somehow for replication?

imap/annotate.c
```diff
@@ -3466,6 +3466,24 @@ EXPORTED int specialuse_validate(const char *mboxname, const char *userid,
...
+                config_getstring(IMAPOPT_SPECIALUSE_NOCHILDREN),
+                NULL,
+                STRARRAY_UNIQ
+            );
+
+            if (strarray_find(forbidden, strarray_nth(valid, j), 0) != -1
+                && mboxlist_haschildren(mboxname))
+            {
+                /* XXX what kind of error should this be? */
+                r = IMAP_MAILBOX_EXISTS;
```
elliefm, line 3481: Hmm, maybe this should be MAILBOX_HASCHILDREN??

lib/strarray.c
brong: Meh - 'n' is approximately exactly equal to one in most of these cases.

elliefm:
Something about this was bothering me, and I finally figured out what: our naive `strarray_intersect()` algorithm doesn't benefit from uniqueness, so ensuring the values are unique first is totally wasted effort.

So I've ripped out my STRARRAY_UNIQ implementation, and also squashed out the original junk-or-trash-only commits.  I've kept the new strarray_split_lcase test (but without the uniqueness test), because the extra coverage is still useful...

I think I'm finally done fiddling with this.  I'd like to tag this include-in-fastmail, and not merge it to master until we've used it ourselves and are confident there's nothing else needed.

brong:
This has been running in prod for Fastmail customers for a while with no complaints!

PR #3296, opened 2020-11-25
JMAP: added JMAP_READ_ONLY and JMAP_NO_CSTATE flags

ksmurchison:


imap/jmap_notes.c
```diff
@@ -84,19 +84,19 @@ static jmap_method_t jmap_notes_methods_nonstandard[] = {
...
         &jmap_notes_get,
-        JMAP_SHARED_CSTATE
+        JMAP_NO_CSTATE|JMAP_READ_ONLY
     },
     {
         "Notes/set",
         JMAP_NOTES_EXTENSION,
         &jmap_notes_set,
-        /*flags*/0
+        JMAP_READ_ONLY
```
rsto, line 93: Notes/set requires an exclusive write lock on conversation.db but is read-only?

imap/jmap_api.h
```diff
@@ -191,7 +191,9 @@ extern void jmap_finireq(jmap_req_t *req);
 
 extern int jmap_is_using(jmap_req_t *req, const char *capa);
 
-#define JMAP_SHARED_CSTATE 1 << 0
+#define JMAP_READ_ONLY     (1 << 0)
```
rsto, line 194: I find this READ_ONLY flag confusing. I guess it means that methods flagged with this are safe to call in a read-only instance? A possible source of subtle errors is that nothing really enforces the read-only part, in contrast to a shared cstate which will certainly bail if the method implementation attempts to create messages or the like. I guess I wonder if we can enforce READ_ONLY on the lower levels, such that no annotations, mailboxes, messages etc can be altered at all.

rsto (commented):
I like this, but I found the READ_ONLY flag a bit confusing. The Notes/set comment is the only one which might require a code change.

rsto (approved):
Looks good to me!

PR #3297, opened 2020-11-30
fix sync_client crash when racing against script updates

elliefm:
Found by Sentry.  What happened was that sync_client went to build a list of the current sieve scripts for a user, while an update was being applied to (one of) that user's script(s).  `sieve_foreach()` saw the "websieve.script.NEW" in `readdir()` and passed it through to the callback; the callback called `sievedir_get_script()` on it, which returned NULL because that filename was already gone; and then the callback crashed because it didn't expect a NULL there.

This does three things:

* The `list_cb()` callback now ignores NULL returns from `sievedir_get_script()` instead of crashing (but perhaps it should also log them?)
* The `sievedir_foreach()` function now ignores filenames ending in ".NEW" -- I believe we don't want this API to see them.  Is that right?
* Fixes an unrelated memory leak in `sievedir_get_script()` when it returns NULL.

This seems to be the smallest change to fix this particular race/crash, but perhaps what we really want is some locking here, I'm not sure.  What do you think?

### DIFF
--- a/imap/sievedir.c
+++ b/imap/sievedir.c
@@ -87,10 +87,14 @@ EXPORTED int sievedir_foreach(const char *sievedir,
 
     while ((dir = readdir(dp)) != NULL) {
         const char *name = dir->d_name;
+        const char *dot;
         char target[PATH_MAX] = "";
         struct stat sbuf;
 
         if (!strcmp(name, ".") || !strcmp(name, "..")) continue;
+
+        /* ignore transient .NEW files */
+        if ((dot = strrchr(name, '.')) && !strcmp(dot, ".NEW")) continue;
 
         strlcpy(path + dir_len, name, sizeof(path) - dir_len);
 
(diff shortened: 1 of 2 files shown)


elliefm:
No tests because I couldn't think of a way to reliably reproduce the race. Open to suggestions!

imap/sievedir.c
```diff
@@ -166,7 +166,10 @@ EXPORTED struct buf *sievedir_get_script(const char *sievedir,
     buf_printf(&buf, "%s/%s", sievedir, script);
 
     int fd = open(buf_cstring(&buf), 0);
```
ksmurchison, line 172: Should we just buf_free(&buf) here since we do it whether open() succeeds or fails

imap/sievedir.c
```diff
@@ -87,11 +87,15 @@ EXPORTED int sievedir_foreach(const char *sievedir,
...
     while ((dir = readdir(dp)) != NULL) {
         const char *name = dir->d_name;
+        const char *dot;
         char target[PATH_MAX] = "";
         struct stat sbuf;
 
         if (!strcmp(name, ".") || !strcmp(name, "..")) continue;
 
+        /* ignore transient .NEW files */
+        if ((dot = strrchr(name, '.')) && !strcmp(dot, ".NEW")) continue;
```
ksmurchison, line 97: In theory this would match foo.NEW.bar.  We have to do enough comparisons of string suffixes that we should probably create a strrncmp() or something

PR #3298, opened 2020-12-01
Sieve in mailbox

ksmurchison:
This passes all Cass tests when using https://github.com/cyrusimap/cassandane/pull/104

I still have work to do with jmap_vacation.c, sieve_rebuild(), and backup/*.c (need to discuss with @elliefm )

I will also remove all legacy sieve code from sync_*.c


lib/imapoptions
```diff
@@ -2405,6 +2405,9 @@ product version in the capabilities
    option has no effect on options that are disabled at compile time
    (e.g., "regex"). */
 
+{ "sieve_folder", "#sieve", STRING, "3.3.1" }
```
elliefm, line 2408: This needs to be "3.3.2" now

sieve/sievec.c
```diff
@@ -91,7 +91,7 @@ int main(int argc, char **argv)
         exit(1);
     }
 
-    instream = fopen(argv[optind++],"r");
+    instream = !strcmp(argv[optind], "-") ? stdin : fopen(argv[optind], "r");
     if(instream == NULL) {
         fprintf(stderr, "Unable to open %s for reading\n", argv[1]);
```
elliefm, line 96: Unrelated to this branch, but while you're in the area anyway, `argv[1]` in this error message should be `argv[optind]`

lib/imapoptions
ksmurchison: Fixed

sieve/sievec.c
ksmurchison: Fixed in master and rebased

lib/imapoptions
elliefm: Ironically, this needs to be 3.5.0 now... but don't worry about it, we can fix it properly when it actually lands on master, it doesn't matter in the meantime.

I'm thinking about a way to make this less of a nuisance for future PR's, and I think I have an idea to play around with.

elliefm (approved):
This seems okay to me.  I've run it through Cassandane with the tests from cyrusimap/cassandane#104 and it all passes.

I've also run it through Cassandane with a 3.2 main instance and this build as the replica, and all the replication tests pass fine, so that's good.

brong (approved):
This all looks really sensible so far.

Yes, stripping the sieve shipping code from sync_client will be good once this is done.

I put some notes about clashes with the dav_reconstruct work which hit the same area dammit :(

ksmurchison:
Superseded by sieve-in-uuid-mailbox

PR #3302, opened 2020-12-07
mappedfile: use xsyslog for IOERRORs

elliefm:
More xsyslog conversions...

I'm particularly interested in thoughts about the lseek errors' "offset" argument.  At first blush I thought the casting was just an old workaround for not having a format string for `off_t`, so I converted them to use `OFF_T_FMT`.  But I've since noticed they were printing in hex too, which is lost by using `OFF_T_FMT`.

So the question is, do we care about these offsets being printed as hex?  It's worth observing that `off_t` is signed, and the signedness is lost when we print as hex...

### DIFF
--- a/lib/mappedfile.c
+++ b/lib/mappedfile.c
@@ -122,14 +122,16 @@ EXPORTED int mappedfile_open(struct mappedfile **mfp,
         }
         r = cyrus_mkdir(mf->fname, 0755);
         if (r < 0) {
-            syslog(LOG_ERR, "IOERROR: cyrus_mkdir %s: %m", mf->fname);
+            xsyslog(LOG_ERR, "IOERROR: cyrus_mkdir failed",
+                             "filename=<%s>", mf->fname);
             goto err;
         }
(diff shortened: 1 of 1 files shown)


ksmurchison (approved):
I personally don't care if the offset is no longer in hex.  I'm not even sure why it was in hex in the first place.  And if we can't convert between the two if we need one or the other, we probably shouldn't be doing this job, right?

elliefm:
Yeah, right?  I'm trying to think why hex would be useful, and I suppose if you wanted to investigate possible file corruption (such as a header record pointing to a bad offset that can't be seeked to), hex values would be what you're searching for in hexdump, so it'd save you a step if you didn't have to convert.  On the other hand, if you were comparing the file size to the bad offset, decimal would save you a conversion.

@rjbs @brong Do you have an opinion here from a plumbing perspective?  I don't have a strong opinion either way, so I'll just do whatever's most useful to whoever actually cares.  And if no-one cares I'll leave it as is, with the unnecessary casts cleaned up and done "right".

brong:
I don't care either way

PR #3303, opened 2020-12-08
lmtp: require ACL_INSERT for fileinto, and ACL_POST for keep.  delive…

ksmurchison:
…ry/keep to user's INBOX is always allowed

### DIFF
--- a/imap/lmtp_sieve.c
+++ b/imap/lmtp_sieve.c
@@ -55,6 +55,7 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 
+#include "acl.h"
 #include "annotate.h"
 #include "append.h"
 #include "assert.h"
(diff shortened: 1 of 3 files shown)


elliefm (approved):
Looks good

brong:
@murch What is the goal of this MR?  lmtp runs as a different user than the mailbox owner, so we would either need to explicitly grant insert to that user for every mailbox, or change that as well.

ksmurchison:
@brong  Per a FM task "set up ACLs such that users can't sieve deliver into other users where they don't otherwise have mailbox insert access"

brong (changes requested):
The reason I haven't approved this is that I don't think the concept is right.

The delivery happens in the context of the delivery user, which is `postman` or something.  If we require user `postman` to have ACL_INSERT on every mailbox that a user can deliver into, then we would need to set up the ACLs correctly for that on every user - both at Fastmail and also tell everyone else upgrading to do so at their site.

Alternatively, we could make a sieve fileinto happen in the context of the mailbox owner rather than `postman` - but we need to think about the implications of that too.

rsto:
This is not required anymore.

PR #3304, opened 2020-12-08
JMAP Email/query: support substring search in headers

rsto:
This patch changes Email/query and JMAP sieve header search to support substring search using operators `startsWith`, `endsWith` and `contains`, which may be specified as a third non-standard header query value. It matches the search and header values by first converting them to canonicalized search form, including Unicode NFC normalization.

Currently, header search in JMAP uses different codepaths, depending on the header and configuration. If a header is a subject or email sender or recipient, it falls back to fuzzy search in Xapian. If instead the header is not one of these special headers, but part of the cache, then a substring match (without NFC) is performed. If however the instance is configured to index all headers in Xapian, then it falls back to fuzzy matching arbitrary header values.

This patch changes that to consistently use JMAP header search for all header queries that use the `header` filter criteria.

Please do note that this merge request is marked as draft. It is ready for review, but I want to revisit the code one last time before merge.

Tested in https://github.com/cyrusimap/cassandane/commit/2adaa2851a292bfbe61616bce11043d6e7af08bd


rsto:
Now updated with further improvements in the charset pipeline: Unicode NFC normalization is now stream-based, reducing the number of required heap-allocations. Similarly, a new charset structure allows for reuse of the charset pipeline internal states, which allows to use the same conversion pipeline for all traversed messages.

PR #3309, opened 2020-12-17
cyrusdb_skiplist: use xsyslog for errors

elliefm:
More of this.  As usual, please try building this too, in case your compiler finds something mine didn't.  Thanks!

### DIFF
--- a/lib/cyrusdb_skiplist.c
+++ b/lib/cyrusdb_skiplist.c
@@ -253,7 +253,8 @@ static int myinit(const char *dbdir, int myflags)
         xclose(fd);
 
         if (r == -1) {
-            syslog(LOG_ERR, "DBERROR: writing %s: %m", sfile);
+            xsyslog(LOG_ERR, "DBERROR: write failed",
+                             "filename=<%s>", sfile);
             xclose(fd);
             return CYRUSDB_IOERROR;
         }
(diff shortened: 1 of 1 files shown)


lib/cyrusdb_skiplist.c
```diff
@@ -1336,7 +1350,8 @@ static int mystore(struct dbengine *db,
     lseek(tid->syncfd, tid->logend, SEEK_SET);
     r = retry_writev(tid->syncfd, iov, num_iov);
     if (r < 0) {
-        syslog(LOG_ERR, "DBERROR: retry_writev(): %m");
+        /* XXX is there a filename or something we can log here */
+        xsyslog(LOG_ERR, "DBERROR: retry_writev failed", NULL);
```
ksmurchison, line 1354: I think we can use db->fname here

lib/cyrusdb_skiplist.c
```diff
@@ -1427,7 +1442,8 @@ static int mydelete(struct dbengine *db,
         /* update end-of-log */
         r = retry_write(tid->syncfd, (char *) writebuf, 8);
         if (r < 0) {
-            syslog(LOG_ERR, "DBERROR: retry_write(): %m");
+            /* XXX is there a filename or something we can log here */
+            xsyslog(LOG_ERR, "DBERROR: retry_write failed", NULL);
```
ksmurchison, line 1446: I think we can use db->fname here

ksmurchison (changes requested):
This looks good (pending my 2 inline comments) and compiles cleanly for me

elliefm:
Thanks.  I've updated those two lines to log the db->fname.  Trivial changes, but we're also not in any hurry here... can you give it another compile check once you're back from break, just in case?

brong:
Gonna add this to our build, assuming that's all happy we'll merge it :)

rsto (commented):
Looks good to me.

PR #3312, opened 2020-12-23
Fix two use after free segfaults

hamarituc:
This PR resolves two use after free conditions in cyrus-imapd 2.4 concerning the libsasl callback data structures, which lead to arbitrary segfaults.

### DIFF
--- a/imap/backend.c
+++ b/imap/backend.c
@@ -284,7 +284,6 @@ static int backend_authenticate(struct backend *s, struct protocol_t *prot,
     struct sockaddr_storage saddr_l, saddr_r;
     char remoteip[60], localip[60];
     socklen_t addrsize;
-    int local_cb = 0;
     char buf[2048], optstr[128], *p;
     const char *mech_conf, *pass;
 
@@ -301,8 +300,8 @@ static int backend_authenticate(struct backend *s, struct protocol_t *prot,
     if(iptostring((struct sockaddr *)&saddr_l, addrsize, localip, 60) != 0)
 	return SASL_FAIL;
 
+    s->sasl_cb = NULL;
     if (!cb) {
-	local_cb = 1;
 	strlcpy(optstr, s->hostname, sizeof(optstr));
 	p = strchr(optstr, '.');
 	if (p) *p = '\0';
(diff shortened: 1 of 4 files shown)


elliefm:
Thanks for the PR.  It looks like the later branches have similar code already, so I think we don't need to forward-port this fix, because it only affects 2.4.  Is that correct?

It looks okay to me, but I'm just back from break and not quite in the swing of things.  @ksmurchison can you also cast an eye over this once you're back please?  Ta

Is there a CVE for this?  And if not, does there need to be?

hamarituc:
> Thanks for the PR. It looks like the later branches have similar code already, so I think we don't need to forward-port this fix, because it only affects 2.4. Is that correct?

Concerning `imap/backend.c` the issue is fixed in later versions. But looking into `backend_authenticate()` the callback data structure isn't freed during error handling. This leads to an memory leak if one of the calls to `sasl_client_new()` or `sasl_setprop()` returns an error code. Furthermore `imap/mupdate-client.c` contains a similiar issue, which is present in all versions beyond 2.4.

So from my point of view it seems advisable to forward-port the respective changes to newer versions. Is this done by the code maintainers or do I have to prepare a pull request for each version? Unfortunately I am not able to test the patch with newer versions in our environment.

> Is there a CVE for this? And if not, does there need to be?

We didn't requested a CVE. I cannot judge whether it is worth of it. We experienced approx. 10 to 20 crashes per day (in a setup 2 servers, 15000 users), so it's at least a denial of service issue. I am unsure if the bug is exploitable in terms of a remote code execution. But since the corrupted data structure contains function pointers, there is a at least a chance for an attacker to alter the execution flow. I would leave this decision to your assessment.

elliefm:
Based on where this occurs (only when some Cyrus process is cleaning up a connection it made to some other Cyrus process; and due to mishandling of cleanup, not due to any particular user action), I don't think this needs a CVE.  Each Cyrus process only services one client at a time, so a hypothetical attacker would mostly just deny their own service.  I think, for the crash to be exploitable in some way, there would need to be some other exploit chained in that allowed an attacker to control the Cyrus processes' cleanup choices, and if there were, then that would be the problem, not the crash.

> So from my point of view it seems advisable to forward-port the respective changes to newer versions. Is this done by the code maintainers or do I have to prepare a pull request for each version? Unfortunately I am not able to test the patch with newer versions in our environment.

That's fine, we can take care of it.  I've split it into its own issue (#3320) so it doesn't get forgotten when this one gets merged.

elliefm (approved):

PR #3316, opened 2021-01-05
squat: fix broken squat engine integration for indexed search

### LINKED ISSUE #2598: squat search_engine not used
TL;DR Searches in cyrus 3.0.8 are slower compared to 2.4.20 and there 
is no indication that cyrus.squat files are used.

I reported this issue on info-cyrus Mailinglist on 17. Sep 2018 with the subject 
"squatter not used after upgrade to cyrus 3.0.8" and reported further test results  
on 25. Oct 2018 

We use cyrus-imapd 3.0.8 in murder setup. (I don't think that "murder" is relevant  
in this case. All test were carried out  on the backend)

We have configured "search_engine: squat" in the backend imapd.conf, 
"search_fuzzy_always:" is NOT enabled (I tested it, but it didn't improve the 
performance and showed wrong results)

"squatter -i" is run in the EVENT section of backend cyrus.conf and the logfiles show 
that squatter is indexing the mailboxes.

Searching for HEADERS or TEXT result in slow searches, compared to cyrus-imapd 2.4.20.
And running strace on the imapd process show that the cyrus.squat files are not 
opend/mmaped and that no attempt was made (no fstat either) while these files 
where accessed in 2.4.20.

Timing Tests on a Mailbox with 4321 Mails (~ 541 MB) from Strace (read "SE SELECT INBOX" 
to write "* BYE LOGOUT received")

SE SELECT INBOX
TS SEARCH TEXT "Test1"    
or  
HS SEARCH HEADER X-comment Unirundmail
L LOGOUT

Cyrus Version | TEXT | HEADER
------------------ | --------- | --------------
2.4.20 | 0.025194s | 0.016047s
3.0.8 | 10.253365s | 1.453782s

[search_text_3.0.8.strace.txt](https://github.com/cyrusimap/cyrus-imapd/files/2639525/search_text_3.0.8.strace.txt)
[search_text_3.0.8.imtest.txt](https://github.com/cyrusimap/cyrus-imapd/files/2639526/search_text_3.0.8.imtest.txt)
[search_text_2.4.20.strace.txt](https://github.com/cyrusimap/cyrus-imapd/files/2639527/search_text_2.4.20.strace.txt)
[search_text_2.4.20.imtest.txt](https://github.com/cyrusimap/cyrus-imapd/files/2639528/search_text_2.4.20.imtest.txt)
[search_header_3.0.8.strace.txt](https://github.com/cyrusimap/cyrus-imapd/files/2639529/search_header_3.0.8.strace.txt)
[search_header_3.0.8.imtest.txt](https://github.com/cyrusimap/cyrus-imapd/files/2639530/search_header_3.0.8.imtest.txt)
[search_header_2.4.20.strace.txt](https://github.com/cyrusimap/cyrus-imapd/files/2639531/search_header_2.4.20.strace.txt)
[search_header_2.4.20.imtest.txt](https://github.com/cyrusimap/cyrus-imapd/files/2639532/search_header_2.4.20.imtest.txt)

rsto:
This PR aims to fix #2598 . As a consequence of the rewrite of the search API for Cyrus v3, text match search expressions were only regarded as indexed if the search operator is a FUZZYMATCH. Consequently, if Cyrus is configured to use the squat search backend, then all text queries are handled as MATCH expressions any text search results in a global scan rather than making make use of squat indexes.

The patch in this PR fixes the issue for the current master branch. I plan to backport it to the upcoming v3.5 release, but had trouble building it due to incompabilities with the sasl API. Is this a known issue, @elliefm ?

The branch https://github.com/cyrusimap/cassandane/tree/fix_squat_search adds a new SearchSquat test module which contains simple regression tests that assert if the squat engine actually is called.

### DIFF
--- a/imap/search_squat.c
+++ b/imap/search_squat.c
@@ -93,17 +93,22 @@ typedef struct {
 static const char *squat_strerror(int err);
 
 static const char * const doctypes_by_part[SEARCH_NUM_PARTS] = {
-    "mh",
-    "f",
-    "t",
-    "c",
-    "b",
-    "s",
-    "h",
-    "m",
-    "o",
-    "a",
-    "ab"
+    "msh", // SEARCH_PART_ANY
+    "f",   // SEARCH_PART_FROM
+    "t",   // SEARCH_PART_TO
+    "c",   // SEARCH_PART_CC
+    "b",   // SEARCH_PART_BCC
+    "s",   // SEARCH_PART_SUBJECT
+    NULL,  // SEARCH_PART_LISTID
+    NULL,  // SEARCH_PART_TYPE
+    "h",   // SEARCH_PART_HEADERS
+    "m",   // SEARCH_PART_BODY
+    "o",   // SEARCH_PART_LOCATION
+    "a",   // SEARCH_PART_ATTACHMENTNAME
+    NULL,  // SEARCH_PART_ATTACHMENTBODY
+    NULL,  // SEARCH_PART_DELIVEREDTO
+    NULL,  // SEARCH_PART_LANGUAGE
+    NULL   // SEARCH_PART_PRIORITY
 };
 
 /* The document name is of the form
@@ -432,10 +437,7 @@ static int run(search_builder_t *bx, search_hit_cb_t proc, void *rock)
     unsigned int uid;
     int r = 0;
 
-#if DEBUG
-    if (bb->verbose > 1)
-        syslog(LOG_NOTICE, "Squat end_search()");
-#endif
+    syslog(bb->verbose > 1 ? LOG_NOTICE : LOG_DEBUG, "Squat run()");
 
     /* check we had balanced ->begin_boolean and ->end_boolean calls */
     if (bb->depth != 1)
@@ -1013,6 +1015,12 @@ static int end_update(search_text_receiver_t *rx)
     return 0;
 }
 
+static int can_match(enum search_op matchop, int partnum)
+{
+    return (matchop == SEOP_MATCH || matchop == SEOP_FUZZYMATCH) &&
+        doctypes_by_part[partnum];
+}
+
 const struct search_engine squat_search_engine = {
     "SQUAT",
     0,
@@ -1028,6 +1036,7 @@ const struct search_engine squat_search_engine = {
     /* compact */NULL,
     /* deluser */NULL,
     /* check_config */NULL,
-    /* langstats */NULL
+    /* langstats */NULL,
+    can_match
 };
 
--- a/imap/search_engines.c
+++ b/imap/search_engines.c
@@ -81,6 +81,7 @@ static const struct search_engine default_search_engine = {
     NULL,
     NULL,
     NULL,
+    NULL,
     NULL
 };
 
@@ -342,4 +343,10 @@ const char *search_op_as_string(int op)
         snprintf(buf, sizeof(buf), "(%d)", op);
         return buf;
     }
+}
+
+EXPORTED int search_can_match(enum search_op matchop, int partnum)
+{
+    const struct search_engine *se = search_engine();
+    return (se->can_match ? se->can_match(matchop, partnum) : 0);
 }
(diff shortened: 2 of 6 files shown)


elliefm:
>I plan to backport it to the upcoming v3.5 release

(Assuming that's a typo and you meant 3.4,)

> but had trouble building it due to incompabilities with the sasl API. Is this a known issue, @elliefm ?

There's no substantial difference between 3.4 and master at the moment, mostly just version numbers, so there shouldn't be any API incompatibility.  I've just tried picking the commit on top of 3.4 and it builds fine for me, and passes the new test.  Maybe something didn't rebuild properly for you?  I usually `git clean -xfd` when switching between branches to avoid messy partial rebuilds.

I also tried cherry-picking it on top of 3.2, but unsurprisingly there's merge conflicts there (in search_engines.[ch]).  I didn't try resolving them yet; I don't think it would be hard to do, but I'll wait till it's landed on master, rather than potentially doing it several times.

imap/search_squat.c
```diff
@@ -93,17 +93,22 @@ typedef struct {
...
+    NULL,  // SEARCH_PART_LISTID
+    NULL,  // SEARCH_PART_TYPE
+    "h",   // SEARCH_PART_HEADERS
+    "m",   // SEARCH_PART_BODY
+    "o",   // SEARCH_PART_LOCATION
+    "a",   // SEARCH_PART_ATTACHMENTNAME
+    NULL,  // SEARCH_PART_ATTACHMENTBODY
+    NULL,  // SEARCH_PART_DELIVEREDTO
+    NULL,  // SEARCH_PART_LANGUAGE
+    NULL   // SEARCH_PART_PRIORITY
```
elliefm, line 111: Not sure if `"mh"` -> `"msh"` is a typo?  Though thinking about it more, I think "msh" means "body or subject or headers", which makes a vague sense.

elliefm (approved):
This all looks fine to me, assuming "msh" is correct

imap/search_squat.c
rsto: Yes, it means to also search subjects for TEXT search. that's consistent with how we do text search for the Xapian backend.

rsto:
I also cherry-picked this on 3.4 now and it builds cleanly (and passes the tests). Backporting to 3.2 is going to more work, so I'd rather prefer if we can keep it at 3.4. Are you going to cherry-pick that commit for the upcoming release, @elliefm ? Or is there some action required on my part?

elliefm:
The test for this is now in Cassandane master, and I've backported the fix to the 3.4 branch.  I don't think there's anything more for you to do here.  Thanks for the research and the fix!

I'm not yet sure how I feel about 3.2, but I think I will probably leave it out.  If for some reason I change my mind, I'll take care of backporting it myself (and if it's non-trivial, then I just won't do it)

pendragonsound:
Thank you, via fix #2598. When I get a moment I'll backport this into our baseline (pre-3.2), as it looks trivial to do so, and test it. 

One minor gripe: in the series of events that added new search engine(s) to cyrus, the squatter -s option was removed, which allowed squatter to skip re-indexing a squat mailbox index when the mailbox had not been modified since the previous squatter execution. This can dramatically reduce squatter execution time for sites with giant mailboxes, and encourages them to run squatter more frequently. Perhaps this was absorbed into other options squatter now provides, but none of them look equivalent to me. Attached find the following simple patch I sketched out for the master branch that restores this option, although I have no means to test it. We are running the equivalent code under 3.0.13, however. I made no attempt to provide the corresponding code for xapian, should it have any value.

[squatter-skip.txt](https://github.com/cyrusimap/cyrus-imapd/files/5810926/squatter-skip.txt)

rsto:
@brong I see that the `-s` squatter option that @pendragonsound mentions got dropped between v2.5 and v3.0 at https://github.com/cyrusimap/cyrus-imapd/commit/3639654092bba4846f9f2e7283d1b3bee8b7317f

Looking at the diff, it looks trivial to re-introduce this option (only for squat backend). What led to the decision to remove it?

PR #3318, opened 2021-01-06
Fix assertion failure in imapd listing subscriptions from twoskip db

CambridgeDave:
I find that running `listmailbox` in cyradm causes the servicing imapd process to immediately exit with an assertion failure.  (Running v3.2.5 on Arch Linux, but same with older versions on FreeBSD.)  I have a twoskip subscription database.

Looking in gdb, the problem seems to be that mboxlist_sublist() provides its callback pointer to cyrusdb_foreach() in the wrong position.  Parameters four and five to cyrusdb_foreach() are both callback pointers, but only parameter four may be null; the twoskip myforeach() asserts this is true, resulting in an assertion failure (in myforeach() in lib/cyrusdb_twoskip.c) if the subscription database is twoskip.

I guess most people don't see this since the subscription database is not twoskip by default.

This pull request swap arguments four and five in the cyrusdb_foreach() call in mboxlist_sublist() so we pass the callback where it is required.

Tested with v3.2.5: cyradm `listmailbox` works as expected.

### DIFF
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -4583,8 +4583,7 @@ EXPORTED strarray_t *mboxlist_sublist(const char *userid)
     r = mboxlist_opensubs(userid, &subs);
     if (r) goto done;
 
-    /* faster to do it all in a single slurp! */
-    r = cyrusdb_foreach(subs, "", 0, subsadd_cb, NULL, list, 0);
+    r = cyrusdb_foreach(subs, "", 0, NULL, subsadd_cb, list, 0);
 
     mboxlist_closesubs(subs);
 
--- a/imap/userdeny_db.c
+++ b/imap/userdeny_db.c
@@ -333,7 +333,7 @@ EXPORTED int denydb_foreach(denydb_proc_t proc, void *rock)
     dr.rock = rock;
 
     return cyrusdb_foreach(denydb, "", 0,
-                           denydb_foreach_cb, NULL, &dr,
+                           NULL, denydb_foreach_cb, &dr,
                            /* txn */NULL);
 }
 
--- a/lib/cyrusdb_flat.c
+++ b/lib/cyrusdb_flat.c
@@ -502,6 +502,8 @@ static int foreach(struct dbengine *db,
 
     struct buf prefixbuf = BUF_INITIALIZER;
 
+    assert(cb);
+
     r = starttxn_or_refetch(db, mytid);
     if (r) return r;
 
--- a/lib/cyrusdb_quotalegacy.c
+++ b/lib/cyrusdb_quotalegacy.c
@@ -581,6 +581,8 @@ static int foreach(struct dbengine *db,
     int i;
     char *tmpprefix = NULL, *p = NULL;
 
+    assert(cb);
+
     /* if we need to truncate the prefix, do so */
     if (prefix[prefixlen] != '\0') {
         tmpprefix = xmalloc(prefixlen + 1);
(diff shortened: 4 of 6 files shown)


elliefm:
Oh, good catch, thanks. :)

This is a little bit interesting, because I believe that the "faster to do it all in a single slurp!" comment on the line above refers specifically to the optimisation of providing `subsadd_cb` as a `p` callback rather than a `cb` callback, which this patch undoes.  So we'll want to remove the comment too.  If you wish to amend your commit and force-push your branch, please do so, otherwise I'll probbaly just merge it next week sometime and remove the comment myself.

@brong though, before I merge this, any gut feel on the complexity of changing twoskip's `myforeach` to support the goodp-with-no-cb invocation?  The "faster to do it all in a single slurp" optimisation was yours (9ed315a586e), so I assume it's a useful enough optimisation to try to keep.  But the fact we assert against it in twoskip makes me wonder if twoskip can't support it for some subtle reason...

CambridgeDave:
You're welcome!

If you think using the p callback was deliberate - and looking at 9ed315a, I tend to agree - I'll look more carefully at the use of these two callbacks to see if the optimisation can be retained.

CambridgeDave:
I've removed the "single slurp" comment, but I actually don't think it's significant.  Slightly further down, `mboxlist_usersubs()` has exactly the same comment, but provides its callback in the cb parameter.

Having taken a look at the database foreach code, it looks like the p callbacks are intended to filter out unneeded cb callbacks, (which are presumably more expensive due to database locking, etc.).

However, I'm not sure the database API intended to allow p callbacks to replace cb callbacks.  foreach() in twoskip and zeroskip  asserts cb is non-null (I'm not sure zeroskip is ever used).  

foreach() in skiplist and flat allows cb to be null, but it's dangerous.  In this case, the p callback *must* return zero, otherwise the null cb pointer will be dereferenced.

There are three places in the code that try this:

  - `mboxlist_sublist()` in mboxlist.c, 
  - `denydb_foreach()` in userdeny_db.c, 
  - a few locations in test/mddb.c.

In all three cases, I believe the p callbacks do currently always return zero.

So I suggest you can either:

  - Always require a non-null cb callback.  You need this patch; userdeny_db.c also needs to be fixed.  skiplist and flat should make the same assertion as twoskip.  You might lose some performance.
  - Remove the cb callback assertion from twoskip (and zeroskip) foreach.  You don't need this patch.  Everything will work, but there's a non-obvious requirement on writers of p callbacks.
  - Adjust all the database foreachs so that they check for non-null cb pointers before proceeding.
 
The last option seems the best, and is reasonably simple, I think:

`    if ((!goodp || goodp(...)) && cb) ...`

instead of:

`    if (!goodp || goodp(...)) ...`

Although introducing errors to the database code would be exciting.

Any thoughts?

CambridgeDave:
Okay, for now I've assumed the database foreach API should consistently *not* allow null cb pointers.

This pull request now also moves the callback to the cb parameter in the two other places it was being passed only in the p parameter, in imap/userdeny_db.c and test/rndddb.c, and adds assertions that cb is non-null to the flat, quotalegacy, and skiplist foreach implementations.

So now the database foreach API is consistent for all the database engines and there should be no assertion failures if a user picks the 'wrong' sort of database.

This should be safe.  Someone more expert on the databases may want to comment on performance.

brong:
Yeah, this looks great.  Thanks!  If you really want to use a goodp only, you can always pass an "assert I didn't get called" callback.

elliefm:
This is now on the 3.4 branch too, which should fix #4055

PR #3321, opened 2021-01-08
fix sasl callback cleanup

### LINKED ISSUE #3320: memory leak during backend auth state cleanup
> But looking into `backend_authenticate()` the callback data structure isn't freed during error handling. This leads to an memory leak if one of the calls to `sasl_client_new()` or `sasl_setprop()` returns an error code. Furthermore `imap/mupdate-client.c` contains a similiar issue, which is present in all versions beyond 2.4.

_Originally posted by @hamarituc in https://github.com/cyrusimap/cyrus-imapd/issues/3312#issuecomment-753217299_

elliefm:
This is the same thing as #3312 (2.4), but for master, which already had some, but not all, of the fixes it contained.

* fixes use after free crash in mupdate-client
* fixes memory leaks in backend when sasl_client_new or sasl_setprop fail

Once this is reviewed and merged I'll cherry-pick it to the other branches

Closes #3320

### DIFF
--- a/imap/backend.c
+++ b/imap/backend.c
@@ -651,6 +651,7 @@ static int backend_authenticate(struct backend *s, const char *userid,
     if (iptostring((struct sockaddr *)&saddr_l, addrsize, localip, 60) != 0)
         return SASL_FAIL;
 
+    s->sasl_cb = NULL;
     if (!cb) {
         strlcpy(optstr, s->hostname, sizeof(optstr));
         p = strchr(optstr, '.');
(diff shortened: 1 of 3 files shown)


elliefm:
Can I get a review of this please?

I'm gonna need to do patch releases of 2.4, 2.5, 3.0 and 3.2 prior to 3.4's real release (such that the update such that XFER knows about the new Cyrus versions) and I would like to get this fix in first, because otherwise it might be another year before most of these branches get another release

ksmurchison (approved):
This looks sane.  Have you tried running it in a test Murder?

elliefm:
Nope, I'm not sure how to deliberately provoke the crash?  If I can manage to do so I'll add a Cassandane test for it, but otherwise I'm trusting the report that it fixes the problem in 2.4, and just bringing it forward.

brong (approved):
yeah, looks good to me

PR #3323, opened 2021-01-13
caldav_alarm.c: limit the number of consecutive alarms that can be pr…

ksmurchison:
…ocessed for a single user

### DIFF
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -1207,6 +1207,8 @@ done:
     mailbox_close(&mailbox);
 }
 
+#define MAX_CONSECUTIVE_ALARMS_PER_USER 50
+
 /* process alarms with triggers before a given time */
 EXPORTED int caldav_alarm_process(time_t runtime, time_t *intervalp)
 {
@@ -1233,14 +1235,30 @@ EXPORTED int caldav_alarm_process(time_t runtime, time_t *intervalp)
 
     caldav_alarm_close(alarmdb);
 
-    int i;
+    int i, cnt = 0;
+    char *prev_userid = NULL;
     for (i = 0; i < rock.list.count; i++) {
         struct caldav_alarm_data *data = ptrarray_nth(&rock.list, i);
+        char *userid = mboxname_to_userid(data->mboxname);
+
+        if (strcmpnull(userid, prev_userid)) {
+            /* different user - reset counter */
+            free(prev_userid);
+            prev_userid = userid;
+            cnt = 1;
+        }
+        else if (++cnt >= MAX_CONSECUTIVE_ALARMS_PER_USER) {
+            /* exceeded per-user limit - move this alarm to end of list */
+            ptrarray_append(&rock.list, data);
+            continue;
+        }
+
         process_one_record(data, runtime);
         caldav_alarm_fini(data);
         free(data);
     }
     ptrarray_fini(&rock.list);
+    free(prev_userid);
 
     syslog(LOG_DEBUG, "done");
 
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -1235,30 +1235,54 @@ EXPORTED int caldav_alarm_process(time_t runtime, time_t *intervalp)
 
     caldav_alarm_close(alarmdb);
 
-    int i, cnt = 0;
-    char *prev_userid = NULL;
-    for (i = 0; i < rock.list.count; i++) {
-        struct caldav_alarm_data *data = ptrarray_nth(&rock.list, i);
-        char *userid = mboxname_to_userid(data->mboxname);
+    int next = 0;
+    char *prev_mboxname = NULL;
+    do {
+        int i = next, cnt = 0;
 
-        if (strcmpnull(userid, prev_userid)) {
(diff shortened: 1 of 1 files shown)


imap/caldav_alarm.c
```diff
@@ -1233,14 +1235,30 @@ EXPORTED int caldav_alarm_process(time_t runtime, time_t *intervalp)
...
         struct caldav_alarm_data *data = ptrarray_nth(&rock.list, i);
+        char *userid = mboxname_to_userid(data->mboxname);
+
+        if (strcmpnull(userid, prev_userid)) {
+            /* different user - reset counter */
+            free(prev_userid);
+            prev_userid = userid;
+            cnt = 1;
+        }
+        else if (++cnt >= MAX_CONSECUTIVE_ALARMS_PER_USER) {
```
brong, line 1250: If there's only a single user with more than 50 alarms, this will cause an infinite loop!  You need to at least zero out `cnt` again.

brong (changes requested):
This needs to be infinite-loop-proofed, but otherwise the concept is good.

brong:
Thinking about this more seriously, I think the idea might be to tack a "NULL" record on the end of the list before processing, and then when we hit that NULL record we can append a new NULL record to the list again. This allows working through in those blocks of 50 items quite cleanly, and resets the counter each time.

imap/caldav_alarm.c
```diff
@@ -1233,14 +1235,54 @@ EXPORTED int caldav_alarm_process(time_t runtime, time_t *intervalp)
...
+
+            if (!data) continue;
+
+            /* XXX  we can do this comparison because the list is sorted */
+            if (!prev_mboxname ||
+                strncmp(data->mboxname, prev_mboxname, strlen(prev_mboxname))) {
+                /* different user/tree - reset counter */
+                cnt = 0;
+            }
+            else if (cnt >= MAX_CONSECUTIVE_ALARMS_PER_USER) {
```
brong, line 1256: My spidey senses detect that this is mis-named, and should be `MAX_CONSECUTIVE_ALARMS_PER_MAILBOX`.

brong (approved):
Overall this looks great - and I think it will be enough.

The one thing it doesn't handle is if there are so many alarms that they run for minutes, because it still processes everything for this timeslice before looking to see if we've advanced into a later timeslice.  But hopefully that's OK.

(actually: it would be even simpler to avoid this timeslice issue, just set a "run again immediately" flag if we ever skip a record.  The fetch code will immediately fetch it still, because the timestamp will still be in the past!)

imap/caldav_alarm.c
ksmurchison: I don't think so.  We only reset the counter when we hit a mailbox that is no longer a child or its predecessor.

But now that I look at this again, I think that logic is faulty.  If we have:
user.foo.A
user.foo.A.B
user.foo.B

This will reset the counter at user.foo.B, which is not what we want

ksmurchison: We should just check if the owner of the mailbox has changed, which is what I think I did originally before I tried to get too cute.

brong: Meh, I think counting by mailbox is reasonable - most people won't have tons of mailboxes each with tons of alarms.

brong:
Btw: my "run again immediately" idea is basically instead of setting 'next = i' just set isdirty = 1, and then return `IMAP_AGAIN` or something and then check the return code of caldav_alarm_process() and don't sleep if it returned `IMAP_AGAIN`.  This would immediately fetch the list of items to run now from the sql db, which would include all the skipped items, because their "next run" time would still be in the past.

This would also pick up new events, since we're doing a time(NULL) again each time through the loop, so we could never starve the other users.

Right now, we keep working on the immediate set that we fetched until it's all completed, even though we're reducing the starvation problem by only doing a small set per user/mailbox - so it doesn't allow new events to mix in if there's a huge amount to do right now.

robmueller:
> Btw: my "run again immediately" idea is basically instead of setting 'next = i' just set isdirty = 1, and then return `IMAP_AGAIN` or something and then check the return code of caldav_alarm_process() and don't sleep if it returned `IMAP_AGAIN`. This would immediately fetch the list of items to run now from the sql db, which would include all the skipped items, because their "next run" time would still be in the past.

IMHO +1 for this approach, I think this is better.

Also the same approach for sending delayed emails. I can see some crappy case in the future where someone uses a dumb bot to submit 5000 emails to be sent 30 seconds in the future. We don't want that to block delayed sending or caldav alarms for all users. So It would be good if we process say 20 of them, then say "that's enough, move on to the next user", but once we've done *everything* else, we immediately say "restart the whole processing loop".

brong:
Sorry to be a pain and snipe your work AGAIN.  I wound up re-implementing this as part of the non-blocking work for dav_reconstruct over in https://github.com/cyrusimap/cyrus-imapd/pull/3380

The specific commit is

https://github.com/cyrusimap/cyrus-imapd/commit/a25cf782dc0a549cbda9f794305c5d05bb86ca21

brong:
This was solved in another PR, so closing this one.

PR #3324, opened 2021-01-14
index: make error handling in getsearchtext more verbose and resilient

rsto:
This patch makes search text extraction more resilient for attachment extracting. Previously, we only a subset of all extractors gracefully if partial message indexing is enabled. Now we treat all extractor errors the same. In addition, we now log erroneous messages to help with debugging.


ksmurchison (approved):
I'm not overly familiar with this code, but the PR looks sane to me.

PR #3325, opened 2021-01-15
message: cache mail address names as quoted-string, if necessary

rsto:
This patch fixes a bug where email addresses with quoted-string names were stored in the index cacherecord without quotes. If such an cached address is reparsed, then the parseaddr lib drops the invalid atoms from the name. In other words: the result of reading an address from cache may differ from the address that was written. This has shown to be a problem in search indexing, but it may just as well have resulted in bugs in other parts of Cyrus.

The cunit tests were updated. In addition, the mailbox minor cache version was bumped.

Addendum: Independently, we may want to make parseaddr more lenient with regards to invalid [atext](https://tools.ietf.org/html/rfc5322#section-3.2.3) names.

### DIFF
--- a/imap/message.c
+++ b/imap/message.c
@@ -2652,7 +2652,34 @@ static void message_write_searchaddr(struct buf *buf,
 
             if (addrlist->name) {
                 tmp = charset_parse_mimeheader(addrlist->name, charset_flags);
-                buf_appendcstr(buf, tmp);
+                /* Determine if name is an atext or quoted-string */
+                static const char *atext_specials = "!#$%&'*+-/=?^_`{|}~";
+                const char *c;
+                for (c = tmp; *c; c++) {
+                    // see RFC 5322, section 3.2.3
+                    if (!isalpha(*c) && !isdigit(*c) && !isspace(*c) &&
+                            !strchr(atext_specials, *c)) {
+                        break;
+                    }
+                }
+                int need_quote = *c;
+                /* Write name */
+                if (need_quote) {
+                    struct buf qtext = BUF_INITIALIZER;
+                    buf_ensure(&qtext, strlen(tmp) + 2);
+                    buf_putc(&qtext, '"');
+                    for (c = tmp; *c; c++) {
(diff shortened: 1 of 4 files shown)


imap/message.c
```diff
@@ -2652,7 +2652,21 @@ static void message_write_searchaddr(struct buf *buf,
 
             if (addrlist->name) {
                 tmp = charset_parse_mimeheader(addrlist->name, charset_flags);
+                /* Determine if name is an atext or quoted-string */
+                static const char *atext_specials = "!#$%&'*+-/=?^_`{|}~";
```
brong, line 2656: What happens if there's a " (DOUBLE QUOTE) in the name?  Is that something that can happen?

brong (approved):
Yeah, this looks really good.  So we're going to have to reconstruct caches first, then reindex?

elliefm:
Can we please get a file under the changes/next directory that mentions the need to rebuild cache and reindex, etc?  This is the exact kind of detail we're trying to capture more effectively. :)

Do we have any opinions about whether or not this fix should be backported to 3.4?  IMO it seems like it ought to be, but I'm not sure of the severity

imap/message.c
rsto: Right, it actually can! I have updated the code and tests accordingly.

rsto:
@elliefm I added a file for upgrade instructions.

rsto:
@brong should we bump the mailbox cache one more time? Otherwise we might have missed a few mails with quotes in email names.

rsto:
One more thing: we do store the quoted strings as UTF-8, not qp-encoded like they would be if we would serialize the email address back to MIME. I think that's a sane choice, as parseaddr can handle these non-ASCII characters in quoted strings.

brong:
> @brong should we bump the mailbox cache one more time? Otherwise we might have missed a few mails with quotes in email names.

No, that should be fine - it's just the future and staff stores at Fastmail that should have seen this by now!

PR #3328, opened 2021-01-19
squatter: support long command-line argument options

rsto:
We've nearly exhausted the alphabet for squatter command-line arguments, and we already prepare to add even more options. This patch adds support for long argument options, and preserves all existing short options.

For reviewers: please have a look at the chosen long argument names. Do they make sense to you? Do you prefer something else?

### DIFF
--- a/docsrc/imap/reference/manpages/systemcommands/squatter.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/squatter.rst
@@ -237,6 +237,18 @@ Options
     Recursively create indexes for all sub-mailboxes of the user,
     mailboxes or mailbox prefixes given as arguments.
 
+.. option:: -s, --squat-skip[=delta]
+
+    Skip mailboxes that have not been modified since last index. This is
+    achieved by comparing the last modification time of a mailbox to
+    the last time the squat index of this mailbox got updated. If the
+    mailbox modification time (plus delta) is less than the squat
+    index modification time, then the mailbox is skipped. The optional
+    argument value delta is defined in seconds and must be equal to or
+    higher than zero, the default value is 60.
+    Squat only.
+    |master-new-feature|
+
 .. option:: -S seconds, --sleep=seconds
 
     After processing each mailbox, sleep for "seconds" before
--- a/docsrc/imap/reference/manpages/systemcommands/squatter.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/squatter.rst
@@ -122,7 +122,7 @@ Options
 
     |cli-dash-c-text|
 
-.. option:: -a
+.. option:: -a, --squat-annot
 
     Only create indexes for mailboxes which have the shared
     */vendor/cmu/cyrus-imapd/squat* annotation set to "true".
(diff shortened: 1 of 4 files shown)


elliefm:
>        getopt_long() and getopt_long_only():
>               These functions are GNU extensions.

Hmm, a quick search isn't revealing a "portable" or standardised API for this... I wonder whether the big non-GNU platforms provide an equivalent?

I would like to see a check in configure.ac, such that configure errors out if the system doesn't provide a compatible `getopt_long()`.  At the moment running configure would succeed, indicating the platform is supported, but then squatter won't compile.

There's not much point if-defing the implementation, since we don't have a "use this other thing instead" option, and "just don't use it" isn't much of an option either.  But we can at least gracefully fail at configure if we know it's gonna fail later anyway.

rsto:
Thanks for the input @elliefm . I have updated the patch with a configure check.

elliefm:
It's not clear to me from the long option names which ones select a mode vs which options tweak the behaviour of that mode... but I don't use it enough to have an intuition, and maybe if you do it's obvious?  And it was never clear as single-character options anyway, so take this as observation rather than critique.

Can you have a look at updating the "Options" section of man page too, to list the long option names alongside the short ones?  The source file is docsrc/imap/reference/manpages/systemcommands/squatter.rst.  I'm not sure what this will need to look like, since this is the first time we're supporting long options...  If you get stuck wrestling with the RST and want to handball it to me to figure, that's fine.

rsto:
@elliefm the mode selection in squatter indeed are confusing, multiple options may select the same mode. We can't change that without breaking backwards-compability, so instead I chose to simplify the naming of long options. While doing that, I also updated the usage string for the repurposed -T option (https://github.com/cyrusimap/cyrus-imapd/commit/b782fcce15a512b12120393191655971bd4228c7) as well as the new -B option (https://github.com/cyrusimap/cyrus-imapd/commit/2d9e371ea32b49c9912106b133c4f2c1ef696add).
The reStructured text source for squatter now includes long options, but I don't think it's using the option directory right. I think there is an option_group directory in docutils, but I didn't manage to use that during a quick attempt. Do you know more about that?

elliefm:
I don't, but I'm checking the docs, and found this:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-option

Looks like just separating them with commas as you have done is fine.

It also says we should put ``<>`` around the arguments (so like: ``.. option:: -n <channel>, --channel=<channel>``), which I think would then allow us to create links to those option arguments from elsewhere.  But we don't use that anywhere else, so I think leave it as is for now, and if later on we want to use that feature, we can do it en-masse.

rsto:
Some squat backend users want the -s option back, so I added it in tandem with this patch. @elliefm I would like to backport both of these to the 3.4 branch.

docsrc/imap/reference/manpages/systemcommands/squatter.rst
```diff
@@ -237,6 +237,18 @@ Options
...
     mailboxes or mailbox prefixes given as arguments.
 
+.. option:: -s, --squat-skip[=delta]
+
+    Skip mailboxes that have not been modified since last index. This is
+    achieved by comparing the last modification time of a mailbox to
+    the last time the squat index of this mailbox got updated. If the
+    mailbox modification time (plus delta) is less than the squat
+    index modification time, then the mailbox is skipped. The optional
+    argument value delta is defined in seconds. It must be non-zero,
```
elliefm, line 247: From the `case 's':` switch block below, it looks like it also has to be positive?  I'm not sure if a negative delta would ever be useful

imap/squatter.c
```diff
@@ -356,6 +356,21 @@ static int index_one(const char *name, int blocking)
         printf("Indexing mailbox %s... ", extname);
     }
 
+    if (skip_unmodified) {
```
elliefm, line 359: An interesting quirk of using `skip_unmodified` like this is that you cannot express "I want to skip the unmodified mailboxes, using a delta of 0".  No idea if this is useful, and I assume most of this implementation is just bringing back code from the old `-s` implementation?  So I'm not sure how fussy I want to be...

I think, _if_ I thought "skip the unmodified mailboxes with a delta of zero" was a useful behaviour, I might do something like:

* initialise skip_unmodified to -1
* in the `case 's':`, change it to either 60 or atoi(optarg) and complain if it's negative
* in the loop, check for `if (skip_unmodified >= 0)`

so that the sign determines whether the flag is enabled, and the magnitude is the delta.

But all that cleverness still doesn't enable negative deltas, and I still have no insight as to whether they're useful, nor whether zero is.   I dunno, what do you think?

elliefm (commented):
If we don't change the implementation to support zero and/or negative deltas, then it just needs the man page to specify "non-zero and positive" instead of just "non-zero".

I don't have an opinion on whether we change the implementation or not, will defer to your judgement. :)

docsrc/imap/reference/manpages/systemcommands/squatter.rst
rsto: No, a negative delta is never useful.

imap/squatter.c
rsto: Zero kind of makes sense and it's simple to support, so why not. I updated the code and documentation accordingly.

elliefm:
Great, this is looking good to me 👍

rsto:
@brong we merged that now to get it backported to the 3.4 branch

elliefm:
This is now on 3.4 too

PR #3329, opened 2021-01-20
allow "UNRELEASED" as imapoptions version string

elliefm:
This resolves the annoying problem of needing to guess what the next Cyrus release version will be when making changes in `lib/imapoptions`, and then needing to keep this updated in your branch if new releases are made before it's merged -- which is becoming more of a nuisance now as our work is likely to sit in a PR for a while before it lands on master.

From now on (well, once this is merged), if you're adding, changing, or deprecating an imapd.conf option, instead of trying to anticipate the next version number, just enter it as the magic string `"UNRELEASED"`.  This works equally well for both the "last_modified" field and the "deprecated_since" field.  Note that the double quotes are required, just as they are for normal version strings.

Part of the process for future releases will be to update these entries with the real release version number, once it is known (i.e. just before release).  The "UNRELEASED" string provides a clear target of which things are new and need to be set correctly.

When building Cyrus, if there are updated config options like this, you will see a warning from make:

```
  GEN      lib/imapopts.c
./tools/config2header: warning: build contains UNRELEASED config options
```

If you're not building a new release, you can ignore this warning.  If you _are_ building a release, you need to resolve this warning (by replacing the "UNRELEASED"s with the new version number) before tagging it.

When using `cyr_info`'s `-s version` switch to look for changed options, UNRELEASED options will have the version 255.255.255, because they needed to have something, and programmatically guessing a version number seemed to defeat the purpose of not guessing a version number...

### DIFF
--- a/tools/config2header
+++ b/tools/config2header
@@ -128,6 +128,7 @@ EXPORTED struct imapopt_s imapopts[] =
 EOF
     ;
 
+my $__warned_unreleased = 0;
 sub parse_last_modified
 {
     my ($version) = @_;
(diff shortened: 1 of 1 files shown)


rsto (approved):
Neat. Looks good to me.

ksmurchison (approved):

PR #3330, opened 2021-01-22
sync_support: downgrade a debug log

robn:
Well, I think its a debug log. Its certainly very noisy, and not
operationally useful.

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -7098,7 +7098,7 @@ static void split_user_mailboxes(const char *key __attribute__((unused)),
     }
 
     if (mboxname_list->count) {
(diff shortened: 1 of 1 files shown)


elliefm:
```
$ack 'syslog.*doing'
imap/sync_support.c
4201:                syslog(LOG_DEBUG, "doing partial sync: %s (%u/%u/%u) (%llu/%llu/%llu)",
7101:        syslog(LOG_NOTICE, "sync_mailboxes: doing %lu",
```

The only other "doing ..." syslog is a LOG_DEBUG already, so yeah, it seems pretty reasonable to me for them both to be!

PR #3335, opened 2021-01-26
search_expr: make nnodes a non-static variable

rsto:
**NOTE**: we might not want to push this ever on the master branch and rather remove the nnodes checks from the search code altogether. This pull request should help us identifying if we can do that safely.

**Description**

Static variable nnodes increments whenever a search expression is allocated. Its purpose is to keep track of the number of search expressions that are created during search expression tree normalisation. While uid-based search resets the counter each time it is called, it may accumulate in process memory as long as only guidsearch is executed.

This could introduce hard-to-track bugs, as the JMAP email search code uses the search expression normalisation code independently from uid-based search.

To avoid this, let's make nnodes a function argument.

### DIFF
--- a/changes/next/search_nonstatic_nnodes
+++ b/changes/next/search_nonstatic_nnodes
@@ -1,0 +1,11 @@
+Description:
+
+Prevents a bug in search, where queries may exceed the search normalization limit.
+
+Config changes:
+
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):
Looks sane to me

rsto:
Now updated with a couple of more improvements in error handling and performance.

PR #3339, opened 2021-02-01
fix parsing of folded leading whitespace in headers

elliefm:
This mainly fixes stuff like this:

```
Message-ID:
 <foo@whatever>
```

being misparsed as `" <foo@whatever>"` rather than `"<foo@whatever>"` –  i.e. the leading whitespace was not eaten if it was also folded.  

One consequence of this was bad formatting of auditlogs for affected messages; another (user visible!) was that duplicate suppression would fail to detect the duplication if one copy was folded and the other not.

The misparsing of folded leading whitespace affects all headers in theory, but other headers usually contain other whitespace, and end up folded later instead of right at the start.  Message-ID is exceptional both in that it doesn't usually contain any other whitespace (so if the sender is going to fold it, they can only fold at the leading whitespace), and in that the value is significant to our handling of the message.

Added two cunit tests: one for this behaviour specifically, and a regression test for empty-header handling to make sure the fix didn't break that.

Obviously this won't do anything to fix up cache/etc for older messages, it'll only help for new incoming messages.  Is there something an admin can do to retroactively fix older messages?

### DIFF
--- a/imap/spool.c
+++ b/imap/spool.c
@@ -207,6 +207,31 @@ static int parseheader(struct protstream *fin, FILE *fout,
             break;
 
         case BODY_START:
+            /* leading whitespace might be folded too */
+            /* XXX mostly copied from BODY, below. could be deduped */
+            if (c == '\r' || c == '\n') {
+                int peek;
+
+                peek = prot_getc(fin);
+
+                if (!skip) buf_appendcstr(&raw, "\r\n");
+                /* we should peek ahead to see if it's folded whitespace */
+                if (c == '\r' && peek == '\n') {
+                    c = prot_getc(fin);
+                }
+                else {
+                    c = peek;
+                }
+
+                if (c != ' ' && c != '\t') {
+                    /* not folded leading ws, actually end of header! */
+                    buf_cstring(&body);
+                    buf_cstring(&raw);
+                    prot_ungetc(c, fin);
+                    goto got_header;
+                }
+            }
+
             if (c == ' ' || c == '\t') /* eat the whitespace */
                 break;
             buf_reset(&body);
--- a/imap/spool.c
+++ b/imap/spool.c
@@ -103,9 +103,10 @@ typedef enum {
 /* we don't have to worry about dotstuffing here, since it's illegal
    for a header to begin with a dot!
 
-   returns 0 on success, filling in 'headname' and 'contents' with a static
-   pointer (blech).
-   on end of headers, returns 0 with NULL 'headname' and NULL 'contents'
+   returns 0 on success, filling in 'headname', 'contents', and 'rawvalue'
+   with malloced pointers (caller must free!).
+   on end of headers, returns 0 with NULL 'headname', NULL 'contents', and
+   NULL 'rawvalue'
 
    on error, returns < 0
 */
--- a/cunit/spool.testc
+++ b/cunit/spool.testc
@@ -279,6 +279,124 @@ static void test_folded_headers(void)
     unlink(tempfile);
 }
 
+static void test_empty_headers(void)
+{
+    static const char MSG[] =
+"From: " HFROM "\r\n"
+"Message-ID: " HMESSAGEID "\r\n"
+"Empty1:\r\n"                               /* not even a leading space */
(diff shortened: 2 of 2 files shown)


imap/spool.c
```diff
@@ -207,6 +207,30 @@ static int parseheader(struct protstream *fin, FILE *fout,
...
+
+                peek = prot_getc(fin);
+
+                if (!skip) buf_appendcstr(&raw, "\r\n");
+
+                if (c == '\r' && peek == '\n') {
+                    c = prot_getc(fin);
+                }
+                else {
+                    c = peek;
```
rsto, line 222: OK, so this allows folded-whitespace to either be CRLF+<WS> or LF+<WS>. Which looks like a sensible thing to do, if somehow a stray newline made it through here. But it should be better included in the test, so at least we can make informed decision if we later decide to change it.

rsto (commented):
The overall logic looks good to me. And yay - tests! I still have a few remarks.

ksmurchison (commented):
I wonder if this is simpler (or even create a new function which checks for newline):

```
diff --git a/imap/spool.c b/imap/spool.c
index d78d69c62..b1f00392d 100644
--- a/imap/spool.c
+++ b/imap/spool.c
@@ -207,11 +207,7 @@ static int parseheader(struct protstream *fin, FILE *fout,
             break;
 
         case BODY_START:
-            if (c == ' ' || c == '\t') /* eat the whitespace */
-                break;
-            buf_reset(&body);
-            s = BODY;
-            /* falls through! */
+            /* leading whitespace might be folded too */
         case BODY:
             /* now we want to convert all newlines into \r\n */
             if (c == '\r' || c == '\n') {
@@ -234,6 +230,14 @@ static int parseheader(struct protstream *fin, FILE *fout,
                     goto got_header;
                 }
             }
+
+            if (s == BODY_START) {
+                if (c == ' ' || c == '\t') /* eat the whitespace */
+                    break;
+                buf_reset(&body);
+                s = BODY;
+            }
+
             if (c >= 0x80) {
                 if (reject8bit) {
                     /* We have been configured to reject all mail of this

```

ksmurchison:
Ugh.  The copy-n-paste of the diff looks crappy.  Essentially move Hopefully, its still understandable

imap/spool.c
```diff
@@ -207,6 +207,30 @@ static int parseheader(struct protstream *fin, FILE *fout,
             break;
 
         case BODY_START:
+            /* leading whitespace might be folded too */
+            if (c == '\r' || c == '\n') {
+                int peek;
```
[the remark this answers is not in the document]
elliefm: This dealing-with-newlines section is more-or-less a direct copy of the same code in the `BODY:` state immediately below. I've deliberately left the style the same so that it's clear to the casual reader that they're the same.

imap/spool.c
elliefm: Yep, again just using the same logic as BODY.  But actually testing it is a great idea, I'll do that.

elliefm:
Hmm.  I was thinking `buf_reset(&body)` still needs to be in BODY START because it's the only place it's initialised/cleared... but it's actually unconditionally reset at the top of the function, before the state machine starts, so that's fine.

But the BODY_START state _only_ initialises `body` and eats the leading whitespace.  Once that's happening in BODY instead, the state machine pattern becomes tangled up and I think harder to follow overall.

Another option might be to get rid of the BODY_START state entirely, and move the "eat an optional leading whitespace" behaviour to the very end, after the parser, before the return -- simplifying the state machine by discarding the leading whitespace after the fact, rather than during the parse.

But I'm going for a light touch here, because I don't think we have enough test coverage of this function to prove a refactor didn't mess things up, and afaik it currently works well other than this one "not noticed in decades" quirk.

imap/spool.c
elliefm: I've extended test_empty_headers to do the same three checks with each of bare CR and LF

elliefm:
Oh hey, our test coverage of parseheader isn't actually too bad (but we do miss a bunch of the error cases).

rsto (approved):
Looks good to me.

PR #3344, opened 2021-02-07
dav sharing: xsyslog-ify an error seen in the wild

rjbs:
@elliefm …just something I saw in the wild while on my rounds.  I think this is right but haven't even compiled.


elliefm (approved):
Looks right, builds fine, passes all our tests...

PR #3346, opened 2021-02-08
lib/imapoptions: deprecate sieve_sasl_send_unsolicited_capability

dilyanpalauzov:
It comply with [RFC 5804 A Protocol for Remotely Managing Sieve Scripts](https://tools.ietf.org/html/rfc5804) this option must be off.

### DIFF
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -2727,9 +2727,8 @@ product version in the capabilities
    \fIfileinto\fR action in scripts to use UTF8 encoding.  Otherwise,
    modified UTF7 encoding should be used. */
 
-{ "sieve_sasl_send_unsolicited_capability", 0, SWITCH, "2.3.17" }
-/* If enabled, timsieved will emit a capability response after a successful
-   SASL authentication, per draft-martin-managesieve-12.txt . */
+{ "sieve_sasl_send_unsolicited_capability", 1, SWITCH, "UNRELEASED", "UNRELEASED" }
+/* Deprecated: this option does nothing. */
 
 { "sieve_use_lmtp_reject", 1, SWITCH, "3.1.1" }
 /* Enabled by default.  If reject can be done via LMTP, then return a 550
--- a/changes/next/deprecate_sieve_sasl_send_unsolicited_capability
+++ b/changes/next/deprecate_sieve_sasl_send_unsolicited_capability
@@ -1,0 +1,13 @@
+Description:
+
+Deprecate sieve_sasl_send_unsolicited_capability in lib/imapoptions and toggle its default.
+
+Config changes:
+
+This enables the RFC 5804 behaviour by default. The old behaviour - per
+draft-martin-managesieve-08 - is not available anymore.
+
+GitHub issue:
+
+https://github.com/cyrusimap/cyrus-imapd/pull/3346
+
--- a/timsieved/parser.c
+++ b/timsieved/parser.c
@@ -890,8 +890,7 @@ static int cmd_authenticate(struct protstream *sieved_out,
   sasl_getprop(sieved_saslconn, SASL_SSF, &val);
   sasl_ssf = *((sasl_ssf_t *) val);
 
-  if (sasl_ssf &&
-      config_getswitch(IMAPOPT_SIEVE_SASL_SEND_UNSOLICITED_CAPABILITY)) {
+  if (sasl_ssf) {
       capabilities(sieved_out, sieved_saslconn, starttls_done, authenticated,
                    sasl_ssf);
       prot_flush(sieved_out);


lib/imapoptions
```diff
@@ -2422,7 +2422,8 @@ product version in the capabilities
    modified UTF7 encoding should be used. */
 
 { "sieve_sasl_send_unsolicited_capability", 0, SWITCH, "2.3.17" }
```
elliefm, line 2424: You should also update the "last_modified" and "deprecated_since" field of this config item, which you do by changing the existing last_modified version (currently "2.3.17") and adding the deprecated_since version after it.  Since you don't know what version this will be released in, you use the string "UNRELEASED" for both fields, and the release manager (me) will sort it out when the time comes.

Which is to say, you should change this line to something like this:

`{ "sieve_sasl_send_unsolicited_capability", 0, SWITCH, "UNRELEASED", "UNRELEASED" }`

This will allow `cyr_info` to recognise that the option has changed, and that it's deprecated, and will enable the deprecated config warning if it's in imapd.conf at startup. :)

Formally deprecating it like this will cause a fatal assertion if any part of Cyrus tries to use the option, so you'll also need to update timsieved/parser.c to no longer try to read it, probably by removing these lines around line 889:

```
  if (sasl_ssf &&
      config_getswitch(IMAPOPT_SIEVE_SASL_SEND_UNSOLICITED_CAPABILITY)) {
      capabilities(sieved_out, sieved_saslconn, starttls_done, authenticated,
                   sasl_ssf);
      prot_flush(sieved_out);
  }
```

Nothing else seems to use the option at all

dilyanpalauzov: > Formally deprecating it like this will cause a fatal assertion if any part of Cyrus tries to use the option, 

What is the difference between “Crashing, because undefined option was used” and “Crashing, because deprecated option was used”?

elliefm: We don't have undefined options, because we deprecate old ones, not remove them entirely.  If we tried to use an option that wasn't defined, it wouldn't even compile, because IMAPOPT_WHATEVER wouldn't exist in the enum.

Asserting if Cyrus still tries to use them after they're deprecated tells us (devs) that we screwed up, and tried to use something that we told sysadmins doesn't exist anymore.  You can see the full range of deprecated behaviours in the test_deprecated_* tests in cunit/libconfig.testc if you're curious.

If you really really need to, you can read an option from imapd.conf that isn't defined in imapoptions using the `config_getoverflowstring()` API, which is how things like the channel prefix on sync options are handled.  But this API doesn't provide the SWITCH/INT/DURATION/etc parsing or type checking, it just gives you the string value from the config file and you have to parse it yourself.  So if we really wanted to deprecate the option but still honour it if it were set, we could convert the existing  `config_getswitch(IMAPOPT_SIEVE_SASL_SEND_UNSOLICITED_CAPABILITY)` check to read it as an overflow string and figure out its true/false value manually, but that seems like a lot of work for something that shouldn't even exist?

dilyanpalauzov:
For me deprecated means: “it will be removed in the future, but currently serves some function”.  As opposed to “NO-OP, kept for backward compatibility”.  Actually I prefer that the unsolicited capability is never sent, when not needed, but as a transition of a year or two I thought updating the documentation to state, that this option is discouraged.  I am fine with any path that abolishes this option.

Please update the code by yourself, as you think is appropriate for the case.  Going over lengthy discussions on a three-lines change is disproportional to the efforts to carry the discussion.

rsto:
I don't see how RFC 5804 forbids sending these capabilities. Instead, it looks to me as if servers MUST send the capabilities after authenticated, e.g. this option MUST be on for a compliant implementation? This is what [RFC 5804 section 1.7](https://datatracker.ietf.org/doc/html/rfc5804#section-1.7) states:

> Server capabilities are sent automatically by the server upon a client connection, or after successful STARTTLS and AUTHENTICATE (which establishes a Simple Authentication and Security Layer (SASL)) commands.  Capabilities may change immediately after a successfully completed STARTTLS command, and/or immediately after a successfully completed AUTHENTICATE command, and/or after a successfully completed UNAUTHENTICATE command (see [Section 2.14.1](https://datatracker.ietf.org/doc/html/rfc5804#section-2.14.1)).  Capabilities MUST remain static at all other times.

@ksmurchison thoughts?

dilyanpalauzov:
RFC 5804 wes preceded by draft-martin-managesieve, which has not required that the server sends unsolicited capability after AUTHENTICATE.  The option enables the pre-RFC 5804 behaviour.  Cf. https://author-tools.ietf.org/iddiff?url1=draft-martin-managesieve-08&url2=draft-martin-managesieve-09&difftype=--html:

draft-martin-managesieve-08 wording
>  Server capabilities are sent by the server upon a client connection.	 

draft-martin-managesieve-09 wording:
> Server capabilities are sent automatically by the server upon a client connection, or after successful STARTTLS and AUTHENTICATE (which establishes a SASL security layer) commands.

As nobody needs the pre-RFC5804 behaviour, this option should disappear.

dilyanpalauzov:
After looking at this again, I think:
* the default value should be changed to 1 (do send unsolicited capabilities after AUTHENTICATE), 
* the draft version in the description should be changed from 12 to 09, and
* this option should be deprecated and eventually removed.

rsto:
Agreed, let's deprecate this option and make it always send the capabilities, as required by the spec.
Do you want to come up with the code changes in this PR or alternatively, we'll close this one and I'll create a new one.

rsto:
Could you please also remove the check for this option?

I mean this line:
https://github.com/dilyanpalauzov/cyrus-imapd/blob/da1411e64bcdb2b0dd5f2922ae84fe2e9820e31d/timsieved/parser.c#L894

rsto (approved):

rsto:
Great, thanks.

rsto:
@elliefm since this is a potential breaking change I did not backport this to 3.10.

PR #3347, opened 2021-02-08
Jmap getblob handlers v2

ksmurchison:
This has multiple changes related to Contacts avatars all with the goal of allowing blobIds of any type to be used for an avatar (e.g. BIMI header blobIds).  To do this the getblob handlers have been made to put the blob content in a struct buf, with the output for an actual download to be handled by jmap_download().  It also gets rid of the hack to conv.db 'G' records to keep track of blobIds for vCard PHOTO properties, which has now been moved to a variant of the 'V' smart blobIds.  As as result, we need to report out that the blobId for an avatar changes from a 'G' or 'H' blobId to a new 'V' blobId during /set{create,update}.

### DIFF
--- a/imap/jmap_calendar.c
+++ b/imap/jmap_calendar.c
@@ -86,8 +86,13 @@ static int jmap_calendarevent_query(struct jmap_req *req);
 static int jmap_calendarevent_set(struct jmap_req *req);
 static int jmap_calendarevent_copy(struct jmap_req *req);
 
-static int jmap_calendarevent_getblob(jmap_req_t *req, const char *blobid,
-                                      const char *accept);
+static int jmap_calendarevent_getblob(jmap_req_t *req,
+                                      const char *from_accountid,
+                                      const char *blobid,
+                                      const char *accept_mime,
+                                      struct buf *blob,
+                                      const char **content_type,
+                                      const char **errstr);
 
 #define JMAPCACHE_CALVERSION 19
 
@@ -1552,7 +1557,7 @@ done:
 
 
 struct calendarevent_getblob_rock {
-    jmap_req_t *req;
+    const char *boundary;
     struct buf *buf;
 };
 
@@ -1567,7 +1572,6 @@ static int _calendarevent_getblob_cb(const char *mailbox __attribute__((unused))
     if (!buf_len(value)) return 0;
 
     struct calendarevent_getblob_rock *rock = vrock;
-    jmap_req_t *req = rock->req;
     struct buf *buf = rock->buf;
 
     /* Parse the value and fetch the patch */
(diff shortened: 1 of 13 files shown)


imap/jmap_calendar.c
```diff
@@ -86,8 +86,13 @@ static int jmap_calendarevent_query(struct jmap_req *req);
 static int jmap_calendarevent_set(struct jmap_req *req);
 static int jmap_calendarevent_copy(struct jmap_req *req);
 
-static int jmap_calendarevent_getblob(jmap_req_t *req, const char *blobid,
-                                      const char *accept);
+static int jmap_calendarevent_getblob(jmap_req_t *req,
```
rsto, line 89: Nit: probably now is a good time to bundle some of these arguments in a getblob_req or the like?

rsto (approved):
This looks good. It's great to detangle the socket from the getblob routines.

ksmurchison:
@rsto   Any reason to not merge this at this time?

rsto:
@ksmurchison no, go ahead!

ksmurchison:
@rsto Can I can a final review, mainly on the last 3 commits (the last kind of supercedes the second to last)

imap/jmap_calendar.c
```diff
@@ -1567,47 +1566,43 @@ static int _calendarevent_getblob_cb(const char *mailbox __attribute__((unused))
...
-static int jmap_calendarevent_getblob(jmap_req_t *req,
-                                      const char *blobid,
-                                      const char *accept_mime)
+static int jmap_calendarevent_getblob(jmap_req_t *req, jmap_getblob_context_t *ctx)
 {
     struct caldav_db *db = NULL;
     struct caldav_data *cdata = NULL;
     struct mailbox *mailbox = NULL;
     icalcomponent *ical = NULL;
+    static struct buf cbuf = BUF_INITIALIZER;
```
rsto, line 1596: Is using a static buffer really necessary here? We are mallocing all over the place in the blob code, so if static cbuf is just an optimization, I don't think it is warranted (and could cause future subtle bugs).

imap/jmap_calendar.c
```diff
@@ -1675,77 +1671,67 @@ static int jmap_calendarevent_getblob(jmap_req_t *req,
...
         annotatemore_findall(cdata->dav.mailbox, cdata->dav.imap_uid,
                              PER_USER_CAL_DATA, 0, _calendarevent_getblob_cb,
                              &rock, 0);
 
-        write_multipart_body(0, req->txn, NULL, 0, NULL);
-
-        free(content_type);
-        res = HTTP_OK;
+        /* Write close-delimiter and epilogue */
+        buf_printf(blob, "\r\n--%s--\r\n%s", boundary, epilogue);
```
rsto, line 1730: Shouldn't the epilogue be followed by the final boundary? The preamble is following the initial boundary string.

rsto (commented):
Generally this looks sane to me (see my 2 code review comments). I know that blobNotFound was requested, but I see two flaws with it:

1. It seems that blobNotFound is not a standard error type, neither in the core spec, nor the interim Contacts spec?
2. The core spec expects the server to return a list of all properties that were invalid during a /set{create,update}. But the blobNotFound error dominates all other invalid properties.

I can't make a judgement call if the specific blobNotFound error warrants to ignore the above points.

Last but not least: I ran this on Cassandane master and the blobId related tests break, as expected. I could not find a branch on Cassandane that validates the new changes? Could you please push them?

ksmurchison:
> Generally this looks sane to me (see my 2 code review comments). I know that blobNotFound was requested, but I see two flaws with it:
> 
>     1. It seems that blobNotFound is not a standard error type, neither in the core spec, nor the interim Contacts spec?

blobNotFOund is here: https://tools.ietf.org/html/rfc8621#section-4.6


>     2. The core spec expects the server to return a list of all properties that were invalid during a /set{create,update}. But the blobNotFound error dominates all other invalid properties.

My intention was to always report invalidProperties, and only report blobNotFound if the request was otherwise correct.  Do you see a specific instance where this is not the case?

> Last but not least: I ran this on Cassandane master and the blobId related tests break, as expected. I could not find a branch on Cassandane that validates the new changes? Could you please push them?

Done.  Branch jmap_getblob_handlers_v2

rsto:
Cool. Looks good to me.

PR #3348, opened 2021-02-09
message: use xsyslog for errors

elliefm:
More of this.  Please try building it, in case your compiler catches things mine doesn't.  Thanks :)


ksmurchison (approved):
Compiles cleanly for me with GCC 10.2.1

PR #3350, opened 2021-02-09
sieve.y: allow :specialuse + :mailboxid until FM UI catches up

ksmurchison:


ksmurchison:
No longer needed as of 2022-06-10

PR #3351, opened 2021-02-10
http_caldav: don't bail for dtstart == dtend

brong:


ksmurchison (approved):

dilyanpalauzov:
Why is it fine to have DTSTART == DTEND ?

Can the line
> //            !icaltime_get_timezone(dtend) != !icaltime_get_timezone(dtstart) ||

be removed?

ksmurchison:
@brong FYI, I forced pushed a rebase of this PR

brong:
> Why is it fine to have DTSTART == DTEND ?
> 
> Can the line
> 
> > //            !icaltime_get_timezone(dtend) != !icaltime_get_timezone(dtstart) ||
> 
> be removed?

It's probably not fine - but I suspected that there were real events like that being stored to our system, so I checked and indeed there were.  So we at least have to make sure our own code is safe, but also maybe make Cyrus fix incoming events by stripping the DTEND if it matches DTSTART.

dilyanpalauzov:
As far as the already uploaded data is concerned, nothing changes.  Fixing CUAs to send correct data and servers to reject invalid data is a loop, that can be demonstrated by examples:
* When I found that DAVx⁵ has not submitted same-ORAGNIZER-in-all-components, both the CUA was fixed and the server started enforcing it
* DAVx⁵ can upload a recurring task with ORGANIZER:mailto:abc, then Evolution downloaded it, changed an instance and uploaded it, Evolution used ORGANIZER:MAILTO:abc .  `MAILTO:` and `mailto:` as schema names are supposed to be the same, but [Cyrus does not think so](https://github.com/cyrusimap/cyrus-imapd/pull/2723), and still rejects valid input (thinks that organizers are different, but they are the same, just mailto is with different case).  Eventually Evolution [was altered](https://gitlab.gnome.org/GNOME/evolution/-/issues/382) to send lowercased `mailto:`
* DAVx⁵ uploaded VTIMEZONES with two X- properties, which was valid iCalendar, but was rejected as invalid.
* Evolution tried to upload [vCard without FN:](https://gitlab.gnome.org/GNOME/evolution/-/merge_requests/41/diffs?commit_id=752d8b2a687d92e2d1f2d73ee454eab73c408416) , the server rejected it and the CUA was fixed.

When one CUA uploads invalid data and another CUAs downloads the invalid data, how shall the latter CUAs interpret the data?  Enforcing on upload any checks is the mean to convince clients to upload valid data.

If you have examples of invalid data, then you see which CUAs have uploaded the data, so you can understand for each CUA what it meant, fix the iCalendar files on the server and communicate the problem to the CUA publisher.

dilyanpalauzov:
Withdrawing this?

dilyanpalauzov:
I changed by mind.  When DTSTART and DTEND are of type VALUE=DATE, then DTSTART must be before DTEND.  Buf if the value type is DATE-TIME, then DTSTART can be the same as DTEND.  (And is the same, per definition, if DURATION and DTEND) are missing.

rjbs:
Today we have tentatively agreed to remove this and declare that we can't fix other people's problems…

imap/http_caldav.c
```diff
@@ -3652,9 +3652,13 @@ static int validate_dtend_duration(icalcomponent *comp, struct error_t *error)
             return HTTP_FORBIDDEN;
         }
         if (icaltime_compare(dtend, dtstart) < 1) {
-            error->desc = "DTEND must occur after DTSTART";
-            error->precond = CALDAV_VALID_DATA;
-            return HTTP_FORBIDDEN;
+            if (icaltime_compare(dtend, dtstart) < 0) {
```
dilyanpalauzov, line 3655: How about storing the result of icaltime_compare() in a variable once, and the use the variable twice?

dilyanpalauzov: What means only used by Fastmail?  Everbody using the public code does have the extra calculation.

ksmurchison: This is a pull request, NOT code in master

rjbs:
We should merge this.  Sadly, this crappy data just exists in the wild.

rjbs:
Actually, we have decided we should drop the syslog *and* abort behavior.  This stinks, but what can you do?  The amount of syslog this generates at present is too much.

rsto (approved):
LGTM

PR #3353, opened 2021-02-14
lib/: move some symbols to the Read-Only segment

dilyanpalauzov:
lib/auth_unix.c:allowedchars and lib/charset.c:base_64.


elliefm:
Great, thanks :)

PR #3355, opened 2021-02-15
propfind_caluseraddr_all(): memory leak

dilyanpalauzov:


dilyanpalauzov:
https://github.com/cyrusimap/cyrus-imapd/pull/3381 takes for this case a different approach.

PR #3357, opened 2021-02-15
Contact blobid: Always return 'blobId' in /set{create.update} and  /get responses.

ksmurchison:
Matching PR for Cassandane exists

### DIFF
--- a/changes/next/contact_blobid
+++ b/changes/next/contact_blobid
@@ -1,0 +1,13 @@
+Description:
+
+Always return 'blobId' in /set{create.update} and  /get responses.
+
+
+Config changes:
+
+None.
+
+
+Upgrade instructions:
+
+None.
--- a/changes/next/contact_blobid
+++ b/changes/next/contact_blobid
@@ -1,6 +1,6 @@
 Description:
 
-Always return 'blobId' in Contact/set{create.update} and Contact/get responses.
+Always return 'blobId' and 'size' in Contact/set{create,update} and Contact/get responses.
 
 
 Config changes:
--- a/changes/next/contact_blobid
+++ b/changes/next/contact_blobid
@@ -1,6 +1,6 @@
 Description:
 
-Always return 'blobId' in /set{create.update} and  /get responses.
+Always return 'blobId' in Contact/set{create.update} and Contact/get responses.
 
 
 Config changes:
--- a/imap/jmap_api.h
+++ b/imap/jmap_api.h
@@ -171,10 +171,18 @@ typedef struct {
     const char *from_accountid;  // input to the handler
     const char *blobid;          // input to the handler
     const char *accept_mime;     // input to the handler
+    unsigned decode : 1;         // input to the handler
     struct buf blob;             // output from the handler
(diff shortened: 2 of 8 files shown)


changes/next/contact_blobid
```diff
@@ -0,0 +1,13 @@
+Description:
+
+Always return 'blobId' in /set{create.update} and  /get responses.
```
rsto, line 3: I'd state the JMAP object type that this change applies to. e.g. "Always return 'blobId' in in JMAP Contact /set...."

imap/jmap_contact.c
```diff
@@ -1031,9 +1032,9 @@ static void _contacts_set(struct jmap_req *req, unsigned kind)
         struct entryattlist *annots = NULL;
         strarray_t *flags = NULL;
 
+        json_t *item = json_pack("{}");
```
rsto, line 1035: `json_pack("{}") ` can be simplified in this case to `json_object()`

rsto (commented):
Generally looks good to me. Just found two nits.

rsto (approved):
Looks good to me

ksmurchison:
@rsto One last review of the final two commits please.  Forgot to use the generic blob fetcher for email parts.

imap/jmap_api.h
```diff
@@ -171,11 +171,19 @@ typedef struct {
...
     const char *accept_mime;     // input to the handler
+    unsigned decode : 1;         // input to the handler
     struct buf blob;             // output from the handler
-    const char *content_type;    // output from the handler
+    char *content_type;          // output from the handler
+    char *encoding;              // output from the handler
     const char *errstr;          // output from the handler
 } jmap_getblob_context_t;
 
+#define GETBLOB_CTX_INITIALIZER(from_accountid, blobid, accept_mime, decode) \
```
rsto, line 181: This deviates from the currently used idioms in the codebase: None of the current INITIALIZER macros takes arguments. Instead they initialize a struct with default values that make the data structure ready for use without further initialization. (And the way the macro currently is defined is not safe with regards to potential duplicate evaluation).

For parameterized initializers there's a couple of foo_init functions, but unfortunately they do not all follow the same pattern. Some are void functions, others return an initialized struct. I tend to use void functions.

For the particular case, though, I would just not use an INITIALIZER at all. Is the blob context initialized at more than one code section?

imap/jmap_api.h
```diff
@@ -171,11 +171,19 @@ typedef struct {
...
+    { from_accountid, blobid, accept_mime, decode,                           \
+            BUF_INITIALIZER, NULL, NULL, NULL };
+
+void jmap_getblob_ctx_free(jmap_getblob_context_t *ctx);
```
rsto, line 185: This should read `_fini` rather than `_free`.

rsto (changes requested):
Found two idiomatic issues that I think should be addressed.

imap/jmap_api.h
ksmurchison: Fixed in latest commit

PR #3363, opened 2021-02-19
xapian_wrap:get_stopper(): skip ctor calls, when the function returns early

dilyanpalauzov:
Skip destructor calls, too.


rsto:
Great, thanks.

PR #3365, opened 2021-02-19
imap:caldav_store_resource() skip a buf_reset(&txn->buf) call

dilyanpalauzov:
Essentially the code contains
```c
buf_reset(&txn->buf); /* this is now removed */
/* txn->buf is not used here */
buf_reset(&txn->buf); /* again */
```


ksmurchison:
Fixed by removing the second call rather than the first

PR #3375, opened 2021-02-21
http_dav.c:principal_search() utilize DAV:displayname from the annotations

dilyanpalauzov:
When calling [REPORT principal-property-search](https://tools.ietf.org/html/rfc3744#section-9.4) and the search criterion is `DAV:displayname`, take the `displayname` from the annotatitonsDB, if available.
```sh
curl -XREPORT -H"content-type:application/xml" -uaaa:bbb --data-binary @- <<EOF https://SERVER/dav/principals/user

<principal-property-search xmlns="DAV:" xmlns:c="urn:ietf:params:xml:ns:caldav">
   <apply-to-principal-collection-set/>
   <property-search>
     <prop>
         <displayname/>
     </prop>
     <match>aaa</match>
   </property-search>
   <prop>
     <c:calendar-user-address-set/>
     <c:calendar-home-set/>
     <c:calendar-user-type/>
     <displayname/>
   </prop>
 </principal-property-search>
```

### DIFF
--- a/changes/next/dispayname_on_principal-property-search_report
+++ b/changes/next/dispayname_on_principal-property-search_report
@@ -1,0 +1,11 @@
+Description:
+
+The DAV:principal-property-search REPORT honours the previously set DAV:displayname property on matches
+
+Config changes:
+
+None
+
+Upgrade instructions:
+
+None
--- a/imap/http_dav.c
+++ b/imap/http_dav.c
@@ -1877,7 +1877,7 @@ HIDDEN void dav_get_principalname(const char *userid, struct buf *buf)
     buf_reset(buf);
 
     if (userid) {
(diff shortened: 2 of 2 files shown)


ksmurchison:
Can you update the patch to apply to current master?

dilyanpalauzov:
I have updated the code, which conflicted/largely coincided with 643e3b0c2d5201babda7e3cac4c.

When the currently logged user has `l` right on the principal/#calendars mailbox, then this principal is considered in the search.

In my eyes it is not nice, if after I contribute some improvement, it is waited until the changeset does not apply anymore straightforward, and then I am asked to adapt it again.  I have forgotten alls the details/context and have to start anew with the changes.  This eats my time, but is at the same time avoidable.

ksmurchison (approved):

PR #3380, opened 2021-02-24
Dav reconstruct audit

brong:
This adds extra smarts around the audit tool for dav-reconstruct, so we know which slot it's running on and which user it's for from the command line.

This will be paired with a Perl tool which can then read said config and actually diff the two databases.


ksmurchison (approved):
This all looks sane.  Thanks for the small commits which make it easy to follow both the changes and logic in doing so.

PR #3381, opened 2021-02-24
For DAV:principal-property-search REPORT match the calendar-user-address-set request with the annotations

dilyanpalauzov:
This change fixes errors, reduces the total number of lines and introduces new feature at the same time.

* introduce get_calendar_user_address_set_for_principal()

* jmap_calendars.c:get_schedule_address_set() used "#calendars" instead of config_getstring(IMAPOPT_CALENDARPREFIX)

* in jmap_calendars.c:get_schedule_address_set() the code used to honour the annotations for httpd_userid, but if they were missing then use the passed userid to generate the default schedule addresses.  This makes no sense.  This changes the behaviour to always check the annotations for "userid" and not for "httpd_userid".

* jmap_calendars.c:get_schedule_address_set() has not iterated over cua_domains.

### DIFF
--- a/changes/next/get_calendar_user_address_set_for_principal
+++ b/changes/next/get_calendar_user_address_set_for_principal
@@ -1,0 +1,11 @@
+Description:
+
+Honour the real CalDAV:calendar-user-address-set WebDAV property for matches in the DAV:principal-property-search report.
+
+Config changes:
+
+None
+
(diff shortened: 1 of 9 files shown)


ksmurchison:
Is the PR outdated or does it need to be updated to apply to current master?

Issue #3148, opened 2020-08-18
HTML-escape the display name of the calendars in list_calendars()


dilyanpalauzov:
imap/http_caldav.c:list_calendars() calls for each calendar
```c
	buf_printf_markup(body, level, "<td>%s%s%s",
                          (cal->flags & CAL_IS_DEFAULT) ? "<b>" : "",
                          cal->displayname,
                          (cal->flags & CAL_IS_DEFAULT) ? "</b>" : "");
```
When cal->displayname contains `&`, `>` or `<` this messes the HTML code and results e.g.
```html
…<tr><td>a'"<.&a<td><select multiple name=comp size=3 onChange="compsetCalendar('/dav/calendars/user/a/a22', 'a'"<.&a', this.options)"><option selected>VEVENT</option><option selected>VTODO</option>…
```
so onChange cannot be called. Likewise in `list_addressbooks()`

dilyanpalauzov:
Moved to https://github.com/cyrusimap/cyrus-imapd/pull/3389.

PR #3389, opened 2021-02-27
http_ca(l,rd)dav: list collections: html escape the displayname

dilyanpalauzov:
Otherwise >, < and & in the displayname of calendars/address books are not converted to `&gt;` `&lt;` or `&amp;` and the generated HTML does not work.

The HTML is rendered when visiting:
* https://server/dav/calendars/user/zzz/ and
* https://server/dav/addressbooks/user/zzz/

Problem description at https://github.com/cyrusimap/cyrus-imapd/issues/3148.


dilyanpalauzov:
The single and double quotes must also be html-escaped with `&apos;` and `&quot;`, by extending `lib/charset.c:uni2html()`, but this code
```html
<html><head>
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
</head><body>
    <button onclick='window.confirm("A&amp;B&quot;C&apos;D")'>E</button>
    <button onclick="window.confirm('F&amp;G&quot;H&apos;I')">J</button>
</body></html>
```
works with no browser: there is always a syntax error.  So html escaping single and double qoutes leads from one problem to another, and can be therefore skipped.

dilyanpalauzov:
The last change solves the problem with ' or " in displayname: it is now shown in dialogs.

When the DAV:displayname contains ' or ", generating valid HTML for `<button onclick='anyFunction("THEDISPLAYNAME")'>B</button>` is very hard.  Therefore the rendered HTML in list_calendars() includes an id attribute to the <td> with the displayname and passes around the id.  The callee uses then `window.confirm(document.getElementById(id).innerText)` which can show ' and " to the mortals.

ksmurchison (approved):

PR #3390, opened 2021-02-28
imap/http_ca(l,rd)dav.js: use "DAV:" as default XML namespace

dilyanpalauzov:
and not in the D: prefix.  Works with Chromium, Firefox and WebKit.


ksmurchison:
Is there a technical reason for making this change?

ksmurchison:
We're only talking about a handful of characters.  I prefer to keep the elements prefixed with the namespace.

PR #3393, opened 2021-03-01
jmap_mail: do not use guidsearch for divergent non-Xapian criteria

rsto:
Fixes a bug in guidsearch where a query may return false results if its DNF clauses contain divergent non-Xapian criteria.                                                                                       

This reduces the number of queries that can be handled by guidsearch and may result in search latency regressions for such queries. 

Tested in https://github.com/cyrusimap/cassandane/commit/89616c9787881e3b20004b66e3d5429e0ec678b5

### DIFF
--- a/changes/next/guidsearch_mixed_clauses
+++ b/changes/next/guidsearch_mixed_clauses
@@ -1,0 +1,14 @@
+Description:
+
+Fixes a bug in guidsearch where a query may return false results if its DNF clauses contain divergent non-Xapian criteria.
+
+This reduces the number of queries that can be handled by guidsearch and may result in search latency regressions for such queries.
+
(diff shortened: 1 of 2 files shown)


elliefm:
Do I want this on 3.4?  I think so, what do you think?

rsto:
Yes, I think, too.

brong (approved):
Oh yeah, I already read this!  Yes, it looks great - I'll tag it in now.

elliefm:
Are we happy with how this is going in our builds?  Can we merge it to master?

rsto:
Done

elliefm:
This is on 3.4 now too, and I've retagged the cass test to include that version

PR #3394, opened 2021-03-01
Jmap contact new blobids

ksmurchison:


imap/jmap_contact.c
```diff
@@ -863,6 +863,34 @@ static int _add_othergroup_entries(struct jmap_req *req,
...
+} property_blob_t;
+
+static property_blob_t *property_blob_new(const char *key, const char *prop,
+                                          char *type, struct buf *data)
+{
+    property_blob_t *blob = xzmalloc(sizeof(property_blob_t));
+
+    blob->key = key;
+    blob->prop = prop;
+    blob->type = type;
```
rsto, line 881: I'd suggest not to transfer ownership here, it's not really clear from the function signature. Since the only caller of property_call_new passes a xstrdup'ed copy, I think that type should become a const pointer and the xstrdup is done within the function.

changes/next/jmap_contact_new_blobids
```diff
@@ -0,0 +1,13 @@
+Description:
+
+Use mailbox id + message UID as blob id
```
rsto, line 3: Nit: this doesn't make clear for what kind of blobs this change applies. It's JMAP CalendarEvent, Contact and ContactGroup, right?

imap/jmap_contact.c
```diff
@@ -863,6 +863,34 @@ static int _add_othergroup_entries(struct jmap_req *req,
...
+    blob->size = buf_len(data);
+    message_guid_generate(&blob->guid, buf_base(data), buf_len(data));
+
+    return blob;
+}
+
+static void property_blob_free(property_blob_t *blob)
```
rsto, line 888: If the blob memory gets freed, then use a `**blobptr` argument and reset it to NULL. Otherwise there could be the risk of double frees later.

rsto (changes requested):
The logic looks good to me, but I found a few non-logic related changes.

PR #3395, opened 2021-03-01
Jmap rawdata blobids

ksmurchison:
Consolidates encoding and decoding of blobIds for raw iCalendar/vCard data

### DIFF
--- a/imap/jmap_util.c
+++ b/imap/jmap_util.c
@@ -698,6 +698,16 @@ done:
     return text;
 }
 
+/*
+ * The blobId syntax for raw message data is:
+ *
+ * <mailbox id> "_" <message UID> [ "_" <userid> [ "_" <subpart> [ "_" <SHA1> ]]]
+ *
+ * <userid> is currently used to personalize iCalendar data and may be empty
+ * <subpart> is currently used to target vCard/iCalendar properties
+ *   with data: URI values (e.g. vCard PHOTO/LOGO/SOUND or iCalendar IMAGE)
+ * <SHA1> is used to target a <subpart> having a specific value
+ */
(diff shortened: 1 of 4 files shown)


imap/jmap_util.c
```diff
@@ -696,3 +697,128 @@ EXPORTED const char *jmap_decode_to_utf8(const char *charset, int encoding,
...
+        size_t len = strlen(base);
+        if (len) {
+            userid = jmap_decode_base64_nopad(base, len);
+            if (!userid) goto done;
+        }
+        base += len;
+    }
+
+    /* Decode subpart */
+    if (subpartptr && *base == '_') {
```
rsto, line 795: Wouldn't this break if the caller passes a null useridptr and a non-null subpartptr (or the other way round) and the blobId contains both? I think that if the caller passes null arguments, then the decoder should validate but skip these parts in the blobId.

rsto (changes requested):
This looks all sane, except for a potential bug in the blobId decoder.

imap/jmap_util.c
ksmurchison: I struggled with how best to handle this, since currently only CalendarEvent uses 'userid' and only Contact uses 'subpart' (which I renamed from 'property').  I thought about not having 'userid' argument and just passing the userid in as the 'subpart' because there currently isn't a use-case for 'userid' with Contact and there may never be one.  However, I could see a use-case for both 'userid' and 'subpart' with CalendarEvent, e.g., A per-user IMAGE property on a VEVENT.

Perhaps, we could write the encoder so that if a subpart is used, the userid MUST be encoded, even if empty, so mboxid__subpart_guid (consecutive underscores between mboxid and subpart).

rsto (approved):
Looks good to me

PR #3396, opened 2021-03-04
jmap_ical: gracefully handle empty property values

rsto:
Fixes a bug where libical returns NULL for empty property values, causing Cyrus to crash.


ksmurchison (approved):
Looks sane to me

PR #3397, opened 2021-03-05
fix loss of replication protocol sync when APPLY MESSAGE fails to write to disk

### LINKED ISSUE #3360: write failure during APPLY MESSAGE causes replication protocol desync
Happens rarely, mostly during disk full events?  Syslog will report the real errors:

```
2021-02-16T15:05:47.198347-05:00 hostname (err) syncserver[2121610]  IOERROR: failed to upload file sha1sum
2021-02-16T15:05:47.199070-05:00 hostname (err) syncserver[2121610]  IOERROR: received bad command: command=<Apply> func=<cmdloop>
```

(aside: it might be useful to log more than just "Apply", if we can -- we're specifically in an APPLY MESSAGE here...)

followed by a load of junk ones:

```
2021-02-16T15:05:47.199163-05:00 hostname (err) syncserver[2121610]  IOERROR: received bad command: command=<Return-path:> func=<cmdloop>
2021-02-16T15:05:47.199194-05:00 hostname (err) syncserver[2121610]  IOERROR: received bad command: command=<Received:> func=<cmdloop>
[...]
```

We try to prevent this, and we mostly get it right, but we miss a detail.

The diagnosis begins in `reservefile()` in dlist.c, where we fail to open a new temporary file for writing, and log the original "failed to upload file" message:

https://github.com/cyrusimap/cyrus-imapd/blob/53e5b3142e91a02baea5a3aad30852c307f3414c/imap/dlist.c#L185-L192

Note the comment: we've thought about this before!

`reservefile()` does the right thing: consumes the entire uploaded file contents off the wire, then returns the error up to the caller, which is `dlist_parse()`.  `dlist_parse()` seems to do the right thing here, the upload has already been eaten so it just returns the error up the chain.

This seems like it's probably fine when the APPLY MESSAGE only contains a single message upload.  But a single APPLY MESSAGE command can contain multiple message uploads.  And it looks like, if any one of them fails, _that one_ gets eaten and the error returned up the chain, but any subsequent uploads remain on the wire.

Here's how `sync_parseline()` reads the dlist parts of the protocol:

https://github.com/cyrusimap/cyrus-imapd/blob/53e5b3142e91a02baea5a3aad30852c307f3414c/imap/sync_support.c#L1858-L1872

Note that if the dlist parse fails, we `eatline()` trying to get back to a sane baseline.  `eatline()` does the right thing with respect to IMAP literals, but replication has richer syntax that it doesn't know about:

https://github.com/cyrusimap/cyrus-imapd/blob/53e5b3142e91a02baea5a3aad30852c307f3414c/imap/imapparse.c#L444-L483

I think what we probably need is a `sync_eatline()`, that does the same thing as the imap-specific one, but also knows the sync syntax extensions.  Though, given imapd can function as a sync server, maybe we really need to add sync awareness to the generic imap eatline, not sure.

This is not urgent; it's easy enough to recover by just restarting the sync server, and letting it catch up on the next replication (assuming the underlying filled disk problem has been resolved).

elliefm:
* Fixes loss of protocol sync in APPLY MESSAGE by adding a `sync_eatline()` that can also consume the remaining dlist file literals correctly
* Also fixes a quiet bug whereby imapd would eatline() twice on sync failures, discarding the next command as well
* And adds a new imapd option "debug_writefail_guid" for simulating write failures for the given guid, to enable regression testing for this case

Corresponding Cassandane tests are here: https://github.com/cyrusimap/cassandane/compare/master...elliefm:v35/3360-apply-message-write-failure-desync

Fixes #3360

### DIFF
--- a/changes/next/3360-apply-message-write-failure-desync
+++ b/changes/next/3360-apply-message-write-failure-desync
@@ -1,0 +1,11 @@
+Description:
+
+Bug fix for #3360
+
+Config changes:
+
+debug_writefail_guid
+
+Upgrade instructions:
+
+None required
--- a/imap/dlist.c
+++ b/imap/dlist.c
@@ -172,9 +172,24 @@ static int reservefile(struct protstream *in, const char *part,
                        struct message_guid *guid, unsigned long size,
                        int isbackup, const char **fname)
 {
+    static struct message_guid debug_writefail_guid = MESSAGE_GUID_INITIALIZER;
     FILE *file;
     char buf[8192+1];
     int r = 0;
+
+    if (debug_writefail_guid.status == GUID_UNKNOWN) {
+        const char *guidstr = config_getstring(IMAPOPT_DEBUG_WRITEFAIL_GUID);
+        if (guidstr) {
+            if (!message_guid_decode(&debug_writefail_guid, guidstr)) {
+                xsyslog(LOG_DEBUG, "debug_writefail_guid: ignoring invalid guid",
+                                   "guid=<%s>", guidstr);
+                message_guid_set_null(&debug_writefail_guid);
+            }
+        }
+        else {
+            message_guid_set_null(&debug_writefail_guid);
+        }
+    }
 
     /* XXX - write to a temporary file then move in to place! */
     *fname = dlist_reserve_path(part, /*isarchive*/0, isbackup, guid);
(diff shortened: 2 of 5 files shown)


elliefm:
I was hoping to also implement debug_writefail_guid for IMAP/JMAP uploads, but when I looked at how to hook it into IMAP there wasn't a great place to do so, since we don't know the message's GUID until after we've already written it to disk successfully, by which point a simulated write failure would be very different from a real one.  I didn't need it immediately anyway so I didn't bother, and since I wasn't bothering to do IMAP uploads, I didn't even look at JMAP.  So it's replication-only for now, and I guess we can figure out how to extend it when we need it.

ksmurchison (approved):
Looks good to me

PR #3399, opened 2021-03-08
JMAP Calendars: guess IANA timezones from embedded VTIMEZONEs

rsto:
VEVENTs may embed VTIMEZONEs which have a non-IANA TZID. In JMAP, these events map to a JSEvent which a custom timezone id and the custom timezone definition defined in the timeZones property. However, the majority of such embedded VTIMEZONEs can be mapped to equivalent IANA timezones, and this patch adds support for this.

Using the new tool `cyr_guesstzdb` it creates a database of timezones based on IANA-based VTIMEZONEs found in zoneinfo_dir. For each timezone it expands the observances of that timezone and stores them in the database along with the timezone identifier.

To guess the IANA name of a embedded VTIMEZONE, the JMAP implementation then looks up all timezones in guesstzd which have the same UTC offset as the start time of the event. if the timezone observances match over the time period of the event, then the IANA timezone is a valid replacement for the embedded timezone.


imap/cyr_guesstzdb.c
```diff
@@ -0,0 +1,187 @@
...
+
+    if (m == CREATE) {
+        /* Determine observance expansion time range */
+        const icaltimezone *utc = icaltimezone_get_utc_timezone();
+        icaltimetype dbstart = {
+            2000, 1, 1, 0, 0, 0, 0, 0, utc
+        };
+        icaltimetype dbend = {
+            2038, 1, 1, 0, 0, 0, 0, 0, utc
+        };
```
ksmurchison, line 143: Not sure how future-proof this type of initialization might be.  Might be easier to just use icaltime_from_string() with a const char *

imap/guesstz.c
```diff
@@ -0,0 +1,914 @@
...
+    }
+
+    int hours = offset / 3600;
+    int minutes = (offset % 3600) / 60;
+    int seconds = offset % 60;
+
+    if (hours < 0 || hours > 23 || minutes < 0 || minutes > 60 ||
+            seconds < 0 || seconds > 60) {
+        return 0;
+    }
```
ksmurchison, line 99: Will these ever be < 0?  offset is always > 0 at this point

imap/guesstz.c
```diff
@@ -0,0 +1,914 @@
...
+    buf_putc(data, '\0');
+
+    size_t tzidlen = buf_len(data);
+    buf_truncate(data, tzidlen + + nobs * sizeof(struct obsrec));
+    struct obsrec *myobs = (struct obsrec *) (data->s + tzidlen);
+    size_t i;
+    for (i = 0; i < nobs; i++) {
+        myobs[i].onset = htonll(obs[i].onset);
+        myobs[i].offset = htonl(obs[i].offset);
+    }
```
ksmurchison, line 201: I understand what's going on here (putting ints into a char array) but I wonder if we couldn't use buf_appendbit32() and buf_appendbit64()

imap/guesstz.c
```diff
@@ -0,0 +1,914 @@
...
+    FTSENT *fe;
+    while ((fe = fts_read(fts))) {
+        if (fe->fts_info != FTS_F) {
+            continue;
+        }
+
+        FILE *stream = fopen(fe->fts_path, "r");
+        if (!stream) continue;
+
+        icalparser *parser = icalparser_new();
```
ksmurchison, line 327: Wouldn't it be easier to just map_refresh() the file and use icalparser_parse_string()?

imap/guesstz.c
```diff
@@ -0,0 +1,914 @@
...
+        strarray_append(&preferred, "America/Nuuk");
+        strarray_append(&preferred, "Atlantic/Azores");
+        strarray_append(&preferred, "Europe/London");
+        strarray_append(&preferred, "Europe/Berlin");
+        strarray_append(&preferred, "Europe/Athens");
+        strarray_append(&preferred, "Indian/Mauritius");
+        strarray_append(&preferred, "Asia/Dhaka");
+        strarray_append(&preferred, "Australia/Melbourne");
+        strarray_append(&preferred, "Pacific/Norfolk");
+    }
```
ksmurchison, line 703: What's the purpose of this list?  Is this just used to select one of multiple matches for a given offset?

imap/guesstz.h
```diff
@@ -0,0 +1,66 @@
...
+extern int guesstz_create(const char *zoneinfo_dir, const char *fname,
+                            icaltimetype start, icaltimetype end);
+
+extern char *guesstz_dump(const char *fname);
+
+extern void guesstz_toiana(struct guesstzdb *gtzdb,
+                           struct buf *idbuf, icalcomponent *vtz,
+                           struct icalperiodtype span,
+                           unsigned is_recurring);
+
```
ksmurchison, line 65: It might be helpful here or in .c to have a comment describing the three different record types and then 'O' records point to 'T' records (and any other subtleties).

lib/dynarray.c
```diff
@@ -48,11 +48,23 @@
...
+    free(da->data);
+    memset(da, 0, sizeof(struct dynarray));
+}
+
+EXPORTED struct dynarray *dynarray_new(size_t membsize)
+{
+    struct dynarray *da = xzmalloc(sizeof(struct dynarray));
+    dynarray_init(da, membsize);
     return da;
 }
```
ksmurchison, line 70: How is dynarray different from the existing ptrarray?

ksmurchison (changes requested):
If this works, then overall I'm happy.  But I have added comments inline.

lib/dynarray.c
dilyanpalauzov: Does `dynarray_new()` call twice `memset(da, 0, sizeof(struct dynarray))` on the same data in a row: once in xzmalloc() and once in `dynarray_init()` (called from dynarray_new)?

imap/cyr_guesstzdb.c
rsto: Done

imap/guesstz.c
rsto: Right, I updated this to only check positive numbers. I also changed it to reject minutes or seconds > 59

imap/guesstz.c
rsto: That's the better approach, indeed. The current approach suffered from struct padding anyway.

imap/guesstz.c
rsto: done

lib/dynarray.c
rsto: Is is more cache-friendly than ptrarray. A ptrarray requires to dereference a pointer for each array member.  A dynarray keeps a contiguous memory block of same-sized elements, so it's faster to traverse them.

rsto: In addition, it allows us to save n malloc/free calls for an array of n members

imap/guesstz.c
rsto: Yes, most likely multiple timezones will match, and we prefer these names.

lib/dynarray.c
rsto: @dilyanpalauzov right, I now changed the `dynarray_new` function to just `xmalloc`, rather than `xzmalloc`.

imap/guesstz.h
rsto: I added both a comment describing the database schema, as well as documented the exported functions.

PR #3405, opened 2021-03-10
sync_support.c: don't try to activate a script on the replica if it d…

ksmurchison:
…oesn't exist


brong (approved):
Yep, this looks good.

PR #3408, opened 2021-03-11
index: use attachment extractor for text/rtf message bodies

rsto:
Before,the raw RTF file including control words could end up in a search snippet preview.

Tested in https://github.com/cyrusimap/cassandane/commit/a95a1b05c1d418104c223bba7d08fe920b620968


ksmurchison (approved):

elliefm:
I've cherry-picked this to 3.4 as 44869b5415 because it was unintrusive and sensible.  The Cassandane test for it already expected it to be present on 3.4, and now it is.

PR #3409, opened 2021-03-11
3.2 - timsieved/actions.c: fix memory leak

dilyanpalauzov:
When UNAUTHENTICATE + AUTHENTICATE is called, the variable `sieved_userid` is leaked.

Backport of 43e145ab7b9f.


dilyanpalauzov:
Apparently backporting memory leaks is not a priority.

PR #3410, opened 2021-03-11
fix up explicit symbol visibility

elliefm:
This fixes the linking issues raised in https://cyrus.topicbox.com/groups/devel/Tf544987d7d6a54fd-M64b7419c76035feff7fb6a22

The Makefile.am changes only apply to imap_libcyrus_imap_la, because that is currently the only target we build that has C++ sources.  If we start including C++ sources in other targets, we'll need to set their target_CXXFLAGS similarly.  Which we would need to do anyway, and now there's a good example to copy from.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -980,7 +980,7 @@ imap_libcyrus_imap_la_LIBADD = \
     lib/libcyrus.la
 
 imap_libcyrus_imap_la_CFLAGS = $(AM_CFLAGS) $(CFLAG_VISIBILITY)
-imap_libcyrus_imap_la_CXXFLAGS = $(AM_CXXFLAGS)
+imap_libcyrus_imap_la_CXXFLAGS = $(AM_CXXFLAGS) $(CFLAG_VISIBILITY)
 
 imap_libcyrus_imap_la_SOURCES = \
 	imap/annotate.c \
--- a/imap/message_test.c
+++ b/imap/message_test.c
@@ -307,7 +307,7 @@ static int usage(const char *name)
     exit(EX_USAGE);
 }
 
-void fatal(const char* s, int code)
+EXPORTED void fatal(const char* s, int code)
 {
     fprintf(stderr, "message_test: %s\n", s);
     cyrus_done();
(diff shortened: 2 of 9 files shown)


Makefile.am
```diff
@@ -980,7 +980,7 @@ imap_libcyrus_imap_la_LIBADD = \
     lib/libcyrus.la
 
 imap_libcyrus_imap_la_CFLAGS = $(AM_CFLAGS) $(CFLAG_VISIBILITY)
-imap_libcyrus_imap_la_CXXFLAGS = $(AM_CXXFLAGS)
+imap_libcyrus_imap_la_CXXFLAGS = $(AM_CXXFLAGS) $(CFLAG_VISIBILITY)
```
dilyanpalauzov, line 984: Does it work, when CFLAG_VISIBILITY is passed instead to imap_libcyrus_LDFLAGS and removed from …CXXFLAGS and …CFLAGS?

imap/message_test.c
```diff
@@ -307,7 +307,7 @@ static int usage(const char *name)
     exit(EX_USAGE);
 }
 
-void fatal(const char* s, int code)
+EXPORTED void fatal(const char* s, int code)
```
dilyanpalauzov, line 310: If all fatal() are explicitly EXPORTED, can CFLAG_VISIBILITY be set in AM_LDFLAGS (globally for all targets) instead per target?

Makefile.am
elliefm: Off the top of my head, probably.  But what we already have is well tested, I intend to backport this fix to the stable branches, so for now I'm keeping things as similar as possible to what we already know works.

imap/message_test.c
elliefm: Yep (either AM_LDFLAGS or AM_CFLAGS/AM_CXXFLAGS), and I plan to do something like this for master-only once 3.4 is released.

dilyanpalauzov: Looking forward to see this in action.  When I did it in 19933ff48200387b the only problem was, as far as I remember, that the `fatal()` were not exported from the executables to the libraries.

brong (approved):

rsto (approved):

imap/message_test.c
dilyanpalauzov: > Yep (either AM_LDFLAGS or AM_CFLAGS/AM_CXXFLAGS), and I plan to do something like this for master-only once 3.4 is released.

Now 3.4 is released.

PR #3411, opened 2021-03-12
Imapd no jmapnotification

brong:
We were missing the header definitions for 'mboxname_isjmapnotificationmailbox' and the inclusion in isnonimapmailbox - but also imapd had the list hard-coded too!  Fix both of those.

Found by inspection on Fastmail dogfood server :)


rsto (approved):
Looks good to me!

PR #3413, opened 2021-03-12
search: do not extract text after the search_maxsize limit

rsto:
This updates the indexing and search snippet code to stop extracting text when at most `search_maxsize` limit octets have been read. Any bytes after that limit already have been ignored during indexing and snippet generation, but the decoder ran through the complete message part, before throwing away the excess bytes.

Before, the limit was hardcoded to 4MB, now it is configurable in imapd.conf.

For discussion: this patch also changes the limit to apply not the the total bytes extracted from a message, but to each message part separately. This has the potential to increase the search index database, but prevents text being omitted depending on the order in which message parts are processed.

Tested in https://github.com/cyrusimap/cassandane/tree/search_large_text

### DIFF
--- a/lib/charset.h
+++ b/lib/charset.h
@@ -142,11 +142,17 @@ extern char *charset_unfold(const char *s, size_t len, int flags);
 
 extern int charset_decode(struct buf *dst, const char *src, size_t len, int encoding);
 
-/* Extract the body text for the message denoted by 'uid', convert its
-   text to the canonical form for searching, and pass the converted text
-   down in a series of invocations of the callback 'cb'.  This is
-   called by index_getsearchtext to extract the MIME body parts. */
-extern int charset_extract(void (*cb)(const struct buf *text, void *rock),
+/* Extract the body text contained in 'data' and with character encoding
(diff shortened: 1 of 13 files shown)


lib/charset.h
```diff
@@ -142,11 +142,17 @@ extern char *charset_unfold(const char *s, size_t len, int flags);
...
-   called by index_getsearchtext to extract the MIME body parts. */
-extern int charset_extract(void (*cb)(const struct buf *text, void *rock),
+/* Extract the body text contained in 'data' and with character encoding
+ * 'charset' and body-part encoding 'encoding'. The 'subtype' argument
+ * defines the MIME subtype (assuming that the main type is 'text').
+ * Extraction is done by a series of invocations of the callback 'cb'.
+ * Extraction stops when either the body text is fully extracted, or
+ * 'cb' returns a non-zero value, which is then returned to the caller.
+ * If 'charset' is unknown, then the function returns early without
+ * errror and never calls 'cb'.
```
elliefm, line 152: "errror" -> "error" 😅

rsto: Fixed, thanks :)

brong (approved):
Yeah, this looks great

PR #3415, opened 2021-03-15
dynarray: add dynamic array helper for same-sized members

rsto:
dynarray is a more cache-friendly variant of good old ptrarray. In almost all cases, we currently use ptrarray with same-sized members. Each member is alloced and freed separately, resulting in both CPU overhead and memory fragmentation. dynarray addresses both of the deficiencies by keeping its members in a contiguous memory buffer.

dynarray got introduced in the jmap-calendars-01 branch but I find it more and more useful to also include in other feature branches already.

### DIFF
--- a/lib/dynarray.c
+++ b/lib/dynarray.c
@@ -1,0 +1,184 @@
+/* dynarray.c -- an expanding array of same-size members
+ *
+ * Copyright (c) 1994-2020 Carnegie Mellon University.  All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ * 1. Redistributions of source code must retain the above copyright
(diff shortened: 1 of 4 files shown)


ksmurchison (approved):

lib/dynarray.c
```diff
@@ -0,0 +1,184 @@
...
+        idx += da->count;
+        if (idx >= 0 && len)
+            ensure_alloc(da, da->count+len);
+    } else if (len) {
+        ensure_alloc(da, da->count+len);
+    }
+    return idx;
+}
+
+EXPORTED int dynarray_size(struct dynarray *da)
```
dilyanpalauzov, line 124: Can the function `dynarray_size()` be converted to macro in dynarray.h?  This saves a function call any time the size is requested.

Likewise for `arrayu64_size()`, `ptrarray_size()`, `strarray_size()`, `bufarray_size()`.

PR #3420, opened 2021-03-19
send BYE from sync_server on clean shutdown

### LINKED ISSUE #3401: sync_server: send some kind of BYE so sync_client can distinguish between clean remote shutdown and error
@robmueller observes that it's "a big and annoying noise generator at the moment when you restart slots"

```
IOERROR: zero length response: command=<MAILBOX> prot_error=<end of file reached> func=<sync_parse_response>
do_folders(): update failed: example.com!user.foo 'Bad protocol'
   Promoting: MAILBOX example.com!user.foo -> USER foo@example.com
IOERROR: zero length response: command=<RESTART> prot_error=<end of file reached> func=<sync_parse_response>
Error in do_sync(): bailing out! Bad protocol
SYNCERROR sync_rightnow: error syncing to: 10.37.129.211 (Bad protocol)
```

> so what i want is that if sync_server is shutdown "cleanly" by shutting down the master, the sync_client on the other end doesn't log an IOERROR due to seeing an EOF condition
> the problem is the whole protocol makes this a bit messy because sync_client waits on a log file, and when the log file appears, it sends commands to syncserver and listens for responses for each command.
> so we can't easily detect a "socket went away" state because it's blocked on waiting for the sync file and doesn't care about the socket until it's ready to do stuff with it

and his suggested "99.9% hack" is:

> * change `signals_set_shutdown(&shut_down)`; to `signals_set_shutdown(&shut_down_via_signal);`
> * make shut_down_via_signal shove a BYE down sync_out and then call shut_down
> 
> then change sync_client.c to detect a BYE message at any point, and if it sees one log a "server gracefully disconnected", and fall as far back up to it's main loop as possible

This seems like there's no issue with back-compat because an older sync_client that doesn't recognise BYE was going to whinge about the EOF and exit anyway, so it whinging about the unrecognised BYE first isn't a new problem...

elliefm:
* When sync_server is shut down cleanly by a signal from master, it now writes "BYE shutting down" to the client socket, if it's connected, before shutting down normally.
* When sync_client sees "BYE shutting down" from sync_server, it knows to expect to be disconnected, and exits cleanly itself (instead of logging a lot of useless IOERRORS).

There are tests here (including infrastructure fixes to make this testable at all): https://github.com/cyrusimap/cassandane/compare/master...elliefm:v35/3401-sync-server-bye

This is upgrade safe:  when the master Cyrus is older than the replica, and doesn't recognise BYE, it IOERRORs about a bad command and bails out (without the BYE, it would have IOERRORed about the zero length command, and bailed out).  When the replica Cyrus is older, and doesn't send BYE, the master will IOERROR about the zero-length command and bail out like it always did.

This doesn't change anything about replication-over-imap, and maybe it should, but I haven't looked at it (yet).

Fixes #3401

### DIFF
--- a/changes/next/sync-server-bye
+++ b/changes/next/sync-server-bye
@@ -1,0 +1,12 @@
+Description:
+
+sync_client can now recognise when the sync_server has been shut down cleanly
+instead of logging a bunch of errors
+
+Config changes:
+
+(none)
+
+Upgrade instructions:
+
+(none)
--- a/imap/sync_server.c
+++ b/imap/sync_server.c
@@ -152,6 +152,7 @@ static void cmd_restore(struct dlist *kin,
 
 static void usage(void);
 void shut_down(int code) __attribute__ ((noreturn));
+void shut_down_via_signal(int code) __attribute__ ((noreturn));
 
 extern int saslserver(sasl_conn_t *conn, const char *mech,
                       const char *init_resp, const char *resp_prefix,
@@ -237,7 +238,7 @@ int service_init(int argc __attribute__((unused)),
     setproctitle_init(argc, argv, envp);
 
     /* set signal handlers */
-    signals_set_shutdown(&shut_down);
+    signals_set_shutdown(&shut_down_via_signal);
     signal(SIGPIPE, SIG_IGN);
 
     /* load the SASL plugins */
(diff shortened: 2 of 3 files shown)


ksmurchison (approved):
Looks good to me

PR #3421, opened 2021-03-20
PROPFIND <c:schedule-calendar-transp>

dilyanpalauzov:
The value of the schedule-calendar-transp property is not a string with possible text values “transparent” or “opaque”, but an XML element in the CalDAV namespace: https://tools.ietf.org/html/rfc6638#section-9.1.

- on PROPPATCH of the transarent property validate the namespace of the value


ksmurchison:
This fixes an obvious bug.

PR #3430, opened 2021-03-24
notify: don't log successful response from notifyd

robn:
There's a lot and they mean nothing!

### DIFF
--- a/imap/notify.c
+++ b/imap/notify.c
@@ -141,8 +141,10 @@ static void notify_dlist(const char *sockpath, const char *method,
     if (c == '\r') c = prot_getc(in);
     /* XXX - do something with the response?  Like have NOTIFY answer */
     if (c == '\n' && res && res->name) {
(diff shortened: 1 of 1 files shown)


imap/notify.c
```diff
@@ -141,8 +141,10 @@ static void notify_dlist(const char *sockpath, const char *method,
     if (c == '\r') c = prot_getc(in);
     /* XXX - do something with the response?  Like have NOTIFY answer */
     if (c == '\n' && res && res->name) {
-        syslog(LOG_NOTICE, "NOTIFY(%s): response %s to method %s",
-               loginfo, res->name, method);
+        if (!strcmp(res->name, "OK")) {
```
brong, line 144: I think you want the other NOT...

brong (changes requested):

brong (approved):
Looks great

PR #3431, opened 2021-03-24
httpd.c:error_response() remove tautology

dilyanpalauzov:


ksmurchison:
This check SHOULD NOT be removed as some parts of the code actually set a more specific error message before error_response() is called (e.g. in starttls()) which we don't want to overwrite.

dilyanpalauzov:
This is the current code:
```c
if (!txn->error.desc) {
    switch (code) {
        case HTTP_SERVER_ERROR:
            if (!txn->error.desc)
                txn->error.desc = "The server encountered an internal error.";
    }
}
```
The switch is not executed, if txn->error.desc is set.

ksmurchison:
Yes, you're correct.

PR #3432, opened 2021-03-25
doc/installing: prefer Let’s Encrypt over self-made certificate

dilyanpalauzov:
Delete the text describing how to set up a self-signed certificate.  Nowadays everybody setting up any form of server shall understand PKI and the terms private key and full chain.


elliefm (approved):

elliefm:
Nice, thanks

PR #3434, opened 2021-03-26
index: handle STORE flags acls per rfc 4314

### LINKED ISSUE #1375: Review RFC 4314 "STORE flags" compliance with no "w" right on folder
From: [Jeroen van Meeuwen (Kolab Systems)](vanmeeuwen@kolabsys.com)
Bugzilla-Id: 3488
Version: 2.5.x (next)
Owner: [Bron Gondwana](brong@fastmail.fm)

### LINKED ISSUE #1723: Compliance problem with RFC-4314
From: [Benoit Tellier](btellier@linagora.com)
Bugzilla-Id: 3895
Version: 2.4.17
Owner: [Bron Gondwana](brong@fastmail.fm)

elliefm:
We have two ancient existing issues open for this so maybe we should just fix it finally...

RFC 4314 says:

> STORE operation SHOULD NOT fail if the user has rights to modify
> at least one flag specified in the STORE, as the tagged NO
> response to a STORE command is not handled very well by deployed
> clients

This changes the rights check in index_store as follows:

* if we are trying to change a flag, but don't have the correct permission for it, remove that flag from the set to change
* if, after removing disallowed flags from the set to change, the set is now empty, then fail with permission denied

These behaviours only apply for "STORE +flags" and "STORE -flags" because these are the only ones it made sense for, though the RFC isn't clear that it should only apply to these.

If the command used is "STORE flags", the old behaviour is preserved: reject the whole change if any of the flags are forbidden.

It looks like STORE is also used for annotations.  I don't think I've changed anything wrt those, but please review closely.

New cassandane test here: https://github.com/cyrusimap/cassandane/compare/master...elliefm:v35/1723-store-flags-acl

Fixes #1723
Fixes #1375

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -1261,13 +1261,60 @@ EXPORTED int index_store(struct index_state *state, char *sequence,
     struct index_modified_flags modified_flags;
     struct index_record record;
 
-    /* First pass at checking permission */
-    if ((storeargs->seen && !(state->myrights & ACL_SETSEEN)) ||
-        ((storeargs->system_flags & FLAG_DELETED) &&
-         !(state->myrights & ACL_DELETEMSG)) ||
-        (((storeargs->system_flags & ~FLAG_DELETED) || flags->count) &&
-         !(state->myrights & ACL_WRITE))) {
-        return IMAP_PERMISSION_DENIED;
+    if (storeargs->operation == STORE_ADD_FLAGS
+        || storeargs->operation == STORE_REMOVE_FLAGS)
+    {
+        int did_limit_flags = 0;
+
+        /* Quoth RFC 4314:
+        *  STORE operation SHOULD NOT fail if the user has rights to modify
+        *  at least one flag specified in the STORE, as the tagged NO
+        *  response to a STORE command is not handled very well by deployed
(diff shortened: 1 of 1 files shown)


ksmurchison (approved):
Looks good to me

elliefm:
> If the command used is "STORE flags", the old behaviour is preserved: reject the whole change if any of the flags are forbidden.

I'm reading this again, and, is there another bug hiding in here?

If a message has, say, the \Deleted flag (somehow), but the user does not currently have the appropriate right to set or unset that flag, and they issue a `STORE flags ...` for some set of flags that they _do_ have permission to modify, are they thus able to implicitly remove the \Deleted flag even though they do not have permission to do so?  And, do we care?

I haven't changed the behaviour of "STORE flags" (with no + or -) so if this bug currently exists then I think it already existed...

elliefm:
> are they thus able to implicitly remove the \Deleted flag even though they do not have permission to do so

Wrote another test, and no, they cannot implicitly remove a flag they're not allowed to edit by setting a new set of flags.  Cool.  And now we have a regression test for the future too :)

PR #3438, opened 2021-03-27
Makefile.am: use TAB in continuation lines

dilyanpalauzov:
for consistency.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -69,19 +69,19 @@ AM_LDFLAGS = $(COV_LDFLAGS) $(ICU_LIBS)
 # have distcheck try to build with all optional components enabled, to aid
 # detection of missing files for these components
 AM_DISTCHECK_CONFIGURE_FLAGS = \
-    --enable-http \
-    --enable-calalarmd \
-    --enable-replication \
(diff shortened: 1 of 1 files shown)


elliefm:
This file is actually kind of a nuisance, cause it's different from everything else, where spaces are used for indentation.

Makefile.am doesn't require tabs (unlike a regular Makefile), _except_ in sections that are excerpts of raw `make` syntax.  So this file is always going to have hard tabs in those sections, but spaces are appropriate everywhere else.  It looks like a lot (but not all) of the automake sections of this file are indented with tabs, probably depending on who wrote it and their editor settings at the time.  It's not safe to run tools/remove-tabs.pl on this file, and so I guess this file was missed when we did the big switch to four space indents.

I think my personal preference would be to use four space indents for this file, just like everything else, and only use hard tabs where they're strictly required by `make` syntax.

dilyanpalauzov:
The indentation is currently inconsistent.

elliefm:
I've created #5038 to do the opposite: replacing tabs with 4-spaces to match our style, except where tabs are required by make syntax

PR #3442, opened 2021-03-29
jmap_contact.c: Switch to using Apple-style labels on vCard ADR/TEL/E…

ksmurchison:
…MAIL properties (via property grouping)


brong (approved):
Looks great.  Looks like it's easy enough to add back LABEL as a param as well if we want, and parse either.

PR #3446, opened 2021-03-31
libcyrus_imap: mark some functions as HIDDEN

dilyanpalauzov:
Remove unused xapian_dbw_cancel_txn() and xapian_doc_reset().

### DIFF
--- a/imap/annotate.c
+++ b/imap/annotate.c
@@ -416,7 +416,7 @@ EXPORTED void setentryatt(struct entryattlist **l, const char *entry,
     }
 }
 
-EXPORTED char *dumpentryatt(const struct entryattlist *l)
+HIDDEN char *dumpentryatt(const struct entryattlist *l)
 {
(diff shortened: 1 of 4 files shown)


elliefm:
I think this needs to be thought about holistically with the other libcyrus_imap problems (e.g. #4798 etc).  It's hard to reason about the ramifications of removing symbol visibility from the library just from looking at a diff, but we don't yet have tests that would prove that the exposed interfaces still work after this change, nor even a clear idea of which things we want to expose anyway.

dilyanpalauzov:
The criterion for hiding a symbol is, that the symbol (function) is not used outside the library.  This makes the list of exported symbol shorter ⇒ speeds up the dynamic linking; and allows some optimizations, like inlining everwhere the function, thus eliminating it, as it cannot be called from outside.

The verification, as for all other HIDDEN and static functions, is to look from where the function is called.

PR #3448, opened 2021-04-06
Jmap shared upload collection

ksmurchison:
Tests added in https://github.com/cyrusimap/cassandane/pull/112

### DIFF
--- a/imap/http_jmap.c
+++ b/imap/http_jmap.c
@@ -625,7 +625,7 @@ static int jmap_download(struct transaction_t *txn)
     return res;
 }
 
-static int sharedrights_cb(const mbentry_t *mbentry, void *vrock)
+static int has_shared_rw_rights_cb(const mbentry_t *mbentry, void *vrock)
 {
     int *rights = (int *) vrock;
 
@@ -650,7 +650,7 @@ static int has_shared_rw_rights(const char *accountid)
 {
     int rights = 0;
 
-    mboxlist_usermboxtree(accountid, NULL, &sharedrights_cb, &rights, 0);
+    mboxlist_usermboxtree(accountid, NULL, &has_shared_rw_rights_cb, &rights, 0);
 
     return rights;
 }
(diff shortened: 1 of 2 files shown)


imap/http_jmap.c
```diff
@@ -625,6 +625,36 @@ static int jmap_download(struct transaction_t *txn)
     return res;
 }
 
+static int sharedrights_cb(const mbentry_t *mbentry, void *vrock)
```
rsto, line 628: Nit: Rename this to has_shared_rw_rights_cb ? This function really just that boolean question, and it's less confusing to than having the same named function to different things in two files.

imap/jmap_api.c
```diff
@@ -3093,6 +3195,14 @@ HIDDEN int jmap_set_sharewith(struct mailbox *mbox,
         }
     }
 
+    if (!r) {
+        /* Set proper access rights on JMAP upload folder */
+        /* XXX This is currently done by brute force.
```
rsto, line 3200: This comment might be better placed at the start of set_upload_rights. Without knowing the internals of that function, it is not clear where the "brute force" is coming from here.

rsto (changes requested):
The logic looks sane to me. Found two nits that I think would make the code clearer to read.

elliefm:
Does this need to be backported to 3.4?  The Cassandane test now checks for 3.4 or greater, but 3.4 doesn't currently have this code, so the test runs and fails.  Not sure whether the test is correct and the changes need to be backported, or if they don't and the test's version check is wrong.

It's not clear to me what the big picture of this change is or how it fits into other JMAP stuff that's happened on master (and not 3.4)

ksmurchison:
This should probably be backported if we support sharing of JMAP resources in 3.4, which I assume we do.

elliefm:
The calendar_set_sharewith test previously worked before the specifics for this PR were added to it, so I'm concluding from that that it does support sharing, and therefore this needs to be backported.  Cool, thanks for the clarification, I'll get it done. :)

elliefm:
This is now on 3.4

PR #3451, opened 2021-04-12
grow dynamic arrays by doubling, rather than fixed size

### LINKED ISSUE #3439: make dynamic arrays grow harder
Our dynamic array types all* grow by only 16 bytes at a time for some reason:

```
$ack QUANTUM
lib/arrayu64.c
76:#define QUANTUM     16
81:    newalloc = ((newalloc + QUANTUM-1) / QUANTUM) * QUANTUM;

lib/bitvector.c
63:#define QUANTUM         (256)
79:        unsigned int newalloc = ((len + QUANTUM-1) / QUANTUM) * QUANTUM;

lib/bufarray.c
86:#define QUANTUM     16
91:    newalloc = ((newalloc + QUANTUM) / QUANTUM) * QUANTUM;

lib/ptrarray.c
81:#define QUANTUM     16
88:    newalloc = ((newalloc + QUANTUM-1) / QUANTUM) * QUANTUM;

lib/dynarray.c
86:#define QUANTUM     16
94:    newalloc = ((newalloc + QUANTUM-1) / QUANTUM) * QUANTUM;

lib/strarray.c
85:#define QUANTUM     16
90:    newalloc = ((newalloc + QUANTUM) / QUANTUM) * QUANTUM;
```

This results in noticeable memcpy overhead from repeated resizing in hot loops.  It's much more normal to have dynamic arrays grow by a factor of 1.5-2 each time they need to grow, to reduce the amount of times they need to grow.

elliefm:
This changes our dynamic array types to double in size each time they need to grow, rather than always growing by 16 elements at a time.  This potentially costs a little more memory usage, but reduces reallocation overheads.  The first allocation is still always 16 elements -- we're not wasting time stepping up through 1, 2, 4, 8 first!

Specifically, these modules have been changed to double in size:

- lib/arrayu64.c
- lib/bufarray.c
- lib/dynarray.c
- lib/ptrarray.c
- lib/strarray.c

I have left lib/bitvector.c alone for now, even though its growth is similar (256 bits at a time), because it's more complicated and I'm not sure it's useful to change -- though let me know if you disagree!

I've added tests for arrayu64 and bufarray, which previously didn't have any, and dynarray, which had very low coverage, and then fixed some bugs that the new tests exposed.

There's a lot of API variation between these modules, especially wrt using `int` vs `size_t`, whether `fooarray_nth()` accepts a perl-style negative index, whether or not there's always a NULL element at the end, etc, even though they're superficially similar.  It would be nice to consolidate them sometime, but that feels like a rabbit hole and I'm leaving it alone for now.

Reviewers: please build this and also run the tests.  Thanks :)

Closes #3439

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -617,6 +617,7 @@ cunit_TESTS = \
 	cunit/binhex.testc \
 	cunit/bitvector.testc \
 	cunit/buf.testc \
+	cunit/bufarray.testc \
 	cunit/byteorder.testc \
 	cunit/charset.testc \
 	cunit/command.testc \
--- a/Makefile.am
+++ b/Makefile.am
@@ -612,6 +612,7 @@ cunit_FRAMEWORK = \
 cunit_TESTS = \
 	cunit/aaa-db.testc \
 	cunit/annotate.testc \
+	cunit/arrayu64.testc \
 	cunit/backend.testc \
 	cunit/binhex.testc \
 	cunit/bitvector.testc \
--- a/cunit/arrayu64.testc
+++ b/cunit/arrayu64.testc
@@ -1,0 +1,747 @@
+#include "cunit/cyrunit.h"
+#include "xmalloc.h"
+#include "bsearch.h"
+#include "arrayu64.h"
+
+static void test_fini_null(void)
+{
+    /* _fini(NULL) is harmless */
+    arrayu64_fini(NULL);
+    /* _free(NULL) is harmless */
+    arrayu64_free(NULL);
+}
+
+static void test_auto(void)
+{
+    arrayu64_t a = ARRAYU64_INITIALIZER;
+    uint64_t u1;
+    uint64_t u2;
+
+    CU_ASSERT_EQUAL(arrayu64_size(&a), 0);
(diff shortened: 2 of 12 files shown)


rsto (approved):
This looks great. Also many thanks for improving test coverage of dynarray!

rsto:
I ran this locally and tests passed. I also valgrinded the dynarray code on the jmap-calendars-01 branch, which actually makes use of that data structure.

brong (approved):
Looks great.  My only nit would be "same code duplicated in lots of places" but that's not a regression.

elliefm:
> Looks great. My only nit would be "same code duplicated in lots of places" but that's not a regression.

Yeah agreed.  If this were C++ these should all be templates (but if this were C++ then the STL would already have us covered).

Templating this stuff in C99+ is possible with clever use of macros (I've done it before, it's even kinda fun), but it's a huge refactor and quite hard to debug, so it kinda needs to be perfect out the gate...

Even without templating it, we could probably genericise the core (like the (`...ensure...()` and `adjust_index...()` functions in particular), but it would require all these slightly different APIs to converge, and even that's quite a refactor...

PR #3453, opened 2021-04-13
Fix JSCalendar default values and use of Participant.email property

rsto:
This fixes two issues with the current JSCalendar implementation that shouldn't need to wait until we land the jmap-calendars-01 branch:

- Previous draft versions defined empty string default values for name and description properties, but this changed to optional properties in the current spec. We should empty values both when reading and writing iCalendar data.

- The current implementation conflates the Participant.email and Participant.sendTo{imip} properties, but they really a separate things. We shouldn't use Participant.email for scheduling and this patch fixes that. We still use Participant.email to validate the custom participantId property, though

Tested in https://github.com/cyrusimap/cassandane/tree/jmap_ical_fixes

### DIFF
--- a/imap/jmap_calendar.c
+++ b/imap/jmap_calendar.c
@@ -1926,6 +1926,31 @@ done:
     return r;
 }
 
+static const char *find_participant_by_addr(json_t *participants, const char *addr)
+{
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):
Looks good to me

PR #3456, opened 2021-04-18
Server annotation replication

brong:
We ran into a production bug where a client was setting a per-server annotation to track some config for itself (a JSON blob).  It turns out that replication died on that.  Ouch.

This fixes this issue, and a related permissions issue found by the associated cassandane tests in https://github.com/cyrusimap/cassandane/pull/114


ksmurchison (approved):
Looks good to me

PR #3458, opened 2021-04-19
jmap_contact_from_vcard: be safe against X-ABLabel with no group

brong:


ksmurchison:
@brong After looking at this again, I wonder if we should just be ignoring X-ABLabel if it doesn't have a group because it can't be assigned to a property anyways.

ksmurchison:
https://github.com/cyrusimap/cyrus-imapd/pull/3459 has a simpler, alternate fix (no need to track labels unassociated with another property).

Tested here:

rsto (approved):
Looks good to me

ksmurchison:
Closing due to #3459 being merged

PR #3459, opened 2021-04-19
JMAP: remove Apple-style labels when removing ADR/EMAIL/TEL properties

ksmurchison:


imap/jmap_contact.c
```diff
@@ -1489,7 +1489,7 @@ static json_t *jmap_contact_from_vcard(struct vparse_card *card,
     hash_table labels = HASH_TABLE_INITIALIZER;
     construct_hash_table(&labels, 10, 0);
     for (entry = card->properties; entry; entry = entry->next) {
-        if (!strcasecmp(entry->name, "X-ABLabel")) {
+        if (!strcasecmp(entry->name, APPLE_LABEL_PROPERTY)) {
```
rsto, line 1492: Is entry->name guaranteed to be non-NULL? Other calls to `strcasecmpsafe` later in this commit suggest it isn't.

lib/vparse.c
```diff
@@ -1032,7 +1033,9 @@ EXPORTED void vparse_replace_entry(struct vparse_card *card, const char *group,
     }
 }
 
-EXPORTED void vparse_delete_entries(struct vparse_card *card, const char *group, const char *name)
+static void _delete_entries(struct vparse_card *card,
```
rsto, line 1036: I wonder if the label counting code in here should be moved to its own function `count_apple_labels`. Saving one iteration of the cards properties doesn't seem to warrant the mental overhead that `_delete_entries` really does not just delete entries anymore.

lib/vparse.c
```diff
@@ -1042,10 +1045,55 @@ EXPORTED void vparse_delete_entries(struct vparse_card *card, const char *group,
...
+                }
+                hash_insert(entry->group, (void *) count, props_using_label_counts);
+            }
+
+            entryp = &((*entryp)->next);
+        }
+    }
+}
+
+EXPORTED void vparse_delete_entries(struct vparse_card *card,
```
rsto, line 1066: Would there ever be a reason to only call `vparse_delete_entries` instead of deleting entries and labels?

lib/vparse.h
```diff
@@ -5,6 +5,8 @@
 #include "util.h"
 #include "strarray.h"
 
+#define APPLE_LABEL_PROPERTY "X-ABLabel"
```
rsto, line 8: I'd prefix that name with either `VCARD_` or `VPARSE_`.

rsto (changes requested):

lib/vparse.c
ksmurchison: Probably not.  I will fold this into one function

lib/vparse.h
ksmurchison: Done

imap/jmap_contact.c
ksmurchison: I don't think any name will ever be NULL, but I'll use safe just to be safe

lib/vparse.c
ksmurchison: This is going to be rolled into one function per the comment below

ksmurchison:
Tested with https://github.com/cyrusimap/cassandane/pull/115

rsto (approved):
Looks good to me

PR #3460, opened 2021-04-20
jmap_mail: replace Email/query on-disk cache with in-memory

rsto:
Replaces the on-disk Email/query cache with an in-memory cache. This allows to speed up consecutive email queries that page through the overall result.

As the cache is stored in memory, this only benefits requests over web sockets or kept-alive HTTP connections.

Config changes:

Deprecates the `jmap_emailsearch_db_path` option.
Introduces the `jmap_querycache_max_age` option.

Tested in https://github.com/cyrusimap/cassandane/tree/jmap_emailquery_cache_inprocess

### DIFF
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -1121,10 +1121,19 @@ Blank lines and lines beginning with ``#'' are ignored.
    the same has to be done (cyr_dbtool) for each subscription database
    See improved_mboxlist_sort.html.*/
 
-{ "jmap_emailsearch_db_path", NULL, STRING, "3.1.6" }
+{ "jmap_emailsearch_db_path", NULL, STRING, "3.1.6", "UNRELEASED" }
 /* The absolute path to the JMAP email search cache file.  If not
    specified, JMAP Email/query and Email/queryChanges will not
    cache email search results. */
+
+{ "jmap_querycache_max_age", "0m", DURATION, "UNRELEASED" }
+/* The duration after which unused cached JMAP query results
+   must be evicted from process memory. If non-zero, then the
+   full result of the last query (before windowing) is stored
+   in-memory. Subsequent queries with the same expression and
+   query state can then page through the cached result.
+   A zero value disables query result caching.
+   This feature currently only is enabled for Email/query. */
 
 { "jmap_preview_annot", NULL, STRING, "3.1.1" }
 /* The name of the per-message annotation, if any, to store message
--- a/Makefile.am
+++ b/Makefile.am
@@ -646,7 +646,8 @@ cunit_TESTS = \
 	cunit/quota.testc \
 	cunit/rfc822tok.testc \
 	cunit/search_expr.testc \
-	cunit/seqset.testc
+	cunit/seqset.testc \
+	cunit/smallarrayu64.testc
 
 if SIEVE
 cunit_TESTS += cunit/sieve.testc
@@ -787,6 +788,7 @@ include_HEADERS = \
 	lib/retry.h \
 	lib/rfc822tok.h \
 	lib/signals.h \
+	lib/smallarrayu64.h \
 	lib/sqldb.h \
 	lib/strarray.h \
 	lib/strhash.h \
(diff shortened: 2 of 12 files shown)


ksmurchison (commented):
Looks sane to me, but I'm not a /query expert

lib/imapoptions
```diff
@@ -1115,11 +1115,20 @@ Blank lines and lines beginning with ``#'' are ignored.
    the same has to be done (cyr_dbtool) for each subscription database
    See improved_mboxlist_sort.html.*/
 
-{ "jmap_emailsearch_db_path", NULL, STRING, "3.1.6" }
+{ "jmap_emailsearch_db_path", NULL, STRING, "3.1.6", "3.5.0" }
```
elliefm, line 1118: Please use "UNRELEASED" rather than "3.5.0".  I'll fix it to whatever is correct at the time when this reaches a real release.

lib/imapoptions
```diff
@@ -1115,11 +1115,20 @@ Blank lines and lines beginning with ``#'' are ignored.
...
 /* The absolute path to the JMAP email search cache file.  If not
    specified, JMAP Email/query and Email/queryChanges will not
    cache email search results. */
 
+{ "jmap_querycache_max_age", "0m", DURATION, "3.5.0" }
```
elliefm, line 1123: Ditto re "UNRELEASED"

elliefm (commented):
(I've only skimmed this so far, haven't looked at the complicated bits, but the versions in imapoptions jumped out at me)

lib/imapoptions
rsto: Will do. I still hadn't grokked our versioning system.

lib/imapoptions
rsto: Ditto

imap/jmap_api.h
```diff
@@ -186,13 +186,31 @@ void jmap_getblob_ctx_fini(jmap_getblob_context_t *ctx);
...
     json_t *server_capabilities;
     long limits[JMAP_NUM_LIMITS];
+    // internal state
     ptrarray_t getblob_handlers; // array of jmap_getblob_handler
+    ptrarray_t handlers; // array of (malloced) jmap_handlers
 } jmap_settings_t;
 
+
+enum jmap_handler_event {
+    JMAP_HANDLE_SHUTDOWN      = (1 << 1), /* executed when httpd is shutdown. req is NULL */
```
ksmurchison, line 203: Not that it matters, but is there a reason why this doesn't start at (1 << 0) ?

imap/jmap_api.h
```diff
@@ -186,13 +186,31 @@ void jmap_getblob_ctx_fini(jmap_getblob_context_t *ctx);
...
+// executed before request-specific handlers
+typedef void jmap_req_preproc(jmap_req_t *req);
+
 typedef struct {
     hash_table methods;
     json_t *server_capabilities;
     long limits[JMAP_NUM_LIMITS];
+    // internal state
     ptrarray_t getblob_handlers; // array of jmap_getblob_handler
+    ptrarray_t handlers; // array of (malloced) jmap_handlers
```
ksmurchison, line 198: Should we call this something more specific like 'state_handlers' or 'context_handlers' ?

imap/jmap_api.h
rsto: No, that was just an oversight. Fixed.

rsto:
I also renamed `JMAP_HANDLER_RESET_CONN` to `JMAP_HANDLER_CLOSE_CONN`

brong:
I had to remove the include-in-fastmail again because it clashes with the uuid mailboxes branch :(

ksmurchison (approved):

rsto:
@brong this is ready for a new review.

Summary of main changes to previous version:

- Uses search result iterators to only traverse as many search hits as requested.
- Maps message part ids to an internal numeric representation for efficient storage.
- Introduces smallarrayu64 for to efficiently store numeric message part ids per match.
- Uses a linked list in contiguous memory for efficiently process findAllInThread requests.

Should we declare smallarrayu64_nth as extern inline?

PR #3466, opened 2021-04-23
more xsyslog conversions

elliefm:
As usual, please also build this, not just eyeball it.

This one might be on the edge of too big to review in one go.  Let me know if that's the case and I'll split it up a bit.

This one required Cassandane updates because a bunch of our replication tests want to check for specific things in syslog.  But I have already made those updates in a backward-compatible fashion and merged them to master, so as long as your Cassandane is up to date, it should still pass as usual.


rsto (approved):
I reviewed all of it and it looks good to me. The error messages now make more clear what went wrong. I built it on gcc 8.3.0 without issues.

PR #3467, opened 2021-04-23
Mboxevent jmap email v2

ksmurchison:
Redo of previous PR.

Adds the 'vnd.cmu.jmapEmail' selector to the 'event_extra_params' option
which will includes a JMAP Email object in the event message.

Also fixes header:asDate to return a Date type rather than a UTCDate type.

Fix for existing Cass test is here: https://github.com/cyrusimap/cassandane/pull/116

### DIFF
--- a/imap/jmap_util.h
+++ b/imap/jmap_util.h
@@ -168,4 +168,13 @@ extern json_t *jmap_header_as_addresses(const char *raw, enum header_form form);
 extern json_t *jmap_emailaddresses_from_addr(struct address *addr,
                                              enum header_form form);
 
+#define JMAP_BLOBID_SIZE 42
+extern void jmap_set_blobid(const struct message_guid *guid, char *buf);
+
+#define JMAP_EMAILID_SIZE 26
+extern void jmap_set_emailid(const struct message_guid *guid, char *buf);
+
+#define JMAP_THREADID_SIZE 18
+extern void jmap_set_threadid(conversation_id_t cid, char *buf);
+
 #endif /* JMAP_UTIL_H */
--- a/imap/jmap_util.h
+++ b/imap/jmap_util.h
@@ -159,12 +159,13 @@ enum header_form {
     HEADER_FORM_GROUPEDADDRESSES = 1 << 6
 };
 
-extern json_t *jmap_header_as_raw(const char *raw, enum header_form form);
-extern json_t *jmap_header_as_text(const char *raw, enum header_form form);
-extern json_t *jmap_header_as_date(const char *raw, enum header_form form);
-extern json_t *jmap_header_as_urls(const char *raw, enum header_form form);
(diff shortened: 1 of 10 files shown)


imap/jmap_util.h
```diff
@@ -147,4 +148,24 @@ extern int jmap_decode_rawdata_blobid(const char *blobid,
...
+    HEADER_FORM_RAW              = 1 << 0,
+    HEADER_FORM_TEXT             = 1 << 1,
+    HEADER_FORM_DATE             = 1 << 2,
+    HEADER_FORM_URLS             = 1 << 3,
+    HEADER_FORM_MESSAGEIDS       = 1 << 4,
+    HEADER_FORM_ADDRESSES        = 1 << 5,
+    HEADER_FORM_GROUPEDADDRESSES = 1 << 6
+};
+
+extern json_t *jmap_header_as_raw(const char *raw, enum header_form form);
```
rsto, line 162: Now that we export these `as_xxx` functions, I think we should remove the `form` argument. All of these functions but the `as_addresses` and `from_addr` functions ignore them, and the reason they all accept an unused argument is because we internally use a hash table of converter functions in `jmap_mail.c`. I wonder if we can just split the `as_addresses` in two functions `as_addresses` and `as_groupaddresses` and do away with that `form` argument.

imap/mboxevent.c
```diff
@@ -947,28 +963,15 @@ EXPORTED void mboxevent_extract_record(struct mboxevent *event, struct mailbox *
...
-            threadid[2] = 'L';
-            threadid[3] = '\0';
-        }
-        else {
-            threadid[0] = 'T';
-            memcpy(threadid+1, conversation_id_encode(record->cid), 16);
-            threadid[17] = '\0';
-        }
-        FILL_STRING_PARAM(event, EVENT_MESSAGE_THREADID, xstrdup(threadid));
+        FILL_STRING_PARAM(event, EVENT_MESSAGE_THREADID,
```
rsto, line 1027: Is it OK to remove the conditional that sets the `NIL` value here? I guess `set_threadid` sets a all-zeroes id if cid is zero.

imap/mboxevent.c
```diff
@@ -1129,19 +1130,7 @@ EXPORTED void mboxevent_extract_msgrecord(struct mboxevent *event, msgrecord_t *
...
-            threadid[2] = 'L';
-            threadid[3] = '\0';
-        }
-        else {
-            threadid[0] = 'T';
-            memcpy(threadid+1, conversation_id_encode(cid), 16);
-            threadid[17] = '\0';
-        }
-        FILL_STRING_PARAM(event, EVENT_MESSAGE_THREADID, xstrdup(threadid));
+        FILL_STRING_PARAM(event, EVENT_MESSAGE_THREADID, xstrdup(threadid(cid)));
```
rsto, line 1203: Ditto for `NIL`

rsto (approved):

PR #3468, opened 2021-04-24
ctl_cyrusdb -r assertion on startup when mboxlist_db = skiplist

felixjogris:
Hi,

ctl_cyrusdb exits during startup:

% /usr/local/cyrus/sbin/ctl_cyrusdb -r
fatal error: Internal error: assertion failed: lib/cyrusdb_skiplist.c: 1469: db && tid
% /usr/local/cyrus/sbin/cyr_info version
cyrus-imapd 3.4.0
% uname -ro
FreeBSD 13.0-RELEASE

I believe this is due to mboxlist_set_racls() and mboxlist_set_runiqueid() may not have done any changes to the mailbox database, e.g. when reverse acls are enabled and have the latest version. In 3.2.x, tid is never NULL set as cyrusdb_fetch() is always called.

BR Felix

### DIFF
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -3312,6 +3312,7 @@ EXPORTED int mboxlist_set_racls(int enabled)
 {
     struct txn *tid = NULL;
     int r = 0;
+    int modified_mbdb = 0;
 
     init_internal();
 
@@ -3320,6 +3321,7 @@ EXPORTED int mboxlist_set_racls(int enabled)
         /* remove */
         r = cyrusdb_foreach(mbdb, "$RACL", 5, NULL, del_cb, &tid, &tid);
         if (!r) have_racl = 0;
+        modified_mbdb = 1;
     }
     if (enabled && !have_racl) {
         /* add */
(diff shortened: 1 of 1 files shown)


elliefm:
This doesn't feel quite right; for one thing, the `cyrusdb_abort(mbdb, tid)` is also going to assert out if it gets in there with NULL tid.

I think an early exit like `if (!tid) return 0;` before the abort/commit section might be better, but I would feel more comfortable with something that was more explicitly "nothing changed" rather than just "somehow we didn't get a tid".

I guess this doesn't bite use at fm because we use twoskip, not skiplist, and it looks at a glance like twoskip's foreach implementation always creates a transaction...

felixjogris:
Agreed. Just checking for tid != NULL depends on cyrusdb_foreach() to set tid. Followed your advise to return r if we don't get a tid or haven't touched mbdb. I'll rebase if ok.

elliefm:
This looks good, please rebase and squash it down to one commit.  Thanks :)

I've got a regression test for this, which proves the crash and the fix, which I'll merge when I merge this one: https://github.com/cyrusimap/cassandane/compare/master...elliefm:v35/3468-ctlcyrusdb-crash

felixjogris:
Done. Thanks.

elliefm (approved):

PR #3471, opened 2021-04-26
WS: use the actual txn for the stream/connection rather than creating a new one

ksmurchison:
This removes an ugly hack where each WS connection would create its own new txn, separating itself from the txn used to bootstrap the WS connection.  The warts that forced this hack have been removed in previous commits on master.

Based on https://github.com/cyrusimap/cyrus-imapd/pull/3477


rsto (approved):
Looks good to me. Builds and tests cleanly on the JMAPEmail test suite.

elliefm (approved):

PR #3473, opened 2021-04-26
Implement JMAP/WS push

ksmurchison:
This branch sits on top of the ws_timeout branch

### DIFF
--- a/imap/http_jmap.c
+++ b/imap/http_jmap.c
@@ -1233,8 +1233,12 @@ static struct prot_waitevent *jmap_ws_pusher(struct protstream *s __attribute__(
     struct mboxname_counters cur_counters;
     struct buf buf = BUF_INITIALIZER;
 
+    xsyslog(LOG_DEBUG, "JMAP WS push", "accountid=<%s>", jpush->accountid);
+
     if (mboxname_read_counters(jpush->inboxname, &cur_counters)) {
         /* Something went wrong - don't reschedule */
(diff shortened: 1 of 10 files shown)


rsto (commented):
I can't comment much on the websocket implementation. Is there any chance to have automated tests for this? I built and ran Cassandane locally, but it looks as if our test suites don't hit any of these code paths.

imap/http_jmap.c
```diff
@@ -1170,45 +1177,201 @@ static int jmap_get_session(struct transaction_t *txn)
...
+        else if (mboxname_read_counters(jpush->inboxname, &cur_counters)) {
+            /* Something went wrong */
+            goto disable;
+        }
+
+        /* Initialize our tracking modseqs to the maximum value */
+        const struct datatype_t *dtype;
+        for (dtype = dataTypes; dtype->name; dtype++) {
+            modseq_t *modseq =
+                (modseq_t *)((size_t) &jpush->counters + dtype->offset);
```
elliefm, line 1300: nit: I think the `(size_t)` cast here (and in the other places that this pattern occurs) should be `(uintptr_t)`, since it looks like the intention is "make sure the addition result is the right type to then be cast to a pointer"?

I think in practice `size_t` and `uintptr_t` are usually the same, but if we're bothering to cast the intermediate result (to shut up some warning, I assume), we might as well cast it to the right thing...

PR #3474, opened 2021-04-26
jmap_api.c Skip special use folders when finding JMAP accounts

ksmurchison:
We shouldn't look at #jmap, #notifications, etc when determining shared accounts.

Tested with https://github.com/cyrusimap/cassandane/pull/117


rsto (approved):
Looks good to me. Also tested it with the Cassandane JMAP suites.

PR #3475, opened 2021-04-27
jmap_api: add event handlers to JMAP settings

rsto:
Adds handlers to handle various stages of the JMAP request processing pipeline. This is an extract from https://github.com/cyrusimap/cyrus-imapd/pull/3460 which we want to land earlier than the email-specific changes.

### DIFF
--- a/imap/jmap_api.h
+++ b/imap/jmap_api.h
@@ -190,8 +190,23 @@ typedef struct {
     hash_table methods;
     json_t *server_capabilities;
     long limits[JMAP_NUM_LIMITS];
+    // internal state
     ptrarray_t getblob_handlers; // array of jmap_getblob_handler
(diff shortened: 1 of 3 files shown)


imap/jmap_api.h
```diff
@@ -186,13 +186,31 @@ void jmap_getblob_ctx_fini(jmap_getblob_context_t *ctx);
 
 typedef int jmap_getblob_handler(jmap_req_t *req, jmap_getblob_context_t *ctx);
 
+// executed before request-specific handlers
+typedef void jmap_req_preproc(jmap_req_t *req);
```
ksmurchison, line 190: Is this still needed?

rsto: right, it isn't

ksmurchison (approved):

PR #3476, opened 2021-04-28
misc: don't use __GNUC_PREREQ() when checking for GCC features

### LINKED ISSUE #3462: __GNUC_PREREQ is not portable
I'm just trying to upgrade cyrus imapd to 3.4.0 on an OmniOS system and am getting build errors due to `__GNUC_PREREQ` not being defined. From what I can tell, this is a macro provided by `features.h` on systems which use glibc, and therefore not portable.

This is the patch I'm using temporarily to get around this:

```diff
-#  if defined(__GNUC__) && __GNUC_PREREQ(4, 3)
+#  if defined(__GNUC__) && ((__GNU__ == 4 && __GNUC_MINOR >= 3) || __GNU__ > 4)
```

elliefm:
This macro is provided by glibc, not gcc, so using it winds up implicitly checking for "gcc+glibc" rather than just "gcc".

Fixes #3462

Need this for 3.4... I would just push it directly to both branches, but I wanted to submit a PR to make sure the DCO exception for org members is working correctly

### DIFF
--- a/lib/byteorder.h
+++ b/lib/byteorder.h
@@ -61,7 +61,7 @@
 #  define ntohll(x) (x)
 
 #else /* small-endian machines */
-#  if defined(__GNUC__) && __GNUC_PREREQ(4, 3)
+#  if defined(__GNUC__) && ((__GNU__ == 4 && __GNUC_MINOR >= 3) || __GNU__ > 4)
      /* Remove existing macros if present */
 #    undef ntohl
 #    undef htonl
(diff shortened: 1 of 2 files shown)


elliefm:
Please do not "Set DCO to pass" on this one, I want to make sure we're still able to merge PR's even if the check failed.

rsto (approved):
Looks good to me. Builds and unit-tests cleanly on my setup (gcc 8.3.0)

ksmurchison (approved):
Also works for me

PR #3477, opened 2021-04-28
Txn/http_conn cleanup callbacks

ksmurchison:
Use an array of auxiliary callbacks to cleanup contexts within txn and http_conn.  This allows us to isolate more code from httpd.c and also will allow cleanup to be run when httpd wouldn't know how to the cleanup the context.


elliefm (approved):
Oh yeah I like this.  Builds and tests fine here (though I don't think we have tests that exercise this stuff)

PR #3478, opened 2021-04-29
(3.4) use Date rather than UTCDate

elliefm:
This updates 3.4 to pass the JMAPTestSuite.t:Email:get:header-header-field-name test, which was recently updated to expect Date objects rather than UTCDate objects.

I tried simply cherry-picking the relevant commit from master (e354baa53abf3a714f1c34b68a68e4b8d1b68d7b), but there is a whole rabbit trail of other commits it's dependent on, so instead of spending the day tracking them down, I just rewrote the same changes to the 3.4 code instead.

But, since I re-wrote it rather than just cherry-picking it, I'd like some eyes on it before I land it please. :)

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -497,18 +497,18 @@ static json_t *_header_as_date(const char *raw, enum header_form form __attribut
 {
     if (!raw) return json_null();
 
-    time_t t;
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):
This looks correct to me

PR #3480, opened 2021-04-29
Reconstruct repair mboxlist

ksmurchison:
@brong This is a (maybe too?) simple fix but it seems to work in my testing.  The only possible downside is that the modseqs for both records get bumped when inserted.

This PR is against one of my intermediate branches, and the only commits worth looking at are the last two.

The penultimate commit is to reconstruct to insert missing I records.
The last commit should fix the issue of deleting an I record for a renamed mailbox.  I was unable to get cyr_expire to attempt to do the deletion, so this fix is untested.

### DIFF
--- a/imap/jmap_backup.c
+++ b/imap/jmap_backup.c
@@ -269,7 +269,7 @@ static json_t *jmap_restore_reply(struct jmap_restore *restore)
 struct restore_rock {
     jmap_req_t *req;
     struct jmap_restore *jrestore;
-    int mbtype;
+    uint32_t mbtype;
     modseq_t deletedmodseq;
     char *(*resource_name_cb)(message_t *, void *);
     int (*restore_cb)(message_t *, message_t *, jmap_req_t *, void *, int);
@@ -348,7 +348,7 @@ static int restore_collection_cb(const mbentry_t *mbentry, void *rock)
     syslog(log_level, "restore_collection_cb: processing '%s'  (type = 0x%03x)",
            mbentry->name, mbentry->mbtype);
 
-    if ((mbentry->mbtype & rrock->mbtype) != rrock->mbtype) {
+    if (mbtype_isa(mbentry->mbtype) != rrock->mbtype) {
         syslog(log_level, "skipping '%s': not type 0x%03x",
                mbentry->name, rrock->mbtype);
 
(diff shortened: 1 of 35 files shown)


imap/jmap_backup.c
```diff
@@ -348,7 +348,7 @@ static int restore_collection_cb(const mbentry_t *mbentry, void *rock)
     syslog(log_level, "restore_collection_cb: processing '%s'  (type = 0x%03x)",
            mbentry->name, mbentry->mbtype);
 
-    if ((mbentry->mbtype & rrock->mbtype) != rrock->mbtype) {
+    if (mbtype_isa(mbentry->mbtype) != rrock->mbtype) {
```
brong, line 351: does this always have a type, or was "0" equal to "all the types" before?

imap/reconstruct.c
```diff
@@ -454,6 +459,19 @@ static int do_reconstruct(struct findall_data *data, void *rock)
 
     name = mbname_intname(data->mbname);
 
+    if (reconstruct_flags & RECONSTRUCT_REPAIR_MBOXLIST) {
```
brong, line 462: I think we should just always do this before loading the mailbox in the regular reconstruct path, rather than adding another flag for the operator to have to find:

`reconstruct -u $user` should fix everything about that user, and likewise `reconstruct $mailbox` should fix everything about that mailbox - we shouldn't need flags unless there's a choice to make or a serious performance issue with running the default, and I don't think this qualifies.

imap/reconstruct.c
```diff
@@ -454,6 +459,19 @@ static int do_reconstruct(struct findall_data *data, void *rock)
...
+        /* Skip tombstones */
+        if (!(data->mbentry->mbtype & MBTYPE_DELETED)) {
+            mbentry_t *mbentry = NULL;
+            r = mboxlist_lookup_by_uniqueid(data->mbentry->uniqueid, &mbentry, NULL);
+            if (r == IMAP_MAILBOX_NONEXISTENT) {
+                mboxlist_update((struct mboxlist_entry *) data->mbentry, 1);
+            }
```
brong, line 469: We should also check that the entry matches, and repair if it doesn't.  Same "fix up the things" argument.

brong (changes requested):
I'd like reconstruct to be more careful about finding all the mismatches than this.

(we may also need a mode which goes through the 'I' records and looks for matching 'N' records at some later date, because that's something that will need its own mode)

brong (approved):
Yep, this looks good.

ksmurchison:
Cherry-picked the required commits onto the current new-mboxlist-recs-bis branch and rebased up to usermeta-by-id-bis

PR #3483, opened 2021-05-05
JMAP Contacts: fix importance property

rsto:
Fixes two bugs with the importance property:

- if no private annotation for importance is set, fall back to reading the shared annotation
- extend the precision of the importance value to DBL_DECIMAL_DIG

Number precision is tested in https://github.com/cyrusimap/cyrus-imapd/tree/jmap_contact_importance_fixes


brong (approved):
Looks great, thanks.

elliefm:
I've pulled the Cassandane branch in, so the importance tests will now pass again

PR #3484, opened 2021-05-05
WIP: Usermeta by id bis

brong:
Creating an MR with this branch, then updating the build process to use this but still base on master - this means that we can add changes to master and we don't need to rebase this branch to build uuidbox builds.


brong:
And now it merges!

PR #3487, opened 2021-05-08
Sieve in UUID mailbox

ksmurchison:
Updated against master 11/22/21


imap/imapd.c
```diff
@@ -11829,7 +11829,6 @@ static int xfer_finalsync(struct xfer_header *xfer)
     if (!r && xfer->userid) {
         r = sync_do_user_seen(&xfer->sync_cs, xfer->userid, replica_seen);
         if (!r) r = sync_do_user_sub(&xfer->sync_cs, xfer->userid, replica_subs);
-        if (!r) r = sync_do_user_sieve(&xfer->sync_cs, xfer->userid, replica_sieve);
```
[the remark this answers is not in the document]
elliefm: I think we're able to find out the version of the destination, aren't we?  So maybe it could be conditional:  if the destination is new enough, don't replicate sieve, but if it's older, it still needs to be replicated.

Ahh, it's `backend_version()` that's used for figuring out the other end's version... and it's based on MAILBOX_MINOR_VERSION (currently 17)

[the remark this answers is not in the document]
ksmurchison: @elliefm @brong I reinstated the sievedir replication code and added a backend_supports_sieve_mailbox() check to NOT do it if the backend is v3.5+.  Cass still passes, but I haven't done any mixed version testing.  Please review.

imap/backend.c
```diff
@@ -1393,3 +1393,12 @@ EXPORTED int backend_version(struct backend *be)
...
 }
+
+EXPORTED int backend_supports_sieve_mailbox(struct backend *be)
+{
+    if (strstr(be->banner, "Cyrus IMAP 3.5")) {
+        return 1;
+    }
+
+    return 0;
+}
```
elliefm, line 1404: Ouch, I don't like this -- this then becomes another function like `backend_version()`, where we have to remember to update it on every branch every time we make a new release.

It's also awkward because "3.5" only supports sieve-in-mailbox if it's a new enough 3.5 to have this PR merged in, but there's a whole years worth of "3.5" builds that didn't.  (Though this is more of a theoretical problem than a real one, since nobody but us uses 3.5 yet, and we don't use XFER.)

I think sieve-in-mailbox warrants bumping MAILBOX_MINOR_VERSION to 18 (and in hindsight I think we probably ought to have bumped it for uuid-mailboxes too, but that's sailed already).

That being the case, this function could then be something like this (or even just a macro):

```
EXPORTED int backend_supports_sieve_mailbox(struct backend *be)
{
    /* sieve mailbox was introduced with MAILBOX_MINOR_VERSION 18 */
    return backend_version(be) >= 18;
}
```

And then the only time we'd ever need to update it again is if some far future version moves sieve storage somewhere else.

I don't remember all the ramifications of bumping MAILBOX_MINOR_VERSION; @brong does this idea seem sane?

elliefm (commented):
I'm about to do some mixed-version testing, will post back with the results in a little while -- but it'll be pure replication tests only, because we don't yet have XFER tests that exercise sieve handling.

elliefm:
Okay, so my cross-version testing setup:

* cassandane is on the sieve-in-uuid-mailbox branch (but rebased on master to get my recent cassanande.ini location changes)
* `[cyrus default]`: /dev/shm/cyrus/main, built and installed from the sieve-in-uuid-mailbox branch
* `[cyrus replica]`: /dev/shm/cyrus/replica, built and installed from the cyrus-imapd-3.4 branch

Most of the `Replication.sieve_*` tests error out, looks like sync_client is failing a `!open_sqldbs` assertion after LOGOUT, and Cassandane notices the non-success exit code:

```
[...]
>1636331814>S5 SYNCAPPLY ACTIVATE_SIEVE %(USERID cassandane FILENAME test1.bc)
<1636331814<S5 OK success
>1636331814>Q01 LOGOUT
<1636331814<* BYE LOGOUT received
Q01 OK Completed
Fatal error: Internal error: assertion failed: lib/sqldb.c: 81: !open_sqldbs
0036520101/sync_client[20512]: Fatal error: Internal error: assertion failed: lib/sqldb.c: 81: !open_sqldbs
[...]
```

It's worth noting that sync_client here is the sieve-in-uuid-mailbox build (only the replica is 3.4).

I have not yet tested the other way (i.e. 3.4 as master, sieve-in-uuid-mailbox as replica).  I'll do that next.

elliefm:
There's a handful of other non-sieve `Replication.*` tests that fail in the above configuration -- but they're all tests marked :min_version_3_5 that are running because the main instance is 3.5, and then getting confused by the replica instance not being 3.5.  I've opened https://github.com/cyrusimap/cassandane/issues/185 to think about later, and am otherwise ignoring them.

elliefm:
With the reversed setup (3.4 as the `[cyrus default]` instance, sieve-in-uuid-mailbox as the `[cyrus replica]` one), the `Replication.sieve*` tests are all fine.

elliefm:
I sketched out an update such that the XFER tests all create a sieve script, and then verify that it was XFER'd correctly: https://github.com/cyrusimap/cassandane/pull/186

I've started out running this on a single build of the sieve-in-uuid-mailboxes Cyrus branch (i.e. all instances are the same build, no cross version stuff), and, this is weird...

* All four of the `MurderIMAP.xfer_user_*` tests successfully complete the XFER, but then when checking whether the new backend has all the data, `assert_sieve_active()` complains that it can't find the "defaultbc" symlink
* Manually checking on disk, the defaultbc symlink is definitely missing
* But I added some extra debug logging to `sievedir_activate_script()` and `sievedir_deactivate_script()`, and it a) creates the symlink with no errors, in the correct place, and b) does not deactivate it later.  So it's baffling that it's not actually there???

It works fine with a Cyrus built from the master branch.

I updated these tests meaning to run them in some cross version configurations, but haven't made it that far yet since I got sidetracked trying to figure out what was going on here.  I don't suppose you have any insight?  I'll keep poking it later in the week, but right now I'm out of ideas what might be going on.

imap/sync_support.c
```diff
@@ -3284,6 +3185,202 @@ int sync_get_uniqueids(struct dlist *kin, struct sync_state *sstate)
...
+            }
+        }
+        else if (S_ISLNK(sbuf->st_mode) && !strcmp(fname, DEFAULTBC_NAME)) {
+            namelen = strlen(link_target) - BYTECODE_SUFFIX_LEN;
+            snprintf(path, sizeof(path), "%.*s", (int) namelen, link_target);
+            if (strarray_find(list, path, 0) >= 0 && path[namelen+1] == 1) {
+                /* Active link for existing script - keep it */
+                return SIEVEDIR_OK;
+            }
+        }
```
elliefm, line 3232: Ahh, here's the cause of the problem with XFER creating a "defaultbc" link during replication,  but not having it anymore later on!

Specifically, this line:

```
            if (strarray_find(list, path, 0) >= 0 && path[namelen+1] == 1) {
 ```

The problem is that `path[namelen+1]` is always `0`, because `path` was just created by `snprintf(..., "%.*s", (int) namelen, ...)` -- we will only ever include namelen bytes, and snprintf always 0-terminates, so the second half of this condition can never succeed -- and therefore we never recognise "Active link for existing script - keep it", and therefore we always throw away the defaultbc if we end up in this callback.

Assuming it's a typo, a quick change to make it compare `== 0` and my XFER test now passes... but at that point the second comparison is completely redundant because it will always be zero anyway, because of the snprintf.  Though the explicit comparison to `1` makes me feel like maybe something more subtle than I'm comprehending was intended here?  So I'm not sure that just removing the second condition is the right fix, but at least we've found the shape of the problem.

File: imap/sync_support.c [hunk 4]
```diff
@@ -3284,6 +3185,202 @@ int sync_get_uniqueids(struct dlist *kin, struct sync_state *sstate)
...
+
+    return 0;
+}
+
+static int remove_cb(const char *sievedir, const char *fname,
+                     struct stat *sbuf, const char *link_target,
+                     void *rock)
+{
+    strarray_t *list = (strarray_t *) rock;
+    char path[PATH_MAX];
```
elliefm, line 3208: Valgrind complains later (at line 3228) that `path` "may be used uninitialised in conditional or jump" (paraphrasing), which is how I found that this function exists and discovered the cause of the problem in my next comment below.

I've checked pretty closely, and I'm pretty sure it's _not_ used uninitialised (it's always initialised by an snprintf)... but we don't check the return value from the snprintf to ensure it succeeded, and it might be that _that_ is what valgrind is really complaining about.

I might be more comfortable checking the return values of snprintf in this function, and also checking that `strlen(...)` is greater than BYTECODE_SUFFIX_LEN (so that the subtractions to calculate `namelen` can't underflow).  But I think we only get in here from `sievedir_foreach()`, and perhaps we trust that it will always give us good input.  I haven't tried to audit it.

But,

```suggestion
    char path[PATH_MAX] = {0};
```

is enough to shut it up.

elliefm: I guess it's showing my comments in the order I composed them rather than in the order the commented-on sections appear in the source, so "in my next comment below" is actually "in the one above" 😅

imap/sync_support.c
ksmurchison: I'm afk at until the morning but almost all of the code in my last commit was taken from master almost verbatim. I'm wondering if this problem also existed there or if the problem is in some changes that I made.

elliefm: master doesn't have `remove_cb()` at all.  I think this code was inherited from the older pre-uuid sieve-in-mailbox implementation, git blame says it's from around May

ksmurchison: OK.  So this issue was twofold:
1. Attempting to use a nasty hack of appending a trailing byte to the script name as an active flag in list_cb()
2. Checking the trailing byte of the wrong string in remove_cb()

I have removed this hack and instead always place the active script first in the array.  If no script is active and empty element will be in the first (zero) position.

Let me know if this solves the issue

elliefm: That seems to have fixed my XFER tests, great.  Now I'll try running them in a mixed version setup and see what happens

elliefm:
Hmm, these two are failing for me:

FastMail.rename_deepfolder_intermediates_rightnow
FastMail.rename_deepfolder_intermediates

Looks like they do a bunch of work with an intermediate folder, replicating and renaming it around, and that's all fine.  But then they each "delete user.cassandane", replicate that, and expect all of cassandane's files to have been cleaned up on the replica -- except "cassandane.dav" remains.

I'm guessing here... maybe the delete codepath is missing consideration of the new sieve_scripts table, and then the database gets left behind because it's not empty?

elliefm:
Replication tests are now fine between 3.4 <-> 3.5 (both ways), cool

For XFER, 3.5 -> 3.4:
* specialuse annotations seem to be lost for some reason, but this happens with master->3.4 too, so it's a separate problem (now #3771)
* sieve replicates properly -- destination server has the correct files and the defaultbc link etc, great!
* but the "#sieve" mailbox is also replicated, and shows up in LIST since 3.4 doesn't know it's special

Haven't tested 3.4->3.5 yet (nor any other versions), will do so after lunch and post the results in a fresh comment.

elliefm:
3.4 -> 3.5 (sieve-in-uuid-mailbox):

XFER tests are failing with an IOERROR in syslog:

```
[30230] =====> Instance::_check_syslog[1392] Cassandane::Instance=HASH(0x55e268d4acc0) syslog error: Nov 12 14:08:29 debian 0308260103/imap[30310]: IOERROR: stat /dev/shm/cass/0308260103/conf/sieve/uuid/f/0/f0b17d2c-6e7e-4f01-ac09-45238b7412a9/harmful.script: No such file or directory
```

("harmful" is a pretty funny word for `random_word()` to choose as the script name, hah)

Looking on disk, that file definitely doesn't exist.  There _is_ a `./tmp/scriptOTym_` tempfile containing the script contents ("keep;").

I think something is going wrong when the old-style sieve replication is processed.  Here's snippets from the telemetry:

First it uploads and activates the sieve, and all seems successful:

```
<1636686509<S5 SYNCAPPLY SIEVE %(USERID cassandane FILENAME harmful.script LAST_UPDATE 1636686508 CONTENT keep;)
>1636686509>S5 OK success
<1636686509<S6 SYNCAPPLY ACTIVATE_SIEVE %(USERID cassandane FILENAME harmful.bc)
>1636686509>S6 OK success
```

Then it fetches the current state:

```
<1636686509<S7 SYNCGET USER cassandane
[ snip MAILBOX responses because they're not interesting right now ]
* SIEVE %(FILENAME harmful.script LAST_UPDATE 1636686509 GUID 1741a5f56c1cfc0f7ef3d0ad53bceb6518bdfdad ISACTIVE 0)
S7 OK success
```

Note "ISACTIVE 0".

The primary notices the activation state doesn't match, and tries again:

```
<1636686509<S8 SYNCAPPLY ACTIVATE_SIEVE %(USERID cassandane FILENAME harmful.bc)
>1636686509>S8 OK success
```

Which succeeds, and then I think we get to the final sync, which we start by checking state again...

```
<1636686509<S9 SYNCGET USER cassandane
[ snip MAILBOX responses again ]
* SIEVE %(FILENAME harmful.script LAST_UPDATE 1636686509 GUID 1741a5f56c1cfc0f7ef3d0ad53bceb6518bdfdad ISACTIVE 0)
S9 OK success
```

"ISACTIVE 0" again.  Send some final updates:

```
<1636686509<S13 SYNCAPPLY LOCAL_UNMAILBOX user.cassandane.#sieve
>1636686509>S13 OK success
<1636686509<S14 SYNCAPPLY ACTIVATE_SIEVE %(USERID cassandane FILENAME harmful.bc)
>1636686509>S14 OK success
```

Hmm, that LOCAL_UNMAILBOX is interesting.  It makes sense because 3.4 doesn't have a "#sieve" mailbox, yet there is one on the replica.  In the above, where I snipped the MAILBOXES responses, the "#sieve" mailbox was included.  I'm not sure if this is a problem or not, maybe the replica should exclude the "#sieve" mailbox from the SYNCGET response if it knows it's talking to a version that doesn't support it.

Once again we activate the sieve script normally and it reports success.  It's worth noting that on disk, defaultbc is indeed a symlink to harmful.bc, so it is correctly active at the end of the run, but I guess some internal state is confused.  But the harmful.script is missing.

The "IOERROR: stat..." looks like it's coming from sieve_rebuild when it wants to know the mtime/etc of the script file, and looking directly at syslog, it seems to be during the final "SYNCAPPLY ACTIVATE", _after_ the "LOCAL_UNMAILBOX".

The LOCAL_UNMAILBOX only happens the one time, I guess it's one of the things that's special about the final sync stage.  But the "ISACTIVE 0" has been occurring all along.  So I think there's two things here when sieve-in-uuid-mailbox is a replica:

* it reports the "#sieve" mailbox existing even when the primary end won't have it or know what to do with it, causing it the primary end to send a LOCAL_UNMAILBOX, which the replica will then honour and lose the original script file
* something in SYNCGET USER is failing to detect the script as being activated, even though the activation itself reported success, and even though on disk the defaultbc links to the bc file correctly

ksmurchison:
@elliefm The FastMail.rename_deepfolder failures should be fixed by https://github.com/cyrusimap/cyrus-imapd/commit/8f48ccf7062663d357889e0bf551d34ff38202e4

imap/user.c
```diff
@@ -328,6 +328,12 @@ EXPORTED int user_deletedata(const mbentry_t *mbentry, int wipe_user)
     search_deluser(mbentry);
 
 #ifdef WITH_DAV
+    /* delete the entire DAV DB */
+    struct buf fname = BUF_INITIALIZER;
+    dav_getpath_byuserid(&fname, userid);
+    unlink(buf_cstring(&fname));
+    buf_free(&fname);
+
```
elliefm, line 336: This does seem to fix the FastMail.rename_deepfolder... failures, but it's odd that it's necessary.  To my reading "dav" is one of the `user_file_suffixes`, so it should have been added to `paths`, and then deleted by the loop over `paths` a few lines below??

ksmurchison: Yeah, you're right.  This was my first attempt at a fix which didn't work.  The real fix was actually NOT creating the DAV DB when the user INBOX doesn't yet exist.  Reverting this commit.

elliefm:
With just this branch (no cross-version) I'm seeing these failures:

* Cyrus::JMAPCore.blob_download

crash from httpd?!:

```
Core was generated by `/dev/shm/cyrus/main/libexec/httpd -C /dev/shm/cass/0310400280/conf/imapd.conf'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f88642e146e in lcase (str=0x0) at lib/util.c:261
261	    while (*scan) {
(gdb) bt
#0  0x00007f88642e146e in lcase (str=0x0) at lib/util.c:261
#1  0x0000556caa69ddf4 in parse_accept (hdr=0x556cac326f50) at imap/httpd.c:2475
#2  0x0000556caa6ca420 in parse_accept_header (hdr=0x556cac326f50)
    at imap/http_jmap.c:542
#3  0x0000556caa6cabbf in jmap_download (txn=0x7ffd44c7b7a0) at imap/http_jmap.c:715
#4  0x0000556caa6c9e0b in meth_get (txn=0x7ffd44c7b7a0, params=0x0)
    at imap/http_jmap.c:392
#5  0x0000556caa69c80d in process_request (txn=0x7ffd44c7b7a0) at imap/httpd.c:1958
#6  0x0000556caa69ce0b in http1_input (txn=0x7ffd44c7b7a0) at imap/httpd.c:2096
#7  0x0000556caa69d179 in cmdloop (conn=0x556caa7b2280 <http_conn>) at imap/httpd.c:2202
#8  0x0000556caa69a0bb in service_main (argc=1, argv=0x556cac2c7eb0, envp=0x7ffd44c80558)
    at imap/httpd.c:1056
#9  0x0000556caa6c9331 in main (argc=3, argv=0x7ffd44c80538, envp=0x7ffd44c80558)
    at master/service.c:647
```

* Cyrus::JMAPSieve.sieve_set_replication

Abnormal exit from sync_client 

* Cyrus::MurderIMAP.xfer_user_altns_unixhs
* Cyrus::MurderIMAP.xfer_user_altns_unixhs_virtdom
* Cyrus::MurderIMAP.xfer_user_noaltns_nounixhs
* Cyrus::MurderIMAP.xfer_user_noaltns_nounixhs_virtdom

errors found in syslog , which all look like this:

```
Nov 16 14:24:27 debian 03104001137/imap[30898]: IOERROR: failed to create user.cassandane.#sieve (unable to reserve mailbox on mupdate server)
```

Note that the MurderIMAP.xfer_user... tests are my modified versions, not the upstream ones (which don't look at sieve at all, so are probably fine).  They're in my branch here if you want to run them yourself: https://github.com/cyrusimap/cassandane/pull/186

I haven't dug in deep on any of these yet, but now that I've at least recorded them for posterity, I will.

elliefm:
Telemetry from the xfer:

```
[ snip replicating messages and mailboxes, which seems fine ]
<1637034972<S5 SYNCAPPLY SIEVE %(USERID cassandane FILENAME yeltsin.script LAST_UPDATE 1637034972 CONTENT keep;)
>1637034972>S5 NO Unknown error
* BYE Fatal error: Internal error: assertion failed: imap/imapd.c: 1276: !open_mailboxes_exist()
<1637034972<S6 SYNCRESTART
```

I think the "NO Unknown error" is when the mupdate reserve has failed.  And then the assertion at imapd.c:1276 is the top of the cmdloop loop -- so I guess when the mupdate reserve fails we're not cleaning something up properly on the way back up the call chain.

It's worth noting that it's the very first pass through the replication that's failing here (so, xfer_initialsync I guess).

Ahh, this is interesting, here's mupdate telemetry:

```
---------- mupduser Tue Nov 16 14:56:12 2021

<1637034972<X0 RESERVE {22+}
user.cassandane.#sieve {22+}
localhost:9101!default
>1637034972>X0 OK "done"
<1637034972<X1 ACTIVATE {22+}
user.cassandane.#sieve {22+}
localhost:9101!default {56+}
cassandane      lrswipkxtecdan  admin   lrswipkxtecdan  anyone  p       
>1637034972>X1 OK "done"
<1637034972<Q01 LOGOUT
>1637034972>Q01 OK "bye-bye"
```

okay, great, it worked,

```
---------- mupduser Tue Nov 16 14:56:12 2021

<1637034972<X0 RESERVE {22+}
user.cassandane.#sieve {22+}
localhost:9103!default
>1637034972>X0 NO "mailbox already exists"
<1637034972<Q01 LOGOUT
>1637034972>Q01 OK "bye-bye"
```

So it's trying to reserve it a second time and not being allowed, that makes sense.

I can't tell from the telemetry log which client it's talking to when.  I guess the first reserve/activate is when the cassandane user is being set up on the primary, and the second one that's rejected is when the XFER target receives the `SYNCAPPLY SIEVE` and needs to create its own `#sieve` folder.  Hmm.

elliefm:
> Note that the MurderIMAP.xfer_user... tests are my modified versions, not the upstream ones (which don't look at sieve at all, so are probably fine). They're in my branch here if you want to run them yourself: [cyrusimap/cassandane#186](https://github.com/cyrusimap/cassandane/pull/186)

I'm not sure what the etiquette is on pushing to each other's branches, but your branch for the cassandane tests is on this remote (unlike mine, which is on my fork), so I _could_ just push my sieve updates to the xfer tests directly to your branch, and then you wouldn't need to stuff around pulling my branch.  Let me know if you'd like me to do this!

ksmurchison:
@elliefm The blob issue is fixed.  I forgot to rebase on master.

I also believe that I have fixed the XFER issue with commit

elliefm:
I've found the problem with `JMAPSieve.sieve_set_replication` --  UNSIEVE is broken (I guess, in the entirety of Cassandane, this is the only test that tries to replicate after deleting or renaming a script!).  It renames the script from "foo" to "bar" and then replicates:

```
>1637108849>S2 SYNCAPPLY SIEVE %(USERID cassandane FILENAME bar.script LAST_UPDATE 1637108848 CONTENT {7+}
keep;
)
<1637108849<S2 OK success
UNSIEVE cassandane foo.script
>1637108849>S3 SYNCAPPLY UNSIEVE %(USERID cassandane FILENAME foo.script)
<1637108849<S3 NO Unknown error
0027260101/sync_client[21993]: UNSIEVE received NO response: Unknown error
Error from sync_do_user(cassandane): bailing out!
0027260101/sync_client[21993]: Error in sync_do_user(cassandane): bailing out!
```

Notice that the filename used over the replication protocol is the ".script" name, not the ".bc" name...  But `sync_apply_unsieve()` just passes the filename it was given straight through to `sync_sieve_delete()`'s new `bcname` argument.

`sync_sieve_delete` carves up the name expecting to convert "foo.bc" -> "foo", but it was actually "foo.script", so it winds up trying to look up "foo.scr" in the database:

```
Nov 17 11:27:29 debian 0027260102/sync[21980]: sqldb_exec(/dev/shm/cass/0027260102/conf/user/uuid/d/5/d5624396-4963-4747-b979-9aea7eed9e09/dav.db): SELECT rowid, creationdate, lastupdated, mailbox, imap_uid,  modseq, createdmodseq, id, name, content, isactive, alive FROM sieve_scripts WHERE name = 'foo.scr';
```

and that isn't found, so it bails out.

The name sent over the replication protocol needs to remain as the "foo.script" version for back compat, so I guess either `sync_sieve_delete()` needs to accept a scriptname instead of a bcname, or `sync_apply_unsieve()` needs to translate before it calls down.  I guess maybe it doesn't matter which, since these functions are deeply coupled anyway.

timsieved/actions.c
```diff
@@ -194,218 +215,182 @@ int capabilities(struct protstream *conn, sasl_conn_t *saslconn,
...
-  prot_printf(conn, "OK\r\n");
+    if (!verify_only) {
+        struct sieve_data *sdata = NULL;
 
-  return TIMSIEVE_OK;
-}
+        result = sievedb_lookup_name(sievedb, mailbox_name(sieve_mailbox),
+                                     buf_cstring(name), &sdata, 0);
+        if (!result || result == CYRUSDB_NOTFOUND) {
+            sdata->name = buf_cstring(name);
```
brong, line 289: I don't love this design where there's a `static struct sieve_data` inside `sievedb_lookup_name` when gets filled and zeroed regardless of whether data is found or not.  Should we be defining the struct out here and passing it into the lookup instead?

elliefm:
XFER, 3.5->3.4: the tests just seem to be failing on the annotations problem now.  Interestingly, the "noaltns nounixhs" test passes, which suggests that the annotations problem is something to do with either altns, unixhs, or virtdomains.  I'll keep digging further into this on the other issue.

I do notice one new oddity in the replication (by inspecting telemetry -- it hasn't caused a problem in anything the test actively exercises):

```
<1637115160<S8 SYNCAPPLY RESERVE %(PARTITION default MBOXNAME (user.cassandane user.cassandane.Drafts) GUID (78eb8071205744da4dbef1adffb2c702eb52bdcf))
>1637115160>* MISSING (78eb8071205744da4dbef1adffb2c702eb52bdcf)
S8 OK success
[snip subsequent replication stages]
```

It's hard to tell after the test finishes (since the 3.5 server no longer has this mailbox, and the 3.4 server does not use guid's for sieve scripts), but I think this is the 3.5 server thinking about sending the one message (script) from the #sieve mailbox over.  It's interesting that it asks the replica if it knows about it, the replica reports it missing, but then it ignores it being missing and carries on with the next sync stages instead.

Anyway, it looks like the "ignore #sieve mailbox for replication purposes" approach is working so far.

imap/mailbox.c
```diff
@@ -3923,6 +3943,233 @@ EXPORTED int mailbox_add_email_alarms(struct mailbox *mailbox)
...
+    if (!old && isexpunged) return 0;
+
+    r = mailbox_cacherecord(mailbox, new);
+    if (r) goto done;
+
+    /* Get script id from filename param in Content-Disposition header */
+    message_read_bodystructure(new, &body);
+    for (param = body->disposition_params; param; param = param->next) {
+        if (!strcmp(param->attribute, "FILENAME")) {
+            ssize_t idlen = strlen(param->value) - strlen(SIEVE_EXTENSION);
```
brong, line 3999: If the ID is the filename, do we need to restrict filenames via timsieved to be legit JMAP OBJECTID (length + valid chars)?  Or is this always a separate JMAP-compatible ID?  (I see it's mainly made by uuidgen)

brong (changes requested):
I have a bunch of comments on here - maybe I should line up a chat about them!

One thing - the "S" prefixed blobId.  I don't think you need that any more.  The body of the script is going to have its sha1 calculated when it's stored as an email on the disk, which means that it will be have a 'G' key and be discoverable by the standard blob mechanisms.

This is excellent because it means that when you create or replace a script with a normal blobId, the resulting script has the same blobId.

Further - I'm not certain that there's any benefit to storing the raw script contents in the sqlite database.  It's easy to get with the standard blob mapping smarts, and we still create a compiled version on disk for the lmtp daemon to use.

elliefm:
XFER 3.4->3.5 (i.e. other way round from my previous comment) is working nicely, cool

timsieved/actions.c
ksmurchison: This is the same as caldav_db and carddav_db.  Should we change all of them?

ksmurchison:
@elliefm The Sieve replication issues should be fixed by my latest commit

imap/mailbox.c
ksmurchison: The filename param is actually the JMAP ID, so its always a GUID that we generate.

imap/sync_support.c
```diff
@@ -2740,6 +2653,11 @@ int sync_apply_mailbox(struct dlist *kin,
     dlist_getatom(kin, "MBOXTYPE", &mboxtype);
     mbtype = mboxlist_string_to_mbtype(mboxtype);
 
+    if (mbtype_isa(mbtype) == MBTYPE_SIEVE) {
```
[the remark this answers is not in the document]
ksmurchison: Fixed

imap/mboxname.c
```diff
@@ -1669,6 +1669,26 @@ EXPORTED int mboxname_isjmapuploadmailbox(const char *name, int mbtype __attribu
...
+EXPORTED int mboxname_issievemailbox(const char *name, int mbtype)
+{
+    if (mbtype_isa(mbtype) == MBTYPE_SIEVE) return 1;  /* Only works on backends */
+    int res = 0;
+
+    mbname_t *mbname = mbname_from_intname(name);
+    const strarray_t *boxes = mbname_boxes(mbname);
+    const char *prefix = config_getstring(IMAPOPT_SIEVE_FOLDER);
+
+    if (strarray_size(boxes) && !strcmpsafe(prefix, strarray_nth(boxes, 0)))
```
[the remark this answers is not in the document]
ksmurchison: So, do want to make sure that strarray_size(boxes) == 1 here, or do you mean actually prevent children to be created?

Same issue with mboxname_is*() functions above this one.

imap/sieve_db.c
```diff
@@ -0,0 +1,1046 @@
...
+        /* Load message containing the resource */
+        data = buf_new();
+
+        r = mailbox_map_record(mailbox, &record, data);
+        if (r) {
+            buf_destroy(data);
+            data = NULL;
+        }
+        else {
+            buf_remove(data, 0, record.header_size);
```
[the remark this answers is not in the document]
ksmurchison: This function isn't actually used anymore, but something like it WILL be needed if we don't store the script content in the DAV DB record

imap/sieve_db.c
```diff
@@ -0,0 +1,1046 @@
...
+        sdata->mailbox =
+            column_text_to_buf((const char *) sqlite3_column_text(stmt, 3),
+                               &db->mailbox);
+        sdata->id =
+            column_text_to_buf((const char *) sqlite3_column_text(stmt, 7),
+                               &db->id);
+        sdata->name =
+            column_text_to_buf((const char *) sqlite3_column_text(stmt, 8),
+                               &db->name);
+        sdata->content =
```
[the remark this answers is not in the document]
ksmurchison: Yes, we discussed this a while ago.  I don't recall exactly why we did this, but its certainly faster than having to mmap the message and use the body.

imap/mboxname.c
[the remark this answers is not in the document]
ksmurchison: Since this same issue occurs in multiple functions, waiting until after this PR lands on master.

imap/sieve_db.c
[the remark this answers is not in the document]
ksmurchison: sieve_script_fetrch() has been totally re-written now that we don't store script content in Sieve DB

timsieved/actions.c
ksmurchison: Since the same issue is present in caldav_db, carddav_cb, and webdav_db, will wait to resolve this until after this PR lands on master.

ksmurchison:
@brong Please review the last 4 commits.
@elliefm Please test this branch again.  I believe all replication issues have been fixed.

imap/mailbox.c
```diff
@@ -4076,7 +4077,9 @@ static int mailbox_update_sieve(struct mailbox *mailbox,
         sdata->isactive = isactive;
         sdata->id = id;
         sdata->name = name;
-        sdata->content = content;
+
+        message_guid_generate(&uuid, content, strlen(content));
```
brong, line 4081: Yeah, this is fine.  We could theoretically read it from the cache datastructure instead of mapping the message again, but this is legit as well.

ksmurchison: Heh. That didn't even occur to me. I was looking for it in the index record instead.

brong (approved):
This looks fantastic.  It was quite neat in the end :)  Thanks for doing this.

ksmurchison:
Cool. If ellie's testing goes ok I'll talk to rjbs about getting it on staff.

elliefm:
All looking good for 3.5->3.5

3.5->3.4 also looking good, only problem remaining here is the annotations one, which Bron now has a fix in progress for

About to test 3.4->3.5...

elliefm:
3.4->3.5 looks good too!

imap/mailbox.c
ksmurchison: @brong https://github.com/cyrusimap/cyrus-imapd/pull/3487/commits/a6b3a468ed1356b49574217b12ac412002c7fb52

brong:
@ksmurchison I just added a commit so that the upgrade code only runs for exactly version 12 databases rather than for anything earlier - because we can't add the content -> content_id translation to a database which doesn't even have the sieve data table!

brong:
actually I'm just gonna merge this one :)  Clashes galore otherwise.

PR #3489, opened 2021-05-11
syslog: Don't define syslog() if glibc already provides a macro

tbaederr:
If the compiler in use does not support `__va_arg_pack()`, GLIBC defines `syslog()` as a macro, and redefining it here breaks the build.

### DIFF
--- a/cunit/syslog.c
+++ b/cunit/syslog.c
@@ -157,6 +157,11 @@ __syslog_chk(int prio, int whatever __attribute__((unused)),
     vlog(prio, fmt, args);
     va_end(args);
 }
+
+/* glibc might define a syslog() macro, which is not wanted here */
+#ifdef syslog
+#undef syslog
+#endif
 #endif
 
 EXPORTED void syslog(int prio, const char *fmt, ...)


elliefm:
The point of this `syslog()` function is to inject our own syslog in front of the system one, so that our unit tests can capture calls to it.  With it ifdef'd out, I expect the unit tests that expect things to be syslogged will fail due to not finding the captured logs?

If the problem is that there is already a `syslog()` macro in the way, perhaps the right thing to do is undef that?  What does that macro do?

The thing being done here is almost certainly somewhat gcc/glibc-specific (because that is what we mostly use).  It would be helpful to know what you're using that's different, and what problems you're seeing.  It might turn out to be fixable, or it might just be that your platform cannot run our unit tests, in which case you can reconfigure without `--enable-unit-tests` and the entire cunit/ directory will be ignored.

tbaederr:
The macro I was talking about is here: https://sourceware.org/git/?p=glibc.git;a=blob;f=misc/bits/syslog.h;h=6f3137a98ee593f3b45f6e7d93dfbabd83e34a44;hb=HEAD#l34

I am using clang, which does not support `__va_arg_pack()`, so the macro is defined. I haven't extensively tested the patch, but it since the cyrusimap implementation of `syslog()` does not use `__va_arg_pack()`, I guess it would be finde to `undef` the macro instead.

elliefm:
Ahh, I see it's just deciding whether to implement syslog() as a function or as a macro, but in either case they both just call `__syslog_chk()`.

And we already have a replacement `__syslog_chk()` identical to our replacement `syslog()`, with this comment above it:

```
/* Under some but not all combinations of options, glibc
 * defines syslog() as an inline that calls this function */
```

So I think compiling out our `syslog()` is okay after all?  Like, if the macro is present, it will call our `__syslog_chk()` replacement, and so our `syslog()` replacement isn't needed?  I need to think about this more though....  I think if we know that the syslog macro will call `__syslog_chk()`, then it's fine to not provide a `syslog()`.  But if we don't know what the syslog macro will do, we'd better undefine it and provide our function instead.  So rather than just `#ifndef syslog`, I think the condition needs to interact with the glibc check above, but I'm distracted at the moment and can't quite see the shape of what I want.

tbaederr:
> So I think compiling out our `syslog()` is okay after all? Like, if the macro is present, it will call our `__syslog_chk()` replacement, and so our `syslog()` replacement isn't needed? I need to think about this more though.... I think if we know that the syslog macro will call `__syslog_chk()`, then it's fine to not provide a `syslog()`. But if we don't know what the syslog macro will do, we'd better undefine it and provide our function instead. So rather than just `#ifndef syslog`, I think the condition needs to interact with the glibc check above, but I'm distracted at the moment and can't quite see the shape of what I want.

I changed the commit to undef the macro instead.  I haven't tested it at all, but one thing I was worried about is other files including `<syslog.h>`, which would still have the macro I think. Either way is fine with me.

PR #3490, opened 2021-05-11
byteorder: fix broken GNUC macros and use builtins for 2-byte values

rsto:


elliefm:
This looks familiar, haven't we fixed this before?

elliefm:
Oh I see: we got a patch in #3462 that changed it to not use `__GNUC_PREREQ()`, which was not portable.  But the patch misspelled `__GNUC__` as `__GNU__` and `__GNUC_MINOR__` as `__GNUC_MINOR`, which this PR fixes.

I've picked this back to 3.4, since that also had the bad patch.

PR #3491, opened 2021-05-11
jmap_ical: round-trip ORGANIZER-only and ATTENDEE-only events

rsto:
VEVENTs may contain only ORGANIZER or ATTENDEEs and the JMAP code currently rejects these events when written back to iCalendar.

This patch allows to update such events. It still does reject such events during creation in JMAP.

Tested in https://github.com/cyrusimap/cassandane/commit/9879f08e72acd47d54b46c5dec6665606ee94fa9 and https://github.com/cyrusimap/cassandane/commit/65c2e29abddd70e9f15e2276ad4166097ca43c09


ksmurchison (approved):
I'm not an expert in the iCal <-> JSCal code but this look good.
Do we have tests?

ksmurchison:
Sorry, missed that.

PR #3493, opened 2021-05-18
jmap_mail: fix break condition for unsupported guidsearch rank

rsto:
Also fixes a memory leak for early returns.


ksmurchison (approved):
Looks sane to me

PR #3497, opened 2021-05-19
squatter: change -s option to require an argument

### LINKED ISSUE #3496: squatter -s/--squat-skip "delta" optarg is supposed to be optional, but throws usage when missing
This came up on the info list: https://cyrus.topicbox.com/groups/info/T290da80a662e5a08-M291cfe06612f34cc1feb1a42

Here's how we document it:

> -s, --squat-skip[=delta]
> Skip  mailboxes  that  have  not  been modified since last index. This is achieved by comparing the last modification time of a mailbox to the last time  the squat index of this mailbox got updated. If the mailbox modification time (plus delta) is less than the squat index modification time, then the mailbox is skipped. The optional argument value delta is defined in seconds and must be equal to or higher than zero, the default value is 60.  Squat only.  This feature is only available on the master branch.

("This feature is only available on the master branch" is a lie, it has been backported to 3.4 as of 3.4.0.)

Here's how we implement it:
https://github.com/cyrusimap/cyrus-imapd/blob/097526869c5be0528a68bd1e245570201c8f93b6/imap/squatter.c#L1107-L1121

and we specify it in the long_options table like: `{"squat-skip", optional_argument, 0, 's' },`

It LOOKS like it should work exactly the way we expect: if there is no delta, use 60; if there is a valid delta, use that; and if there is a junk delta, throw a usage error.  But instead, it seems to be regarding "no delta" and "junk delta" as the same, and throwing the usage error instead of defaulting to 60.  I guess we're misunderstanding something about how to use optional optargs with `getopt_long()`.

We have a test for this in SearchSquat.skip_unmodified_slow (which, because of the _slow suffix, will only run if you run it by name, or provide the --slow argument to testrunner), but it explicitly specifies a delta of 0.  So we have a test that confirms the functionality works, but the test doesn't exercise the optionality, which I guess is how we missed this.

rsto:
It didn't work with optional arguments anyway.

Fixes #3496

### DIFF
--- a/changes/next/squatter_s_argument_required
+++ b/changes/next/squatter_s_argument_required
@@ -1,0 +1,12 @@
+Description:
+
+Changes the squatter -s option to require an argument
+
+Config changes:
+
+None.
+
+
+Upgrade instructions:
+
+None. Optional argument handling of the -s option was broken already.
--- a/docsrc/imap/reference/manpages/systemcommands/squatter.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/squatter.rst
@@ -237,17 +237,18 @@ Options
     Recursively create indexes for all sub-mailboxes of the user,
     mailboxes or mailbox prefixes given as arguments.
 
-.. option:: -s, --squat-skip[=delta]
+.. option:: -s delta, --squat-skip=delta
 
     Skip mailboxes that have not been modified since last index. This is
     achieved by comparing the last modification time of a mailbox to
     the last time the squat index of this mailbox got updated. If the
-    mailbox modification time (plus delta) is less than the squat
-    index modification time, then the mailbox is skipped. The optional
-    argument value delta is defined in seconds and must be equal to or
-    higher than zero, the default value is 60.
+    mailbox modification time plus delta is less than the squat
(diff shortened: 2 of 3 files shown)


elliefm (changes requested):
This looks good, thanks.  The "requested changes" are just nit picking the documentation, since we're in there anyway we might as well nail it.

I have a question about the `SearchSquat.skip_unmodified_slow` cassandane test.  It only takes a couple of seconds to run, so I guess the "_slow" tag is probably a leftover from some old work-in-progress version that took a lot longer, before it was cut down to the current version.

I'd like to remove the _slow tag so we just run this test all the time, what do you think?

For comparison, the other "_slow" tests are `Bug3072.copy_longset_slow` (runtime: 1m47s) and `ClamAV.remove_infected_slow`, which is very quick because it errors out!  I don't remember how long it takes when it works, but it's a good argument against hiding stuff behind "_slow" unless it's really obnoxious, because we won't necessarily notice when it breaks... 😅

rsto:
Thanks a lot @elliefm ! I have updated the man page with both of your review comments. I also removed the `_slow` magic from the Cassandane test.

elliefm (approved):
Looks good

PR #3498, opened 2021-05-19
CardDAV, JMAP: Add 'vcard_max_size' option

ksmurchison:
Tested with https://github.com/cyrusimap/cassandane/pull/123


ksmurchison:
Should the option name be 'contact_max_size' rather than 'vcard_max_size'?  I don't really care, but will admins know that the storage format is vCard, or what vCard actually is?

imap/carddav_db.c
```diff
@@ -1120,12 +1130,18 @@ EXPORTED int carddav_store(struct mailbox *mailbox, struct vparse_card *vcard,
     time_to_iso8601(now, datestr, sizeof(datestr), 0);
     vparse_replace_entry(vcard, NULL, "REV", datestr);
 
+    /* Check size of vCard */
+    struct buf buf = BUF_INITIALIZER;
+    vparse_tobuf(vcard, &buf);
```
rsto, line 1135: For CardDAV, this check, including the VCARD dump, would be done twice, right?

rsto (approved):
This looks good to me. I wonder if we could do the size limit check just once for CardDAV (if my assumption is right)

imap/carddav_db.c
ksmurchison: carddav_store() appears to only be called by jmap_contact().  We should probably try to find a way to consolidate that and http_carddav.c:store_resource()

rsto: Thanks. Agree that we should consolidate, but also that this PR isn't the right place to do it.

PR #3499, opened 2021-05-20
CalDAV, JMAP: Add 'icalendar_max_size' option

ksmurchison:
Tested with https://github.com/cyrusimap/cassandane/pull/124


imap/jmap_calendar.c
```diff
@@ -2578,6 +2578,12 @@ static int setcalendarevents_create(jmap_req_t *req,
     const char *calendarId = NULL;
     strarray_t schedule_addresses = STRARRAY_INITIALIZER;
 
+    static int icalendar_max_size = -1;
```
rsto, line 2581: Nit: I might have missed during the review for CardDAV: it's a pity that libconfig only supports signed int, which on my machine is a 4-byte signed integer. I guess any meaningful event will be way smaller that 2GB. Should we enforce an explicit maximum?

rsto (approved):
Looks good to me, only have a comment on max_size limits

imap/jmap_calendar.c
ksmurchison: Not sure what you mean by explicit maximum.  Do you mean within libconfig?

rsto: I meant to state a maximum in the doc. But it's really not that important.

PR #3500, opened 2021-05-20
Add FROMADDR= parameters to SMTP SCHK command for JMAP future delivery

ksmurchison:


imap/smtpclient.c
```diff
@@ -394,12 +394,25 @@ static int smtpclient_rset(smtpclient_t *sm)
     return r;
 }
 
-static int smtpclient_schk(smtpclient_t *sm)
+static int smtpclient_schk(smtpclient_t *sm, json_t *fromaddr)
```
rsto, line 397: Is this the first time we pull in json into smtplib? If so, could we reuse the existing smtp_addr_t type (or a list of it) instead of a JMAP EmailAddress object?

imap/smtpclient.c
```diff
@@ -863,6 +891,19 @@ EXPORTED int smtp_is_valid_esmtp_value(const char *val)
     return 1;
 }
 
+/* Encodes into the current location of the struct buf.
+   The caller must buf_reset() before calling is necessary. */
```
rsto, line 895: Nit type: 'is' should mean 'if'. I am generally bad with comma, but I think the "if necessary" should be preceded by one.

imap/smtpclient.h
```diff
@@ -108,6 +112,11 @@ extern int smtpclient_open_host(const char *addr, smtpclient_t **smp);
 extern int smtpclient_send(smtpclient_t *sm, smtp_envelope_t *env, struct buf *data);
 extern int smtpclient_sendprot(smtpclient_t *sm, smtp_envelope_t *env, struct protstream *data);
 
+/* Perform SMTP envelope (and optionally size and/or From: header address checks)
```
rsto, line 115: This is find hard to parse. Would "Check the SMTP envelope" work instead of "Perform .."

rsto (commented):
Logic looks good to me. Found some nits.

imap/smtpclient.c
ksmurchison: I'm not too keen on passing json into smtpclient.c either but I also didn't want to have to translate the json array into some other format, but its probably best that I do.

ksmurchison: Changed to using a strarray_t

rsto (approved):
Looks good to me

PR #3501, opened 2021-05-25
search_xapian: upgrade indexed.db version during uuid relocation

rsto:
This fixes the indexed.db implementation for uuid mailboxes:

- Relocates the xapian directories for all active tiers
- Only upgrades indexed.db entries from mailbox name to mailbox-id format during relocate_by_id
- Supports existing and new indexed.db files for both relocated and not-relocated users

### DIFF
--- a/imap/relocate_by_id.c
+++ b/imap/relocate_by_id.c
@@ -458,7 +458,7 @@ static void get_searchparts(const char *key, const char *val, void *rock)
 
     basedir = user_hash_xapian(srock->userid, val);
     if (basedir) {
-        struct buf buf;
+        struct buf buf = BUF_INITIALIZER;
 
         buf_initm(&buf, basedir, strlen(basedir));
         buf_appendcstr(&buf, XAPIAN_DIRNAME);
--- a/imap/relocate_by_id.c
+++ b/imap/relocate_by_id.c
@@ -292,6 +292,20 @@ int main(int argc, char **argv)
                 }
                 if (namelock) mboxname_release(&namelock);
             }
+            else if (mbentry->mbtype & MBTYPE_INTERMEDIATE) {
+                if (!quiet) printf("\tPromoting intermediary: %s\n", extname);
+
+                if (!nochanges) {
+                    r = mboxlist_promote_intermediary(mbentry->name);
+                    if (namelock) mboxname_release(&namelock);
+
+                    if (r) {
+                        fprintf(stderr,
+                                "\tFailed to promote intermediary %s: %s\n",
+                                extname, error_message(r));
+                    }
+                }
+            }
             else if (!nochanges) {
                 /* Rewrite mbentry */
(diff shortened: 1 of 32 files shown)


rsto:
@brong do we also want squatter to support indexed.db format upgrades? Right now, the only way to trigger them is by either creating a new user, or relocating an existing one.

ksmurchison (commented):
I'm not an expert on the Xapian code, but this looks sane to me.  And if it works, it works.

imap/relocate_by_id.c
```diff
@@ -255,9 +258,46 @@ int main(int argc, char **argv)
...
+                    int j;
+                    for (j = 0; j < strarray_size(items); j++) {
+                        const char *item = strarray_nth(items, j);
+                        const char *col = strrchr(item, ':');
+                        if (!col) continue;
+                        buf_setmap(&buf, item, col-item);
+                        strarray_append(&rock.tiernames, buf_cstring(&buf));
+                        arrayu64_append(&rock.tiergens, atoi(col+1));
+                    }
+                    config_foreachoverflowstring(get_searchparts, &rock);
```
brong, line 292: I suspect I would have done this the other way out - just resolved the on-disk path for each `tier:gen` with old-style and new-style names and renamed those (if old existed and new didn't) rather than using overflowstrings at all.

imap/search_xapian.c
```diff
@@ -587,199 +614,116 @@ static int tierexists_cb(void *rock, const char *key, size_t keylen,
...
-                      data, datalen, &mr->txn);
-    buf_truncate(mr->buf, prefix_len);
+    buf_appendmap(rock->buf, key, keylen);
+    r = cyrusdb_store(idb->db, buf_base(rock->buf), buf_len(rock->buf),
+            data, datalen, &idb->txn);
+    buf_truncate(rock->buf, prefix_len);
     if (r) {
         syslog(LOG_ERR, "cachetier_cb: could not save key %.*s for tier %s: %s",
-                (int) keylen, key, buf_cstring(mr->buf), cyrusdb_strerror(r));
+                (int) keylen, key, buf_cstring(rock->buf), cyrusdb_strerror(r));
```
brong, line 644: heh - this syslog still has the same bug that the previous version had - it's printing the key buffer AFTER it truncates it back to just the prefix.  Not a regression, but still wrong.

brong (approved):
Yep, this looks good.

(nice PR number by the way...)

imap/search_xapian.c
rsto: right! fixed that

imap/relocate_by_id.c
rsto: True, but traversing the config saves us figuring out the partition names and their respective searchpartition config keys. Of course if you want, we can change the current scheme.

PR #3503, opened 2021-05-26
jmap_mailbox: prefer to report changes in updated if modseqs match

rsto:
Mailbox renames result in two mboxlist entries with the same uniqueid and modseq, one of them being deleted. Until now, Mailbox/changes either reports this mailbox id in `updated` or `destroyed`, depending on mboxlist traversal order. This patch changes this to always prefer reporting equal modseq changes in `updated` over `created` or `destroyed`.

Tested in https://github.com/cyrusimap/cassandane/commit/08d5b4227964c43b5986a7e6e2dbfe4c2d34acf1

### DIFF
--- a/imap/jmap_mailbox.c
+++ b/imap/jmap_mailbox.c
@@ -4158,6 +4158,19 @@ static int _mbox_changes_cb(const mbentry_t *mbentry, void *rock)
         *(data->only_counts_changed) = 0;
     }
 
+    /* Determine where to report that update. Note that we even report
+     * hidden mailboxes in order to allow clients remove unshared and
+     * deleted mailboxes */
+    json_t *dest = NULL;
+    if (mbentry->createdmodseq <= data->since_modseq) {
+        if ((mbentry->mbtype & MBTYPE_DELETED) ||
+                !jmap_hasrights_mbentry(req, mbentry, JACL_LOOKUP)) {
+            dest = data->destroyed;
+        }
+        else dest = data->updated;
+    }
+    else dest = data->created;
+
     /* Is this a more recent update for an id that we have already seen?
      * (check all three) */
     json_t *old[3];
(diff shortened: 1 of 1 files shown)


imap/jmap_mailbox.c
```diff
@@ -4158,6 +4158,20 @@ static int _mbox_changes_cb(const mbentry_t *mbentry, void *rock)
         *(data->only_counts_changed) = 0;
     }
 
+    /* Determine where to report that update. Note that we even report
+     * hidden mailboxes in order to allow clients remove unshared ando
```
ksmurchison, line 4162: Small typo "ando"

imap/jmap_mailbox.c
```diff
@@ -4158,6 +4158,20 @@ static int _mbox_changes_cb(const mbentry_t *mbentry, void *rock)
...
+     * deleted mailboxes */
+    json_t *dest = NULL;
+    if ((mbentry->mbtype & MBTYPE_DELETED) || !jmap_hasrights_mbentry(req, mbentry, JACL_LOOKUP)) {
+        if (mbentry->createdmodseq <= data->since_modseq)
+            dest = data->destroyed;
+    } else {
+        if (mbentry->createdmodseq <= data->since_modseq)
```
ksmurchison, line 4169: Nit: Can't this if/else be moved up to be else if/else ?

ksmurchison (approved):
Other than 2 nits, looks sane to me

imap/jmap_mailbox.c
rsto: fixed

imap/jmap_mailbox.c
rsto: thanks. I simplified that if expression.

ksmurchison (approved):

brong (commented):
This looks good - though hopefully eventually this will all be looking at the "I" keys in the new style mailboxes.db, and hence we can tell if the mailbox still exists or not from there!

PR #3505, opened 2021-05-26
jmap_contact.c: strip weird Apple (localization?) wrapper from labels

ksmurchison:
and consolidate address, email, phones gathering code into a single for loop

### DIFF
--- a/imap/jmap_contact.c
+++ b/imap/jmap_contact.c
@@ -1549,180 +1549,157 @@ static json_t *jmap_contact_from_vcard(const char *userid,
     json_object_set_new(obj, "jobTitle", jmap_utf8string(item));
 
     json_t *adr = json_array();
-
-    for (entry = card->properties; entry; entry = entry->next) {
-        if (strcasecmp(entry->name, "adr")) continue;
-        json_t *item = json_object();
-
-        /* XXX - type and label */
-        const strarray_t *a = entry->v.values;
-
-        const struct vparse_param *param;
-        const char *type = "other";
-        const char *label = NULL;
-        if (entry->group) {
-            label = hash_lookup(entry->group, &labels);
-        }
-        for (param = entry->params; param; param = param->next) {
-            if (!strcasecmp(param->name, "type")) {
-                if (!strcasecmp(param->value, "home")) {
-                    type = "home";
-                }
-                else if (!strcasecmp(param->value, "work")) {
-                    type = "work";
-                }
-                else if (!strcasecmp(param->value, "billing")) {
-                    type = "billing";
-                }
-                else if (!strcasecmp(param->value, "postal")) {
-                    type = "postal";
-                }
-            }
-            else if (!strcasecmp(param->name, "label")) {
-                label = param->value;
-            }
-        }
-        json_object_set_new(item, "type", json_string(type));
-        json_object_set_new(item, "label", label ? json_string(label) : json_null());
-
-        const char *pobox = strarray_safenth(a, 0);
-        const char *extended = strarray_safenth(a, 1);
-        const char *street = strarray_safenth(a, 2);
-        buf_reset(&buf);
-        if (*pobox) {
-            buf_appendcstr(&buf, pobox);
-            if (extended || street) buf_putc(&buf, '\n');
-        }
-        if (*extended) {
-            buf_appendcstr(&buf, extended);
-            if (street) buf_putc(&buf, '\n');
-        }
-        if (*street) {
-            buf_appendcstr(&buf, street);
-        }
-
-        json_object_set_new(item, "street",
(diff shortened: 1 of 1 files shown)


imap/jmap_contact.c
```diff
@@ -1549,180 +1549,157 @@ static json_t *jmap_contact_from_vcard(const char *userid,
...
         const char *label = NULL;
+        size_t label_len = 0;
+
+        /* Apple label */
         if (entry->group) {
             label = hash_lookup(entry->group, &labels);
+            label_len = strlen(label);
+
+            /* Check and adjust for weird (localized?) labels */
+            if (label_len > 8 && !strncmp(label, "_$!<", 4)) {
```
rsto, line 1568: I don't know how these weird labels look like (is there a test for it?). But the pointer arithmetic here looks confusing. Why is label increment by 4, but its length decrement by 8? Is there some hidden assumption?

Alternatively: couldn't the code just skip any non-alphanumeric characters at the start of a label?

imap/jmap_contact.c
```diff
@@ -1782,6 +1753,18 @@ static json_t *jmap_contact_from_vcard(const char *userid,
         }
     }
 
+    if (defaultEmailIndex < 0)
```
rsto, line 1756: If defaultEmailIndex is reset to 0 here, why first initialize it to -1?

imap/jmap_contact.c
```diff
@@ -1782,6 +1753,18 @@ static json_t *jmap_contact_from_vcard(const char *userid,
...
+        defaultEmailIndex = 0;
+    int i, size = json_array_size(emails);
+    for (i = 0; i < size; i++) {
+        json_t *item = json_array_get(emails, i);
+        json_object_set_new(item, "isDefault",
+                            i == defaultEmailIndex ? json_true() : json_false());
```
rsto, line 1762: There always is exactly one defaultEmailIndex, so couldn't this loop be replaced with

`json_object_set_new(json_array_get(emails, defaultEmailIndex), "isDefault", json_true())`

and set "isDefault" false earlier when the email JSON object is created?

(even if there is not a single email, it is safe to call array_get and set_new on NULL values)

rsto (changes requested):
I found the pointer arithmetic confusing, but I might not know the full story.

imap/jmap_contact.c
ksmurchison: The Apple labels look something like _$!<Foo>!$_

imap/jmap_contact.c
ksmurchison: The logic is directly from the original code, but changing the variable name from defaultIndex to defaultEmailIndex

imap/jmap_contact.c
ksmurchison: Yes, we can make that change.  This code is a direct copy of the original.

rsto: Yeah, let's use the opportunity to clean this up.

imap/jmap_contact.c
rsto: So should we just skip special chars at the start of the label? And decrement label_len by the count of skipped characters?

ksmurchison: Damn it, GitHub tried to translate the special chars that I typed.  The label starts with _$!< and ends with >!$_, which is where the 8 comes from

ksmurchison: GitHub still stripping the underscores

imap/jmap_contact.c
ksmurchison: I think the logic is to only flag the first PREF that we see as the default, in case there are more than one.  If we know that if we initialize to zero and blindly set defaultIndex each time we see PREF, then we will get the last rather than the first.

I don't know if it makes any difference.

rsto: Right! I did not think of that. So let's keep it initialized to -1.

imap/jmap_contact.c
rsto: Today I learned that Github does some funky stuff when displaying code snippets.

rsto (approved):
Looks good to me

imap/jmap_contact.c
elliefm: If you're commenting in the web interface, you can make it format it weird stuff properly (skipping the automatic markdown italics/bold/etc) by wrapping it in backticks -- so `_$!<` and `>!$_` -- and you can check the Preview tab before submitting to make sure it understands you correctly.

But if you're commenting by replying to the notification email, it's anyone's guess what it'll do.  I think backticks will still work the same way, in the same way as _wrapping something in underscores italicises it_ etc.  But without the Preview button, you don't get an opportunity to check that it understood you correctly before you commit to posting it like that.

PR #3508, opened 2021-05-31
backups: use xsyslog for IOERRORs

elliefm:
As usual, please both read and build test this.  This is the old replication-backups stuff, not the JMAP one.


rsto (commented):
Looks good to me. Builds and tests cleanly.

PR #3514, opened 2021-06-03
jmap_backup.c: when restoring drafts track actual Message-ID rather than conversation ID

ksmurchison:
This fixes the problem where drafts in a thread wouldn't be restored.  Using CID exclusively was faulty logic.

Tested with https://github.com/cyrusimap/cassandane/pull/128


rsto (approved):
Looks good to me.

PR #3515, opened 2021-06-04
Possible fix for failing SearchSquat.simple test on master

elliefm:
The test was broken by 5f6bc5eaaeca1fcc65fb3f55352a21a2a5c097ce, which among other things removed an `SEOP_MATCH:` case from `search_build_query()`.  I don't understand why it was removed, maybe it was accidental?  Restoring it seems to fix the broken test, while not breaking anything else.


rsto (approved):
Thanks a lot @elliefm. This must have sneaked in when I ported experimental search code into the caching branch.

PR #3516, opened 2021-06-04
JMAP: enforce unique uid in Contact and CalendarEvent /set{create}

rsto:
Tested in https://github.com/cyrusimap/cassandane/pull/129


ksmurchison (commented):
Just curious: is there a reason why we xstrdup() the uid before looking it up in the DAV DB?

Otherwise, looks good.

rsto:
Assuming that the uid is not a duplicate we have to strdup it, and so we don't have to think much of if uid got duped or not  before we attempt to free it.

ksmurchison (approved):

PR #3517, opened 2021-06-04
http_ws: don't assert transaction buffer zero length for logging

rsto:
Sending an invalid JMAP Request payload leaves string data in the transaction buffer. Do not assert its length to be zero, but rather reset the contents before logging.

### DIFF
--- a/imap/http_ws.c
+++ b/imap/http_ws.c
@@ -560,7 +560,7 @@ static void on_msg_recv_cb(wslay_event_context_ptr ev,
             struct iovec iov[2];
             int niov = 0;
 
-            assert(!buf_len(&txn->buf));
+            buf_reset(&txn->buf);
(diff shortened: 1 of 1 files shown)


ksmurchison (commented):
The assert() is there because txn->buf is used in multiple places and we don't want to stomp on data that is needed elsewhere.  I'd prefer that the previous user of the buffer do the reset.  Does that make sense or is it complete nonsense?

rsto:
I usually also follow this approach. In the particular case we then need to find out where the buffer is not reset. I'll get back to this later.

ksmurchison:
@rsto Do you have a sample request that triggers this?

rsto:
I think just sending a JMAP request without the `@type: Request` entry is enough

PR #3521, opened 2021-06-09
Jmap cache one at a time

brong:
This changes the _next functions to only return a single item at a time, and will only process exactly as many records as needed to fill the requested window.


rsto (approved):
Looks good to me. Let's see if the function calls are causing overhead.

PR #3522, opened 2021-06-09
conversations: cache altrep (uniqueid or name) in the conversations object

brong:
This seems like a much nicer central cache, and it means we can always get the name or the uniqueid and we don't need to prework if not asked for it.

### DIFF
--- a/imap/conversations.c
+++ b/imap/conversations.c
@@ -285,22 +285,52 @@ EXPORTED int conversations_num_folders(struct conversations_state *state)
     return strarray_size(state->folders);
 }
 
-EXPORTED const char* conversations_folder_name(struct conversations_state *state,
-                                               int foldernum)
+EXPORTED const char *conversations_folder_mboxname(const struct conversations_state *state,
+                                                   int foldernum)
 {
-    const char *folder = strarray_safenth(state->folders, foldernum);
-    if (state->folders_byname) {
-        return folder;
-    }
-    else {
-        static char fname[MAX_MAILBOX_NAME+1];
-        mbentry_t *mbentry = NULL;
+    const char *val = strarray_nth(state->folders, foldernum);
+    if (state->folders_byname)
+        return val;
 
-        mboxlist_lookup_by_uniqueid(folder, &mbentry, NULL);
-        strlcpy(fname, mbentry ? mbentry->name : "", sizeof(fname));
+    // make it writeable even though const, because we're fiddling cache
+    struct conversations_state *backdoor = (struct conversations_state *)state;
+    if (!backdoor->altnames) backdoor->altnames = strarray_new();
+    const char *res = strarray_nth(backdoor->altnames, foldernum);
+    if (!res) {
+        if (!val) return NULL;
+        // cache lookup
+        mbentry_t *mbentry = NULL;
+        int r = mboxlist_lookup_by_uniqueid(val, &mbentry, NULL);
+        if (r || !mbentry) return NULL;
+        strarray_set(backdoor->altnames, foldernum, mbentry->name);
         mboxlist_entry_free(&mbentry);
-        return fname;
+        res = strarray_nth(backdoor->altnames, foldernum);
     }
+    return res;
+}
+
+EXPORTED const char *conversations_folder_uniqueid(const struct conversations_state *state,
+                                                   int foldernum)
+{
+    const char *val = strarray_nth(state->folders, foldernum);
+    if (!state->folders_byname)
+        return val;
+
+    // make it writeable even though const, because we're fiddling cache
+    struct conversations_state *backdoor = (struct conversations_state *)state;
+    if (!backdoor->altnames) backdoor->altnames = strarray_new();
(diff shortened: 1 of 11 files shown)


brong:
Looking buggy with the xapian stuff :(  Oh well - I'll look more later.

imap/conversations.h
```diff
@@ -97,6 +97,7 @@ struct conversations_state {
     char *annotmboxname;
     strarray_t *counted_flags;
     strarray_t *folders;
+    strarray_t *altnames;  // cache the alternative mboxname or uniqueid
```
rsto, line 100: nit: might want to rename this to altrep

imap/conversations.c
```diff
@@ -285,22 +285,52 @@ EXPORTED int conversations_num_folders(struct conversations_state *state)
...
+    const char *val = strarray_nth(state->folders, foldernum);
+    if (state->folders_byname)
+        return val;
+
+    // make it writeable even though const, because we're fiddling cache
+    struct conversations_state *backdoor = (struct conversations_state *)state;
+    if (!backdoor->altnames) backdoor->altnames = strarray_new();
+    const char *res = strarray_nth(backdoor->altnames, foldernum);
+    if (!res) {
+        if (!val) return NULL;
```
rsto, line 300: nit: could merge this line with the above as `if (!res && val)`

imap/conversations.h
```diff
@@ -139,8 +140,8 @@ struct conv_folder {
 #define CONV_GUIDREC_BYNAME_VERSION 0x1   // last folders byname version
 
 struct conv_guidrec {
+    const struct conversations_state *cstate;  // this conversationsdb!
     const char      *guidrep; // [MESSAGE_GUID_SIZE*2], hex-encoded
-    const char      *mailbox;       // if version >= 2 mboxid, else mboxname
```
rsto, line 143: ahh, sanity!

imap/conversations.h
```diff
@@ -269,6 +269,19 @@ extern int conversations_iterate_searchset(struct conversations_state *state,
...
+/* lookup the matching name or uniqueid */
+#define conv_guidrec_mboxname(rec) conversations_folder_mboxname((rec)->cstate, (rec)->foldernum)
+#define conv_guidrec_uniqueid(rec) conversations_folder_uniqueid((rec)->cstate, (rec)->foldernum)
+#define conv_guidrec_mbentry(rec, mbentryp) ( \
+  ((rec)->version > CONV_GUIDREC_BYNAME_VERSION) ? \
+   mboxlist_lookup_by_uniqueid(conv_guidrec_uniqueid(rec), mbentryp, NULL) : \
+   mboxlist_lookup(conv_guidrec_mboxname(rec), mbentryp, NULL) \
+ )
+#define conv_guidrec_mboxcmp(rec, mbox) ( \
+  ((rec)->version > CONV_GUIDREC_BYNAME_VERSION) ? \
```
rsto, line 281: these makes the macros arguments potentially be evaluated twice

rsto (changes requested):
Looking forward to see this land. The macro definitions look suspicious to me.

imap/conversations.h
brong: done.

imap/conversations.h
brong: yeah, all these similar things do.  We never call them with an expression, and I'm not sure how to make them safe against that while still returning a single value...

https://stackoverflow.com/questions/2679182/have-macro-return-a-value

Without using either an inline function or a GCC extension...

imap/conversations.c
```diff
@@ -285,22 +285,56 @@ EXPORTED int conversations_num_folders(struct conversations_state *state)
...
+        if (r || !mbentry) return NULL;
+        strarray_set(backdoor->altrep, foldernum, mbentry->name);
+        mboxlist_entry_free(&mbentry);
+        res = strarray_nth(backdoor->altrep, foldernum);
+    }
+    return res;
+}
 
-        mboxlist_lookup_by_uniqueid(folder, &mbentry, NULL);
-        strlcpy(fname, mbentry ? mbentry->name : "", sizeof(fname));
```
wolfsage, line 300: This method seems like it always used to return a string or an empty string, never NULL (unless ` const char *folder = strarray_safenth(state->folders, foldernum);` could have been null?).

This new version returns NULL which you've checked for in one place, but there are others that need checking too in imap/jmap_mail.c just above it I think.

Should it instead return "" like it used to or do we always need to guard the responses from this with NULL checks?

brong: I'd prefer to guard with NULL checks - because "" is clearly not a thing.

imap/conversations.c
brong: yeah true - or we could just put a `if (!val) return NULL;` way up top, because we don't even need to look for res - it's not going to be possible to resolve!

imap/conversations.c
brong: I have done a commit that puts those checks in!

PR #3524, opened 2021-06-10
jmap_mail: fix 'hasKeyword:$seen' queries for shared mailboxes

rsto:
This patch adds a new `seen` search attribute which correctly determines if to read the `\Seen` flag from the index record or seen database. The Email/query `hasKeyword:$seen` filter is updated to make use of this new attribute.

The implementation is somewhat lacking in that it opens the seen database for each inspected mailbox. This unnecessarily incurs CRC check costs. However, the current search attribute design does not allow for stateful attributes across a whole query. We might need to come up with such, if the CRC checks are proofing to be to costly. 

Tested in https://github.com/cyrusimap/cassandane/pull/131

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -1414,13 +1414,16 @@ static void _email_search_type(search_expr_t *parent, const char *s, strarray_t 
     strarray_fini(&types);
 }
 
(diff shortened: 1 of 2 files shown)


brong (approved):
Yep, this looks great.  It only adds expense for mailboxes without internal seen anyway, which is rare.  I like it.

PR #3529, opened 2021-06-17
UUID legacy paths

brong:
This fixed MOST of the legacy paths handling.

There's still a serious bug with user_deletedata, where it doesn't know the correct paths any more because we've already nuked the mailboxes.db entry.  In general, we're going to have to change all the underlying APIs so we can get a clean userid -> path mapping cheaply and reliably.  Possibly even store the user's uniqueid into every mailbox header file.

### DIFF
--- a/imap/append.c
+++ b/imap/append.c
@@ -885,9 +885,7 @@ static int findstage_cb(const conv_guidrec_t *rec, void *vrock)
 
     if (!strcmp(rock->partition, mbentry->partition)) {
         struct stat sbuf;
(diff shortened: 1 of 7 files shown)


ksmurchison (approved):
Looks good to me, especially if it works

brong:
@ksmurchison one more commit - turns out that intermediates are now getting a partition in them, and that's breaking the logic that says they don't exist when trying to open them!

ksmurchison (approved):

PR #3530, opened 2021-06-17
ical_support.c: add missing END:VCALENDAR

ksmurchison:
Fixes issue with events created/sent by Exchange


rsto (approved):

brong:
@ksmurchison maybe we should try to parse first, and only if the parse fails then check for the missing string, append, try parsing again.

ksmurchison:
@brong Updated to only append missing END:CALENDAR is the initial parse fails

rsto (approved):
LGTM

PR #3532, opened 2021-06-22
New mboxlist_create() api

ksmurchison:
@brong Just making sure that we're on the same page before I really dig into this.  mboxlist_createmailbox_full() will eventually become just mboxlist_createmailbox() and all callers will have to be changed similar to what I did with mboxlist_createmailbox_opts()  and mboxlist_createsync()

### DIFF
--- a/imap/http_dav_sharing.c
+++ b/imap/http_dav_sharing.c
@@ -352,9 +352,10 @@ static int lookup_notify_collection(const char *userid, mbentry_t **mbentry)
             goto done;
         }
 
-        if (*mbentry) free((*mbentry)->name);
-        else *mbentry = mboxlist_entry_create();
+        mboxlist_entry_free(mbentry);
+        *mbentry = mboxlist_entry_create();
         (*mbentry)->name = xstrdup(notifyname);
+        (*mbentry)->mbtype = MBTYPE_COLLECTION;
     }
 
   done:
@@ -379,9 +380,9 @@ static int _create_notify_collection(const char *userid, struct mailbox **mailbo
             goto done;
         }
 
-        r = mboxlist_createmailbox(mbentry->name, MBTYPE_COLLECTION,
-                                   NULL, 1 /* admin */, userid, NULL,
-                                   0, 0, 0, 0, mailbox);
+        r = mboxlist_createmailbox(mbentry, 0/*options*/, 0/*highestmodseq*/,
+                                   1/*isadmin*/, userid, NULL/*authstate*/,
+                                   0/*flags*/, mailbox);
         /* we lost the race, that's OK */
         if (r == IMAP_MAILBOX_LOCKED) r = 0;
         if (r) syslog(LOG_ERR, "IOERROR: failed to create %s (%s)",
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -1861,22 +1861,14 @@ EXPORTED int mboxlist_createmailbox_opts(const char *name, int mbtype,
         }
     }
 
-    const mbentry_t mbentry = {
-        (char *) name,
-        NULL, // ext_name
-        0,    // mtime
-        uidvalidity,
-        createdmodseq,
-        0,    // foldermodseq
-        mbtype,
-        (char *) partition,
-        NULL, // server
-        NULL, // acl
-        (char *) uniqueid,
-        NULL, // root_mailboxid
-        NULL, // legacy_specialuse
-        PTRARRAY_INITIALIZER
-    };
+    mbentry_t mbentry = MBENTRY_INITIALIZER;
+
+    mbentry.name = (char *) name;
+    mbentry.uidvalidity = uidvalidity;
+    mbentry.createdmodseq = createdmodseq;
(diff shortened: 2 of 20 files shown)


imap/mboxlist.c
```diff
@@ -242,6 +242,9 @@ static struct dlist *mboxlist_entry_dlist(const char *dbname,
     if (mbentry->uniqueid)
         dlist_setatom(dl, "I", mbentry->uniqueid);
 
+    if (mbentry->root_uid)
```
brong, line 245: I know it's kind of obnoxiously long, but I'd avoid `uid` and either say "root_uniqueid" or "root_mailboxid" with an end-goal of renaming every "uniqueid" to "mailboxid", the RFC8474 blessed name.

imap/mboxlist.c
```diff
@@ -1849,11 +1861,26 @@ EXPORTED int mboxlist_createmailbox_opts(const char *name, int mbtype,
         }
     }
 
-    r = mboxlist_createmailbox_full(name, mbtype, partition,
-                                    isadmin, userid, auth_state,
-                                    options, uidvalidity, createdmodseq, 0, 0, NULL,
-                                    uniqueid, localonly,
-                                    forceuser, dbonly, 0, &mailbox);
+    const mbentry_t mbentry = {
```
brong, line 1864: Oh god.  Really?  I'd be tempted to go wtih mbentry = MBENTRY_INITIALIZER; then explicitly set the fields.  It's fewer lines and more "mbentry gets changed later" safe.

brong (changes requested):
This is definitely the kind of thing I'm thinking of :)  Looks good modulo the couple of comments.

ksmurchison:
Specialized variants have been replaced by _full().  Will replace all other calls next.

imap/http_dav_sharing.c
```diff
@@ -352,9 +352,10 @@ static int lookup_notify_collection(const char *userid, mbentry_t **mbentry)
             goto done;
         }
 
-        if (*mbentry) free((*mbentry)->name);
-        else *mbentry = mboxlist_entry_create();
+        mboxlist_entry_free(mbentry);
+        *mbentry = mboxlist_entry_create();
         (*mbentry)->name = xstrdup(notifyname);
+        (*mbentry)->mbtype = MBTYPE_COLLECTION;
     }
```
brong, line 359: This is a slight behaviour change - we would have retained all the fields in the mbentry before if one was passed in, except for the 'name'.

But then - I checked the callsites, and they ALL make sure it's NULL before this is called, so that behaviour was never used.  You could even just `assert(!*mbentry)` to be really sure that the behaviour stays the same!

imap/mboxlist.c
```diff
@@ -1770,129 +1813,22 @@ static int mboxlist_createmailbox_full(const char *mboxname, int mbtype,
...
-                           int localonly, int forceuser, int dbonly,
-                           int notify, struct mailbox **mailboxptr)
+EXPORTED int mboxlist_createmailboxlock(const mbentry_t *mbentry,
+                                        unsigned options, modseq_t highestmodseq,
+                                        unsigned isadmin, const char *userid,
+                                        const struct auth_state *auth_state,
+                                        unsigned flags, struct mailbox **mboxptr)
 {
-    struct mboxlock *namespacelock = mboxname_usernamespacelock(name);
+    struct mboxlock *namespacelock = mboxname_usernamespacelock(mbentry->name);
```
brong, line 1822: I expect we'll want to do mbentry_usernamespacelock(mbentry) at some point :)  And lock by uniqueid.

brong (approved):
Yeah, this looks great.  Good first steps along they way to a saner API :)

ksmurchison:
@brong Do we want to fold mboxlist_createmailboxlock() into _createmailbox() with a new MBOXLIST_CREATE_NAMELOCK flag?

brong:
> @brong Do we want to fold mboxlist_createmailboxlock() into _createmailbox() with a new MBOXLIST_CREATE_NAMELOCK flag?

I'd say "not yet" - there's a bunch of other functions that do the same thing with wrapping a lock, and this is consistent with that.  We want to keep some level of focus on the main goal :)

ksmurchison:
@brong new option to ctl_mboxlist to add root mailbox ids to children

brong (approved):
This all looks great, putting my approval on it :)

brong:
Hmm this is super-clashy with jmap-calendars-01!

ksmurchison:
@brong Do we still need/want the 'R' field in mailboxes.db entries?  I guess I should audit our code and see if/where it might save a lookup or two.

brong:
> @brong Do we still need/want the 'R' field in mailboxes.db entries? I guess I should audit our code and see if/where it might save a lookup or two.

@ksmurchison - I would probably say let's drop it for now, and bring it back later if we need.  Simple is good.

ksmurchison:
I have gone ahead and removed the root_mailboxid code

brong (commented):
Yep, this all looks good - it's a good step along the way to tidying up these APIs!

PR #3534, opened 2021-06-23
Fix --disable-sieve --enable-jmap build

### LINKED ISSUE #3527: ./configure --without-sieve --with-jmap fails
When cyrus is built --without-sieve --with-jmap, the file imap/sieve_err.h is not generated.  imap/jmap_vacation.c imports sieve/sieve_interface.h and thus also sieve/sieve_err.h.  Perhaps jmap_vacation shall be build only, when sieve is enabled.

elliefm:
Make would fail in imap/jmap_vacation.c and imap/sievedir.c when configured with --disable-sieve and --enable-jmap, because they were pulling in sieve headers that would try to include sieve_err.h, which hadn't been generated.

It looks like both of these source files are already conditionalised to handle sieve being unavailable, so I've just done the same for the sieve header includes.

Needed to update some Cassandane tests that would crash out during setup when timsieved didn't exist, those fixes are already merged.  Seems to pass everything else.

Fixes #3527

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1082,7 +1082,7 @@ imap_libcyrus_imap_la_SOURCES = \
 	imap/sequence.c \
 	imap/sequence.h \
 	imap/setproctitle.c \
-        imap/sievedir.c \
+	imap/sievedir.c \
 	imap/sievedir.h \
 	imap/spool.c \
 	imap/spool.h \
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):
Works for me in all 4 combinations of JMAP/Sieve (with https://github.com/cyrusimap/cyrus-imapd/commit/32b60da32da9f61f54432320d00483c8436da40e)

PR #3536, opened 2021-06-25
docs: remove references to freenode

elliefm:
Since we don't use it anymore, stop saying we do.

Will need to pick this back to all the old-stable branches too, for the website.


elliefm:
Also fixed up the README.md build status buttons from .org => .com and 3.0 => 3.4, since I was in here anyway

ksmurchison (approved):

elliefm:
Merged and picked back to all the branches that the website builder uses.  2.5 was quite different so I just did it all again instead of figuring out the conflicts/renames/etc.

PR #3537, opened 2021-06-29
jmap_mail: use custom heuristic for Email.hasAttachment property

rsto:
This patch replaces the current logic to set the `hasAttachment` property with a Cyrus-custom heuristic.

The JMAP spec recommends the `Email.hasAttachment` property to be set if its `attachments` property includes at least one bodypart. Cyrus dissects body parts in `textBody`, `htmlBody` and `attachments` according to the reference algorithm in the [JMAP Email spec](https://jmap.io/spec-mail.html#properties-of-the-email-object), but this may miss a couple of cases which we expect to be attachments.

Tested in https://github.com/cyrusimap/cassandane/pull/132


ksmurchison (approved):
Looks good to me

PR #3538, opened 2021-06-29
mbpath: free string returned by user_hash_meta after use

rsto:
Fixes a memory leak in mbpath. In a perfect world, jansson would provide us with an API to transfer ownership of string values to their JSON type. But [this section](https://github.com/akheron/jansson/blob/master/src/value.c#L747) in the upstream source suggests we won't get that. Defining our own `json_stringm` in json_support.h would allow us to fake this API, but this might mislead callers to assume no allocations are made.


ksmurchison (approved):
Good catch.  Straight forward fix

PR #3539, opened 2021-06-29
Contact avatar mediatype

ksmurchison:
Better handling of missing mediatype on vCard PHOTO property

### DIFF
--- a/imap/vcard_support.c
+++ b/imap/vcard_support.c
@@ -126,19 +126,39 @@ EXPORTED struct vparse_card *record_to_vcard(struct mailbox *mailbox,
     return vcard;
 }
 
+struct image_magic {
+    size_t offset;
+    size_t len;
+    uint8_t data[8];
+};
+
 static const struct image_signature {
-    const uint8_t magic[8];
-    size_t magiclen;
     const char *mediatype;
+    struct image_magic magic[2];
 } image_signatures[] = {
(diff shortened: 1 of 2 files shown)


imap/vcard_support.c
```diff
@@ -126,6 +126,21 @@ EXPORTED struct vparse_card *record_to_vcard(struct mailbox *mailbox,
     return vcard;
 }
 
+static const struct image_signature {
+    const uint8_t magic[8];
+    size_t magiclen;
+    const char *mediatype;
+} image_signatures[] = {
```
rsto, line 133: Wikipedia lists a couple of additional magic bytes for the respective types:

# JPG
```
FF D8 FF DB
FF D8 FF E0 00 10 4A 46
49 46 00 01
FF D8 FF EE
FF D8 FF E1 ?? ?? 45 78
69 66 00 00
```

where `??` is the data size.

# Google WebP format (`image/webp`)
```
52 49 46 46 ?? ?? ?? ??
57 45 42 50
```

where `??` is the file size.

rsto (approved):
Nice! I found some additional magic markers that we might want to support. In any case, this looks good to me.

ksmurchison:
Added support for 'split' magic numbers and a couple more formats.  I may add support for SVG as well

rsto (approved):
Looks good to me.

rsto (approved):

PR #3540, opened 2021-06-30
mboxlist: allow RACL entries where 'l' permission exists

### LINKED ISSUE #3528: "Lookup" ACL stopped working in 3.4
After a recent update to Cyrus imapd 3.4.1 the "List" commands fail to list folders with 'l' (lookup), but without 'r' (read) ACLs.
As I understand they should list them, as described in the "Access Control Lists Rights Reference", as well as in rfc4314.

This behaviour shows up only if "reverseacls" is enabled.

The responsible change is this:

```
--- cyrus-imapd-3.2.6/imap/mboxlist.c   2021-03-09 05:32:13.000000000 +0000
+++ cyrus-imapd-3.4.1/imap/mboxlist.c.orig      2021-05-10 06:34:24.000000000 +0000
@@ -693,11 +698,13 @@
     }
 }
 
-static int user_is_in(const strarray_t *aclbits, const char *user)
+static int user_can_read(const strarray_t *aclbits, const char *user)
 {
     int i;
     if (!aclbits) return 0;
     for (i = 0; i+1 < strarray_size(aclbits); i+=2) {
+        // skip ACLs without read bit
+        if (!strchr(strarray_nth(aclbits, i+1), 'r')) continue;
         if (!strcmp(strarray_nth(aclbits, i), user)) return 1;
     }
     return 0;
@@ -737,9 +744,10 @@
     if (oldusers) {
         for (i = 0; i+1 < strarray_size(oldusers); i+=2) {
             const char *acluser = strarray_nth(oldusers, i);
+            if (!strchr(strarray_nth(oldusers, i+1), 'r')) continue;
             if (!strcmpsafe(userid, acluser)) continue;
             if (strarray_find(admins, acluser, 0) >= 0) continue;
-            if (user_is_in(newusers, acluser)) continue;
+            if (user_can_read(newusers, acluser)) continue;
             mboxlist_racl_key(!!userid, acluser, name, &buf);
             r = cyrusdb_delete(mbdb, buf.s, buf.len, txn, /*force*/1);
             if (r) goto done;
@@ -750,9 +758,10 @@
     if (newusers) {
         for (i = 0; i+1 < strarray_size(newusers); i+=2) {
             const char *acluser = strarray_nth(newusers, i);
+            if (!strchr(strarray_nth(newusers, i+1), 'r')) continue;
             if (!strcmpsafe(userid, acluser)) continue;
             if (strarray_find(admins, acluser, 0) >= 0) continue;
-            if (user_is_in(oldusers, acluser)) continue;
+            if (user_can_read(oldusers, acluser)) continue;
             mboxlist_racl_key(!!userid, acluser, name, &buf);
             r = cyrusdb_store(mbdb, buf.s, buf.len, "", 0, txn);
             if (r) goto done;
```

elliefm:
Since ff1442271ae144fb84fcf58f0e258f54d8c11164 we have only stored RACL entries where the ACL contained the 'r' permission.  IMAP LIST needs to be able to return mailboxes where the user has 'l' but not 'r', but when reverseacls is enabled, it can't see them at all because there's no RACL entry.

This fix addresses the issue by creating RACL entries if either the 'r' or 'l' permission is present.

Added a lot of tests for this specific case: https://github.com/cyrusimap/cassandane/compare/master...elliefm:v34/3528-list-lookup-only

It doesn't seem to break anything else, by which I mean, none of our existing tests break with this change.  But if you can think of something that might break with this, we should add a test for it!

I thought RACL and RUNQ were versioned, and thought I might need to bump the RACL version as part of this, but it looks like that was reverted by 3d0abf7e0c.

Fixes #3528, needs to be picked back to 3.4 once merged.

### DIFF
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -922,8 +922,8 @@ static int user_can_read(const strarray_t *aclbits, const char *user)
     int i;
     if (!aclbits) return 0;
     for (i = 0; i+1 < strarray_size(aclbits); i+=2) {
-        // skip ACLs without read bit
-        if (!strchr(strarray_nth(aclbits, i+1), 'r')) continue;
+        // skip ACLs with neither read nor lookup bit
+        if (!strpbrk(strarray_nth(aclbits, i+1), "lr")) continue;
         if (!strcmp(strarray_nth(aclbits, i), user)) return 1;
     }
     return 0;
@@ -967,7 +967,7 @@ static int mboxlist_update_racl(const char *dbname, const mbentry_t *oldmbentry,
     if (oldusers) {
         for (i = 0; i+1 < strarray_size(oldusers); i+=2) {
             const char *acluser = strarray_nth(oldusers, i);
-            if (!strchr(strarray_nth(oldusers, i+1), 'r')) continue;
+            if (!strpbrk(strarray_nth(oldusers, i+1), "lr")) continue;
             if (!strcmpsafe(userid, acluser)) continue;
             if (strarray_find(admins, acluser, 0) >= 0) continue;
             if (user_can_read(newusers, acluser)) continue;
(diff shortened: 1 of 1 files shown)


brong (approved):
Yeah, this looks great.  Indeed, 'l' is almost more the permissions that's needed for the reverse ACL to make sense!  But either is good too.  Thanks.

This was done in the first place to stop the `anyone p` ACL generating zillions of spurious records.

elliefm:
This is now on 3.4 too, and bumps RACL_VERSION there.

PR #3541, opened 2021-07-02
jmap_mail: fix erroneous use of conversation_folder_number

rsto:


ksmurchison (approved):
Do we even have to make such an obvious fix a PR?

PR #3543, opened 2021-07-05
caldav_sched: pass the acting userid through to the imip sender

brong:
This came up today with calendar invitations on shared calendars being rejected because they were authenticated as the person creating the request rather than the calendar owner.

Given how calendar requests are created, I think the correct answer is to just authenticate as the calendar owner.  You can only send as them if you have the `8` ACL on their Outbox, but after that the email is coming from the calendar owner for all intents and purposes, so the system should authenticate as them.

### DIFF
--- a/imap/http_caldav_sched.h
+++ b/imap/http_caldav_sched.h
@@ -169,9 +169,9 @@ extern int isched_send(struct caldav_sched_param *sparam, const char *recipient,
 
 extern int sched_busytime_query(struct transaction_t *txn,
                                 struct mime_type_t *mime, icalcomponent *comp);
-extern void sched_request(const char *onuserid, const strarray_t *schedule_addresses, const char *organizer,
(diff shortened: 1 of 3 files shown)


imap/http_caldav_sched.h
```diff
@@ -173,7 +173,8 @@ extern void sched_request(const char *onuserid, const strarray_t *schedule_addre
                           icalcomponent *oldical, icalcomponent *newical);
 extern void sched_reply(const char *onuserid, const strarray_t *schedule_addresses,
```
rsto, line 174: Nit: the `onuserid` argument name got renamed to `userid` in the function implementation.

rsto (approved):
This looks good to me. I checked the JMAP calendars code and it looks sane to me: we only do `sched_reply` and `sched_request` calls there, which pass the authenticated req->userid as userid argument.

PR #3547, opened 2021-07-09
Sieve jmapquery: plumb conversations_state down to the callback

ksmurchison:
@rsto It turns out that lmtpd already opened conv.db so I just plumbed it down to the jmapquery() callback

### DIFF
--- a/imap/jmap_sieve.c
+++ b/imap/jmap_sieve.c
@@ -1599,14 +1599,22 @@ static int getmetadata(void *sc, const char *extname,
     return r ? 0 : 1;
 }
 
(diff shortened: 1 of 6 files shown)


rsto (approved):
Looks good to me. Thanks. I suggest we do not land this directly on main branch. Rather, I'll cherry-pick it on my feature branch where I'll build someInThread filtering in Sieve.

ksmurchison:
Yeah, I had to intentions of landing this.  It was just a start point for you to do the thread work.

rsto (commented):
Not going to deploy, as we'll work on it on another feature branch.

PR #3548, opened 2021-07-10
http_jmap.c: initial implementation of proxying JMAP (based on authenticated user)

ksmurchison:
Each JMAP method should probably be proxied by accountId, but this is a start.

### DIFF
--- a/imap/http_jmap.c
+++ b/imap/http_jmap.c
@@ -89,6 +89,7 @@ static void jmap_reset(void);
 static void jmap_shutdown(void);
 
 /* HTTP method handlers */
+static int meth_connect_jmap(struct transaction_t *txn, void *params);
 static int meth_get(struct transaction_t *txn, void *params);
 static int meth_options_jmap(struct transaction_t *txn, void *params);
 static int meth_post(struct transaction_t *txn, void *params);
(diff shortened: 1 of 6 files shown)


rsto (commented):
The code changes look sane to me. Two questions, though:

- Will this work with web sockets, too?
- Has this been tested? I guess we don't have automated tests for proxying JMAP, so I assume you did manual testing?

ksmurchison:
This currently will not work with WebSockets.
I only tested it by hand.

rsto:
OK, I can't tell if not supporting WebSockets is good enough, but the code at hand looks sane. I'll approve.

rsto (approved):
Looks good to me.

ksmurchison:
Updated to proxy JMAP/WS over HTTP/1.1.  JMAP/WS over HTTP/2 still to be done

elliefm:
I'd like to see at least a basic cassandane test for this.  We already have the Murder test suite which knows how to set up a frontend/mupdate server and a couple of backends, so the infrastructure work is already done.

A copy of Murder.frontend_commands that uses JMAP instead of IMAP would be a great start, just so we can establish the baseline of "stuff is proxied to the right place, responses come back, actions take effect, and we'll notice if it breaks later"

imap/http_jmap.c
```diff
@@ -327,6 +329,29 @@ static int jmap_parse_path(struct transaction_t *txn)
...
 
+    if (config_mupdate_server && !config_getstring(IMAPOPT_PROXYSERVERS)) {
+        /* Locate the authenticated user's INBOX */
+        char *inbox = mboxname_user_mbox(httpd_userid, NULL);
+        int r = proxy_mlookup(inbox, &tgt->mbentry, NULL, NULL);
+
+        free(inbox);
+
+        if (r) {
+            syslog(LOG_ERR, "mlookup(%s) failed: %s", inbox, error_message(r));
```
rsto, line 340: could use xsyslog here

rsto (changes requested):
The code looks good me, except for using syslog. I agree with @elliefm that we it needs at least a basic Cassandane test. Also, I think this feature warrants an entry in the changes directory.

PR #3550, opened 2021-07-12
jmap_ical: use lookup table for VEVENT recurrence exceptions

rsto:
Rewrites the overrides_from_ical function to read overrides from a pre-filled list, rather than looking them up within the function.

Passes all JMAP Email and Calendar tests.


ksmurchison (approved):
This looks sane to me

PR #3553, opened 2021-07-14
preload string comparisons that will the same for every message

brong:
This speeds up index_msgdata_load for the very common request which the Fastmail UI uses to load every single mailbox!

```
        "sort": [
          {
            "property": "snoozedUntil",
            "isAscending": false,
            "mailboxId": "567d9ea54215e1e9"
          },
          {
            "isAscending": false,
            "property": "receivedAt"
          }
        ],
```

Instead of having to do a strcmp of the mailbox_uniqueid against the parameter for every single record in the mailbox, it's done just once at the top of the loop.

I also picked out the other two cases where we do string comparisons that aren't going to change for every message in the mailbox, because why not!

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -6121,6 +6121,7 @@ MsgData **index_msgdata_load(struct index_state *state,
     struct index_record record;
     struct conversations_state *cstate = NULL;
     conversation_t conv = CONVERSATION_INIT;
+    int *preload = NULL;
 
     if (!n) return NULL;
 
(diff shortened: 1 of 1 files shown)


imap/index.c
```diff
@@ -6132,6 +6133,32 @@ MsgData **index_msgdata_load(struct index_state *state,
     if (found_anchor)
         *found_anchor = 0;
 
+    /* set mailbox level states */
+    for (j = 0; sortcrit[j].key; j++); // count how many we need
+    preload = xzmalloc(j * sizeof(int));
```
rsto, line 6138: j can be zero, which should be fine with malloc(0) returning either NULL or a free-able pointer. But I remember we had issues with freeing a zero size malloc result (and fixed that during a late evening hack session at a bar in Zurich :))

imap/index.c
```diff
@@ -6132,6 +6133,32 @@ MsgData **index_msgdata_load(struct index_state *state,
...
+#ifdef WITH_JMAP
+        case SORT_SNOOZEDUNTIL:
+#endif
+            preload[j] = !strcmpnull(mailbox_uniqueid(mailbox), sortcrit[j].args.mailbox.id);
+            break;
+
+        case SORT_HASCONVFLAG:
+            preload[j] = -1;
+            if (!cstate) cstate = conversations_get_mbox(index_mboxname(state));
+            assert(cstate);
```
rsto, line 6152: Are we sure that this code is not reachable for builds without conversations? The alternative would be to gracefully not preload if cstate is null.

rsto (commented):
This looks like a good win for performance. Just have two remarks that I'd like to see addressed.

imap/index.c
brong: A little later (inside the loop):

```
            if ((label == SORT_HASCONVFLAG || label == SORT_CONVMODSEQ ||
                label == SORT_CONVEXISTS || label == SORT_CONVSIZE) && !did_conv) {
                if (!cstate) cstate = conversations_get_mbox(index_mboxname(state));
                assert(cstate);
```

So if so, this isn't a regression.

imap/index.c
brong: Good point, fixed in the updated commit.

brong:
@rsto I have resolved your two comments!  One with a fix, another with a justification ;)

imap/index.c
rsto: ack

rsto (approved):
Looks good to me

PR #3554, opened 2021-07-14
jmap_ical: hash previous VEVENT versions by uid and recurid

rsto:
Similar to what we did in https://github.com/cyrusimap/cyrus-imapd/pull/3550 this patch organizes the old main VEVENT and overrides in a hash table for efficient lookup. Previously, this code could suffer from performance degradation for large VCALENDAR data.

This is an internal refactor, it passes all calendar regression tests, including Valgrind.

### DIFF
--- a/imap/jmap_ical.c
+++ b/imap/jmap_ical.c
@@ -110,11 +110,119 @@ static int is_valid_jmapid(const char *s)
     return i > 0 && s[i] == '\0';
 }
 
+/* A helper structure to organize iCalendar components */
+struct icalcomps {
+    hash_table by_uidrecurid; /* pointer to icalcomponent */
+    hash_table by_uid; /* ptrarray_t of icalcomponent */
+    struct buf buf;
+};
+
+#define ICALCOMPS_INITIALIZER { \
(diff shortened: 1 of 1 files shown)


imap/jmap_ical.c
```diff
@@ -110,11 +110,125 @@ static int is_valid_jmapid(const char *s)
...
+static void icalcomps_init(struct icalcomps *comps, icalcomponent *ical)
+{
+    // hash table requires us to preallocate its size
+    size_t ncomps = 0;
+    icalcomponent *comp;
+    for (comp = icalcomponent_get_first_component(ical, ICAL_VEVENT_COMPONENT);
+         comp;
+         comp = icalcomponent_get_next_component(ical, ICAL_VEVENT_COMPONENT)) {
+        ncomps++;
+    }
```
ksmurchison, line 151: This loop can probably be replaced with icalcomponent_count_components()

ksmurchison (approved):
Other than the one small change for counting components, this looks good.

imap/jmap_ical.c
rsto: great, done

PR #3557, opened 2021-07-16
cunit: clamp snprintf arguments to their expected format width

rsto:
Gcc's stack protector picked this up in the buf cunit tests.


brong (approved):
Gotta love the signed sprintf!  Good find.

PR #3558, opened 2021-07-17
Iter seqset

brong:
No idea why I wrote this (back in 2017), but it's simple and passes all the tests, so cleaning this up from my laptop!

I guess in theory we can make a more efficient iterator that skips ahead to each next UID when we change to new database structures.


imap/mailbox.c
```diff
@@ -7628,6 +7629,12 @@ EXPORTED void mailbox_iter_startuid(struct mailbox_iter *iter, uint32_t uid)
     iter->recno = uid ? mailbox_finduid(mailbox, uid-1) : 0;
 }
 
+EXPORTED void mailbox_iter_uidset(struct mailbox_iter *iter, struct seqset *seq)
```
rsto, line 7632: It's be nicer to user `const seqset *` to make ownership of seq clear. But this will require to update the signature of `seqset_ismember`. The the implementation of ismember could do the usual backdoor cast to non-const.

rsto (approved):
Looks good to me, but we can use the occasion to sanitize the sequence set API.
Too bad we can't cunit test mailbox.c. Luckily, the SearchFuzzy cassandane tests cover this codepath.

imap/mailbox.c
brong: Yeah, that's a great idea.  I'll do that, because I love this API trick.

brong: Done and also now 'seqset_t' everywhere.  Also it's `lib/seqset.[ch]` and such.

rsto (approved):
Looks good to me. Builds and tests locally and passes JMAPEmail and SearchFuzzy tests.

PR #3563, opened 2021-07-20
Lockfixes

brong:
This fixes a bunch of lock inversion, mostly by making less happen within the locked sections!

(it's also a step towards reduction of mboxname lookups in the counters code - but mostly hoisting the name lookup outside the locked section).


ksmurchison (approved):
This all looks good to me

PR #3564, opened 2021-07-20
Proxy JMAP requests in a Murder

ksmurchison:
[edit from elliefm: There's tests for this in cyrusimap/cassandane/pull/151]

### DIFF
--- a/imap/http_proxy.c
+++ b/imap/http_proxy.c
@@ -1329,7 +1329,9 @@ EXPORTED int http_proxy_h2_connect(struct backend *be, struct transaction_t *txn
     http_err = http_status_to_code(code);
     send_response(txn, http_err, resp_hdrs, &resp_body.payload);
 
-  done:
+    log_proxy_request(http_err, txn, resp_hdrs, &resp_body);
+
+ done:
     buf_free(&resp_body.payload);
     if (resp_hdrs) spool_free_hdrcache(resp_hdrs);
 
--- a/imap/http_proxy.c
+++ b/imap/http_proxy.c
@@ -48,6 +48,8 @@
 #endif
 #include <assert.h>
 #include <ctype.h>
+#include <errno.h>
+#include <sysexits.h>
 #include <syslog.h>
 #include <sasl/sasl.h>
 #include <sasl/saslutil.h>
@@ -599,6 +601,9 @@ static void write_cachehdr(const char *name, const char *contents,
 
     /* Ignore private headers in our cache */
     if (name[0] == ':') return;
+
+    if (txn->meth == METH_CONNECT && !strcasecmp(name, "Sec-WebSocket-Accept"))
+        return;
 
     for (hdr = hop_by_hop; *hdr && strcasecmp(name, *hdr); hdr++);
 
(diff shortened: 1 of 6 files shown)


ksmurchison:
I think this is finally done (other than Cass test).  It properly proxys JMAP requests and JMAP/WS connections over both HTTP/1.1 and HTTP/2

imap/http_proxy.c
```diff
@@ -1243,3 +1261,212 @@ EXPORTED int http_proxy_copy(struct backend *src_be, struct backend *dest_be,
...
+    protgroup_reset(protin);
+    protgroup_insert(protin, clientin);
+
+    for (n = 0; n < ptrarray_size(pipes); n++) {
+        txn = ptrarray_nth(pipes, n);
+        protgroup_insert(protin, txn->be->in);
+
+        if (txn->flags.ver < VER_2) {
+            /* stream client input directly to server */
+            serverout = txn->be->out;
```
elliefm, line 1374: This feels like, if there's more than one txn in `pipes`, it could end up replacing `serverout` a bunch of times.  Does that need to be guarded against?

imap/http_proxy.c
```diff
@@ -1243,3 +1261,212 @@ EXPORTED int http_proxy_copy(struct backend *src_be, struct backend *dest_be,
...
+    }
+
+    n = prot_select(protin, PROT_NO_FD, &protout, NULL,
+                    timeout_sec ? &timeout : NULL);
+    if (n == -1 && errno != EINTR) {
+        syslog(LOG_ERR, "prot_select() failed in proxy_check_input(): %m");
+        fatal("prot_select() failed in proxy_check_input()", EX_TEMPFAIL);
+    }
+
+    if (n && protout) {
```
elliefm, line 1385: Does this need to be `if (n > 0 && protout) {`?  If the prot_select above is interrupted then `n` could be -1 without fataling, which is truthy, and then the loop just below will iterate backwards from -1 to INT_MIN, underflow to INT_MAX, and keep looping until it reaches 0 (unless it crashes earlier than that due to an out of bounds access somewhere)

elliefm (commented):
I don't know the bigger picture well enough to spot potential abstraction problems, but taken line by line it seems okay except for a couple of possible edge cases

imap/http_proxy.c
ksmurchison: Good question, but for HTTP/1.1 there can only be one txn at a time due to the nature of the protocol.  HTTP/2+ is multiplexed, but we can't stream directly to the server here.  We need to return to the caller so that we can parse the framing and determine what stream we are dealing with.  I'll add a comment to the code to explain this.

imap/http_proxy.c
ksmurchison: Good catch.  This code is a specialized copy of proxy.c:proxy_check_input() which does the same thing.  However, it looks like protout will be NULL if prot_select( ) returns -1

rsto (commented):
This looks sane to me, but bear in mind that I haven't grokked all of the murder code, yet. I'd like to approve when we see a Cassandane test working.
I can confirm that my local build does not yield any compiler errors. However, the whole current Cassandane Murder test suite fails with valgrind errors.

ksmurchison:
I've looked into adding a JMAP Murder test to Cass and it looks non-trivial. I might need some help in getting the infrastructure built

elliefm:
> I've looked into adding a JMAP Murder test to Cass and it looks non-trivial. I might need some help in getting the infrastructure built, mainly getting a second httpd instance running.

Do you have a working local setup?  If so, can you send me your imapd.conf files from the frontend, backend, and mupdate server?  (If the frontend and mupdate server are the same server, that's actually great, because that's the setup Cassandane uses.)

I figured out the long way how to get it working for IMAP back when I wrote the Murder suite, but I'm not sure what the config needs to look like to have it proxy HTTP as well -- but I know the infrastructure fairly well, so once I know what the config needs to look like I think I can make Cassandane do it.

ksmurchison:
I have frontend and mupdate as the same server.  Other than that, its the same setup as IMAP in a Murder, with httpd on frontend and backend (with JMAP module enabled)

elliefm:
The trouble is, I don't understand the IMAP murder config quite well enough to know how to generalise it, I just know what I needed to do to wrestle it into working 😅   So it'd really help to see examples of working HTTP, cause e.g. I'm not sure where I'll need to configure the port numbers?

Though I suppose you might be using port 80/443 by default and not actually have port numbers explicitly configured, whereas Cassandane needs to sticks everything on :910x.  Hmm.

ksmurchison:
Fair enough.  I will cobble together a minimal config for you

elliefm:
Great, thanks!

ksmurchison:
Actually, do any of the tests in Murder.pm actually proxy a client request from the frontend to the backend?  They all appear to talk to a frontend or backend directly.

elliefm:
No idea really.  The ones that talk to the frontend expect the frontend to do whatever it needs to do to service the request.  If none of them exercise anything that needs to be proxied, then we need to add some tests for things that do get proxied.

I might be misunderstanding how murder proxying works.  I wrote these tests mostly in the dark, so I don't expect them to be either correct or thorough as they are.  They're better than the nothing we previously had, but they still need a lot more work.

ksmurchison:
@elliefm So, I took a look at the Cass Murder config and I think we can use the "include port in servername" trick for JMAP, but this trick can only work for one protocol at a time.  So, we wouldn't be able to have an IMAP service run on the backend, just HTTP.  This is because any service running on the proxy (IMAP, JMAP, POP) will try to connect to the port encoded in servername.

elliefm:
Ahh interesting.  So I think, then, that instead of having just `$want->{murder}` as the only flag, we'd want to have separate `$want->{imapmurder}` and `$want->{jmapmurder}` (and corresponding `:IMAPMurder` and `:JMAPMurder` magics to set those flags), and then have _create_instances set up whichever one is wanted, and die hard if both are wanted.  And then the existing murder tests would need to be switched to be explicitly IMAP, and then we'd be able to add new tests that are explicitly JMAP.  I don't know if we'd want to split the IMAP and JMAP murder tests into two separate suites, or keep them in the Murder suite and just use the :IMAPMurder / :JMAPMurder magics to configure them correctly.

ksmurchison:
This sounds like a reasonable plan to me.

elliefm:
@ksmurchison A first stab at what we've discussed about testing this is in https://github.com/cyrusimap/cassandane/pull/151

elliefm:
Hmm.

Running `MurderJMAP.frontend_commands` with this branch but rebased onto current master, under valgrind, finds a memory leak in the frontend httpd.

Since the source for the build I'm testing doesn't exist anywhere as its own entity, the line numbers quoted below might not correspond with line numbers in your local files.  So you'll probably wanna run this yourself to get real line numbers...

```
logs[1470] Valgrind errors from file /dev/shm/cass/0301420304/vglogs/httpd.9891
logs[1470] ==9891== 32 bytes in 1 blocks are definitely lost in loss record 79 o
logs[1470] ==9891==    at 0x483577F: malloc (vg_replace_malloc.c:299)
logs[1470] ==9891==    by 0x4CC7477: xmalloc (xmalloc.c:55)
logs[1470] ==9891==    by 0x4C0829F: prot_addwaitevent (prot.c:548)
logs[1470] ==9891==    by 0x1AD8CA: proxy_findserver (proxy.c:185)
logs[1470] ==9891==    by 0x1CE7A0: meth_post (http_jmap.c:481)
logs[1470] ==9891==    by 0x1A15E6: process_request (httpd.c:1953)
logs[1470] ==9891==    by 0x1A1BEF: http1_input (httpd.c:2091)
logs[1470] ==9891==    by 0x1A1F5D: cmdloop (httpd.c:2197)
logs[1470] ==9891==    by 0x19EE34: service_main (httpd.c:1054)
logs[1470] ==9891==    by 0x1CD8EB: main (service.c:647)
logs[1470] ==9891== 
```

Looks like the thing that's being leaked is a `struct prot_waitevent` in `(struct backend *)->timeout`.  It's usually cleaned up by `proxy_downserver()` specifically, but is notably left alone by `backend_disconnect()`.  So I wonder if we're winding up on a codepath where backend_disconnect is called rather than proxy_downserver.

Though, I've just noticed that `cmdloop (httpd.c:2197)` is:

```
        if (ret == HTTP_SHUTDOWN) {
            syslog(LOG_WARNING,
                   "Shutdown file: \"%s\", closing connection", conn->close_str);
===>        shut_down(0);
        }
```

So we're actually in the process of shutting down when we allocate this new prot_waitevent that then gets leaked? That's weird.

It's also weird that I don't see that "Shutdown file: ..., closing connection" log line anywhere.  There is a "Received bad client magic byte string, closing connection" from the backend1 httpd, and an "end of file reached, closing connection" from the frontend httpd.

For what it's worth it's the frontend httpd that the leak is being reported in.

I'm really confused by the valgrind backtrace telling me we came from the "Shutdown file:" block, but that log line not appearing in syslog.  So I don't trust the report, but I don't know how to proceed further.

elliefm:
> It's also weird that I don't see that "Shutdown file: ..., closing connection" log line anywhere. There is a "Received bad client magic byte string, closing connection" from the backend1 httpd, and an "end of file reached, closing connection" from the frontend httpd.

Oh, I changed branch while briefly distracted with something else and forgot to change back, so disregard everything I said about being in a shut down, I had the wrong line numbers myself! 😅 

`cmdloop (httpd.c:2197)` is actually this, which is just ordinary input, nothing to do with shut_down.  That makes more sense!

```
            else if (!ret) {
                /* HTTP/1.x request */
===>            http1_input(&txn);
            }
```

PR #3565, opened 2021-07-21
proxy: don't cache the same backend multiple times

elliefm:
This fixes a double-free crash on cleanup because the same backend was being cached multiple times, introduced in #3560.

I _think_ this is the only place this needs to be changed, but I'm not entirely certain, so even though this is only a one-line fix, please look closely 😅

### DIFF
--- a/imap/proxy.c
+++ b/imap/proxy.c
@@ -192,7 +192,7 @@ EXPORTED struct backend * proxy_findserver(const char *server,          /* hostn
     ret->inbox = inbox;
 
     /* insert server in list of cache connections */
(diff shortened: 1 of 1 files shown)


elliefm:
This was found by the Cassandane Murder tests erroring due to core dumps. The fix makes the Murder tests pass, but my full run hasn't finished yet, will update when it does...

brong (approved):
Yeah, that looks great.  Definitely the right thing!

brong (approved):

elliefm:
Full Cassandane run looks good 👍

PR #3569, opened 2021-07-21
squatter: lock conversations db in mailbox_begin to fix lock ordering

brong:
This should hopefully fix the remaining deadlock error I'm seeing in the logs!  Basically, it tries to lock conversations AFTER the xapianactive file is locked, but that reverses the ordering from when we do searches - so instead, always take the lock if we're XAPINDEXed.  Which should hopefully mean that is_indexed doesn't deadlock.  Yay.

[ not to be merged:  this is exploratory work ]

### DIFF
--- a/imap/search_xapian.c
+++ b/imap/search_xapian.c
@@ -2641,12 +2641,26 @@ static int begin_mailbox_update(search_text_receiver_t *rx,
     char *userid = NULL;
 
     tr->flags = flags;
+    tr->mode = (flags & (SEARCH_UPDATE_XAPINDEXED|SEARCH_UPDATE_AUDIT)) ?
+        XAPIAN_DBW_XAPINDEXED : XAPIAN_DBW_CONVINDEXED;
 
     /* not an indexable mailbox, fine - return a code to avoid
      * trying to index each message as well */
     if (!fname) {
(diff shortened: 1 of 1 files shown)


imap/search_xapian.c
```diff
@@ -2658,6 +2660,18 @@ static int begin_mailbox_update(search_text_receiver_t *rx,
         goto out;
     }
 
+    /* make sure the conversations are open before we start indexing
+     * to avoid deadlocking against the search state */
+    if (tr->mode == XAPIAN_DBW_CONVINDEXED) {
+        struct conversations_state *cstate = mailbox_get_cstate(tr->super.mailbox);
```
rsto, line 2666: This should get the cstate from `mailbox` instead of `tr->super.mailbox`. The latter only is set at the end of this function.

rsto (changes requested):

imap/search_xapian.c
brong: Yeah!  Duh.  Fixed.

rsto (approved):

rsto:
Turns out I already approved that in 2021! And now that I re-reviewed I still do.

PR #3570, opened 2021-07-22
search_xapian: more deadlock prevention

brong:
Yeah, another one - turns out the read cases can be called with or without a conversations.db lock too.


brong:
Actually, I'm merging these commits to the other PR.

PR #3571, opened 2021-07-22
Mboxlist root mailboxid

ksmurchison:
Add a 'R' (root mailboxid) field to mbentry records.

Also add 's' option to ctl_mboxlist to ass 'R' field to existing records

### DIFF
--- a/imap/ctl_mboxlist.c
+++ b/imap/ctl_mboxlist.c
@@ -96,6 +96,7 @@ enum mboxop { DUMP,
               CHECKPOINT,
               UNDUMP,
               VERIFY,
+              ADD_ROOTS,
               NONE };
 
 struct dumprock {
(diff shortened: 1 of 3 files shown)


imap/ctl_mboxlist.c
```diff
@@ -96,6 +96,7 @@ enum mboxop { DUMP,
               CHECKPOINT,
               UNDUMP,
               VERIFY,
+              ADD_ROOTS,
```
brong, line 99: This really does feel like what we need is a `$V` key or something in mailboxes.db, rather than using the `AS ` key as the upgrade indicator, and then we can confirm that roots have been added.

imap/ctl_mboxlist.c
```diff
@@ -882,6 +883,91 @@ static void do_verify(void)
...
+};
+
+static int addroots_cb(const mbentry_t *mbentry, void *rockp)
+{
+    mbname_t *mbname = mbname_from_intname(mbentry->name);
+    struct root_rock *rrock = rockp;
+    int is_root = 0, is_child = 0;
+
+    /* Skip DELETED.* mailboxes */
+    if (mbname_isdeleted(mbname)) {
```
brong, line 898: I think maybe we do want to add the INBOX root_mailboxid to DELETED.user.foo mailboxes as well, so we know to consider them in the same userid, since they're findable with mboxlist_usermboxtree and flags.

imap/ctl_mboxlist.c
```diff
@@ -882,6 +883,91 @@ static void do_verify(void)
...
+    }
+    else {
+        /* Shared mailbox */
+        const strarray_t *boxes = mbname_boxes(mbname);
+
+        if (strarray_size(boxes) == 1) {
+            /* Toplevel mailbox */
+            is_root = 1;
+        }
+        else if (rrock->mbname &&
```
brong, line 924: does this also need `&& !mbname_userid(rock->mbname)` for maximum safety?

imap/ctl_mboxlist.c
```diff
@@ -882,6 +883,91 @@ static void do_verify(void)
...
+
+    if (is_root) {
+        /* Switch to new root */
+        mbname_free(&rrock->mbname);
+        free(rrock->mboxid);
+
+        rrock->mbname = mbname;
+        rrock->mboxid = xstrdup(mbentry->uniqueid);
+    }
+    else if (is_child) {
```
brong, line 940: I believe we should also be setting root_mailboxid on the INBOX or top-level folder, so we can just use that field rather than having to interpret the lack of one as "this is the root".

Particularly since that's what we do further down when we maintain this field in the mailbox creation code.

imap/ctl_mboxlist.c
```diff
@@ -882,6 +883,91 @@ static void do_verify(void)
...
+
+    return 0;
+}
+
+static void do_addroots(int intermediary)
+{
+    struct root_rock rrock = { NULL, NULL };
+    int flags = MBOXTREE_TOMBSTONES;
+
+    if (intermediary) flags |= MBOXTREE_INTERMEDIATES;
```
brong, line 963: I think we should be always doing INTERMEDIATES here, while they still exist.  We don't want to leave them rootless.

brong (approved):
A bunch of little bits of edge-case feedback here, but it's looking like it'll be useful soon :)

The big question is whether we want to version the mailboxes.db with a single key rather than using the presence of `AS ` as the upgraded marker and needing yet another upgraded marker later.  This would allow us to rev the database with more improvements without needing this upgrade magic each time - just have `ctl_cyrusdb -r` check if it's current and upgrade it if not.

brong (changes requested):
Sorry I clicked the wrong button - I meant to request changes!  I think we do need to fix these few things before deploying

ksmurchison:
Superseded by #4838

PR #3572, opened 2021-07-23
imapd.c:cmd_sync…(): avoid double checking for imapd_userisadmin

dilyanpalauzov:
The cmd_syncapply(), cmd_syncget() and cmd_syncrestore() functions check, whether imapd_userisadmin, so the same check is not necessary, before these functions are called.


ksmurchison:
Superseded by #5039

PR #3573, opened 2021-07-23
jmap_mail: load the cache correctly

brong:
Turns out we were calculating how much MORE to cache rather than what to cache up to, but the "loadn" API takes "up to here", not "this much more".  Doh.

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -3858,24 +3858,29 @@ static void emailquery_cache_slice(struct emailquery *q,
                                    json_t **errp)
 {
     if (!qc->have_total) {
-        size_t n = SIZE_MAX;
+        /* conditions for doing a partial load:
+         * 1) we don't need the total
+         * 2) we don't need to find an anchor
+         * 3) the offset is from the start, not from the end
+         * 4) we have a limit, so we're not fetching everything
+         * 5) we don't need all in thread
+         * if all these conditions are met, then we can fill just to the
+         * end of the requested window */
         if (!q->super.calculate_total && !q->super.anchor &&
             q->super.position >= 0 && q->super.limit && !q->findallinthread) {
-            /* Lazily load at most n matches into cache */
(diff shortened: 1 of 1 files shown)


wolfsage:
I'll leave the code review to someone else - but for the commit message, can we include an explanation of what was wrong and how this fixes it?

Otherwise a year from now we'll just shake our heads at `jmap_mail: load the cache correctly` and wonder

Thanks!

brong:
> I'll leave the code review to someone else - but for the commit message, can we include an explanation of what was wrong and how this fixes it?
> 
> Otherwise a year from now we'll just shake our heads at `jmap_mail: load the cache correctly` and wonder
> 
> Thanks!

Yep, done!  I've updated the commit message and also fixed the flipped logic of my last minute rewrite from the one I'd tested by hand on my testbed without good variable names.

imap/jmap_mail.c
```diff
@@ -3858,24 +3858,29 @@ static void emailquery_cache_slice(struct emailquery *q,
...
-            if (lastpos > have_len) {
-                n = lastpos - have_len;
+            size_t have_len = q->collapse_threads ?
+                              qc->collapsed_len :
+                              qc->uncollapsed_len;
+            size_t need = q->super.position + q->super.limit;
+            // is the window already within the loaded section?
+            if (have_len < need) {
+                // fill just to the end of the window
+                emailquery_cache_loadn(q, qc, need);
```
rsto, line 3878: The `loadn` function is a misnomer I introduced, and now I can now how it is misleading. A better name might be `ensuren` or even `ensure`, like we do in the buf API.

rsto (approved):
Looks good to me, although we might want to rename `loadn`.

imap/jmap_mail.c
brong: I'll rename it!

rsto (approved):

PR #3574, opened 2021-07-26
jmap_mail: reset mboxname iterator in email_append

rsto:
Resets the mailbox name iterator when we deallocate its memory.


ksmurchison (approved):
Make sense

PR #3575, opened 2021-07-27
docs: replace FastMail with Fastmail

rjbs:
(except in changelog)


docsrc/overview/who_is_cyrus.rst
```diff
@@ -1,21 +1,20 @@
...
-**FastMail**
+**Fastmail**
 
-FastMail_ is a leading e-mail hosting provider that runs on Cyrus software.
+Fastmail_ is a leading e-mail hosting provider that runs on Cyrus software.
 
 *	Bron Gondwana
-*   Ellie Timoney
-*   Conrad Kleinespel
-*   Chris Davies
```
elliefm, line 17: There's some kinda whitespace mismatch here, does the Bron line have a tab in it?

elliefm: Scrolling further, the same comment applies to most/all the other bulleted lists.  I'd like to see it to be consistent, and I don't like the hard tab so I would lean toward replacing the old tabs rather than making the new lines match the old

elliefm (approved):
This looks good, thanks.

The `doc/` changes are correct; we don't* have generated files in version control, so the fact those files are committable as diffs rather than new files means they needed fixing too. :)

* mostly 😅

docsrc/overview/who_is_cyrus.rst
rjbs: I've replaced all the tabs with single spaces to match the majority of bullets.

Issue #3240, opened 2020-10-12
Seen/Unseen Flag not working correctly if sharedseen of a shared mailbox is not set to true


guimard:
From [Debian #972079](https://bugs.debian.org/972079):

> Dear Maintainer,
>
> User subscribed to a shared mailbox, which has sharedseen set to false, can not set Seen/Unseen as expected.
> Selecting messages as read or unread,to state of seen/unseen is not keeping as expected.No error/fail/warn messages in logs.
> User has acl all on shared mailbox. Seen/Unseen on a shared mailbox with sharedseen is set to true,is working as expected.
>
> Counter check: cyrus-imapd version 3.0.8 is working as expected with a shared mailbox with sharedseen set to false.

guimard:
Hi, this issue affects Cyrus-Imapd 3.2.x

guimard:
@elliefm: hi, did you see this issue ?

elliefm:
I've just tried reproducing the issue with the following Cassandane test: https://github.com/cyrusimap/cassandane/compare/master...elliefm:v32/3240-sharedseen-on-shared-mailbox

But the test passes okay on both 3.2 and master, so I can't see an obvious problem here...

> User subscribed to a shared mailbox

It would be useful to know whether this "shared mailbox" is a mailbox at the top level namespace (not belonging to any particular user), or whether it is a mailbox belonging to a second user that they have given the first user access to.

> User has acl all on shared mailbox

It would be useful to know exactly what the ACL is set to.  `all` is not a sensible ACL string (it means "admin and lookup only", and it might complain about `l` being set twice).  So I assume they mean "all the settings are enabled", and am testing with `lrswipkxtecdan`.  But it would help to know exactly what their ACL is, in case the difference is the problem.

tpayen:
Hi, it seems we encounter a similar problem with version 3.2 on Debian. We have recently updated 2 of our cyrusimap servers from 2.5 to 3.2.6 and our Thunderbird clients are loosing Seen/Unseen status when using shared mailboxes. 

We have activated  imap debug log to try to figure out the problem and it seems to come from the UID STORE command which is acting like SILENT option is always enable when using shared mailboxes.

For example in Cyrus 2.5, STORE command result was :
```
SendData: 30 uid store 492 -Flags (\Seen)
CreateNewLineFromSocket: * 1 FETCH (FLAGS () UID 492)
CreateNewLineFromSocket: 30 OK Completed
```

[...]

PR #3577, opened 2021-07-27
Fixing issue #3240 Seen/Unseen Flag not working correctly if sharedseen of a shared mailbox is not set to true

tpayen:
Following the issue #3240, when using \Seen flag with shared seen disable, modseq isn't updated so store command never return the fetch flags. So we force the update of modseq value by calling index_rewrite_record function

This is probably unoptimized patch as we can unify the call of index_rewrite_record with others functions

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -4708,6 +4708,7 @@ static int index_storeflag(struct index_state *state,
     int dirty = 0;
     modseq_t oldmodseq;
     struct index_map *im = &state->map[msgno-1];
+    struct index_record *record;
     int r;
 
     memset(modified_flags, 0, sizeof(struct index_modified_flags));
@@ -4739,6 +4740,9 @@ static int index_storeflag(struct index_state *state,
     uint32_t system_flags;
     uint32_t internal_flags;
     uint32_t user_flags[MAX_USER_FLAGS/32];
+    
+    r = msgrecord_get_index_record_rw(msgrec, &record);
+    if (r) return r;
 
     r = msgrecord_get_systemflags(msgrec, &system_flags);
     if (r) return r;
@@ -4870,6 +4874,8 @@ static int index_storeflag(struct index_state *state,
     r = msgrecord_set_internalflags(msgrec, internal_flags);
     if (r) return r;
     r = msgrecord_set_userflags(msgrec, user_flags);
+    if (r) return r;
+    r = index_rewrite_record(state, msgno, record, /*silent*/0);
     if (r) return r;
 
     /* if it's silent and unchanged, update the seen value, but
--- a/imap/index.c
+++ b/imap/index.c
@@ -4740,7 +4740,7 @@ static int index_storeflag(struct index_state *state,
     uint32_t system_flags;
     uint32_t internal_flags;
     uint32_t user_flags[MAX_USER_FLAGS/32];
-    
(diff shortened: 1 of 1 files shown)


elliefm:
Have confirmed by testing that this patch does seem to fix the problem, and doesn't seem to introduce any new ones (at least: not that our tests find), but I'm keeping the full discussion in the linked issue #3240 for now.

imap/index.c
```diff
@@ -4739,6 +4740,9 @@ static int index_storeflag(struct index_state *state,
     uint32_t system_flags;
     uint32_t internal_flags;
     uint32_t user_flags[MAX_USER_FLAGS/32];
+
```
elliefm, line 4743: nit: this line has trailing spaces

elliefm:
This looks okay, and my tests ([here](https://github.com/cyrusimap/cassandane/compare/master...elliefm:v32/3240-sharedseen-on-shared-mailbox)) confirm that it a) fixes the problem, and b) doesn't break anything else that we have a test for.  It also cherry picks cleanly onto the later branches, and a) and b) still apply for each.

But I don't know the msgrecord or index APIs in enough depth to tell whether this is the right way to do it, so can we get some experienced eyes on this please?

ksmurchison (commented):
This also looks ok to me, but I'll defer to @brong to determine if there is a better way to bump the modseq.

imap/index.c
tpayen: Cleaned with last commit

elliefm: Thanks!

brong (changes requested):
This was fixed slightly differently in later code.  This code will bounce the modseq twice, while moving the "get_index_record" and "rewrite_record" after the msgrecord_save() like the current code does will be slightly more efficient.

elliefm:
@brong

> This was fixed slightly differently in later code.

Looks like it wasn't fully, though!  The new `Flags.seen_sharedmb_nosharedseen` test in https://github.com/cyrusimap/cassandane/pull/142 (not merged yet) fails on current Cyrus master.

It fails at this line -- the \Seen flag is stored okay, but we expect an untagged response and don't get one:
https://github.com/cyrusimap/cassandane/pull/142/files#diff-2d74fb1901913e00a6db529589e7c4665a613b273867e331d1c814ad7e94df9eR299

I reckon this might be a case of "it's already fixed for user/ mailboxes, but not shared ones specifically" -- Cyrus master passes all the other Flags tests, even with the  updates to aggressively check for the expected untagged responses, suggesting that it is shared mailboxes specifically that are the problem here.

I'm not sure whether this means the modseq is not being bumped, or if it's just the untagged response that's missing.  I think I thought the two were inherently related but I don't really remember.  What do you think?

I guess we might have two things here, and one is already fixed on master, the other not.  Locally merging this PR on top of master allows the new test to pass, but we don't wanna bump the modseq twice...  guess the tests should also check for that!

elliefm:
Thanks for the ping, I've put this onto the agenda to discuss next time we meet.  Usually that would be next week, but I'm not currently certain if next week's call is going ahead, it might not be until the week after.

brong:
Since this fixes things reasonably in 3.2, I'm going to just go ahead and merge it with apologies for taking so long.

PR #3578, opened 2021-07-27
httpd.c: add precondition error to log line

ksmurchison:
This might aid in diagnosing client issues without having to enable telemetry

### DIFF
--- a/imap/http_dav.c
+++ b/imap/http_dav.c
@@ -451,6 +451,31 @@ static const struct precond_t {
     { "max-resource-size", NS_CARDDAV },
 };
 
+void dav_precond_as_string(struct buf *buf, struct error_t *err)
+{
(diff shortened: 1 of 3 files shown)


imap/http_dav.c
```diff
@@ -451,6 +451,31 @@ static const struct precond_t {
...
+{
+    const struct precond_t *precond = &preconds[err->precond];
+
+    switch (precond->ns) {
+    case NS_DAV:
+        buf_appendcstr(buf, "DAV:");
+        break;
+    case NS_CALDAV:
+        buf_appendcstr(buf, "CALDAV:");
+        break;
```
ksmurchison, line 464: Note that I've chosen to use these strings as namespaces rather than the actual urns for brevity

brong (approved):

PR #3580, opened 2021-07-28
http_h2.c: reverse end of headers logic

ksmurchison:
There are fewer instances to leave stream open than to close it so
changing the logic to reflect this seems to make sense


rsto (changes requested):
I hate to be that guy, but I can't determine what this patch wants to achieve. Could you please trim down the commit message to a single line and add clarifying info either in the commit message body, or in the PR description?

ksmurchison:
@rsto commit message changed

rsto (approved):
Looks good to me.

PR #3581, opened 2021-07-28
http_h2.c: also log stream-id for HTTP/1.1 -> HTTP/2 upgrade request

ksmurchison:


imap/http_h2.c
```diff
@@ -874,6 +874,11 @@ HIDDEN int http2_start_session(struct transaction_t *txn,
         txn->strm_ctx = strm;
         txn->flags.ver = VER_2;
         ptrarray_add(&txn->done_callbacks, &stream_free);
+
+        /* Tell syslog our stream-id */
+        buf_printf(&txn->buf, "%d", strm->id);
```
wolfsage, line 879: The rest of the code above uses buf instead of txn->buf, does that matter?

ksmurchison: It doesn't, but I should stay consistent

rsto (approved):
Looks good to me

PR #3589, opened 2021-08-02
HTTP transfer encoding fix

ksmurchison:
This fixes a crasher when compression is enabled and the client requests gzip transfer-encoding.

The only clients that I'm aware of that can/do request gzip transfer-encoding are curl and libwww-perl


rsto (approved):
This looks sane to me, but I can't run this locally ATM.

PR #3590, opened 2021-08-03
Improve the efficiency of annotations DB lookups

brong:
It's probably worth looking at the diffs independently here.  Two are of particular interest - one changes the _annotate_getdb API, and the other splits the annotatemore_findall API.

There's still more work to do, but this is a couple of big ones :)


ksmurchison (approved):
This looks good

PR #3591, opened 2021-08-03
UUID bugfixes

brong:
This is quite urgent!  Well, the two bugfixes are.  The boring syslog line fixer isn't.

This fixes two observed bugs when upgrading a server.

1) the name-history 'dbname' field was storing intname instead of dbname values.

2) when a DELETED mailbox was created by sync_server, it wasn't correctly finding that the INBOX was still legacy, because INBOX isn't a parent of the DELETED namespace.


ksmurchison (approved):
This all looks sane

PR #3596, opened 2021-08-04
lock user namespace while upgrading 

brong:
This should fix an issue we currently have where multiple processes try to upgrade the subscriptions DB at once.

By using the user_namespacelock, we make sure two processes don't upgrade at the same time!

We also check the xVERx key exactly rather than doing a foreach and hoping we never create an earlier key in the database, and check it again after taking the namelock to be sure that it hasn't been upgraded by a racing process.


ksmurchison (approved):
LGTM

PR #3597, opened 2021-08-04
http_caldav.c: skip "mailto:" when comparing ORGANIZERs in components

ksmurchison:
This fixes a bug where the ORGANIZER in the master component is "MAILTO:foo@example.com" and the ORGANIZER in an override is "mailto:foo@example.com"


rsto (approved):
Looks good to me

dilyanpalauzov:
This is more or less duplicate of https://github.com/cyrusimap/cyrus-imapd/pull/2723/files .  I think the email addresses shall also be compared in case-insensitive manner.

PR #3599, opened 2021-08-05
http_caldav.c: protect against empty ORGANIZER properties

ksmurchison:
At least one stupid client tries to store resources with an empty ORGANIZER property, which we need to protect against if libical was compiled to allow empty props (until libical is fixed to only allow empty text props)


rsto (approved):
LGTM. We really need strncasecmpsafe.

PR #3600, opened 2021-08-05
jmap_calendar: fix CalendarEvent.blobId property for non-uuid

rsto:
Fixes the CalendarEvent.blobId for users which did not get relocated to uuid. Passes all regression tests.


ksmurchison (approved):
Makes sense to me

Issue #3592, opened 2021-08-04
Include sessionid in cmdtimer log


robn:
If `cmdtimer:` logs included the current session ID, it would make some kinds of log search 0.1% easier for me!

PR #3602, opened 2021-08-06
Add sessionid field to cmdtimer: logs

elliefm:
This adds a `sessionid=<...>` field to the "cmdtimer:" log lines, which was requested by @robn in #3592.  I also switched it to xsyslog while I was in there anyway.

I was briefly confused as to why the `commandmintimer` option (which controls this behaviour) was doing its own value parsing, rather than being an OPT_DURATION, until I realised this option accepts a floating point value.  So I also updated the docs for the option to make this clear.

Added a basic test in: https://github.com/cyrusimap/cassandane/pull/144

Since this is also an xsyslog conversion, which we know are prone to subtle typos, please do a build test of this -- in case your compiler catches anything mine didn't.

Thanks!

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -2453,12 +2453,17 @@ static void cmdloop(void)
         if (commandmintimer && strcmp("idle", cmdname)) {
             double cmdtime, nettime;
             const char *mboxname = index_mboxname(imapd_index);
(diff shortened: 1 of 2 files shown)


brong (approved):
Looks great - builds and runs tests fine on my testbed.

PR #3603, opened 2021-08-06
xsyslog: don't log syserror/func fields for LOG_NOTICE/LOG_INFO

### LINKED ISSUE #3601: make xsyslog only include syserror when the log level is error/warning/debug
Our errno discipline isn't great yet and we sometimes leak ignored errors into log lines where it's not relevant.  If we only included it when it might be relevant, this would avoid confusing/misleading "errors" in what are supposed to be informational log lines.

elliefm:
This prevents confusion caused when automated diagnostic stuff (especially _wrong_ diagnostic stuff, like leaked errno values) winds up in informational log messages.

So far we've mostly used xsyslog for error conditions, but this makes it a little nicer to use for everything else too.

For whatever it's worth, the full set of syslog log levels is:

- LOG_EMERG      system is unusable
- LOG_ALERT      action must be taken immediately
- LOG_CRIT       critical conditions
- LOG_ERR        error conditions
- LOG_WARNING    warning conditions
- LOG_NOTICE     normal, but significant, condition
- LOG_INFO       informational message
- LOG_DEBUG      debug-level message

And I think we still want syserror=<> and func=<> for all of these, except LOG_NOTICE and LOG_INFO.

This also reverts a commit that was working around the problem by unconditionally resetting errno in one particular place.

Closes #3601

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -2099,6 +2099,8 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
 {
     static struct buf buf = BUF_INITIALIZER;
     int saved_errno = errno;
+    int want_diag = (LOG_PRI(priority) != LOG_NOTICE
+                     && LOG_PRI(priority) != LOG_INFO);
 
     buf_reset(&buf);
     buf_appendcstr(&buf, description);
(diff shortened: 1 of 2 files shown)


lib/util.c
```diff
@@ -2099,6 +2099,8 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
 {
     static struct buf buf = BUF_INITIALIZER;
     int saved_errno = errno;
+    int want_diag = (LOG_PRI(priority) != LOG_NOTICE
+                     && LOG_PRI(priority) != LOG_INFO);
```
rsto, line 2103: Nit: I think this is a rare occasion where the combining operator of a multi-line expression is put at the second line. I did a quick, non-exhaustive scan of the codebase and I only saw it a the first line (inluding util.c)

rsto (approved):
Looks great to me! Just found one nit.

lib/util.c
elliefm: If you want me to change it I will; but, I think this is the best way to do it, and while we're thinking about code style, it should be our style to do it this way, even though it hasn't consistently been so in the past.

Having the operator on the left, where it's aligned and easiest to see, makes the relationship between the surrounding lines a bit clearer at a glance, which saves some mental cycles for parsing the subexpressions.

For whatever it's worth, `ack '^\s*[&|]{2}'` finds hundreds of examples like this already present.  Though `ack '[&|]{2}\s*$'` finds thousands the other way, too!

rsto: I am used to the other notation, but I am not invested in either. I'd say leave as you prefer and we address this in our upcoming style guide.

PR #3605, opened 2021-08-09
imapd: don't check index if status on a different mailbox

brong:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1708981

The likelyhood that a client is waiting on this data and
won't issue a NOOP or otherwise check for it is much lower
than the likelyhood that the client won't think to look in
the status response to a different mailbox for an update to
the index state.

If we are doing a STATUS on the selected mailbox, we still
need to do the `index_check` though, because that will make
sure the data is up-to-date.

... this passes all the Cassandane tests (ImapTest, Idle and Expunge being the most interesting probably)

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -9326,8 +9326,6 @@ static void cmd_status(char *tag, char *name)
 
     /* local mailbox */
 
-    imapd_check(NULL, 0);
-
     c = parse_statusitems(&statusitems, &errstr);
     if (c == EOF) {
         prot_printf(imapd_out, "%s BAD %s\r\n", tag, errstr);
@@ -9351,6 +9349,10 @@ static void cmd_status(char *tag, char *name)
                 IMAP_PERMISSION_DENIED : IMAP_MAILBOX_NONEXISTENT;
         }
     }
+
+    // status of selected mailbox, we need to refresh
+    if (!strcmpsafe(mbentry->name, index_mboxname(imapd_index)))
+        imapd_check(NULL, 0);
 
     if (!r) r = imapd_statusdata(mbentry, statusitems, &sdata);
 


elliefm:
So we previously would do an index check (and therefore return unsolicited updates) on any STATUS command, regardless of the currently selected mailbox;

but this seems to be confusing to clients who have a mailbox selected and don't expect/ignore updates about other mailboxes;

and later, when they do actually ask for STATUS of those other mailboxes, there aren't any updates because we've already sent them (but they were ignored).

And so this patch changes STATUS to only return updates for the currently selected mailbox, not everything else, which means when they're explicitly asked for, they're ready to send (and not already-sent-and-ignored).

Am I understanding this correctly?

The code looks fine, just trying to figure out what it does! :)

brong:
> So we previously would do an index check (and therefore return unsolicited updates) on any STATUS command, regardless of the currently selected mailbox;
> 
> but this seems to be confusing to clients who have a mailbox selected and don't expect/ignore updates about other mailboxes;
> 
> and later, when they do actually ask for STATUS of those other mailboxes, there aren't any updates because we've already sent them (but they were ignored).
> 
> And so this patch changes STATUS to only return updates for the currently selected mailbox, not everything else, which means when they're explicitly asked for, they're ready to send (and not already-sent-and-ignored).
> 
> Am I understanding this correctly?
> 
> The code looks fine, just trying to figure out what it does! :)

Yep, precisely that.  We're actually "allowed" to send unsolicited updates at any time, and the client is supposed to update its state machine - but of course most clients aren't listening at any random time - so we generally only send them in response to a command by the client.

So this change will meant that we just don't send the unsolicited updates - which we're also allowed to NOT send them if we'd prefer.  Only NOOP requires us to check and send them.

https://datatracker.ietf.org/doc/html/rfc3501#section-6.1.2
```
      Since any command can return a status update as untagged data, the
      NOOP command can be used as a periodic poll for new messages or
      message status updates during a period of inactivity (this is the
      preferred method to do this).  The NOOP command can also be used
      to reset any inactivity autologout timer on the server.
```

elliefm (approved):

elliefm:
This ought to be backported to 3.4 once it's confirmed that it resolves the Thunderbird issue report.

I have stuff in flight on that branch at the moment, so I will take care of it myself after that stuff lands; please don't backport it in the meantime.

elliefm:
Will also need #3606

elliefm:
This has been cherry-picked to 3.4 and 3.2.

It didn't pick cleanly to 3.0 (at a glance: mbentry API changes) and I didn't bother going further

PR #3606, opened 2021-08-09
status: need to not dereference mbentry on error

brong:
This fixes a crash when you try to STATUS a folder that doesn't exist, introduced by 6a00f5b0b0402d7023bc7e5e6c0c49d23567c5a8


elliefm (approved):
Looks good to me (but so did the last one 😅 )

PR #3607, opened 2021-08-10
mboxlist: make space for a full 490 character folder name

brong:
This caused a problem for a user with a mailbox that's exactly 490 characters long!  The name got trimmed by one character by this function because `snprintf` needs the entire space PLUS an extra byte for the `\0`.  The size to `snprintf` should be the entire size of the buffer, not one fewer.

```
       The functions snprintf() and vsnprintf() write at most size bytes (including the terminating null byte ('\0')) to str.
```

This fixes the broken user sync.

### DIFF
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -590,7 +590,7 @@ static int mboxlist_parse_entry(mbentry_t **mbentryptr,
 
     if (name) {
       /* copy name */
-        snprintf(mboxname, MAX_MAILBOX_NAME+1, "%.*s",
+        snprintf(mboxname, sizeof(mboxname), "%.*s",
                  (int) (namelen ? namelen : strlen(name)), name);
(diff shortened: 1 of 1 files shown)


imap/mboxlist.c
```diff
@@ -590,7 +590,7 @@ static int mboxlist_parse_entry(mbentry_t **mbentryptr,
 
     if (name) {
       /* copy name */
-        snprintf(mboxname, MAX_MAILBOX_NAME, "%.*s",
+        snprintf(mboxname, MAX_MAILBOX_NAME+1, "%.*s",
```
elliefm, line 593: If this had been `snprintf(mboxname, sizeof(mboxname), ...` in the first place it never would have been a problem, so maybe the fix should be to use sizeof ;)

brong: Done!

elliefm (approved):

PR #3611, opened 2021-08-12
Quota rename

brong:
This fixes the issues identified by:  https://github.com/cyrusimap/cassandane/pull/146

There's also a bunch of other small improvements to the safety and behaviour of renames in uuidland, including:

* you can rename a subfolder, and it will still detect the right INBOX
* per-mailbox annotations won't get nerfed
* replication will detect and repair renamed folders so long as the change happened on the master first

### DIFF
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -170,10 +170,10 @@ EXPORTED void mboxlist_entry_free(mbentry_t **mbentryptr)
 
     free(mbentry->legacy_specialuse);
 
-    former_name_t *fname;
-    while ((fname = ptrarray_pop(&mbentry->name_history))) {
-        free(fname->name);
-        free(fname);
+    former_name_t *histitem;
+    while ((histitem = ptrarray_pop(&mbentry->name_history))) {
+        free(histitem->name);
+        free(histitem);
     }
     ptrarray_fini(&mbentry->name_history);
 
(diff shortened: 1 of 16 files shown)


imap/mboxname.c
```diff
@@ -2292,11 +2292,32 @@ EXPORTED char *mboxname_conf_getpath(const mbname_t *mbname, const char *suffix)
...
+            if (!r) {
+                int i;
+                for (i = 0; i < ptrarray_size(&mbentry_byid->name_history); i++) {
+                    const former_name_t *fname = ptrarray_nth(&mbentry_byid->name_history, i);
+                    char *olduserid = mboxname_to_userid(fname->name);
+                    char *oldinboxname = mboxname_user_mbox(olduserid, NULL);
+                    r = mboxlist_lookup_allow_all(oldinboxname, &mbentry, NULL);
+                    free(oldinboxname);
+                    free(olduserid);
+                    if (r != IMAP_MAILBOX_NONEXISTENT) break;
```
elliefm, line 2315: This line looks like it'll break out of the loop on success OR any other error, is that the intent?  If so it could use a comment, or maybe an explicit `if (r == 0 || r != IMAP_MAILBOX_NONEXISTENT) ...`, to make it clear to the reader that both cases have been considered and this choice is deliberate.

Or, I guess maybe mboxlist_lookup_allow_all only ever returns 0 or IMAP_MAILBOX_NONEXISTENT, so not being nonexistent means success?  Not sure, I haven't looked. Even if that is the case now, being explicit about the intent here will protect us in the future.

elliefm:
Most of annotate.testc is failing, I guess they're noticing the annotate_delete_mailbox change

imap/mboxname.c
ksmurchison: I think this entire chunk of code could be eliminated if we also use #3571

imap/mboxlist.c
```diff
@@ -1056,17 +1059,15 @@ static int mboxlist_update_entry(const char *name,
                 /* Existing mailbox */
                 while (oldid->name_history.count) {
                     former_name_t *fname = ptrarray_shift(&oldid->name_history);
-                    add_former_name(name_history, fname->dbname,
+                    add_former_name(name_history, fname->name,
                                     fname->mtime, fname->foldermodseq);
                     free(fname);
```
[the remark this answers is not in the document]
ksmurchison: Agreed

ksmurchison (commented):
This all looks good modulo the fname->name leak and the comments that @elliefm and I have on the mboxname_conf_getpath() changes.

imap/mboxname.c
brong: Yep, that's definitely a good plan.

PR #3617, opened 2021-08-18
Fix fromAnyContact Email filters to work with shared addressbooks

rsto:
The Cyrus JMAP email extension supports lookup of emails, where a sender (recipient) may match any contact of that user, including shared addressbooks. However, shared addressbook lookup was broken, both because

- the carddav API was still using mailboxnames for lookup
- the shared addressbook routines where missing an authstate

### DIFF
--- a/imap/carddav_db.c
+++ b/imap/carddav_db.c
@@ -688,7 +688,7 @@ static int groupmembers_cb(sqlite3_stmt *stmt, void *rock)
 
 EXPORTED strarray_t *carddav_getgroup(struct carddav_db *carddavdb,
                                       const mbentry_t *mbentry, const char *group,
-                                      mbname_t *othermb)
+                                      const mbentry_t *othermb)
 {
     int r = 0;
     int isshared = 0;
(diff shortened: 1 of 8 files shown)


rsto:
@ksmurchison I have marked this as draft still, as the code currently does not compile due to `sieve/test.c` breaking. The test code is missing a legit authstate and namespace. We could make that work by making these arguments optional, when they really aren't for regular use. Maybe you have another idea how we can fix this?

```sieve/test.c: In function ‘jmapquery’:
sieve/test.c:601:19: error: ‘script_data_t’ {aka ‘struct <anonymous>’} has no member named ‘authstate’
                 sd->authstate, sd->ns, time(NULL), &err);
                   ^~
sieve/test.c:601:34: error: ‘script_data_t’ {aka ‘struct <anonymous>’} has no member named ‘ns’
                 sd->authstate, sd->ns, time(NULL), &err);
```

Feel free to pull the branch and update (but please let me know, if you do so)

ksmurchison:
I'm actually sick of trying to maintain sieve/test.c.  I'd be happy if it just went away.  SieveScript/test is a better tool at this point anyway.  

Other than that, I'd hack it up however needed to make it compile.  I don't have any clever ideas at ths point.

ksmurchison (approved):
This looks good to me

brong:
I have updated sieve/test.c with code to create a namespace and an authstate.  We're still using it at Fastmail for now, so as much as it's a pain to keep updated, let's keep it around for at least another release.

brong:
I just rebased this and fixed up the API changes on master

PR #3618, opened 2021-08-19
mboxname.c: hash the mboxname before creating fname for lock

ksmurchison:
This is a quick fix for mailbox names that are too long to create a lock file


rsto (approved):
Looks good to me.

brong:
The only downside is yet another thing where you can't tell the username being processed (even more so than uniqueid, where you can at least do a reverse lookup).

On the plus side, better privacy for people - their usernames don't scatter the filesystem as lock files!

I'm tempted though to split this into two things - a section that gets hashed and a section that doesn't, because right now everything from sync channel keys to user lock keys will all be totally opaque, which might be too much un-debugable.  Maybe even create them with the hash and then the first N characters of the lock name, so we can see some detail.

brong:
We fixed this by locking by uniqueid instead for non-legacy mailboxes

PR #3619, opened 2021-08-20
ptloader valgrind fixes

elliefm:
This fixes a bunch of memory errors and leaks in ptloader, found by valgrinding Cassandane.

Notably, they're all in the LDAP PTS module, because this is the only PTS module we have any tests for...

This particular source file is pretty ghastly.  I've tried to make the smallest fixes possible, rather than getting bogged down refactoring for weeks/months...

### DIFF
--- a/ptclient/ldap.c
+++ b/ptclient/ldap.c
@@ -1154,7 +1154,7 @@ static int ptsmodule_make_authstate_attribute(
 
             *dsize = sizeof(struct auth_state) +
                 (numvals * sizeof(struct auth_ident));
-            *newstate = xmalloc(*dsize);
+            *newstate = xzmalloc(*dsize);
             if (*newstate == NULL) {
                 *reply = "no memory";
                 rc = PTSM_FAIL;
@@ -1188,7 +1188,7 @@ static int ptsmodule_make_authstate_attribute(
                 if (numvals == 1) {
                     if(!*newstate) {
                         *dsize = sizeof(struct auth_state);
-                        *newstate = xmalloc(*dsize);
+                        *newstate = xzmalloc(*dsize);
 
                         if (*newstate == NULL) {
                             *reply = "no memory";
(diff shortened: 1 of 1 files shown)


ptclient/ldap.c
```diff
@@ -1173,9 +1176,9 @@ static int ptsmodule_make_authstate_attribute(
                 }
                 strlcat((*newstate)->groups[i].id, rdn[0], sizeof((*newstate)->groups[i].id));
                 (*newstate)->groups[i].hash = strhash((*newstate)->groups[i].id);
+                ldap_value_free(rdn);
```
rsto, line 1179: `rdn` is declared in the function scope, but it looks as if there is no need for that. It could instead be declared inside the block of this for loop.

ptclient/ldap.c
```diff
@@ -1489,21 +1489,17 @@ static int ptsmodule_make_authstate_group(
...
-        rc = ptsmodule_expand_tokens(ptsm->group_base, canon_id, NULL, &base);
-        if (rc != PTSM_OK)
-            goto done;
     }
 
-    syslog(LOG_DEBUG, "(groups) about to search %s for %s", base, filter);
-
-
+    /* XXX i guess canon_id+6 skips over leading "group:" */
     rc = ptsmodule_expand_tokens(ptsm->group_base, canon_id+6, NULL, &base);
```
rsto, line 1498: I think the commit message should reflect that its this function that got called twice. I wasn't clear to me at the start of review, *which* call is duplicated.

rsto (changes requested):
Found two nits that I think are worth changing.

ptclient/ldap.c
elliefm: I agree -- I initially didn't do this cause when I started looking at the variable scoping generally, it's all a huge mess, and I didn't want to complicate things by refactoring it all.  But there's no reason I can't just fix this one variable while I'm looking directly at it, so I'll do so!

elliefm:
@rsto Thanks! I've addressed those two issues, and rebased on current master while there anyway.  Can you give it another look please?

elliefm:
It's probably a good idea to also run `./testrunner.pl --valgrind LDAP` over it, just in case your setup finds something that mine didn't!  There's only 7 tests, so it takes under a minute even with valgrind.

elliefm:
Bump -- can I get a re-review on this please?

ksmurchison (approved):
LGTM

ksmurchison (approved):

PR #3622, opened 2021-08-23
Annotate improvements

brong:
This has two commits.

One fixes the sort order so that diff_annotations and apply_annotations is actually reliable, and the other makes sure that we only copy annotations from the replica if the replica is newer and we're copying everything else from the replica too!

I have a separate commit to make the value a dlist, but it turns out that binary values and dlistsax don't work, so I have to fix that first.


ksmurchison (approved):
looks good

PR #3623, opened 2021-08-23
Annotate dlist

brong:
This is amazingly totally independent of the other sync stuff!  But this makes the annotation format a dlist, which is much nicer - along with some dlist module API updates too.

### DIFF
--- a/cunit/dlist.testc
+++ b/cunit/dlist.testc
@@ -42,9 +42,10 @@ static void test_nil(void)
     dlist_setatom(dl, "STRING", "NIL");
     dlist_setatom(dl, "EMPTY", NULL);
     dlist_setatom(dl, "DOUBLEWHAMMY", "\"NIL\"");
+    dlist_setmap(dl, "BINARY", "\0NIL\0", 5);
     dlist_printbuf(dl, 0, &b);
 
-    CU_ASSERT_STRING_EQUAL(buf_cstring(&b), "(\"NIL\" NIL {5+}\r\n\"NIL\")");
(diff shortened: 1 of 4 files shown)


ksmurchison (commented):
This looks good.  Do we have a tool to convert the existing annotations dbs to use the dlist format?

brong:
> This looks good. Do we have a tool to convert the existing annotations dbs to use the dlist format?

Nope - but I think if we wanted to, we'd probably just have reconstruct notice old-format annotations and upgrade them.  The code reads both formats just fine, so there's no real reason to upgrade everything straight away.

brong:
I don't want to deploy this one straight away because it's not backwards compatible, so for sure I want to have the upgrade that adds this be as simple as possible.

ksmurchison (approved):

elliefm (changes requested):
This one will need an entry in changes/next.  We're changing a storage format and will need to provide instructions to people upgrading.  Even if those instructions are just "you don't need to do anything, it'll upgrade in place as necessary", we still need to capture that before we forget it.

brong:
> This one will need an entry in changes/next. We're changing a storage format and will need to provide instructions to people upgrading. Even if those instructions are just "you don't need to do anything, it'll upgrade in place as necessary", we still need to capture that before we forget it.

Yeah good point!  OK!  I'll come back to it, I'm not in a super hurry to apply this anyway, but it's good to remember that.

brong:
Keeping this as a draft for now - don't wanna change format until we're ready!

PR #3626, opened 2021-08-24
imap/annotate: remove unused appendstrlistpat() and comp_pat from struct strlist

dilyanpalauzov:
• imap/imapd.c: remove unnecessary freestrlist declaration


brong (approved):
Hey, great - this is a good piece of excess code to get rid of.  Thanks.

(at which point - we probably should just be using a strarray for these things, but that's a job for another day)

brong:
> @brong: Are you still working on annotations stuff? Does this PR intersect with things you have in progress?

nothing that intersects with this.  In fact I like this so much I'm going to look at how much trouble it is to turn all these into strarrays.

PR #3627, opened 2021-08-24
JMAP Backup relaxed locking

ksmurchison:
Process messages in batches so we don't deadlock and/or hold locks for a long time.  All Cass tests succeed even with a batch size of 1

### DIFF
--- a/imap/jmap_backup.c
+++ b/imap/jmap_backup.c
@@ -80,30 +80,34 @@ jmap_method_t jmap_backup_methods_standard[] = {
     { NULL, NULL, NULL, 0}
 };
 
+/* NOTE: we don't set flags to require CSTATE, because that holds
+ * a user lock (exclusive if READ_WRITE is requested) for the entire
+ * time the method is running.  Backup restores can be quite slow,
+ * and we release locks in batches so that the user can keep working */
 jmap_method_t jmap_backup_methods_nonstandard[] = {
     {
         "Backup/restoreContacts",
         JMAP_BACKUP_EXTENSION,
         &jmap_backup_restore_contacts,
-        JMAP_NEED_CSTATE | JMAP_READ_WRITE
+        /*flags*/0
     },
     {
         "Backup/restoreCalendars",
         JMAP_BACKUP_EXTENSION,
         &jmap_backup_restore_calendars,
-        JMAP_NEED_CSTATE | JMAP_READ_WRITE
+        /*flags*/0
     },
     {
         "Backup/restoreNotes",
         JMAP_BACKUP_EXTENSION,
         &jmap_backup_restore_notes,
-        JMAP_READ_WRITE
+        /*flags*/0
     },
     {
         "Backup/restoreMail",
         JMAP_BACKUP_EXTENSION,
         &jmap_backup_restore_mail,
-        JMAP_NEED_CSTATE | JMAP_READ_WRITE
+        /*flags*/0
     },
     { NULL, NULL, NULL, 0}
 };
--- a/imap/jmap_backup.c
+++ b/imap/jmap_backup.c
@@ -280,6 +280,7 @@ struct restore_rock {
     void *rock;
     struct mailbox *mailbox;
     int keep_open;
+    int result;
 };
 
 struct restore_info {
@@ -288,8 +289,8 @@ struct restore_info {
     uint32_t uid_torecreate;
 };
 
-static void restore_resource_cb(const char *resource __attribute__((unused)),
-                                void *data, void *rock)
+static int restore_resource_cb(const char *resource __attribute__((unused)),
+                               void *data, void *rock)
 {
     struct restore_info *restore = (struct restore_info *) data;
     struct restore_rock *rrock = (struct restore_rock *) rock;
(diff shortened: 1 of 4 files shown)


imap/jmap_backup.c
```diff
@@ -1548,22 +1562,19 @@ static int restore_message_list_cb(const mbentry_t *mbentry, void *rock)
...
+            struct mailbox_plan *plan =
+                hash_lookup(mailbox->name, mrock->mailboxes);
+            if (!plan) {
+                /* Create a plan for this mailbox */
+                plan = xzmalloc(sizeof(struct mailbox_plan));
+                hash_insert(mailbox->name, plan, mrock->mailboxes);
             }
+
+            /* Add this msgno to the unflag array */
+            arrayu64_append(&plan->unflag, record->recno);
```
brong, line 1577: Unfortunately recno isn't stable over an unlock :(  - if the mailbox repacks, the recno will be wrong - so we're going to have to record them by UID instead.

imap/jmap_backup.c
```diff
@@ -1726,15 +1714,15 @@ static void restore_mailbox_plan_cb(const char *guid __attribute__((unused)),
...
+                if (!plan) {
+                    /* Create a plan for this mailbox */
+                    plan = xzmalloc(sizeof(struct mailbox_plan));
+                    hash_insert(rmail->mboxname, plan, mailboxes);
                 }
 
-                /* Add this msgno to the mailbox */
-                arrayu64_append(msgnos, rmail->msgno);
+                /* Add this msgno to the restore array */
+                arrayu64_append(&plan->restore, rmail->msgno);
```
brong, line 1725: Which raises a naming convention issue here that I didn't pick up before.  What's `msgno`?  The naming conventions say that `msgno` is the IMAP message number in a selected mailbox, i.e. an `index_state` offset.

imap/jmap_backup.c
```diff
@@ -1957,9 +1968,27 @@ static void restore_mailbox_cb(const char *mboxname, void *data, void *rock)
 
         message_set_from_mailbox(mailbox, msgno, msg);
```
brong, line 1969: Right, the signature on this is:
```
EXPORTED void message_set_from_mailbox(struct mailbox *mailbox, unsigned int recno, message_t *m)
```

So it's a recno.

I think the best approach might just be to iterate the mailbox and check for next uid - otherwise we could use the finduid lookup for each in order to find the correct record.  There is a question whether we should try to improve the APIs here first, or hack around them :(  But it turns out this is all broken if it races with a cyr_expire repack!

brong (changes requested):
Sorry :(  I found another bit issue.  The use of `recno` isn't safe across unlock/relock, so we're going to have to rearchitect that to use UIDs instead too.  Boo :(

An option would be to use seqsets, and then we could create a merged seqset of UIDs that are interesting and pass that to `mailbox_iter_uidset` - though with restarting, maybe that's too much messing around as well.

imap/jmap_backup.c
```diff
@@ -321,12 +321,16 @@ static void restore_resource_cb(const char *resource __attribute__((unused)),
         goto done;
     }
 
-    if (restore->msgno_torecreate) {
-        recreatemsg = message_new_from_mailbox(mailbox, restore->msgno_torecreate);
+    if (restore->uid_torecreate) {
+        struct index_record record;
+        mailbox_find_index_record(mailbox, restore->uid_torecreate, &record);
```
wolfsage, line 326: Here and elsewhere we don't check the return value of mailbox_find_index_record. Should we? Or is the point that if we got here and it does fail, something is awful messed up anyway and crashing later is okay?

brong: Yeah, we need to check that it succeeds.  It could be that the UID we're planning to work with got removed by a repack.

imap/jmap_backup.c
```diff
@@ -1966,32 +1970,31 @@ static void restore_mailbox_cb(const char *mboxname, void *data, void *rock)
             }
         }
 
-        message_set_from_mailbox(mailbox, msgno, msg);
+        struct index_record record;
+        mailbox_find_index_record(mailbox, uid, &record);
+        message_set_from_record(mailbox, &record, msg);
```
brong, line 1975: ditto here - we need to check that we found it.

brong (changes requested):
Mostly looks good, just note the "we can fail to find it" error case

imap/jmap_backup.c
wolfsage: Oh actually there's something worse too in not checking that. We had an issue awhile back because of this bit of mailbox_find_index_record:
```
    /* wasn't the actual record? */
    if (record->uid != uid) return IMAP_NOTFOUND;

```
So it might fill in `&record` and then return saying "I didn't find it" and then later code could work on the wrong record

brong (approved):
Yep, that looks good.  I think it's OK to bail out if we fail to find any of the UIDs we looked for - it'll be rare, and we can just run the restore again.

ksmurchison:
@brong One final commit per your last comment

PR #3628, opened 2021-08-25
changes/next/TEMPLATE: rename to 0000-TEMPLATE to fix ls sort order

elliefm:
Looks like `ls` these days may default to a case-insensitive sort, so an uppercase name is no longer enough to force your important files to the top of the listing.

Expected result: `ls changes/next` lists the template file first, before all the changes files.  It works for me; please test this and report back whether it works for you.  I assume some of us might have different `ls` defaults, or have manually configured different behaviours.


brong:
my ls defaults to case insensitive, so TEMPLATE appears half way through.

```
brong@zula:~/src/cyrus-imapd$ ls changes/next/
3360-apply-message-write-failure-desync  message_cache_quotedaddr
contact_blobid				 message_parse_bogus_charset_params
create_no_intermediates			 propfind_caltransp
fix_write_searchaddr_quoted		 RFC8970
guidsearch_mixed_clauses		 search_large_text
icalendar_max_size			 squatter_s_argument_required
invalid_rrule_handling			 sync-client-retry-wait-limit
jmap_contact_importance_peruser		 sync-server-bye
jmap_contact_new_blobids		 TEMPLATE
jmap_contacts_apple_labels		 usermeta-by-id-bis
jmap_getblob_handlers_v2		 vcard_max_size
jmap_push				 ws_jmap_push
mboxevent_jmap_email			 ws_timeout
```

brong (approved):
Yes please, we should definitely do this

PR #3637, opened 2021-08-28
Delayed actions - don't repack until we've replied to the user!

brong:
This set of commits creates a list of "delayed actions" deep in libcyrus, where all the libraries can call it.  It then runs them either during shutdown, or you can call them at any time.  By doing this, both mailbox cleanup and twoskip checkpoints can be delayed until after we've replied to the user, meaning that the user doesn't have to wait on background tasks.

an alternative option is to log all these things somewhere and have a maintenance job that reads the log file sync_log style and does the actions, but this seems a nice sweet spot that doesn't require another daemon.

### DIFF
--- a/imap/cyr_expire.c
+++ b/imap/cyr_expire.c
@@ -87,7 +87,6 @@
 #define SECS_IN_A_DAY (24 * SECS_IN_AN_HR)
 
 /* global state */
-static volatile sig_atomic_t sigquit = 0;
 static int verbose = 0;
 static const char *progname = NULL;
 static struct namespace expire_namespace; /* current namespace */
@@ -380,7 +379,7 @@ static int archive(const mbentry_t *mbentry, void *rock)
     struct mailbox *mailbox = NULL;
     int archive_seconds = -1;
 
-    if (sigquit)
+    if (in_shutdown)
         return 1;
 
     if (mbentry->mbtype & MBTYPE_DELETED)
@@ -452,7 +451,7 @@ static int expire(const mbentry_t *mbentry, void *rock)
     int expire_seconds = 0;
     int did_expunge = 0;
 
-    if (sigquit) {
+    if (in_shutdown) {
         /* don't care if we leak some memory, we are shutting down */
         return 1;
     }
@@ -587,7 +586,7 @@ static int expire_conversations(const mbentry_t *mbentry, void *rock)
     unsigned int nseen = 0, ndeleted = 0;
     char *filename = NULL;
 
-    if (sigquit)
+    if (in_shutdown)
         return 1;
 
     if (mbentry->mbtype & MBTYPE_DELETED)
@@ -627,7 +626,7 @@ done:
 
 static void sighandler(int sig __attribute((unused)))
 {
-    sigquit = 1;
+    in_shutdown = 1;
     return;
 }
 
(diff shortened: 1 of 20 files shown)


brong:
(I've marked this as a draft for now, it's not ready to merge yet, but would still love reviews on the design)

brong:
Ok, I figured it out :) Gotta drop the transaction if not making changes, cyrusdb is a bit more strict about this than mailboxes

imap/cyr_expire.c
```diff
@@ -624,7 +626,7 @@ static int expire_conversations(const mbentry_t *mbentry, void *rock)
 
 static void sighandler(int sig __attribute((unused)))
 {
-    sigquit = 1;
+    in_shutdown = 1;
```
elliefm, line 629: If we start setting `in_shutdown` from a signal handler like this, then it needs to change type to `volatile sig_atomic_t`, like `sigquit` was.  Looks like `in_shutdown` is just an `int`, and comes from imap/global.c

Though... it feels wrong that `cyr_expire` even has its own signal handler.  Shouldn't it be using the lib/signals.c functions like everything else?

imap/pop3d.c
```diff
@@ -603,6 +603,8 @@ void shut_down(int code)
 
     in_shutdown = 1;
 
+    libcyrus_run_delayed();
```
elliefm, line 606: Some of these shut_down functions set `in_shutdown=1` before calling libcyrus_run_delayed(), others do it in the other order.

Might not be significant but it smells funny?

imap/sync_server.c
```diff
@@ -395,6 +395,9 @@ static void usage(void)
  */
 void shut_down(int code)
 {
+    /* run any delayed commands before we go into shutdown */
+    libcyrus_run_delayed();
+
     in_shutdown = 1;
```
elliefm, line 401: Like this one is careful to not set `in_shutdown = 1` until after the libcyrus_run_delayed, and even has a comment about it

lib/libcyr_cfg.c
```diff
@@ -213,12 +225,56 @@ EXPORTED void libcyrus_config_setswitch(enum cyrus_opt opt, int val)
...
+    delayed_actions = action;
+}
+
+EXPORTED void libcyrus_run_delayed(void)
+{
+    while (delayed_actions) {
+        struct delayed_action *action = delayed_actions;
+        delayed_actions = action->next;
+        action->cb(action->rock);
+        action->myfree(action->rock);
```
elliefm, line 266: myfree is allowed to be NULL, so need to check for that before calling it

imap/sync_server.c
brong: Yeah, I reverted that behaviour - and I'll drop this.

The `in_shutdown` flag says "finish fast", so repacking is a bad choice then!  It happens because we've had a sigquit.

imap/pop3d.c
brong: Actually, we should always call it after setting in_shutdown, if the work hasn't been explicitly done already - because in_shutdown gets called by

lib/libcyr_cfg.c
brong: yes good point!

imap/cyr_expire.c
brong: yeah, it's probably old.  I'll change that.

(sig_atomic_t probably isn't needed given that it only ever goes 0 to 1 and it isn't the end of the world to miss the value temporarily, but also - may as well change the global to be that too)

ksmurchison (approved):
LGTM

imap/global.c
```diff
@@ -93,7 +93,7 @@ static enum {
 
 static int cyrus_init_nodb = 0;
 
-EXPORTED int in_shutdown = 0;
+EXPORTED volatile sig_atomic_t in_shutdown = 0;
```
elliefm, line 96: This change isn't actually needed anymore now that you've made cyr_expire just use lib/signals 😅

PR #3638, opened 2021-08-30
testsaslauthd needs socket path info (-f)

bpcamac:
testsaslauthd will complain with "connect() : No such file or directory" if it doesn't know the socket path via (-f).


elliefm (approved):
This seems fine to me...

The manpage for testsaslauthd (in the cyrus-sasl repo) says not to use testsaslauthd unless you know what you're doing, and to read the source code 😬

I expect the exact path of the socket probably varies from system to system, but at least documenting what it should vaguely look like (giving you a hint what to look for) is probably more help than leaving it off entirely

PR #3639, opened 2021-08-31
Intermediary repair

brong:
This is a work in progress!  The first commit is urgent, because it fixes a corruption bug that's in the wild.

The second commit converts all intermediate creation to be full mailboxes - which is maybe actually all we need.

The next step if we decided to do it, would be to extract the magic code from cmd_create in imapd.c and put that into an `mboxlist_createtree(` wrapper around `mboxlist_createmailbox`.

Then we just need to make the replication protocol reject anything which breaks these guarantees, and make the sync_client always order things so it creates parents before children and deletes children before parents.


ksmurchison (commented):
This is looking sane

brong:
Two of these commits have already been deployed to Fastmail production to fix the immediate issues :)  The rest is all ready for review and hopefully deploy soon too!

ksmurchison (approved):

PR #3643, opened 2021-09-02
Add 'caldav_accept_invalid_rrules' option.

ksmurchison:
Adds a setting to accept invalid RRULEs (e.g. FREQ=WEEKLY;BYMONTHDAY=15) rather than rejecting them.


brong (approved):
Yeah, looks great.  I guess we need a new cyruslibs too?

PR #3644, opened 2021-09-03
TEMPORARY: Revert iana zones

brong:
This is just here until we fix up the zone names from libical so that the prefixed versions don't leak out to the world via notifyd.


brong:
no longer needed, a better fix has been done.

PR #3648, opened 2021-09-06
docs: call manpages index "index"

elliefm:
So that navigating to http://www.cyrusimap.org/imap/reference/manpages/ works!

It's currently called "[commands](https://www.cyrusimap.org/imap/reference/manpages/commands.html)", which is discoverable via the sidebar links but hard to find by typing in an address.

Not really sure what I expect a reviewer to do with this...


brong (approved):
Very sensible, thanks.

PR #3649, opened 2021-09-08
document that carddav httpmodule requires caldav

elliefm:
This updates the description of `httpmodules` to reflect that the "carddav" module is dependent on the "caldav" module being enabled too.

Though I'm not sure if this is expected or not.  If they're supposed to be independent, then we have a bug to fix instead...


brong:
This got fixed by https://github.com/cyrusimap/cyrus-imapd/pull/3650 so we don't need it any more

PR #3653, opened 2021-09-08
http_dav.c: lookup INBOX when parsing principal URLs, rather than blindly looking up calendar-home-set

ksmurchison:
This allows the CardDAV module to be used indepedently of CalDAV.  All Cass tests still pass

### DIFF
--- a/imap/http_dav.c
+++ b/imap/http_dav.c
@@ -67,7 +67,6 @@
 #include "annotate.h"
 #include "acl.h"
 #include "append.h"
-#include "caldav_db.h"
 #include "dlist.h"
 #include "global.h"
 #include "http_dav.h"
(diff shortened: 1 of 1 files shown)


brong (approved):
Oh hey - yeah, definitely better.

elliefm:
`JMAPCalendars.calendar_set_sharewith` is erroring for me today, and bisect tells me it was this one that broke it.  I haven't looked into it in any depth, just ran a bisect.

brong:
> `JMAPCalendars.calendar_set_sharewith` is erroring for me today, and bisect tells me it was this one that broke it. I haven't looked into it in any depth, just ran a bisect.

This is what I get for not running the tests first.

(and then... for not reading my email until I tried to debug the same test thinking it was something I'd broken in my work on another PR)

The underlying bug is that the principal access control is by ACLs on `INBOX.#calendars`, which gets updated for every user who has read access on any of the sub items - so this change also updated the mbentry which is later checked for access control.  Doh.

PR #3654, opened 2021-09-09
Always compile autocreate in

robn:
Since it requires config settings to actually enable it, and those are
off by default, there's no good reason to not have it always there.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -957,6 +957,8 @@ imap_calalarmd_SOURCES = imap/calalarmd.c imap/mutex_fake.c
 imap_calalarmd_LDADD = $(LD_SERVER_ADD)
 
 imap_imapd_SOURCES = \
+	imap/autocreate.h \
+	imap/autocreate.c \
 	imap/imap_proxy.c \
 	imap/imap_proxy.h \
 	imap/imapd.c \
@@ -967,12 +969,6 @@ imap_imapd_SOURCES = \
 	imap/sync_support.c \
 	imap/sync_support.h \
 	master/service.c
-
-if AUTOCREATE
-imap_imapd_SOURCES += \
-	imap/autocreate.h \
-	imap/autocreate.c
-endif
 
 imap_imapd_LDADD = $(LD_SIEVE_ADD) $(LD_SERVER_ADD)
 
@@ -1171,6 +1167,8 @@ endif
 endif
 
 imap_lmtpd_SOURCES = \
+	imap/autocreate.c \
+	imap/autocreate.h \
 	imap/lmtpd.c \
 	imap/lmtpd.h \
 	imap/lmtpengine.c \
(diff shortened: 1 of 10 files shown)


robn:
There is a risk, of course, that even when disabled by config, the now-present code will cause quirks or problems in other things that have received a lot more love recently. Fastmail in particular does not normally compile with `--enable-autocreate`, so there is a change there. So I would suggest that the autocreate code paths at least get a light review for sensibleness.

robn:
As to "why": when this is not compiled in, the autocreate config options do not flag any kind of error, they just .. don't work. This led me down a merry path to debug why. And since autocreate is a very convenient way to experiment with Cyrus without surrounding supporting user management infrastructure, it seems more interesting to just make it available rather than try to put warnings on the config items if used when they can't possibly work.

Makefile.am
```diff
@@ -1171,6 +1167,8 @@ endif
 endif
 
 imap_lmtpd_SOURCES = \
+	imap/autocreate.c \
+	imap/autocreate.h \
```
dilyanpalauzov, line 1171: The .h files need to be mentioned in a single _SOURCES place in the whole Makefile.am .  The sole purpose of mentioning them ever, is to bundle them on `make dist`.

Makefile.am
```diff
@@ -1351,19 +1343,15 @@ endif # CUNIT
 endif # JMAP
 
 imap_pop3d_SOURCES = \
+	imap/autocreate.c \
+	imap/autocreate.h \
```
dilyanpalauzov, line 1347: Since autocreate.c is used by many executables, I propose moving it to libimap.so .

configure.ac
```diff
@@ -1538,15 +1538,6 @@ AC_ARG_WITH(pidfile,
 MASTERPIDFILE="\"$MASTERPIDFILE\""
 AC_DEFINE_UNQUOTED(MASTER_PIDFILE, $MASTERPIDFILE,[Name of the pidfile for master])
 
-dnl
-dnl see if we're compiling with autocreate support
-dnl
-AC_ARG_ENABLE(autocreate,
-        [AS_HELP_STRING([--enable-autocreate], [enable autocreate support])],,[enable_autocreate="no";])
```
elliefm, line 1545: Rather than removing all this, I think I'd prefer if we just flipped the default to "yes".  That way anyone who maybe doesn't trust the autocreate code (reasonable!) can still configure with `--disable-autocreate` and not have to think about it.

Longer term, removing a bunch of conditional compilation is one of our goals, because it reduces the size of the matrix of configurations that hypothetically need testing.  But the first step towards that is including it by default, not including it unconditionally. :)

elliefm (changes requested):

elliefm:
Just gonna close this, it's bitrotted and I don't think it's worth the work to resuscitate.  Like, it's not a bad idea, and we do intend to get rid of a lot of these configure options, but I still don't trust autocreate enough to make it always available.  When we do get there, it's probably easier to just start again.

PR #3655, opened 2021-09-10
.travis.yml: notify every success on slack

elliefm:
We've gotten used to "silence means success", but these days silence
usually means "travis is out of credit and builds are no longer
running".

This makes success noisy too, so when things get quiet we're probably
out of credits.


ksmurchison (approved):

brong (approved):
Yes, this is good

PR #3656, opened 2021-09-10
mailbox: don't leak header name!

elliefm:
Found by valgrind


brong (approved):
Yep, this patches the "fix" I did the other day.

PR #3657, opened 2021-09-10
HTTP module for PTS

ksmurchison:
This allows fetching group membership via a simple HTTP GET request, with JSON response


configure.ac
```diff
@@ -1058,6 +1058,14 @@ if test "x$enable_afs" = "xyes"; then
 fi
 AM_CONDITIONAL([USE_AFSKRB], [grep HAVE_AFSKRB confdefs.h >/dev/null])
 
+AC_ARG_ENABLE(httppts,
+        [AS_HELP_STRING([--enable-httppts], [enable HTTP PTS support])],, [enable_httppts=no])
```
brong, line 1062: Should we just default this on - or even require it?  It's annoying to add yet another compile time flag for a tiny bit of inoffensive code.

ptclient/http.c
```diff
@@ -0,0 +1,396 @@
...
+    if (!strncasecmp(p, "https://", 8)) {
+        state.uri.host = p += 8;
+        state.uri.port = 443;
+        state.uri.https = 1;
+    }
+    else if (!strncasecmp(p, "http://", 7)) {
+        state.uri.host = p += 7;
+        state.uri.port = 80;
+    }
+    else fatal("Invalid 'httppts_uri' scheme", EX_CONFIG);
```
brong, line 260: is the some URI parser library we could/should be using here?

brong (approved):
This looks like it's gonna work fine, just would be nice to use some more library code for all the encoding and URI parsing.

configure.ac
ksmurchison: I was thinking the same thing as I coded this up.  Initially I thought it might have to be gated due to using optional 3rd paty libraries, but that doesn't match reality.  I'll make the necessary changes.

ptclient/http.c
ksmurchison: I tried using xmlParseURI, which we use in httpd.c, but it doesn't like the {} in the template.  I didn't want to add another dependency for something that is easily parsable

configure.ac
brong: Great!  Fewer config options is always good.

ksmurchison: ptloader (and friends) is no built by default, with HTTP support

elliefm (commented):
I wonder how we'd test this from Cassandane.  Probably need a fake data-driven httpd, like how we have `utils/fakeldapd` and `data/directory.ldif`, and then (as a start) similar tests to what's in LDAP.pm...

elliefm:
This will need a changes/next file too

brong (approved):
At a read through, this seems fine now.  It's also entirely optional and turned on by config, so I think it's quite safe to merge and iterate on if we hit issues.

configure.ac
```diff
@@ -1109,8 +1109,6 @@ AC_SUBST(LDAP_LDFLAGS)
 AC_SUBST(LDAP_LIBS)
 AM_CONDITIONAL([HAVE_LDAP], [test "$have_ldap" = "yes"])
 
-AM_CONDITIONAL([PTCLIENT], [test "x$enable_afs" = "xyes" -o "$have_ldap" = "yes"])
-
```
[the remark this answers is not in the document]
ksmurchison: --disable-httpd is already broken in master (due to libxml2 and libical used in support files) and its no worse in this PR

elliefm (approved):
Looks good, though I notice there's no tests.  Looks like we talked about that last time, and we would have needed some way to hook it up, which didn't exist at the time.

I think @wolfsage's new Util::TestUrl from #3857 might make it easy to test this now, if you wanted to have a go?

ksmurchison:
Added HTTPPTS.pm which is a copy of LDAP, but using the HTTP PTS module

ksmurchison:
@wolfsage Would you mind looking at HTTPPTS.pm and make sure I'm using TestUrl properly?

wolfsage:
Yes that looks right to me

elliefm (approved):
There's a problem in the tests that prevents the LDAP and HTTPPTS suites being run in the same run (details inline).

Everything else seems fine, and the fix for the test problem will be trivial and won't need re-review, so I'm approving this in advance to save another round-trip.

cassandane/Cassandane/Cyrus/HTTPPTS.pm
```diff
@@ -0,0 +1,355 @@
...
+use lib '.';
+use base qw(Cassandane::Cyrus::TestCase);
+use base qw(Cassandane::Unit::TestCase);
+use Cassandane::Util::Log;
+
+Cassandane::Cyrus::TestCase::magic(AltPTSDBPath => sub {
+    shift->config_set(
+        'ptscache_db_path' => '@basedir@/conf/non-default-ptscache.db'
+    );
+});
```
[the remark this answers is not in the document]
ksmurchison: Fixed

PR #3658, opened 2021-09-10
Namespacelock mbentry updates

brong:
This whole series of commits is (apart from a couple of bugfixes at the top) all about making sure that there's a namespace lock held every time we update the mailboxes.db.  It also wrap a transaction around all the updates related to a single mbentry (N key, I key, ACLs).

This is a precursor to the work to make renames atomic.  We need the namelocks coming in, so a lot of the commits are bugfixes around the namelocking further out in the callers.

### DIFF
--- a/imap/jmap_backup.c
+++ b/imap/jmap_backup.c
@@ -1143,7 +1143,6 @@ static int recreate_calendar(const mbentry_t *mbentry,
     if (!r && !*newmailbox) {
         /* Create the calendar */
         char *newmboxname = caldav_mboxname(req->accountid, makeuuid());
-        struct mboxlock *namespacelock = user_namespacelock(req->accountid);
         mbentry_t newmbentry = MBENTRY_INITIALIZER;
         newmbentry.name = newmboxname;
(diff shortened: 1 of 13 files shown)


imap/jmap_backup.c
```diff
@@ -1143,15 +1143,13 @@ static int recreate_calendar(const mbentry_t *mbentry,
     if (!r && !*newmailbox) {
         /* Create the calendar */
         char *newmboxname = caldav_mboxname(req->accountid, makeuuid());
-        struct mboxlock *namespacelock = user_namespacelock(req->accountid);
```
ksmurchison, line 1146: Don't we still want a namelock when recreating the mailbox?

brong: This is locked in the caller, so there's no need to lock it again.

brong:
The one big question as raised by Ken on slack when this first came up is... should we be doing these locks in murder frontends and mupdate masters?  Con: an extra locking overhead, Pro: consistency of code and extra guarantee that we don't try to fiddle mbentries under things that aren't expecting it.

ksmurchison (approved):

PR #3660, opened 2021-09-10
CardDAV without CalDAV bis

ksmurchison:
This switches back to looking up a "home-set" collection when parsing principal URLs, defaulting to calendar-home-set if CalDAV is enabled, then dropping to addressbook-home-set if enabled, and finally using the root of a user's #drive hierarchy


brong (approved):
Yeah looks good - covers all the bases!  Thanks.

PR #3661, opened 2021-09-10
HTTP/1.1 Keep-Alive

ksmurchison:
HTTP/1.1 requests can also use the Keep-Alive header and we should respond in kind.  Very trivial patch.

### DIFF
--- a/imap/httpd.c
+++ b/imap/httpd.c
@@ -2392,12 +2392,15 @@ static int parse_connection(struct transaction_t *txn)
                 if (!strcasecmp(token, "upgrade")) {
                     /* Client wants to upgrade */
                     parse_upgrade(txn);
+                    break;
                 }
                 else if (!strcasecmp(token, "close")) {
                     /* Non-persistent connection */
                     txn->flags.conn |= CONN_CLOSE;
+                    break;
                 }
-                break;
+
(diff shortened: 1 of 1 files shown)


imap/httpd.c
```diff
@@ -2392,12 +2392,15 @@ static int parse_connection(struct transaction_t *txn)
...
+                    break;
                 }
                 else if (!strcasecmp(token, "close")) {
                     /* Non-persistent connection */
                     txn->flags.conn |= CONN_CLOSE;
+                    break;
                 }
-                break;
+
+                GCC_FALLTHROUGH
```
brong, line 2403: If we look at the line a few below this (I can't tag it in the comment, because it's outside the context, doh) we see:

```
                    /* Persistent connection */
                    txn->flags.conn = CONN_KEEPALIVE;
```

Does that need to be changed to:

```
                    txn->flags.conn |= CONN_KEEPALIVE;
```

Since http/1.1 allows additional connection flags.

brong (changes requested):

imap/httpd.c
ksmurchison: This logic change was incorrect and has been fixed in subsequent commit

brong (approved):

PR #3662, opened 2021-09-10
prot_waitevent-leak-fix

ksmurchison:
proxy.c: don't leak prot_waitevent when a connection to a backend has gone away


elliefm (approved):
Great, this fixes the leak that MurderJMAP.frontend_commands found.  Thanks!

PR #3663, opened 2021-09-11
Memory leaks

brong:
Two annoying memory leaks introduced by my delayed actions work :(  Found by valgrind.


elliefm (approved):
Looks good

PR #3664, opened 2021-09-11
imapd: flush output when starting output so changes are told immediately

brong:
This fix is good to backport to 3.0, 3.2 and 3.4 as well.

As discovered in https://bugzilla.mozilla.org/show_bug.cgi?id=1708981

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -3279,6 +3279,7 @@ static void cmd_idle(char *tag)
         idling = 1;
 
         index_release(imapd_index);
+        prot_flush(imapd_out);
         while ((flags = idle_wait(imapd_in->fd))) {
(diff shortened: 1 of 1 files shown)


elliefm (commented):
>imapd: flush output when starting output so changes are told immediately #3664

Did you mean "when starting idle" rather than "when starting output"?

Code looks good, but can you force push with the commit message fixed please :)

brong:
> > imapd: flush output when starting output so changes are told immediately #3664
> 
> Did you mean "when starting idle" rather than "when starting output"?
> 
> Code looks good, but can you force push with the commit message fixed please :)

Done - also put the detail of the Thunderbird bugzilla into the commit message.

elliefm (approved):

PR #3669, opened 2021-09-13
jmap_backup: always use the rrock mailbox pointer

brong:
This fixes a "use after free" when releasing and recreating the
mailbox where `struct mailbox *mailbox` pointed at the old copy.


elliefm (approved):
Looks good.  I've also run this through the JMAPBackup test suite, both with and without valgrind, and it comes up clean.

PR #3670, opened 2021-09-13
Sync empty names

brong:
Replication has been bailing out on:

MAILBOX ""

We're not quite sure where those are coming from - so by adding an assertion we'll find out :)

Also, fix the reader so it just skips such bogus entries.

### DIFF
--- a/imap/cyr_synclog.c
+++ b/imap/cyr_synclog.c
@@ -51,6 +51,7 @@
 #include <unistd.h>
 #endif
 
+#include "assert.h"
 #include "global.h"
 #include "sync_log.h"
 #include "util.h"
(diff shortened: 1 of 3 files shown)


elliefm (commented):
This seems fine to me.  Is this a temporary-only one, until we find the source of the empty names?

Leaving it without an approval tick for now so we don't accidentally merge it later, if we weren't intending to do so!

brong:
> This seems fine to me. Is this a temporary-only one, until we find the source of the empty names?
> 
> Leaving it without an approval tick for now so we don't accidentally merge it later, if we weren't intending to do so!

I think both bits are probably OK - the "don't crash if you see the wrong thing in the log file" is good safety, and the "crash if you're about to generate a wrong thing" is a good belt-and-suspenders that we have a in a few places in the code now.

elliefm (approved):

PR #3671, opened 2021-09-13
jmap_ical: ignore ORGANIZER and ATTENDEE if value is empty

rsto:
Depending on libical settings an empty-valued ORGANIZER or ATTENDEE property may return NULL or an empty string as
calendar address. This patch unifies handling such values as NULL values.

Tested in https://github.com/cyrusimap/cassandane/tree/jmap_ical_ignore_empty_caladdress

### DIFF
--- a/imap/jmap_ical.c
+++ b/imap/jmap_ical.c
@@ -241,13 +241,23 @@ static const char *sha1key(const char *val, char *keybuf)
 
 static char *normalized_uri(const char *uri)
 {
-    const char *col = strchr(uri, ':');
-    if (!col) return xstrdupnull(uri);
+    if (!uri) return NULL;
 
     struct buf buf = BUF_INITIALIZER;
-    buf_setmap(&buf, uri, col - uri);
-    buf_lcase(&buf);
-    buf_appendcstr(&buf, col);
+    const char *col = strchr(uri, ':');
+    if (col) {
+        /* Normalize URI scheme to lower case */
+        buf_setmap(&buf, uri, col - uri);
+        buf_lcase(&buf);
+        buf_appendcstr(&buf, col);
+    }
+    else buf_setcstr(&buf, uri);
+
+    buf_trim(&buf);
+    if (!buf_len(&buf)) {
+        buf_free(&buf);
+        return NULL;
+    }
     return buf_release(&buf);
 }
 
@@ -1291,8 +1301,13 @@ rsvpto_from_ical(icalproperty *prop)
     /* Read URI from property value and check if this URI already is defined.
      * If it isn't, this could be because an iCalendar client updated the
      * property value, but kept the RSVP x-params. */
-    const char *caladdress = icalproperty_get_value_as_string(prop);
-    if (!caladdress) goto done;
+    const char *val = icalproperty_get_value_as_string(prop);
+    if (!val) goto done;
+    buf_setcstr(&buf, val);
+    buf_trim(&buf);
+    if (!buf_len(&buf)) goto done;
+
+    const char *caladdress = buf_cstring(&buf);
     int caladdress_is_defined = 0;
     json_t *jval;
     const char *key;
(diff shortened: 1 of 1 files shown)


imap/jmap_ical.c
```diff
@@ -3356,11 +3376,18 @@ participants_to_ical(icalcomponent *comp,
...
+                icalproperty *attd = icalcomponent_get_first_property(tmp, ICAL_ATTENDEE_PROPERTY);
+                allow_organizer_attendee_only = (orga == NULL) != (attd == NULL);
+                if (!allow_organizer_attendee_only) {
+                    char *orgauri = normalized_uri(icalproperty_get_value_as_string(orga));
+                    char *attduri = normalized_uri(icalproperty_get_value_as_string(attd));
+                    allow_organizer_attendee_only = (orgauri == NULL) != (attduri == NULL);
+                    free(orgauri);
+                    free(attduri);
                 }
+                if (allow_organizer_attendee_only)
```
ksmurchison, line 3389: Couldn't this be 'else if'

ksmurchison (commented):
One comment inline

imap/jmap_ical.c
rsto: Not sure I understand: Line 3395 updates `allow_organizer_attendee_only` and we want to break if either the property isn't set at all, or has a null value.

brong: I'd probably recommend an extra blank link between the blocks in that case, so it's clear it's not a continuation case :)  First item for our code review review maybe!

* if it's an `else if` there should be no blank line, if it's a new `if` then there should be a blank line before the new block.

imap/jmap_ical.c
```diff
@@ -241,13 +241,22 @@ static const char *sha1key(const char *val, char *keybuf)
...
 {
-    const char *col = strchr(uri, ':');
-    if (!col) return xstrdupnull(uri);
+    if (!uri) return NULL;
 
     struct buf buf = BUF_INITIALIZER;
-    buf_setmap(&buf, uri, col - uri);
-    buf_lcase(&buf);
-    buf_appendcstr(&buf, col);
+    const char *col = strchr(uri, ':');
```
brong, line 247: this could probably use a comment along the lines of "lowercase just the uri scheme" to explain why we're splitting at the colon.

imap/jmap_ical.c
```diff
@@ -241,13 +241,22 @@ static const char *sha1key(const char *val, char *keybuf)
...
+        buf_appendcstr(&buf, col);
+    }
+    else buf_setcstr(&buf, uri);
+
+    buf_trim(&buf);
+    if (!buf_len(&buf)) {
+        buf_free(&buf);
+        return NULL;
+    }
     return buf_release(&buf);
```
brong, line 260: if we wanted to do the "don't allocate memory if using a short struct buf" magic, we'd need to make APIs like this write into a buf rather than returning a malloced `char *`.  That does mean you'd need to use a flag value for "error" though.

imap/jmap_ical.c
ksmurchison: @rsto My mistake, I missed the fact that allow_organizer_attendee_only was being set in the if {} block.  A blank line would definitely help

imap/jmap_ical.c
rsto: yes, changing existing char* return types will require lots of code change. It will be somewhat easy to implement the internal optimization in struct buf. But to really make use of it, we'll have to update existing code. An initial profiling run could tell us where to start looking for the biggest gain.

ksmurchison (approved):

imap/jmap_ical.c
rsto: Right, added a comment

rsto: And the next for code style review will be to decide if a comment should go above or inside the `if` block :)

rsto:
@brong are you OK with the updated patch?

PR #3674, opened 2021-09-15
ical_support: do not use icaltimezone_get_tzid

rsto:
Both icaltime_get_tzid and icaltimezone_get_tzid cause libical to return timezone identifiers prefixed with the hardcoded value `/freeassociation.sourceforge.net/`.

This even happens if
* a custom zoneinfo directory is set
* the libical tzid_prefix is set to the empty string

Consequently, both functions should not be used within Cyrus.

We could patch our libical fork in cyruslibs, but this won't help installations that build on upstream libical versions.

Tested in https://github.com/cyrusimap/cassandane/pull/165

### DIFF
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -377,7 +377,7 @@ static int send_alarm(struct get_alarm_rock *rock,
     else
         timezone = "[floating]";
     FILL_STRING_PARAM(event, EVENT_CALENDAR_TIMEZONE,
-                      xstrdup(timezone));
+                      xstrdupsafe(timezone));
     FILL_STRING_PARAM(event, EVENT_CALENDAR_START,
                       xstrdup(icaltime_as_ical_string(start)));
     FILL_STRING_PARAM(event, EVENT_CALENDAR_END,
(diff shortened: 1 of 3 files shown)


ksmurchison (commented):
This looks fine, but I still don't see how we get prefixes if we're properly using set_zone_directory().  If this is because its use in Cyrus is gated by the tzdist module, then we should fix that.

rsto:
This is the critical code section:
https://github.com/libical/libical/blob/15cf1ddf5331f828adc59ac74be83c7b119f5542/src/libical/icaltimezone.c#L1875-L1891

Regardless of the zoneinfo value it prefixes the tzid value with the BUILTIN_TZID prefix, ignoring if we set the prefix to empty string previously: https://github.com/libical/libical/blob/master/src/libical/icaltimezone.c#L205

ksmurchison:
Perhaps we should we be calling icaltimezone_set_tzid_prefix("")

rsto:
ical_support_init already does this

dilyanpalauzov:
See also https://github.com/cyrusimap/cyrus-imapd/pull/3238 - The vendor prefix in TZID must be the empty string.

ksmurchison (approved):

PR #3675, opened 2021-09-15
Sieve iMIP

ksmurchison:
New Sieve action that will process iMIP messages

Cass test here: https://github.com/cyrusimap/cassandane/pull/167


sieve/interp.h
```diff
@@ -49,7 +49,7 @@
 struct sieve_interp {
     /* standard callbacks for actions */
     sieve_callback *redirect, *discard, *reject, *fileinto, *snooze, *keep;
-    sieve_callback *notify;
+  sieve_callback *notify, *imip;
```
dilyanpalauzov, line 52: Indentation.

elliefm (changes requested):
This will need a file in changes/next

Haven't reviewed properly yet, but writing this thought down before I forget it

sieve/script.c
```diff
@@ -841,6 +842,19 @@ static int do_action_list(sieve_interp_t *interp,
...
             break;
+        case ACTION_IMIP:
+            if (interp->imip)
+                ret = interp->imip(NULL, interp->interp_context,
+                                   script_context,
+                                   message_context,
+                                   &errmsg);
+            free(interp->lastitem);
+            interp->lastitem = NULL;
+            if (ret == SIEVE_OK)
```
elliefm, line 853: If we somehow get here without an `interp->imip` callback, then we won't set `ret` ourselves.  And if the previous action had been successful, then we'll misreport this one as success too, even though we did nothing.

I notice the ACTION_VACATION case below returns `SIEVE_INTERNAL_ERROR` if there's no callback, should this do the same?

elliefm:
MurderJMAP.frontend_commands fails on this branch, looks like the blob upload is rejected with 401 Unauthorized.  I guess it's something to do with the HTTP auth change but I don't understand what.

sieve/script.c
ksmurchison: Fixed here: https://github.com/cyrusimap/cyrus-imapd/pull/3699/commits/5bf165415498497bf4e2f81fe3a9a8b50dbdd104

ksmurchison:
Superseded by #3699

PR #3680, opened 2021-09-22
mailbox: lock by name if legacy_dirs

brong:
This makes rename/delete locks work correctly for legacy mailboxes.

I believe this fixes both the failing tests.  At least, it does for me :)

### DIFF
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -962,6 +962,7 @@ static int mailbox_open_index(struct mailbox *mailbox)
 static int mailbox_mboxlock_reopen(struct mailboxlist *listitem, int locktype, int index_locktype)
 {
     struct mailbox *mailbox = &listitem->m;
+    uint32_t legacy_dirs = (mailbox_mbtype(mailbox) & MBTYPE_LEGACY_DIRS);
     int r;
 
     mailbox_release_resources(mailbox);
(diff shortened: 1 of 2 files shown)


imap/mailbox.c
```diff
@@ -962,6 +962,7 @@ static int mailbox_open_index(struct mailbox *mailbox)
 static int mailbox_mboxlock_reopen(struct mailboxlist *listitem, int locktype, int index_locktype)
 {
     struct mailbox *mailbox = &listitem->m;
+    uint32_t legacy_dirs = (mailbox_mbtype(mailbox) & MBTYPE_LEGACY_DIRS);
```
elliefm, line 965: It's a little jarring to see a variable that's only used as a boolean being declared as an unsigned type with a fixed-size.  I guess this is because mbtype itself is uint32_t, and the compiler complains if the type doesn't match?

If we wanted to use a normal int instead, I think we could do something like:

```
    int legacy_dirs = !!(mailbox_mbtype(mailbox) & MBTYPE_LEGACY_DIRS);
```

Looks like we already use the `!!` construct for this sort of thing in a few other places, so it's not without precedent.

elliefm (commented):
My eyes don't see anything wrong here, but all of annotate.testc's tests are now failing in set_up() and I don't understand why

`FastMail.cyr_expire_delete_findpaths_legacy` now passes, but `CaldavAlarm.simple` still fails for me too.

imap/mailbox.c
brong: yeah, we could - there was already one that was this, so I just copied it

elliefm (approved):

PR #3682, opened 2021-09-22
Calalarmd use our own VTIMEZONEs

ksmurchison:
calalarmd needs to call ical_support_init() to use our own VTIMEZONEs


rsto (approved):

brong:
This appears to have solved the crasher by itself with no other changes needed.

brong:
BUT - do we need to initialise libical everywhere?  imap could theoretically touch it, and sync_server

PR #3683, opened 2021-09-22
Plug iSchedule leak

ksmurchison:
http_ischedule.c: don't leak XML response


brong (approved):
This appears to match the behaviour of all the other similar cases.  I don't quite understand why 'root' itself isn't freed but root->doc is, I guess that's how the library works.

PR #3685, opened 2021-09-26
jmap_core: update all the Blob APIs to the latest draft-ietf-jmap-blob-04 draft

brong:
This changes Blob/get to Blob/lookup, and adds the new Blob/get and Blob/set logic from the IETF draft.

### DIFF
--- a/imap/jmap_core.c
+++ b/imap/jmap_core.c
@@ -623,9 +623,10 @@ static int _parse_datatypes(jmap_req_t *req __attribute__((unused)),
 {
     int32_t *datatypesp = rock;
 
-    if (!strcmp(key, "types")) {
+    // support both "types" and "typeNames" selectors for now
+    if (!strcmp(key, "types") || !strcmp(key, "typeNames")) {
         if (!json_is_array(arg)) {
-            jmap_parser_invalid(parser, "types");
+            jmap_parser_invalid(parser, key);
             // field known, type wrong
             return 1;
         }
@@ -647,7 +648,7 @@ static int _parse_datatypes(jmap_req_t *req __attribute__((unused)),
                 *datatypesp |= typenum;
             }
             else {
-                jmap_parser_push_index(parser, "types", i, NULL);
+                jmap_parser_push_index(parser, key, i, NULL);
                 jmap_parser_invalid(parser, NULL);
                 jmap_parser_pop(parser);
             }
@@ -688,10 +689,12 @@ static int jmap_blob_lookup(jmap_req_t *req)
     }
 
     if (!datatypes) {
-        err = json_pack("{s:s s:[s]}", "type", "invalidArguments", "arguments", "types");
+        err = json_pack("{s:s s:[s]}", "type", "invalidArguments", "arguments", "typeNames");
         jmap_error(req, err);
         goto done;
     }
+
+    const char *resname = json_object_get(req->args, "typeNames") ? "matchedIds" : "types";
 
     /* Sort blob lookups by mailbox */
     hash_table getblobs_by_uniqueid = HASH_TABLE_INITIALIZER;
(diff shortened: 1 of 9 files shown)


imap/jmap_core.c
```diff
@@ -399,6 +415,171 @@ static int getblob_cb(const conv_guidrec_t* rec, void* vrock)
...
+static int _parse_xrange(jmap_req_t *req __attribute__((unused)),
+                         struct jmap_parser *parser,
+                         const char *key,
+                         json_t *arg,
+                         void *rock)
+{
+    struct blob_xrange *rangep = rock;
+
+    if (!strcmp(key, "offset")) {
+        if (json_is_integer(arg))
```
rsto, line 461: Both "offset" and "length" arguments should be checked for negative values.

imap/jmap_core.c
```diff
@@ -399,6 +415,171 @@ static int getblob_cb(const conv_guidrec_t* rec, void* vrock)
...
+            }
+            else {
+                const char *base = buf_base(&ctx.blob);
+                size_t len = buf_len(&ctx.blob);
+                if (range.offset <= len) {
+                    base += range.offset;
+                    len -= range.offset;
+                }
+                else {
+                    r = -1;
```
rsto, line 514: I assume that's copied from old code, still I think that `-1` isn't a good choice here. Instead it should be set to HTTP_SERVER_ERROR or any more appropriate error. In this particular case wouldn't it even result in an `invalidArguments` JMAP error?

Remark: for new JMAP methods, I even tend to make them return `void` and take a `json_t **err` argument. That way I force the method handler to take care or error reporting.

imap/jmap_core.c
```diff
@@ -399,6 +415,171 @@ static int getblob_cb(const conv_guidrec_t* rec, void* vrock)
...
+                else {
+                    json_array_append_new(get.not_found, json_string(blob_id));
+                }
+            }
+            jmap_getblob_ctx_fini(&ctx);
+        }
+    }
+
+    /* Reply */
+    jmap_ok(req, jmap_get_reply(&get));
```
rsto, line 575: See my comment above, we might want to `jmap_error` here in case of error.

rsto (changes requested):
Looks good, with a few caveats. Argument validation could be improved. Mainly though, it seems to me that this patch only works for regular `G` blobs? Should the `lookup` method also use getblob_ctx to allow for handling smart blob ids? Probably I just don't see it...

imap/jmap_core.c
brong: My plan here was that if there's an error, we add the blob to notFound, on the grounds that there was no blob data found within the specified range.  It would also be possible to just return `null` for the data bits if the range doesn't intersect but the blob does exist - I think that's reasonable.

imap/jmap_core.c
brong: I think if we parsed OK, then unless there's a really bad error, we just want to return notFound, but... yeah, maybe we could bail and error if there's a real system error or something.

brong:
Many updates pushed!  @rsto I've also pushed an updated copy of the RFC, and an updated copy of the MR for fastmail hm to match the new spec.

brong:
@rsto two bugfixes as found by https://github.com/cyrusimap/cassandane/pull/173.  I'll keep revising both these PRs to actually match the spec before merging them, but build the the intermediate packages for Fastmail so we can convert the middleware first.

brong:
Once again, I've churned this based on list feedback.  

@rsto, I'll ask for a final review once I've rebased everything together, but let's wait until next week for feedback at the interim.

brong:
This is now in working group last call :)  Looks like it's unlikely to change

brong:
Gonna merge this so we can rebase other work on top - if there are any more changes, we can do them later.

PR #3689, opened 2021-09-28
JMAPEmail: fix broken fromContactGroupId filter for shared addressbooks

rsto:
Before, it erroneously used the mailbox uniqueid as username to lookup contact groups.

We already had a test for shared addressbooks but it did not make it on the main branch: https://github.com/cyrusimap/cassandane/pull/168


brong (approved):
I was there when this was debugged - definitely good, and this is a bugfix for an incorrect change made not long ago.

PR #3690, opened 2021-10-01
Fix compiler warnings (for 3.4)

robohack:
- fix most missing prototype warnings (though not all properly)
- fix many (but not nearly all) const qualifier warnings
- fix shadowed variable warnings
- use ISO C90 declarations (mixing decls and code, especially with C90
  initializers, leads to code that cannot be reasoned about and which
  very often illicits impossible to predict, and incorrect, compiler
  optimizations)

Using the following compiler flags with GCC 9.3.0 (on NetBSD 9.99.x):

CWARNFLAGS='-Wall -Wextra -Wimplicit -Wreturn-type -Wswitch -Wcomment
-Wshadow -Wstrict-prototypes -Wcast-align -Wchar-subscripts
-Wmissing-declarations -Wmissing-prototypes -Wno-long-long
-Wformat-extra-args -Wundef -Wbad-function-cast -Wwrite-strings
-Wdeclaration-after-statement -Wno-stack-protector -Wno-sign-conversion'

Note the following GCC warnings cannot be used for this code:

  -Wcast-qual        ### impossible with the very bad code in skiplist
  -Wconversion       ### code is too broken
  -Wpointer-arith    ### code is too broken (void* used in arithmetic!)
  -Waggregate-return ### code returns structs, but should be ok for C90


elliefm (changes requested):
This has too many unrelated changes smashed into one commit to review properly.  Can you please break it up into multiple commits -- I would suggest one commit per class of warning fixed might be a good starting point.

I will also note that Cyrus is C99, not C90.  At a glance, there's a bunch of fixes in here (e.g. moving variable declarations to the top of the block) that are style choices in C99, not problems to fix.

There's also a bunch of tabs added.  Our style is four-space indents, no tabs.  There is a `tools/remove-tabs.pl` script in the repository that can clean these up for you -- it has no documentation, but if you can read perl you can figure it out.

Finally, 3.4 is the stable branch.  Please rebase this onto master.  Once the fixes are landed on master we can consider selectively backporting them 3.4 as necessary, but I'm not going to land this straight onto 3.4.

backup/backupd.c
```diff
@@ -894,6 +896,8 @@ static int cmd_apply_message(struct dlist *dl)
     struct sync_msgid_list *guids = sync_msgid_list_create(0);
     struct dlist *di;
```
dilyanpalauzov, line 897: I am against such changes.  The closest the variable declaration are to the first variable use, the easiest the code is to read and understand. Writing on one line both declaration and definition is just less text in total.  Yes, other places in the code do pit the declarations to the very top, but I do not like such places, too.

dilyanpalauzov:
> * use ISO C90 declarations (mixing decls and code, especially with C90 initializers, leads to code that cannot be reasoned about and which very often illicits impossible to predict, and incorrect, compiler optimizations)

Can you provide a concrete example?

robohack:
I beg to differ on almost all of everyone's points.

First off though I do agree this should be merged to master and then back-ported to supported release branches.  I would be happy to do that but I won't do that until/unless I have a concrete guarantee that it will be committed and that there will also be a commitment from project leaders to ask all developers with commit access to enable and use all relevant compiler warnings.  The Cyrus code has been maintained and extended without enough compiler warnings enabled for too many decades already.

The appearance of tab characters are intentional very carefully placed reminders (which show up quite visibly in any editor capable of hi-lighting such otherwise invisible extraneous whitespace, and which are also trivial to search for with grep) to re-indent code where new blocks are necessary.

All changes on this branch/commit are directly related.  They do not introduce (intentionally) any functional changes.  They clean up compiler warnings only.  There is no logical way to separate them that doesn't just make more work. for everyone  The diff isn't that hard for any experience C programmer to review, especially if they are familiar with the code base.  It took me about three hours to do a careful review after leaving it sit untouched for several days.

The Cyrus code is most definitely **not** C99 -- it abuses a few dubious C99 features, that is all.  Meanwhile if it is subjected to the full force of a C99-only compiler with full optimization enabled I can guarantee there are several instances of hidden Undefined Behaviour that will cause bugs.  I wouldn't be surprised if even with these fixes that it still couldn't even get up and running with a good UB sanitizer runtime (e.g. the Clang-LLVM one).  I will endeavour to run a test instance with a UB sanitizer runtime, but I'm not yet in the phase of porting that I can get any decent code coverage with the tests I've done to date.

I have nearly 40 years of experience writing C code for every type of system and application imaginable from embedded systems and operating system kernels to graphical user interfaces.  I've ported vast amounts of commercial and open-source code to at least an order of magnitude more types of extremely varied systems than exist and run today outside of museums.  Based on my experience I submit that all of these fixes are extremely necessary for a code base like this one.

Defining variables at any point other than the start of a block only leads to hard-to-find bugs and mistaken assumptions, _especially_ in code as complex and poorly factored as most of this code is.  Shadowed variables are another symptom of this practice, and they _always_ cause confusion.  There are at least two concrete examples of such bugs in the fixes on this branch, and numerous examples where shadowed variable names can clearly lead to confusion and hidden bugs.  These should all be very obvious to any reviewer who is actually familiar with the code base.  Obviously there is debate in the community about this practice, but those of us with ample experience in dealing with the consequences of such messy code disagree vehemently with everyone from the pundits who learned C++ first to those compiler vendors on the ISO committee who forced these features into C99 against the recommendations of those with ample experience on the outside of compilers.  Google for discussions about this if you want to read a wider variety of opinions -- there are many to choose from.

dilyanpalauzov:
There is no guarantee that change proposals will be considered within reasonable time.

As far as UBSAN detected errors are found, or static analyzer errors, these go beyond splitting variable declaration and variable definition because of 'compiler warnings'.  As such these could be submitted separately.

elliefm:
If you have 40 years experience then you already know we're not going to accept ~4000 lines of changes from an unknown developer without careful review, no matter how much that unknown developer insists they are trivial and obvious and easy to review.

I have requested that you break this massive single commit into reasonable, discrete pieces, so that it can be reviewed in reasonable, discrete chunks, without our eyes glazing over.  If you can provide these changes in a reviewable format, then they can be reviewed.

I'm not entering into a discussion about changing our code style here, so save your breath.  In the end, we will amend our style as we amend our style, or won't as we don't.  But whatever we do, it won't be on the basis of one contribution that can't be bothered to match what's already there.  You earn respect by showing it.

For whatever it's worth, @dilyanpalauzov does not have merge rights on this repository.  His contributions are almost always useful and appreciated, but in the end, you do not need to persuade him of anything.  Argue with him if you want, but understand that you're not moving anything forward by doing so.

robohack:
I can't imagine how anyone can think that breaking up a single-purpose change into multiple changes will in any way decrease the amount of work needed to apply, review, and test.

I can't even imagine any sensible way to break it into multiple change sets, never mind how many iterations of the same commit would be reasonable.  It's one logical set of changes to fix one problem.  The changes are very simple, non-functional, and very easy to understand and review, observe in action, and test.

Perhaps we should start with you enabling all the warnings I've used and try a build (of the un-patched source) yourself.  Be prepared for nearly a quarter million lines of output from any modern version of GCC.

Issue #2383, opened 2018-05-30
[3.0.7] Xfer for a single user doesn't work


toony:
Platform: Debian 8.10, Ubuntu 14.04.5
Cyrus: 3.0.7 from github
Build options: --libexecdir /usr/lib/cyrus --enable-autocreate --enable-replication --enable-murder --enable-idled --disable-xapian --disable-squat --with-perl --with-cyrus-user=cyrus --with-cyrus-group=mail --with-syslogfacility=MAIL --with-lock=fcntl
Configuration: altnamespace: yes, unixhierarchysep: yes, virtdomains: userid, no murder

2 cyrus: 1 on debian 8.10, 1 on ubuntu 14.04.5, proxy_password set
Cyrus partition _apr-vmnet\_loc_ exists on both side.

Trying to xfer a mailbox from one server to the other one:
`localhost> xfer user/test@domain.tld 172.16.20.55`
Mailbox isn't transfered to other server.

Cyrus logs:
> May 30 11:50:39 bluemind-debian cyrus/imap[10043]: XFER: connecting to server '172.16.20.55'
May 30 11:50:39 bluemind-debian cyrus/imap[10043]: XFER: destination supports replication
May 30 11:50:39 bluemind-debian cyrus/imap[10043]: XFER: disconnecting from servers

Xfer is ok when doing a global transfert:
`localhost> xfer * 172.16.20.55`
Cyrus logs:
> May 30 11:57:21 bluemind-debian cyrus/imap[11032]: requested partition directory for unknown partition '*'
May 30 11:57:21 bluemind-debian cyrus/imap[11032]: XFER: connecting to server '172.16.20.55'
May 30 11:57:21 bluemind-debian cyrus/imap[11032]: XFER: destination supports replication
May 30 11:57:21 bluemind-debian cyrus/imap[11032]: XFER: user 'test@domain.tld' -> 172.16.20.55!apr-vmnet_loc
May 30 11:57:21 bluemind-debian cyrus/imap[11032]: XFER: initial sync of user test@domain.tld
May 30 11:57:21 bluemind-debian cyrus/imap[11032]: USER test@domain.tld
May 30 11:57:21 bluemind-debian cyrus/imap[11032]: LOCAL_MAILBOX domain.tld!user.test
May 30 11:57:21 bluemind-debian cyrus/imap[11032]: LOCAL_MAILBOX domain.tld!user.test.Draft
[...]

PR #3693, opened 2021-10-01
XFER fixes

elliefm:
This fixes two bugs and one source of confusion in XFER:

* The `name` argument was converted to the wrong namespace before the mboxlist_findall call, so if you specified a user or mailbox name (rather than a pattern or partition name), it wouldn't find anything to do.  This was already fixed in 3.0.15, but since we didn't have any tests for XFER at the time, it wasn't clear whether the same fix was necessary or correct to forward port, so I didn't and forgot about it.  We now have such tests, and it turns out to be necessary and correct!  This is aka #2383 
* If the destination backend did not support using the replication protocol for XFER, the XFER would mostly succeed, but it would try to send a sync restart when it was finished, and hang indefinitely waiting for a response.  We no longer try to send a sync restart unless we're using the replication protocol for the XFER.
* If the `name` argument didn't match anything (e.g. due to the first bug, or just user error), previously it would do nothing, and respond "OK Completed".  Now it responds "NO Mailbox does not exist".

I expect to backport these fixes as far back as 3.0 (and have already tested them, including back and forth in mixed-version setups).

2.5 remains a bad XFER target for various reasons, but it seems to be able to XFER out, at least to 3.0.

There is still more work to do, but this is a good foundation on which to build.  So I'd like to get this much merged before diving any deeper.

These commits are tiny; the much larger part of this PR is the new Cassandane tests for XFER:  https://github.com/cyrusimap/cassandane/pull/169

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -12109,7 +12109,8 @@ static void cmd_xfer(const char *tag, const char *name,
             list.allow_usersubs = 1;
         }
 
-        mboxlist_findall(NULL, mbname_intname(mbname), 1, NULL, NULL, xfer_addmbox, &list);
+        /* admin namespace, use original name */
+        mboxlist_findall(NULL, name, 1, NULL, NULL, xfer_addmbox, &list);
     }
 
     r = xfer_init(toserver, &xfer);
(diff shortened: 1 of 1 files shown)


brong (approved):
Wow, those are all really simple looking fixes, but I know it was a lot of work to figure them out :)  Good work!

PR #3696, opened 2021-10-04
jmap_mailbox: unify mboxlist callback filters for /query and /get

rsto:
We want both methods to operate on the same set of mboxlist entries. Currently, the gap is a missing check for `mbtypes_unavailable` in Mailbox/query, which might have caused a (non-reproducible) issue a couple of weeks ago. Using the same logic in Mailbox/query and /get should avoid such errors also for future changes.

Passes all JMAP Cassandane tests.

### DIFF
--- a/imap/jmap_mailbox.c
+++ b/imap/jmap_mailbox.c
@@ -752,37 +752,46 @@ struct jmap_mailbox_get_cb_rock {
     strarray_t *sublist;
 };
 
+static int is_jmap_mailbox(const mbentry_t *mbentry, int tombstones)
+{
+    /* Don't list special-purpose mailboxes. */
+    if (mbtypes_unavailable(mbentry->mbtype) ||
+        mbtype_isa(mbentry->mbtype) != MBTYPE_EMAIL ||
+        ((mbentry->mbtype & MBTYPE_DELETED) && !tombstones))
+        return 0;
+
+    // No more returns from here
+
+    mbname_t *mbname = mbname_from_intname(mbentry->name);
+    const char *topbox = strarray_nth(mbname_boxes(mbname), 0);
+    int ret = 0;
+
+    /* skip INBOX.INBOX magic intermediate */
+    if (strarray_size(mbname_boxes(mbname)) == 1 && !strcmp(topbox, "INBOX"))
+        goto done;
+
+    /* skip any of our magic mailboxes */
+    if (!strcmpsafe(topbox, config_getstring(IMAPOPT_CALENDARPREFIX))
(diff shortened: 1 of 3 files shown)


imap/jmap_mailbox.c
```diff
@@ -752,37 +752,48 @@ struct jmap_mailbox_get_cb_rock {
...
-    json_t *list = (json_t *) rock->get->list, *obj;
-
     /* Don't list special-purpose mailboxes. */
     if (mbtypes_unavailable(mbentry->mbtype) ||
-        mbtype_isa(mbentry->mbtype) != MBTYPE_EMAIL) {
+        mbtype_isa(mbentry->mbtype) != MBTYPE_EMAIL)
         return 0;
-    }
+
+    if ((mbentry->mbtype & MBTYPE_DELETED) && !tombstones)
```
brong, line 762: This doesn't do anything, because:

```
#define mbtypes_unavailable(mbtype) \
  ((mbtype) & (MBTYPE_REMOTE | MBTYPE_RESERVE | MBTYPE_MOVING | MBTYPE_DELETED))
```

We already returned on the previous check.

imap/jmap_mailbox.c
```diff
@@ -1292,25 +1303,14 @@ static int _mboxquery_cb(const mbentry_t *mbentry, void *rock)
...
 
-    if (mbtype_isa(mbentry->mbtype) != MBTYPE_EMAIL)
-        return 0;
-
-    if ((mbentry->mbtype & MBTYPE_DELETED) && !q->include_tombstones)
+    if (!is_jmap_mailbox(mbentry, q->include_tombstones))
         return 0;
 
-    enum shared_mbox_type shared_mbtype = _shared_mbox_type(q->shared_mboxes, mbentry->name);
+    enum shared_mbox_type shared_mbtype = _shared_mbox_type(q->shared_mboxes, mbentry->name);
```
brong, line 1309: This line appears to be adding a trailing whitespace

brong (changes requested):
This appears to introduce a bug around tombstone handling, per comments.

imap/jmap_mailbox.c
rsto: Thanks! I added a commit that removes `DELETED` from the `unavailable` check in mboxlist.h. We only used it at two locations, once in jmap_mail and once here in jmap_mailbox.

imap/jmap_mailbox.c
rsto: Removed

brong (approved):
Thanks for fixing that up.  Sorry I didn't get back to this earlier.

PR #3697, opened 2021-10-04
No more strlist

brong:
Gonna ask both of you to review this.  Ken because you were the major user of strlist throughout the HTTP code, and ellie because you've been doing lots of really careful review and I appreciate it :)

There's two commits here - the first turns tok_next into a thing that returns a 'const char *', because mostly that's all we need, and it reminds us not to free or screw with it.  The couple of places that do make changes can then be much more carefully reviewed.

The second commit replaces 'struct strlist' with a strarray_t in the couple of places that it was being used.  It largely impacts the query arg parsing code and the timezone code.

This work is based on top of https://github.com/cyrusimap/cyrus-imapd/pull/3626 at which time strlist became a very boring type.

### DIFF
--- a/imap/http_tzdist.c
+++ b/imap/http_tzdist.c
@@ -97,7 +97,7 @@ static int action_expand(struct transaction_t *txn);
 static int json_response(int code, struct transaction_t *txn, json_t *root,
                          char **resp);
 static int json_error_response(struct transaction_t *txn, long tz_code,
-                               struct strlist *param, icaltimetype *time);
+                               const strarray_t *param, icaltimetype *time);
 static struct buf *icaltimezone_as_tzif(icalcomponent* comp);
 static struct buf *icaltimezone_as_tzif_leap(icalcomponent* comp);
 static struct buf *_icaltimezone_as_tzif(icalcomponent* ical, bit32 leapcnt,
@@ -816,7 +816,7 @@ static int action_capa(struct transaction_t *txn)
         if ((r = zoneinfo_lookup_info(&info))) return HTTP_SERVER_ERROR;
 
         buf_reset(&txn->buf);
-        buf_printf(&txn->buf, "%s:%s", info.data->s, info.data->next->s);
+        buf_printf(&txn->buf, "%s:%s", strarray_nth(&info.data, 0), strarray_nth(&info.data, 1));
 
         /* Construct our response */
         root = json_pack("{ s:i"                        /* version */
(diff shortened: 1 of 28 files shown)


ksmurchison (approved):
LGTM

imap/http_tzdist.c
```diff
@@ -1073,24 +1077,26 @@ static int action_list(struct transaction_t *txn)
...
 
-        if (param->next                         /* once only */
-            || strncmp(param->s, "geo:", 4)) {  /* value value */
+        if (strarray_size(param) != 1      /* once only */
+            || strncmp(val, "geo:", 4)) {  /* value value */
             return json_error_response(txn, TZ_INVALID_LOCATION, param, NULL);
         }
 
-        latitude = strtod(param->s + 4, &endptr);
+        latitude = strtod(val, &endptr);
```
elliefm, line 1099: Lost a `+4` here?  Looks like we were trying to skip over "geo:"

imap/http_tzdist.c
```diff
@@ -1133,14 +1139,15 @@ static int action_list(struct transaction_t *txn)
     }
     else if ((param = hash_lookup("changedsince", &txn->req_qparams))) {
         unsigned prefix = 0;
+        const char *val = strarray_nth(param, 0);
 
-        if (param->next) {  /* once only */
+        if (strarray_size(param) != -1) {  /* once only */
```
elliefm, line 1144: `-1`??

elliefm (changes requested):
Couple of small-but-scary things that look like typos (comments inline).  Both in http_tzdist.c, curiously, so maybe we don't have enough test coverage for that module.

More concerningly, cunit/tok.testc doesn't build.  At a glance, it looks like it's all just complaints about the API changing to `const char *` and the tests not having been updated to match yet.

Though: when I run Cassandane across this, I get core files left behind from `JMAPCore.eventsource`, `JMAPTestSuite.t:core:download`, and `JMAPTestSuiteWS.t:core:download`.  They're all "double free or corruption" during httpd `transaction_free()`, or something like that (I've lost my terminal scrollback, whoops).

For whatever it's worth, I built and tested this with the branch rebased on top of current master (9c128d0a26), and with Cassandane at ec84f488.  The failing tests are fine without this PR on top -- so perhaps there's something subtly wrong with the tok changes, and the problem is being masked by the trivial testc build problems?  I'm not sure.  I don't see anything wrong by eye, other than the http_tzdist things I already mentioned.

elliefm:
@brong Still want this?  No longer merges, but my review at the time reckoned there were a bunch of broken tests anyway.  If we wanna keep it, it'll need a rebase and some attention.

brong:
Yeah, I don't care enough to restore this :)

PR #3699, opened 2021-10-04
Sieve iMIP

ksmurchison:
New Sieve action that will process iMIP messages

Al lot of the iTIP and [Cal]DAV code was factored out into separate source files so they could be linked in by lmtpd


Makefile.am
```diff
@@ -1275,6 +1278,8 @@ imap_httpd_SOURCES = \
 	imap/http_ws.h \
 	imap/httpd.c \
 	imap/httpd.h \
+	imap/itip_support.c \
+	imap/itip_support.h \
 	imap/jcal.c \
```
dilyanpalauzov, line 1283: Since itip_support.c and caldav_util..c are used from lmtpd and httpd, it makes sense to move both files into libcyrus_imap.  It is a library for shared code.

ksmurchison: The problem with that is that there are 2 global variables (httpd_userid and namespace_calendar) that would have to defined by every binary that links libcyrus_imap.  I'd rather not have to do that.

ksmurchison: I guess its possible to include this everywhere, but do IMAP, and POP really care about iTIP?

dilyanpalauzov: The overall shared code library handling is a bit random.  E.g. spool.c is included in libcyrus_imap and as source file in httpd and others.  I think the overall criterion to include something in a shared library is, if a source file is needed by more than one executable. The httpd_userid and imapd_userid global variables serve the same purpose, so they can share also the same name.

I think the ultimate goal shall be to move all shared code in at most two libraries.  As far as I remember I implemented autocreate (of the imap and sieve user) on the first CalDAV login and this made libcyrus_imap depend on libcyrus_sieve and libcyrus_sieve depend on libcyrus_imap…

imap/zoneinfo_db.c is another candidate to put in a shared library.

As a rule of thumb global symbols are not good.  If the fatal() implementation can be put (as WEAK symbol in some shared library, them most of the fatal() implementations could be removed.

imap/caldav_db.c
```diff
@@ -155,6 +155,17 @@ EXPORTED struct caldav_db *caldav_open_userid(const char *userid)
     /* Construct mbox name corresponding to userid's scheduling Inbox */
     caldavdb->sched_inbox = caldav_mboxname(userid, SCHED_INBOX);
 
+    if (db->version >= DB_MBOXID_VERSION) {
+        /* Lookup mailbox ID of scheduling Inbox */
+        mbentry_t *mbentry = NULL;
+        int r = mboxlist_lookup(caldavdb->sched_inbox, &mbentry, NULL);
+        if (!r) {
+            free(caldavdb->sched_inbox);
```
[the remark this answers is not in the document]
ksmurchison: Done

imap/caldav_util.c
```diff
@@ -0,0 +1,996 @@
...
+                                              strarray_t *schedule_addresses)
+{
+    icalcomponent *ical = NULL;
+    struct index_record record;
+    int r;
+
+    /* Fetch index record for the cal resource */
+    r = mailbox_find_index_record(mailbox, cdata->dav.imap_uid, &record);
+    if (r) return NULL;
+
```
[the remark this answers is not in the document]
ksmurchison: Done

Makefile.am
ksmurchison: @brong Are you still wanting code changes?  Are you happy with the tests in https://github.com/cyrusimap/cassandane/pull/167?

imap/caldav_util.c
```diff
@@ -0,0 +1,1179 @@
...
+    return ret;
+}
+
+
+/*
+ * Handle stripping per-user data from existing and/or new shared resource.
+ *
+ * Logic is as follows:
+ *
```
dilyanpalauzov, line 549: terminating space

sieve/sieve-lex.l
```diff
@@ -317,7 +317,7 @@ ws              [ \t]+
     /* notify - draft-martin-sieve-notify-01 */
 <INITIAL>denotify             return DENOTIFY;
 <INITIAL>:method              return METHOD;
-<INITIAL>:id                  return ID;
+<INITIAL>:id                  return NID;
 <INITIAL>:low                 return yylval->nval = LOW;
```
dilyanpalauzov, line 321: Perhaps sieved.c shall be changed to print NID instead of ID?

sieve/sieve_interface.h
```diff
@@ -97,6 +97,18 @@ typedef int sieve_list_comparator(const char *text, size_t tlen,
...
+    unsigned updates_only    : 1;
+    unsigned delete_canceled : 1;
+    const char *calendarid;
+    struct buf errstr;
+} sieve_imip_context_t;
+
+typedef int sieve_processimip(void *interp_context, void *script_context,
+                              void *message_context,
+                              sieve_imip_context_t *imip_context);
+
```
dilyanpalauzov, line 111: Many terminating spaces.

sieve/sieve.y
```diff
@@ -3335,3 +3439,23 @@ static test_t *build_jmapquery(sieve_script_t *sscript, test_t *t, char *json)
...
+    unsigned flags = 0;
+
+    assert(t && t->type == BC_PROCESSIMIP);
+
+    if (t->u.imip.invites_only) flags |= IMIP_INVITESONLY;
+    if (t->u.imip.updates_only) flags |= IMIP_UPDATESONLY;
+    if (t->u.imip.delete_canceled) flags |= IMIP_DELETECANCELED;
+
+    if (t->u.imip.errstr_var) verify_identifier(sscript, t->u.imip.errstr_var);
+
```
dilyanpalauzov, line 3454: terminating space

sieve/sieve-lex.l
ksmurchison: No, because the actual parameter is :id.  Support for draft-martin-sieve-notify-01 will probably be stripped out soon.

Makefile.am
brong: Have replied on that - the tests look OK.  Gonna read the code for careful now.

imap/itip_support.c
```diff
@@ -0,0 +1,1053 @@
...
+    if (!attendee) return 1; // something weird is going on, treat it as a change
+    return (get_partstat(oldcomp, attendee) != get_partstat(newcomp, attendee));
+}
+
+/* Returns the calendar collection name to use as scheduling default.
+ * This is just the *last* part of the complete path without trailing
+ * path separator, e.g. 'Default' */
+HIDDEN char *caldav_scheddefault(const char *userid)
+{
+    const char *annotname =
```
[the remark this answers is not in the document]
brong: There's probably a ton of places where that change could be made throughout the code if we wanted.  We do use this pattern a lot.  It's probably not worth doing it one-off here though, because that just makes more different versions to look for if we decide on a different pattern than this.

sieve/message.h
```diff
@@ -67,7 +67,7 @@ typedef enum {
     ACTION_UNMARK,
     ACTION_ENOTIFY,
     ACTION_NOTIFY,
-    ACTION_DENOTIFY
+    ACTION_DENOTIFY,
```
brong, line 70: Is this extra trailing comma deliberate?  Presumably it might annoy some compilers?

brong (approved):
I haven't reviewed the large swathes of copy-pasted code to make sure that they didn't change at all during the copy/pastes - and I reviewed the change as a whole rather than individual commits.

I think this design looks good, and this is ready to work with.  Couple of maintainability comments inline.

sieve/message.h
elliefm: This is perfectly valid, and excellent C style (C++ and perl too).  I do this all the time, deliberately, and have done so for years.  Future additions to this list will diff as added lines only, which is very readable.

Without the trailing comma, the next time someone adds to the list, the diff will have a noisy delete-and-two-adds, because you can't just add to the list, you also have to modify the old last one.

(JavaScript notably doesn't like this sort of thing; you might have tripped over that in the past.)

PR #3704, opened 2021-10-06
update README.md

### LINKED ISSUE #2983: README.md needs updating
The versions of this file in the version branches need to be checked for correctness wrt that version

The version in master needs to be checked for correctness wrt master, but also the project generally, as it is what is shown on the repository main page!

At a glance, it looks like it hasn't been updated since 3.0 was released.  Needs clarification about what type of bug/security fixes to expect for the versions that are supported (and which versions are supported)

elliefm:
Mostly motivated by the list of current and supported versions being wildly out of date, but I also fixed up a few other details while in there anyway.

There's two commits here: one has real changes; the other just fixes up some trailing whitespace and wraps long lines.  If you're viewing the diffs on github, adding `?w=1` to the end of the URL will hide lines where only whitespace changed (just like `git diff -w`).  Or just view one commit at a time.

Closes #2983 (though I will need to backport similar fixes to the other branches).

I'd like review from _both_ @brong and @rjbs on this one please.  Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,51 +17,58 @@ active development.
 What is Cyrus
 =============
 
-Cyrus is an IMAP server, where IMAP (Internet Message Access Protocol) 
+Cyrus is an IMAP server, where IMAP (Internet Message Access Protocol)
 is a protocol for accessing mail.
 
-The Cyrus IMAP server differs from other IMAP server implementations in 
-that it is generally intended to be run on "sealed" servers, where 
-normal users are not permitted to log in. The mailbox database is stored 
-in parts of the filesystem that are private to the Cyrus IMAP system. 
-All user access to mail is through the IMAP, NNTP, or POP3 protocols. 
+The Cyrus IMAP server differs from other IMAP server implementations in
+that it is generally intended to be run on "sealed" servers, where
+normal users are not permitted to log in. The mailbox database is stored
+in parts of the filesystem that are private to the Cyrus IMAP system.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,62 +6,69 @@
...
-The Cyrus IMAP server differs from other IMAP server implementations in 
-that it is generally intended to be run on "sealed" servers, where 
-normal users are not permitted to log in. The mailbox database is stored 
-in parts of the filesystem that are private to the Cyrus IMAP system. 
-All user access to mail is through the IMAP, NNTP, or POP3 protocols. 
+The Cyrus IMAP server differs from other IMAP server implementations in
+that it is generally intended to be run on "sealed" servers, where
+normal users are not permitted to log in. The mailbox database is stored
+in parts of the filesystem that are private to the Cyrus IMAP system.
+All user access to mail is through the IMAP, NNTP, or POP3 protocols.
```
brong, line 27: and JMAP!  It's an RFC protocol.

Cyrus also has an HTTP server and CalDAV/CardDAV implementations.

brong (approved):
This looks great as an immediate fixup.  Thanks.  We could write more about the other new features too.

README.md
rjbs, line 27:
```suggestion
All user access to content is through JMAP, IMAP, NNTP, POP3, CalDAV, CardDAV,
and WebDAV protocols.
```

As per Bron's suggestion.

PR #3707, opened 2021-10-11
cyr_dbtool: fix batch mode leaks

elliefm:
Fixes a few simple leaks in cyr_dbtool's batch mode, found by valgrinding Cassandane.

Looks like these leaks have existed for ages, but for some reason Cassandane only recently started triggering them -- probably due to the addition of the `run_dbcommand()` APIs that use it.


brong (approved):
Yeah, that looks great - I bet I was just lazy about cleaning up memory because the process closes immediately afterwards anyway.

PR #3709, opened 2021-10-12
Charset decode sha1

brong:
This is an efficiency gain for the GUID key creation during message parsing - instead of having to malloc space to decode the message, we can do it in constant space.

No new cassandane tests, but it does add cunit tests around the sha1 framework, which exercise both decode codepaths across both the internal buffer boundary and the internal sha1 engine buffers, as well as the testcases from the Wikipedia page with all the codepaths :)


lib/charset.h
```diff
@@ -142,6 +142,8 @@ extern char *charset_unfold(const char *s, size_t len, int flags);
 
 extern int charset_decode(struct buf *dst, const char *src, size_t len, int encoding);
 
+extern int charset_decode_sha1(uint8_t *dest, size_t *outlen, const char *src, size_t len, int encoding);
```
rsto, line 145: I find the name `outlen` confusing, I wondered why it is needed at all a 20-byte SHA1. Renaming it to `decodedlen` could avoid that confusion.

imap/message.c
```diff
@@ -738,21 +738,20 @@ message_header_lookup(const char *buf, const char **valp)
 static void body_add_content_guid(const char *base, struct body *body)
 {
     int encoding = ENCODING_NONE;
-    char *decbuf = NULL;
+    uint8_t dest[20];
```
rsto, line 741: `xsha1.h` already defines the macro `SHA1_DIGEST_LENGTH`, prefer to use it here instead of magic `20`?

brong: Yep, updated here and in the cunit tests too

elliefm (approved):
Looks good; few comments inline

PR #3712, opened 2021-10-13
http_ws: fix invalid read during clientless shutdown

ksmurchison:


imap/httpd.h
```diff
@@ -344,7 +344,8 @@ struct http_connection {
 
     void *tls_ctx;                      /* TLS context */
     void *sess_ctx;                     /* HTTP/2+ session context */
-    void **ws_ctx;                      /* WebSocket context (HTTP/1.1 only) */
+    const void *ws_ctx;                 /* Reference to WebSocket context
+                                           (HTTP/1.1 only) */
```
elliefm, line 348: I think this doesn't need to be `void` anymore either, since it's not used polymorphically, so it could just be a `const ws_context *`... though everything else here is `void *`, and maybe it's easier to just keep thinking about `void *` once you are already.

elliefm (commented):
Cool, this is a better approach than mine, I'll close mine.

I'm not certain about the `_end_channel()` bit specifically (commented inline) -- but if that's fine actually, then this is approved.  I did run it through all the tests, as well as JMAPTestSuiteWS with valgrind, and it all passes, I'm just a little worried about a possible edge case that our tests might not exercise.

imap/httpd.h
ksmurchison: Its easier to keep it void* in case we're compiled without wslay.

imap/http_ws.c
```diff
@@ -708,15 +708,17 @@ static void _end_channel(struct transaction_t *txn)
 
     free(ctx);
 
-    txn->ws_ctx = NULL;
+    /* Clear all references to WS context */
+    txn->conn->ws_ctx = txn->ws_ctx = NULL;
```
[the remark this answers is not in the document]
ksmurchison: Its 1:1 for HTTP/1.1, which is the only time conn->ws_ctx is set.

PR #3718, opened 2021-10-23
vCard validation - when one of UID/FN/VERSION is missing, spell exactly what is missing

dilyanpalauzov:
After such utterly clear error messages, fixing any problem is easier.


ksmurchison:
This is resolved with the addition of the vCard4 code

PR #3722, opened 2021-10-26
http_caldav_sched.c:propcmp() compare the DATE-TIME properties by absolute value

dilyanpalauzov:
The effect is that changing the DTSTART timezone and value do not trigget iTIP message, if the start/end/due times converted to UTC do not change.

Comparing only the value with icalproperty_get_value_as_string() ignores the TZ information and does not trigger a change, when the timezone changes, the relative value remains the same, but the value converted to UTC does change.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1624,6 +1624,8 @@ dnl
                 AC_MSG_ERROR([Need libical 3.0.0 for http]))
 
         dnl icalproperty_new_color was overlooked, and eventually added in 3.0.5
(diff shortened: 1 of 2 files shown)


rsto:
Closed in favor of https://github.com/cyrusimap/cyrus-imapd/pull/5057

dilyanpalauzov:
> On the other side, if the initial value is changed to:
>
>>    DTSTART;TZID=Europe/Sofia:20210101T000000
> 
> as far as I remember no iTIP message was triggered, but there should be one.

This was addressed in 2022 by ac18bf3163ea9cc657435d9.

PR #3723, opened 2021-10-26
http_caldav_sched.c:imip_send_sendmail() No need for <> in From: and To: when display-name is absent

dilyanpalauzov:
No need to render the display-name, when it is identical to the email address.  The latter does happen, when some CUA inserts under ATTENDEE/ORGANIZER’s CN= property parameter value the property value (with mailto: stripped).


dilyanpalauzov:
My understanding is that this was addressed by `format_address()` in https://github.com/cyrusimap/cyrus-imapd/pull/4956/.

As this change is trivial, handling is slightly faster would have been reasonable.

PR #3724, opened 2021-10-27
imtest.c: resolve -Wdiscarded-qualifiers -Wwrite-strings warnings

dilyanpalauzov:
Compiling with CFLAGS set to the above produces many warnings on imtest.c .

### DIFF
--- a/imtest/imtest.c
+++ b/imtest/imtest.c
@@ -158,18 +158,18 @@ struct protocol_t;
 
 struct banner_t {
     u_char is_capa;     /* banner is capability response */
-    char *resp;         /* end of banner response */
+    const char *resp;   /* end of banner response */
     void *(*parse_banner)(char *str);
                         /* [OPTIONAL] parse banner, returns 'rock' */
 };
 
 struct capa_cmd_t {
-    char *cmd;          /* capability command string (NULL = no capa cmd) */
-    char *resp;         /* end of capability response */
-    char *tls;          /* [OPTIONAL] TLS capability string */
-    char *login;        /* [OPTIONAL] plaintext login cmd capability string */
-    char *auth;         /* [OPTIONAL] AUTH (SASL) capability string */
-    char *compress;     /* [OPTIONAL] COMPRESS capability string */
+    const char *cmd;      /* capability command string (NULL = no capa cmd) */
+    const char *resp;     /* end of capability response */
+    const char *tls;      /* [OPTIONAL] TLS capability string */
+    const char *login;    /* [OPTIONAL] plaintext login cmd capability string */
+    const char *auth;     /* [OPTIONAL] AUTH (SASL) capability string */
+    const char *compress; /* [OPTIONAL] COMPRESS capability string */
     void (*parse_mechlist)(struct buf *list, const char *str,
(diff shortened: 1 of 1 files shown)


rsto:

elliefm:
I fetched the branch and resolved the conflicts locally, and these changes look good to me by eye.  They do indeed resolve the discarded qualifiers warnings when compiling imtest with `-Wwrite-strings`.  But I don't think we have anything that tests `imtest`, so I can't tell whether it still _works_, or merely compiles.  @ksmurchison I think I recall you use imtest at times, any thoughts on testing it?

I did some reading about `-Wwrite-strings`.  Turns out it's not really a warning option, but rather it enables a non-standard C dialect in which string literals have type `const char[]`, which is similar to how C++ works.  When you misuse a string literal in this dialect, you get a "discarded qualifiers" warning.  In standard C, string literals have type `char[]`, so there's no warning when they're misused, it will just crash if anything tries to write to them. Apparently the C standards people decided updating this detail to match C++ would break too much old code.

I would be tempted to propose we add `-Wwrite-strings` to our standard build options, because once all the warnings it produces are fixed, we will have compile time protection against misused string literals, instead of them being sources of potential crashes.  But the C standards people weren't wrong... it does break a lot of old code!  Here's the counts of problems detected when I build master with this option turned on:

```
$make -j4 -k 2>&1 | grep error: | cut -d: -f1 | sort | uniq -c
      1 ./config.h
     14 imap/autocreate.c
      3 imap/ctl_cyrusdb.c
      2 imap/cyr_dbtool.c
      1 imap/deliver.c
      3 imap/fetchnews.c
      6 imap/http_caldav.c
      4 imap/http_caldav_sched.c
      4 imap/http_carddav.c
      6 imap/http_dav.h
      4 imap/http_dav_sharing.c
      2 imap/httpd.c
      6 imap/http_rss.c
      4 imap/http_webdav.c
      2 imap/idled.c
      1 imap/imapd.c
      1 imap/index.c
      1 imap/itip_support.h
      1 imap/jmap_mail.c
      1 imap/lmtpd.c
      9 imap/mbdump.c
      6 imap/message.c
      1 imap/mupdate-client.c
     11 imap/nntpd.c
      1 imap/pop3d.c
      3 ./imap/prometheus.h
      1 imap/quota_db.c
      3 imap/smtpclient.c
      2 imap/sync_support.c
      4 imap/tls.c
      4 imap/xcal.c
    195 imtest/imtest.c
    488 lib/chartable.c
      8 lib/imclient.c
      6 lib/imparse.c
      4 lib/libconfig.c
      2 lib/parseaddr.c
      1 notifyd/notifyd.c
      2 notifyd/notify_log.c
      2 ptclient/ldap.c
      1 sieve/bc_eval.c
     28 sieve/script.c
      3 sieve/sieved.c
      7 sieve/sieve-lex.l
      1 timsieved/parser.c
```

That'll be a big job to clean all that up, but maybe it's worth it for the protection `-Wwrite-strings` will get us.  imtest is the second worst culprit in this regard, so it'll be good to have that fixed at least.

elliefm (approved):

elliefm:
Manually merged as b93fac6b6

Issue #3694, opened 2021-10-03
Update EmailSubmission/set identityId handling


robmueller:
Based on the discussion we had:

1. cyrus no longer requires the identityId in EmailSubmission/set.create to be the same as the cyrus username, it must be present but can be any string value
2. When cyrus tries to send the email (or tries to validate the email at EmailSubmission/set creation time) using it's SMTP module, if in response to EHLO it sees the JMAPIDENTITY extension, it passes the EmailSubmission/set.create identityId parameter as an extra SMTP MAIL FROM parameter of IDENTITY=X where X is the xtext encoded string from the EmailSubmission identityId parameter.

That seems reasonable to me. A bunch of edge cases of course:

1. If envelope.mailFrom.parameters.identity ("identity" being any case e.g identity, Identity, IDENTITY, IdEnTiTy, etc) exists, EmailSubmission/set should return an invalidProperties error
2. The IDENTITY=X MAIL FROM parameter is in addition to any MAIL FROM parameters passed in the EmailSubmission/set.create envelope.mailFrom.parameters object
3. If the JMAPIDENTITY SMTP extension isn't present, then the EmailSubmission/set.create identityId parameter is effectively ignored and lost

@rjbs Tagging you to make sure I haven't missed anything.

PR #3728, opened 2021-10-27
Fix EmailSubmission/set identityId handling

ksmurchison:
For https://github.com/cyrusimap/cyrus-imapd/issues/3694


brong (approved):
Looks good - this fixes delivery problems for events in shared calendars, right?  Nice.

PR #3730, opened 2021-10-28
JMAPEmail: include Trash unseen counts in threadKeyword search

rsto:
Up until now, evaluation of the threadKeyword query criteria did not take the `Seen` state of messages in the "Trash"
mailbox into account. The JMAP specification does not define this special treatment for the Trash mailbox.

This patch updates threadKeyword search to include the `Seen` state of messages within Trash. To preserve former behaviour, queries must explicitly exclude the "Trash" mailbox, using either an `inMailboxOtherThan` or a negated `inMailbox` criteria. For such queries, the `Seen` count of messages with "Trash" is ignored.

Note that this patch only applies to the hard-coded "Trash" mailbox name. Mailboxes with other name and `role=trash` always counted towards the total `Seen` count of a thread.

To make this work for guidsearch,  DNF search expression normalisation also removes trivial conditionals from the search tree.

Tested in https://github.com/cyrusimap/cassandane/pull/180

### DIFF
--- a/cassandane/Cassandane/Cyrus/JMAPEmail.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPEmail.pm
@@ -19492,7 +19492,15 @@ EOF
     }, $res->[0][1]{list}[0]{mailboxIds});
     $self->assert_equals(JSON::false, $res->[1][1]{performance}{details}{isGuidSearch});
     $self->assert_deep_equals([], $res->[1][1]{ids});
-    $self->assert_equals(JSON::true, $res->[2][1]{performance}{details}{isGuidSearch});
+    my ($maj, $min) = Cassandane::Instance->get_version();
+    if ($maj < 3 || ($maj ==3 && $min < 5)) {
+        $self->assert_equals(JSON::true, $res->[2][1]{performance}{details}{isGuidSearch});
+    }
+    else {
+        # Due to improved JMAP Email query optimizer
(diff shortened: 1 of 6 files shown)


brong (approved):
The logic looks good - the only question (I've emailed about it) is whether we want this behaviour change.

neilj:
@brong Yes this is the right behaviour; the unread status of the messages in trash is still ignored if the trash folder is excluded from the search (which is the default), but this fixes global unread searches that include trash.

@rsto I would have thought we should be looking at the role attribute rather than the name to determine the Trash folder? (It's always called "Trash" in Fastmail so not really a problem, but would be better for localisation. But perhaps it's hardcoded in a bunch of places?)

rsto:
@neilj I also think that we should look at the Trash role, rather than the hardcoded name. The decision to hardcode "Trash" was taken for conversations.db a good while ago. We could change that, but it bears of the risk of running into lock inversions with annotations.db. @brong what do you prefer?

neilj:
Right, I thought it would be something like that. Probably best to just leave it for now then; I think there was a plan to redo conversations.db in the future anyway (merging it with a bunch of other indexes)?

rsto:
Yes, merging a couple of databases in on the roadmap. That would be a good occasion to fix this.

brong:
@rsto I rebased this on master and force-pushed the update, it was clashing with another JMAPEmail change

PR #3736, opened 2021-10-30
Remove unreachable-code

dilyanpalauzov:
Reported by clang 12 called with

CFLAGS="-Wunreachable-code -Wunreachable-code-aggressive -Wunreachable-code-break -Wunreachable-code-loop-increment -Wunreachable-code-return"
CXXFLAGS="-Wunreachable-code -Wunreachable-code-aggressive -Wunreachable-code-break -Wunreachable-code-loop-increment -Wunreachable-code-return"

### DIFF
--- a/backup/ctl_backups.c
+++ b/backup/ctl_backups.c
@@ -1026,7 +1026,6 @@ static int lock_run_sqlite(const char *userid, const char *fname,
         /* execlp never returns */
         perror("execlp sqlite3");
         _exit(EX_SOFTWARE);
-        break;
 
     default:
         /* parent */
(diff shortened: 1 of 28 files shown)


mildred:
This is mostly break statements and control flow statement. I see no value of removing them, if for any reason they become reachable, they guard against other bugs.

rsto:
Most of these changes are too aggressive, such as removing `break` from case without clear benefit. `-Wunreachable-code-return` looks more useful but makes my autoconf configure script bail, so let's get not down that rabbit hole either.

dilyanpalauzov:
> removing break from case without clear benefit. 

`return; break;` or `exit(); break;` is the same as having `break; break;`.  The benefit are less compiler warnings on `clang -Weverything` and less lines of code, without changing functionality.

PR #3737, opened 2021-10-30
Annotate usage(), service_abort() with __attribute__((noreturn))

dilyanpalauzov:
as suggested by clang 12 -Wmissing-noreturn

Make dav_reconstruct.c:usage() and cyr_synclog.c:usage() static.

In imapd.c, lmtpd.c and mupdate.c the declarations of shut_down() were not correct.

### DIFF
--- a/backup/backupd.c
+++ b/backup/backupd.c
@@ -208,7 +208,7 @@ EXPORTED int service_init(int argc __attribute__((unused)),
 }
 
 /* Called by service API to shut down the service */
-EXPORTED void service_abort(int error)
+EXPORTED __attribute__((noreturn)) void service_abort(int error)
 {
     shut_down(error);
 }
--- a/backup/ctl_backups.c
+++ b/backup/ctl_backups.c
@@ -75,7 +75,7 @@ EXPORTED void fatal(const char *error, int code)
 }
 
 static const char *argv0 = NULL;
-static void usage(void)
+static __attribute__((noreturn)) void usage(void)
 {
     fprintf(stderr, "Usage:\n");
     fprintf(stderr, "    %s [options] compact [mode] backup...\n", argv0);
--- a/backup/cyr_backup.c
+++ b/backup/cyr_backup.c
@@ -78,7 +78,7 @@ EXPORTED void fatal(const char *error, int code)
 }
 
 static const char *argv0 = NULL;
-static void usage(void)
+static __attribute__((noreturn)) void usage(void)
 {
     fprintf(stderr, "Usage:\n");
     fprintf(stderr, "    %s [options] [mode] backup list chunks\n", argv0);
--- a/backup/restore.c
+++ b/backup/restore.c
@@ -71,7 +71,7 @@ EXPORTED void fatal(const char *s, int code)
 }
 
 static const char *argv0 = NULL;
-static void usage(void)
+static __attribute__((noreturn)) void usage(void)
 {
     fprintf(stderr, "Usage:\n");
     fprintf(stderr, "    %s [options] server [mode] backup [mboxname | uniqueid | guid]...\n", argv0);
--- a/imap/chk_cyrus.c
+++ b/imap/chk_cyrus.c
@@ -63,7 +63,7 @@
 #include "map.h"
 #include "xmalloc.h"
 
-static void usage(void)
+static __attribute__((noreturn)) void usage(void)
 {
     fprintf(stderr, "chk_cyrus [-C <altconfig>] partition\n");
     exit(-1);
--- a/imap/ctl_cyrusdb.c
+++ b/imap/ctl_cyrusdb.c
@@ -118,7 +118,7 @@ static int compdb(const void *v1, const void *v2)
     return ((char *)db1->archiver - (char *)db2->archiver);
 }
 
-static void usage(void)
+static __attribute__((noreturn)) void usage(void)
 {
     fprintf(stderr, "ctl_cyrusdb [-C <altconfig>] -c\n");
     fprintf(stderr, "ctl_cyrusdb [-C <altconfig>] -r [-x]\n");
--- a/imap/ctl_deliver.c
+++ b/imap/ctl_deliver.c
@@ -59,7 +59,7 @@
 #include "util.h"
 #include "xmalloc.h"
 
-static void usage(void)
+static __attribute__((noreturn)) void usage(void)
 {
     fprintf(stderr,
             "ctl_deliver [-C <altconfig>] -d [-f <dbfile>]\n");
--- a/imap/ctl_mboxlist.c
+++ b/imap/ctl_mboxlist.c
@@ -882,7 +882,7 @@ static void do_verify(void)
     mboxlist_allmbox("", &verify_cb, &found, MBOXTREE_TOMBSTONES);
 }
 
-static void usage(void)
+static __attribute__((noreturn)) void usage(void)
 {
     fprintf(stderr, "DUMP:\n");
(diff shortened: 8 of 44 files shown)


elliefm:
This is fine, but for `service_abort()`, I wonder if it's enough to just add the `__attribute__((__noreturn__))` to the declaration in master/service.h, rather than needing to add it to every definition:

https://github.com/cyrusimap/cyrus-imapd/blob/12094319c334c9aec6215cf2f444b42c6fe940f1/master/service.h#L60

I don't have clang lying around, otherwise I'd just try it myself

dilyanpalauzov:
`#include "service.h"` exists only in master/master.c:, master/service-thread.c and master/service.c .

rsto:
I am against merging this, but maybe I do not understand its benefit. Maybe @elliefm or @ksmurchison want to chime in, too.

This is what the current [GCC documentation (14.2.0)](https://gcc.gnu.org/onlinedocs/gcc-14.2.0/gcc/Common-Function-Attributes.html) states about `noreturn`:

> A few standard library functions, such as abort and exit, cannot return. GCC knows this automatically. Some programs define their own functions that never return. [...] The noreturn keyword tells the compiler to assume that [function] cannot return. It can then optimize without regard to what would happen if [function] ever did return. This makes slightly better code. More importantly, it helps avoid spurious warnings of uninitialized variables. 

I do not see "spurious warnings of uninitialized variables" by compiling the current codebase with gcc (Debian 12.2.0-14)  with `-Wall,-Werror` so this does not seem to apply here. That leaves "slightly better code", but that's quite vague.

In addition, we do not have the tooling set up to assert that validity of these assumptions. Just check out this PR, which shows the difference how gcc and clang deal with this attribute: https://github.com/cyrusimap/cyrus-imapd/pull/5056. This built fine with gcc with anyone noticing.

dilyanpalauzov:
[Clang adds](https://clang.llvm.org/docs/AttributeReference.html#noreturn-noreturn):

>  The compiler will generate a diagnostic for a function declared as [[noreturn]] that appears to be capable of returning to its caller.

I do not understand your argument against. Can you explain it again, why noreturn is used on some places, but should not be used on all places, where appropriate?

> In addition, we do not have the tooling set up to assert the validity of the assumption that a function does not return.

A function with `noreturn` generates better code.  Why do you need tooling to detect, that better code was generated?

> … Just check out this PR, which shows the difference how gcc and clang deal with this attribute: https://github.com/cyrusimap/cyrus-imapd/pull/5056. This built fine with gcc with anyone noticing.

Let me ask rhetorically, if somebody notices things, but nothing happens with the findings three or five years, why should somebody look for discrepancies, such as in https://github.com/cyrusimap/cyrus-imapd/pull/5056?

I see now [gcc -Wsuggest-attribute=noreturn](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wsuggest-attribute_003d).  Is this the tooling you asked for?

rsto:
> [Clang adds](https://clang.llvm.org/docs/AttributeReference.html#noreturn-noreturn):
> 
> > The compiler will generate a diagnostic for a function declared as [[noreturn]] that appears to be capable of returning to its caller.
> 
> I do not understand your argument against. Can you explain it again, why noreturn is used on some places, but should not be used on all places, where appropriate?

I can not, but I am also not the one who had started putting `noreturn` in the codebase seemingly inconsistently. I am just against introducing more of this or any other attributes, unless we *first* agree that they bring more value than effort. By effort I mean setting up our CI to actually deal with this, make sure that it compiles fine under at least clang and gcc using the relevant compiler diagnostics, update the developer documentation. At least @elliefm and @ksmurchison should have their say in that, too.

> Let me ask rhetorically, if somebody notices things, but nothing happens with the findings three or five years, why should somebody look for discrepancies, such as in #5056?

I noticed because it broke my build with clang-14 using `-Wall,-Werror`.

> I see now [gcc -Wsuggest-attribute=noreturn](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wsuggest-attribute_003d). Is this the tooling you asked for?

Yes, that's part of what I meant with tooling.

ksmurchison:
I have a couple general observations/questions:

1. Should we be using the standard C syntax for attributes, or continue to use the GNU syntax?
2. I really dislike having the prepended attributes to the function declaration on the same line.  Having to look in the middle of the line for the function is really annoying to me.  I'd prefer to have the prepended attributes on the preceding line.

In the case of `noreturn`, I'm fine using it, provided we use it consistently, and subject to my notes above.  As far as using other attributes, I agree with Robert that the level of optimizations to the code needs to outweigh cost of updating and maintaining the code.  Also, using GNU-specific attributes is a non-starter for me, as I don't want to dictate what compiler needs to be used.

rsto:
I'm fine with `unused` and `noreturn` too, but only if we actually treat them as any other code element, and breaking them should mean breaking the builds and rejecting to merge that code, too. That means we need to update our CI to actually check for these attributes to be used appropriately.

Just look at https://github.com/cyrusimap/cyrus-imapd/pull/3744: we plastered these `__unused__` attributes across the codebase to make gcc's `-Wunused` (as part of `-Wall`) diagnostics happy. But then breaking the actual semantics of this attribute is silently ignored by the compiler? What's even the equivalent of clangs `-Wused-but-marked-unused warnings` in gcc, if there is any?

I'm generally fine with tightening the builds with these diagnostics, but we should make an informed decision which diagnostics we want. Then we establish them as baseline in CI and require any new code contributions to actually meet their criteria. One-off pull requests that fix these diagnostics for a snapshot in time without further enforcement are no real solution.

dilyanpalauzov:
Both clang and gcc have `-Wno-attributes`, so compilers have option to disable (this particular) warning.  The current code:
```
  CC       imap/libcyrus_imap_la-partlist.lo
../imap/partlist.c: In function 'partlist_selectpart_index':
../imap/partlist.c:287:27: warning: cast from function call of type 'int' to non-matching type 'double' [-Wbad-function-cast]
  287 |             val = 100. * ((double)rand() / (RAND_MAX + 1.));
      |                           ^        
  CC       imap/libcyrus_imap_la-search_expr.lo
../imap/search_expr.c: In function 'maxcost':
../imap/search_expr.c:811:25: warning: cast from function call of type 'void *' to non-matching type 'long int' [-Wbad-function-cast]
  811 |         intptr_t cost = (intptr_t) hashu64_lookup((uint64_t)(uintptr_t) e, costcache);
      |                         ^
  CC       imap/libcyrus_imap_la-seen_db.lo
```
shows how easy it is to find out what to pass to the compiler to silence other warnings today (by adding `-Wno-bad-function-cast` to CFLAGS/CXXFLAGS).  The parameters to pass to the compiler, when compiling Cyrus IMAP, are anyway not managed by Cyrus IMAP itself.  That said, I do not agree that using attributes imposes anyhow what compiler to use.

I find no equivalent in GCC for clang’s `-Wused-but-marked-unused`, but I do not think this is relevant here.

I also do not think, that only warnings which are triggered by the CI should be handled, as the versions behind running the CI can get out of date, but this is not yet a criterion to do or not a reasonable change.  Irrespective of this, it will be good, if the versions of all the software in the CI is updated (yearly).  In addition, I cannot say anything about the CI criteria, as I have no impact on the CI itself, but at the same time I state that for exactly this reason the CI should be an irrelevant criterion.

Adding attribute for ignored return values can show memory leaks as in https://github.com/cyrusimap/cyrus-imapd/issues/4061.

elliefm:
This is useful.  I've redone it on current master as #5069

PR #3744, opened 2021-10-31
Silence clang -Wused-but-marked-unused warnings

dilyanpalauzov:


rsto:
Would you still want to get this merged? If so, would you rebase to latest master and return with `-Wused-but-marked-unused warnings`?

dilyanpalauzov:
I still want to have this merged, as this improves code quality, or rather — amount of code reduces without changing functionality.  Actually, compiling the code with `-Weveryting` / `-Wall -Wextra` leads then to less warnings and allows to concentrate on the remaining, potentially serious compiler warnings.

When I submitted this three years ago, it had applied correctly towards the latest master branch and likely addressed the warning on all places.   I am repeatedly asked for a plethora for pull requests from me in this repository, made long time in the past, if I can rework them, so that they apply clearly towards the master branch today.  It is not my fault, that today this changeset does not apply straightforward towards the master branch, there is huge time lapse between the moment I submitted the change and the moment it is reviewed. As you can imagine, I have also changesets, which are not yet submitted, because there is realistic probability, that these will get attention only after they do not apply anymore towards the master branch.  At the same time in this concrete case it is clear what needs to be done, in order to create a changeset, which applies towards the master branch.

That said, I prefer if somebody, not me, clears all places producing used-but-marked-unused warnings.

rsto:

elliefm:
I'll be very happy when GCC starts detecting and warning about this case with `-Wall -Wextra`, and when that happens, whoever is the first to upgrade will probably clean it all up right away (because we all build with `-Werror` too, so warnings break our builds), and then it won't be an issue ever again.

In the meantime, we neglected this one and it got stale, sorry about wasting your effort like that.

PR #3746, opened 2021-10-31
jmap_mail.c: silence two -Wconditional-uninitialized warnings

dilyanpalauzov:


imap/jmap_mail.c
```diff
@@ -1275,9 +1275,9 @@ static int _email_is_expunged_cb(const conv_guidrec_t *rec, void *rock)
             uint32_t internal_flags;
             modseq_t createdmodseq;
             r = msgrecord_get_systemflags(mr, &flags);
-            if (!r) msgrecord_get_internalflags(mr, &internal_flags);
-            if (!r) msgrecord_get_createdmodseq(mr, &createdmodseq);
             if (!r) {
+                msgrecord_get_internalflags(mr, &internal_flags);
+                msgrecord_get_createdmodseq(mr, &createdmodseq);
```
elliefm, line 1280: Huh.  I wonder if it was deliberate that we were ignoring errors from `msgrecord_get_internalflags()` and `msgrecord_get_createdmodseq()`.  If it was deliberate, then this fix is good, and much clearer.

But the repeated `if (!r) ...;` constructs smell like they were meant to be our common `if (!r) r = ...;` pattern, but mistyped...

The first of these came from 77246af0f3a (when we split the flags field into separate system and internal flags), and the second from b2531deceff (when we started comparing against createdmodseq).  Though the second might have just been a copy and paste bug from the line above.  @ajaysusarla @brong -- these were your commits, what do you think?

ajaysusarla: that sounds perfectly reasonable to me

rsto:
We should not ignore the return values. Also, let's rather fix all -Wconditional-uninitialized warning errors than just these two. Closing this in favor of https://github.com/cyrusimap/cyrus-imapd/pull/5053

PR #3749, opened 2021-11-01
Quota log

brong:
This is not necessarily for merging, but maybe is after a bit of tidying.  It adds a bunch of extra syslogs to quota, but mostly in errors cases, which are good to keep.

We're trying to track down a rare issue in the Fastmail infrastructure where the quota file itself (quotalegacy backend) goes missing, but the mailboxes themselves still have a quotaroot on them, leading to LOSTQUOTA errors, and also it crashes out on rename which is pretty sucky.

### DIFF
--- a/imap/quota_db.c
+++ b/imap/quota_db.c
@@ -234,6 +234,7 @@ out:
 EXPORTED int quota_read_withconversations(struct quota *quota)
 {
     int r = quota_read(quota, NULL, 0);
+    if (r) return r;
 
     if (config_getswitch(IMAPOPT_QUOTA_USE_CONVERSATIONS)) {
         struct conversations_state *local_cstate = NULL;
@@ -251,7 +252,7 @@ EXPORTED int quota_read_withconversations(struct quota *quota)
         if (local_cstate) conversations_commit(&local_cstate);
     }
 
-    return r;
+    return 0;
 }
 
 /*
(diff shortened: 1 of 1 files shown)


imap/quota_db.c
```diff
@@ -478,6 +480,8 @@ EXPORTED int quota_write(struct quota *quota, int silent, struct txn **tid)
 
     dlist_printbuf(dl, 0, &buf);
 
+    syslog(LOG_NOTICE, "QUOTADEBUG: quota write root=<%s> data=<%.*s>", quota->root, (int)buf.len, buf.s);
```
rjbs, line 483: This should be LOG_DEBUG.

imap/quota_db.c
```diff
@@ -618,6 +624,8 @@ EXPORTED int quota_deleteroot(const char *quotaroot, int silent)
 
     r = cyrusdb_delete(qdb, quotaroot, strlen(quotaroot), NULL, 0);
 
+    syslog(LOG_NOTICE, "QUOTADEBUG: removed quota root=<%s>", quotaroot);
```
rjbs, line 627: This should be LOG_DEBUG.

brong:
@rjbs I've now removed all the DEBUG logs, we're only logging actual errors now

rjbs (approved):
LGTM 👍🏽

PR #3752, opened 2021-11-02
JMAP Email/set{update}: invalidate updated email ids that belong to erroneous plan

rsto:
Fixes a bug in Email/set{update} for shared mailboxes. Root cause identified together with @wolfsage 

The commit message has it all:
    
    The Email/set{update} planer creates a plan for each mailbox where
    at least one email is updated, copied or moved. If a user does not
    have the required ACLs or quota for a mailbox, then all updated
    emails in that mailbox are invalidated in a set error. But updated
    emails in other mailboxes that depended on that mailbox are currently
    not reported as invalid.
    
    This patch invalidates also updated emails that depend the mailbox
    of an erroneous update plan, but do not belong to it.

Also fixes an issue with Email/set{update} not validating boolean values in overwritten `mailboxIds`.

Tested in https://github.com/cyrusimap/cassandane/pull/182

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -10763,6 +10763,17 @@ static void _email_update_parse(json_t *jemail,
     else if (!json_object_size(mailboxids)) {
         jmap_parser_invalid(parser, "mailboxIds");
     }
+    else {
+        const char *mailboxid = NULL;
+        json_t *jval;
+        json_object_foreach(mailboxids, mailboxid, jval) {
+            if (jval != json_true()) {
+                jmap_parser_push(parser, "mailboxIds");
+                jmap_parser_invalid(parser, mailboxid);
+                jmap_parser_pop(parser);
+            }
+        }
+    }
     update->mailboxids = mailboxids;
 
     /* Is snoozed being overwritten or patched? */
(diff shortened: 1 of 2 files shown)


imap/jmap_mail.c
```diff
@@ -11644,26 +11644,64 @@ static int _email_bulkupdate_plan(struct email_bulkupdate *bulk, ptrarray_t *upd
...
+                for (j = 0; j < ptrarray_size(uidrecs); j++) {
+                    struct email_uidrec *uidrec = ptrarray_nth(uidrecs, j);
+                    intptr_t errcode = (intptr_t)hash_lookup(uidrec->mboxrec->mbox_id, &erroneous_plans);
+                    if (errcode && !json_object_get(bulk->set_errors, uidrec->email_id)) {
+                        json_object_set_new(bulk->set_errors, uidrec->email_id,
+                                jmap_server_error(errcode));
+                    }
+                }
+            }
+            ptrarray_t *uidrecs[3] = { &plan->setflags, &plan->delete, &plan->snooze };
```
wolfsage, line 11672: Why is plan->copy handled differently above and not part of this loop? It looks the same to me

wolfsage (commented):
Seems good to me (minus 1 question and 2 nits).

But I'd prefer someone with more familiarity with cyrus to approve!

imap/jmap_mail.c
rsto: Not quite. plan->copy is a list of a list of uidrecs. The others are list of records.

rsto:
Thanks @wolfsage . Updated the commit message.

ksmurchison (approved):
LGTM

PR #3753, opened 2021-11-02
jmap_util: do not guess charset of valid UTF-8 with replacement chars

rsto:
Fixes a bug in the JMAP internal utf8 decoder, where replacement characters in a  base64-encoded UTF-8 MIME body could cause the decoder to incorrectly guess the source charset.

Tested in https://github.com/cyrusimap/cassandane/pull/183


brong (approved):
Yeah, this looks good.

rjbs:
(5572ac3 should be removed ?)

PR #3754, opened 2021-11-02
tautology in http_dav_share: xml_add_shareeaccess() and propfind_shareaccess()

dilyanpalauzov:
xml_add_shareaccess() and propfind_shareaccess() call
> if (mboxname_userownsmailbox(fctx->req_tgt->userid, fctx->mbentry->name)) …

where req_tgt->userid and mbentry are both set based on the requested path.  In turn mboxname_userownsmailbox() is always true.

Since the ACL and PROPFIND on DAV:acl http verbs are supposed to return answer only if the user has admin rights, and since add_shareaccess() and propfind_shareaccess() return the same information, as PROPFIND on DAV:acl, this patch changes propfind_shareaccess() to return NO_PRIVILEGES, when the requesting user has no admin rights.

xml_add_shareaccess() and propfind_shareaccess() return now sharing information, if the authenticated user has admin rights on the mailbox.

### DIFF
--- a/imap/http_dav_sharing.c
+++ b/imap/http_dav_sharing.c
@@ -1413,7 +1413,9 @@ static void xml_add_sharee(const char *userid, void *data, void *rock)
 HIDDEN void xml_add_shareaccess(struct propfind_ctx *fctx,
                                 xmlNodePtr resp, xmlNodePtr node, int legacy)
 {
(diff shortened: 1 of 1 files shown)


ksmurchison:
@brong Do you have any thoughts on this one?  Seems sane to me unless I'm missing something.

brong (approved):
This looks fine to me; going to merge it

PR #3755, opened 2021-11-03
configure.ac: the SIZEOF_(INT,SIZE_T,OFF_T) variables are nowhere used

dilyanpalauzov:


elliefm:
The defines aren't necessarily used, but these statements also produce `configure` output like this:

```
...
checking size of int... 4
checking size of long... 8
checking size of size_t... 8
checking size of off_t... 8
checking size of time_t... 8
checking size of long long int... 8
checking size of unsigned long long int... 8
checking whether byte ordering is bigendian... no
checking alignment of uint32_t... 4
...
```

which we use all the time when tracking down weird problems on other platforms.

dilyanpalauzov:
The disappearing lines, by this change, are:
```
checking size of int... 4
checking size of size_t... 8
checking size of off_t... 8
```
If these lines are essential for debugging, then the bugs are, that software depends implicitly on the int/size_t/off_t sizes.  The solution is not the output of `./configure`, but removing the implicit dependencies.

elliefm:
Yep, but to find and remove them, we have to know about them, and this is how we come to know about them.  When some weird thing happens, and someone reports it, and we look at their configure output and see "size of size_t... 4" (or whatever), we can know pretty quickly that we've probably reintroduced this class of bug, and can fix it.

Without this line in the configure output, we could be chasing our tails for ages not realising that their weird platform has a four byte size_t, because the chances that _they_ know and  just tell us up front are slim.

PR #3763, opened 2021-11-08
http_*dav.c: don't need to double check ACLs in per-resource-type GET/HEAD callback

ksmurchison:


dilyanpalauzov:
How about meth_get_head_fb() and http_rss.c:meth_get()?

rsto (approved):
Looks good to me.

ksmurchison:
> How about meth_get_head_fb() and http_rss.c:meth_get()?

Neither of those use meth_get_head() in http_dav.c

PR #3765, opened 2021-11-09
Remove some references to Berkeley DB

dilyanpalauzov:
Remove some referencs to Berkeley DB.

Can now the restriction about the absolute file names in cvt_cyrusdb and cyr_dbtool now be removed?


elliefm (approved):

elliefm:
Thanks :)

With regard to the absolute file names: I'm not sure.  I believe we plan to spend some time on the command line tools next year -- reviewing what's there, what's missing, what's redundant, etc, and in particular adding support for long option names, since some of them are running out of single-character switches for new features.  I guess we'll probably revisit the absolute file names restriction at the same time.

PR #3769, opened 2021-11-10
lib/lock_flock.c: export debug_locks_longer_than

dilyanpalauzov:
Fixes a problem when flock(2) is used instead of fcnlt(2).

The symbol `debug_locks_longer_than` is required by imap/global.c.

lib/lock_flock.c does not log long locks.

### DIFF
--- a/lib/lock_flock.c
+++ b/lib/lock_flock.c
@@ -53,6 +53,8 @@
 
 EXPORTED const char lock_method_desc[] = "flock";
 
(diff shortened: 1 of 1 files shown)


elliefm (approved):

elliefm:
Looks good, thanks.  I've cherry-picked this back to 3.4 and 3.2, which seems to be as far back as it exists.

PR #3770, opened 2021-11-11
CardDAV PUT code reorganization

ksmurchison:
carddav_store_resource() should just store the resource.  carddav_put() should do data sanity checks itself as they (and any performance hit) aren't needed by COPY/MOVE.

This commit will be used by the vcard4 PR to sanity check the PUT Content-Type against the vCard VERSION itself.


rsto (approved):
Looks good to me

PR #3773, opened 2021-11-15
cyrusdb: remove bitrotted old test suite

elliefm:
This used to live in lib/test/ but it's been broken since bf641dc1114911c7496a22f4353f8cf7e4db8672 (2012), which incompatibly changed the cyrusdb API.

Nowadays cyrusdb is fairly comprehensively tested by cunit/aaa-db.testc, we don't need to bring this back.


ksmurchison (approved):

brong (approved):
Yeah, looks great.

Issue #3771, opened 2021-11-12
XFER 3.5->3.4 loses specialuse annotations


elliefm:
I'm not sure why, but it appears to do it deliberately!

First it correctly replicates the mailboxes, including the special use annotation (in this case `\Drafts` on `user.cassandane.Drafts`).

But then, near the end of the process, it explicitly checks for the annotation and UNANNOTATION's it.  Here's a snippet from telemetry:

```
<1636677874<S9 SYNCGET ANNOTATION user.cassandane.Drafts
>1636677874>* ANNOTATION %(MBOXNAME user.cassandane.Drafts ENTRY /specialuse USERID cassandane VALUE {7}
\Drafts)
S9 OK success
<1636677874<S10 SYNCAPPLY UNANNOTATION %(MBOXNAME user.cassandane.Drafts ENTRY /specialuse USERID cassandane)
>1636677874>S10 OK success
<1636677874<S11 SYNCGET ANNOTATION user.cassandane
>1636677874>S11 OK success
```

???

It's not a problem for 3.5->3.5, nor is it a problem for 3.4->3.4.

elliefm:
Vague thought: I wonder if this is related to the thing where annotations have modseqs now

elliefm:
3.5->3.5 final sync:

```
<1637038947<S8 SYNCGET USER cassandane
>1637038947>* MAILBOX %(UNIQUEID 619daa68-0607-4c6a-a5c1-2a4c2a3e503f MBOXNAME user.cassandane SYNC_CRC 597298260 SYNC_CRC_ANNOT 2750922030 LAST_UID 3 HIGHESTMODSEQ 8 RECENTUID 3 RECENTTIME 1637038947 LAST_APPENDDATE 1637038947 POP3_LAST_LOGIN 0 POP3_SHOW_AFTER 0 UIDVALIDITY 1637038947 PARTITION default ACL "cassandanelrswipkxtecdan	admin	lrswipkxtecdan	anyone	p	" OPTIONS P CREATEDMODSEQ 1 FOLDERMODSEQ 4)
[...]

PR #3781, opened 2021-11-17
Annotatemore mboxname

brong:
Don't use mboxlist_findall for the mailbox name in replication for annotations.

Should hopefully fix https://github.com/cyrusimap/cyrus-imapd/issues/3771

### DIFF
--- a/imap/annotate.c
+++ b/imap/annotate.c
@@ -1207,7 +1207,7 @@ static int annotatemore_findall_full(const char *pattern, /* internal */
      * server annotations, which are recognised by an empty pattern by the
      * code below */
     if (!mailbox && !pattern && !*mboxname) {
-        mailbox = NULL;
+        mboxname = NULL;
         pattern = "";
     }
 
--- a/imap/annotate.c
+++ b/imap/annotate.c
@@ -1175,6 +1175,7 @@ static int _findall(struct findall_data *data, void *rock)
 
 static int annotatemore_findall_full(const char *pattern, /* internal */
                          const struct mailbox *mailbox,
+                         const char *mboxname,
                          unsigned int uid,
                          const char *entry,
                          modseq_t since_modseq,
@@ -1184,10 +1185,11 @@ static int annotatemore_findall_full(const char *pattern, /* internal */
 {
     int r;
     struct find_rock frock;
+    mbentry_t *mbentry = NULL;
 
     init_internal();
 
-    assert(pattern || mailbox);
+    assert(pattern || mailbox || mboxname);
     assert(entry);
 
     frock.pattern = pattern;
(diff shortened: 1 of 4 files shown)


elliefm (commented):
This looks good, and fixes the problem I was seeing, but the tests that touch global annotations now fail.  Looks like mboxname in sync_get_annotation can be an empty string -- maybe this is why it was using annotatemore_findall_pattern!

Metadata.comment_repl chokes here:

```
>1637124856>S2 SYNCGET ANNOTATION ""
<1637124856<S2 NO IMAP_MAILBOX_NONEXISTENT No Such Mailbox
0454140101/sync_client[5897]: ANNOTATION received NO response: IMAP_MAILBOX_NONEXISTENT No Such Mailbox
0454140101/sync_client[5897]: Error in do_sync(): bailing out! Mailbox does not exist
```

And Metadata.private_global_annot_replication and Metadata.shared_global_annot_replication fall over in the same way

imap/annotate.c
```diff
@@ -1200,22 +1202,22 @@ static int annotatemore_findall_full(const char *pattern, /* internal */
     frock.since_modseq = since_modseq;
     frock.flags = flags;
 
-    r = _annotate_getdb(mailbox ? mailbox_uniqueid(mailbox) : NULL, mailbox, uid, 0, &frock.d);
+    if (mailbox || mboxname) {
```
elliefm, line 1205: Something like this here

```suggestion
    if (mailbox || (mboxname && *mboxname)) {
```

sorts out the first part of the "mboxname might be `""`" problem (but is incomplete, more below)

imap/annotate.c
```diff
@@ -1228,6 +1230,7 @@ static int annotatemore_findall_full(const char *pattern, /* internal */
     }
```
elliefm, line 1231: And then once we stop exiting early above, this if/else chain then needs to change shape too.  I think the three cases are something like:

* mbentry? => it's a single mailbox annotation
* pattern && *pattern? => it's a pattern
* otherwise => it's a server entry

In particular `else if (!*pattern) {` is bad now: we no longer get an empty string in the "pattern" argument during replication, because it's in the mboxname argument now instead, and pattern is NULL so *pattern segfaults.

I notice a few other invocations of `annotatemore_findall_pattern` (jmap, nntpd).  At a glance, two of them call it with an argument called "pattern", but the third calls it with an mbname_intname(mbname), and might also be a candidate for using the `annotatemore_findall_mboxname` version instead.

imap/annotate.c
brong: Yep, fixed this up so the other variables are right too

imap/annotate.c
brong: Yep, patched the other annotatemore_findall_pattern too

ksmurchison (approved):
LGTM

imap/annotate.c
```diff
@@ -1200,22 +1202,30 @@ static int annotatemore_findall_full(const char *pattern, /* internal */
     frock.since_modseq = since_modseq;
     frock.flags = flags;
 
-    r = _annotate_getdb(mailbox ? mailbox_uniqueid(mailbox) : NULL, mailbox, uid, 0, &frock.d);
+    /* special case where mboxname is "" and others are empty, this means
+     * server annotations, which are recognised by an empty pattern by the
+     * code below */
+    if (!mailbox && !pattern && !*mboxname) {
+        mailbox = NULL;
```
elliefm, line 1209: Same tests still failing... ah, looks like a typo.  This fixes it:

```suggestion
        mboxname = NULL;
```

elliefm (changes requested):

brong:
@elliefm - fixed and pushed, it's passing all the tests for me now.

elliefm (approved):
Looks good!

elliefm:
Can this be merged?  Looks like it's just been forgotten about, rather than held up for any particular reason

I'll need to backport it to 3.6 once it's merged

PR #3788, opened 2021-11-22
dlist: use xsyslog for IOERRORs

### LINKED ISSUE #3784: reservefile: log errno when file ops fail
I ran into a sync problem, which turned out to be a filesystem issue, such that the `fopen` call in `reservefile` (`dlist.c`) failed. `errno` was `ENOSPC`, and it would have helped me a lot if that had been logged!

elliefm:
More xsyslog conversions -- as usual for these, please build this, in case your compiler catches something mine didn't.  Thanks

Fixes #3784


brong (approved):
Looks good, and compiles on my setup.

PR #3791, opened 2021-11-24
some cassandane integration cleanups

elliefm:
Some miscellaneous cleanups.

Was hoping to integrate the build systems as part of this, but it turns out that's going to be complicated and probably take a while to get right, and I don't want to hold up the trivial stuff waiting for it.


ksmurchison (approved):

elliefm:
Need to pick this back to the older branches too...

PR #3798, opened 2021-11-26
lib/imapoptions: add deprecated entry for reverseuniqueuids

### LINKED ISSUE #3616: reverseuniqueids setting removed in 3.5
it might need to be brought back as "deprecated", since i think it's present in 3.4

elliefm:
Accidentally removed entirely by bfb0061cf64ff878b27e0d3e8e2d5549c77b563f
(#3484), when it should have been updated with a deprecated-since version
instead.

Fixes #3616

### DIFF
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -2156,6 +2156,9 @@ If all partitions are over that limit, this feature is not used anymore.
 /* At startup time, ctl_cyrusdb -r will check this value and it
    will either add or remove reverse ACL pointers from mailboxes.db */
 
+{ "reverseuniqueids", 1, SWITCH, "UNRELEASED", "UNRELEASED" }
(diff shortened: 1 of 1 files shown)


brong (approved):
Yep, this is fine - does cyr_info lint mode detect this as a deprecated key?

elliefm:
I can't remember off the top of my head whether the lint modes say anything in particular about deprecated options, but being able to bump the timestamp fields is important for the `-s version` option (which highlights configured settings whose behaviour has changed since that version, and whose documentation you therefore need to review), and the documentation will be generated to say that it's deprecated, instead of it just disappearing entirely

elliefm:
Ah I see it's not super obvious from the diff alone... yes, the second "UNRELEASED" version stamp is the deprecated_since field.

Issue #3782, opened 2021-11-17
test failures when mailbox_legacy_dirs = yes


elliefm:
To reproduce:

* add "mailbox_legacy_dirs = yes" to the `[config]` section of cassandane.ini
* run testrunner.pl normally

Here's the failures and errors I saw when I ran this this afternoon (sorry, I've lost the "error" vs "failed" status).  This won't be quite exactly the same as your cassandane cause I happened to run it on a local branch where I have some extra WIP tests for an unrelated thing I'm chasing, but it'll probably be close.

```
Cyrus::Delete.admin_inbox_imm
Cyrus::FastMail.rename_deepfolder_intermediates_rightnow
Cyrus::FastMail.rename_deepfolder_intermediates
Cyrus::FastMail.rename_deepuser_standardfolders
Cyrus::FastMail.rename_deepuser_standardfolders_rightnow
Cyrus::JMAPBackup.restore_mail_submailbox
Cyrus::JMAPBackup.restore_mail_simple
Cyrus::JMAPBackup.restore_mail_full
Cyrus::JMAPCalendars.calendar_changes
Cyrus::JMAPMailbox.mailbox_changes
Cyrus::JMAPMailbox.mailbox_changes_counts
Cyrus::JMAPMailbox.mailbox_changes_shared
Cyrus::JMAPMailbox.mailbox_querychanges_name
Cyrus::JMAPMailbox.mailbox_set
Cyrus::JMAPMailbox.mailbox_set_destroy_empty
Cyrus::JMAPMailbox.mailbox_set_destroy_movetomailbox
Cyrus::JMAPMailbox.mailbox_set_destroy_movetomailbox_empty
Cyrus::JMAPMailbox.mailbox_set_destroy_removemsgs
Cyrus::JMAPMailbox.mailbox_set_name_collision
Cyrus::JMAPMailbox.mailbox_set_name_interop
Cyrus::JMAPMailbox.mailbox_set_name_swap
Cyrus::JMAPMailbox.mailbox_set_order
Cyrus::JMAPMailbox.mailbox_set_parent
Cyrus::JMAPMailbox.mailbox_set_protected_move_parent
Cyrus::JMAPMailbox.mailbox_set_role_move_destroy
Cyrus::JMAPMailbox.mailbox_set_role_protected_destroy
Cyrus::JMAPMailbox.mailbox_set_subscriptions_destroy
Cyrus::JMAPMailbox.mailbox_set_shared
[...]

PR #3799, opened 2021-11-26
partial fixes for mailbox_legacy_dirs=yes problems

elliefm:
Two small fixes, found by running cassandane with `mailbox_legacy_dirs: yes` in the `[config]` section.

Part of #3782.  There's more still to be fixed, but this is what I've tracked down so far.


brong (approved):
They both look really good.  The Delete one I had independently found (and indeed, committed a change for from my local repo and forgot that I didn't remove it before deploying - so it's in our production builds!)

Issue #3756, opened 2021-11-03
Upgrade 3.2 → 3.4 does not drop INDEX idx_res_uid ON dav_objs 


dilyanpalauzov:
`git diff origin/cyrus-imapd-3.2..origin/cyrus-imapd-3.4 -- dav_db.c` shows than an sqlite `CREATE INDEX IF NOT EXISTS idx_res_uid ON dav_objs ( res_uid )` is not called any more, but if it existed in the past, the index is not deleted.  I guess
```c
+#define CMD_DBUPGRADEv10                                        \
+    "CREATE UNIQUE INDEX IF NOT EXISTS idx_ical_imapuid ON ical_objs ( mailbox, imap_uid );" \
+    "CREATE UNIQUE INDEX IF NOT EXISTS idx_vcard_imapuid ON vcard_objs ( mailbox, imap_uid );" \
+    "CREATE UNIQUE INDEX IF NOT EXISTS idx_object_imapuid ON dav_objs ( mailbox, imap_uid );"
```
should be extended to delete the index.

PR #3800, opened 2021-11-27
dav_db:CMD_DBUPGRADEv10 — DROP INDEX IF EXISTS idx_res_uid

dilyanpalauzov:
Since d35f0dac07904bef the index is not created for new setups.  The index was however not removed from older setups.

This change does not help for upgrades to 3.4, which upgrades were performed before this change.  For such setups, thas is if you already run 3.4, the command

> for x in $(find /var/imap -name '*.dav'); do sqlite3 $x "DROP INDEX IF EXISTS idx_res_uid"; done

executed under the 'cyrus' user, does perform the change, where '/var/imap' is the configdirectory in imapd.conf .

Fixes https://github.com/cyrusimap/cyrus-imapd/issues/3756.


brong (approved):
Yep, this looks good.  The code has changed a little underneath and this will break some other branches, so I'll massage this into mergeable shape, merge it, then fix the other branches.

PR #3795, opened 2021-11-24
caldav_put() DUE must be after DTSTART in VTODO, both must have the same value-type

dilyanpalauzov:
This is postulated by [RFC 5545 Section 3.8.2.3 Date-Time Due](https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.2.3).

[Evolution can use](https://gitlab.gnome.org/GNOME/evolution/-/issues/1715) different value-types.

PR #3802, opened 2021-11-29
caldav_put() DUE must be after DTSTART in VTODO, both must have the same value-type

ksmurchison:
This is postulated by [RFC 5545 Section 3.8.2.3 Date-Time Due](https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.2.3).

[Evolution can use](https://gitlab.gnome.org/GNOME/evolution/-/issues/1715) different value-types.

(This is a replacement for #3795 )

### DIFF
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -4307,6 +4307,49 @@ static int personalize_resource(struct transaction_t *txn,
 }
 
 
+static const struct endtime_errors_t {
+    const char *type_mismatch;
+    const char *bad_value;
+} endtime_errors[] = {
+    { "DTEND must have same value type as DTSTART",
+      "DTEND must occur after DTSTART" },
+    { "DUE must have same value type as DTSTART",
+      "DUE must occur after DTSTART"}
+};
+
+/* Make sure DTEND|DUE > DTSTART, and both values have value same type */
+static int check_endtime(icalcomponent *comp, const char **errstr)
+{
+    const struct endtime_errors_t *errors = NULL;
+    icaltimetype dtstart, dtend;
+
+    dtstart = icalcomponent_get_dtstart(comp);
+    if (icaltime_is_null_time(dtstart)) return 0;
+
+    if (icalcomponent_isa(comp) == ICAL_VTODO_COMPONENT) {
+        dtend = icalcomponent_get_due(comp);
+        errors = &endtime_errors[1];
+    }
+    else {
+        dtend = icalcomponent_get_dtend(comp);
+        errors = &endtime_errors[0];
+    }
+
+    if (icaltime_is_null_time(dtend)) return 0;
+
+    if (icaltime_is_date(dtend) != icaltime_is_date(dtstart)) {
+        *errstr = errors->type_mismatch;
+        return HTTP_FORBIDDEN;
+    }
+
+    if (icaltime_compare(dtend, dtstart) < 1) {
+        *errstr = errors->bad_value;
+        return HTTP_FORBIDDEN;
+    }
+
+    return 0;
+}
+
 /* Perform a PUT request
  *
  * preconditions:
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
This does not produce any warnings:
```c
#include <stdio.h>

static const struct endtime_errors_t {
    const char type_mismatch[43];
    const char bad_value[31];
} endtime_errors[] = {
    { "DTEND must have same value type as DTSTART",
      "DTEND must occur after DTSTART" },
    { "DUE must have same value type as DTSTART",
      "DUE must occur after DTSTART"}
};

int main() {
  const struct endtime_errors_t *a;
  for (int i = 0; i < 2; i++) {
    a = &endtime_errors[i];
    printf("%s\n", a->type_mismatch);
    printf("%s\n", a->bad_value);
  }
  return 0;
}
```

brong:
This is going to hit the same problem that https://github.com/cyrusimap/cyrus-imapd/pull/3351 is there to deal with - zero length events.

imap/http_caldav.c
```diff
@@ -4307,6 +4307,49 @@ static int personalize_resource(struct transaction_t *txn,
...
+    const char type_mismatch[43];
+    const char bad_value[31];
+} endtime_errors[] = {
+    { "DTEND must have same value type as DTSTART",
+      "DTEND must occur after DTSTART" },
+    { "DUE must have same value type as DTSTART",
+      "DUE must occur after DTSTART" }
+};
+
+/* Make sure DTEND|DUE > DTSTART, and both values have value same type */
```
dilyanpalauzov, line 4320: ... both values have same type.  One 'value' is too.much.

imap/http_caldav.c
```diff
@@ -4367,6 +4409,13 @@ static int caldav_put(struct transaction_t *txn, void *obj,
         rrule = icalcomponent_get_first_property(comp, ICAL_RRULE_PROPERTY);
     }
 
+    /* Make sure DTEND|DUE > DTSTART, and both values have value same type */
```
dilyanpalauzov, line 4412: Here is one extra 'value’, too.

imap/http_caldav.c
```diff
@@ -4428,22 +4458,11 @@ static int caldav_put(struct transaction_t *txn, void *obj,
...
-                goto done;
-            }
-            if (icaltime_compare(dtend, dtstart) < 1) {
-                txn->error.desc = "DTEND must occur after DTSTART";
-                txn->error.precond = CALDAV_VALID_OBJECT;
-                ret = HTTP_FORBIDDEN;
-                goto done;
-            }
+        /* Make sure DTEND|DUE > DTSTART, and both values have value same type */
+        ret = check_endtime(comp, &txn->error.desc);
```
dilyanpalauzov, line 4462: One ’value’ too much.

brong (changes requested):
We need to solve the "same start and end" problem so we don't break FM production again, so putting that as a blocker on this.

dilyanpalauzov:
From the following statements:
* In Tasks, if DUE and DTSTART are present, both must have the same value-type
* In Tasks, if DUE and DTSTART are present, DUE must be after DTSTART, as required by RFC 5545
* In Events, DTSTART must be before (and different from) DTEND, despite RFC 5545 allowing DTSTART==DTEND

shall be enforced by Cyrus IMAP?

imap/http_caldav.c
dilyanpalauzov: > ret = check_endtime(comp, &txn->error.desc);

should be
> ret = check_endtime(nextcomp, &txn->error.desc);

imap/http_caldav.c
ksmurchison: Fixed

imap/http_caldav.c
ksmurchison: Fixed

imap/http_caldav.c
ksmurchison: Fixed

ksmurchison:
This PR has been superseded by later commits

PR #3804, opened 2021-11-30
http_dav_sharing: decode sharer URI before parsing path

rsto:
Fixes parsing of sharer URIs with escaped characters, which resulted in mailbox lookup failures later in the codepath.


ksmurchison (approved):
LGTM, especially if tests pass

PR #3805, opened 2021-11-30
GET DAV Collection Fix

ksmurchison:
http_dav.c: always return 204 (No Content) for pseudo-collections (those with no backing mailbox)

### DIFF
--- a/imap/http_dav.c
+++ b/imap/http_dav.c
@@ -5247,11 +5247,11 @@ int meth_get_head(struct transaction_t *txn, void *params)
     r = dav_parse_req_target(txn, gparams);
     if (r) return r;
 
-    if (txn->req_tgt.namespace->id == URL_NS_PRINCIPAL) {
-        /* Special "principal" */
-        if (txn->req_tgt.flags == TGT_SERVER_INFO) return get_server_info(txn);
+    /* Special "principal" */
+    if (txn->req_tgt.flags == TGT_SERVER_INFO) return get_server_info(txn);
 
-        /* No content for principals (yet) */
+    if (!txn->req_tgt.mbentry) {
+        /* No content for non-mailbox URLs (yet) */
         return HTTP_NO_CONTENT;
     }
 


imap/http_dav.c
```diff
@@ -5247,11 +5247,11 @@ int meth_get_head(struct transaction_t *txn, void *params)
     r = dav_parse_req_target(txn, gparams);
     if (r) return r;
 
-    if (txn->req_tgt.namespace->id == URL_NS_PRINCIPAL) {
```
wolfsage, line 5250: What does getting rid of the principal check do? Are we sure that can go (maybe it's just encompassed by not having a mbentry?)

Can TGT_SERVER_INFO somehow be set outside of a URL_NS_PRINCIPAL namespace->id, and now we're going to start intercepting requests we otherwise wouldn't?

wolfsage (commented):
A few questions - but otherwise I don't understand this code well enough to approve this MR - is there someone else who can look?

Thanks

imap/http_dav.c
ksmurchison: TGT_SERVER_INFO can not be set for any other namepsace, which is why I removed the namespace check.  It was there previously to return 204 for any other principal, but now that is captured by any URL that doesn't map to a mbentry.

brong:
This PR could do with a bit more description about what was causing the issue that it's fixing.  The code looks good, but the reasoning is unclear from just the PR description itself.

brong (approved):
Yep, looks good.  I assume this is a crasher fix.

ksmurchison:
This actually restores previous behavior that regressed in a previous commit. alh noticed today that a robm test was failing because we were responding with 403 rather than 204

brong:
Ahh right.  Yeah, all good - let's go ahead and merge this then :)

elliefm:
> This PR could do with a bit more description about what was causing the issue that it's fixing. The code looks good, but the reasoning is unclear from just the PR description itself.

One thing that can work well in these case is creating an issue that describes the problem, and then the PR description can be in terms of "Fixes #xyz by scrobbling the widget earlier", or whatever.  It's also nice for release notes, because the release notes can then say "Fixed #xyz: scrobble widget earlier", and the curious can click through to the issue describing the problem that was solved.

I'm reluctant to insist on this though, cause making an issue for every small fix is probably just gonna feel like extra busywork and a waste of time.  I'm sure there's a point where the bug is complex enough that the separate issue is almost essential, but I don't know where the threshold between extremes lies, or what side of it this one lies on.

PR #3806, opened 2021-11-30
Sieve user rename fix

ksmurchison:
The change to sieve_db.c turned out to be unnecessary, since I didn't use it outside of that file.  I intended to use it in sync_support.c, but I thought it made sense to create a single mbname_t rather than using separate mboxname_*() functions that all create and destroy a private mbname_t.

Also, I wasn't sure how to actually check for the presence of INBOX.#sieve on the replica in the Cass test.  I verified that it indeed does exist by hand, but the test should be augmented.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Rename.pm
+++ b/cassandane/Cassandane/Cyrus/Rename.pm
@@ -610,6 +610,53 @@ sub test_rename_deepuser
     $self->check_replication('newuser');
 }
 
+sub test_rename_user_sieve
+    :AllowMoves :Replication :SyncLog :needs_component_sieve
+{
+    my ($self) = @_;
+
+    xlog $self, "Test user rename with Sieve script";
+
+    my $user = 'cassandane';
+    my $newuser = 'newuser';
+    my $scriptname = 'test1';
+    my $scriptcontent = <<'EOF';
+keep;
+EOF
+
+    # install sieve script on master
+    $self->{instance}->install_sieve_script($scriptcontent, name=>$scriptname);
+
+    # verify that sieve script exists on master
+    $self->assert_sieve_exists($self->{instance}, $user, $scriptname, 0);
+
+    # replicate and check initial state
+    $self->run_replication();
+    $self->check_replication($user);
+
+    # verify that sieve script exists on both master and replica
+    $self->assert_sieve_exists($self->{instance}, $user, $scriptname, 1);
+    $self->assert_sieve_exists($self->{replica}, $user, $scriptname, 1);
+
+    # rename user
+    my $admintalk = $self->{adminstore}->get_client();
+    my $res = $admintalk->rename('user.cassandane', 'user.newuser');
+    $self->assert(not $admintalk->get_last_error());
+
+    # verify that sieve script exists on master
+    $self->assert_sieve_exists($self->{instance}, $newuser, $scriptname, 1);
+    $self->assert_sieve_not_exists($self->{instance}, $user, $scriptname, 1);
+
+    # replicate and check the renames
+    my $synclogfname = "$self->{instance}->{basedir}/conf/sync/log";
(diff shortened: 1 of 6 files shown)


imap/sync_support.c
```diff
@@ -3652,6 +3652,46 @@ int sync_apply_rename(struct dlist *kin, struct sync_state *sstate)
...
+
+                if (!r) {
+                    /* Take care of changing quotaroot and ACL */
+                    user_copyquotaroot(oldmboxname, newmboxname);
+                    user_renameacl(NULL /*namespace*/, newmboxname,
+                                   mbname_userid(oldmbname),
+                                   mbname_userid(newmbname));
+                }
+            }
+            mbname_free(&newmbname);
```
brong, line 3690: I think you mean oldmbname here?

imap/sync_support.c
```diff
@@ -3652,6 +3652,46 @@ int sync_apply_rename(struct dlist *kin, struct sync_state *sstate)
...
+                                           sstate->authstate, NULL,
+                                           sstate->local_only, 1, 1,
+                                           1/*keep_intermediaries*/,
+                                           0/*move_subscription*/,
+                                           1/*silent*/);
+
+                if (!r) {
+                    /* Take care of changing quotaroot and ACL */
+                    user_copyquotaroot(oldmboxname, newmboxname);
+                    user_renameacl(NULL /*namespace*/, newmboxname,
```
brong, line 3685: Are these needed for the other mailboxes as well?  Right now this only applies to the sieve mailbox.

brong (changes requested):
There appears to be one incorrect free (safe, but it'll leak oldmbname) - and I have a question about logic.

imap/sync_support.c
ksmurchison: Do you mean for mailboxes like INBOX.foo?  Those should be handled by normal replication right?  Presumably the rename_deepuser test already proves this.

brong: Yep, I guess so.  This is the minimal change.  Good.

brong (approved):
Looks good with this change.

elliefm:
> Also, I wasn't sure how to actually check for the presence of INBOX.#sieve on the replica in the Cass test. I verified that it indeed does exist by hand, but the test should be augmented.

Have a look at `read_mailboxes_db()` in Instance.pm -- it parses the output of `ctl_mboxlist -d` and returns it as a hash, that you can then assert for expected values within.

Something like (untested): 

```
my $mailboxesdb = $self->{replica}->read_mailboxes_db();
$self->assert_not_null($mailboxesdb->{'user.cassandane.#sieve'});
$self->assert_equals($mbtype_sieve, $mailboxesdb->{'user.cassandane.#sieve'}->{mbtype});
```

might do the trick.

There's some examples of this sort of thing in the List "tombstones" tests, and the MurderIMAP "xfer" tests.

brong:
I've added the necessary magic to the test :)

cassandane/Cassandane/Cyrus/Rename.pm
```diff
@@ -610,6 +610,75 @@ sub test_rename_deepuser
...
+    $self->{instance}->install_sieve_script($scriptcontent, name=>$scriptname);
+
+    # verify that sieve script exists on master
+    $self->assert_sieve_exists($self->{instance}, $user, $scriptname, 0);
+
+    # check if we have the new-style sieve
+    my $mailboxesdb = $self->{instance}->read_mailboxes_db();
+    my $have_new_sieve = $mailboxesdb->{'user.cassandane.#sieve'} ? 1 : 0;
+    xlog "Checking for new sieve resulted in $have_new_sieve";
+    die Dumper($mailboxesdb);
```
elliefm, line 637: Whoops, don't think you meant to leave this in...!

PR #3807, opened 2021-12-01
cassandane List: handle altnamespace being off in config

brong:


elliefm (approved):

elliefm:
For what it's worth, I don't think we generally need the "cassandane:" at the start of cassandane commits, at least when the first word would be a suite name that starts with an uppercase (e.g. "List"), since it's already visually distinct from our C sources, which are all lowercase.

And if (like me) you're already in a habit of scanning for that case distinction when reading lists of commits, then "cassandane:" being lowercase breaks it!

PR #3808, opened 2021-12-01
CalDAV Unique Object Fix

ksmurchison:
This fixes a crasher when we find a conflicting iCal UID in DAV DB, but we are unable to look up the offending mailbox record.

In the crasher that we had, the DAV DB record had the mailbox name (not id), but the byname flag was not set, so we tried to lookup the mailbox by id using the name as the id, which failed.

### DIFF
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -4502,26 +4502,27 @@ static int caldav_put(struct transaction_t *txn, void *obj,
         }
     }
     if (ret) {
-        char *owner;
-        const char *mboxname;
+        const char *mboxname = NULL;
         mbentry_t *mbentry = NULL;
 
         if (cdata->dav.mailbox_byname)
             mboxname = cdata->dav.mailbox;
         else {
             mboxlist_lookup_by_uniqueid(cdata->dav.mailbox, &mbentry, NULL);
-            mboxname = mbentry->name;
+            if (mbentry) mboxname = mbentry->name;
         }
-        owner = mboxname_to_userid(mboxname);
 
-        buf_reset(&txn->buf);
-        buf_printf(&txn->buf, "%s/%s/%s/%s/%s",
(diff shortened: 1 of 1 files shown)


imap/http_caldav.c
```diff
@@ -4502,26 +4502,27 @@ static int caldav_put(struct transaction_t *txn, void *obj,
...
+            buf_reset(&txn->buf);
+            buf_printf(&txn->buf, "%s/%s/%s/%s/%s",
+                       namespace_calendar.prefix, USER_COLLECTION_PREFIX, owner,
+                       strrchr(mboxname, '.') + 1, cdata->dav.resource);
+            txn->error.resource = buf_cstring(&txn->buf);
+            free(owner);
+        }
         mboxlist_entry_free(&mbentry);
-        free(owner);
-        ret = HTTP_FORBIDDEN;
```
rsto, line 4524: The previous code hardcoded the return value to HTTP_FORBIDDEN. Now it looks as if to return whatever non-zero `ret` was before. Is that on purpose?

rsto (commented):
Not sure if that return code would introduce a regression. Also, is mboxname really needed? Isn't the raw resource path already in the transaction of the HTTP request URL?

imap/http_caldav.c
ksmurchison: Right.  But it appears to me that it was redundant.  If we get here, ret is already set to HTTP_FORBIDDEN.  Unless I'm missing something.

ksmurchison:
> Not sure if that return code would introduce a regression. Also, is mboxname really needed? Isn't the raw resource path already in the transaction of the HTTP request URL?

mboxname is for the mailbox that contains the resource with the conflicting iCal UID, which may NOT be the target of the request.

rsto (approved):

PR #3809, opened 2021-12-01
Propfind data crasher fix

ksmurchison:
We had a crasher in a CARDDAV:addressbook-query REPORT in which we trying to fetch the CARDDAV:address_data property.  vcard_parse_string() was returning NULL, which we then try to dereference.

The first 2 commits will now return a DAV:propstat with a DAV:status of 501 in such a case rather than crash.

We believe the crash was a result of parsing non-NULL-terminated data, which should be fixed by the last 2 commits.


rsto (approved):
LGTM

PR #3810, opened 2021-12-01
vCard default to v3

ksmurchison:
Default to v3 for records stored before we actually set the version

### DIFF
--- a/imap/carddav_db.c
+++ b/imap/carddav_db.c
@@ -230,9 +230,14 @@ static int read_cb(sqlite3_stmt *stmt, void *rock)
     cdata->dav.creationdate = sqlite3_column_int(stmt, 1);
     cdata->dav.imap_uid = sqlite3_column_int(stmt, 4);
     cdata->dav.lock_expire = sqlite3_column_int(stmt, 8);
-    cdata->version = sqlite3_column_int(stmt, 9);
     cdata->kind = sqlite3_column_int(stmt, 11);
     cdata->jmapversion = sqlite3_column_int(stmt, 18);
+
+    cdata->version = sqlite3_column_int(stmt, 9);
+    if (cdata->version == 0) {
+        /* Default to v3 for records stored before we actually set the version */
+        cdata->version = 3;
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
> This will prevent a needless vcard_to_v3() translation

Is the translation done only once, and then the version in the database updated?

ksmurchison:
We store whatever version the client feeds us via PUT.  The v3 <-> v4 translations are done on-the-fly based on the stored version and what specific version, if any, the client requests via GET/REPORT.  This PR fixes the case where we have a legacy DB record (before we actually started setting the version field) and we treat v0 != v3, thus calling vcard_to_v3().

dilyanpalauzov:
Thanks for the answer.  The question was, if for legacy records, vcard_to_v3(). is called only once, and afterwards the legacy records are not any more legacy, but current.

dilyanpalauzov:
What happens to legacy vCard-records, when the uploaded vCard was version 4?

ksmurchison:
Currently for legacy records vcard_to_v3() would be called every time the vCard is fetched.  We only set the version field on the DB record when the vCard is updated.

I'm not sure if we even accepted v4 cards previously, but if we do have legacy v4 records, they would be treated as v3 with or without this patch.  If we think that we have legacy v4 cards, then the proper fix is to parse the vCard data to get the version, if the DB record has version==0.

dilyanpalauzov:
I have both vCard 3.0 and vCard 4.0 for long time, where the database column has version=0.  It would be best, if there is some migration tool, like cvt_xlist_specialuse, which updates the database column and afterwards there is no run-time overhead and code.

ksmurchison:
dav_reconstruct will do that right now.

dilyanpalauzov:
If `dav_reconstruct` removes all cases with version==0, why is there code needed, to handle the case version==0

brong (commented):
This definitely makes sense to me.  We only stored v3 before.

brong (approved):

PR #3811, opened 2021-12-02
Sieve in mailbox replication fix

ksmurchison:
Revert to replicating #sieve mailbox as a MAILBOX.  This also includes in-protocol detection of #sieve support.


brong (approved):
Looks great.  I'm gonna spin this up today

elliefm:
This can't go to master until the in-protocol detection part is in place and working, because it'll break back-compat.

ksmurchison:
@brong Please review the v14 commit.  SQLite doesn't support dropping a column, so this seemed the easiest way to handle things so we don't have a bogus mailbox name after a user rename.

brong:
This has been running in Fastmail production for a bit now - gonna merge it to lock it into place, we can do another PR if there's more changes to make.

elliefm:
@ksmurchison asked me to run this through mixed-version Cassandane to/from 3.4.  Since it's already been merged I used master instead.  Here's the results:

3.5->3.4:

* Bunch of failures from tests that call run_replication with the "nosyncback" flag, because the 3.4 replica doesn't recognise the new "FORCE" command, and 3.5 sync_client hard exits if trying to set FORCE fails
* The MurderIMAP.xfer_user... tests still mostly choke on "\Drafts" because of #3771.  @brong has a fix for this in #3781, but it's changes-requested and maybe been forgotten about
* BUT, one of the MurderIMAP.xfer_user... tests fails because the #sieve mailbox has been replicated to 3.4 (doh!), so something is not quite right here.  I think the other ones would also fail for this reason, except they happen to complain about \Drafts first instead.

3.4->3.5:

* A couple of failures from tests that can't work in this combination anyway
* These sieve ones fail and might be relevant, haven't investigated deeper:
    JMAPSieve.sieve_set
    JMAPSieve.sieve_set_replication
    Rename.rename_user_sieve
* The XFER tests and most of the replication tests seem happy in this direction

PR #3814, opened 2021-12-06
Remove 'static const struct buf' tripping hazard

### LINKED ISSUE #3813: buf_cstring can crash on "novalue"
This doesn't currently occur in practice, but I tripped over it while debugging something unrelated, and would like to not trip over it again!

`novalue` is declared here.  Note that it's declared `static const`:

https://github.com/cyrusimap/cyrus-imapd/blob/fc57d4b15dad2850843a84a188e276b0ba283fd2/imap/sync_support.c#L1476-L1486

`buf_cstring()` is defined here:

https://github.com/cyrusimap/cyrus-imapd/blob/fc57d4b15dad2850843a84a188e276b0ba283fd2/lib/util.c#L1104-L1110

Its intention is to always return a valid cstring, and convert the `struct buf` to contain one internally at the same time.  It takes a `const struct buf*` argument so that it can be used in const-correct code, and casts away the constness internally to implement the behaviour that makes sure the buffer is 0-terminated.  This is usually fine.

But objects declared `static const` (like `novalue`) are probably stored in a read-only segment, not on the stack or heap.  So it doesn't matter that we cast away the constness -- the memory is still read only, and it segfaults.

I think what I'd like to do is move `novalue` into lib/util.c, so that everything can use it if it wants (not just `apply_annotations()`), and then special-case buf_cstring to recognise this constant and just return `""` without trying to modify it internally.

elliefm:
Fixes #3813 

Variables declared `static const` are stored in read-only memory, but the `buf_...()` functions expect to be able to change a buf's in-memory representation, even when they offer `const` as a promise the data itself won't change.  If the buf in question is in read-only memory, they'll crash when they try.

This PR removes the `static` keyword from the one `static const struct buf` that we had -- before we accidentally treat it like any other struct buf and crash!

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -1482,7 +1482,7 @@ int apply_annotations(struct mailbox *mailbox,
     const struct sync_annot *local = (local_annots ? local_annots->head : NULL);
     const struct sync_annot *remote = (remote_annots ? remote_annots->head : NULL);
     const struct sync_annot *chosen;
-    static const struct buf novalue = BUF_INITIALIZER;
+    const struct buf novalue = BUF_INITIALIZER;
     const struct buf *value;
     int r = 0;
     int diff;


brong (approved):
Yes great, thanks for doing this.

PR #3815, opened 2021-12-06
squatter.c: don't index CalDAV schedule Inbox

ksmurchison:
Self explanatory PR

### DIFF
--- a/imap/squatter.c
+++ b/imap/squatter.c
@@ -204,6 +204,7 @@ static int should_index(const char *name)
     int ret = 1;
     mbentry_t *mbentry = NULL;
     mbname_t *mbname = mbname_from_intname(name);
+    const char *userid = mbname_userid(mbname);
     /* Skip remote mailboxes */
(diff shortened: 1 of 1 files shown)


imap/squatter.c
```diff
@@ -245,6 +246,13 @@ static int should_index(const char *name)
         goto done;
     }
 
+    // skip CalDAV scheduling Inbox
+    if ((mbtype_isa(mbentry->mbtype) == MBTYPE_CALENDAR) && userid &&
+        !strcmp("Inbox", strarray_nth(mbname_boxes(mbname), 1))) {
```
rsto, line 251: Is `Inbox` hard-coded here to not include http_dav.h for `SCHED_INBOX`?

rsto (approved):
LGTM, but maybe we can skip the hard-coded mailbox name?

imap/squatter.c
ksmurchison: Yeah.  Oversight on my part

rsto (approved):

PR #3816, opened 2021-12-06
MKCOL namespacelock fix

ksmurchison:
This fixes a crasher where we call goto done before namespacelock has been declared, E.g. if parse_xml_body() fails


brong (approved):
Yep, that's great

PR #3817, opened 2021-12-07
jmap_api: ignore shared calendar/addressbook home for capabilities

rsto:
Before, if only the `#calendars` or `#addressbooks` mailboxes where shared, then the account still got reported to share calendars or addressbooks.


ksmurchison (approved):
LGTM

PR #3819, opened 2021-12-08
jmap_calendar: treat account as empty if not calendar is shared

rsto:


brong (approved):
This all looks sensible at a read through - certainly the Cassandane bugfixes do!  I'm assuming the behaviour we get is what we want.

(Also: this is deployed to all Fastmail users now)

PR #3821, opened 2021-12-09
DAV DB v12 -> v14 upgrade fix

ksmurchison:
Upgrading an empty v12 sieve_scripts table was causing a crash: mboxlist_lookup_allow_all() on a NULL mboxname fails, causing sievedb_upgrade() to fail, causing sqldb_open() to return a NULL DB handle to mailbox_update_dav|sieve()

### DIFF
--- a/imap/sieve_db.c
+++ b/imap/sieve_db.c
@@ -1071,8 +1071,6 @@ EXPORTED int sieve_script_rebuild(const char *userid,
     " WHERE rowid = :rowid;"
 
 #define CMD_UPDATE_v13_TABLE             \
-    "DELETE FROM sieve_scripts WHERE (rowid, modseq) NOT IN" \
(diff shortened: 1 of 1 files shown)


ksmurchison:
@brong We saw another issue with upgrades today.  Arenamed user (pending delete) failed to upgrade because setting mailbox to the mailbox id caused a UNIQUE constraint to be broken.  The user had 2 sieve_script records with the same IMAP UID (one with old mailbox name and one with the new).

My latest patch should remove all duplicates, leaving the record having the highest modseq for the given IMAP UID.  I assume that this is what we want, but in this user's case, the pendingdelete record had a lower modseq.  So that was somewhat confusing.

ksmurchison:
This code has been reviewed a couple times by Bron and has been running in production for over a week. Self-approving

PR #3823, opened 2021-12-09
more mailbox_legacy_dirs fixes

elliefm:
This fixes most (but not all) of the errors found by running cassandane with "mailbox_legacy_dirs = yes" (#3782).  Specifically, it fixes:

* quota_annot_used getting out of sync when "displayname" annotation is removed during rename
* List "tombstones" tests checking for MBTYPE_DELETED incorrectly, and failing when the MBTYPE_LEGACY_DIRS bit was also set

It also updates the FastMail.rename_deepfolder_intermediates and FastMail.rename_deepfolder_intermediates_rightnow tests to produce a useful error message for the particular failure they're experiencing -- I have no insight to the underlying problem yet, but at least now the errors are comprehensible for future debugging.

We might be able to do the quota_annot_used fix better, perhaps by adding a `delete_cb()` instead of continuing to abuse `rename_cb()`, but for now this works and is heavily commented.

I'll update #3782 with the status of the remaining failures.

### DIFF
--- a/cassandane/Cassandane/Cyrus/FastMail.pm
+++ b/cassandane/Cassandane/Cyrus/FastMail.pm
@@ -708,7 +708,7 @@ sub test_rename_deepfolder_intermediates
     $self->assert_null(grep { m/deleting intermediate with no children/ } @syslog);
 
     xlog $self, "Make sure there are no files left with cassandane in the name";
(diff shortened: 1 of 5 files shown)


ksmurchison (approved):
LGTM

PR #3824, opened 2021-12-10
XFER to legacy sieve

ksmurchison:
This properly XFERs a user from Cyrus 3.5 to 3.4.  The #sieve mailbox remains in the user mailbox list so we can properly do the required functions:

After initial sync, the #sieve mailbox is DEACTIVATEd on MUPDATE
During final sync, the #sieve mailbox is skipped (its contents have been sync'd via SIEVE commands)
On success, the #sieve mailbox is DELETEd both on the source server and on MUPDATE
On failure, the #sieve mailbox is reACTIVATEd on the source server on MUPDATE

This has been tested successfully with MurderIMAP.xfer_user_noaltns_nounixhs
(the other altns/unixhs combination tests fail and I'm not sure why yet -- annotations?)

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -11198,7 +11198,13 @@ struct xfer_header {
     struct xfer_item *items;
 };
 
-static int xfer_mupdate(int isactivate,
+enum {
+    MUPDATE_ACTIVATE,
+    MUPDATE_DEACTIVATE,
+    MUPDATE_DELETE
+};
+
+static int xfer_mupdate(int op,
                         const char *mboxname, const char *part,
                         const char *servername, const char *acl)
 {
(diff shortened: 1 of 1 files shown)


elliefm:
> This has been tested successfully with MurderIMAP.xfer_user_noaltns_nounixhs
(the other altns/unixhs combination tests fail and I'm not sure why yet -- annotations?)

That'll just be #3771 / #3781.  In the meantime, I know how to fiddle with those tests locally to avoid tripping the annotations bug, so I'll do that to make sure they don't break on something else instead

imap/imapd.c
```diff
@@ -11198,7 +11198,13 @@ struct xfer_header {
...
 };
 
-static int xfer_mupdate(int isactivate,
+enum {
+    MUPDATE_ACTIVATE,
+    MUPDATE_DEACTIVATE,
+    MUPDATE_DELETE
+};
+
+static int xfer_mupdate(int op,
```
elliefm, line 11207: I think I would name this enum, and then use it as the op type instead of `int`.  It'll get us a few extra compiler protections if the function knows it's an enumerated type and not just some number.

Something like:

```
enum mupdate_op {
   ...
};

static int xfer_mupdate(enum mupdate_op op,
```

imap/imapd.c
```diff
@@ -11411,8 +11420,8 @@ static int xfer_deactivate(struct xfer_header *xfer)
 
     /* Step 3: mupdate.DEACTIVATE(mailbox, newserver) */
     for (item = xfer->items; item; item = item->next) {
-        r = xfer_mupdate(0, item->mbentry->name, item->mbentry->partition,
-                         config_servername, item->mbentry->acl);
+        r = xfer_mupdate(MUPDATE_DEACTIVATE, item->mbentry->name,
+                         item->mbentry->partition, config_servername, NULL);
```
elliefm, line 11424: No longer using the ACL for deactivate?

elliefm (commented):
Looking good, couple of nits.  Will post back again later once tested

ksmurchison:
ACL isn't used for deactivate

elliefm (approved):
Looks good under testing!

imap/imapd.c
ksmurchison: Done

PR #3825, opened 2021-12-13
TestCase: clean up properly when _start_instances fails

elliefm:
By moving the eval from `_setup_http_service_objects()` up a layer, to encapsulate both calls.

This diff will be much easier to read if you add ?w=1 to the url (equivalent to `git diff -w`), since most of the lines changed are just indentation level changes.

Fixes a particular edge case where a failure late in `_start_instances()` would leave things partially started without cleaning up, and then later tests would fail due to the port they wanted to use being already in use.


rsto (approved):
LGTM

ksmurchison (approved):

Issue #3035, opened 2020-05-06
lmtpd crash with Cyrus 3.2.0


hagedose:
On a RHEL 7  test system with 3.2.0 I see the following behavior when I deliver a message to lmtpd with swaks:

```
swaks --to xxx --protocol LMTP --socket /var/lib/imap/socket/lmtp
=== Trying /var/lib/imap/socket/lmtp...
=== Connected to /var/lib/imap/socket/lmtp.
<-  220 cyrus3-staging.rrz.uni-koeln.de Cyrus LMTP 3.2.0-1.el7 Fedora server ready
 -> LHLO cyrus3-staging.rrz.uni-koeln.de
<-  250-cyrus3-staging.rrz.uni-koeln.de
<-  250-8BITMIME
<-  250-ENHANCEDSTATUSCODES
<-  250-PIPELINING
<-  250-SIZE
<-  250-AUTH EXTERNAL
<-  250-IGNOREQUOTA
<-  250 Ok SESSIONID=<cyrus-1588692869-52497-1-10721630997991278449>
 -> MAIL FROM:<root@cyrus3-staging.rrz.uni-koeln.de>
<-  250 2.1.0 ok
 -> RCPT TO:<xxx>
<-  250 2.1.5 ok
 -> DATA
<-  354 go ahead
 -> Date: Tue, 05 May 2020 17:34:29 +0200
 -> To: xxx
 -> From: root@cyrus3-staging.rrz.uni-koeln.de
 -> Subject: test Tue, 05 May 2020 17:34:29 +0200
 -> Message-Id: <20200505173429.052655@cyrus3-staging.rrz.uni-koeln.de>
 -> X-Mailer: swaks v20170101.0 jetmore.org/john/code/swaks/
 ->
 -> This is a test mailing
 ->
 -> .
*** Socket closed connection unexpectedly.
```

The corresponding log looks like this:

```
Mai 05 17:34:29 cyrus3-staging.rrz.uni-koeln.de cyrus/lmtpunix[52497]: command: LHLO cyrus3-staging.rrz.uni-koeln.de
Mai 05 17:34:29 cyrus3-staging.rrz.uni-koeln.de cyrus/lmtpunix[52497]: TLS is available.
Mai 05 17:34:29 cyrus3-staging.rrz.uni-koeln.de cyrus/lmtpunix[52497]: command: MAIL FROM:<root@cyrus3-staging.rrz.uni-koeln.de>
[...]

PR #3826, opened 2021-12-13
Avoid platform-dependent strerror(0)

cwalther:
As discussed in #3035:

On macOS, the Cyrus LMTP server responds `250 2.1.5 Undefined error: 0 SESSIONID=<cyrus-...>` in case of success. While there is nothing technically wrong with that as the protocol allows anything in the human-readable part of the message, it is somewhat confusing to read the word “error” in a mail server log when there is nothing wrong.

This comes from `send_lmtp_error()` → `error_message()` → `strerror(0)`, which on macOS returns “Undefined error: 0”, whereas on Linux/glibc it returns an appropriate “Success”. Other platforms may return something different entirely, the standards do not seem to specify it.

Avoid this by returning a constant “Success” from `error_message(0)`, which, matching the most common platform Linux, appears to be a known-good value also in the few other places than `send_lmtp_error()` that can call `error_message()` with 0.

### DIFF
--- a/com_err/et/error_message.c
+++ b/com_err/et/error_message.c
@@ -72,6 +72,14 @@ long code;
     int started = 0;
     char *cp;
 
+    /* Support for error number 0 differs among implementations of
+       strerror(). Linux/glibc returns "Success", which is nice, while
+       macOS returns "Undefined error: 0", which is rather confusing
+       in an LMTP success response "250 2.1.5 Undefined error: 0
+       SESSIONID=...". Return a known, consistent string instead. */
+    if (code == 0) {
+        return "Success";
+    }
     l_offset = code & ((1<<ERRCODE_RANGE)-1);
     offset = (int) l_offset;
     table_num = code - l_offset;


ksmurchison:
I'm not sure we want to go messing with the com_err source.  I think something like this would be sufficient:

[strerror.diff.txt](https://github.com/cyrusimap/cyrus-imapd/files/7705665/strerror.diff.txt)

cwalther:
Happy to do it that way if @elliefm who I was discussing this solution with agrees.

elliefm:
I'm also generally reluctant to mess around in `com_err`...

But I think our general expectation is that `error_message(0)` is sensible, because it is on Linux.   And so I'd rather encapsulate the MacOS workaround in `error_message()` itself, so we can continue to just expect that, rather than every caller needing to remember to think about it and potentially work around it locally.

ksmurchison:
Is our copy of error_message() always compiled in, or could certain platforms link in a system version?  It's been a while since I thought about com_err so I don't recall.

elliefm:
We do allow a system version in some cases, but I don't remember which cases, and the configure checks for it are complex, as I recall.  I do vaguely remember seeing comments to the effect of system com_err on Mac being bad, so I think we always use our version on Mac -- so it's not an unreasonable spot to drop in a Mac-specific workaround.

I can't remember whether my own builds on Debian use our com_err or a system one, but it doesn't matter anyway since my strerror(0) is fine...

cwalther:
I had no idea com_err was a system library, but sure enough, there it is:
```
# macOS 10.14
mojave:~ cwalther$ ls -l /usr/lib/libcom_err.dylib 
lrwxr-xr-x  1 root  wheel  54 Oct 10  2018 /usr/lib/libcom_err.dylib -> /System/Library/Frameworks/Kerberos.framework/Kerberos
# macOS 12.0:
cwalther@monterey ~ % ls -l /Library/Developer/CommandLineTools/SDKs/MacOSX12.0.sdk/usr/lib/libcom_err.tbd
lrwxr-xr-x  1 root  wheel  63 Nov  1 09:28 /Library/Developer/CommandLineTools/SDKs/MacOSX12.0.sdk/usr/lib/libcom_err.tbd -> ../../System/Library/Frameworks/Kerberos.framework/Kerberos.tbd
```

The corresponding header seems missing though:
```
cwalther@monterey ~ % cat comerr.c 
#include <stdio.h>
#include <string.h>
#include <et/com_err.h>

int main() {
	for (int i = 0; i < 3; i++) {
		printf("error_message(%d): %s   strerror(%d): %s\n", i, error_message(i), i, strerror(i));
	}
	return 0;
}
cwalther@monterey ~ % cc -o comerr comerr.c -lcom_err 
comerr.c:3:10: fatal error: 'et/com_err.h' file not found
#include <et/com_err.h>
         ^~~~~~~~~~~~~~
1 error generated.
```

I can use the one from MacPorts to make it compile, but the result seems somewhat useless:
```
cwalther@monterey ~ % cc -o comerr -I/opt/local/include comerr.c -lcom_err                              
cwalther@monterey ~ % ./comerr                                                                          
error_message(0): Unknown error 0   strerror(0): Undefined error: 0
error_message(1): Unknown error 1   strerror(1): Operation not permitted
error_message(2): Unknown error 2   strerror(2): No such file or directory
```

(Just in case this helps clear anything up.)

cwalther:
Correction: the header does exist as `<com_err.h>` (without the `et/`). Doesn’t change the useless output though, as the library is obviously the same.

cwalther:
I recently noticed that if you have [libcomerr](https://ports.macports.org/port/libcomerr/details/) from MacPorts installed, `configure` will pick up that one by default instead of using the bundled one, making my proposed fix ineffective. It has the same problem with error number 0 (presumably because it also just [calls](https://git.kernel.org/pub/scm/fs/ext2/e2fsprogs.git/tree/lib/et/error_message.c?h=v1.45.6&id=29d22c467547f11f4db31331549ee4e304eeb220#n139) `strerror()`). So maybe @ksmurchison’s fix from https://github.com/cyrusimap/cyrus-imapd/pull/3826#issuecomment-992700043 is better after all. Alternatively, replace all calls to `error_message()` by calls to a wrapper function that checks for 0 and otherwise calls the (either bundled or system-provided) `error_message()`? Or additionally implement a configure check for the output of `error_message(0)` – when it‘s bad, revert to the bundled one?

elliefm:
Trying to swap this back into memory...

> Or additionally implement a configure check for the output of `error_message(0)` – when it‘s bad, revert to the bundled one?

I think that would require:

* your patch to the cyrus com_err, such that our `error_message(0)` always returns "Success", skipping the `strerror()` call; and
* a configure check such that, when some pre-existing com_err is found, its `error_message(0)` is tested, and if it's found lacking, we ignore the pre-existing one and build the cyrus one instead

Is that what you had in mind?  That feels okay to me.

Alternatively, we could just make a point of considering `error_message(0)` to have undefined behaviour, and in the cases where the string output can be user-visible (e.g. lmtpd), explicitly handle `0` separately.  That would make Ken's patch in https://github.com/cyrusimap/cyrus-imapd/pull/3826#issuecomment-992700043 the right solution.  Though if we were doing that, I would like to also update our `error_message()` implementation with a big comment explaining that our `0` behaviour is not portable and should not be relied on where the text of the message will be user-visible, so that when we use our own copy's code as documentation of how to use it, we don't continue to assume `0` is fine because it works for us.

cwalther:
Yes, these are the options I had in mind.

ksmurchison:
@elliefm Do you want to work on this one, or should we just merge my patch to lmtpengine.c?

elliefm:
@ksmurchison For the amount of configure.ac work it'd take to do this cleverly, I don't think it's worth it.  I think we just apply your lmtpengine patch, perhaps with a comment that we can't rely on error_message(0) to produce a success string.

elliefm:
This is now #5073

cwalther:
Thanks for taking this up!

PR #3828, opened 2021-12-14
Caldav calendar query fix

ksmurchison:
This fixes an issue where the calendar-query wasn't being executed on certain resources because of a bad initialization of 'ical'.

This also adds a calendar-query test which exercises time-range and component filtering, recurrence expansion, and partial event retrieval.


rsto (approved):
LGTM

PR #3829, opened 2021-12-14
Flags: check for corrrect untagged flags responses from STORE/FETCH

elliefm:
This was previously: https://github.com/cyrusimap/cassandane/pull/142

Two commits here: one is a new regression test for #3240 / #3577, the other adds similar hardening to the tests that already existed.

At the moment, the regression tests proves that the bug #3240 still exists, and also proves that #3577 fixes it.  Though the fix in #3577 is believed to also double-bump modseqs, which is bad, and is why it hasn't landed.

I probably want to update the tests further to ensure modseqs aren't double-bumped, but I have not yet done so.


elliefm:
These tests were merged in #4240

PR #3832, opened 2021-12-16
json_is_date()

ksmurchison:
Add json_is_date() function for future use (will be used by jmap_files branch)

### DIFF
--- a/imap/json_support.c
+++ b/imap/json_support.c
@@ -49,15 +49,42 @@
 
 #include "json_support.h"
 
-int json_is_utcdate(json_t *json)
+static int parse_date(json_t *json, unsigned utc)
 {
     const char *s = NULL;
     struct tm date;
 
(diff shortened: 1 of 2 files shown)


rsto (approved):
Looks good to me. Note that there is something very similar in the jmap_ical code: https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/jmap_ical.h#L112

ksmurchison:
@rsto  I hadn't noticed those functions.  jmapical_utcdatetime_from_string() would certainly work.  Do JMAP Date values NOT have/allow numeric TZ offsets?

rsto:
@ksmurchison JMAP `Date` values do allow an offset, but JSCalendar `LocalDateTime` and `UTCDateTime` do not. Cool, so the `json_is_date` function does close that gap!

PR #3836, opened 2021-12-21
Quota f u fix

brong:
This fixes the use of `quota -f -u` so it doesn't wipe sub quotaroots, as found in Fastmail production.

### DIFF
--- a/imap/quota.c
+++ b/imap/quota.c
@@ -315,12 +315,15 @@ static void test_sync_done(const char *mboxname)
 /*
  * A quotaroot was found, add it to our list
  */
-static int fixquota_addroot(struct quota *q,
-                            void *rock __attribute__((unused)))
+static int fixquota_addroot(struct quota *q, void *rock)
 {
     struct quota localq;
     struct txn *tid = NULL;
+    const char *userid = (const char *)rock;
     int r;
+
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):

elliefm:
Can I get a ping when this is merged please?  I'll need to cherry-pick or backport the fix to 3.2 and 3.4, which I believe are affected by this bug too

elliefm (approved):

imap/quota.c
```diff
@@ -387,7 +390,7 @@ int buildquotalist(char *domain, char **roots, int nroots, int isuser)
 
     /* basic case - everything (potentially limited by domain still) */
     if (!nroots) {
-        r = quota_foreach(buf, fixquota_addroot, buf, NULL);
+        r = quota_foreach(buf, fixquota_addroot, NULL, NULL);
```
elliefm, line 393: Hmm.  Does this changed line mean `quota -d domain` is no longer limited by domain?

We only have one test that exercises the `-d` option, and it doesn't check that other domains aren't affected.

elliefm: Oh no, the first argument is still "buf". The changed argument is just the fixquota_addroot rock, which was previously unused and is now a userid.  Confused myself!

Issue #3835, opened 2021-12-20
rename user loses sieve bytecode when mailbox_legacy_dirs = yes


elliefm:
To reproduce:

* add `mailbox_legacy_dirs = yes` to the `[config]` section of cassandane.ini
* run the `Rename.rename_user_sieve` test

The test will fail like this:

```
1) Rename.rename_user_sieve
 /dev/shm/cass/0230330101/conf/sieve/n/newuser/test1.bc: file not found at /usr/share/perl5/Test/Unit/Exception.pm line 13.
	Test::Unit::Exception::throw_new(Test::Unit::Failure=HASH(0x55deaab1f7f0), "-package", "Cassandane::Cyrus::TestCase", "-file", "Cassandane/Cyrus/TestCase.pm", "-line", 1389, "-object", ...) called at /usr/share/perl5/Test/Unit/Assert.pm line 85
	Test::Unit::Assert::do_assertion(Cassandane::Cyrus::Rename=HASH(0x55deaaab5078), Test::Unit::Assertion::Boolean=SCALAR(0x55deaab04130), "Cassandane::Cyrus::TestCase", "Cassandane/Cyrus/TestCase.pm", 1389, "/dev/shm/cass/0230330101/conf/sieve/n/newuser/test1.bc: file "...) called at /usr/share/perl5/Test/Unit/Assert.pm line 19
	Test::Unit::Assert::assert(Cassandane::Cyrus::Rename=HASH(0x55deaaab5078), undef, "/dev/shm/cass/0230330101/conf/sieve/n/newuser/test1.bc: file "...) called at Cassandane/Cyrus/TestCase.pm line 1389
	Cassandane::Cyrus::TestCase::assert_sieve_exists(Cassandane::Cyrus::Rename=HASH(0x55deaaab5078), Cassandane::Instance=HASH(0x55deaaac3450), "newuser", "test1", 1) called at Cassandane/Cyrus/Rename.pm line 658
	Cassandane::Cyrus::Rename::test_rename_user_sieve(Cassandane::Cyrus::Rename=HASH(0x55deaaab5078)) called at /usr/share/perl5/Test/Unit/TestCase.pm line 75
	[...framework calls elided...]
```

which is coming from here:
https://github.com/cyrusimap/cyrus-imapd/blob/522d3786d94cd61af5209a5b5d4c13c7acf63d0d/cassandane/Cassandane/Cyrus/Rename.pm#L658

The `rename` command has succeeded:

```
[...]

PR #3837, opened 2021-12-21
Legacy user rename fix

ksmurchison:
Fix for #3835


elliefm:
This looks okay, but it's still failing for me in the same way (bytecode directory does not get renamed).  Looking closely, it logs this error:

```
Dec 22 11:20:47 debian 0020450101/imap[3436]: error renaming /dev/shm/cass/0020450101/conf/sieve/c/cassandane to /dev/shm/cass/0020450101/conf/sieve/n/newuser: No such file or directory
```

Looks like it's complaining that the `/n/` subdirectory doesn't exist yet.  On a real system this probably usually exists (I guess it's rare to rename someone to an initial that's never been used before), but under Cassandane when the whole data layout is new at the start of the test, it doesn't exist yet.

This local patch got me moving forward:

```
diff --git a/imap/user.c b/imap/user.c
index 4b71201748..7f66a1bf3e 100644
--- a/imap/user.c
+++ b/imap/user.c
@@ -431,11 +431,14 @@ static int user_renamesieve(const char *olduser, const char *newuser
                        "oldpath=<%s> newpath=<%s>",
                        oldpath, newpath);
 
+    /* make sure newpath's directory components exist */
+    r = cyrus_mkdir(newpath, 0755 /* unused */);
+
     /* rename sieve directory
      *
      * XXX this doesn't rename sieve scripts
      */
-    r = rename(oldpath, newpath);
+    if (!r) r = rename(oldpath, newpath);
     if (r < 0) {
         if (errno == ENOENT) {
             syslog(LOG_WARNING, "error renaming %s to %s: %m",
```

With this patch, the `master` copy of the bytecode directory is now renamed correctly and the test can proceed....

And at this point, the test now fails because the bytecode directory on the `replica` hasn't been renamed.

This is an interesting position, because this version supports the SIEVE-MAILBOX capability, so sync_client doesn't try to do any APPLY SIEVE commands, it just sends a RENAME for the #sieve mailbox.

Looks like `sync_apply_rename` doesn't have any special-casing for when the mailbox is MBTYPE_SIEVE, it just treats it as a normal mailbox -- which works fine for uuid-mailboxes, but for legacy mailboxes it's missing the user_renamesieve handling just like `cmd_rename` was.

imap/imapd.c
```diff
@@ -7910,6 +7910,10 @@ static void cmd_rename(char *tag, char *oldname, char *newname, char *location)
     if (!r && rename_user) {
         user_copyquotaroot(oldmailboxname, newmailboxname);
         user_renameacl(&imapd_namespace, newmailboxname, olduser, newuser);
+        if (config_getswitch(IMAPOPT_MAILBOX_LEGACY_DIRS)) {
```
elliefm, line 7913: Hmmm.  Do we also need to check for MBTYPE_LEGACY_DIRS here?  Like, maybe the server is configured to use uuid-mailboxes, but this particular user hasn't been converted yet.  I'm not sure what the between-states look like but it feels important to catch them, if they can exist.

ksmurchison: Good point.  Fixed

ksmurchison:
@elliefm I applied your cyrus_mkdir() patch and pushed 3 others which now allow all Rename tests to pass.

I mistakenly had mailbox_legacy_dirs=yes in the wrong section of cassandane.ini when I tested yesterday, so I wasn't really testing that config at all.

brong:
Yeah, we can't rely on "letter has probably been used before", because for sure it will make things messy.

elliefm (approved):

imap/mboxlist.c
```diff
@@ -2814,9 +2814,15 @@ EXPORTED int mboxlist_renamemailbox(const mbentry_t *mbentry,
...
+            if (mbtype_isa(mailbox_mbtype(newmailbox)) == MBTYPE_SIEVE) {
+#ifdef USE_SIEVE
+                mailbox_add_sieve(newmailbox);
+#endif
 #ifdef WITH_DAV
-            mailbox_add_dav(newmailbox);
+            } else {
+                mailbox_add_dav(newmailbox);
 #endif
+            }
```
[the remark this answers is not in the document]
ksmurchison: Compiles successfully on top of #3845 (in all 4 combos).

brong (approved):
Yeah, looks good.  It's annoying supporting so many modes, but this does let us keep running cleanly for users who have just upgraded Cyrus and not yet upgraded all their locations!

PR #3840, opened 2021-12-22
Quota: test quota -d

elliefm:
Confused myself and thought I'd found a bug in `quota -d`.  Wrote a test to prove it, instead proved it was fine.  But at least we have a new test now.

This tests `quota -d domain` to ensure that both:

* quotas within the specified domain are reported
* quotas in other domains are not


brong (approved):
Great!  Thanks for doing this :)  Looks good.

PR #3284, opened 2020-11-09
JMAP: only send alerts for VEVENTs

brong:
This will make dogfood less horrible for people in the meanwhile!

[ not to be merged:  problem will be solved with an alternate strategy ]

PR #3841, opened 2021-12-23
Fastmail-only: ignore RFC-draft calendars capability

rsto:
Ignores the future JMAP Calendars IETF capability rather than rejecting requests.

Do not merge this PR.


ksmurchison (approved):

rsto:
We already have something similar at https://github.com/cyrusimap/cyrus-imapd/pull/3284

PR #3845, opened 2021-12-29
DAV/Sieve config fixes

ksmurchison:
Fixes to allow all four combinations of DAV +/- Sieve to be compiled


rsto:
I think I miss the back story to what motivated this change. Is it just for compilation? Or are there further requirements how to handle Cyrus data for all these combinations?

imap/dav_db.c
```diff
@@ -256,14 +256,39 @@ struct sqldb_upgrade davdb_upgrade[] = {
   /* Don't upgrade to version 11.  We only jump to 11 on CREATE */
   /* Don't upgrade to version 12.  This was an intermediate Sieve DB version */
   /* Don't upgrade to version 13.  This was an intermediate Sieve DB version */
-  { 14, CMD_DBUPGRADEv14, &sievedb_upgrade },
+  { 14, CMD_DBUPGRADEv14,
```
brong, line 259: Oh man - I hate this complexity level :(  We should just require sieve to be included.

brong (approved):
This looks fine - but I'd also prefer not to have so many optional compilation units!

imap/dav_db.c
ksmurchison: I think my last commit makes this "better" for some value of "better".  However, it does guarantee that the Sieve table schema is correct if some installation was at v12/v13 and disabled Sieve for v14 (highly unlikely, but possible).

elliefm:
> I think I miss the back story to what motivated this change. Is it just for compilation? Or are there further requirements how to handle Cyrus data for all these combinations?

It came up because I noticed a potentially-weird option combination while reviewing #3837 a couple of weeks ago, did a quick test to see if it was a real problem, and found that it wasn't, but that the same combination triggered compilation problems elsewhere.  Maybe I should've opened an issue for it; instead the history is buried in one of my "[outdated]" review comments on that PR.

PR #3850, opened 2022-01-04
Mboxlist tooling

brong:
This does some initial rewrites of mboxlist handling for history and updates:

- adds partition, mbtype, uidvalidity to the history entries.  This will come in handy later
- makes sure that the N key and I key updates are atomic by starting a transaction if there isn't one
- refactors the code a bit to support the above


ksmurchison (approved):
This all looks sane, provided it checks out in testing.

brong:
It all passes Cassanane OK :)

PR #3851, opened 2022-01-04
Fastmail-only: rollback jmap-calendars-01 deployment

rsto:
This PR is not meant to merge on the main branch. It rolls back dav.db changes of the jmap-calendars-01 branch which were introduced in version 15 and sets the current version to 16.

I manually (and only rudimentary) tested this by running CalendarEvent/get on this branch both on a Cyrus environment with current master and jmap-calendars-01.

This should cover the majority of changes introduced by jmap-calendars-01. It does not rollback newly introduced ACLs (like mayUpdatePrivate), so these only should be set once we are sure to not want to rollback (or defined a downgrade path for ACLs).

This PR is more a work in progress, but I'd appreciate an early review.


ksmurchison (approved):
LGTM.

rsto:
Closing, we are not going to roll back.

PR #3855, opened 2022-01-07
sync_client: in -A mode, reconnect if error

brong:
I added the "don't bail on error" code a few months back, but if the error caused the connection to be dropped by the replica, we just keep trying over and over on a dead connection.  Instead, if there's an error we should reconnect before moving along to the next user.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Replication.pm
+++ b/cassandane/Cassandane/Cyrus/Replication.pm
@@ -261,7 +261,7 @@ EOF
 # Test handling of replication when append fails due to disk error
 #
 sub test_syncall_failinguser
-    :NoStartInstances :min_version_3_5
+    :NoStartInstances :min_version_3_6
 {
     my ($self) = @_;
 
--- a/cassandane/Cassandane/Cyrus/Replication.pm
+++ b/cassandane/Cassandane/Cyrus/Replication.pm
@@ -258,6 +258,121 @@ EOF
 }
 
 #
+# Test handling of replication when append fails due to disk error
+#
+sub test_syncall_failinguser
+    :NoStartInstances :min_version_3_5
+{
+    my ($self) = @_;
+
+    my $canary = << 'EOF';
+From: Fred J. Bloggs <fbloggs@fastmail.fm>
+To: Sarah Jane Smith <sjsmith@tard.is>
+Subject: this is just to say
+X-Cassandane-Unique: canary
+
+I have eaten
+the canary
+that was in
+the coal mine
+
+and which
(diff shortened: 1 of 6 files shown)


elliefm:
We talked about this out of band and you were going to add a Cassandane test based on the mechanism used by the `Replication.append*_diskfull` tests to make sure this works as expected, and continues with the remaining users even after one fails.

I don't see it in the new commits since last time I looked though, is it still in progress?

brong:
@elliefm yeah, I'll do that still!

brong:
@elliefm test case added too :)

cassandane/Cassandane/Cyrus/Replication.pm
```diff
@@ -257,6 +257,121 @@ EOF
     }
 }
 
+#
+# Test handling of replication when append fails due to disk error
+#
+sub test_syncall_failinguser
+    :NoStartInstances :min_version_3_5
```
elliefm, line 264: I plan to cherry-pick these fixes so it _will_ be on 3.6, otherwise I'd be asking for `:min_version_3_7` here...

```suggestion
    :NoStartInstances :min_version_3_6
```

cassandane/Cassandane/Cyrus/TestCase.pm
```diff
@@ -1179,6 +1181,7 @@ sub run_replication
     push(@cmd, '-O') if defined $nosyncback;
     push(@cmd, '-u', $user) if defined $user;
     push(@cmd, '-m', $mailbox) if defined $mailbox;
+    push(@cmd, '-A') if defined $allusers;
```
elliefm, line 1184: I think this `defined` is nqr... if you `run_replication(allusers => 0);` then you'll unexpectedly get the -A behaviour, even though you though you explicitly said you didn't want it, because `0` is not `undef`.

`defined` is kinda bogus for the existing options too, but it's less likely to cause a real problem because `run_replication(user => 0)`, `run_replication(mailbox => '')`, etc aren't really coherent things to try to do anyway.

```suggestion
    push(@cmd, '-A') if $allusers;  # n.b. boolean
```

... or fix them all, so that $allusers doesn't need a comment about why it's different from the `defined` ones

elliefm (approved):
Looks good!  Couple of nits in the tests, posted fixes as "suggestions".   Approving now so we don't need another cycle around. :)

brong:
Thanks @elliefm I included both those suggestions :)

elliefm:
This has been cherry-picked to cyrus-imapd-3.6

PR #3856, opened 2022-01-07
trivial "make dist" fixes

elliefm:
Well, one fix, one itch scratched:

* `make dist` was failing because imap/http_caldav.h no longer exists (was renamed to "caldav_util.h"), but it was still listed in a `..._SOURCES` line in Makefile.am.  So make would try to include it in the tarball, not find it, try to build it, not find a rule to build it, and complain that there was no rule...  
* imap/rfc822_header.c is automatically generated, but we keep it in the repository for the reasons described in 4f4c46d388.  Since I updated to buster (and a newer `gperf` that produces slightly different code), every time I make a release, my repository ends up dirty, which confuses me for a moment.  So this time I'm committing the new version to save myself the confusion for a while.  We previously did the same thing in 6ed3e47e95.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1255,7 +1255,6 @@ imap_httpd_SOURCES = \
 	imap/http_admin.c \
 	imap/http_applepush.c \
 	imap/http_caldav.c \
-	imap/http_caldav.h \
 	imap/http_caldav_js.h \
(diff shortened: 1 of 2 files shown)


rsto (approved):
LGTM

PR #3857, opened 2022-01-07
cassandane: Add new_test_url() and Cassandane::Test::URL for easy HTT…

wolfsage:
…P servers

See t/new-test-url.t for more examples, but basically this let's you:

    my $server = $self->new_test_url("some string response");

    # Now $server->url contains an http://... url to a running service
    # that will respond to HTTP requests

You can also have the http service running custom code (which is a Plack
application):

    my $server = $self->new_test_url(sub {
      my $env = shift;

      my $req = Plack::Request->new($env);

      # ...

      return [
        200,
        [ "Content-Type" => "text/plain" ],
        [ "The body" ],
      ];
    });

You may also update the responses:

    $server->update("new response");

... which does this by stopping/restarting the service with the new code.

These servers automatically go away when the object goes out of scope.
Alternatively, you can call $server->unregister to end them.

### DIFF
--- a/cassandane/Cassandane/Test/NewTestUrl.pm
+++ b/cassandane/Cassandane/Test/NewTestUrl.pm
@@ -1,0 +1,133 @@
+#!/usr/bin/perl
+
+package Cassandane::Test::NewTestUrl;
+use strict;
+use warnings;
+
+use JSON;
+use LWP::UserAgent;
+
+use lib '.';
+use base qw(Cassandane::Unit::TestCase);
+
+sub new
+{
+    my $class = shift;
+    my $self = $class->SUPER::new(@_);
+    return $self;
+}
+
+sub test_basic
+{
+    my ($self) = @_;
+
+    my $string_based_url = $self->new_test_url("string based");
+    my $code_based_url = $self->new_test_url(sub {
+        return [
+            201,
+            [],
+            [ "code based" ],
+        ],
+    });
+
+    my $lwp = LWP::UserAgent->new;
+
+    {
+        my $res = $lwp->get($string_based_url->url);
+        $self->assert_str_equals('200', $res->code);
+        $self->assert_str_equals('string based', $res->decoded_content);
+    }
+
+    {
+        my $res = $lwp->get($code_based_url->url);
+        $self->assert_str_equals('201', $res->code);
+        $self->assert_str_equals('code based', $res->decoded_content);
+    }
+
+    $string_based_url->unregister;
+
+    {
+        # Unregistering one shouldn't affect the other
+        my $res = $lwp->get($code_based_url->url);
+        $self->assert_str_equals('201', $res->code);
+        $self->assert_str_equals('code based', $res->decoded_content);
+    }
+
+    $code_based_url->update("newval");
+
+    {
+        my $res = $lwp->get($code_based_url->url);
+        $self->assert_str_equals('200', $res->code);
+        $self->assert_str_equals('newval', $res->decoded_content);
+    }
+
+    {
+        eval { $string_based_url->url };
+        my $err = $@;
+        $self->assert_matches(
+            qr/\QCannot call ->url after ->unregister has been called\E/,
+            $err,
+        );
+    }
+
+    {
+        eval { $string_based_url->update("foo") };
+        my $err = $@;
+        $self->assert_matches(
+            qr/\QCannot call ->update after ->unregister has been called\E/,
+            $err,
+        );
+    }
+
+    # Plack example
+    my $plack_based_url = $self->new_test_url(sub {
+        my $env = shift;
+        my $req = Plack::Request->new($env);
+
+        my $payload = decode_json($req->raw_body);
+
+        my $res;
+
+        if ($payload->{good}) {
+            $res = Plack::Response->new(200);
+            $res->content_type('application/json');
+            $res->body(encode_json({ good => "job" }));
+        } else {
+            $res = Plack::Response->new(400);
+            $res->content_type('application/json');
+            $res->body(encode_json({ tough => "luck" }));
+        }
+
+        return $res->finalize;
+    });
+
+    {
+        my $res = $lwp->post(
+            $plack_based_url->url,
+            Content => encode_json({ good => 1 }),
+        );
+        $self->assert_str_equals('200', $res->code);
+
+        my $json = decode_json($res->decoded_content);
+        $self->assert_deep_equals(
+            { good => "job" },
+            $json,
+        );
+    }
+
+    {
+        my $res = $lwp->post(
+            $plack_based_url->url,
+            Content => encode_json({ good => 0 }),
+        );
+        $self->assert_str_equals('400', $res->code);
+
+        my $json = decode_json($res->decoded_content);
+        $self->assert_deep_equals(
+            { tough => "luck" },
+            $json,
+        );
+    }
+}
+
+1;
--- a/cassandane/Cassandane/Unit/TestCase.pm
+++ b/cassandane/Cassandane/Unit/TestCase.pm
@@ -46,6 +46,7 @@ use Data::Dumper;
 
 use lib '.';
 use Cassandane::Util::Log;
+use Cassandane::Util::TestUrl;
 
 my $enabled;
 my $buildinfo;
@@ -327,6 +328,14 @@ sub assert_num_lte
 
     $self->assert(($actual <= $expected),
                   "$actual is not less-than-or-equal-to $expected");
+}
+
+sub new_test_url {
+    my ($self, $content_or_app) = @_;
+
+    return Cassandane::Util::TestURL->new({
+        app => $content_or_app,
+    });
 }
 
 1;
--- a/cassandane/Cassandane/Util/TestUrl.pm
+++ b/cassandane/Cassandane/Util/TestUrl.pm
@@ -1,0 +1,112 @@
+package Cassandane::Util::TestURL;
+
+use Plack::Loader;
+use Plack::Request;
+use Plack::Response;
+
+use Test::TCP;
+use Carp qw(croak);
+
+use lib '.';
+use Cassandane::PortManager;
+
+sub new
+{
+    my ($pkg, $args) = @_;
+
+    my %attrs;
+
+    for my $required (qw(app)) {
+        $attrs{$required} = delete $args->{$required};
+        unless ($attrs{$required}) {
+            croak("'$required' required for Cassandane::Test::URL->new");
+        }
+    }
+
+    my $self = bless {}, __PACKAGE__;
+
+    $self->update($attrs{app});
+
+    return $self;
+}
+
+sub url
+{
+    my ($self) = @_;
+
+    if ($self->was_unregistered) {
+        croak("Cannot call ->url after ->unregister has been called!");
+    }
+
+    $self->{url};
+}
+
+sub _guard { shift->{_guard} }
+
+sub unregister
+{
+    my ($self) = @_;
+
+    delete $self->{_guard};
+    $self->{was_unregistered} = 1;
+}
+
+sub was_unregistered { shift->{was_unregistered} }
+
+sub update
+{
+    my ($self, $content_or_app) = @_;
+
+    unless (ref $content_or_app) {
+        my $content = $content_or_app;
+        # Plain ol' successful response
+        $content_or_app = sub {
+            return [
+                200,
+                [],
+                [ $content ],
+            ];
+        };
+    }
+
+    my $app = sub {
+        my $sock_or_port = shift;
+        my $server = Plack::Loader->auto(
+            host => '0.0.0.0',
+            port => $sock_or_port,
+        );
+        $server->run($content_or_app);
+        exit;
+    };
+
+    unless ($self->_guard) {
+        if ($self->was_unregistered) {
+            # We've already been unregistered. It's no longer safe to call
+            # update because our port may have been reused by someone else
+            # and our url is no longer useable
+            croak("Cannot call ->update after ->unregister has been called!");
+        }
+
+        my $host = "127.0.0.1";
+
+        my $guard = Test::TCP->new(
+            host => $host,
+            code => $app,
+            port => Cassandane::PortManager::alloc(),
+        );
+
+        my $port = $guard->port;
+
+        $self->{url} = "http://$host:$port/";
+        $self->{_guard} = $guard;
+    } else {
+        $self->_guard->{code} = $app;
+
+        $self->_guard->stop;
+        $self->_guard->start;
+    }
+
+    return;
+}
+
+1;
--- a/cassandane/Cassandane/Util/TestUrl.pm
+++ b/cassandane/Cassandane/Util/TestUrl.pm
@@ -69,14 +69,38 @@ sub update
         };
     }
 
+    {
+      package Shutdown::Guard;
+
+      use POSIX qw(_exit);
+
+      sub new { bless {}, __PACKAGE__ }
+
+      sub DESTROY {
+          _exit(0);
+      }
+    }
+
     my $app = sub {
+        # No matter how we we leave this block, the $guard will go out
+        # of scope while in the run phase, which means it can exit Perl
+        # before any END block or object destructors are run.
+        my $guard = Shutdown::Guard->new;
+
+        # Test::TCP uses Term to stop us. We could just let that
+        # terminate Perl cleanly, but in case something has put in a
+        # signal handle or something else, let's just call exit here.
+        # This will also cause our $guard above to go out of scope
+        # in a timely manner during the run phase.
+        local $SIG{TERM} = sub { exit; };
+
         my $sock_or_port = shift;
         my $server = Plack::Loader->auto(
             host => '0.0.0.0',
             port => $sock_or_port,
         );
+
         $server->run($content_or_app);
-        exit;
     };
 
     unless ($self->_guard) {
--- a/cassandane/doc/README.deps
+++ b/cassandane/doc/README.deps
@@ -2,6 +2,12 @@ SOFTWARE DEPENDENCIES
 
 Cassandane needs the following software to work:
 
+ * LWP::UserAgent
+ * Plack
+ * Test::TCP
+   Used for standing up local test HTTP servers for testing various bits of
+   Cyrus callout capabilities
+
  * Mail::IMAPTalk
     This is a fully featured IMAP client package, available on CPAN.
 


wolfsage:
This includes a few new deps (though I suspect some like Test::More/LWP::UserAgent would already be installed). The new t/ file is there purely to test this functionality so making changes to it is easy, but won't be part of normal cassandane runs. (We could switch that to be a proper cassandane unit test file if we want...)

With this, it should be very easy to write tests for things like Group ACLs which can talk to an HTTP service

elliefm (commented):
> (We could switch that to be a proper cassandane unit test file if we want...)

Yes please.  If we've got it it might as well always run, like the rest of the self-tests.  Also, the existence of that "t/" directory might mislead future delvers into thinking Cassandane is anything like normal perl. ;)

If the self-test is absurdly expensive for some reason (it doesn't look it), and undesirable for it to always run, we have a mechanism for marking tests to only run when testrunner.pl is called with `--slow`.  We don't have a strictly defined threshold for what constitutes "absurdly expensive and undesirable to always run", but `ack 'sub test_.*_slow' cassandane/` shows the three tests we think this about.

> These servers automatically go away when the object goes out of scope.

Cassandane usually wants to control the port numbers of things that spawn underneath it, partly so that it can make sure there aren't conflicts/overlaps between parallel-running tests, but also so it can make sure they don't hang around after the test finishes.  It does a middling job at these tasks.  

It does expect to own everything in the 91xx range (or `base_port` and the next hundredish ports).  I assume Plack or Test::TCP is going to bind a random port since none is given explicitly; if it chooses one in the range Cassandane expects to own, Cassandane may be confused.  So it would be nice if `new_test_url` (or something underneath it) would use Cassandane::PortManager to get a port to use, rather than letting one be chosen at random.

wolfsage:
> > (We could switch that to be a proper cassandane unit test file if we want...)
> 
> Yes please. If we've got it it might as well always run, like the rest of the self-tests. Also, the existence of that "t/" directory might mislead future delvers into thinking Cassandane is anything like normal perl. ;)
Done.

> It does expect to own everything in the 91xx range (or `base_port` and the next hundredish ports). I assume Plack or Test::TCP is going to bind a random port since none is given explicitly; if it chooses one in the range Cassandane expects to own, Cassandane may be confused. So it would be nice if `new_test_url` (or something underneath it) would use Cassandane::PortManager to get a port to use, rather than letting one be chosen at random.

Done. Though one thing Cassandane will do is attempt to reuse ports without checking if they are open (which can cause failures between test runs if something is left over).

Test::TCP will only ever give you an open port (it tries really hard.)

cassandane/Cassandane/Unit/TestCase.pm
```diff
@@ -329,4 +329,120 @@ sub assert_num_lte
                   "$actual is not less-than-or-equal-to $expected");
 }
 
+{
+  package Cassandane::Test::URL;
```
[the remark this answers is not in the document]
wolfsage: Often these little packages are so niche and just provide the underlying behaviour of something returned by a small function, so having them in their own file can be a bit much, but in this case I don't think it's a problem so I've moved it to Cassandane::Util::TestURL

cassandane/Cassandane/Unit/TestCase.pm
```diff
@@ -329,4 +329,120 @@ sub assert_num_lte
...
+    my ($pkg, $args) = @_;
+
+    my %attrs;
+
+    for my $required (qw(app)) {
+      $attrs{$required} = delete $args->{$required};
+      unless ($attrs{$required}) {
+        require Carp;
+
+        Carp::croak("'$required' required for Cassandane::Test::URL->new");
```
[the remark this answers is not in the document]
wolfsage: Bad old habit - it's common in perl core not to import carp until you need it to avoid bloating memory etc etc. It's silly and I've removed it

cassandane/Cassandane/Unit/TestCase.pm
```diff
@@ -329,4 +329,120 @@ sub assert_num_lte
...
+    }
+
+    my $app = sub {
+      my $sock_or_port = shift;
+      my $server = Plack::Loader->auto(
+        host => '0.0.0.0',
+        port => $sock_or_port,
+      );
+      $server->run($content_or_app);
+      exit;
```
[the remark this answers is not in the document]
wolfsage: > I don't know Plack or Test::TCP. Are they well-behaved with respect to... actually I don't have a word for this...
> 
> Okay let's imagine this scenario:
> 
>     * I have written a Cassandane test
> 
>     * When the test runs, the entirety of Cassandane, including all its own globals, is in scope
> 
>     * My test uses this to spawn off a fake http server for the Cyrus-under-test to interact with
> 
>     * Eventually, that server calls `exit` here
> 
>     * Is the server's exit going to try to `DESTROY` everything it has in scope, including all of Cassandane's globals?
> 
> 
> I can't see where the fork happens (I'm not about to go reading the Plack or Test::TCP code), so I can't see whether the fork execs a new image (good), or keeps running a clone of the original image (bad). I ask because I've already cleaned up a few messes like this in Cassandane's internals, and there's at least one remaining that I haven't gotten to yet, so I'm wary of another one sliding in unobserved.
> 
> I would like to assume respected modules like Plack/Test::TCP behave well, but I don't _know_ that, so I'm asking. :)

I don't think its on any module that forks to behave well, because there's not an easy to *to* behave well.

Any time perl forks and does not exec, you'll have this problem. I think its on the destructors to ensure they aren't being in run in the process they need to be run in if it can cause problems.

Usually I do this by taking the parent PID when the destructor is set up and refusing to run if the current pid does not match:

```
package Foo;
my $PARENT = $$;

sub DESTROY {
  return if $$ != $PARENT;
}
```

etc etc

You could attempt to have all of your children that fork and don't exec exit using POSIX::_exit(), but you'd have to account for any signal handlers that cause perl to exit normally or die. (IE: If there's an exception, perl will throw a $SIG{__DIE__}, then your _exit() won't be called unless you set up something to catch __DIE__, because even die() calls destructors on exit)

I see the few places where we fork we do something with $SIG{TERM} to call _exit(0). We could add that, because Test::TCP uses $SIG{TERM} when stop() is called or when its guard goes out of scope.

wolfsage:
@elliefm I think this is ready for re-review - I've fixed most of the things you mentioned and left the `fork()` behaviour as an open discussion, where would you like to go from here?

Once we're happy with everything I can squash all commits into one and sign it off.

Thanks.

cassandane/Cassandane/Unit/TestCase.pm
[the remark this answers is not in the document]
elliefm: I guess I see that; on the other hand I find it deeply confusing when I'm in an unfamiliar file hunting around trying to solve a bug, or understand a behaviour, and I search for something like "^sub new" and end up reading entirely the wrong constructor, or whatever.

(This gets worse when there's a base class and subclasses of it defined in the same file, because then there's _lots_ of "subroutine with the same name but it's not the one you're looking for", but that's not the case here.)

cassandane/Cassandane/Unit/TestCase.pm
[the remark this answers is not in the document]
elliefm: We (newly) do something like this already in the one place it's burned us so far: https://github.com/cyrusimap/cyrus-imapd/blob/5fa7876cff6671794412f65e8241b398cbf0e4d8/cassandane/Cassandane/Instance.pm#L1538-L1541

But Cassandane is big and thorny and I'm sure there's many more delicate bits, they just haven't been kicked yet.

The two(?) things I'm aware of that still fork-without-exec are `Cassandane::Instance::_start_smtpd` and `Cassandane::Instance::start_httpd`, and I have littered both with comments reminding myself to rip them out into standalone scripts that can be fork-and-exec'd (or started by Cyrus rather than by Cassandane), like I already did with `utils/fake*d`.

It's interesting to dig around, because now that I have, I think `start_httpd` could be replaced by this thing you're making.  For that matter, in hindsight I think `start_httpd` is what I was thinking of when I told Ken that I thought we already had something like this.

> I see the few places where we fork we do something with `$SIG{TERM}` to call `_exit(0)`. We could add that, [...]

If those "few places" are `_start_smtp` and `start_httpd`, then I consider them terrible models we shouldn't copy ;)

In the case of this thing (and `start_httpd`), where the behaviour of the spawned server is defined by a sub provided by the caller, I'm not sure off the top of my head how to handle it cleanly.  It's easier for the other two, cause one's behaviour is static, and the other's is defined by a file.

The alternative -- protecting DESTROY -- is awkward since Cassandane expects that it's controlling all the forking (except for that done by Cyrus), and we'd probably need to do this in every extant file if we're relinquishing that.

cassandane/Cassandane/Test/NewTestUrl.pm
```diff
@@ -0,0 +1,132 @@
...
+}
+
+sub test_basic
+{
+    my ($self) = @_;
+
+    my $ts = bless {}, 'Cassandane::Unit::TestCase';
+
+    my $string_based_url = $ts->new_test_url("string based");
+    my $code_based_url = $ts->new_test_url(sub {
```
elliefm, line 26: Is there a reason you're creating `$ts`, and not just using `$self`, which already is a Cassandane::Unit::TestCase subclass, and so inherits `new_test_url`?

Someday when I end up needing this thing myself, I'm going to come looking at these tests as an example of how to drive it, and this is gonna confuse me...

I guess maybe you're trying to isolate the thing under test from the test that's testing it?  Which I think makes some sense, but if so it's tricky and could use a comment.

In practice, with `sub new_test_url` being defined as a method on Cassandane::Unit::TestCase, any test can simply call it like `$self->new_test_url(...)`.  But someone coming here for an example may just copy what's here, not realising it doesn't have to be that hard.  Or, realise it could be simpler, and get stuck second guessing why it's not. ;)

cassandane/Cassandane/Test/NewTestUrl.pm
```diff
@@ -0,0 +1,132 @@
...
+        ],
+    });
+
+    my $lwp = LWP::UserAgent->new;
+
+    {
+        my $res = $lwp->get($string_based_url->url);
+        $self->assert_str_equals('200', $res->code);
+        $self->assert_str_equals('string based', $res->decoded_content);
+    }
```
elliefm, line 40: I like this pattern, of putting each "do a thing, get a result, check it" in its own block, rather than having to reuse $res over and over

cassandane/Cassandane/Unit/TestCase.pm
```diff
@@ -329,4 +330,12 @@ sub assert_num_lte
...
 }
 
+sub new_test_url {
+  my ($self, $content_or_app) = @_;
+
+  return Cassandane::Util::TestURL->new({
+    app => $content_or_app,
+  });
+}
+
```
elliefm, line 340: Four space indents in cyrus ;)

elliefm (commented):
Couple of nits, but this seems fine, modulo the ongoing fork discussion

cassandane/Cassandane/Test/NewTestUrl.pm
wolfsage: Oops. Hold over from when this was in t/.

Fixed!

cassandane/Cassandane/Unit/TestCase.pm
wolfsage: Yeah whoops! I thought I had fixed everywhere but missed this one, thanks

wolfsage: Fixed

cassandane/Cassandane/Unit/TestCase.pm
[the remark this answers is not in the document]
wolfsage: Okay so I've added back in a pattern you were dubious about before, but I think in the end its worth it here (that is, declaring a tiny package within the same file). See https://github.com/cyrusimap/cyrus-imapd/pull/3857/commits/d789a2fc3ed5be300743f3a46d0358e49c30340a

What it does is create an object that calls _exit() when it goes out of scope, which is, as far as I can tell, always guaranteed to happen during the run phase of perl, which means it will happen when this block ends, before any other END/DESTROY block gets to run, effectively preventing any other cleanup code from ever running.

What do you think?

cassandane/Cassandane/Test/NewTestUrl.pm
elliefm: Yeah I love Test::More subtests!  But it's been a while since I've written much perl that wasn't "Cassandane perl", and for good or ill we're wedded to Test::Unit

cassandane/Cassandane/Unit/TestCase.pm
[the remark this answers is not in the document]
elliefm: Oh that's cute.  My pure perl isn't sharp enough to anticipate any edge cases you probably haven't yourself, but this seems like it should play nice with Cassandane.

I think the embedded package thing bothers me a lot less when the whole thing is small enough to fit on screen at once (even in the worst case of a 24-row terminal).  By contrast, Cassandane/Unit/TestPlan.pm has a bunch of multi-hundred-line packages in the same file and it's a nightmare to keep track of what's going on.

I'm gonna pull this branch and build/run it to see what happens, though I don't really expect we'd see weird edge cases turn up until we start using it heavily.  I think the main thing I'll be looking for (assuming it passes) is that I don't see any "ignoring DESTROY from invalid caller" (or whatever it was) in the logs.

cassandane/Cassandane/Test/NewTestUrl.pm
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/perl
+
+package Cassandane::Test::NewTestUrl;
+use strict;
+use warnings;
+
+use JSON;
+use LWP::UserAgent;
+
+use base qw(Cassandane::Unit::TestCase);
```
elliefm, line 10: We don't expect Cassandane modules to be in a place where perl needs to look for them, so you gotta `use lib '.';` before referencing them

```suggestion
use lib '.';
use base qw(Cassandane::Unit::TestCase);
```

cassandane/Cassandane/Util/TestUrl.pm
```diff
@@ -0,0 +1,135 @@
+package Cassandane::Util::TestURL;
+
+use Plack::Loader;
+use Plack::Request;
+use Plack::Response;
+
+use Test::TCP;
+use Carp qw(croak);
+
+use Cassandane::PortManager;
```
elliefm, line 10: Ditto

```suggestion
use lib '.';
use Cassandane::PortManager;
```

cassandane/Cassandane/Unit/TestCase.pm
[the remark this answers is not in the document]
elliefm: Of course I won't see any "ignoring DESTROY...", because the test isn't one that starts up any Instances at all, hah.  Hmm

cassandane/Cassandane/Test/NewTestUrl.pm
```diff
@@ -0,0 +1,130 @@
...
+        $self->assert_str_equals('200', $res->code);
+        $self->assert_str_equals('newval', $res->decoded_content);
+    }
+
+    {
+        eval { $string_based_url->url };
+        my $err = $@;
+        $self->assert(
+            $err =~ /\QCannot call ->url after ->unregister has been called\E/,
+        );
```
elliefm, line 68: Oh you could use `$self->assert_matches(...)` for this, which does the same thing but gets you a good error message if it fails (and the opposite, `$self->assert_does_not_match(...)`)

cassandane/Cassandane/Test/NewTestUrl.pm
wolfsage: Do you need use lib though? These files are pulled in by other cassandane files which are run by testrunner which has already set up the lib paths at that point?

cassandane/Cassandane/Test/NewTestUrl.pm
wolfsage: Ah wasn't aware of that. I stole this pattern from another test file

cassandane/Cassandane/Test/NewTestUrl.pm
elliefm: You do if you want to 'perl -c' all the modules before you run it, which I do, to make sure it isn't gonna trip over compilation failures halfway through (and is how I found it was missing)

cassandane/Cassandane/Test/NewTestUrl.pm
wolfsage: Fixed!

cassandane/Cassandane/Test/NewTestUrl.pm
wolfsage: Okay, I've added that

wolfsage:
Alright, rebased and rewrote history to clean it up and fixed the last remaining things (I believe).

The diff from what was there to this force push:
```
diff --git a/cassandane/Cassandane/Test/NewTestUrl.pm b/cassandane/Cassandane/Test/NewTestUrl.pm
index 6a85b77..f3dcb26 100644
--- a/cassandane/Cassandane/Test/NewTestUrl.pm
+++ b/cassandane/Cassandane/Test/NewTestUrl.pm
@@ -7,6 +7,7 @@ use warnings;
 use JSON;
 use LWP::UserAgent;
 
+use lib '.';
 use base qw(Cassandane::Unit::TestCase);
 
 sub new
@@ -63,16 +64,18 @@ sub test_basic
     {
         eval { $string_based_url->url };
         my $err = $@;
-        $self->assert(
-            $err =~ /\QCannot call ->url after ->unregister has been called\E/,
+        $self->assert_matches(
+            qr/\QCannot call ->url after ->unregister has been called\E/,
+            $err,
         );
     }
 
     {
         eval { $string_based_url->update("foo") };
         my $err = $@;
-        $self->assert(
-            $err =~ /\QCannot call ->update after ->unregister has been called\E/,
+        $self->assert_matches(
+            qr/\QCannot call ->update after ->unregister has been called\E/,
+            $err,
         );
     }
 
diff --git a/cassandane/Cassandane/Util/TestUrl.pm b/cassandane/Cassandane/Util/TestUrl.pm
index 66df5b4..00c953a 100644
--- a/cassandane/Cassandane/Util/TestUrl.pm
+++ b/cassandane/Cassandane/Util/TestUrl.pm
@@ -7,6 +7,7 @@ use Plack::Response;
 use Test::TCP;
 use Carp qw(croak);
 
+use lib '.';
 use Cassandane::PortManager;
 
 sub new


```

I think this is all set? Thanks!

elliefm (approved):
Great, thanks :)

wolfsage:
Awesome, thank you!

PR #3858, opened 2022-01-07
sqldb_rollbackl() fix

ksmurchison:
We had a httpd crasher via assert() in mailbox_open_webdav() -> webdav_abort() -> sqldb_rollback(), presumably because we failed to obtain a writelock and therefore didn't execute a SAVEPOINT command.  This patch bails out of sqldb_rollback() entirely if we don't have a writelock, because we can't have a transaction without one.

### DIFF
--- a/lib/sqldb.c
+++ b/lib/sqldb.c
@@ -483,6 +483,8 @@ EXPORTED int sqldb_commit(sqldb_t *open, const char *name)
 
 EXPORTED int sqldb_rollback(sqldb_t *open, const char *name)
 {
+    if (!open->writelock) return 0;
+
     assert(open->trans.count);
     char *prev = strarray_pop(&open->trans);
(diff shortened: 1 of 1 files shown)


brong:
That sounds reasonable, part of the issue with sqldb stuff is that it's supposed to be a stack, so you can abort part way up the stack.  So there should be NO transactions if there's no writelock.  So we probably want to think through the whole cleanup codepath.

rsto:
@murch is this PR still required for review?

elliefm:
psst, ken's github name is @ksmurchison, not `@murch`

rsto:
@ksmurchison is this PR still required to review?

ksmurchison:
@brong We had another assert() in sqldb_rollback(), this time via CalDAV.  Should we merge this or do some further digging?

rsto (approved):

PR #3859, opened 2022-01-09
Use non-blocking locks for exclusive mailbox actions

brong:
This fixes the lock inversion where cyr_expire triggers a mailbox repack, and then EVERY process tries to repack the mailbox, but if there's a long-running sync_client or a pop3d holding the mailbox namelock open, then everything else just blocks.  This is particularly bad with cyr_expire, because it holds an exclusive namelock while it tries this, which then blocks everything ELSE behind it.

### DIFF
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -1259,7 +1259,8 @@ static void _delayed_cleanup(void *rock)
      * if we are in the middle of a shutdown */
     if (in_shutdown) goto done;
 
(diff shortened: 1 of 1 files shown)


elliefm (approved):

rjbs:
@ksmurchison Please give this a review, and we'll merge it if all good.  (There was some discussion of the state of this today, @brong has asked you to approve.)

brong:
This no longer does anything - was merged as part of https://github.com/cyrusimap/cyrus-imapd/pull/4359

PR #3865, opened 2022-01-12
mboxlist: don't truncate over-long mailbox names

brong:
This fixes a crash where mailbox opens correctly, but the name in the mbentry doesn't match the name in the mailbox then when we close the mailbox, we can't find the listitem.

### DIFF
--- a/imap/mailbox.h
+++ b/imap/mailbox.h
@@ -57,8 +57,9 @@
 #include "seqset.h"
 #include "util.h"
 
-#define MAX_MAILBOX_NAME 490
+#define MAX_MAILBOX_CREATENAME 490
 /* enough space for all possible rewrites and DELETED.* and stuff */
+#define MAX_MAILBOX_NAME 510
 #define MAX_MAILBOX_BUFFER 1024
 #define MAX_MAILBOX_PATH 4096
 
--- a/cassandane/Cassandane/Cyrus/Replication.pm
+++ b/cassandane/Cassandane/Cyrus/Replication.pm
@@ -2139,4 +2139,31 @@ sub test_splitbrain_different_uniqueid_used
     $self->assert_str_equals($mid, $rid);
 }
 
+sub test_delete_longname
+    :AllowMoves :Replication :SyncLog :DelayedDelete :min_version_3_3
+{
+    my ($self) = @_;
+
+    my $mtalk = $self->{master_store}->get_client();
+
+    #define MAX_MAILBOX_NAME 490
+    my $name = "INBOX.this is a really long name 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1.2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2.3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3 3.foo";
+    my ($success) = $mtalk->create($name);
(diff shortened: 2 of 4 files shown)


ksmurchison (approved):

elliefm:
This branch somehow has Dilyan's sendmail documentation updates in it?

elliefm:
And ignoring the sendmail documentation updates, the only changes are a Cassandane test and a #define being renamed -- but no functional changes?

Smells like a rebase gone wrong and the substance of the change is lost!

elliefm (changes requested):
Putting a red flag on this cause I think something has gone wrong, and the actual change isn't in here anymore

docsrc/assets/setup-sendmail.rst
```diff
@@ -1,39 +1,149 @@
...
+Install Sendmail
+################
 
-Next, we run a script that takes the ``/etc/mail/sendmail.mc`` file and and prepares it for use by Sendmail. This may take some time.
+We'll set up LMTP with the Sendmail SMTP server.
 
 ::
 
-    sudo sendmailconfig
+    sudo apt-get install -y sendmail
```
dilyanpalauzov, line 30: This file shall be removed from the PR.

brong:
> And ignoring the sendmail documentation updates, the only changes are a Cassandane test and a #define being renamed -- but no functional changes?
> 
> Smells like a rebase gone wrong and the substance of the change is lost!

No, the #define being renamed is all that's needed!

```
-#define MAX_MAILBOX_NAME 490
+#define MAX_MAILBOX_CREATENAME 490
 /* enough space for all possible rewrites and DELETED.* and stuff */
+#define MAX_MAILBOX_NAME 510
```

Basically this means that all the buffers now have 510 bytes in them, but the checks are still at 490 (the two callsites where user-facing checks are done were updated to use `CREATENAME`)

docsrc/assets/setup-sendmail.rst
brong: Done.

brong:
> Basically this means that all the buffers now have 510 bytes in them, but the checks are still at 490 (the two callsites where user-facing checks are done were updated to use `CREATENAME`)

Oh I should describe why this is enough :)  (and chosen so even the +2 buffers in some places fit inside 512 bytes).

When a mailbox is renamed to a DELETED mailbox, it adds:

DELETED. (8 bytes) after the domain and before the mailbox part; as well as

.12345678 (32 bit epoch timestamp in hex - so 9 bytes) at the end.

So this adds a total of 17 bytes.  You can rename `DELETED` if you fiddle the config, but I figure nobody will do that or if they do they won't make it massively longer!  Maybe I should update the docs as well.

imap/mailbox.h
```diff
@@ -57,8 +57,9 @@
 #include "seqset.h"
 #include "util.h"
 
-#define MAX_MAILBOX_NAME 490
+#define MAX_MAILBOX_CREATENAME 490
 /* enough space for all possible rewrites and DELETED.* and stuff */
+#define MAX_MAILBOX_NAME 510
 #define MAX_MAILBOX_BUFFER 1024
 #define MAX_MAILBOX_PATH 4096
```
elliefm, line 65: The "enough space..." comment used to appear to describe MAX_MAILBOX_BUFFER, but now it appears to describe MAX_MAILBOX_NAME, and there's no longer an apparent explanation of why MAX_MAILBOX_BUFFER is twice the size.

The explanation in your previous comment makes sense, but a comment on a PR will be lost to history almost immediately...

Since these four defines are all very similarly named but different in intent, it'd be a lot clearer to future developers (or forgetful ones, like me) to just have a little comment block above them describing what they're for.

Maybe something like:

```
/* MAX_MAILBOX_CREATENAME - max length of a mailbox name during creation
 * MAX_MAILBOX_NAME - max length of a mailbox name, with space for delayed_delete munging
 * MAX_MAILBOX_BUFFER - buffer size that's large enough for all possible mailbox name operations
 * MAX_MAILBOX_PATH - max length of a mailbox's on disk path
 */
```

(hope I've got all those right, but if I didn't then that means the comment is even more important)

I wonder if we can define MAX_MAILBOX_NAME in terms of `(MAX_MAILBOX_CREATENAME + strlen(config_getstring(IMAPOPT_DELETEDPREFIX)) + expected_length_of_timestamp_suffix)` or something, instead of just hoping nobody configures a longer prefix than we anticipated.  I don't think it would be too expensive, libconfig tries quite hard to be very cheap to look up once loaded.  If we don't do this, we should probably update the deletedprefix entry in imapoptions to say that it shouldn't be longer than _n_.

elliefm (approved):
Somehow when I looked at this last week I thought MAX_MAILBOX_NAME was being _renamed_ to MAX_MAILBOX_CREATENAME, and didn't catch that actually it was still there, but moved and with a new value.

Looks okay now, approving, but I would like to see a comment describing each of the MAX_MAILBOX... defines instead of just a PR comment.  More detail in an inline comment, but I don't remember whether github will show the inline comments first or this "main" one, so you might have already seen the detail before you see this.

PR #3866, opened 2022-01-13
cyr_expire: add config options to control expunging

brong:
1) expire_by_savedate (default true)

If set, check the savedate rather than gmtime on records to decide
whether to expunge them

2) expire_keep_flagged (default false)

If set, never expunge messages with the \Flagged system flag set.

### DIFF
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -969,6 +969,15 @@ Blank lines and lines beginning with ``#'' are ignored.
 /* Notifyd(8) method to use for "EVENT" notifications which are based on
    the RFC 5423.  If not set, "EVENT" notifications are disabled. */
 
+{ "expire_by_savedate", 0, SWITCH, "UNRELEASED" }
+/* If true, calculate age of messages for expiry by the savedate
+   (timestamp when the message was added to this folder) rather
+   than gmtime (date header on the message).  Default is true. */
+
+{ "expire_keep_flagged", 0, SWITCH, "UNRELEASED" }
+/* If true, don't expire messages which have the `\Flagged` IMAP
+   flag set.  Default is false. */
+
 { "expunge_mode", "delayed", ENUM("immediate", "semidelayed", "delayed"), "3.1.1" }
 /* The mode in which messages (and their corresponding cache entries)
    are expunged.  "semidelayed" mode is the old behavior in which the
--- a/changes/next/cyr-expire-options
+++ b/changes/next/cyr-expire-options
@@ -1,0 +1,27 @@
+Description:
+
+Options to control cyr_expire's mailbox cleanup behaviour.
+
+
+Config changes:
+
+Adds two config options:
+
+1) expire_by_savedate (default false)
+
+If set, check the savedate rather than gmtime on records to decide
+whether to expunge them
+
+2) expire_keep_flagged (default false)
+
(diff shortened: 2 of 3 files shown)


lib/imapoptions
```diff
@@ -922,6 +922,15 @@ Blank lines and lines beginning with ``#'' are ignored.
 /* Notifyd(8) method to use for "EVENT" notifications which are based on
    the RFC 5423.  If not set, "EVENT" notifications are disabled. */
 
+{ "expire_by_savedate", 1, SWITCH, "3.7.0" }
```
elliefm, line 925: This needs to be the string "UNRELEASED", not "3.7.0"
```suggestion
{ "expire_by_savedate", 1, SWITCH, "UNRELEASED" }
```

lib/imapoptions
```diff
@@ -922,6 +922,15 @@ Blank lines and lines beginning with ``#'' are ignored.
...
+/* If true, calculate age of messages for expiry by the savedate
+   (timestamp when the message was added to this folder) rather
+   than gmtime (date header on the message).  Default is true. */
+
+{ "expire_keep_flagged", 0, SWITCH, "3.7.0" }
```
elliefm, line 930: Ditto
```suggestion
{ "expire_keep_flagged", 0, SWITCH, "UNRELEASED" }
```

elliefm (changes requested):
Unreleased imapd.conf option changes shouldn't have a version number, they should have "UNRELEASED" -- the release manager (me) will fill them in appropriately when they make it into some actual release

It looks like the default expiry calculation behaviour is changing?  This will need to be remarked in release notes eventually, which means there needs to be a changes/next file for this PR so that that detail doesn't get forgotten over the coming year.

ksmurchison:
@brong do we still want this functionality or should we close this PR?

brong:
@ksmurchison I think we do want this behaviour!  Then we can use cyrus's built in expunging rather than what we do now, which is use SAVEDATE in a external script.

brong:
@elliefm I have fixed imapoptions and made a changes file, and also made the default keep the old behaviour

elliefm (approved):
Looks good, thanks

dilyanpalauzov:
The text in the PR says the defaults are for expire_by_savedate → true, and for expire_keep_flagged → false.  changes/next/cyr-expire-options says for both the default is false.  The lines:
 
 > { "expire_by_savedate", 0, SWITCH, "UNRELEASED" }
 > { "expire_keep_flagged", 0, SWITCH, "UNRELEASED" }
 
suggest that for both options the default is false.  But for the former the documentation states the opposite:
 
 > /* If true, calculate age of messages for expiry by the savedate (timestamp when the message was added to this folder) rather than gmtime (date header on the message).  Default is true. */
 
 I suggest removing the text “The default is … ” in imapoptions for both options, as in most cases there is no need to spell the default values there.  The reason is that the rendered documentation includes the default value anyway, cf. https://www.cyrusimap.org/imap/reference/manpages/configs/imapd.conf.html for allowapop: and allowdeleted:.
 
 Moreover, for binary option I do not think that the spelling “If true" or "If set” is necessary in the documentation.  It is good enough to spell what the option does, it is obvous from the context that the option does that thing only when it is enabled.

Issue #3843, opened 2021-12-27
Missing manual pages


prlw1:
I believe your intention is that if cyrus imapd is built without certain options, e.g., NNTPD, then the manual pages which are dependent on those options, e.g., fetchnews.8, shouldn't be installed.

Thankfully, you also don't expect me, when building from your tarball, to install manual page building dependencies, to generate them from `.rst` files in docsrc.

The inconsistency is that when you `make dist`, without enabling all of the options, those pregenerated manual pages don't make it into the tarball, as the `dist` part of `dist_man8_MANS` would have to be outside of any conditions.

elliefm:
Should be fixed in future releases from 3.2 forward.

The fix doesn't cherry-pick cleanly to 3.0 -- there have been other changes in the meantime.  If someone is affected by this on 3.0 and reopens the issue, I'll be happy to spend a little effort backporting it to fix 3.0 too.  But if that doesn't happen, I'll save my energy.

PR #3867, opened 2022-01-14
Ensure manpages for optional features are always included in release tarball

elliefm:
Turns out (#3843) our release tarballs have not included man pages for fetchnews, nntpd, and some others, because they're for "optional" features, and when making a release we don't turn on or off the optional features, because we're not building any code and the distribution shouldn't depend on them...

This change makes sure that in maintainer mode (used for releases) the man pages for these features are included in the release tarball, regardless of whether the feature has been "enabled" or not.

When an end user extracts the tarball and runs configure themselves (not in maintainer mode), the pre-generated man pages will be present, ready to be installed if the feature they're for has been enabled.

We plan to make most of these optional features non-optional sooner or later, but right now I'm fixing the bug as it exists.

This will be backported to the various older branches too.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1680,18 +1680,18 @@ dist_man8_MANS += \
 	man/httpd.8
 endif # HTTPD
 
-if BENCH
-check_PROGRAMS += bench/cyrdbbench
-bench_cyrdbbench_SOURCES = bench/cyrdbbench.c imap/mutex_fake.c
-bench_cyrdbbench_LDADD = $(LD_BASIC_ADD)
-endif # BENCH
-
 if REPLICATION
 dist_man8_MANS += \
 	man/sync_client.8 \
 	man/sync_reset.8 \
 	man/sync_server.8
-endif
+endif # REPLICATION
+
+if BENCH
+check_PROGRAMS += bench/cyrdbbench
+bench_cyrdbbench_SOURCES = bench/cyrdbbench.c imap/mutex_fake.c
+bench_cyrdbbench_LDADD = $(LD_BASIC_ADD)
+endif # BENCH
 
 master_master_SOURCES = \
 	master/master.c \
--- a/Makefile.am
+++ b/Makefile.am
@@ -1687,6 +1687,19 @@ dist_man8_MANS += \
 	man/sync_server.8
 endif # REPLICATION
 
+if MAINTAINER_MODE
+## make sure feature-dependent man pages are included in distribution
+dist_man8_MANS += \
+	man/ctl_zoneinfo.8 \
+	man/fetchnews.8 \
+	man/httpd.8 \
+	man/nntpd.8 \
+	man/squatter.8 \
+	man/sync_client.8 \
+	man/sync_reset.8 \
+	man/sync_server.8
+endif # MAINTAINER_MODE
+
 if BENCH
 check_PROGRAMS += bench/cyrdbbench
 bench_cyrdbbench_SOURCES = bench/cyrdbbench.c imap/mutex_fake.c


rsto (approved):
LGTM

Makefile.am
```diff
@@ -1680,18 +1680,31 @@ dist_man8_MANS += \
...
-
 if REPLICATION
 dist_man8_MANS += \
 	man/sync_client.8 \
 	man/sync_reset.8 \
 	man/sync_server.8
-endif
+endif # REPLICATION
+
+if MAINTAINER_MODE
```
dilyanpalauzov, line 1690: The `if MAINTAINER_MODE` shall be removed and the files moved to the dist_man8_MANS variable on line 1627.

The conditionals from lines 1678, 1672 and others shall be removed.  As a rule of thumb, it makes no sense to have a `dist_` prefix within Automake conditional.

elliefm: The conditional lines need to be conditional to prevent the man pages being _installed_ for features that haven't been built.

The conditionalised lines could equally be `man8_MANS` (without the `dist_`) in this case, but that would mean more changes beyond what's necessary, and more testing.

Our plan for some point in the future is that none of these features will be optional anymore, and at that point all this conditionalisation will go away, so it's not worth the work right now to tinker with it beyond just fixing the bugs.

dilyanpalauzov:
Calling
```
./configure --with-com_err --enable-cmulocal --enable-idled --with-sqlite --enable-replication --enable-maintainer-mode --enable-calalarmd --enable-xapian --enable-gssapi=/usr/local --with-ldap --enable-jmap
make install DESTDIR=/z
```
prints
```
 /usr/bin/install -c -m 644 man/sync_client.8 man/sync_reset.8 man/sync_server.8 man/ctl_zoneinfo.8 man/fetchnews.8 man/httpd.8 man/nntpd.8 man/squatter.8 man/sync_client.8 man/sync_reset.8 man/sync_server.8 '/z/usr/local/share/man/man8'
/usr/bin/install: will not overwrite just-created '/z/usr/local/share/man/man8/sync_client.8' with 'man/sync_client.8'
/usr/bin/install: will not overwrite just-created '/z/usr/local/share/man/man8/sync_reset.8' with 'man/sync_reset.8'
/usr/bin/install: will not overwrite just-created '/z/usr/local/share/man/man8/sync_server.8' with 'man/sync_server.8'
make[3]: *** [Makefile:6932: install-man8] Error 1
make[3]: Leaving directory '/git/cyrus/cyrus-imapd'
make[2]: *** [Makefile:7431: install-am] Error 2
make[2]: Leaving directory '/git/cyrus/cyrus-imapd'
make[1]: *** [Makefile:7111: install-recursive] Error 1
make[1]: Leaving directory '/git/cyrus/cyrus-imapd'
make: *** [Makefile:7425: install] Error 2
```
With this change, man pages get installed twice with the same `install` invocation, and this fails.

elliefm:
It's because you're passing `--enable-maintainer-mode`.  You should only use that if you're going to `make dist` or `make distcheck`; and if you're going to do that, you shouldn't pass all the other options.

dilyanpalauzov:
This is very unusual for automake packages.  The common patterns is to have `make diat` always work, irrespective of how ./configure was called.

elliefm:
I think the current wisdom is that packages just shouldn't use maintainer mode -- at least, while I haven't looked into it deeply, that's the impression I have come by while learning how it worked.

I'd like to remove it from cyrus-imapd sometime, but it's low priority, will require a bunch of tedious testing to make sure I didn't make it worse, etc, so I keep putting it off...

dilyanpalauzov:
As far as I remember with automake man pages cannot be installed conditionally in a way that is reliable under all circumstances.  Currently, for one year, `./configure --enable-maintainer-mode --more-options && make install DESTDIR=/tmp/BB` fails with
```
 /usr/bin/install -c -m 644 man/imtest.1 man/installsieve.1 man/httptest.1 man/lmtptest.1 man/mupdatetest.1 man/nntptest.1 man/pop3test.
1 man/sieveshell.1 man/sivtest.1 man/smtptest.1 man/synctest.1 man/dav_reconstruct.1 man/dav_reconstruct.1 '/tmp/BB/usr/local/share/man/man1'
/usr/bin/install: will not overwrite just-created '/tmp/BB/usr/local/share/man/man1/dav_reconstruct.1' with 'man/dav_reconstruct.1'
make[3]: *** [Makefile:8212: install-man1] Error 1
make[3]: Leaving directory '/git/cyrus/cyrus-imapd.38/build-gcc-asan'
```
I consider it as more important to have `make install` always completing without error, at the price of installing some man pages for programs (binaries) which are not compiled/installed.

`--enable-maintainer-mode` is a legitimate options, I do not see why shall it cause problems.

elliefm:
> As far as I remember with automake man pages cannot be installed conditionally in a way that is reliable under all circumstances.

Hmm.  Yeah I'd be happy if we just always installed all man pages, even if the tools they document haven't been built/installed.  We already do that with the html documentation anyway.

elliefm:
⇒ #4797

PR #3868, opened 2022-01-14
backend: update backend_version for 3.6 and 3.7

elliefm:
So XFER between versions can recognise the new versions.

This will be cherry-picked to 3.6 and also backported to the older release branches (as far back as 2.4).


rsto (commented):
I do not enough about this corner of Cyrus to make a decent review.

brong (approved):
Yep, that looks great - good plan to backport this everywhere.

PR #3871, opened 2022-01-19
Prepare for Cyrus-SASL 2.1.28

quanah:
Signed-off-by: Quanah Gibson-Mount <quanah@symas.com>

### DIFF
--- a/docsrc/conf.py
+++ b/docsrc/conf.py
@@ -443,7 +443,7 @@ rst_prolog = """
 .. |imap_utopic_stock_version| replace:: 2.4.17+caldav~beta10-5
 .. |imap_vivid_stock_version| replace:: 2.4.17+caldav~beta10-17
 .. |imap_wily_stock_version| replace:: 2.4.17+caldav~beta10-17
-.. |sasl_current_stable_version| replace:: 2.1.27
+.. |sasl_current_stable_version| replace:: 2.1.28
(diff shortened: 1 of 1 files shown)


elliefm (approved):

elliefm:
Do we have an anticipated release date, or just "when it's ready"?

This PR should theoretically be merged around the time it's actually released, so the /dev website update is synchronised... but I'll also need to cherry-pick the commit to the stable branch so the main website gets it too, which introduces a delay, since we're in different timezones and I'm only available three days a week.

So maybe we should plan to merge and cherry-pick it a little earlier than the actual release, and the website will be ahead of reality for a bit?  The alternative is that the release happens, but the main website lags behind a day or two.  I'm not really sure which is preferable, but I guess either is okay with me.  Do you have a preference?

Also, do you get a "Merge pull request" button?

elliefm:
I've backported this to the 2.5, 3.0, 3.2, 3.4, and 3.6 branches.  The website changes should take affect a little after the next hour... will check later and make sure it looks okay.

quanah:
Awesome, thanks @elliefm !

elliefm:
Looks like it's worked okay, https://www.cyrusimap.org now says (near the bottom):

> The latest stable version of Cyrus SASL is 2.1.28.

As do the /2.5, /3.0, /3.2, /3.4, /3.6, and /dev versions of the site.

PR #3872, opened 2022-01-21
Debug sequence bump

ksmurchison:
Added debug syslog()s to see why replies to invites bumps sequence number


ksmurchison:
Issue resolved by #3876

PR #3877, opened 2022-01-24
jmap_ical: preserve LOCATION in CalendarEvent/set{update}

rsto:
Previously, we non-deterministically chose any Location as the
main LOCATION property. Now we attempt to keep the old LOCATION.

If a X-APPLE-STRUCTURED-LOCATION exists for this LOCATION, then
we preserve its values and merge it in the same JSCalendar Location.


ksmurchison (approved):
I haven't compared the new code to the spec, but the structure looks sane

PR #3879, opened 2022-01-25
caldav_alarm.c: ensure alarm record corresponds to current version of event

ksmurchison:
protect from triggering alarms for deleted/replaced event resources


brong (approved):
Yeah, this looks great :)

PR #3881, opened 2022-01-26
jmap_mailbox.c tweaks

ksmurchison:


rsto (approved):
Looks good to me. That size calculation using logical operator was insane!

PR #3884, opened 2022-01-27
imapd.c, user.c: rewrite ACLs when a sharee gets renamed

ksmurchison:


imap/user.c
```diff
@@ -510,21 +508,56 @@ EXPORTED int user_renameacl(const struct namespace *namespace, const char *name,
...
+    return _user_renameacl(rrock->namespace, mbentry,
+                           rrock->olduser, rrock->newuser);
+}
+
+EXPORTED int user_sharee_renameacls(const struct namespace *namespace,
+                                    const char *olduser, const char *newuser)
+{
+    struct rename_rock rrock = { namespace, olduser, newuser, NULL, NULL, 0 };
+
+    return mboxlist_usermboxtree(olduser, NULL, &sharee_rename_cb, &rrock,
```
brong, line 554: Oh wow - yeah, that's basically "do all the things I have an ACL to", tricky!

brong (approved):
This looks great - efficient and useful :)  Reverse ACLs are excellent.

I do wish we had a better API for "stuff shared with me" than `mboxlist_usermboxtree` with all the skips, but it's clean and works correctly.

PR #3886, opened 2022-01-30
fix Carddav.sharing_rename_sharee

elliefm:
This test expects "altnamespace: no", so I've added the :NoAltNamespace magic


rsto (approved):
That explains it!

PR #3887, opened 2022-01-31
Remove support for DAV:displayname on JMAP Mailboxes

ksmurchison:


rsto (approved):
Looks good to me. There is the slight risk that some mailboxes were created during the time where Cyrus allowed to duplicate names, which would have IMAP mailbox names like mailbox and mailbox1. If keeping the read part of the displayname annotation is worth it to prevent that, I don't know.

ksmurchison:
If we keep the read part, what do we do for a mailbox rename?  That's what brought this to our attention.  If we can solve the mailbox rename issue, we can keep the annotation, but so far trying to delete/rename the annotation has been causing other issues

rsto:
Right, so in that case I'd say it's not worth the effort to keep that annotation. I don't even think we ever released a stable Cyrus that included the support for duplicate names.

ksmurchison:
If we never released it then I think we're safe with this.

rsto:
Agreed

PR #3888, opened 2022-01-31
Add missing requirements from CalendarEvent/set specification

rsto:
The latest CalendarEvent/set specification includes a couple of requirements that are not met by the current Cyrus implementation.

The latest specification is available here: https://github.com/jmapio/jmap/blob/master/spec/calendars/event.mdown

This PR implements these missing requirements, where each commit implements a single requirement (except for the JNULL and variable scope redeclarations).


rsto:
Rebased to latest jmap-calendars-01 branch

ksmurchison (approved):
I didn't verify that the technical changes match the current draft, but the code looks sane.

imap/http_caldav.c
```diff
@@ -2270,7 +2270,7 @@ static int list_calendars(struct transaction_t *txn)
     buf_printf(&txn->buf, "%s://%s%s", proto, host, txn->req_tgt.path);
 
     memset(&lrock, 0, sizeof(struct list_cal_rock));
-    lrock.scheddefault = caldav_scheddefault(httpd_userid);
+    lrock.scheddefault = caldav_scheddefault(httpd_userid, 0);
```
dilyanpalauzov, line 2273: Why is this `0` and not `1`?  My understanding is that even if the user does not set a “Default Calendar Collection”, Cyrus IMAP does provide one and that one cannot be deleted and is highlighted.

rsto: I agree with you, this is a regression. Will you provide a pull request?

dilyanpalauzov: > Will you provide a pull request?

No.

dilyanpalauzov: How about here: https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/http_caldav.c#L1059 ?

PR #3889, opened 2022-02-01
JMAP Calendars: workaround stale dav.db data and fix empty links property

rsto:
All regression tests pass correctly.


ksmurchison (approved):
LGTM

PR #3890, opened 2022-02-01
jmap_calendar: set auto-expire on JMAP notifications mailbox

rsto:
This sets the Cyrus-internal "expire" mailbox annotation on
newly created #jmapnotifications mailboxes. The default is
to expire after 30 days, but this value may be overriden
with the jmap_notifications_expire config option.


elliefm:
I'm not sure I like this approach -- it's already pretty hard to keep track of what `cyr_expire` will actually do any given time it runs, and adding automatic expire annotations and an imapd.conf option adds to that complexity.

I don't really have any better ideas though.

To the extent that anyone even knows expire annotations exist, I think they're expected to be managed by the admin (or by some external automation the admin creates).  What happens if an admin uses setmetadata to change the annotation that cyrus set, will cyrus reject it?  Quietly change it back later?  Will we (cyrus implementors) or cyrus be confused by it having been changed to something other than the configured default?

rsto:
These are all good points, but as you said, we didn't find a better solution, yet.

The previous attempt was to expire notifications when a JMAP client dismisses a notification as read. This won't cut it for installations that have Cyrus built with JMAP, but only use CalDAV for calendaring.

An alternative that could work is to expire notifications when a new notification is created, be that through JMAP or CalDAV. This will incur some latency cost, but it should be negligible for the majority of requests.

rsto:
@ksmurchison I converted this PR to draft until we are 100% sure how we want to expire notifications.

elliefm:
One of the things cyr_expire does -- I can't remember which, but it might be the "move message to archive partition after 7 days" action -- is just hard coded, always happens on the back of one of the other actions, there's no configuration for the "7 days", etc.  I don't like that either... but I think something like it would be less confusing than a thing that automatically creates expire annotations, which then has it's own configuration elsewhere, and relies on the expire annotation behaviour that might be not well known.

It looks like the #jmapnotifications mailboxes are MBTYPE_JMAPNOTIFY?  Perhaps you could add a new action to cyr_expire that specifically targets MBTYPE_JMAPNOTIFY mailboxes, similar to the other cyr_expire actions.  An admin would need to add the appropriate switch to their cyr_expire event (or run it again as a separate event), but at least this is what they're already used to.

ksmurchison:
@rsto can you rebase this?  I'm doubting that this is really 153 commits

rsto:
Will do. But since this PR is a draft it will needs some further code changes anyway.

rsto:
@rsto what's the state of jmap_expire now?

PR #3892, opened 2022-02-02
Squatter: Don't crash on assert() in lib/cyrusdb_twoskip.c:fetch()->assert(keylen)

mosvald:
This checks return value from mboxname_from_external() and if it is NULL
or empty string, just inform about the invalid mailbox name instead of
crashing.

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1911689


brong (approved):
Good catch - yes, that's reasonable.  Maybe twoskip fetch should just return NOTFOUND rather than aborting on `fetch("")`  - though running mboxtree on that will do EVERYTHING, so this patch is good anyway.

PR #3894, opened 2022-02-02
VALARM absolute trigger fix

ksmurchison:
calalarmd would spin on a non-terminating recurring event  having an alarm with an absolute trigger, because we would never hit the exit clause (an alarm more than 2 months in the future)

E.g.:

BEGIN:VTODO
CREATED:20211011T104314Z
DTSTAMP:20220201T203307Z
DTSTART;TZID=America/Toronto:20220202T113000
DUE;TZID=America/Toronto:20220202T113000
LAST-MODIFIED:20220201T203257Z
RRULE:FREQ=DAILY
STATUS:NEEDS-ACTION
SUMMARY:D3 pill
UID:8F8917BD-0DD4-4B7B-9245-AF8BBC42E493
X-APPLE-SORT-ORDER:3
BEGIN:VALARM
ACTION:DISPLAY
DESCRIPTION:Reminder
TRIGGER;VALUE=DATE-TIME:20220202T163000Z
UID:C9C19C74-EA8E-4A22-8BD2-C68C970644E0
X-WR-ALARMUID:C9C19C74-EA8E-4A22-8BD2-C68C970644E0
END:VALARM
END:VTODO

### DIFF
--- a/cassandane/Cassandane/Cyrus/CaldavAlarm.pm
+++ b/cassandane/Cassandane/Cyrus/CaldavAlarm.pm
@@ -235,6 +235,77 @@ EOF
     $self->assert_alarms({summary => 'Simple', start => $start, timezone => 'Australia/Sydney'});
 }
 
+sub test_recurring_absolute_trigger
+    :min_version_3_7 :needs_component_calalarmd
+{
+    my ($self) = @_;
+
+    my $CalDAV = $self->{caldav};
+
+    my $CalendarId = $CalDAV->NewCalendar({name => 'foo'});
+    $self->assert_not_null($CalendarId);
+
+    my $now = DateTime->now();
+    $now->set_time_zone('Australia/Sydney');
+    # bump everything forward so a slow run (say: valgrind)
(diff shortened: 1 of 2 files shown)


ksmurchison:
I just occurred to me that if Apple continues to create these, and the user expects to get an alarm every day, perhaps we should bump the trigger time by the difference between DTSTART and the start time of the recurrence rather than just triggering once.

brong (approved):
This looks good - thanks!  Definitely looks like this stuff is happening in the wild already.

rsto (approved):
LGTM

rsto:
> I just occurred to me that if Apple continues to create these, and the user expects to get an alarm every day, perhaps we should bump the trigger time by the difference between DTSTART and the start time of the recurrence rather than just triggering once.

I don't think we should do that, at least not before we know why Apple chose to use an absolute time trigger. My guess is that they just set this VALARM because they want to keep track of its UID across Apple devices. The alarms are then generated on the device. But this is all just guesswork for now.

PR #3899, opened 2022-02-04
caldav_db: resolve duplicate recurrence ids conflicts

rsto:
iCalendar allows multiple calendar objects with the same
recurrence id embedded within a VCALENDAR. But for JMAP
Calendars these duplicates are not allowed and cause the
jscal_objs table to raise a conflict.

Resolve such write conflicts by ignoring all but the first
recurrence id for all duplicate entries.

In order to gracefully handle any existing database entries,
use UPSERT rather than INSERT or UPDATE.


ksmurchison (approved):
LGTM

PR #3905, opened 2022-02-07
Revert "jmap_calendar: add MBTYPE_JMAPNOTIFICATION"

rsto:
It only was committed on the main branch for integration testing.

This reverts commit 92a22afbe7073ba0481e07a5b4f758a021e82a3f.

### DIFF
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -225,9 +225,6 @@ EXPORTED const char *mboxlist_mbtype_to_string(uint32_t mbtype)
     case MBTYPE_ADDRESSBOOK:
         buf_putc(&buf, 'a');
         break;
-    case MBTYPE_JMAPNOTIFY:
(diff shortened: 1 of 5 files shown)


elliefm (approved):
Thanks!

elliefm:
It's merged now, can you delete the branch so the name doesn't get in the way later please? (I don't think it will, but it might)

rsto:
I can't delete the branch because it is protected. Most likely it preserved the protected status when I forked it off the 3.6 branch. I don't have the admin rights on the repo to unprotect branches.

elliefm:
Ahh, I bet it's protected because it matches the pattern cyrus-imapd-*!  I'll just delete it myself, just didn't want to do so out from under you.

I'm not sure why you don't have admin rights, we should probably just give them to you.

PR #3906, opened 2022-02-07
caldav_sched: preserve PARTSTAT for rescheduled iTIP messages

rsto:
Before, every change to a scheduled iCalendar component replaced the
PARTSTAT of existing ATTENDEEs to be reset to NEEDS-ACTION. This
patch changes that to only set NEEDS-ACTION if the ATTENDEE either
has no PARTSTAT parameter set.

### DIFF
--- a/cassandane/Cassandane/Cyrus/JMAPCalendars.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPCalendars.pm
@@ -17902,4 +17902,114 @@ EOF
 }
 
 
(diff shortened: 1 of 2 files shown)


elliefm:
The `JMAPCalendars.calendarevent_guesstz` failure was due to my cyruslibs being out date.  I've updated and it's working now.  `JMAPCalendars.rscale_in_jmap_hidden_in_caldav` still fails.

PR #3910, opened 2022-02-09
JMAPCalendars: fix rscale_in_jmap_hidden_in_caldav

elliefm:
Looks like the recurrenceRule->recurrenceRules JMAP rename was mistakenly applied to a CalDAV query as well, causing this test to fail


rsto (approved):
OK, let's do that. Still, CalDAVTalk needs an update.

PR #3913, opened 2022-02-14
caldav_db.c: initialize upsert ptrarray early

ksmurchison:
This fixes a crasher if CMD_SELJSCALOBJS fails and we goto done early

PR #3914, opened 2022-02-14
caldav_writeical_jmap: Don't attempt to free uninitialized memory

wolfsage:
Initialization should happen before any gotos, otherwise we crash.


wolfsage:
(Have no compiled/tested)

wolfsage:
Dupe of https://github.com/cyrusimap/cyrus-imapd/pull/3913!

PR #3920, opened 2022-02-15
processimip: assume that the script owner is indeed a recipient

ksmurchison:
Until we have a good mechanism for specifying/querying a user's scheduling addresses, assume that the script owner is the intended recipient


elliefm:
I've cherry-picked this to the cyrus-imapd-3.6 branch

PR #3921, opened 2022-02-16
properly sanity check DTEND/DURATION

ksmurchison:
Recent changes to libical now have icalcomponent_get_dtend() return the value of DTSTART if neither DTEND nor DURATION is present in a VEVENT.  This now causes Cyrus to reject such VEVENTs with "DTEND must occur after DTSTART" errors.

This PR avoids using icalcomponent_get_dtend() and checks each property individually.

### DIFF
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -3635,7 +3635,7 @@ static int caldav_patch(struct transaction_t *txn, void *obj)
 }
 
 
-static int validate_end_props(icalcomponent *comp, struct error_t *error)
+static int validate_dtend_duration(icalcomponent *comp, struct error_t *error)
 {
     icalproperty *prop;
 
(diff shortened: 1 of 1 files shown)


rsto (commented):
Just a nit.

imap/http_caldav.c
```diff
@@ -3635,6 +3635,45 @@ static int caldav_patch(struct transaction_t *txn, void *obj)
...
+        }
+        if (icaltime_compare(dtend, dtstart) < 1) {
+            error->desc = "DTEND must occur after DTSTART";
+            error->precond = CALDAV_VALID_DATA;
+            return HTTP_FORBIDDEN;
+        }
+    }
+    else {
+        /* Make sure DURATION > 0 */
+        prop = icalcomponent_get_first_property(comp, ICAL_DTEND_PROPERTY);
```
dilyanpalauzov, line 3662: May be ICAL_DURATION_PROPERTY?

PR #3922, opened 2022-02-16
http_caldav_sched: percent-decode iMIP notification recipient

rsto:
The 'recipient' field in the notification for outgoing iMIP messages
requires an email address as value. But Cyrus copied the mailto:
encoded URI of the addressee in that field. This breaks for email
addresses containing a '+' character in the localpart.


imap/http_caldav_sched.c
```diff
@@ -594,14 +594,20 @@ static int imip_send(const char *userid, struct sched_data *sched_data,
                      const char *sender, const char *recipient)
 {
     const char *notifier = config_getstring(IMAPOPT_IMIPNOTIFIER);
+    int r = 0;
+
+    struct buf myrecipient = BUF_INITIALIZER;
+    charset_decode_percent(&myrecipient, recipient);
```
brong, line 600: I prefer not to reuse the same variable name for data that has different levels of decoding, so if we called it `const char *rawrecipient` in the signature and then have a `const char *recipient` for the rest.

brong (approved):

imap/http_caldav_sched.c
rsto: thanks, fixed

PR #3924, opened 2022-02-17
jmap_ical: do not implicitly reset libical iterator

rsto:
Reading the component-scoped sequence timestamp inside the
ATTENDEE loop messes up the libical-internal iterator.

### DIFF
--- a/imap/jmap_ical.c
+++ b/imap/jmap_ical.c
@@ -2497,7 +2497,29 @@ participants_from_ical(icalcomponent *comp, json_t *linksbyparticipant)
         free(uri);
     }
 
+    /* Read scheduling fields */
+    char *schedule_updated = NULL;
+    icaltimetype icaldt = icalcomponent_get_dtstamp(comp);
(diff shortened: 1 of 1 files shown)


imap/jmap_ical.c
```diff
@@ -2561,8 +2569,10 @@ participants_from_ical(icalcomponent *comp, json_t *linksbyparticipant)
         json_decref(participants);
         participants = json_null();
     }
+
     free_hash_table(&attendee_by_uri, NULL);
     free_hash_table(&id_by_uri, free);
+    free(schedule_updated);
```
ksmurchison, line 2575: Do we need to free schedule_comment as well?

ksmurchison (changes requested):

imap/jmap_ical.c
rsto: right. fixed.

ksmurchison (approved):

PR #3926, opened 2022-02-17
ContactGroup/get fix for v4 vCards

ksmurchison:
A user has a vCard containing the following and we were showing an empty contact list because we were only looking for the Apple-specific group properties and not the formalized v4 ones:

KIND:group
MEMBER:urn:uuid:60f60d95-1f33-480c-bfd6-02b93a07aefc
MEMBER:urn:uuid:3e7cfbaf-3199-41bd-8749-38b8d1c89605


rsto (approved):
LGTM

PR #3930, opened 2022-02-21
jmap_calendar.c: explicitly initialize struct createevent

ksmurchison:
We had a CalendarEvent/set {create} fail with error=<NOT NULL constraint failed: ical_objs.resource>.  This probably won't fix that, but I think its a little cleaner.

### DIFF
--- a/imap/jmap_calendar.c
+++ b/imap/jmap_calendar.c
@@ -4497,13 +4497,14 @@ static void setcalendarevents_create(jmap_req_t *req,
                                      json_t **errptr)
 {
     struct jmap_parser parser = JMAP_PARSER_INITIALIZER;
+    struct createevent create;
     int r;
 
-    struct createevent create = {
-        .jsevent = jsevent,
-        .db = db,
-        .serverset = serverset
-    };
+    memset(&create, 0, sizeof(struct createevent));
+
+    create.jsevent = jsevent;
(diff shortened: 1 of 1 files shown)


imap/jmap_calendar.c
```diff
@@ -4497,13 +4497,14 @@ static void setcalendarevents_create(jmap_req_t *req,
                                      json_t **errptr)
 {
     struct jmap_parser parser = JMAP_PARSER_INITIALIZER;
+    struct createevent create;
```
elliefm, line 4500: You can just do `struct createevent create = {0};` here, and avoid needing the `memset()` syscall, if you don't want to define a CREATEEVENT_INITIALIZER for it.

On the other hand, such an initialiser could also just be `#define CREATEEVENT_INITIALIZER {0}`, so it wouldn't be a nuisance to create or maintain

rsto: I agree with elliefm that the `memset` is my least preferred option. However, I do not see a reason for this pull request: what makes initializing the variable with `{0}` (or memset), then setting some of its fields in a separate statements clearer? The original assignment using designated initializer gives us both, and we have that feature since C99.

ksmurchison:
Closing this because its unnecessary and does fix anything.

PR #3931, opened 2022-02-21
caldav_util.c: reset errno before storing resource

ksmurchison:


brong (approved):
This seems obviously fine.  It would be nice for either the commit message or the PR description to explain how this was found and what issue it fixes!

PR #3932, opened 2022-02-22
Sieve :matches fix

ksmurchison:
A script like 

if header :matches "subject" `"\\\\foo"` {

was NOT matching 

Subject: \\foo

The reason was that we were skipping '\' regardless of the following character.

Strings in :matches only need to escape '?' and '*' (match explicity against themselves).

If correct, this fix should be backported.

### DIFF
--- a/sieve/comparator.c
+++ b/sieve/comparator.c
@@ -297,7 +297,10 @@ static int octet_matches_(const char *text, size_t tlen,
             set_var(var_num, val_start, t, match_vars);
             break;
         case '\\':
-            c = *p++;
+            if (*p == '?' || *p == '*') {
+                /* escaped wildcard character */
+                c = *p++;
+            }
             /* falls through */
         default:
             if ((c == *t) || (casemap && (toupper(c) == toupper(*t)))) {


sieve/comparator.c
```diff
@@ -296,7 +296,10 @@ static int octet_matches_(const char *text, size_t tlen,
             set_var(var_num, val_start, t, match_vars);
             break;
         case '\\':
-            c = *p++;
+            if (*p == '?' || *p == '*') {
+                /* escaped wildcard character */
+                c = *p++;
+            }
```
kjetilho, line 302: I do not think this is correct.  In a match pattern, `\` is used to quote any following character, and that is what happens in the original code: the `\` in c is replaced by the next character in the pattern, and then it falls through to process it as a verbatim character.

I think there is a small bug here in the original code, though: If the `\` is last in the pattern string (e.g., `*foo\`, c becomes 0, and it will enter the `else` for "literal char does not match" in the default block.  Behaviour is undefined here, ie. the pattern is invalid and probably should throw an error during upload.  At least it is good it never returns a successful match.

kjetilho: Actually - after reading the RFC again, I still think the patch is wrong, but also the existing code.

It can be argued that in a match context *only* `?` and `*` can be escaped.  In other words, `"\\a"` is `\a` and also `"\\\\"` is `\\`.  `"\\?"` is however just `?`.

So - the code should *rewind* the pointer when the character following `\` is not a wildcard.

```c
    } else {
        c = *--p;
    }
```

PR #3936, opened 2022-02-24
conversations: zero out cids list if no match for message-id

brong:
This was found by a user reporting that messages were being merged
despite having a different subject.  It was because the second one
had an x-me-message-id, and the lookup didn't find it - but the
`cids` list still had the value from the earlier references lookup,
which meant that `i == 3` was true but the message-id was wrong.


brong:
@ksmurchison adding you as a reviewer as well :)  This is a pretty short and obvious fix thankfully.

ksmurchison (approved):
Makes sense to me

PR #3939, opened 2022-02-28
jmap_mail: supported extended filenames for calendar attachments

rsto:
Cyrus already determined an email attachment to be a calendar
attachment if its name ends with the .ICS suffix. But names
encoded as extended MIME parameters were not supported.

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -6800,6 +6800,47 @@ static int _email_get_headers(jmap_req_t *req __attribute__((unused)),
     return r;
 }
 
+
+static void emailbodypart_read_name(struct buf *buf, const struct body *part)
+{
+    const char *fname = NULL;
+    int is_extended = 0;
+
+    /* Lookup name parameter. Disposition header has precedence */
+    const struct param *param;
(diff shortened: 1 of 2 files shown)


imap/jmap_mail.c
```diff
@@ -7304,18 +7315,13 @@ static int _email_get_bodies(jmap_req_t *req,
...
-                        if (len > 4 && !strcasecmp(param->value + len-4, ".ICS")) {
-                            is_icsbody = 1;
-                            break;
-                        }
-                    }
-                    param = param->next;
+                struct buf buf = BUF_INITIALIZER;
+                emailbodypart_read_name(&buf, part);
+                if (buf.len > 4 &&
+                        !strcasecmp(buf_cstring(&buf) + buf.len-4, ".ICS")) {
```
ksmurchison, line 7321: I know we use buf.len in a lot of places in the code, but should we standardize on using buf_len()?

ksmurchison (commented):
LGTM other than my comment

ksmurchison (approved):

PR #3943, opened 2022-03-01
JMAP Calendars: only send iTIP for new or updated instances

rsto:
Before, the iTIP notification contained all standalone instances having that iCalendar UID.

The PR also includes a code-hygiene commit of the CalendarEvent/set{update} code.


rsto:
Reverted to draft, I want to check one more thing.

ksmurchison (approved):
LGTM

PR #3944, opened 2022-03-01
Inherit attendee user data from master when creating iTIP override

ksmurchison:
When the organizer sent an iTIP request with a new overridden instance the attendee's personal data (alarms, transparency) wasn't being inherited from the master component


ksmurchison:
@rsto I will add a check in the imip_override test

rsto (approved):

PR #3945, opened 2022-03-02
jmap_ical: fix conditional using unrelated global variable

rsto:
The conditional erroneously checked the `timezone` variable, which
is defined in the libc time.h header and unrelated to the actual
control logic.


ksmurchison (approved):
Clearly a correct fix.

PR #3946, opened 2022-03-02
JMAP Calendars: ignore scheduling inbox when looking up jscal objects

rsto:


ksmurchison (changes requested):
The first 2 commits looks good.  I think the last two leaked in from another PR which has since been merged.

ksmurchison (approved):

PR #3949, opened 2022-03-02
Another iMIP override test

ksmurchison:
This tests an issue previously found by Neil which appears to have been fixed by #3942 and #3944


cassandane/Cassandane/Cyrus/Sieve.pm
```diff
@@ -4358,6 +4358,163 @@ EOF
     $self->assert_str_equals('DISPLAY', $ical->{entries}[1]{entries}[0]{properties}{action}[0]{value});
 }
 
+sub test_imip_override2
```
rsto, line 4361: could rename that to something more specific? From what I see in the test, it's something like `test_imip_preserve_alerts` ?

rsto (changes requested):
Test logic looks good to me, but please make the test name more clear.

cassandane/Cassandane/Cyrus/Sieve.pm
ksmurchison: Done

rsto (commented):
Sorry to be that guy, but could you please fixup the second commit on the first?

rsto (approved):

PR #3950, opened 2022-03-03
proof of concept: don't allow "timezone" as a name

elliefm:
libc's time.h gives us this:

```
#if defined __USE_MISC || defined __USE_XOPEN
extern int daylight;
extern long int timezone;
#endif
```

We don't use it, but since we always include time.h (via config.h), we always have this `timezone` name kicking around in the global scope.  We do a lot of work with timezones, and problems with names or scoping can slip beneath the radar when this global exists as a fallback -- e.g. the bug fixed by #3945, which the compiler couldn't detect because `timezone` was valid (but irrelevant).

This fix defines `timezone` to be a comment, so that it's no longer a usable name.  There were a few places where we did try to use timezone as a local variable name; these have been renamed to `tz`.  If some source file ever needs to use the libc global (seems unlikely), it can access it by `#define REALLY_USE_LIBC_TIMEZONE` before including config.h.

With this fix, new code that tries to use `timezone` as a variable name will fail to compile, a bit like this:

```
imap/caldav_alarm.c: In function ‘send_alarm’:
imap/caldav_alarm.c:387:26: error: expected identifier or ‘(’ before ‘=’ token
     const char *timezone = NULL;
                          ^
```

I would really like a more helpful message here, but I haven't yet come up with a clever way to provoke one.  I am very open to suggestions!

What do we think of this?

An alternate possibility might be to `#undef __USE_MISC` or `#undef __USE_XOPEN` before including time.h.  That way time.h wouldn't give us the global `timezone`, and the problem would go away.  But I'm not sure what else we would lose.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2322,6 +2322,15 @@ AH_BOTTOM([
 # endif
 #endif
 
+/* time.h declares a `timezone` global which causes stupid problems
+ * when we accidentally use it as a local name too, but mess up the
+ * scoping.  So, make sure we can not use it as a local name too,
+ * except very carefully!
+ */
+#ifndef REALLY_USE_LIBC_TIMEZONE
+#define timezone /* dont call things this! */
+#endif
+
 /* com_err.h, as needed */
 #ifndef IN_COM_ERR
 #ifdef HAVE_ET_COM_ERR_H
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -384,17 +384,17 @@ static int send_alarm(struct get_alarm_rock *rock,
     FILL_STRING_PARAM(event, EVENT_CALENDAR_ORGANIZER,
                       xstrdup(prop ? icalproperty_get_value_as_string(prop) : ""));
 
-    const char *timezone = NULL;
+    const char *tz = NULL;
     if (!icaltime_is_date(start) && icaltime_is_utc(start))
-        timezone = "UTC";
+        tz = "UTC";
     else if (icaltime_get_timezone(start))
-        timezone = icaltime_get_location_tzid(start);
+        tz = icaltime_get_location_tzid(start);
     else if (rock->floatingtz)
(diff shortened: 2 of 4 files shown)


rsto (dismissed):
I like this a lot!

configure.ac
```diff
@@ -2322,6 +2322,15 @@ AH_BOTTOM([
 # endif
 #endif
 
+/* time.h declares a `timezone` global which causes stupid problems
+ * when we accidentally use it as a local name too, but mess up the
+ * scoping.  So, make sure we can not use it as a local name too,
+ * except very carefully!
+ */
+#ifndef REALLY_USE_LIBC_TIMEZONE
+#define timezone /* dont call things this! */
```
brong, line 2331: You can even do this:

```
#define timezone "Don't use timezone as a variable name"
```



Which leads to:
```
In file included from imap/jmap_ical.c:44:
imap/jmap_ical.c: In function ‘location_from_ical’:
./config.h:781:18: error: expected identifier or ‘(’ before string constant
  781 | #define timezone "Don't use timezone as a variable name"
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./config.h:781:18: note: in definition of macro ‘timezone’
  781 | #define timezone "Don't use timezone as a variable name"
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

brong (dismissed):
This looks great - either the comment or string variation.  Nice idea.

configure.ac
elliefm: oh that's amazing, thank you

elliefm: Hmm.  I tried that out by undoing the change in #3945, and it wouldn't have caught that case! :(

The problem there was:

```
     if (timezone /* whoops, wrong name! */) {
```

and it slides by unremarked because a string constant is true in a boolean context.

PR #3952, opened 2022-03-04
Adapt annotation parsing regexps to version 3.2

lmctv:
The enclosed commit unbreaks cyradm's info command when talking to version 3.2 servers, which omit the quotes around the mailbox name, and can also send unquoted ```NIL``` as can be seen in the following imtest excerpts:

```l00 getannotation user.test-user-32 "*" "value.shared"
* ANNOTATION user.test-user-32 "/vendor/cmu/cyrus-imapd/squat" ("value.shared" NIL)
* ANNOTATION user.test-user-32 "/vendor/cmu/cyrus-imapd/size" ("value.shared" "6007")
```
```l00 getannotation user.test-user-24 "*" "value.shared"
* ANNOTATION "user.test-user-24" "/vendor/cmu/cyrus-imapd/squat" ("value.shared" "true")
* ANNOTATION "user.test-user-24" "/vendor/cmu/cyrus-imapd/size" ("value.shared" "7006")
```
Sending it upstream, since I think it can help someone else's version migration.

### DIFF
--- a/perl/imap/IMAP/Admin.pm
+++ b/perl/imap/IMAP/Admin.pm
@@ -724,7 +724,7 @@ sub getinfo {
 			# this is the only possible response.
 
 		        if ($text =~
-			       /^\s*\"([^\"]*)\"\s+\"([^\"]*)\"\s+\(\"([^\"]*)\"\s+\"([^\"]*)\"\)/) {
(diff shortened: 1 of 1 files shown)


elliefm (approved):
Looks good, thanks!

elliefm:
I've had a look at the 2.5 and 3.0 branches, and it doesn't look like it would be useful to port forward.  Looks like a similar-but-different set of changes had already been applied for those branches.

PR #3955, opened 2022-03-07
jmap_calendar: pick new default calendar if current one is destroyed

rsto:
Before, CalendarEvent/set{destroy} rejected deleting the default
calendar with a 'forbidden' set error. Now it allows to delete
the default calendar, and will attempt to pick a new default
calendar to be set for CalDAV.


ksmurchison (approved):
LGTM

PR #3957, opened 2022-03-08
jmap_ical: set Participant schedule properties for invitee iTIP

rsto:
Before, the Participant scheduleComment et al were only set for
iTIP REPLY. Now they also are set for COUNTER and REFRESH.


ksmurchison (approved):
Makes sense

PR #3961, opened 2022-03-09
Properly count how many components are canceled via iTIP

ksmurchison:
This fixes a bug where `processimip :deletecanceled` would delete the entire resource when just a single instance was canceled


brong (approved):

rsto (approved):
LGTM

PR #3962, opened 2022-03-09
jmap_calendar: share full recurrence set for private events

rsto:
See https://www.rfc-editor.org/errata/eid6872


ksmurchison (approved):
Nice and simple

brong (approved):
This looks like what I saw agreed upon via slack :)

PR #3963, opened 2022-03-09
jmap_ical: only server-set "updated" property is source of event

rsto:
JMAP spec now says: "If (and only if) the server is the source of the
event (see Section XXX), the “updated” property MUST be set to the
current time by the server whenever an event is created or updated."


brong (approved):
This looks good.

PR #3964, opened 2022-03-09
JMAP scheduled send

ksmurchison:
Title describes it all.  @brong @rsto documentation for this is in MR #10032 in FM GitLab


imap/jmap_mail.c
```diff
@@ -11153,7 +11169,14 @@ static int _email_bulkupdate_plan_mailboxids(struct email_bulkupdate *bulk, ptra
...
                     /* Delete message from mailbox */
                     ptrarray_append(&plan->delete, uidrec);
-                    plan->needrights |= JACL_REMOVEITEMS;
+
+                    /* Don't require rights to remove items from the
+                       normally read-only $scheduled mailbox
+                       since we have already verified that the emails
+                       in this plan are indeed scheduled for sending */
+                    if (strcmpnull(scheduled_uniqueid, mboxrec->mbox_id)) {
+                        plan->needrights |= JACL_REMOVEITEMS;
```
rsto, line 11194: The comment would warrant this to be written as

    if (!strcmpnull(scheduled_uniqueid, mbox_id))
        plan->needrights &= ~JACL_REMOVEITEMS;

imap/jmap_mail.c
```diff
@@ -11233,7 +11256,13 @@ static int _email_bulkupdate_plan_mailboxids(struct email_bulkupdate *bulk, ptra
                 ptrarray_append(&plan->copy, pick_uidrecs);
             }
             ptrarray_append(pick_uidrecs, pick_uidrec);
-            plan->needrights |= JACL_ADDITEMS;
+
+            /* Don't require rights to add items to the normally
+               read-only $scheduled mailbox since we have already verified
+               that the emails in this plan are indeed scheduled for sending */
+            if (strcmpnull(scheduled_uniqueid, dst_mbox_id)) {
+                plan->needrights |= JACL_ADDITEMS;
```
rsto, line 11280: same as above

imap/mboxlist.h
```diff
@@ -196,6 +196,8 @@ int mboxlist_createmailboxcheck(const char *name, int mbtype,
 #define MBOXLIST_CREATE_SYNC                (1<<4)
 /* don't promote intermediary into a real mailbox */
 #define MBOXLIST_CREATE_KEEP_INTERMEDIARIES (1<<5)
+/* create mailbox as read only by everyone */
+#define MBOXLIST_CREATE_READ_ONLY           (1<<6)
```
rsto, line 200: nit: the flag `MBOXLIST_CREATE_READ_ONLY`only is introduced here, but the previous two commits already use it. that would Cyrus leave in non-compilable state. Better to merge these three commits, as they really are one thing?

imap/caldav_alarm.c
```diff
@@ -1271,12 +1271,79 @@ static int move_to_mailboxid(struct mailbox *srcmbox,
...
+    char *mboxname;
+    uint32_t uid;
+};
+
+static int find_sched_cb(const conv_guidrec_t *rec, void *rock)
+{
+    struct find_sched_rock *frock = (struct find_sched_rock *) rock;
+    struct buf attrib = BUF_INITIALIZER;
+    mbentry_t *mbentry = NULL;
+
```
rsto, line 1285: guid_foreach will also call the callback for deleted and expunged messages, so better the check the rec's system flags here?

imap/caldav_alarm.c
```diff
@@ -1359,6 +1426,51 @@ static int process_futurerelease(struct mailbox *mailbox,
 
     caldav_alarm_delete_record(mailbox_name(mailbox), record->uid);
 
+    onSend = json_object_get(submission, "onSend");
+    if (onSend) {
+        json_t *emailid = json_object_get(submission, "emailId");
```
rsto, line 1431: Better to pull the string value out of the JSON value here, and for destmboxid and setkeywords later? This would also allow to better handle NULL values in the "that-never-happens" case.

imap/caldav_alarm.c
```diff
@@ -1132,6 +1132,145 @@ static int process_valarms(struct mailbox *mailbox,
 }
 
 #ifdef WITH_JMAP
+static int move_to_mailboxid(struct mailbox *srcmbox,
```
rsto, line 1135: nit: is move_to_mailbox introduced in this commit, but used in the previous commit already? If so, they should be one commit.

imap/jmap_mail_submission.c
```diff
@@ -555,6 +555,35 @@ static void _emailsubmission_create(jmap_req_t *req,
...
+                const char *keyword;
+                json_t *jbool;
+
+                jmap_parser_push(&parser, "setKeywords");
+                json_object_foreach(jval, keyword, jbool) {
+                    if (!json_is_boolean(jbool) ||
+                        !jmap_email_keyword_is_valid(keyword)) {
+                        jmap_parser_invalid(&parser, keyword);
+                    }
+                }
```
rsto, line 580: isn't an

    else jmap_parser_invalid(parser, field)

missing here for unknown arguments?

rsto (commented):
I don't have the full details on the Scheduled Email requirements, but from what I know the logic looks sane.

Still found a few of places, where I think the code could be clearer or improved.

Probably I read the commits in the wrong direction, so please excuse it that's the case (you'll know what I mean when you see it)

cassandane/Cassandane/Cyrus/JMAPEmailSubmission.pm
```diff
@@ -1311,5 +1311,218 @@ sub test_emailsubmission_onsuccess_not_using
...
+                    role => "scheduled"
+                },
+                "3" => {
+                    name => "Sent",
+                    role => "sent"
+                }
+            }
+         }, "R1"]
+    ]);
+    $self->assert_str_equals('Mailbox/set', $res->[0][0]);
```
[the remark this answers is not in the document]
ksmurchison: Yeah, you're right.

cassandane/Cassandane/Cyrus/JMAPEmailSubmission.pm
```diff
@@ -1311,5 +1311,218 @@ sub test_emailsubmission_onsuccess_not_using
...
+    $self->assert_equals(JSON::false, $mbox->{myRights}->{mayRename});
+    $self->assert_equals(JSON::false, $mbox->{myRights}->{mayDelete});
+    $self->assert_num_equals(0, $mbox->{totalEmails});
+    $self->assert_num_equals(0, $mbox->{unreadEmails});
+    $self->assert_num_equals(0, $mbox->{totalThreads});
+    $self->assert_num_equals(0, $mbox->{unreadThreads});
+
+    xlog $self, "Generate a email via IMAP";
+    $self->{store}->set_folder("Inbox.Drafts");
+    $self->make_message("foo", body => "foo\r\n") or die;
```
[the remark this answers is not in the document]
ksmurchison: Nope

imap/jmap_mail.c
ksmurchison: I understand your comment, but we haven't added JACL_REMOVEITEMS at this point and we DO want it for non-scheduled mailboxes.  I will try to fix the comment

imap/caldav_alarm.c
ksmurchison: This was committed before it was used in process_futurerelease()

imap/caldav_alarm.c
ksmurchison: Good catch

imap/jmap_mail_submission.c
ksmurchison: Yep

imap/jmap_mail.c
ksmurchison: Clarified the logic and removing confusing comments

cassandane/Cassandane/Cyrus/JMAPEmailSubmission.pm
```diff
@@ -1311,5 +1311,218 @@ sub test_emailsubmission_onsuccess_not_using
...
+    $self->assert_not_null($res->[0][1]{created});
+    my $draftsid = $res->[0][1]{created}{"1"}{id};
+    my $schedid = $res->[0][1]{created}{"2"}{id};
+    my $sentid = $res->[0][1]{created}{"3"}{id};
+
+    xlog $self, "get mailbox $schedid";
+    $res = $jmap->CallMethods([['Mailbox/get', { ids => [$schedid] }, "R1"]]);
+    $self->assert_str_equals($schedid, $res->[0][1]{list}[0]->{id});
+
+    my $mbox = $res->[0][1]{list}[0];
```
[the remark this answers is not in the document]
ksmurchison: Done

rsto (approved):
LGTM. I suggest we squash and merge the commits, as only the final commit provides the new feature.

rjbs:
I have removed include-in-fastmail for the time being while I do some new builds.

elliefm (changes requested):
This looks like a new feature?  It needs a changes/next file for next year's release notes

elliefm:
Could also update the "JMAP implementation status" to mention it at the same time, so the docs don't immediately become stale again.  It's in docsrc/imap/download/installation/http/jmap.rst

ksmurchison:
@elliefm I will definitely add a changes/next
As far as JMAP implementation status, this is a FM feature which doesn't have an external specification, kind of like Snooze.  Do we document FM/Cyrus specific JMAP stuff?

elliefm:
Good question; I'm not sure!

If it's not now and never will be a standard part of JMAP, then it probably doesn't need to be on the "JMAP implementation status" page.

I still think it's worth having documentation for what it is and how to use it, since if we've gone to all the trouble to build it we might as well finish the job and make it useful to others too.  But that falls into a more general documentation category that we're usually pretty behind on anyway; I don't even know where we would put such documentation.  And if there's no spec for this feature then I imagine it'll change a lot as we work out what we need from it, and maybe it makes more sense to save documentation effort until the target stops moving.

elliefm (approved):
I've only looked at the changes/next file since @rsto already approved the rest of it,  and that's now fine, thanks!

ksmurchison:
@brong Can you review today's (March 14) commits?

imap/caldav_alarm.c
```diff
@@ -909,7 +978,13 @@ EXPORTED int caldav_alarm_sync_nextcheck(struct mailbox *mailbox, const struct i
...
 HIDDEN int caldav_alarm_delete_record(const char *mboxname, uint32_t imap_uid)
 {
-    return update_alarmdb(mboxname, imap_uid, 0);
+    return update_alarmdb(mboxname, imap_uid, 0, 0, 0);
+}
+
+static int caldav_alarm_bump_nextcheck(const char *mboxname,
+                                       uint32_t imap_uid, time_t nextcheck)
+{
+    return update_alarmdb(mboxname, imap_uid, nextcheck, 0, 0);
```
brong, line 987: Should this be using a constant for `atype`?

imap/caldav_alarm.c
```diff
@@ -870,13 +905,39 @@ static int read_lastalarm(struct mailbox *mailbox,
...
+        atype = ALARM_CALENDAR;
+        break;
+    case MBTYPE_EMAIL:
+        atype = ALARM_SNOOZE;
+        break;
+    case MBTYPE_JMAPSUBMIT:
+        atype = ALARM_SEND;
+        break;
+    default:
+        assert(0);
```
brong, line 923: This should be `fatal("reason", EX_SOFTWARE);`

imap/caldav_alarm.c
brong: Ahh, I have followed up - it's OK :)  If `atype` is zero, then we do an update which doesn't change the type.  Fair enough.

brong (approved):
I had a concern, but I checked and it wasn't.  The `fatal` rather than `assert(0)` is still a good change suggestion, but not a blocker to deploy.

imap/jmap_mailbox.c
```diff
@@ -2920,6 +2922,11 @@ static void _mbox_destroy(jmap_req_t *req, const char *mboxid,
         goto done;
     }
 
+    /* Do not allow to remove $scheduled */
+    if (mboxname_isscheduledmailbox(mbentry->name, mbentry->mbtype)) {
```
brong, line 2926: We shouldn't allow deletion of specialuse folders at all, so this should be subsumed by that rather than an explicit check for the scheduled mailbox.

imap/caldav_alarm.c
```diff
@@ -1660,8 +1664,8 @@ static int process_futurerelease(struct caldav_alarm_data *data,
         if (setkeywords) json_decref(setkeywords);
         mailbox_close(&sched_mbox);
         free(frock.mboxname);
-        free(frock.userid);
     }
+    free(userid);
```
brong, line 1663: yeah this is nice not calculating it twice :)

imap/caldav_alarm.c
```diff
@@ -1783,7 +1783,7 @@ static void process_one_record(struct caldav_alarm_data *data, time_t runtime, i
 #ifdef WITH_JMAP
     else if (mbtype_isa(mailbox_mbtype(mailbox)) == MBTYPE_JMAPSUBMIT) {
         if (record.internaldate > runtime || dryrun) {
-            caldav_alarm_bump_nextcheck(data, record.internaldate, runtime, error_message(r));
+            caldav_alarm_bump_nextcheck(data, record.internaldate, runtime, NULL);
```
brong: Is this OK to be running a replace after a past error, or will that wipe the error?

(yeah, I know I'm being annoying, but I'm not sure the various orders in which this can happen)

imap/caldav_alarm.c
```diff
@@ -1702,7 +1697,7 @@ static int process_snoozed(struct caldav_alarm_data *data,
 
     /* Check runtime against wakeup and adjust as necessary */
     if (dryrun || wakeup > runtime) {
-        caldav_alarm_bump_nextcheck(data, wakeup, runtime, NULL);
+        caldav_alarm_bump_nextcheck(data, wakeup, 0, NULL);
```
brong, line 1700: For dryrun you need the zero here (don't change it) but for `wakeup greater` don't we actually need to track the runtime?

brong (commented):
Couple of comments on the new commits.  I'm not 100% sure on the runtime tracking changes yet, but maybe just a chat would clear it up.

imap/caldav_alarm.c
ksmurchison: I don't think we want to adjust the last_run with runtime here because we actually haven't tried to unsnooze.  We got to this point earlier than the actual wakeup time.

imap/caldav_alarm.c
ksmurchison: @brong If we pass a null err_str to bump_nextcheck, we use the current data->err_str in the replace.  Same if last_run is 0 (zero).

But, your comment highlights an issue with the above line in that I should be passing last_run as 0 so we don't change it, since we haven't actually tried to send the message here.

imap/jmap_mail_submission.c
```diff
@@ -1121,6 +1121,18 @@ static int getsubmission(struct jmap_get *get,
             json_object_del(sub, "envelope");
         }
 
+        /* created */
+        if (jmap_wantprop(get->props, "created")) {
+            char datestr[RFC3339_DATETIME_MAX];
```
rsto, line 1126: All the other TIME_MAX macros define the maximum length excluding the nul character, I assume that's also the case for RFC3339.

rsto (approved):
Once it's clear that there is no risk of buffer overflow in the RFC3339 buffer, it looks good to me.

imap/jmap_mail_submission.c
ksmurchison: RFC3339_DATETIME_MAX is is used as above throught Cyrus.  Its defined as 21 and time_to_rfc3339() always writes a UTC time string of length 20

rsto: OK, great. it's weird that it is the only time MACRO that includes the nul byte (or the comments for the other macros are wrong). but I always found the MAX length definitions without nul byte to be asking for trouble anyway.

cassandane/Cassandane/Cyrus/JMAPEmailSubmission.pm
```diff
@@ -1857,5 +1857,154 @@ sub test_emailsubmission_scheduled_send_fail
...
+            properties => ["mailboxIds"],
+        }, "R2"],
+    ] );
+    $self->assert_num_equals(1, scalar @{$res->[0][1]->{list}});
+    $self->assert_str_equals('final', $res->[0][1]->{list}[0]->{undoStatus});
+
+    $self->assert_num_equals(1, scalar @{$res->[1][1]->{notFound}});
+    $self->assert_equals($emailid1, $res->[1][1]->{notFound}[0]);
+
+
```
wolfsage, line 2003: nit: extra newline

wolfsage (commented):
LGTM

cassandane/Cassandane/Cyrus/JMAPEmailSubmission.pm
ksmurchison: Fixed

ksmurchison: Fixed

rjbs:
@ksmurchison "caldav_alarm.cL" in the latest commit message is a typo.

PR #3965, opened 2022-03-10
caldav_sched: send iTIP REPLY if ORGANIZER changed

rsto:
Each calendar in Google Calendar has a separate organizer. If the
owner of a scheduled event moves their event from one calendar
to the other, an invite with the new ORGANIZER is sent out.

The processimip code rejects this change, but a Cyrus user may add
the updated event to their calendar. In this case, send out a
REPLY as otherwise the invitee would not show up as attending in
Google Calendar.


ksmurchison (approved):
LGTM

PR #3967, opened 2022-03-10
ci: Travis to use cyrus-buster instead of cyrus-jessie

ajaysusarla:
While the tag was `cyrus-jessie`, we were building `buster`, so named
the tag appropriatedly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,10 @@ before_install:
 - echo '{"ipv6":true,"fixed-cidr-v6":"2001:db8:1::/64"}' | sudo tee /etc/docker/daemon.json
 - echo 'core' | sudo tee /proc/sys/kernel/core_pattern
 - sudo service docker restart
-- docker pull cyrusimapdocker/cyrus-jessie
+- docker pull cyrusimapdocker/cyrus-buster
 script:
 - docker run --env TRAVIS_BRANCH --env TRAVIS_PULL_REQUEST --env TRAVIS_PULL_REQUEST_BRANCH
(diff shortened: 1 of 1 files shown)


ajaysusarla:
The newly tagged docker images are available here:https://hub.docker.com/repository/docker/cyrusimapdocker/cyrus-buster

elliefm (approved):

elliefm:
Do you happen to know if Travis always uses the ".travis.yml" from the master branch, or does it use the ".travis.yml" from the branch it's building?

In the latter case I guess I'll need to cherry-pick this back to the stable branches too

elliefm:
No worries, I'll get it picked back shortly then

ajaysusarla:
> No worries, I'll get it picked back shortly then

@elliefm do we want that though? The old tag will continue to exist. The reason I think we should not do it is because, the cyrus-libs that are built into the images will always be the latest. So we might have a state where the branches rely on some versions of cyrus-libs, which don't exist anymore.
Although, we have that problem even now.

elliefm:
For me it's mostly about not wanting the additional load of keeping track of which branches use which docker image.

If we wind up with breaking changes due to cyruslibs moving forward, IMO we can just deal with that when it happens... perhaps by saying "this release is so old that we're no longer going to bother running CI for it".

We already have problems like this for 3.0 (which uses libical features that are now deprecated), and we get around that by building 3.0 with -Wno-error so the deprecated functions are just warnings instead of errors.

ajaysusarla:
ack! that sounds reasonable!

elliefm:
This is now on the 3.6, 3.4, and 3.2 branches

For 3.0, looks like we don't use docker at all (the .travis.yml is very different and the cherry-pick fails with loads of conflicts), so it can just stay how it is.

PR #3968, opened 2022-03-11
cassandane: murder tests require --enable-murder

elliefm:
Turns out Cassandane::Instance breaks really badly if it tries to start up with murder configuration and there's no "mupdate" binary...

This fix simply annotates the MurderIMAP and MurderJMAP tests with the :need_component_murder feature test, so that if the installed Cyrus was built without --enable-murder, they won't try to run (and no Cassandane::Instances will be instantiated).

### DIFF
--- a/cassandane/Cassandane/Cyrus/MurderIMAP.pm
+++ b/cassandane/Cassandane/Cyrus/MurderIMAP.pm
@@ -202,6 +202,7 @@ sub check_user
 }
 
 sub test_aaasetup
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):
@elliefm rjbs and I just ran into this on inabox

elliefm:
Oh yeah, that's how I found it too!

elliefm:
This is now on 3.6 too

PR #3970, opened 2022-03-12
Cassandane::Tiny - cassandane tests in many small files

rjbs:
🧪 This is sort of a proof of concept, but possibly good enough to move forward.  (I do think we can improve on this quite a bit, though.)

## Why?

The chief problem I'm trying to solve is merge conflicts.  Right now, we have a few *very* large test files (like JMAPEmail) to which tests get appended.  When there are several active branches, this can result in a bunch of conflict with the tests all being added in one place.  To avoid that, we make every test a single file.  Then the main conflict is that we must not use the same name more than once.

## Cassandane::Tiny

The mechanism here is pretty gross, but I've tried to make it sort of safe for future fix-ups.

Cassandane tests use Test::Unit, in which subroutines with names beginning with `test_` are tests.  The test runner loads the classes, gets a list of their methods, and runs them.  This branch introduces **Cassandane::Tiny::Loader**, which is invoked with a path to a tree of test files.  Each test file is loaded into a test class, and is expected to contain one test subroutine.  Nothing enforces this, but every test must indicate that it's meant to be loaded this way by including `use Cassandane::Tiny`.  Later, we may provide an alternate boilerplate module.

I split JMAPCalendar and JMAPEmail to use tiny tests, found in ./cassandane/tiny-tests subdirectories.

## Selective Running

It's already possible to run `testrunner JMAPEmail/email_set_foo`.  This branch adds, as somewhat unrelated work, the ability to run `testrunner JMAPEmail/email_set_*`, which will match anything starting with `email_set`.  You can also use negated tests, so that…

```
$ ./testrunner JMAPEmail '!JMAPEmail/email*'
```

…means "run all the JMAPEmail tests that don't have a name starting with "email".

## Next Steps

I'd like to be able to ditch Test::Unit over time, but this seems like a very big project.

We should convert more tests into tiny tests.  (Process described below.)

We should eliminate the difference between any two classes loading tiny tests.  That is: test selection by wildcard means that there's not much reason for individual test classes.  We should be able to make a single class that loads all tests.  This means figuring out what methods in each class can be made common, and how to resolve conflicts.  *We can do this over a long time if we like.*

I'd like to ensure that the filename and test name are directly related, rather than being similar by convention.  This can be done later.

## How to convert a Cassandane test class to tiny tests

This could be more automated, but I'm not sure it'd be worth the time…

1. reorganize the `.pm` file so that all the `test_` subroutines are at the end and all the non-test subroutines are ahead
2. write all the test subroutines to a new file (called, say "foo")
3. delete them from the `.pm` file — make sure you leave the `1;` in the pm file
4. in place of the test subroutines, add `use Cassandane::Tiny::Loader 'foo';`
5. run `perl tiny-test-splitter.pl foo foo`
6. add `tiny-tests/foo` and commit everything
7. make sure it works!

My most common mistake was getting warnings out of step 5 and realizing that I'd missed some subs in 2/3.


cassandane/tiny-tests/calendar/account-get-capabilities
```diff
@@ -0,0 +1,52 @@
+#!perl
```
elliefm, line 1: Historically, we have a massive copyright comment blob at the top of every file.  I'd love to be rid of them...

elliefm (commented):
[inline comments, mostly of the "that's interesting, I wonder why you did it like that" variety]

elliefm:
The github review UI was really struggling to cope with so many changes, so I didn't want to spend too much time typing into the "review" box in case I lost it all...

The way the tiny-tests subdirectory contents are namespaced, without direct reference to the suite the test belongs to, seems like it would imply that multiple test suites could as equally pull in the same test function, but I don't think that's true?  The tests are very dependent on the `sub new`, `sub set_up` and `sub tear_down` they were called with, and I'm not sure it's useful for them to be theoretically-generalised.  But given that, if you're starting by looking in the tiny-tests directory at e.g.  "calendar/account-get-capabilities", it's not apparent that it comes from JMAPCalendars specifically ("calendar" could easily be a CalDAV test, for example).  The prefix is helpful but I think it ought to match the suite name?

The individual test files fail a "perl -c filename" test.  First, because they `use Cassandane::Tiny` without first `use lib '.';`, and so can't find that.  But if I manually fix that in one file for experiments' sake, it then fails on our custom attributes:

```
Invalid CODE attributes: min_version_3_1 : needs_component_sieve : needs_component_jmap at tiny-tests/email/thread-changes line 233.
```

I rely a lot on being able to syntax-check tests with `perl -c` to catch compilation errors/missing dependencies/etc _without_ the whole cassandane infrastructure needing to bootstrap first, and at the moment this breaks that, and it would be nice if there were a way it could not break it.  It might even be simple?  I don't know much about CODE attributes -- I've tinkered with ours a lot but I've mostly been copying what was already there.

The suites themselves (e.g. JMAPEmail.pm) are still "syntax OK", cool.

I'm currently running the JMAPEmail suite and it seems to be working fine...  as does JMAPCalendars.  Cool.

I'm curious whether there's any difference in overall run time -- so here's the timing from a random other full run earlier this morning, before I lose it from my terminal history:

```
Time: 1234 wallclock secs ( 1.62 usr  1.19 sys + 499.31 cusr 654.75 csys = 1156.87 CPU)
```

I'll post back in a bit with a full runtime from this version.  I don't really expect much change, but I was gonna do a full run anyway, so I might as well compare.

elliefm:
```
Time: 1216 wallclock secs ( 1.59 usr  1.33 sys + 483.29 cusr 644.82 csys = 1131.03 CPU)
```

same/same

cassandane/Cassandane/Tiny/Loader.pm
```diff
@@ -0,0 +1,41 @@
...
+sub import {
+  my ($class, $path) = @_;
+
+  my $into = caller;
+
+  unless (-d $path) {
+    Carp::confess(qq{can't find path "$path" for loading tests; Cassandane expects to be run from the ./cyrus-imapd/cassandane directory"});
+  }
+
+  my @tests = `find $path -type f`;
```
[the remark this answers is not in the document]
rjbs: Well, mostly it's laziness.  File::Find, in my opinion, has an obnoxious interface that I don't like thinking about.  Anything else would've required I think about how to get another prereq in place for anybody running these tests, so I didn't.  The Perl wisdom is:  "Perl is a glue language.  Don't forget that you still have all that stuff for using the glue *on*!"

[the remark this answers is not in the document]
rjbs: Also, I thought maybe you'd be interested in this old gem:

https://rjbs.cloud/blog/2013/01/the-speed-of-perl-file-finders/

https://rjbs.cloud/blog/2013/01/more-on-the-speed-of-file-finding/

cassandane/Cassandane/Unit/TestPlan.pm
```diff
@@ -569,8 +589,16 @@ sub _parse_test_spec
             my $test;
             ($fpath, $test) = ($fpath =~ m/^(.*)\/([^\/]+)$/);
             next unless defined $test;
-            return ($neg, 'f', "$fpath.pm", $test)
-                if ( -f "$fpath.pm" );
+
+            if ( -f "$fpath.pm" ) {
+                if ($test =~ /\*/) {
+                    my @hunks = split /\*+/, $test, -1;
+                    my $regex = join q{.*}, map {; quotemeta } @hunks;
```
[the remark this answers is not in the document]
rjbs: Here, not much.  But this is a defensive measure against the problem that `{...}` are used for both blocks and anonymous hash references.  A few constructs in Perl can accept either a block or an expression, like map.  Example:

```
my @result = map { $_, quotemeta } @hunks;
my @result = map { $_, quotemeta }, @hunks;
```

There are two possible invocations of `map`:

* map BLOCK LIST
* map EXPR,LIST

Note that the expression form is always followed by a comma… but the grammar is just not good enough to always do exactly what you might expect.  Some day you will write `map { ... } ...` and you'll get what you didn't expect.  Putting a `;` inside the first brace is meant to say for sure "this is a block, not an expression".  I'm not sure it's fullproof.  You will sometimes see this approached from the other angle, writing `+{...}` to say "this is an expression, not a block".

cassandane/tiny-tests/calendar/account-get-capabilities
rjbs: I can put them back if we want (or need) to, but I'm hoping we can avoid it. :)

rjbs:
> But given that, if you're starting by looking in the tiny-tests directory at e.g. "calendar/account-get-capabilities", it's not apparent that it comes from JMAPCalendars specifically ("calendar" could easily be a CalDAV test, for example). The prefix is helpful but I think it ought to match the suite name?

Fair point, I can do this.

rjbs:
> I rely a lot on being able to syntax-check tests with perl -c to catch compilation errors/missing dependencies/etc without the whole cassandane infrastructure needing to bootstrap first, and at the moment this breaks that, and it would be nice if there were a way it could not break it.

I would like to make this work, but the short form is:  If you `perl -c .../JMAPCalendars.pm` it will load all of the subfiles.  Stick to doing that, imho.  Over time, we can tweak the sitch.

rjbs:
@elliefm I have renamed the little directories.

Do you think anything else stands in the way of this being merged?

cassandane/Cassandane/Unit/TestPlan.pm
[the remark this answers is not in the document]
elliefm: Oh that's excellent intel, and I'm glad I asked!

cassandane/tiny-tests/calendar/account-get-capabilities
elliefm: I'm not comfortable making copyright decisions, so I just meticulously copy what was already being done, rather than risk becoming responsible for a bad choice. 😅 

Interestingly, this is Cassandane, which is variously Copyright Fastmail/Opera, _not_ CMU, so the calculus is different from what it would be for Cyrus itself.

I do not like having copyright statements at the top of every file, but I will defer to whatever decision someone else makes.  I bring up the issue to observe that there's a decision point here, so it can be made consciously rather than accidentally.

elliefm:
> I would like to make this work, but the short form is: If you `perl -c .../JMAPCalendars.pm` it will load all of the subfiles. Stick to doing that, imho. Over time, we can tweak the sitch.

Oh, that's pretty reasonable, thanks!

cassandane/Cassandane/Cyrus/JMAPEmail.pm
```diff
@@ -381,6 +381,6 @@ sub _set_quotalimits
     $self->assert_str_equals('ok', $admintalk->get_last_completion_response());
 }
 
-use Cassandane::Tiny::Loader 'tiny-tests/email';
+use Cassandane::Tiny::Loader 'JMAPEmail';
```
elliefm, line 384: whoops, missing `tiny-tests/`

cassandane/Cassandane/Tiny.pm
```diff
@@ -0,0 +1,10 @@
+package Cassandane::Tiny;
+use strict;
+use warnings;
+
+sub import {
+  $Cassandane::Tiny::Loader::RELOADED = 1;
```
elliefm, line 6: This could use a `no warnings 'once';` to shush the "possible typo" warning

elliefm:
> > I would like to make this work, but the short form is: If you `perl -c .../JMAPCalendars.pm` it will load all of the subfiles. Stick to doing that, imho. Over time, we can tweak the sitch.
> 
> Oh, that's pretty reasonable, thanks!

Alas, it doesn't work as nicely as I'd like:  I introduced some junk into tiny-tests/JMAPCalendars/calendarevent-blobid, but `perl -c [...]/JMAPCalendars.pm` still reported syntax OK.  So I guess the syntax check is not that deep. :(

> Do you think anything else stands in the way of this being merged?

Nope, this is pretty good, just nits left

cassandane/Cassandane/Tiny.pm
rjbs: Done.

PR #3972, opened 2022-03-13
Fix build with jmap disabled

buschmann23:
When jmap is disabled as compile option, build fails with missing includes and failed linking.


elliefm:
This is now on 3.6 too

PR #3973, opened 2022-03-14
Support Cyrus build on Debian arm64

rsto:
Some sections in the Cyrus codebase implicitly include assumptions that are only guaranteed on the x86 platform. This patch changes these to work on both x86 and arm64. Also includes fixes that showed up when compiling with clang instead of gcc.

The test system used was

    Linux xxx 5.10.0-12-arm64 #1 SMP Debian 5.10.103-1 (2022-03-07) aarch64 GNU/Linux

All cunit and Cassandane tests pass on both platforms.

### DIFF
--- a/imap/cyr_ls.c
+++ b/imap/cyr_ls.c
@@ -71,6 +71,7 @@
 #include <sys/param.h>
 #include <pwd.h>
 #include <grp.h>
+#include <inttypes.h>
 
 #include "bsearch.h"
 #include "util.h"
@@ -182,7 +183,7 @@ static void long_list(struct stat *statp)
 
     strftime(datestr, 13, datefmt, localtime(&(statp->st_ctime)));
 
-    printf("%c%c%c%c%c%c%c%c%c%c %lu %-8s %-8s % 10ld %s ",
+    printf("%c%c%c%c%c%c%c%c%c%c %" PRIuMAX " %-8s %-8s % 10ld %s ",
            S_ISDIR(statp->st_mode) ? 'd' : '-',
            (statp->st_mode & S_IRUSR) ? 'r' : '-',
            (statp->st_mode & S_IWUSR) ? 'w' : '-',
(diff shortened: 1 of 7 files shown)


ksmurchison (approved):
Looks good other than my one comment

imap/cyr_ls.c
```diff
@@ -182,7 +182,7 @@ static void long_list(struct stat *statp)
 
     strftime(datestr, 13, datefmt, localtime(&(statp->st_ctime)));
 
-    printf("%c%c%c%c%c%c%c%c%c%c %lu %-8s %-8s % 10ld %s ",
+    printf("%c%c%c%c%c%c%c%c%c%c %ju %-8s %-8s % 10ld %s ",
```
elliefm, line 185: `j` isn't portable here.  Please use the `PRIuMAX` macro (may need to include `<inttypes.h>` if it's not already included)

Something like this (untested):

```
    printf("%c%c%c%c%c%c%c%c%c%c %" PRIuMAX "u %-8s %-8s % 10ld %s ",
```

elliefm (changes requested):
Looks good, but one portability concern (ironically!)

elliefm:
builds and tests fine on :

Linux xxx 4.19.0-18-amd64 #1 SMP Debian 4.19.208-1 (2021-09-29) x86_64 GNU/Linux

imap/cyr_ls.c
rsto: Thanks! Did not know about `PRIuMAX`

elliefm (approved):

imap/cyr_ls.c
elliefm: There's one of these `PRI[iu][size]` portable printf width macros for each of the `[u]int[size]_t` types, and a set of corresponding `SCN...` ones for use with scanf (which are usually the same as the printf one, but can theoretically be different).  I think they were introduced in C99?

brong (approved):
This looks good.

PR #3974, opened 2022-03-15
cunit: don't interpolate the arguments into the format string!

brong:
This was causing crashers with newer compilers.  It's such an obnoxious thing.

Much better to pass the strings to the formatter as separate strings.


rsto (approved):
Compiles and runs cleanly for me, using `gcc (Debian 10.2.1-6)`

elliefm (approved):

PR #3975, opened 2022-03-15
PortManager: start at port 29100

brong:
This is higher and less likely to clash with other things

### DIFF
--- a/cassandane/Cassandane/PortManager.pm
+++ b/cassandane/Cassandane/PortManager.pm
@@ -61,9 +61,8 @@ sub alloc
         die "Invalid TEST_UNIT_WORKER_ID - code not run in Worker context"
             if (defined($workerid) && $workerid eq 'invalid');
         my $cassini = Cassandane::Cassini->instance();
-        my $cassini_base_port = $cassini->val('cassandane', 'base_port') // 0;
-        $base_port = 0 + $cassini_base_port || 9100;
-        $base_port += $max_ports * ($workerid-1);
+        my $cassandane_base_port = 0 + $cassini->val('cassandane', 'base_port', '29100');
+        $base_port = $cassandane_base_port + $max_ports * ($workerid-1);
     }
     for (my $i = 0 ; $i < $max_ports ; $i++)
     {
--- a/cassandane/Cassandane/Test/Cassini.pm
+++ b/cassandane/Cassandane/Test/Cassini.pm
@@ -192,7 +192,7 @@ sub test_environment_override
         'cassandane.pwcheck' => 'alwaystrue',
         'cassandane.cleanup' => 'no',
         'cassandane.maxworkers' => '1',
-        'cassandane.base_port' => '9100',
+        'cassandane.base_port' => '29100',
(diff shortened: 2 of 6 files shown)


rsto:
I'm going to check this on my VM in a bit, but it will take a while.

ksmurchison:
We should also change the comment on the default value for `base_port` in `cassandane.ini.example`

rsto:
Tests passed. Way better than it was before.

rsto (approved):
LGTM

elliefm (changes requested):
Tested, works fine.

Agreed with Ken's comment though: please update the other `9100`s too... `ack 9100` finds these:

```
cassandane/cassandane.ini.example
74:##base_port = 9100

cassandane/genmail3.pl
66:        port => 9100,

cassandane/imap-append.pl
59:        port => 9100,
```

The latter two look like utility scripts which are assuming the default is 9100, but that assumption will no longer hold, and they'll break if used.  I don't think we use them, but we can save our future selves a few minutes of confusion if we ever do, by keeping their assumptions updated.

`ack '\b91\d\d\b'` additionally finds these ones:

```
cassandane/Cassandane/Test/MessageStoreFactory.pm
94:    $ms = Cassandane::MessageStoreFactory->create(uri => 'imap://victoria:secret@foo.com:9143/inbox.foo');
99:    $self->assert($ms->{port} == 9143);
102:    $ms = Cassandane::MessageStoreFactory->create(uri => 'imap://victoria@foo.com:9143/inbox.foo');
107:    $self->assert($ms->{port} == 9143);
110:    $ms = Cassandane::MessageStoreFactory->create(uri => 'imap://foo.com:9143/inbox.foo');
115:    $self->assert($ms->{port} == 9143);
```

... though I don't think these ones actually start a service on that port, they're just testing that the object is created correctly, so they won't actually clash with anything nor care that the default range has changed.

cassandane/Cassandane/PortManager.pm
```diff
@@ -58,7 +58,7 @@ sub alloc
             if (defined($workerid) && $workerid eq 'invalid');
         my $cassini = Cassandane::Cassini->instance();
         my $cassini_base_port = $cassini->val('cassandane', 'base_port') // 0;
-        $base_port = 0 + $cassini_base_port || 9100;
+        $base_port = 0 + $cassini_base_port || 29100;
```
elliefm, line 61: Heh... `$cassini->val()` takes an optional third argument, the default value to use if no value was configured.  So these two lines could have been one all along, like:

```
$base_port = 0 + $cassini->val('cassandane', 'base_port', '[2]9100');
```

elliefm (approved):

brong:
I saw we ran into a 9100 port issue again recently, so would be great to merge this!  @elliefm I did the patch for the default value thing you suggested

cassandane/Cassandane/PortManager.pm
brong: Yeah!  Done :)

elliefm (approved):
Looks good, thanks :)

PR #3976, opened 2022-03-15
JMAP Backup/restoreEmail fix

ksmurchison:
This fixes a bug where if the same message is destroyed + restored twice, the $restored flag would be missing on the message after the second restore.

### DIFF
--- a/cassandane/Cassandane/Cyrus/JMAPBackup.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPBackup.pm
@@ -2177,7 +2177,7 @@ sub test_restore_notes_all_dryrun
     $self->assert_num_equals(0, scalar @{$res->[0][1]{destroyed}});
 }
 
-sub test_restore_mail_twice
+sub test_restore_mail_twice_slow
     :min_version_3_3 :needs_component_jmap
 {
     my ($self) = @_;
@@ -2229,12 +2229,12 @@ sub test_restore_mail_twice
     ]);
 
     # need a gap between move and destroy otherwise we will restore both copies
-    sleep 3;
+    sleep 1;
 
     xlog "destroy email1";
     $res = $jmap->CallMethods([
         ['Email/set', {
-            destroy => ["$emailId1"]
+            destroy => [$emailId1]
          }, "R5"],
     ]);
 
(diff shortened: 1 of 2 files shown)


cassandane/Cassandane/Cyrus/JMAPBackup.pm
```diff
@@ -2177,4 +2177,136 @@ sub test_restore_notes_all_dryrun
...
+        ['Email/set', {
+            update => { $emailId1 => {
+                "mailboxIds/$inbox->{id}" => undef,
+                "mailboxIds/$trash->{id}" => JSON::true
+                } }
+         }, "R5"]
+    ]);
+
+    # need a gap between move and destroy otherwise we will restore both copies
+    sleep 3;
```
rsto, line 2232: Is one second enough, too? If not, you could append `_slow` to the test function name to make it only run when slow tests are enabled.

cassandane/Cassandane/Cyrus/JMAPBackup.pm
```diff
@@ -2177,4 +2177,136 @@ sub test_restore_notes_all_dryrun
...
+         }, "R5"],
+    ]);
+
+    xlog "restore mail prior to most recent changes";
+    $res = $jmap->CallMethods([
+        ['Backup/restoreMail', {
+            undoPeriod => "PT1S"
+         }, "R7"],
+        ['Email/get', {
+            ids => ["$emailId1"],
```
rsto, line 2247: nit: there's no need for the double quotes here and later in the code

cassandane/Cassandane/Cyrus/JMAPBackup.pm
```diff
@@ -2177,4 +2177,136 @@ sub test_restore_notes_all_dryrun
...
+
+    $self->assert_num_equals(1, scalar(@{$res->[1][1]{list}}));
+    $self->assert_str_equals("$emailId1", $res->[1][1]{list}[0]{id});
+    $self->assert_str_equals("$emailAt1", $res->[1][1]{list}[0]{receivedAt});
+    $self->assert_null($res->[1][1]{list}[0]{keywords}->{'$restored'});
+    $self->assert_equals(JSON::true, $res->[1][1]{list}[0]{mailboxIds}{$trash->{id}});
+    $self->assert_null($res->[1][1]{list}[0]{mailboxIds}->{$inbox->{id}});
+
+    # need a gap between destroys otherwise we will restore both copies
+    sleep 3;
```
rsto, line 2282: This test definitely needs `_slow` added.

rsto (changes requested):
The code looks good to me, but the test takes too much time to include it in regular cassandane runs.

rsto (approved):
LGTM

PR #3977, opened 2022-03-15
iTIP REQUEST Fix

ksmurchison:
Organizers MUST send the entire per-attendee view of an event in an iTIP REQUEST.
Therefore, any component NOT in the request MUST be removed from the attendee's existing iCalendar resource.

This PR also fixes existing Cass tests to properly include ALL components in each iTIP REQUEST.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Sieve.pm
+++ b/cassandane/Cassandane/Cyrus/Sieve.pm
@@ -4507,7 +4507,7 @@ EOF
     xlog $self, "Check that the message made it to INBOX";
     $self->check_messages({ 1 => $msg }, check_guid => 0);
 
-    xlog $self, "Check that the updated event made it to calendar";
+    xlog $self, "Check that the standalone instance was dropped from the calendar";
     $events = $CalDAV->GetEvents($CalendarId);
 
     $self->assert_equals(1, scalar @$events);
--- a/cassandane/Cassandane/Cyrus/Sieve.pm
+++ b/cassandane/Cassandane/Cyrus/Sieve.pm
@@ -4328,6 +4328,20 @@ CREATED:20210923T034327Z
 UID:$uuid
 DTEND;TZID=America/New_York:20210923T183000
 TRANSP:OPAQUE
+SUMMARY:An Event
+DTSTART;TZID=American/New_York:20210923T153000
+DTSTAMP:20210923T034327Z
+SEQUENCE:0
+RRULE:FREQ=WEEKLY
+ORGANIZER;CN=Test User:MAILTO:foo\@example.net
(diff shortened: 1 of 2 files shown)


cassandane/Cassandane/Cyrus/Sieve.pm
```diff
@@ -4371,6 +4385,138 @@ EOF
...
+                        flags => [ '\\Recent', '\\Flagged' ]);
+    $self->{instance}->deliver($msg);
+
+    xlog $self, "Check that the message made it to INBOX";
+    $self->check_messages({ 1 => $msg }, check_guid => 0);
+
+    xlog $self, "Check that the updated event made it to calendar";
+    $events = $CalDAV->GetEvents($CalendarId);
+
+    $self->assert_equals(1, scalar @$events);
```
rsto, line 4513: I'd add a "check that the standalone instance got dropped from the calendar" or the like, as this seems to be test all about.

imap/itip_support.c
```diff
@@ -714,6 +714,21 @@ static int deliver_merge_request(const char *attendee,
         icalcomponent_add_component(ical, new_comp);
     }
 
+    /* Remove components of old object that still remain in the hash table */
```
rsto, line 717: This is the case of "Remove components of old object that are not present in the new object"? If so, the comment should just tell so, I had to read the code to figure out what's the purpose of the hash table.

imap/itip_support.c
```diff
@@ -594,13 +594,13 @@ static int deliver_merge_request(const char *attendee,
     for (; itip; itip = icalcomponent_get_next_component(request, kind)) {
         icalcomponent *new_comp = icalcomponent_clone(itip);
 
-        /* Lookup this comp in the hash table */
+        /* Lookup and remove this comp from the hash table */
```
rsto, line 597: Probably this should read "Lookup and remove this comp in the old object"?

rsto (commented):
Code LGTM, but I think a few comments could be made clearer for future readers.

cassandane/Cassandane/Cyrus/Sieve.pm
ksmurchison: Yep

imap/itip_support.c
ksmurchison: Actually we are just removing it from the hash table, NOT the old object (yet).  I added clarifying comment

imap/itip_support.c
ksmurchison: Added clarifying comment

rsto (approved):
LGTM

brong (approved):
This looks right to me - if the organizer sends a full REQUEST then it replaces anything currently stored locally.

PR #3978, opened 2022-03-15
RENAME INBOX INBOX.foo fix

ksmurchison:
Need an index_init so index_check() has a protstream.
Fixes a bug in RENAME INBOX INBOX.foo if INBOX hasn't been selected.
This bug was reported by Sentry via an assert()

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -7780,11 +7780,11 @@ static void cmd_rename(char *tag, char *oldname, char *newname, char *location)
             state = imapd_index;
         }
         else {
-            struct index_init init;
-            memset(&init, 0, sizeof(init));
-            init.userid = imapd_userid;
-            init.authstate = imapd_authstate;
-            init.out = imapd_out;
+            struct index_init init = {
+                .out = imapd_out,
+                .userid = imapd_userid,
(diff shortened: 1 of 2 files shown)


imap/imapd.c
```diff
@@ -7780,7 +7780,13 @@ static void cmd_rename(char *tag, char *oldname, char *newname, char *location)
             state = imapd_index;
         }
         else {
-            r = index_open(oldmailboxname, NULL, &state);
+            struct index_init init;
+            memset(&init, 0, sizeof(init));
```
elliefm, line 7784: this could be `struct index_init init = {0};`

ksmurchison: Yep.  Will fix.  We just discussed this over dinner.  Will ad this to coding standards

brong: This feels very amenable to a test too!

brong (changes requested):

imap/imapd.c
ksmurchison: Initialization fixed

brong (approved):
Looks great.  Thanks for the test :)

PR #3980, opened 2022-03-16
SASL initial response fix

ksmurchison:
prevent crash if initial response is incomplete base64


brong (approved):
Could do with a little more description in the PR or commit message about why this change is needed - but it's also pretty clear from the code.

PR #3981, opened 2022-03-16
Jmap blob 09

brong:
This adds the new Blob/upload JMAP endpoint and updates all the tests to use it

### DIFF
--- a/cassandane/Cassandane/Cyrus/JMAPCalendars.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPCalendars.pm
@@ -7804,11 +7804,11 @@ EOF
     ];
 
     my $res = $jmap->CallMethods([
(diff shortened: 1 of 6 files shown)


brong:
@elliefm this plus the "remove the old way" and probably the "update to the IETF urn" will want to be backported to 3.6 as well.

ksmurchison (approved):
LGTM

elliefm:
> @elliefm this plus the "remove the old way" and probably the "update to the IETF urn" will want to be backported to 3.6 as well.

@brong Which ones are these?  The latter might be #3685?  Not sure what the former refers to though

brong:
I've removed the backwards compat code now :)

rjbs:
We think this is ready, per Cyrus dev call today.

PR #3982, opened 2022-03-16
carddav_update()

ksmurchison:
Add and use carddav_update() when only \Flagged has changed on a record.
This saves us having to parse the vCard unnecessarily


brong (approved):
This looks good.

rsto (approved):
LGTM

PR #3983, opened 2022-03-16
conversations: count email as unseen if unseen in any folder

rsto:
Before, conversation folder counts only counted an email as unseen
if it was unseen in that folder. This conflicts with JMAP because
an Email object is reported as unseen if it unseen in any mailbox
where a copy of that email exists in.

Since folder unseen counts only are used for JMAP, this patch
changes the unseen count to implement the JMAP Email semantics.
Only the Trash folder counts unseen as previously.


cassandane/Cassandane/Cyrus/JMAPMailbox.pm
```diff
@@ -4188,11 +4189,19 @@ EOF
     # T3 in (a,b) unseen
     # T4 in (a,trash) seen
 
-    $self->_check_counts('Initial Test',
-        a => [ 5, 2, 4, 2 ],
-        b => [ 3, 1, 2, 2 ],
-        trash => [ 1, 0, 1, 0 ],
-    );
+    if ($maj > 3 || ($maj == 3 && $min >= 7)) {
```
brong, line 4192: This feels like it should be backported to 3.6 - since it's a bugfix for JMAP behaviour.

brong (approved):
This looks great, thanks.

@elliefm - I raised a comment about backporting this to 3.6 in the version checks for the tests.  I think we should backport this, since it makes JMAP more correct and doesn't affect any other codepaths.

elliefm:
Sure, shall do.  I've added an assignment to myself and the 3.6 label, so this PR will sit in my list until I do it (even after it's already been merged to master).

@brong @ksmurchison @rsto It's fine for you all to do this (i.e. assign to me and label) if you have a bug fix you already know needs to be backported to 3.6.  Not merely fine, very appreciated!  Cause it saves me having to figure it out myself, and reduces the risk that I miss it.

Assignments are separate from (just below) the "reviewers" list.

If you're not sure if the affected code is in 3.6 or not, it's okay to just assign it to me anyway, and say you're not sure, so that if it doesn't backport cleanly I don't get confused. :)

Rule of thumb for whether something is in 3.6 or not: if it was new this year, no.  If it was new last year (or earlier), probably yes.

Thanks!

elliefm:
This is now on 3.6, though the JMAPMailbox.mailbox_counts test fails because of hardcoded checks for 3.7 in the test.  I'll have a new PR with a fix for that up shortly.

PR #3984, opened 2022-03-17
Negative system user rights

ksmurchison:
Also include negative rights in is_system_user()


brong (approved):
Clearly and obviously good!  Good catch.

PR #3985, opened 2022-03-17
Mailbox/set{create} server-set properties

ksmurchison:
Mailbox/set{create} must return all server-set properties


brong (approved):
Yep, that looks good.

PR #3986, opened 2022-03-17
Calendar/set{create} myRights

ksmurchison:
Calendar/set{create} must return myRights

### DIFF
--- a/imap/jmap_calendar.c
+++ b/imap/jmap_calendar.c
@@ -2080,7 +2080,7 @@ static void setcalendars_create(struct jmap_req *req,
     }
 
     /* Report calendar as created. */
-    *record = json_pack("{s:s s:o }", "id", uid,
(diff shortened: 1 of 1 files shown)


imap/jmap_calendar.c
```diff
@@ -2080,7 +2080,10 @@ static void setcalendars_create(struct jmap_req *req,
     }
 
     /* Report calendar as created. */
-    *record = json_pack("{s:s}", "id", uid);
+    *record = json_pack("{s:s s:o }", "id", uid,
```
brong, line 2083: Just so I can complain about something, there's excess whitespace after the `:o` here...

brong (approved):
Looks good.  On tiny nit.

imap/jmap_calendar.c
ksmurchison: Fixed

PR #3987, opened 2022-03-18
cassandane: fix up "cleanup" bugs

elliefm:
Here's how the cleanup option is supposed to work, according to documentation:

https://github.com/cyrusimap/cyrus-imapd/blob/5b2bf0f12955469241a6c75c5d88f1781377c13c/cassandane/doc/running_tests.txt#L155-L160

> Firstly, it immediately cleans up any files left over in /var/tmp/cass/.

This wasn't working consistently, because I guess at some point we changed how we name the instance directories without changing the regex we use to recognise them at cleanup, and so they would only match (and be cleaned up) by accident.

The first commit here fixes the regex to match how these names are created, so that leftovers from previous runs are correctly removed at startup.  Also adds comments cross-referencing both places that need to be updated next time we change either.

It turns out this is obscenely noisy, so the second commit only logs that it's doing this if we're running verbosely.

> Secondly, it cleans up any such files after each test, unless the test fails.

This was a lie.  If cleanup was enabled, instance directories were always cleaned up after the test ran, regardless of whether it passed or failed.  This made the cleanup option pretty useless -- if some test failed, you couldn't debug it, because the state was already gone.

The third commit here fixes this by moving the cleanup out of `tear_down()`, which has no way of knowing the result, and into `post_tear_down()`, which does.  It's worth observing that `post_tear_down()` only applies to test suites that inherit from Cassandane::Cyrus::TestCase (not Cassandane::Unit::Testcase), but that's fine, because only Cassandane::Cyrus::TestCase tests create Instance objects in the first place.

I plan to backport this to 3.6, and maybe earlier.

### DIFF
--- a/cassandane/Cassandane/Cyrus/TestCase.pm
+++ b/cassandane/Cassandane/Cyrus/TestCase.pm
@@ -46,6 +46,7 @@ use Scalar::Util qw(refaddr);
 use List::Util qw(uniq);
 use Digest::file qw(digest_file_hex);
 use File::Temp qw(tempfile);
+use File::Path qw(rmtree);
 
 use lib '.';
 use base qw(Cassandane::Unit::TestCase);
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):
This didn't appear to break anything on my system

elliefm:
@rsto Could you have a look at this one too please?  You were particularly interested in the cleanup feature, so I'd like to know it works for you

elliefm:
This is now on the old branches as far back as 2.5

PR #3988, opened 2022-03-19
jmap_ical: allow creation time of override to differ from main event

rsto:
Before, the 'created' timestamp of a recurrence override was not
allowed to differ from the main event. Now, the created property
can be set for new overrides, but the value must not be changed
later. As for main events, a created timestamp in the future is
clamped to server time.


ksmurchison (approved):
LGTM

Issue #3941, opened 2022-02-28
cyrusimap-3.4.2 sieve delivery to inbox as well as fileinto subfolder


dekeonus:
During delivery / sieve processing lmtpunix is delivering to INBOX _as well as_ folder set by fileinto

```
Mar  1 07:38:18 MAILHOST cyrus_imap/lmtpunix[23570]: Delivered: <E1nOmk0-0007qO-U7@smtp-06-tuk1.steampowered.com> to mailbox: user.MYUSER
Mar  1 07:38:18 MAILHOST cyrus_imap/lmtpunix[23570]: Delivered: <E1nOmk0-0007qO-U7@smtp-06-tuk1.steampowered.com> to mailbox: user.MYUSER.email-accts.gmail
Mar  1 07:38:18 MAILHOST cyrus_imap/lmtpunix[23570]: USAGE MYUSER user: 0.497246 sys: 2.105787
```

This is definitely the same messsage, as when I then log into the imap server using my mail client and move the erroneous message from INBOX to email-accts.gmal I get the following

`Mar  1 07:39:07 MAILHOST cyrus_imap/imap[23586]: user.MYUSER.email-accts.gmail: same message appears twice 4898 4899`

Note: it **Always** delivers to INBOX first then the subfolder second (or at the very least, that's always the logged order)
I'm not sure how to get more verbose logging for the sieve filter processing and final delivery action choice.

The active/default sieve script for user:
```
# Sieve Filter: default

##only 'include' as this should just pull in other scripts
require ["include"];

## include fails silently if script does not exist
# Global Amavis filters
include :global "amavisd";

# Webmail generated rules
include :personal "webmail";
```
The global script included above
```
# Sieve Filter: amavisd

require ["fileinto", "regex"];

## only enabled (in amavisd) for select(special) accounts
# File virii into folder Virii
if header :comparator "i;ascii-casemap" :matches "X-Amavis-Alert" "INFECTED*"  {
    fileinto "INBOX/Virii";
    stop;
}

# File spam into folder Junk
if header :comparator "i;ascii-casemap" :matches "X-Spam-Status" "YES*"  {
    fileinto "INBOX/Junk";
    stop;
}
```
The included personal (actual) filter script
```
# Sieve Filter: Personal Rules
# Generated by Ingo (http://www.horde.org/ingo/)
# hand edited by MYUSER

require ["fileinto", "regex"];

[...]

PR #3990, opened 2022-03-21
Cyrus Imapd 3.4.3 - proposed fix for sievebug #3941

slangenhorst:
proposed fix for #3941


elliefm:
@ksmurchison does this look okay to you?  The reasoning behind it is in the linked issue, #3941.

I wonder whether 3.6/master are affected as well?  I guess we probably want a Cassandane test for this, which would tell us pretty quickly

ksmurchison:
@elliefm I think the real fix for this and a larger problem is #3995

ksmurchison:
Should be fixed by #3995

PR #3991, opened 2022-03-22
Allow changing text uuid <-> urn:uuid

ksmurchison:
User ran into an issue where they create a contact via Fastmail UI which writes a v3 vCard having UID with a text value uuid.
emClient tries to update this contact via CardDAV as a v4 vCard having a UID with a urn:uuid (with same uuid).

Cyrus would previously reject this thinking the client is trying to change the UID value, but in actuality the default UID value type for v3 is text and for v4 is URL.


rsto (approved):
LGTM

PR #3992, opened 2022-03-22
jmap_calendar: expose JMAP metadata and UTC times for private events

rsto:
The JMAP Calendars specification requires the following properties
to be public:

- calendarIds
- isDraft
- utcStart
- utcEnd


ksmurchison (approved):
LGTM

PR #3994, opened 2022-03-23
IMAP: forbid non-admins to delete non-IMAP mailboxes

ksmurchison:
Previously, a user could DELETE 'INBOX.#sieve' even though it was hidden from LIST

### DIFF
--- a/cassandane/Cassandane/Cyrus/Delete.pm
+++ b/cassandane/Cassandane/Cyrus/Delete.pm
@@ -910,4 +910,17 @@ sub test_no_delete_with_children
     $self->assert_str_equals('no', $talk->get_last_completion_response());
 }
 
+sub test_no_delete_non_imap
+    :min_version_3_7
+{
+    my ($self) = @_;
+
+    my $store = $self->{store};
+    my $talk = $store->get_client();
+
+    $talk->delete('INBOX.#calendars.Default');
+    $self->assert_str_equals('no', $talk->get_last_completion_response());
+    $self->assert_matches(qr/Permission denied/i, $talk->get_last_error());
+}
+
 1;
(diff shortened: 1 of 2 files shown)


cassandane/Cassandane/Cyrus/Delete.pm
```diff
@@ -910,4 +910,17 @@ sub test_no_delete_with_children
...
+{
+    my ($self) = @_;
+
+    my $store = $self->{store};
+    my $talk = $store->get_client();
+
+    $talk->delete('INBOX.#calendars.Default');
+    $self->assert_str_equals('no', $talk->get_last_completion_response());
+    $self->assert_matches(qr/Permission denied/i, $talk->get_last_error());
+}
```
rsto, line 924: The test could also assert that admins are still able to delete the non-email mailbox. The admin IMAPTalk instance can be obtained with `my $admintalk = $self->{adminstore}->get_client();`

imap/mboxlist.c
```diff
@@ -2156,6 +2156,12 @@ mboxlist_delayed_deletemailbox(const char *name, int isadmin,
     r = mboxlist_lookup_allow_all(name, &mbentry, NULL);
     if (r) goto done;
 
+    /* Don't allow users to delete non-IMAP mailboxes */
+    if (!isadmin && (mbtype_isa(mbentry->mbtype) != MBTYPE_EMAIL)) {
```
rsto, line 2160: Should the `MBOXLIST_DELETE_FORCE` flags allow to delete as non-admin?

rsto (changes requested):

imap/mboxlist.c
ksmurchison: The flag is only allowed for admins, so we don't have to check for its presence here.

PR #3995, opened 2022-03-23
Sieve include and implicit keep fix

ksmurchison:
sieve_eval_bc() should not be handling implicit keep. Implicit keep is handled by do_action_list() and do_sieve_error().
This fixes an issue where implicit keep was being done on an included script rather than just the toplevel script

### DIFF
--- a/cassandane/Cassandane/Cyrus/Sieve.pm
+++ b/cassandane/Cassandane/Cyrus/Sieve.pm
@@ -5810,4 +5810,32 @@ EOF
     }, $res->[0][1]{list}[1]{calendarIds});
 }
 
+sub test_include_cancel_implicit_keep
+    :needs_component_sieve
+{
+    my ($self) = @_;
+
+    xlog $self, "Install a script which includes another";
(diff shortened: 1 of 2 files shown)


elliefm (approved):
Looks good

elliefm:
Once this lands on master, I'll backport it to 3.6 and 3.4

dilyanpalauzov:
This was copied to the 3.4 and 3.6 branches.  Then it was reverted on master and the 3.4 branch and was refixed on master and 3.4 branches by https://github.com/cyrusimap/cyrus-imapd/pull/4052 .   Will the refix land on the 3.6 branch?

elliefm:
Yup.  When I start getting ready for 3.6.0-beta3 I'll work through all the merged PR's with label:3.6, and the refixes are in this set.  Shouldn't be too long, now that we seem to have good fixes for the mboxlist_upgrade problems, but I'm doing new 3.2 and 3.4 releases first, and then I'll focus on 3.6.0-beta3

PR #3998, opened 2022-03-24
Revert "IMAP: forbid non-admins to delete non-IMAP mailboxes (#3994)"

rsto:
This reverts commit 16889c7d3a44c623a2eeea0d57a2293d02c88aa2.

Turns out we still want to be able to delete non-email mailboxes.


ksmurchison (approved):
The test is still valid but the rest is not

PR #3979, opened 2022-03-16
Cassini: check environment for config overrides

elliefm:
This changes allows environment variables to take precedence over options specified in cassandane.ini

The main use case I imagine for this is where your Cyrus build wrapper installs Cyrus to a prefix named for the version of the build: in this case, it's annoying to have to edit cassandane.ini to update the `[cyrus default].prefix` path every time you want to test a new build.  With environment variables supported as configuration sources, you could have your build wrapper also scaffold a CASSINI_CYRUS_DEFAULT_PREFIX environment variable into place as it does the install, and let Cassandane use that to find Cyrus.

This adds a new occurrence of the string "9100", so if #3975 lands first I'll need to fix it up, or vice-versa.

I plan to backport this to the release branches, so that if I start using it in my setup, it doesn't break when I'm doing releases of older versions.

PR #3999, opened 2022-03-25
Cassini: option to permit environment-only config

elliefm:
Since 407000f0af, Cassandane has refused to start without a cassandane.ini, since its absence usually meant no configuration had been done, and nothing would work correctly.

Since #3979, Cassandane has been able to read configuration values from environment variables.  But it is still mandatory to have a cassandane.ini file somewhere, even if all your config was set in environment variables.

This PR adds a new option, "cassandane.allow_noinifile", which defaults to 'no'.  If it is set to 'yes', then Cassandane allows itself to start up, even though there is no cassandane.ini file to be found.  This means if you want to set up your entire config in the environment, without needing a cassandane.ini file, you can do so as long as CASSINI_CASSANDANE_ALLOW_NOINIFILE=yes is also in your environment.

### DIFF
--- a/cassandane/Cassandane/Cassini.pm
+++ b/cassandane/Cassandane/Cassini.pm
@@ -93,8 +93,7 @@ sub new
         }
     }
 
-    die "couldn't find a cassandane.ini file" if not $filename;
-    $filename = abs_path($filename);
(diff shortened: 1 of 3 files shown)


rsto (approved):
LGTM

PR #4000, opened 2022-03-25
http_caldav_sched: properties only are equal if parameters match

rsto:
Before, when determining if a REQUEST should be sent for a changed
iCalendar property, only the value of the old and new properties
were compared. Now also their parameters are included in comparison.

### DIFF
--- a/cassandane/Cassandane/Cyrus/JMAPCalendars.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPCalendars.pm
@@ -20888,6 +20888,81 @@ sub test_calendarevent_get_reset_iter
         $res->[2][1]{list}[1]{start});
 }
 
+sub test_itip_request_tzid_change
+    :min_version_3_7 :needs_component_jmap
+{
+    my ($self) = @_;
+    my $jmap = $self->{jmap};
+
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):
LGTM

dilyanpalauzov:
https://github.com/cyrusimap/cyrus-imapd/pull/3722 approaches the same problem in a different way, or expressed in other words it also takes into account TZID when comparing DATE-TIME values.

dilyanpalauzov:
> properties only are equal if parameters match

The title is misleading: `DSTART:20220327T130000Z` and `DTSTART;TZID=Europe/Sofia:20220327T180000` are equal (the unix timestamps behind both values are equal), but the parameters do not match.

rsto:
> > properties only are equal if parameters match 
> The title is misleading

Equal here means: the string representation of the property value and all its parameters must match. I do not see how the title is misleading in regards to that.

I also do not agree that `DSTART:20220327T130000Z` and `DTSTART;TZID=Europe/Sofia:20220327T180000` should be considered equal when determining if to sendi iTIP messages. Informing attendees about the TZID is crucial, as the timezone definition might change before the scheduled event happens.

PR #4006, opened 2022-03-29
jmap_calendar: reset iCalendar iterator state for each calendar

rsto:
This applies to CalendarEvent/get:

The iCalendar iterator state keeps the current iCalendar data
in-memory as long as the IMAP uid of the event resource remains
unchanged. But the iterator state was not reset when the iterator
switches to the next calendar mailbox.

Consequently, if the last IMAP uid of the previously processed calendar
matched the first IMAP uid in the new calendar, then CalendarEvent/get
erroneously returned the contents of the former event with only the
calendarIds and iCalendar uid having the correct value.

This patch also bumps the cache version of the JMAP CalendarEvent
cache, so that any cached bogus events are flushed from it.

### DIFF
--- a/cassandane/Cassandane/Cyrus/JMAPCalendars.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPCalendars.pm
@@ -20694,4 +20694,65 @@ EOF
     $self->assert_not_null($notif);
 }
 
+sub test_calendarevent_get_reset_iter
+    :min_version_3_7 :needs_component_jmap
+{
+    my ($self) = @_;
+    my $jmap = $self->{jmap};
+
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):
makes sense

elliefm:
Does this need to be backported to 3.6?  I will have a look myself later, but if you happen to just know one way or the other that would be helpful

rsto:
It does not, as we didn't include the iterator optimization in 3.6. This might turn out to be an issue for installations running 3.6 in case they have lots of calendar events with a large number of standalone recurrence instances in them. For now, I think it's enough we bundle the correct version of the iterator in the next release.

PR #4008, opened 2022-03-30
Sanity check moveToMailboxId at awaken time

ksmurchison:
Make sure we don't move the awakened message into a mailbox that we can't/shouldn't

### DIFF
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -1311,8 +1311,23 @@ static int process_snoozed(struct mailbox *mailbox,
     /* Determine destination mailbox of awakened email */
     destmboxid = json_object_get(snoozed, "moveToMailboxId");
     if (destmboxid) {
-        destname = mboxlist_find_uniqueid(json_string_value(destmboxid),
-                                          userid, authstate);
+        mbentry_t *mbentry = NULL;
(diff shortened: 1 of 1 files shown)


rsto (approved):
LGTM

imap/caldav_alarm.c
```diff
@@ -1311,8 +1311,23 @@ static int process_snoozed(struct mailbox *mailbox,
...
+            // MUST be an email mailbox
+            (mbtype_isa(mbentry->mbtype) == MBTYPE_EMAIL) &&
+            // MUST NOT be deleted
+            !(mbentry->mbtype & MBTYPE_DELETED) &&
+            // MUST be able to append messages
+            (cyrus_acl_myrights(authstate, mbentry->acl) & ACL_INSERT) &&
+            // MUST NOT be DELETED mailbox
+            !mboxname_isdeletedmailbox(mbentry->name, NULL) &&
+            // MUST NOT be \Snoozed mailbox
+            strcmp(mbentry->name, mbname_intname(mbname))) {
```
wolfsage, line 1327: mbname here is actually just the INBOX so this doesn't check against the snoozed mailbox. (This will be fixed later)

ksmurchison: Fixed by #4010

PR #4010, opened 2022-03-30
Check awaken mbox fix

ksmurchison:
Actually check awaken mailbox name against \Snoozed mailbox name (NOT Inbox)
Also test awakening into a destroyed mailbox, which should fallback to Inbox


wolfsage:
LGTM

wolfsage (approved):

PR #4011, opened 2022-03-31
caldav_alarm: set main event id in Alert.calendarEventId property

rsto:
Also encodes recurrenceId as JMAP LocalDateTime.


ksmurchison (approved):
Straightforward

PR #4012, opened 2022-03-31
iTIP Reschedule Fix

ksmurchison:
Per RFC 5545, Section 4.4.7:  remove overrides if not included with reschedule of master

### DIFF
--- a/cassandane/Cassandane/Cyrus/Sieve.pm
+++ b/cassandane/Cassandane/Cyrus/Sieve.pm
@@ -4371,6 +4371,124 @@ EOF
     $self->assert_str_equals('DISPLAY', $ical->{entries}[1]{entries}[0]{properties}{action}[0]{value});
 }
 
+sub test_imip_invite_single_then_master
+    :needs_component_sieve :needs_component_httpd :min_version_3_7
+{
+    my ($self) = @_;
+
+    my $IMAP = $self->{store}->get_client();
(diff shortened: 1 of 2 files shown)


cassandane/Cassandane/Cyrus/Sieve.pm
```diff
@@ -4371,6 +4371,138 @@ EOF
...
+EOF
+
+    xlog $self, "Deliver iMIP update";
+    $msg = Cassandane::Message->new(raw => $imip);
+    $msg->set_attribute(uid => 2,
+                        flags => [ '\\Recent', '\\Flagged' ]);
+    $self->{instance}->deliver($msg);
+
+    xlog $self, "Check that the message made it to INBOX";
+    $self->check_messages({ 1 => $msg }, check_guid => 0);
```
rsto, line 4494: Is that right to check `1` here?

rsto (commented):
I have a hard time figuring out the logic is sane, and I think that's just due to the variable names and (probably outdated) comments.

cassandane/Cassandane/Cyrus/Sieve.pm
ksmurchison: This check isn't required for this test so I removed it.

rsto (approved):
Clearer to me now. LGTM

PR #4015, opened 2022-04-04
docs: update weekly meeting time

elliefm:
What it says on the tin, just switching over to the mid-year meeting time

I'll need to cherry-pick this back to all the stable branches too


brong (approved):

elliefm:
This is now on all the branches that we build website content from

PR #4016, opened 2022-04-04
CaldavAlarm: fix recurring_absolute_trigger tz offset

elliefm:
"Australia/Sydney" is only +11 for half the year.  Replace hardcoded compensation with calculated one.


elliefm:
There's probably a better way to do this that doesn't require pretending Australia/Sydney is Z in the first place, but I don't understand the shape of the test well enough to see it.

rsto (approved):
lgtm. but isn't this just `- ($now->offset())` ?

elliefm:
I find non-literal negations a little easier to read if they're expressed a subtraction-from-zero, rather than one more punctuation character in a sea of punctuation characters

PR #4018, opened 2022-04-04
jmap_mail: implement JMAP "$seen" keyword semantics in Email/query

rsto:
Before, a "$seen" keyword filter in Email/query only inspected the
seen flag per mailbox. This contradicts the semantics of the
"$seen" keyword for JMAP Emails, where an email only is seen if
all of its copies in any mailbox have the seen flag set.

Determining the seen state of an email potentially is costly,
as it may require to open each mailbox where a copy of the message
is stored in. To keep the incurred runtime costs low, the patch
uses a query-scoped cache of seen.db entries. At worst, each mailbox
is opened twice during the whole query, doubling the previous
worst time runtime cost of opening each mailbox once.

### DIFF
--- a/cassandane/Cassandane/Cyrus/JMAPEmail.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPEmail.pm
@@ -22143,6 +22143,197 @@ sub test_email_query_seen_shared
     $self->assert_deep_equals([$email2], $res->[3][1]{ids});
 }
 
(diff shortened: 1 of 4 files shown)


brong (approved):
Looks good - the comment on the free function is kinda whatever.

rjbs:
There have been no new tickets related to problems regarding seen counts in weeks.  Can be merged?

PR #4019, opened 2022-04-05
jmap_calendar: default scheduling calendar must be writable

rsto:
The default scheduling calendar defined in CalendarPreferences
must be writable, as does any calendar that Cyrus picks internally
if no default calendar is set. This change also updates
CalendarPreferences/set to immediately pick a new default calendar,
if the user explicitly sets the defaultCalendarId property to null.


ksmurchison (approved):
LGTM

PR #4020, opened 2022-04-06
relocate_by_id fixes

### LINKED ISSUE #4005: relocate_by_id won't relocate user without a xapian activefile
Looks like if the activefile is missing, `relocate_by_id` bails out on the whole operation?

```
cyrus@debian:~$ /dev/shm/cyrus/main/sbin/relocate_by_id -C /dev/shm/cass/casscmd/conf/imapd.conf user.cassandane

Relocating: user.cassandane
Failed to open activefile for /dev/shm/cass/casscmd/conf/user/c/cassandane.xapianactive: Unknown code ____ 254
```

https://github.com/cyrusimap/cyrus-imapd/blob/1e8c4ba7ff5b2b2b44fcf16cc620340606864eb6/imap/relocate_by_id.c#L265-L276

I guess the activefile is missing in this case because squatter has never run.

Looks like `mappedfile_open()` will accept `MAPPEDFILE_CREATE` in its `flags` argument, and then instead of returning an error when the file is missing, it'll just create it.  Could we do this here, instead of bailing out?  It looks like it'll handle it sensibly -- there won't be any items in the newly-created activefile, so it won't try to do anything with them.  And then it'll be able to continue with the rest of the relocate?

If I manually create an empty activefile where it expects it to be, then `relocate_by_id` again, it succeeds:

```
cyrus@debian:~$ mkdir -p /dev/shm/cass/casscmd/conf/user/c/
cyrus@debian:~$ touch /dev/shm/cass/casscmd/conf/user/c/cassandane.xapianactive
cyrus@debian:~$ /dev/shm/cyrus/main/sbin/relocate_by_id -C /dev/shm/cass/casscmd/conf/imapd.conf user.cassandane

Relocating: user.cassandane
	Renaming: /dev/shm/cass/casscmd/data/user/cassandane	 -> /dev/shm/cass/casscmd/data/uuid/b/a/bad4b9d7-16da-4993-8ff7-cf90f350edad
	Renaming: /dev/shm/cass/casscmd/conf/user/c/cassandane.xapianactive	 -> /dev/shm/cass/casscmd/conf/user/uuid/b/a/bad4b9d7-16da-4993-8ff7-cf90f350edad/xapianactive.db
```

So the approach seems fine.... assuming that it's okay for this file to be empty, which I don't know.

### LINKED ISSUE #4014: relocate_by_id command line niceties
From reports on info-cyrus:

* It doesn't seem to complain when it doesn't find any work to do.  If you spell the mailbox name incorrectly, and it finds no matches, then it successfully does nothing.  It should instead complain that the mailbox does not exist.
* I think it might be expecting mailbox names on the command line to be in the internal namespace.  But our other tools expect them to be in the admin namespace, so it's confusing to admins when the format they're used to typing doesn't work (and moreso when coupled with the previous point).  We've dealt with this sort of thing before in e.g. #1699

Our tests only invoke it with the `-u userid` invocation, which seems to work fine.  But we ought to test both invocations.

Also, the man page only provides a `-u` example.

elliefm:
This fixes a few edge case and UX bugs in `relocate_by_id`.  It's probably easiest to review by stepping through a commit at a time, since there's a few not-entirely-related changes to the same code in this one PR.

* It would fail to relocate a user who didn't have a xapian active file (#4005), probably due to the user not having been indexed yet.  I've fixed this by having the mappedfile_open call just create the file if it's missing, which seems to be fine -- the newly created active file is empty, so the subsequent loop over its contents does nothing.  I could instead make it skip the xapian work if it's missing, let me know which you prefer.
* If we're not search_engine=xapian, we definitely won't have a xapian active file, but also shouldn't create one.  So it now checks for search_engine=xapian before doing any xapian work.
* It didn't do namespace handling of mailbox arguments, so admins were expected to provide mailbox arguments in the internal namespace, rather than the admin namespace the other tools want (#4014).  It now does the namespace handling, and therefore accepts admin namespace mailbox arguments.
* If the user or mailbox you asked it to relocate didn't exist (perhaps you provided it the admin namespace instead of the internal namespace it previously expected) it would do nothing, say nothing, and exit success (#4014).  It now tells you if your user/mailbox arguments don't match anything, or the thing they match has already been relocated.

Fixes #4005 
Fixes #4014

### DIFF
--- a/cassandane/Cassandane/Cyrus/FastMail.pm
+++ b/cassandane/Cassandane/Cyrus/FastMail.pm
@@ -1582,7 +1582,7 @@ sub test_sync_reset_legacy
     close(FH);
 
     xlog $self, "files exist";
-    $self->assert(@files);
+    $self->assert_not_equals(0, scalar @files);
 
     $self->{instance}->run_command({ cyrus => 1 }, 'sync_reset', '-f' => 'magicuser' );
 
@@ -1591,7 +1591,7 @@ sub test_sync_reset_legacy
     close(FH);
 
     xlog $self, "no files left for this user";
-    $self->assert(not @files);
+    $self->assert_equals(0, scalar @files);
 }
 
 sub test_sync_reset_nolegacy
@@ -1671,7 +1671,7 @@ sub test_relocate_legacy_nodomain
     close(FH);
 
     xlog $self, "files exist";
-    $self->assert(@files);
+    $self->assert_not_equals(0, scalar @files);
 
     $self->{instance}->run_command({ cyrus => 1 }, 'relocate_by_id', '-u' => "magicuser" );
 
@@ -1680,7 +1680,7 @@ sub test_relocate_legacy_nodomain
     close(FH);
 
     xlog $self, "no files left for this user";
-    $self->assert(not @files);
+    $self->assert_equals(0, scalar @files);
 }
 
 sub test_relocate_legacy_domain
@@ -1712,7 +1712,7 @@ sub test_relocate_legacy_domain
     close(FH);
 
     xlog $self, "files exist";
-    $self->assert(@files);
+    $self->assert_not_equals(0, scalar @files);
 
     $self->{instance}->run_command({ cyrus => 1 }, 'relocate_by_id', '-u' => "magicuser\@example.com" );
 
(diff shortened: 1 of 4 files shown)


ksmurchison (approved):
This all looks great.  Thanks!

elliefm:
This is now on the 3.6 branch too

PR #4024, opened 2022-04-07
jmap_calendar: only create EventNotification for shared calendars

rsto:
Before, an event notification was created for any CalendarEvent/set
regardless if the calendar was shared to any user. This caused
notifications to pile up that no user ever sees before they get
expunged.


ksmurchison (approved):
:+1:

PR #4028, opened 2022-04-08
relocate_by_id: do not implicitly cast from imapopt to enum value

rsto:
gcc (Debian 10.2.1-6) 10.2.1 20210110 rightly complains with `-Wall` about 

```
imap/relocate_by_id.c: In function ‘main’:
imap/relocate_by_id.c:114:44: error: implicit conversion from ‘enum imapopt’ to ‘enum enum_value’ [-Werror=enum-conversion]
  114 |     enum enum_value config_search_engine = IMAPOPT_ZERO;
      |                                            ^~~~~~~~~~~~
cc1: all warnings being treated as errors
```

PR #4030, opened 2022-04-10
relocate_by_id: fix typo in initialisation

elliefm:
Fixes a typo where an enum was zero-initialised using the wrong enum's constant.  Newer compilers than mine apparently have a warning about it (or an error with -Werror).

Replaces previous fix from #4028, which applied a function call to the typo to produce a correct value.  This fix just directly uses the correct value in the first place.

This should be good, but please compile check it, since my compiler couldn't see the original problem, so I can't definitively prove it's "fixed"

### DIFF
--- a/imap/relocate_by_id.c
+++ b/imap/relocate_by_id.c
@@ -111,7 +111,7 @@ int main(int argc, char **argv)
     struct buf buf = BUF_INITIALIZER;
     char *alt_config = NULL;
     mbentry_t *mbentry;
-    enum enum_value config_search_engine = config_getenum(IMAPOPT_ZERO);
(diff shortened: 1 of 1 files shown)


ksmurchison (approved):
WFM

elliefm:
@rsto Pretty sure this will also fix the `fatal error: Internal error: assertion failed: lib/libconfig.c: 163: config_loaded` in the relocate tests.

You can't call `config_getfoo()` until after `cyrus_init()`, so your `config_getenum(IMAPOPT_ZERO)` workaround for the initialisation compile error was the cause of the crash. ;)

elliefm:
Yep, it does :)

PR #4031, opened 2022-04-11
jmap_calendar: preserve custom timezones in CalendarEvent/set

rsto:
Before, the event timezones got rewritten to IANA timezones
even if the client did only update non-time related properties.


ksmurchison (approved):
Seems sane

rjbs:
To be merged.

PR #4033, opened 2022-04-11
Support iTIP replies to RDATE occurrences

ksmurchison:


imap/itip_support.c
```diff
@@ -403,36 +403,56 @@ HIDDEN icalcomponent *master_to_recurrence(icalcomponent *master,
...
             startzone = (icaltimezone *) icaltime_get_timezone(dtstart);
-        }
-        else {
-            /* Convert RECURRENCE-ID to UTC */
-            recurid = icaltime_convert_to_zone(recurid, utc_zone);
-            hash_insert(icaltime_as_ical_string(recurid), comp, &comp_table);
+
+            /* Get RRULE, if any */
+            prop = icalcomponent_get_first_property(master, ICAL_RRULE_PROPERTY);
+            if (prop) rrule = icalproperty_get_rrule(prop);
```
rsto, line 439: This will work for 99.9% of all events, but it does not handle multiple RRULEs. We already have support for multiple recurrence rules:

- ical_support.c defines `struct multirrule_iterator` which we would need to export
- `icalcomponent_myforeach` uses the multirrule, but you would need to strip the RDATEs first from the iCalendar data

Probably for the current patch, it would just be enough to gather all rrule iterators into a list and check each once, as you do not need to generate multiple values from a rrule anyway.

rsto (changes requested):
Mostly looks good to me, but we should support the multiple-RRULE edge case still.

imap/itip_support.c
ksmurchison: Done.  Also uses just single loop to fetch all recurrence properties

PR #4039, opened 2022-04-14
jmap_mail: do not create or move emails in deleted mailboxes

rsto:
The code didn't take delayed delete for mailboxes into account.


ksmurchison (approved):
LGTM

PR #4042, opened 2022-04-18
http_tzdist.c: remove useless assignments

dilyanpalauzov:
as suggested by clang.


dilyanpalauzov:
The clang static analyzer calls this “Dead nested assignment”, and more precisely “Although the value stored to 'ZZZ' is used in the enclosing expression, the value is never actually read from 'ZZZ'”.

ksmurchison (approved):

PR #4048, opened 2022-04-21
http_caldav: only create EventNotification for shared calendars

ksmurchison:
Before, an event notification was created for any PUT/DELETE
regardless if the calendar was shared to any user. This caused
notifications to pile up that no user ever sees before they get
expunged.


wolfsage:
Seems right

wolfsage (approved):

rsto (approved):
LGTM

PR #4051, opened 2022-04-25
Revert "Sieve include and implicit keep fix"

ksmurchison:
Reverts cyrusimap/cyrus-imapd#3995

Although this PR fixed the issue with "include", it broke the why that IMAP flags were handled during implicit keep.
A new fix is forthcoming.


elliefm:
This is now on the 3.4 branch

elliefm:
This is now on the 3.6 branch

PR #4052, opened 2022-04-25
Sieve include and implicit keep fix (bis)

ksmurchison:
Second attempt at fixing this, which also doesn't break handling of IMAP flags on implicit keep

### DIFF
--- a/cassandane/Cassandane/Cyrus/Sieve.pm
+++ b/cassandane/Cassandane/Cyrus/Sieve.pm
@@ -6158,4 +6158,32 @@ EOF
     }, $res->[0][1]{list}[1]{calendarIds});
 }
 
+sub test_include_cancel_implicit_keep
+    :needs_component_sieve
+{
+    my ($self) = @_;
+
+    xlog $self, "Install a script which includes another";
+    $self->{instance}->install_sieve_script(<<EOF
+require ["include"];
+include "foo";
+EOF
+    );
+
+    xlog $self, "Install a script to be included";
+    $self->{instance}->install_sieve_script(<<EOF
+# This should cancel implicit keep
+discard;
+EOF
+    , name=>'foo');
+
+    xlog $self, "Deliver a message";
+    my $msg = $self->{gen}->generate(subject => "Message 1");
+    $self->{instance}->deliver($msg);
+
+    xlog $self, "Check that no messages are in INBOX";
+    $self->{store}->set_folder('INBOX');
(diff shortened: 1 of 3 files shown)


elliefm (changes requested):
This looks okay, and the tests pass...

> which also doesn't break handling of IMAP flags

But the one test case doesn't do anything with IMAP flags.  Is it just the same test as from the previous, reverted, implementation?

I'd like to see a second test which does do something with IMAP flags that triggers the bug in the old implementation, to prove the new implementation fixes the problem (and to prevent us breaking it again next time we're in here)

ksmurchison:
@elliefm Added test_include_fileinto_implicit_keep_flags

cassandane/Cassandane/Cyrus/Sieve.pm
```diff
@@ -6158,4 +6158,66 @@ EOF
...
+{
+    my ($self) = @_;
+
+    $self->{store}->set_fetch_attributes(qw(uid flags));
+
+    xlog $self, "Install a script which includes another";
+    $self->{instance}->install_sieve_script(<<EOF
+require ["include"];
+include "foo";
+# Should implicit keep without \Seen, but with \Flagged
```
elliefm, line 6200: Perl complains about the \S (and the \F) here:

> Unrecognized escape \S passed through at Cassandane/Cyrus/Sieve.pm line 6198.

This is because the here-doc is interpolated (like a `"double quoted string"`).  If I recall correctly, you can force a here-doc to be uninterpolated (like a `'single quoted string'`) by saying `<<'EOF'` instead of `<<EOF`, or something like that.  Or you can just escape the `\`'s the same as you would in any other double-quoted string, like this:

```suggestion
# Should implicit keep without \\Seen, but with \\Flagged
```

elliefm (approved):
Thanks! One nit (with suggestion), see comment inline.

Assuming you apply the suggestion, then this doesn't need another review cycle, so I'm marking it as approved now.

elliefm:
note to self: when picking this back to the stable branches, also include 56796d4601

cassandane/Cassandane/Cyrus/Sieve.pm
ksmurchison: Fixed

elliefm:
This is now on the 3.4 branch

elliefm:
This is now on the 3.6 branch

PR #4056, opened 2022-04-27
promstatsd: can't accumulate stats from files that couldn't be locked

### LINKED ISSUE #4054: Reading possibly uninitialized values in promdir_foreach/accum_stats
Clang Static Analyzer 14 sees this possible flow in imap/promstatsd.c:

On line 272 is called `promdir_foreach(&accum_stats, PROMDIR_FOREACH_DONEPROCS, &all_stats);`.  It jumps to line 146.

Line 163:
```c
struct prom_stats stats;
r = mappedfile_readlock(mf);
if (!r) {
    memcpy(&stats, mappedfile_base(mf), mappedfile_size(mf));
}
r = proc(&stats, rock) // if mappedfile_readlock(mf) returned NON-ZERO, stats is not initialized; proc is accum_stats.
```
Continue on line 195:
```c
    stats_copy = hash_lookup(stats->ident, h); // stats is not initialized
    for (i = 0; i < PROM_NUM_METRICS; i++) {
        stats_copy->metrics[i].value += stats->metrics[i].value; //stats->metrics[i].value is not initialized
    }
```

elliefm:
Found by @dilyanpalauzov using Clang's static analyser: if a stats file is successfully `mappedfile_open()`d, but then the subsequent `mappedfile_readlock()` failed, we would end up trying to read uninitialised memory and probably crash.

I'm not sure what it even means if `mappedfile_open()` succeeds and `mappedfile_readlock()` fails (though the `mappedfile_readlock()` will log an IOERROR explaining why)...

With this fix, if it ever happens, we'll just ignore that file this time, rather than crashing.  The resultant metrics report will be under-counted by whatever counts were in that file, but it should self-correct on the next update, assuming the cause was ephemeral.

Fixes #4054

### DIFF
--- a/imap/promstatsd.c
+++ b/imap/promstatsd.c
@@ -177,10 +177,12 @@ static int promdir_foreach(promdir_foreach_cb *proc, enum promdir_foreach_mode m
         r = mappedfile_open(&mf, fname, 0);
         if (r) continue;
         r = mappedfile_readlock(mf);
-        if (!r) {
-            memcpy(&stats, mappedfile_base(mf), mappedfile_size(mf));
-            mappedfile_unlock(mf);
+        if (r) {
+            mappedfile_close(&mf);
+            continue;
         }
+        memcpy(&stats, mappedfile_base(mf), mappedfile_size(mf));
+        mappedfile_unlock(mf);
         mappedfile_close(&mf);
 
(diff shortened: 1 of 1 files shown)


ksmurchison (approved):
LGTM

PR #4057, opened 2022-04-28
mboxname_isXXXmailbox() rewrite

ksmurchison:
Consolidates code and improves performance of mboxname_isnonimapmailbox() and mboxname_isnondeliverymailbox() by not creating/destroying an mbname_t for every mailbox type that they check.

Also prohibits IMAP DELETEing toplevel non-IMAP mailboxes (e.g. #sieve, #calendars, #jmap)

Unit tests will be introduced by #4090

### DIFF
--- a/imap/mboxname.c
+++ b/imap/mboxname.c
@@ -1499,60 +1499,94 @@ EXPORTED int mboxname_isdeletedmailbox(const char *name, time_t *timestampp)
     return res ? 1 : 0;
 }
 
-/*
- * If (internal) mailbox 'name' is a CALENDAR mailbox
- * returns boolean
- */
-EXPORTED int mboxname_iscalendarmailbox(const char *name, int mbtype)
+#define PREFIX_FLAG_NONIMAP     (1<<0)
+
+static const struct mailbox_prefix {
+    enum imapopt opt;
+    unsigned flags : 1;
+
+} mailbox_prefixes[] = {
+    { IMAPOPT_NOTESMAILBOX,               0                   },
+    { IMAPOPT_CALENDARPREFIX,             PREFIX_FLAG_NONIMAP },
+    { IMAPOPT_ADDRESSBOOKPREFIX,          PREFIX_FLAG_NONIMAP },
+    { IMAPOPT_DAVDRIVEPREFIX,             PREFIX_FLAG_NONIMAP },
+    { IMAPOPT_DAVNOTIFICATIONSPREFIX,     PREFIX_FLAG_NONIMAP },
+    { IMAPOPT_JMAPNOTIFICATIONFOLDER,     PREFIX_FLAG_NONIMAP },
+    { IMAPOPT_JMAPSUBMISSIONFOLDER,       PREFIX_FLAG_NONIMAP },
+    { IMAPOPT_JMAPPUSHSUBSCRIPTIONFOLDER, PREFIX_FLAG_NONIMAP },
+    { IMAPOPT_JMAPUPLOADFOLDER,           PREFIX_FLAG_NONIMAP },
+    { IMAPOPT_SIEVE_FOLDER,               PREFIX_FLAG_NONIMAP },
+    { IMAPOPT_ZERO,                       0                   }
+};
+
+static int _mbname_prefix_isanyof(const mbname_t *mbname,
+                                  const struct mailbox_prefix prefixes[],
+                                  unsigned flags)
 {
-    if (mbtype_isa(mbtype) == MBTYPE_CALENDAR) return 1;  /* Only works on backends */
-    int res = 0;
-
-    mbname_t *mbname = mbname_from_intname(name);
     const strarray_t *boxes = mbname_boxes(mbname);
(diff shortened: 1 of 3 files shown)


imap/mboxname.c
```diff
@@ -1499,190 +1499,121 @@ EXPORTED int mboxname_isdeletedmailbox(const char *name, time_t *timestampp)
...
+    IMAPOPT_SIEVE_FOLDER,
+    IMAPOPT_ZERO
+};
 
-/*
- * If (internal) mailbox 'name' is a ADDRESSBOOK mailbox
- * returns boolean
- */
-EXPORTED int mboxname_isaddressbookmailbox(const char *name, int mbtype)
+static int _mbname_isa(const mbname_t *mbname, const enum imapopt prefix_opts[])
```
elliefm, line 1517: This returns true if the `mbname`'s prefix matches _any of_ the specified `prefix_opts`.  That's fine, but the other way "isa" could be interpreted is that it returns true only if if matched _all of_ them.

So to make it clearer to future readers, I think I would call this `_mbname_isanyof` instead of `_mbname_isa`.  That way they don't have to stop and read the implementation to figure out which kind of "isa" is it.

elliefm (commented):
This looks sensible, I like the approach, even though it's a bit tricky... it's hard to tell from visual inspection whether any behaviour accidentally changed.

I'd like to see some cunit tests that exercise the public `mboxname_isXXXmailbox()` interfaces for various inputs, and verify that they produce results consistent with our expectations.  The tests could be added to the existing cunit/mboxname.testc suite.

Having tests will also be really useful next time we need to tweak stuff in here, to make sure we didn't break it -- especially with regard to the significant ordering of `mailbox_prefix_opts[]` and the 1-offset for non-IMAP mailboxes.

imap/mboxname.c
```diff
@@ -1719,6 +1647,17 @@ EXPORTED int mboxname_isscheduledmailbox(const char *name, int mbtype)
     return res;
 }
 
+/*
+ * If (internal) mailbox 'name' is a user's \Scheduled mailbox
+ * returns boolean
```
rsto, line 1652: Better to replace `boolean` with `non-zero`?

imap/mboxname.c
```diff
@@ -1499,204 +1499,132 @@ EXPORTED int mboxname_isdeletedmailbox(const char *name, time_t *timestampp)
...
-    mbname_t *mbname = mbname_from_intname(name);
-    const strarray_t *boxes = mbname_boxes(mbname);
-    const char *prefix = config_getstring(IMAPOPT_DAVNOTIFICATIONSPREFIX);
-    if (strarray_size(boxes) && !strcmpsafe(prefix, strarray_nth(boxes, 0)))
-        res = 1;
-
-    mbname_free(&mbname);
-    return res;
+    return (strarray_size(mbname_boxes(mbname)) == 1 &&
+            _mbname_prefix_isanyof(mbname, mailbox_prefix_opts+1));  // skip Notes
```
rsto, line 1544: This `+1` is a bit magic, and should we ever update the prefix_opts array it could be wrong altogether. For future maintenance it might cleaner to either

- define the offset of the first non-imap mailbox close to the definition of `mailbox_prefix_opts`
- mailbox_prefix an array of structs, where each struct denotes if the prefix is for a non-imap mailbox
- split the definition of mailbox_prefix_opts in IMAP and non-imap prefixes and only use the according prefix list

rsto (changes requested):
I think it's a good idea, but I agree with @elliefm that a couple of regression tests would help with future maintenance.

ksmurchison:
Tests are in #4090

imap/mboxname.c
ksmurchison: Is now ` _mbname_prefix_isanyof()`

imap/mboxname.c
ksmurchison: Used an array of structs with a flag

imap/mboxname.c
ksmurchison: Done (for all functions)

elliefm:
As part of my re-review I'd like to run it through the new tests from #4090, but there's merge conflicts when I rebase this branch on top of that one.  I could probably just resolve them to run the tests, but maybe I'd resolve them differently and then I wouldn't be testing the same thing anyway...

I've approved #4090, so can you merge that to master, and then rebase this PR on master?  And then I'll be able to run the tests knowing that I'm running them on the right code.

ksmurchison:
I already ran the cunit tests on the new code and they pass

elliefm:
I would assume so!  But we also read our own code before we commit it -- the point of the code review is to have a second set of eyes/hands/hardware do the same thing, and expect the same result.

And in cases like this where the diff is difficult to read (not your fault, just the diff algorithm doesn't cope very well with these kinds of changes), and I might have missed something visually, then I want to also build it and run the tests as part of my review.

ksmurchison:
What I did was check out the branch with the cunit tests and make a copy of mboxname.testc 
The checkout this branch and copy mboxname.testc into place

rsto (approved):
LGTM

PR #4059, opened 2022-04-30
httpd/parse_accept() is never called with NULL parameter

dilyanpalauzov:
The Clang Static Analyzer claims in the call of parse_accept(NULL), that qsort(NULL, …) will be executed, while the first parameter of qsort() may not be null.

This changeset tells the compiler, that the invocation of parse_accept(), or its callers, will not pass NULL as parameter.

### DIFF
--- a/imap/http_dav.h
+++ b/imap/http_dav.h
@@ -601,7 +601,7 @@ int dav_check_precond(struct transaction_t *txn, struct meth_params *params,
                       const char *etag, time_t lastmod);
 int dav_premethod(struct transaction_t *txn);
 unsigned get_preferences(struct transaction_t *txn);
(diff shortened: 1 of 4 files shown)


ksmurchison (changes requested):
If this silences clang, I'm fine with it, even if __attribute__ is GCC compiler specific.

I think, for readability, my preference would be to have the __attribute__ at the end of the function declaration, e.g.:

extern struct accept *parse_accept(const char **hdr) __attribute__((nonnull));

dilyanpalauzov:
The initial warning was that when `parse_accept(NULL)` is called, then `qsort(NULL, …)` will be invoked, which is bad.

With this change, the new warning is that if hdr[0] is NULL, then calling `parse_accept(hdr)` will execute `qsort(NULL, …)`.

PR #4062, opened 2022-04-30
Add some nonnull/returns_nonnull/warn_unused_result attributes

dilyanpalauzov:
This allows the compilers to emit more warnings.

### DIFF
--- a/lib/util.h
+++ b/lib/util.h
@@ -289,13 +289,13 @@ struct buf {
 #define buf_ensure(b, n) do { if ((b)->alloc < (b)->len + (n)) _buf_ensure((b), (n)); } while (0)
 #define buf_putc(b, c) do { buf_ensure((b), 1); (b)->s[(b)->len++] = (c); } while (0)
 
-void _buf_ensure(struct buf *buf, size_t len);
-const char *buf_cstring(const struct buf *buf);
-const char *buf_cstringnull(const struct buf *buf);
-const char *buf_cstringnull_ifempty(const struct buf *buf);
-char *buf_release(struct buf *buf);
-char *buf_newcstring(struct buf *buf);
-char *buf_releasenull(struct buf *buf);
+__attribute__((nonnull)) void _buf_ensure(struct buf *buf, size_t len);
+__attribute__((nonnull, returns_nonnull)) const char *buf_cstring(const struct buf *buf);
+__attribute__((nonnull)) const char *buf_cstringnull(const struct buf *buf);
+__attribute__((nonnull)) const char *buf_cstringnull_ifempty(const struct buf *buf);
+__attribute__((nonnull, returns_nonnull, warn_unused_result)) char *buf_release(struct buf *buf);
+__attribute__((nonnull, returns_nonnull, warn_unused_result)) char *buf_newcstring(struct buf *buf);
+__attribute__((nonnull, warn_unused_result)) char *buf_releasenull(struct buf *buf);
 void buf_getmap(struct buf *buf, const char **base, size_t *len);
 int buf_getline(struct buf *buf, FILE *fp);
 size_t buf_len(const struct buf *buf);
--- a/lib/xmalloc.h
+++ b/lib/xmalloc.h
@@ -48,15 +48,15 @@
 /* for free() */
 #include <stdlib.h>
 
-extern void *xmalloc (size_t size);
-extern void *xzmalloc (size_t size);
-extern void *xcalloc (size_t nmemb, size_t size);
-extern void *xrealloc (void *ptr, size_t size);
-extern char *xstrdup (const char *str);
+__attribute__((malloc, warn_unused_result)) void *xmalloc (size_t size);
+__attribute__((malloc, warn_unused_result)) void *xzmalloc (size_t size);
+__attribute__((malloc, warn_unused_result)) void *xcalloc (size_t nmemb, size_t size);
+__attribute__((warn_unused_result)) void *xrealloc (void *ptr, size_t size);
+__attribute__((malloc, warn_unused_result)) char *xstrdup (const char *str);
 extern char *xstrdupnull (const char *str);
-extern char *xstrdupsafe (const char *str);
-extern char *xstrndup (const char *str, size_t len);
-extern void *xmemdup (const void *ptr, size_t size);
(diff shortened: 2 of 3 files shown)


imap/jmap_mdn.c
```diff
@@ -454,12 +454,10 @@ static json_t *send_mdn(struct jmap_req *req, struct mdn_t *mdn,
...
 
-        syslog(LOG_ERR, "MDN/send failed: %s", desc ? desc : error_message(r));
+        syslog(LOG_ERR, "MDN/send failed: %s", desc);
 
-        if (desc) {
-            err = json_pack("{s:s, s:s}", "type", "serverFail",
-                            "description", desc);
-        }
+        err = json_pack("{s:s, s:s}", "type", "serverFail",
+                        "description", desc);
```
elliefm, line 460: This isn't the right fix.  Even though `desc` will never be NULL, it may point to an empty string `""`.

I think the condition in the ternary should be `desc[0]` rather than `desc`, something like `syslog(..., desc[0] ? desc : error_message(r));`

I'm not sure what the right fix for the `if (...)` block is.  We could make the condition `desc[0]` like the ternary, and therefore only report an error if there is an error string. But even if there is no error string, we know there was an error because of the `if (r)` above.  So it seems strange to maybe not report it.  Maybe this part should also fall back to `error_message(r)` when desc is the empty string.  But I don't know the larger context here -- input from @ksmurchison or @rsto would clear this up.

lib/xmalloc.h
```diff
@@ -48,15 +48,15 @@
...
 
-extern void *xmalloc (size_t size);
-extern void *xzmalloc (size_t size);
-extern void *xcalloc (size_t nmemb, size_t size);
-extern void *xrealloc (void *ptr, size_t size);
-extern char *xstrdup (const char *str);
+__attribute__((malloc, warn_unused_result)) void *xmalloc (size_t size);
+__attribute__((malloc, warn_unused_result)) void *xzmalloc (size_t size);
+__attribute__((malloc, warn_unused_result)) void *xcalloc (size_t nmemb, size_t size);
+__attribute__((malloc, warn_unused_result)) void *xrealloc (void *ptr, size_t size);
```
elliefm, line 54: `xrealloc()` must not have the `malloc` attribute, according to gcc docs (emphasis mine):

> Using the attribute with no arguments is designed to improve optimization by relying on the aliasing property it implies. Functions like malloc and calloc have this property because they return a pointer to uninitialized or zeroed-out, newly obtained storage. **However, functions like realloc do not have this property, as they may return pointers to storage containing pointers to existing objects**. Additionally, since all such functions are assumed to return null only infrequently, callers can be optimized based on that assumption.

The other usages of `malloc`, and the other attributes, all look good.

elliefm (changes requested):
Thanks, this is good use of attributes (with one nit).

There's a problem with the jmap_mdn.c part that I'm not sure how to resolve, because I'm not familiar with the broader context, so I would like @ksmurchison and/or @rsto to also look at this and advise.  Details in the line-by-line comments.

imap/jmap_mdn.c
rsto: Right after `desc` is assigned, it should read `if (!desc[0]) desc = error_message(r)` or the like.

lib/util.h
```diff
@@ -289,13 +289,13 @@ struct buf {
...
 #define buf_putc(b, c) do { buf_ensure((b), 1); (b)->s[(b)->len++] = (c); } while (0)
 
-void _buf_ensure(struct buf *buf, size_t len);
-const char *buf_cstring(const struct buf *buf);
-const char *buf_cstringnull(const struct buf *buf);
-const char *buf_cstringnull_ifempty(const struct buf *buf);
-char *buf_release(struct buf *buf);
-char *buf_newcstring(struct buf *buf);
-char *buf_releasenull(struct buf *buf);
+__attribute__((nonnull)) void _buf_ensure(struct buf *buf, size_t len);
```
rsto, line 292: I'm all for helping the compiler but it seems arbitrary why to just annotate these functions. And if we want annotate all headers, then we should discuss this with the rest of the Cyrus team. The cyrus-devel list is a better place to agree on this.

rsto (commented):
Agree with @elliefm that this needs further tweaking.

lib/util.h
dilyanpalauzov: The initial problem was, that the Clang Static Analyzer complained in one of the `desc ? desc : x` checks, that in case desc is NULL, later in the code bad things will happen (I do not remember which), and looking further it turend out, that desc cannot be NULL.  Hence I added annotations to all involved functions.

rsto: After more thought I come to agree and my initial comment is moot. Setting the attributes to help the compiler for these functions is definitely going to be useful, thanks. The point that @elliefm raised in https://github.com/cyrusimap/cyrus-imapd/pull/4062#discussion_r862542301 still stands.

dilyanpalauzov: > The point that @elliefm raised in [#4062 (comment)](https://github.com/cyrusimap/cyrus-imapd/pull/4062#discussion_r862542301) still stands.

The point I raised in https://github.com/cyrusimap/cyrus-imapd/pull/4062#discussion_r868007376 still stands.

rsto: This is not how we do code reviews in this project. It is great that your patch highlights an issue in the current error handling, but your proposed code change does not actually fix the issue. In such a situation, we typically either include the fix right with the pull request (as a separate commit), or we keep the current erroneous code section and create a new pull request for the fix.

I suggest you either remove the disputed code from your PR, or you change the code to include the actual fix. It is very rare that we have two authors on one pull request and I do not see a reason to change that for this one.

dilyanpalauzov: In this project it is random, whether code review will be done.

I meant you change the code as you think it shall be changed.  Once that changes are upstream, I will adapt the changes here.

Besides, the number of authors of a pull requests is irrelevant.

rsto: I just landed https://github.com/cyrusimap/cyrus-imapd/commit/f5258901cb66e4b0084da9035309630ff26ca17d upstream

dilyanpalauzov:
Can this make progress?

PR #4065, opened 2022-04-30
jmap_ical.c: return value of xjmapid_from_icalm() is never used

dilyanpalauzov:


elliefm:
@rsto This seems inconsequential, a style change really.  The change itself seems fine to me, but maybe the original code was the way it was for consistency with something else, and I value consistency.  You spend the most time in here, so I'm happy to defer to whatever you prefer.

rsto (approved):
Thanks. LGTM

PR #4066, opened 2022-04-30
jmap_calendar:createevent_store() use free() instead of xzfree()

dilyanpalauzov:


elliefm:
@rsto This seems inconsequential, a style change really.  The change itself seems fine to me, but maybe the original code was the way it was for consistency with something else, and I value consistency.  You spend the most time in here, so I'm happy to defer to whatever you prefer.

rsto (commented):
I prefer `xzfree` for new or updated code, as the runtime overhead is neglible almost all cases and it helps with use-after-free bugs when the pointer is reset to NULL. Using free should be used mostly if the variable is part of a small scoped block, or the code is in a hot loop.

dilyanpalauzov:
Now the variable is part of a small scoped code block.

Using `xzfree()` instead of `free()` to hide use-after-free bugs, does not fix the use-after-free bugs, but hides them.

rsto (approved):
Fine

Issue #3896, opened 2022-02-03
ctl_mboxlist dump/undump miss new fields


elliefm:
`ctl_mboxlist -d` dumps mailboxes.db to a flat file, and `ctl_mboxlist -u` undumps such a file into a new mailboxes.db

We document these as the correct tools for doing conversions of mailboxes.db (e.g. "improved_mboxlist_sort" says it requires a dump/undump when changing its value).

However, since some version long ago (3.0 maybe? 2.5?) this tool has become stale.  It does not know about all the things we store in mailboxes.db, and consequently that data is lost across a dump/undump.

There is already a fix for this for 3.4 in #3384, but 3.6 and master introduce additional new fields that this fix doesn't take into account, so it's already stale as soon as 3.6 lands.  The new fields are multi-valued, so "just" extending this work to accommodate them will be quite complicated.

After some discussion at https://cyrus.topicbox.com/groups/devel/Tea9b3c8c5728d1e8-M7b68e56d5c41dafa2acbba63, the plan for 3.6 is to change the dump/undump format to use json and make sure all fields are properly dumped and undumped.  Ideally we'll also preserve the ability to read the older format, if possible, so that people can dump/undump as part of an upgrade process if they wish (though they will still lose any data their older server failed to dump).

I expect to start working on this after the first 3.6 beta is released, but will intend to backport it into whichever 3.6 release comes out after it's done.

elliefm:
There are recover (-r) and checkpoint (-c) implementations in here that have been "deprecated" in favour of ctl_cyrusdb since 2001, and aren't documented anywhere else.  I'm gonna just rip them out while I'm in here.

For some bizarre historical reason the dump implementation is deeply interwoven with the populate-mupdate implementation?  So I guess I can see why this has been neglected so long, it's a mess.  Gonna separate the two actions first, so when I refactor dump, I'm not tripping over unrelated mupdate stuff.

elliefm:
WIP here: https://github.com/cyrusimap/cyrus-imapd/compare/master...elliefm:v37/3896-ctl_mboxlist-dump-undump

elliefm:
It occurs to me that, since the dump loop is initiated by `mboxlist_allmbox()`, that this still can only dump records that are actually mailboxes, and misses all the other stuff that mailboxes.db stores.

I think the right thing to do will be to cyrusdb_foreach over everything in there, and implement the dump/undump as dlist<->json converters?

elliefm:
[...]

PR #4069, opened 2022-05-02
ctl_mboxlist: dump and undump all fields (as JSON)

elliefm:
Full context is in #3896.   Short summary:

* The `ctl_mboxlist` dump (`-d`) and undump (`-u`) actions are stale, and miss many fields -- leading to lost data if mailboxes.db needs to be dumped and undumped
* We recommend using ctl_mboxlist dump and undump for certain conversions, e.g. when turning on or off improved_mboxlist_sort, so it's not good that it loses data
* @dilyanpalauzov provided a patch in #3384 which brings dump and undump up to date for 3.4.  But 3.6 has more new fields, so this patch is already stale.
* The new fields in 3.6+ are multi-valued, so extending the existing file format to support them is non-trivial.  Instead, the dump/undump format is being changed to JSON.

The first commit in this PR is Dilyan's patch for 3.4, rebased to fix a conflict.  I expect, once this PR lands, to backport Dilyan's patch to 3.4 and maybe 3.2.

The conversion to JSON format sits on top of this, to keep the history sane.  I expect to backport this PR to 3.6.

There's a lot of change here... I'd suggest reviewing commit by commit.

I'm getting hung up on how much error checking to implement in the new undump.  The previous undump had very little, and I have implemented at least as much as it had, but I'm not sure how much more to do.  On the one hand, what we had was presumably "good enough" once; on the other hand, with the format being JSON now, people are more likely to tinker with it on the way through.

Putting this up for review and feedback now, even though I don't feel like undump (last commit) is quite finished yet.  If we're happy with the level of error checking, I can remove the "wip" from the commit message.  If we want more, let's talk specifics.


ksmurchison (commented):
This is looking good to me

rsto (commented):
This definitely is going to be an improvement. I skimmed the code and it does not seem to address dumping all db entries using `cyrusdb_foreach`, yet? Or did you decide not to do that?

elliefm:
@rsto I decided not to do that -- the appropriate tool for a context-agnostic database dump/undump is probably `ctl_cyrusdb`, `cvt_cyrusdb`, or `cyr_dbtool`.  None of these directly have dump/undump commands as such, but `cvt_cyrusdb` implicitly has the functionality in that it can convert any database format to/from any other database format -- including "flat", which is just a text file.

The point of `ctl_mboxlist` is that it knows the semantic meaning of mailboxes.db's contents, so it's correct for it to use the mboxlist APIs, and it's good for it to represent the dumped data semantically.  If there happens to be any other stuff in there that would only be found by a `cyrusdb_foreach`, that stuff is presumably junk and can be disregarded.

elliefm:
No longer "draft" status.

The main change since last time is that the documentation now stresses the need for care when undumping, that it won't/can't protect you from putting junk into your mailboxes.db if you ask it to -- I have a vague recollection that we talked about this a few weeks ago and decided that was reasonable.

elliefm:
This is now on the 3.6 branch

Backport of the non-JSON version to 3.4 in progress at https://github.com/cyrusimap/cyrus-imapd/compare/cyrus-imapd-3.4...elliefm:v34/3896-ctl_mboxlist-dump-undump -- I think Cassandane::Instance will need an update to accommodate the changed file format, so I've gotta do/test that yet

elliefm:
This doesn't backport cleanly to 3.4 or earlier, and requires too much new and untested code to make it happen, and too much development time, and nobody's asking for it, so I'm removing it from the 3.4 and 3.2 backport lists instead of just ignoring it every release

PR #4073, opened 2022-05-04
Replicate to archive

### LINKED ISSUE #3830: Add an ENABLE command for sync_client to turn on "spool to archive partition"
When syncing up a new server, you can easily blow out the non-archive space when copying in a big user, before cyr_expire can move all the records.  With this flag, we could force messages to write to the correct place without needing "this came from archive" magic in the wire protocol.

robn:
This adds a capability to the replication protocol that, when enabled, causes sync files to be written to the archive partition instead of the main spool. Closes #3830.

This compiles and has been lightly tested by hand in a test environment. I have tested:

- setting/clearing `archive_enabled`, confirming the the `REPLICATION-ARCHIVE` capability is offered correctly
- running `sync_client` with and without `-a`, and confirming with `strace` that `sync_server` stores and deletes incoming files to the correct sync dir.
- running `sync_server` without `archive_enabled` set, and confirming with `strace` that `sync_server` stores and deletes incoming files to the main spool regardless of the presence or absence of the `-a` switch.

So I'm inclined to say this is plausibly correct, but I don't know the fine details of the replication system and would love someone to read it through properly.

I haven't tested `imapd`-based replication at all as I'm totally unfamiliar with it.

As you see, there's no automated testing here as I'm not sure where to start. All guidance welcome!

I didn't really love the way I've done it, but I think there's no choice but to plumb the archive flag down through the dlist parser, since that's where the sync file stuff is all handled. Switching between the global and `sync_state` flag seems weird too; I've pretty much copied the sieve-mailbox model for that. I guess what I'm saying is, this is a bit yuck, but not new forms of yuck, and nothing that can be fixed without reworking a bit of internals, which is definitely not why I'm here!

### DIFF
--- a/docsrc/imap/reference/manpages/systemcommands/sync_client.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/sync_client.rst
@@ -168,6 +168,17 @@ Options
     compression if both ends support it.  Set this flag when you want
     to abort if compression is not available.
 
+.. option:: -a
+
+    Request the replicate-to-archive feature. If the remote end has the
+    ``archive_enabled`` option set, then it will stage incoming replication on
+    the archive partition instead of the spool partition. If the remote end
+    does not support it, replication will proceed as though **-a** was not
+    provided.  This option is useful when standing up a new replica of an
+    existing server, as most of the stored mail is likely older than the
+    archive threshold and so is destined for the archive partition anyway. By
+    staging on that partition, Cyrus can avoid a cross-partition copy for every
+    message.
 
 Examples
 ========
--- a/backup/backupd.c
+++ b/backup/backupd.c
@@ -645,7 +645,7 @@ static void cmdloop(void)
             if (!backupd_userid) goto nologin;
             if (!strcmp(cmd.s, "Apply")) {
                 struct dlist *dl = NULL;
-                c = dlist_parse(&dl, /*parsekeys*/ 1, /*isbackup*/ 1, backupd_in);
+                c = dlist_parse(&dl, /*parsekeys*/ 1, /*isarchive*/ 0, /*isbackup*/ 1, backupd_in);
                 if (c == EOF) goto missingargs;
                 if (c == '\r') c = prot_getc(backupd_in);
                 if (c != '\n') goto extraargs;
@@ -681,7 +681,7 @@ static void cmdloop(void)
             if (!backupd_userid) goto nologin;
             if (!strcmp(cmd.s, "Get")) {
                 struct dlist *dl = NULL;
-                c = dlist_parse(&dl, /*parsekeys*/ 1, /*isbackup*/ 1, backupd_in);
+                c = dlist_parse(&dl, /*parsekeys*/ 1, /*isarchive*/ 0, /*isbackup*/ 1, backupd_in);
                 if (c == EOF) goto missingargs;
                 if (c == '\r') c = prot_getc(backupd_in);
                 if (c != '\n') goto extraargs;
--- a/backup/lcb_internal.c
+++ b/backup/lcb_internal.c
@@ -77,7 +77,7 @@ HIDDEN int parse_backup_line(struct protstream *in, time_t *ts,
     if (c == EOF)
         goto fail;
 
-    c = dlist_parse(&dl, /*parsekeys*/ 1, 1, in);
+    c = dlist_parse(&dl, /*parsekeys*/ 1, /*isarchive*/ 0, 1, in);
 
     if (!dl) {
         fprintf(stderr, "\ndidn't parse dlist, error %i\n", c);
--- a/imap/append.c
+++ b/imap/append.c
@@ -439,7 +439,7 @@ static int callout_receive_reply(const char *callout,
     prot_setisclient(p, 1);
 
     /* read and parse the reply as a dlist */
-    c = dlist_parse(results, /*parsekeys*/0, /*isbackup*/0, p);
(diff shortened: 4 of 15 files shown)


docsrc/imap/reference/manpages/systemcommands/sync_client.rst
```diff
@@ -168,6 +168,17 @@ Options
     compression if both ends support it.  Set this flag when you want
     to abort if compression is not available.
 
+.. option:: -a
+
+    Request the replicate-to-archive feature. If the remote end has the
+    ``archive_enabled`` option set, then it will stage incoming replication on
+    the archive partition insteazd of the spool partition. If the remote end
```
elliefm, line 175: > insteazd

elliefm (commented):
This looks good, but I don't have any intuition about if anything might be missing, so instead I'm thinking about testing...

Cassandane already has an `:ArchivePartition` magic attribute that theoretically makes it pretty easy to set up tests where the Cyrus instances have archive partitions, so it shouldn't be hard to set up some tests that prove that replicating with `-a` doesn't break, with and without archive_enabled.  "Theoretically" because I'm not sure how it interacts with replication tests (which spin up two separate cyrus instances per test: one primary, one replica).  My gut feel says if you use the magic, both instances will end up with an archive partition configured, but I haven't looked to confirm.

The much more interesting tests, though, will be ones that confirms whether replication with -a does in fact replicate straight to the archive partition, bypassing the main partition.  You mentioned needing strace to verify this by hand, hmm...

One possibility could be making the deep dlist reserve code syslog (at LOG_DEBUG level) its choice about where to write to.  Cassandane catches syslog, so the test could make sure debug:yes is enabled, do the replication, then look for the syslog line and see if it matches what it expected.  If we think this logging would be a nuisance for other debugging purposes, we could potentially gate that particular syslog call behind its own imapd.conf option, "debug_log_sync_partition_choice" or something.  It wouldn't be the first "debug_foo" imapd.conf option that only exists to provide a hook for Cassandane (debug_writefail_guid is the other one).

If you wanted to have a go at writing tests, have a look at cassandane/Cassandane/Cyrus/Replication.pm.  I'm happy to provide any guidance/etc as needed.  On the other hand, if you don't particularly want to have a go, it seems reasonable to delegate that part to us, in which case it'll probably end up with me anyway. :)

There's also cunit tests for dlist in cunit/dlist.testc.  But I don't think it's possible to test this part of dlist from there, since cunit tests don't run with an installation or configuration.

robn:
> Cassandane already has an :ArchivePartition magic attribute

This is a good pointer, thank you. I will get reading!

> You mentioned needing strace to verify this by hand, hmm...

I wouldn't say I _needed_ it, just that I did it that way :)

> One possibility could be making the deep dlist reserve code syslog (at LOG_DEBUG level) its choice about where to write to.

That's probably not bad. I think testing it in isolation is better, but as you say, the dlist cunit tests don't have configuration so that's not really gonna work. This reserve stuff deep inside dlist really keeps telling us help, I'm trapped in the wrong place and I can't get out.

> If you wanted to have a go at writing tests, have a look at cassandane/Cassandane/Cyrus/Replication.pm. I'm happy to provide any guidance/etc as needed. On the other hand, if you don't particularly want to have a go, it seems reasonable to delegate that part to us, in which case it'll probably end up with me anyway. :)

I don't _particularly_ want to have a go, as I mostly just want the feature. But also it would not hurt me to have a go and understand a bit more of the water I swim in, and I do have a spare afternoon tomorrow so I'll take a swing at it and let you know how I get on. Thanks!

robn:
@elliefm I did not let you know, hey. So I got a test env up and running, but I didn't really know how to express what wanted at the time, and now I've just been reminded of this and I've pretty much forgotten what I was doing. So there's that.

elliefm:
@robn Since your last comment I keep periodically trying to swap this back in, but it's not sticking.

I think I thought the changes were fine, but if there was anything missing I didn't trust myself to see that.  I'll page in the others and see what they think, maybe they'll be confident in the changes and not mind there not being tests.

robn:
@elliefm "not sticking" right :confused:. Thanks!

elliefm:
@ksmurchison @rsto can you both please cast an eye over this?  This week preferably, so if we need to discuss it at next week's call, we've already done our homework.

I've looked at it myself and I think it looks okay, but without accompanying tests I'm not confident that I'm not missing something.  I'd like to know what you both think.

If we want tests, then it'll probably fall on us (not robn) to write them, which is fine, it just means the question at this stage is "do we want tests to proceed", and if so, we'll assign someone to write them and go from there.

elliefm:
Working on some tests for this.  So far it's looking like:

1) the replica instance uses its archive partition for the sync staging directory, regardless of how the message was stored on the master instance (good)
2) once staged, the messages are copied to the replica's data partition, regardless of how they were stored on the master instance (surprising)

I'm not entirely certain, but I suspect that 2) might be that the replica instance is applying its own logic to determine whether or not the message should be stored on the archive or data partition.  And since these messages are only seconds old, they get put on the data partition rather than the archive one, of course.

I had assumed on-archive status was a detail that was replicated, but I guess maybe it's not, and replicas just make their own decisions.  So my tests will need a slightly different shape.

elliefm:
Okay, confirmed that that was what was happening, and found a way to deal with it. Cool.

elliefm:
This plus tests: #4193

elliefm:
This has been merged as part of #4193

robn:
Thanks for this @elliefm, really appreciate you doing the lift!

PR #4075, opened 2022-05-04
http_jwt: OpenSSL 3 compat EVP_PKEY_base_id

dbungert:
The WARNING section of the EVP_PKEY_base_id manpage in OpenSSL 3 says
that EVP_PKEY_base_id is "only reliable with EVP_PKEYs that have been
assigned an internal key with EVP_PKEY_assign_*", and the usage here of
base_id seems to be without EVP_PKEY_assign usage.

This same warning points to EVP_PKEY_is_a, which seems fine for this
purpose but is part of OpenSSL 3.

### DIFF
--- a/imap/http_jwt.c
+++ b/imap/http_jwt.c
@@ -328,14 +328,14 @@ static int validate_pkey_type(struct jwt *jwt, EVP_PKEY *pkey)
     if (!jwt->nid)
         return 0;
 
-    if (jwt->nid == EVP_PKEY_base_id(pkey))
-        return 1;
-
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
     if (jwt->nid == EVP_PKEY_HMAC && EVP_PKEY_is_a(pkey, "HMAC"))
         return 1;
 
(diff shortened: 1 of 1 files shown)


imap/http_jwt.c
```diff
@@ -318,6 +323,25 @@ static int parse_token(struct jwt *jwt, const char *in, size_t inlen)
     return 1;
 }
 
+static int validate_pkey_type(struct jwt *jwt, EVP_PKEY *pkey)
+{
+    if (!jwt->nid)
+        return 0;
+
+    if (jwt->nid == EVP_PKEY_base_id(pkey))
```
rsto, line 331: Shouldn't this be in an `#else` branch, similar to how it is done in `read_public_key`? Otherwise the use of `ENV_PKEY_base_id` isn't guaranteed to be safe in OpenSSL v3?

rsto (commented):
Thanks! Just have one question.

rsto:
Many thanks again. I now merge this as-is, but I plan to get rid of the numeric `nid` identifiers altogether. We only support HMAC and RSA currently, so using `EVP_PKEY_is_a` for all OpenSSL versions should be trivial to implement.

elliefm:
This is now on the 3.6 branch

PR #4076, opened 2022-05-05
Support decoding vCard v4 image properties

ksmurchison:
Adds support for decoding data URIs

Probably best to review by ignoring whitespace changes


rsto (approved):
LGTM

PR #4077, opened 2022-05-05
httpd.c: return value of httpdate_gen() always ignored

dilyanpalauzov:


ksmurchison:
We might have future code where we do use the return value.
Is some compiler complaining?

dilyanpalauzov:
No compiler is complaining.

Ideally the return form of all functions shall be evaluated and marked with `__attribute__((warn_unused_result))`, `__attribute__((returns_nonnull))`, `__attribute__((malloc))`, etc.  Afterwards the compiler can give more useful hints or perform optimizations.  In case of `httpdate_gen()` the return value is not needed.

Once the return-value can be utilized, the function can be altered.  But honestly, the return value can be used even now, but it is not used, the same usage-schema will be very likely applied in the future, so I do not think that the return value will be used.

ksmurchison (approved):

PR #4078, opened 2022-05-05
mboxevent.c: string params steal the string

ksmurchison:
This fixes a double free issue


wolfsage:
Sure. Looking at usage of FILL_* it seems like they more typically take copies of strings rather than ownership, but this is probably fine as we're keeping previous behaviour of it anyway and just allowing it to work with buf_free...

wolfsage (approved):

Issue #4060, opened 2022-04-30
Memory leak in xapian_query_new_match()


dilyanpalauzov:
[xapian_query_new_match()](https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/xapian_wrap.cpp#L1949) contains this code:
```c++
Xapian::Query *q = xapian_query_new_match_internal(db, partnum, str);
Xapian::Query *qq = xapian_query_new_match_internal(db, partnum, mystr);
*q |= *qq;
```
The last line invokes a copy constructor for [operator |=](https://xapian.org/docs/apidoc/html/classXapian_1_1Query.html#abb6f891e0fd7ea0b8e61413ebe7614c6).

The Clang Static Analyzer 14 says, that the memory pointed by qq is never freed.  I believe it.  While Xapian does reference counting, it does not call a `delete` as counterpart of `new` from xapian_query_new_match_internal().

Issue #4070, opened 2022-05-02
Memory leak in jmap_api.c:extract_value()


dilyanpalauzov:
`jmap_pointer_decode()` does [`return buf_release(&buf)`](https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/jmap_util.c#L162).  [jmap_api.c:extract_value()](https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/jmap_api.c#L146) has:
```c
p = jmap_pointer_decode(path, top - path);
if (*p == '\0') { //p is not NULL.
    return NULL;
}
```
`free(p)` is missing.

PR #4081, opened 2022-05-09
Fix memory leaks issues 4060 4070

rsto:
Fixes two memory leaks that @dilyanpalauzov found, one in the Xapian wrapper, one in the JMAP result references handling code.

### DIFF
--- a/imap/xapian_wrap.cpp
+++ b/imap/xapian_wrap.cpp
@@ -1955,6 +1955,7 @@ xapian_query_new_match(const xapian_db_t *db, int partnum, const char *str)
             Xapian::Query *qq = xapian_query_new_match_internal(db, partnum, mystr);
             if (qq && q) {
                 *q |= *qq;
+                delete qq;
             }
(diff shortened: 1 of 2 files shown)


imap/xapian_wrap.cpp
```diff
@@ -1955,6 +1955,7 @@ xapian_query_new_match(const xapian_db_t *db, int partnum, const char *str)
             Xapian::Query *qq = xapian_query_new_match_internal(db, partnum, mystr);
             if (qq && q) {
                 *q |= *qq;
+                delete qq;
             }
             else if (!q) q = qq;
         }
```
dilyanpalauzov, line 1961: A final `else delete qq;`?

elliefm:
Fixes #4060 #4070

[putting these in text, rather than the title, so the issues and PR can be crossed linked for history navigation]

ksmurchison (approved):

elliefm:
This is now on the 3.6 branch

PR #4084, opened 2022-05-13
cassandane: suppress libnss Valgrind errors on aarch64

rsto:
Tested on GNU Linux Debian 11.3

### DIFF
--- a/cassandane/vg.supp
+++ b/cassandane/vg.supp
@@ -80,3 +80,17 @@
     fun:service_init
     fun:main
 }
+{
+   # should only be necessary for valgrind versions < 3.17.0
+   # https://valgrind.org/docs/manual/dist.news.html:
+   #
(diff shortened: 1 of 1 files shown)


elliefm (approved):
Seems fine, but it might be nice to have a comment describing the not-problem that's being suppressed (like how `mupdate_detached_threads_leak` above does), so that next time someone audits the suppressions file they can more easily figure out whether this suppression is still needed or not.

Last time I audited it, I removed loads of old suppressions that seemingly weren't needed anymore, and it would've been a quicker job if the weirder ones had been well-described, instead of having to decipher them to figure out what they had been for.

rsto:
Right. I think I found the culprit and updated the suppression file accordingly.

elliefm:
Awesome, thanks!

PR #4086, opened 2022-05-15
Remove useless assignments

dilyanpalauzov:
or, as the Clang Static Analyzer calls them, “Dead nested assignments”.


elliefm (approved):
Looks good, thanks

PR #4087, opened 2022-05-16
cassandane: verify that smtp service actually started

elliefm:
Previously, if Cassandane's fake smtp service failed to start for any reason (e.g. the assigned port was already in use), we wouldn't actually notice this.  We would naively record the port we tried to use as the smtp port, and then tests which required the smtp service would connect to whatever was there... with interesting results depending on what, if anything, was listening.

This PR makes sure the smtpd actually starts, and bails out the whole test if it didn't.

There's a couple of other tweaks to make sure that bailing out during `start()` is safe -- without them, Cassandane's internal worker pool state gets confused, and it can't shut itself down properly.

### DIFF
--- a/cassandane/Cassandane/Instance.pm
+++ b/cassandane/Cassandane/Instance.pm
@@ -604,8 +604,10 @@ sub _list_pid_files
     my ($self) = @_;
 
     my $rundir = $self->{basedir} . "/run";
-    opendir(RUNDIR, $rundir)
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):
LGTM

elliefm:
This is now on the 3.4 branch

elliefm:
This is now on the 3.6 branch

PR #4090, opened 2022-05-16
mboxname.testc: added mboxname_isXXX() tests

ksmurchison:
Precursor to having tests for #4057 

The fixes in here mboxname.c  are to make the unit tests pass (multiple types of mailboxes all use MBTYPE_COLLECTION), but will be overwritten by #4057


elliefm (approved):
Looks good

PR #4092, opened 2022-05-17
jmap_mail: do not try to read seen.db if no database found

rsto:
This makes Cyrus fail with an internal error if a mailbox unexpectedly
requires an open seen.db but no database was found for this user.
Before, Cyrus failed with an assertion which aborted the HTTP request.


ksmurchison (approved):
LGTM

PR #4094, opened 2022-05-18
Remove useless assignments

dilyanpalauzov:
or, as the Clang Static Analyzer calls them, “Dead initializations”.


imap/sievedir.c
```diff
@@ -303,7 +303,8 @@ EXPORTED int sievedir_put_script(const char *sievedir, const char *name,
     /* parse the script */
     sieve_script_t *s = NULL;
     char *myerrors = NULL;
-    int r = sieve_script_parse_string(NULL, content, &myerrors, &s);
```
ksmurchison, line 306: I don't understand how this is considered a “Dead initialization”

dilyanpalauzov: The value stored in `r` is never read.

ksmurchison (approved):

PR #4095, opened 2022-05-18
imtest: memory leak

dilyanpalauzov:


ksmurchison (commented):
We could actually use xzfree(*line) here

ksmurchison (approved):

Issue #4035, opened 2022-04-13
ctl_cyrusdb -r crash on first start after upgrade to 3.6 betas


elliefm:
From the list: https://cyrus.topicbox.com/groups/info/T3e85440ddbb44ec6-Maf769decd0dd45d4572145b8

There is a Debian bug report for this already: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1007965

We're not sure precisely where the crash is happening, because so far nobody who has reported it has had core dumps enabled.  Anecdotally, it appears to be from the mboxlist_upgrade (which upgrades mailboxes.db to support both legacy-storage and uuid-storage of mailboxes).  It also appears that any mailboxes that haven't already been converted yet at the time the crash occurs are left out of the new mailboxes.db, and thus appear to have been "lost".

If the admin has the presence of mind to `reconstruct -f ...`, then the missing mailboxes will come back fine.  If they instead start tinkering on the file system, they can create themselves a mess...

[One user on the Debian bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1007965#25) suggests that it seems to occur for very old mailboxes that do not have a uniqueid.  It's not clear whether this is a guess, or they managed to get a core dump and find the source.  They've provided a patch that creates a uniqueid during the process if there isn't one.

elliefm:
The same user also suggests that if the upgrade process left the original mailboxes.db file in place, and do the conversion into a mailboxes.db.NEW, and finally rename on success, then this wouldn't be a problem.  But instead we rename the original mailboxes.db to mailboxes.db.OLD, and convert into a new mailboxes.db, so after the failure we're left with the half-converted/half-empty version instead of the original file.

elliefm:
@ksmurchison can you have a look at this please, particularly the linked comment on the Debian bug?

I'm not sure whether it's correct to just generate a uniqueid if one is missing, or if the fact it's missing could be indicative of other problems.  Perhaps if the uniqueid is missing, we should bail out (cleanly!) rather than just smoothing it over?  I'm not sure, what do you think?

I think the suggestion about doing the conversion into a mailboxes.db.NEW is a good one, and we should definitely do that if we can.  But I'm not sure if we can, maybe there's a good reason it is the way it is.  What do you think?

ksmurchison:
[...]

PR #4096, opened 2022-05-18
mboxlist_upgrade() fixes

ksmurchison:
Fixes to deal with missing uniqueids and to be more crash resilient #4035

### DIFF
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -1005,6 +1005,7 @@ static int mailbox_mboxlock_reopen(struct mailboxlist *listitem, int locktype, i
 static int mailbox_open_advanced(const char *name,
                                  int locktype,
                                  int index_locktype,
+                                 const mbentry_t *mbe,
                                  struct mailbox **mailboxptr)
 {
     mbentry_t *mbentry = NULL;
@@ -1050,7 +1051,8 @@ static int mailbox_open_advanced(const char *name,
         free(userid);
     }
 
-    r = mboxlist_lookup_allow_all(name, &mbentry, NULL);
+    if (mbe) mbentry = mboxlist_entry_copy(mbe);
+    else r = mboxlist_lookup_allow_all(name, &mbentry, NULL);
 
     if (!r && mbentry->mbtype & MBTYPE_DELETED)
         r = IMAP_MAILBOX_NONEXISTENT;
@@ -1058,6 +1060,7 @@ static int mailbox_open_advanced(const char *name,
     if (r) {
         if (mailbox->local_namespacelock)
             mboxname_release(&mailbox->local_namespacelock);
+        mboxlist_entry_free(&mbentry);
         remove_listitem(listitem);
         return r;
     }
@@ -1123,13 +1126,13 @@ done:
 EXPORTED int mailbox_open_irl(const char *name, struct mailbox **mailboxptr)
 {
     return mailbox_open_advanced(name, LOCK_SHARED, LOCK_SHARED,
-                                 mailboxptr);
+                                 NULL, mailboxptr);
 }
 
 EXPORTED int mailbox_open_iwl(const char *name, struct mailbox **mailboxptr)
 {
     return mailbox_open_advanced(name, LOCK_SHARED, LOCK_EXCLUSIVE,
-                                 mailboxptr);
+                                 NULL, mailboxptr);
 }
 
 EXPORTED int mailbox_open_irlnb(const char *name, struct mailbox **mailboxptr)
@@ -1138,13 +1141,19 @@ EXPORTED int mailbox_open_irlnb(const char *name, struct mailbox **mailboxptr)
                                  LOCK_SHARED|LOCK_NONBLOCK,
                                  /* cannot do nonblocking lock on index...why? */
                                  LOCK_SHARED,
-                                 mailboxptr);
+                                 NULL, mailboxptr);
 }
 
 EXPORTED int mailbox_open_exclusive(const char *name, struct mailbox **mailboxptr)
 {
     return mailbox_open_advanced(name, LOCK_EXCLUSIVE, LOCK_EXCLUSIVE,
-                                 mailboxptr);
+                                 NULL, mailboxptr);
+}
+
+EXPORTED int mailbox_open_from_mbe(const mbentry_t *mbe, struct mailbox **mailboxptr)
+{
+    return mailbox_open_advanced(mbe->name, LOCK_EXCLUSIVE, LOCK_EXCLUSIVE,
+                                 mbe, mailboxptr);
 }
 
 EXPORTED const char *mailbox_name(const struct mailbox *mailbox)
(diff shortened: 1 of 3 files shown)


elliefm:
Oh cool!  At a first glance the approach looks really good, amazing what a fresh set of eyes can see!  But I'm realising as I glance at it that I'm gonna need to look at it again properly when I'm fresh myself, I think I'm about done for today.

I have a bunch of new Cassandane tests from my 3.4 version of these fixes that I've been working on today, so first thing next week I'll run this through those and see what shakes out.  I'll also test it by hand on a deliberately-broken 3.4 upgrade.

imap/mboxlist.c
```diff
@@ -5423,6 +5423,21 @@ static int _foreach_cb(void *rock,
     mbentry->name = xstrndup(key, keylen);
     mbentry->mbtype |= MBTYPE_LEGACY_DIRS;
 
+    if (!mbentry->uniqueid) {
+        /* Fetch uniqueid from cyrus.header */
+        struct mailbox *mailbox = NULL;
+        int r = mailbox_open_from_mbe(mbentry, &mailbox);
+        if (r) {
+            mboxlist_entry_free(&mbentry);
+            return r;
```
elliefm, line 5432: I think this needs to log the error and return 0 rather than r.  Otherwise the cyrusdb_foreach will bail out at the first problem, and whatever mbentries we didn't get to will be missing from the upgraded db later.  Same applies at the end of this function (which doesn't change in this diff).  The overall return value is eventually ignored by process_mboxlist() anyway, so it will look like it succeeded, even if it craps out early and does very little.

imap/mailbox.c
```diff
@@ -1050,14 +1051,16 @@ static int mailbox_open_advanced(const char *name,
         free(userid);
     }
 
-    r = mboxlist_lookup_allow_all(name, &mbentry, NULL);
+    if (mbe) mbentry = mboxlist_entry_copy(mbe);
+    else r = mboxlist_lookup_allow_all(name, &mbentry, NULL);
```
elliefm, line 1055: Oh, this is a very neat solution, nice

elliefm (commented):
I've tested this on a data directory from 3.4 with:

* uniqueid missing from mbentry (but RUNQ entry still present)
* uniqueid missing from both mbentry and header (but RUNQ entry still present)

and in both cases `ctl_cyrusdb -r` successfully upgrades the database without losing the affected user.  Nice!

But if I remove the RUNQ entry as well (since, if there wasn't a uniqueid, there could never have been one), then the upgrade never happens.  I think this is because we only upgrade at all if there are RACL or RUNQ entries found by `_check_rec_cb`.  Under my test, Cyrus isn't configured to keep reverseacl's, so there's no RACL entries, and the only user is the cassandane user, so if its RUNQ entry is missing (due to me having deleted it) then there are no RUNQ entries at all...

Though there is still the `$RUNQ => 1` record saying that reverse uniqueids are enabled, so maybe `_check_rec_cb` could look for that as well?  Or we could invert the logic, and assume we need to upgrade unless we find new-style entries.

One other safety nit, commented inline

imap/mboxlist.c
ksmurchison: So, if we fail to open the mailbox, do we assume that is doesn't exists and ignore the mbentry?

If we do bail, then the original mailboxes.db remains as-is.  We could/should report an error.

elliefm: It's a tricky one -- on one hand, maybe we should crap out completely (i.e. fail all the way up until we exit non-zero from ctl_cyrusdb, so that master doesn't even start services if any part of the upgrade didn't succeed).  The admin will be forced to report/fix whatever is preventing that mailbox from opening, and try again.

On the other hand, if a mailbox fails to open, is that even a class of error that the sysadmin can repair?  Especially in a state where Cyrus can't start up and the cmdline tools can't function because the upgrade hasn't succeeded yet!  I guess it would depend on why it failed to open, which we don't know yet -- it's a hypothetical edge-case question at the moment.

I think for now, the best behaviour is "do everything we possibly can, but log and skip individual mailboxes that don't work for some reason", and that means returning zero even when they fail, so we at least continue with the rest.  I imagine this is probably what you've just done, I haven't looked at your latest updates yet.

I guess the next 3.6 beta will give us some data about what sort of cases, if any, it still trips on, and then having that data we can work out how to deal with it.

ksmurchison: What do we do if  we fail to parse the mbentry?  Try to piece one together from the mailbox header, or log and skip?

ksmurchison:
@elliefm Changed the logic to assume that we need to upgrade unless we find an A or I record.  This should solve the problem when there are no $RACL or $RUNQ records.

imap/mboxlist.c
elliefm: Log and skip too, I think -- at least for now.  Maybe if we see a bunch of failure patterns that are obviously recoverable we can think about whether it's worth it, otherwise let it fall through to reconstruct to handle after the fact.

That being said, I'm not sure how much reconstruct can currently handle -- first it needs the upgrade to succeed and produce a good database, so in my previous work it didn't have anything to work with.  But once this PR is merged, with the upgrade improvements and the new `mailbox_open_mbe()` API, I think I should be able to forward-port my reconstruct and ctl_cyrusdb -r work from 3.4 and have them actually function -- and if that turns out to be the case, then the upgrade skipping some individual entries it couldn't cope with won't be such a big deal, cause the admin can just reconstruct to get them back.

elliefm (commented):
Ahh, turned out there was another bug around `_check_rec_cb()`, but it's not one I can comment on inline cause it's in a line that wasn't changed.  Here's the patch I needed to get it working:

```
diff --git a/imap/mboxlist.c b/imap/mboxlist.c
index 69fe0c256a..c975d5f90d 100644
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -5502,7 +5502,7 @@ EXPORTED int mboxlist_upgrade(int *upgraded)
     r = cyrusdb_foreach(mbdb, "", 0, NULL, _check_rec_cb, &do_upgrade, NULL);
     mboxlist_close();
 
-    if (r != CYRUSDB_DONE) return r;
+    if (r && r != CYRUSDB_DONE) return r;
     else if (!do_upgrade) return 0;
 
     /* create db file names */
```

(though: maybe it should also be `fatal(...)`  rather than `return r`, since the return value is ignored, and if we get an error here something is seriously wrong)

With that patch, I've been able to test and confirm:

* 3.4 with no uniqueid in mbentry, and no RACL entry -> upgrade to this branch:
  success, uniqueid is correctly read from cyrus.header, the overall upgrade succeeds, and the new mailboxes.db has N and I records for user.cassandane
* 3.4 with no uniqueid in mbentry, nor in cyrus.header, and no RACL entry -> upgrade to this branch:
  success, the overall upgrade succeeds, the new mailboxes.db has N and I records for user.cassandane with a brand-new uniqueid, and the cyrus.header has been rewritten with the new uniqueid as well

So it's looking promising!

imap/mboxlist.c
ksmurchison: Logging and skip is now implemented

elliefm:
This is now on the 3.6 branch

PR #4098, opened 2022-05-19
smtpclient: return NULL instead empty SMTP response text

rsto:
Callers assume that the string returned by smtpclient_get_resp_text
actually contains a descriptive response, but the returned string
could have been empty.

Requires https://github.com/cyrusimap/cyrus-imapd/pull/4097 for proper testing.

### DIFF
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -1605,14 +1605,15 @@ static int process_futurerelease(struct caldav_alarm_data *data,
             /* Get the response code and error text.
                We treat anything other than 5xx as a temp failure */
             code = smtpclient_get_resp_code(sm);
+            if (code >= 500) {
+                /* Permanent failure */
+                cancel = 1;
+            }
+            err = NULL;
             if (code) {
                 err = smtpclient_get_resp_text(sm);
-                if (code >= 500) {
-                    /* Permanent failure */
-                    cancel = 1;
-                }
             }
-            else {
+            if (!err) {
                 err = error_message(r);
             }
             syslog(LOG_ERR, "smtpclient_send failed: %s", err);
--- a/imap/smtpclient.c
+++ b/imap/smtpclient.c
@@ -794,7 +794,7 @@ EXPORTED unsigned smtpclient_get_resp_code(smtpclient_t *sm)
 
 EXPORTED const char *smtpclient_get_resp_text(smtpclient_t *sm)
 {
-    return buf_cstring(&sm->resp.text);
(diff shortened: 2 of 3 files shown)


ksmurchison (changes requested):
Looks like caldav_alarm.c:process_futurerelease() needs to be tweaked to support err == NULL

rsto:
Thanks, I had looked at that code section but missed that part.

imap/caldav_alarm.c
```diff
@@ -1605,14 +1605,15 @@ static int process_futurerelease(struct caldav_alarm_data *data,
             /* Get the response code and error text.
                We treat anything other than 5xx as a temp failure */
             code = smtpclient_get_resp_code(sm);
+            if (code >= 500) {
```
ksmurchison, line 1608: Is there a reason that this check was moved from below?

imap/caldav_alarm.c
```diff
@@ -1605,14 +1605,15 @@ static int process_futurerelease(struct caldav_alarm_data *data,
...
+            err = NULL;
             if (code) {
                 err = smtpclient_get_resp_text(sm);
-                if (code >= 500) {
-                    /* Permanent failure */
-                    cancel = 1;
-                }
             }
-            else {
+            if (!err) {
```
ksmurchison, line 1616: This appears to be the only required change

ksmurchison (commented):
Two inline comments

imap/caldav_alarm.c
rsto: the conditional `if (code)` was redundant, as it is always true if `if (code >= 500)`

imap/caldav_alarm.c
rsto: I found it better to split the cancellation conditional from the error reporting. Also, `err` is a reused variable, so I rather want to reset it here before use.

ksmurchison (approved):

dilyanpalauzov:
jmap_mail_submission.c:_emailsubmission_create() contains:
```c
const char *desc = smtpclient_get_resp_text(*sm);
p = strchr(desc, ' ');
if ((p = strstr(desc, "[jmapError:"))) {…}
*set_err = json_pack("{s:s s:s}", "type", "forbiddenToSend", "description", desc);
*set_err = json_pack("{s:s s:s}", "type", err, "description", desc);
```
`desc` can be `NULL`, but the first parameter of `strstr()` may not be `NULL`.

Issue #4085, opened 2022-05-15
icaltimezone_truncate_vtimezone_advanced() stores the address of local variable in the result


dilyanpalauzov:
In [`icaltimezone_truncate_vtimezone_advanced()`](https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/ical_support.c#L2850) ends with `*proleptic = &tombstone;` where `prolepic` is parameter and `tombstone` — a local variable.

GCC 12 reports:
```
imap/ical_support.c: In function ‘icaltimezone_truncate_vtimezone_advanced’:                                                         
imap/ical_support.c:2850:31: warning: storing the address of local variable ‘tombstone’ in ‘*proleptic_277(D)’ [-Wdangling-pointer=] 
 2850 |     if (proleptic) *proleptic = &tombstone;                                                                                     
      |                    ~~~~~~~~~~~^~~~~~~~~~~~                                                                                      
imap/ical_support.c:2371:23: note: ‘tombstone’ declared here                                                                         
 2371 |     struct observance tombstone;                                                                                                
      |                       ^~~~~~~~~                                                                                                 
imap/ical_support.c:2371:23: note: ‘proleptic_277(D)’ declared here                                                                  
```

PR #4099, opened 2022-05-19
squash GCC 12.1.1 complaints

ksmurchison:
Just upgraded OS on my laptop which came with newer GCC

Should be fix for #4085

### DIFF
--- a/imap/ical_support.c
+++ b/imap/ical_support.c
@@ -2359,7 +2359,7 @@ static void icalproperty_get_isstd_isgmt(icalproperty *prop,
 EXPORTED void icaltimezone_truncate_vtimezone_advanced(icalcomponent *vtz,
                                                        icaltimetype *startp, icaltimetype *endp,
                                                        icalarray *obsarray,
-                                                       struct observance **proleptic,
+                                                       struct observance *proleptic,
                                                        icalcomponent **eternal_std,
                                                        icalcomponent **eternal_dst,
                                                        icaltimetype *last_dtstart,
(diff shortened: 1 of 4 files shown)


imap/ical_support.c
```diff
@@ -2368,7 +2368,7 @@ EXPORTED void icaltimezone_truncate_vtimezone_advanced(icalcomponent *vtz,
     icaltimetype start = *startp, end = *endp;
     icalcomponent *comp, *nextc, *tomb_std = NULL, *tomb_day = NULL;
     icalproperty *prop, *proleptic_prop = NULL;
-    struct observance tombstone;
+    static struct observance tombstone;
```
rsto, line 2371: Could we instead require callers to pass a `struct observance*` instead?

rsto (changes requested):

imap/ical_support.c
ksmurchison: Fixed

rsto (approved):
LGTM

imap/ical_support.c
```diff
@@ -2847,7 +2847,9 @@ EXPORTED void icaltimezone_truncate_vtimezone_advanced(icalcomponent *vtz,
         }
     }
 
-    if (proleptic) *proleptic = &tombstone;
+    if (proleptic) {
+        memcpy(proleptic, &tombstone, sizeof(struct observance));
+    }
```
dilyanpalauzov, line 2852: Will `if (proleptic) *proleptic = tombstone` work?

PR #4100, opened 2022-05-20
3.4: ensure mbentries have uniqueids

elliefm:
This improves the safety of upgrades from 3.4 to 3.6+, by making 3.4 fastidious about keeping uniqueid present and up to date (even though 3.4 itself doesn't strictly need it, unlike 3.6+).

Reviewers: please pay close attention to the locking and ordering of the mbentry/mailbox updates.  I think I got it correct, but I'm not totally confident.

* `ctl_cyrusdb -r` will now ensure every mbentry has a uniqueid field at startup.  It will use the one from the cyrus.header if the cyrus.header has one, otherwise it will generate a new uniqueid and save it to both
* `reconstruct` will do the same thing.  This is important for handling the case where an old uniqueid-less mailbox has been restored from backup (though it would also be taken care of by `ctl_cyrusdb -r` at next restart, this can avoid needing a restart)
* `reconstruct -M` ("prefer mailboxes.db") will write the mbentry's uniqueid down to the cyrus.header, if it was missing.  (If both are missing, then same as before: generate a new one, write it to both locations.)
* We now log (LOG_NOTICE) a message any time we open a mailbox and discover that it is missing a uniqueid from either location
* Cassandane tests exercising all cases
* Imported Cyrus::HeaderFile from Fastmail-internal repository, for use by the Cassandane tests

This is complementary to the fixes in #4096 and is part of the solution to #4035 

Once approved and merged, I plan to backport these changes to 3.2, and make new releases from both branches.

I also plan to forward port whatever's possible to forward port for 3.6 and master, once this and #4096 have both independently landed.

### DIFF
--- a/imap/ctl_cyrusdb.c
+++ b/imap/ctl_cyrusdb.c
@@ -129,7 +129,7 @@ static void usage(void)
 static int fixmbox(const mbentry_t *mbentry,
                    void *rock __attribute__((unused)))
 {
-    int r;
+    int r, r2;
 
     /* if MBTYPE_RESERVED, unset it & call mboxlist_delete */
     if (mbentry->mbtype & MBTYPE_RESERVE) {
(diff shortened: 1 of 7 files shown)


elliefm:
D'oh, forgot to add a changes file.  Will do so next week.

imap/ctl_cyrusdb.c
```diff
@@ -173,12 +173,69 @@ static int fixmbox(const mbentry_t *mbentry,
...
+                              __func__, mbentry->name, error_message(r));
+            goto skip_uniqueid;
+        }
+
+        if (!mailbox->uniqueid) {
+            /* yikes, no uniqueid in header either! */
+            mailbox_make_uniqueid(mailbox);
+            xsyslog(LOG_INFO, "mailbox header had no uniqueid, creating one",
+                              "mboxname=<%s> newuniqueid=<%s>",
+                              mbentry->name, mailbox->uniqueid);
```
rsto, line 197: During the mailbox-id rollout we changed the codebase to always use the `mailbox_name()` and `mailbox_uniqueid` functions, rather than accessing these fields directly. Are these function already available in 3.4? If so, we don't need to change this part of the code when/if we pull this patch also into the main branch.

rsto (approved):
LGTM, if possible use the newish mailbox API. +1 for the extensive tests, must have been interesting to write them :)

imap/ctl_cyrusdb.c
elliefm: Two things:

* those accessor functions aren't available in 3.4 anyway, but,
* even if they were, they're ambiguous with respect to whether they return the value from the _mbentry_ or from the _header_, so they're not safe when specifically trying to resolve differences between the two copies. (In practice, they currently return the _mbentry_ copy... but if that implementation detail changed, the behaviour here could break, and we wouldn't notice, because we don't experience the problems this set of fixes addresses.)

rsto: Great, thanks for clarifying.

elliefm:
Realised late last night that my tests do not verify that the `$RUNQ$...` entry is (re)created correctly, and I don't even remember if I looked for that detail while testing manually, before writing the Cassandane tests.  They also don't properly verify that that mbentry and header match after repair (in 3.4, `getmetadata .../uniqueid` only looks at the cyrus.header value, not the mbentry one)

So pulling this back to draft for the moment, so I can extend the tests to verify success more thoroughly, and potentially fix the implementation if it the updates expose a bug...

elliefm:
Updated the tests to check the mbentry and `$RUNQ$` record -- both were fine, no new bug 😌

qyb:
I've built v34/4035-populate-uniqueids (commit 30227af81f822daed757aa50c789a44d027079a6) , `ctl_cyrusdb -r` to upgrade my 3.4 db.  Then switching origin/mboxlist_upgrade_fixes (commit e2c07f90d43c0bacc9fca4d1b9ce1abfe2cb65aa) to try upgrade 3.6. But the db still broken, lost one record

elliefm:
@qyb I appreciate your enthusiasm, but do you mind keeping discussion of the general issue to the issue itself?  It's #4035.  It'll probably take several PRs before this issue is properly resolved, and it'll be a lot easier to keep track of the broader discussion if it happens in one place, not scattered among a bunch of PRs.  Thanks!

I'd like to ask you for more details about your mailboxes.db state before and after the test you just tried, but this PR isn't the place for it.

qyb:
> @qyb I appreciate your enthusiasm, but do you mind keeping discussion of the general issue to the issue itself? It's #4035. It'll probably take several PRs before this issue is properly resolved, and it'll be a lot easier to keep track of the broader discussion if it happens in one place, not scattered among a bunch of PRs. Thanks!
> 
> I'd like to ask you for more details about your mailboxes.db state before and after the test you just tried, but this PR isn't the place for it.

I got it. I became a C programmer and a sendmail administrator in 1999. I hope my testing will help you and the project.

rsto (approved):
LGTM

elliefm:
This is now on the 3.4 and 3.2 branches.

Parts still need forward-porting to master, and then backporting from there to 3.6

PR #4101, opened 2022-05-20
JMAP Mail Submission: sanitize recent change to smtpclient_get_resp_text

rsto:
Recent change f5258901cb66e4b0084da9035309630ff26ca17d introduced a NULL-pointer read which now is fixed. Also, both @ksmurchison and @dilyanpalauzov noticed an unnecessary assignment during code review, so let's just remove it.


ksmurchison (approved):
LGTM

PR #4102, opened 2022-05-20
JMAP Email/query must ignore $seen keyword for JMAP blobs

rsto:
A MIME message uploaded to the `#jmap` and then imported using Email/import still contributed to the $seen counts of the imported email. This could lead to false results in Email queries that made use of the `hasKeyword: $seen` or `xxxInThreadHaveKeyword: $seen` filter criteria.


ksmurchison (approved):
Looks sane

PR #4105, opened 2022-05-24
ci: add github workflow

ajaysusarla:


elliefm:
I've skimmed [some documentation](https://docs.github.com/en/enterprise-server@3.5/actions/migrating-to-github-actions/migrating-from-travis-ci-to-github-actions) quickly, and it looks like this would replace our use of Travis CI with "Github Actions"?  Interesting...

It would be nice to not have to deal with Travis CI anymore, assuming Github Actions turn out to be less of a nuisance

PR #4107, opened 2022-05-24
caldav_db: compare guid of VCALENDAR to detect main event changes

rsto:
The recently added jscal_objs table stores calendar event metadata
for both standalone recurrence instances and main events. Up until
now, for both types of events a change to an existing entry was
detected by comparing the guid of the old and new iCalendar string
of the respective VEVENT component. This introduced a bug when
the main VEVENT component in VCALENDAR remains octet-equal to the
old entry, and only its recurrence overrides got changed.

This patch fixes that by comparing recurrence override by the guid
of the respective VEVENT, but comparing main events by the guid of
the whole VCALENDAR component.


ksmurchison (approved):
LGTM

PR #4108, opened 2022-05-24
OpenSSL 3.0 fixes

ksmurchison:
This PR allows Cyrus to compile with OpenSSL 3.0.

Note that it really only fixes the issues with the MD5 and SHA1 APIs , by using the newer EVP APIs.

The issues with the TLS APIs have been largely avoided at this point by simply commenting out the code that uses legacy APIs.  Someone with a better understanding of DH params and elliptic curves (and the new APIs) will need to fix the rest.

@rsto @elliefm This (rewritten) code compiles both on my laptop with OpenSSL 3.0.2 and my desktop which still has OpenSSL 1.1.1n.  Please try this on your platforms.

### DIFF
--- a/lib/xsha1.h
+++ b/lib/xsha1.h
@@ -43,8 +43,9 @@
 #ifndef LIB_XSHA1_H
 #define LIB_XSHA1_H
 
-#include <assert.h>
 #include <config.h>
+
+#include "lib/assert.h"
 
 #ifdef HAVE_SSL
 
--- a/lib/xsha1.h
+++ b/lib/xsha1.h
@@ -48,12 +48,23 @@
 #ifdef HAVE_SSL
 
 #include <openssl/sha.h>
+#include <openssl/evp.h>
 
 #ifndef SHA1_DIGEST_LENGTH
 #define SHA1_DIGEST_LENGTH (SHA_DIGEST_LENGTH)
 #endif
 
-#define xsha1 SHA1
+#define xsha1(d,l,h)       EVP_Digest(d, l, h, NULL, EVP_sha1(), NULL)
+
+#define SHA1_CTX           EVP_MD_CTX*
+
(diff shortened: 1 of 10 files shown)


lib/xsha1.h
```diff
@@ -45,14 +45,29 @@
 
 #include <config.h>
 
+/*
+ * Use the non-deprecated OpenSSL SHA1 API is its avaiable.
```
elliefm, line 49:
```suggestion
 * Use the non-deprecated OpenSSL SHA1 API if it's available.
```

elliefm:
```
$openssl version
OpenSSL 1.1.1n  15 Mar 2022

$cat /etc/debian_version 
10.12
```

* configures and builds fine
* cassandane mostly passes, and the failures don't look related to this

rsto:
Still builds and tests correctly.

rsto (approved):

elliefm (approved):
Still builds and tests correctly for me too

simo5:
A few questions and comments:
- you are not checking result s on Init, with the EVP API both MD5 and SHA1 can (and are in some systems in FIPS) be blocked and will fail at init time. You should check the return and decide what to do if the desired digest algorithm is not available.
- would it be possible to use this occasion to move to a better digest than SHA-1, say at least SHA-256? (related to the above, there are system where SHA-1 is not allowed).
- I am not familiar with the the TLS code in cyrus imap but loding custom dh params is frowned upon thee days. You should use the well defined groups instead, so losing the ability to load a file with dhparams is probably a good thing.

elliefm (approved):
Looks good -- but two nits/suggestions wrt `assert()`, which has been on my mind lately.  I don't think this needs re-review after those changes, so approving now.

Still builds and tests correctly on my system (which is still the same as it was last time)

PR #4113, opened 2022-05-26
http_dav.c: properly report bad multiget hrefs

ksmurchison:
Non-get-able hrefs were being reported as the target URL o fthe REPORT method and not the actual href in the body of the multiget


rsto (approved):
LGTM

PR #4114, opened 2022-05-26
processimip: ignore iTIP send from and to the same address

ksmurchison:
Fixes a problem that we have seen in the wild with a CANCEL message being redirected back to the ORGANIZER

### DIFF
--- a/imap/itip_support.c
+++ b/imap/itip_support.c
@@ -1007,9 +1007,11 @@ HIDDEN enum sched_deliver_outcome sched_deliver_local(const char *userid,
     icalcomponent_kind kind;
     icalcomponent *comp;
     icalproperty *prop;
-    struct transaction_t txn;
     enum sched_deliver_outcome result = SCHED_DELIVER_ERROR;
     strarray_t recipient_addresses = STRARRAY_INITIALIZER;
+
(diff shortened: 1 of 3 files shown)


imap/itip_support.c
```diff
@@ -1016,6 +1016,12 @@ HIDDEN enum sched_deliver_outcome sched_deliver_local(const char *userid,
     syslog(LOG_DEBUG, "sched_deliver_local(%s, %s, %X)",
            sender, recipient, sparam->flags);
 
+    if (!strcmp(sender, recipient)) {
```
wolfsage, line 1019: This will skip nulling out txn just below on line 1023/1029, and then at the end we do:
```
done:
...
    spool_free_hdrcache(txn.req_hdrs);
...
```

Which I think might now call spool_free_hdrcache(...) on uninitialized data?

rsto (changes requested):
Needs to fix the memory corruption issue addressed by @wolfsage , other than that it looks good to me.

imap/itip_support.c
ksmurchison: Fixed

rsto (approved):

PR #4119, opened 2022-06-02
jmap_util: charset detection must not decrease printable characters  …

rsto:
In some cases, the charset detection library might erroneously
guess a text to be encoded in a single byte encoding such as
ISO-8859-2. The resulting UTF-8 string then trivially contains no
invalid or replacement byte sequences but a large number of
single byte control characters.

To prevent this, this patch adds the heuristic that the amount of
printable characters must not decrease in relation to the total
number of all valid characters.

### DIFF
--- a/cunit/jmap_util.testc
+++ b/cunit/jmap_util.testc
@@ -145,7 +145,6 @@ static void test_decode_to_utf8(void)
         int want_is_encoding_problem;
     };
 
-
     // this is all about "Adélaïde"
 
     struct testcase tcs[] = {{
@@ -189,6 +188,18 @@ static void test_decode_to_utf8(void)
         0.51,
         "Ad""\xef\xbf\xbd""la""\xef\xbf\xbd""de",
         0
+    }, {
+        // Multi-byte UTF-8 with invalid byte sequence:
+        // "Hello,😛🪐world🍎🏓!💾," "\xc3\x28"
+        "SGVsbG8s8J+Ym/CfqpB3b3JsZPCfjY7wn4+TIfCfkr4swyg=",
+        "utf-8",
+        ENCODING_BASE64,
+        0.0,
+        "\x48\x65\x6c\x6c\x6f\x2c\xf0\x9f\x98\x9b\xf0\x9f"
+        "\xaa\x90\x77\x6f\x72\x6c\x64\xf0\x9f\x8d\x8e\xf0"
+        "\x9f\x8f\x93\x21\xf0\x9f\x92\xbe"
+        "," "\xef\xbf\xbd" "(",
+        1
     }, {
         NULL, NULL, ENCODING_UNKNOWN, 0.0, NULL, 0
     }};
--- a/imap/jmap_util.c
+++ b/imap/jmap_util.c
@@ -702,15 +702,7 @@ EXPORTED void jmap_decode_to_utf8(const char *charset, int encoding,
         }
     }
     else if (!charset_id || !strcasecmp("US-ASCII", charset_id)) {
-        int has_cntrl = 0;
-        size_t i;
-        for (i = 0; i < textlen; i++) {
-            if (iscntrl(text[i])) {
-                has_cntrl = 1;
-                break;
-            }
-        }
-        if (has_cntrl) {
+        if (counts.cntrl) {
             /* Could be ISO-2022-JP */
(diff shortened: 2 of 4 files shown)


imap/jmap_util.c
```diff
@@ -751,14 +744,29 @@ EXPORTED void jmap_decode_to_utf8(const char *charset, int encoding,
...
-                        free(guess);
+                            (obj->confidence >= confidence)) {
+                        // libchardet might have guessed a single-byte character encoding,
+                        // which will result in a large number of valid characters and no
+                        // invalid or replacement characters. An indicator that the guess
+                        // was wrong is if the number of printable characters decreased
+                        // in relation to all valid characters, including control chars.
+                        double printable = counts.valid ?
+                            (counts.valid - counts.cntrl) /
+                            ((double) counts.valid) : 0;
```
elliefm, line 755: I might write this as something like:

```
double printable = counts.valid
    ? 1.0 * (counts.valid - counts.cntrl) / counts.valid
    : 0;
```

My thinking is:
* the `?:` being left-aligned in a column make them very easy to scan for and see the relationship between the expressions.  I'd prefer to align them with the `=` too, but I don't think there is enough horizontal space here.
* the true-expression being on its own line rather than wrapped helps too
* we're constrained for horizontal space, and `1.0 *` is fewer chars than `(double)`.  The compiler will probably optimise out the multiplication, and even if it doesn't, the multiplication costs nothing compared to the division we're about to do anyway.
* any term in an expression being `double` is enough to promote every other term to `double`, so it doesn't matter whether we explicitly promote the dividend or the divisor (as long as we promote one of them). Sticking the promotion out the front makes it a little clearer up front that this will be a floating point division.  (In either of our implementations, the integer subtraction in the bracketed subexpression won't be promoted to floating point until after its integer result is computed, but it doesn't matter either way.)

This isn't a "you must change it", just a "these are the things I think about when deciding how to wrap a long line". :)

The `guess_printable` computation below might not work quite the same, because I think the true-expression may be too long to fit on a single line at this level of indentation.  In that case I would wrap the true-expression with the second line indented further and starting from `/ ...`.

cunit/jmap_util.testc
```diff
@@ -189,6 +188,18 @@ static void test_decode_to_utf8(void)
...
+        // "Hello,😛🪐world🍎🏓!💾," "\xc3\x28"
+        "SGVsbG8s8J+Ym/CfqpB3b3JsZPCfjY7wn4+TIfCfkr4swyg=",
+        "utf-8",
+        ENCODING_BASE64,
+        0.0,
+        "\x48\x65\x6c\x6c\x6f\x2c\xf0\x9f\x98\x9b\xf0\x9f"
+        "\xaa\x90\x77\x6f\x72\x6c\x64\xf0\x9f\x8d\x8e\xf0"
+        "\x9f\x8f\x93\x21\xf0\x9f\x92\xbe"
+        "," "\xef\xbf\xbd" "(",
+        1
```
elliefm, line 202: I'm not familiar with these tests....  This chunk at least looks consistent with the neighbouring chunks, though I can't see whether it's good or bad by just looking at it, nor do I understand how it proves the solution.  Do you want me to dig deeper, or do you think it's enough that you understand it and know it's good?  If I was going to dig deeper, I think I would run this test suite under coverage, and see if the new code was covered.

elliefm (approved):
This looks good, though I'm trusting that the test proves it works (and that the test passes).  I can dig deeper into that if you want it verified, but I don't understand it enough to do so visually.

I notice this PR only applies to JMAP.  Will we need to make similar changes elsewhere?  Perhaps IMAP, search indexing, etc.  I imagine you've already thought about it, and the answer was "no", but better to ask than assume.

I have a style (wrapping) comment inline.

imap/jmap_util.c
rsto: Thanks, the line break just was my editor applying its formatting rules. Also applied your other suggested changes.

cunit/jmap_util.testc
rsto: I think it is good enough as regression test as it is. Should the test ever fail we'll see a nicely formatted comparison of strings in a UTF-8 enabled terminal. I also ran the test on a Cyrus instance that does not have the fix applied to prove that it breaks there.

PR #4120, opened 2022-06-03
caldav_alarm: disable alarms for MINUTELY and SECONDLY recurrences

rsto:
calalarmd holds a user lock while processing alarms, which could
take too long for high frequently occurring calendar entries.
This disables alarms for such events altogether. A more lenient
solution can be to optimize the codepath that determines the
`nextcheck` column entry.

### DIFF
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -868,13 +868,15 @@ static int has_alarms(void *data, struct mailbox *mailbox,
 
                 struct icalrecurrencetype rrule = icalproperty_get_rrule(prop);
                 int recur_interval = rrule.interval;
-                char *bypart = "";
+                const char *bypart = "";
                 int disable = 0;
 
                 switch (rrule.freq) {
                 case ICAL_YEARLY_RECURRENCE:
                 case ICAL_MONTHLY_RECURRENCE:
                 case ICAL_WEEKLY_RECURRENCE:
+                    /* Any frequency over weekly is pointless,
+                       but we still want to check BY parts for any insanity */
                     recur_interval *= 7;
 
                     GCC_FALLTHROUGH
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -103,6 +103,7 @@ struct get_alarm_rock {
 };
 
 static struct namespace caldav_alarm_namespace;
+static int min_interval;
 
 EXPORTED int caldav_alarm_init(void)
 {
@@ -113,6 +114,10 @@ EXPORTED int caldav_alarm_init(void)
         syslog(LOG_ERR, "%s", error_message(r));
         fatal(error_message(r), EX_CONFIG);
     }
+
+    min_interval =
+        config_getduration(IMAPOPT_CALENDAR_MINIMUM_ALARM_INTERVAL, 'm');
+    if (min_interval < 0) min_interval = 0;
 
     return sqldb_init();
 }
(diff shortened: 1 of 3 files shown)


imap/caldav_alarm.c
```diff
@@ -844,6 +844,30 @@ static time_t process_alarms(const char *mboxname, uint32_t imap_uid,
...
+
+            struct icalrecurrencetype rrule = icalproperty_get_rrule(prop);
+            if (rrule.freq == ICAL_SECONDLY_RECURRENCE ||
+                rrule.freq == ICAL_MINUTELY_RECURRENCE) {
+                xsyslog(LOG_NOTICE,
+                        "Disabling alarms for high frequence calendar entry",
+                        "freq=<%s> mboxname=<%s> imap_uid=<%d>",
+                        icalrecur_freq_to_string(rrule.freq),
+                        mboxname, imap_uid);
+                return 0;
```
wolfsage, line 866: Ah, a 0 nextcheck will cause the alarmdb row for the event to be deleted (which is *not* the event itself, right?)

Cool

wolfsage (approved):
Seems right ot me

imap/caldav_alarm.c
rsto: Correct

dilyanpalauzov:
This is going to skip `FREQ=MINUTELY;INTERVAL=60`, which is in practice repeating hourly.

rsto:
> This is going to skip FREQ=MINUTELY;INTERVAL=60, which is in practice repeating hourly.

Correct. This patch is not meant to get merged, but is currently required by Fastmail.

rjbs:
@rsto Let's rework this into something like

`minimum_alarm_interval_seconds` ?

ksmurchison:
@rsto please review my re-write.  Once approved we can squash and merge.  Also, does 5m default seem reasonable?

rsto:
@murch this LGTM. I only think that the documentation is a bit misleading: the implementation does not take into the any of the `BYxxx` RRULE rules. I agree that this is overkill, but we could say so. "The BY parameters of the RFC 5545 RRULE are not taken into account when the minimum interval is calculated".

ksmurchison:
@rsto Damn, I hadn't thought about BYSECOND and BYMINUTE.  Do you think we should add support for those?

rsto:
I mean for BYSECOND and BYMINUTE is does look simple enough to implement, but the config option allows for arbitrarily long intervals, so the lower frequency BY rules could take some effort to calculate right.

ksmurchison:
@rsto Added support for BY parts.  Please take another look

imap/caldav_alarm.c
```diff
@@ -809,6 +857,78 @@ static int has_alarms(void *data, struct mailbox *mailbox,
...
+            /* Disable alarms that fire too frequently */
+            for (prop =
+                     icalcomponent_get_first_property(comp, ICAL_RRULE_PROPERTY);
+                 prop;
+                 prop =
+                     icalcomponent_get_next_property(comp, ICAL_RRULE_PROPERTY)) {
+
+                struct icalrecurrencetype rrule = icalproperty_get_rrule(prop);
+                int recur_interval = rrule.interval;
+                char *bypart = "";
```
rsto, line 871: this can be `const`

lib/imapoptions
```diff
@@ -654,6 +654,12 @@ Blank lines and lines beginning with ``#'' are ignored.
    if the CalDAV client hasn't specified a restriction in the creation
    request. */
 
+{ "calendar_minimum_alarm_interval", "5m", DURATION, "UNRELEASED" }
+/* The minimum allowed interval between alarms for a recurring event.
+   Primarily used to suppress alarms for MINUTELY and SECONDLY recurrences.
+   The default is 5 minutes.  The minimum value is 0, which will
+   allow all alarms, regardless of frequency. */
```
rsto, line 661: `, regardless of frequency` seems unnecessary here.

imap/caldav_alarm.c
```diff
@@ -809,6 +857,78 @@ static int has_alarms(void *data, struct mailbox *mailbox,
...
+                int disable = 0;
+
+                switch (rrule.freq) {
+                case ICAL_YEARLY_RECURRENCE:
+                case ICAL_MONTHLY_RECURRENCE:
+                case ICAL_WEEKLY_RECURRENCE:
+                    recur_interval *= 7;
```
rsto, line 878: Maybe just always allow YEARLY, MONTHLY and WEEKLY without check? And state so in the documentation?

rsto (commented):
Just found some nits though that still would improve it. Other than that LGTM.

imap/caldav_alarm.c
ksmurchison: Fixed

lib/imapoptions
ksmurchison: Removed

imap/caldav_alarm.c
ksmurchison: I think we still want to check the BY parts for these frequencies.  FREQ=WEEKLY;BYSECOND=1,2,3,4,5 is ridiculous but possible

ksmurchison:
@rsto Looking at this code again, I wonder if it makes sense to only do the interval check if the recurring component contains an alarm.

PR #4125, opened 2022-06-07
Use [LIMIT] response code for IMAP APPEND where appropriate

ksmurchison:
This gives a more machine-readable response when we fail due to too many copies of the same message


elliefm (approved):

rsto (approved):
LGTM

PR #4126, opened 2022-06-08
JMAPComposite: remove redundant test suite

elliefm:
It breaks because it includes JMAP Calendars tests without properly setting up the environment they need, so they all fail.

But Rik tells me this was made as a proof of concept that wasn't supposed to stay around, so instead of fixing it up I'm just removing it.


cassandane/Cassandane/Cyrus/JMAPComposite.pm
```diff
@@ -1,386 +0,0 @@
...
-use URI;
-use URI::Escape;
-
-use lib '.';
-use base qw(Cassandane::Cyrus::TestCase);
-use Cassandane::Util::Log;
-
-use charnames ':full';
-
-sub new
```
rsto, line 61: Does the JMAPComposite even need a `new` function? I haven't dug into the recent Cassandane refactor, but don't we still have distinct JMAPTalk instances per JMAPEmail, JMAPCalendars, ... test suites?

ksmurchison (approved):
Removing the file seems right to me

cassandane/Cassandane/Cyrus/JMAPComposite.pm
rsto: I think I should have first had my morning coffee before doing code review. Sorry for the noise.

rsto (approved):

PR #4127, opened 2022-06-08
mbexamine: tighten up error handling

elliefm:
This is part of some work that @rjbs requested.   I'd like to get it merged now because otherwise it'll start to conflict with the long-options work that I'm also doing.

This:
* reports any errors that occur on stderr
* treats invocations that don't find any work to do (i.e. when there are no matches for the provided mailbox pattern) as failures rather than successes

We don't have any existing tests for mbexamine, and I haven't bothered to write any.  I expect this interface to change further when we do the big tool audit/update, so in my mind tests can wait.  Let me know if you disagree!

### DIFF
--- a/imap/mbexamine.c
+++ b/imap/mbexamine.c
@@ -111,6 +111,7 @@ int main(int argc, char **argv)
     char buf[MAX_MAILBOX_PATH+1];
     char *alt_config = NULL;
     int (*cb)(struct findall_data *, void *) = &do_examine;
+    int ok_count = 0;
 
     while ((opt = getopt(argc, argv, "C:u:s:qc")) != EOF) {
         switch (opt) {
@@ -155,16 +156,31 @@ int main(int argc, char **argv)
 
     if (optind == argc) {
         strlcpy(buf, "*", sizeof(buf));
-        mboxlist_findall(&mbexamine_namespace, buf, 1, 0, 0, cb, NULL);
+        r = mboxlist_findall(&mbexamine_namespace, buf, 1, 0, 0, cb, &ok_count);
+        if (r) {
+            fprintf(stderr, "%s: %s\n", buf, error_message(r));
+        }
     }
 
     for (i = optind; i < argc; i++) {
-        mboxlist_findall(&mbexamine_namespace, argv[i], 1, 0, 0, cb, NULL);
+        r = mboxlist_findall(&mbexamine_namespace, argv[i], 1, 0, 0, cb, &ok_count);
+        if (r) {
+            fprintf(stderr, "%s: %s\n", argv[i], error_message(r));
+        }
     }
 
     cyrus_done();
 
-    exit(0);
+    if (r && !ok_count) {
+        return r;
+    }
+    else if (!ok_count) {
+        fprintf(stderr, "No matching mailboxes found\n");
+        return EX_NOUSER; /* XXX i guess? */
+    }
+    else {
+        return 0;
+    }
 }
 
 static void usage(void)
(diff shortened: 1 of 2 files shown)


imap/mbexamine.c
```diff
@@ -111,6 +111,7 @@ int main(int argc, char **argv)
     char buf[MAX_MAILBOX_PATH+1];
     char *alt_config = NULL;
     int (*cb)(struct findall_data *, void *) = &do_examine;
+    int ok_count = 0;
```
rsto, line 114: Something bugs me with that variable name. I'm sure if I read that code again in 2 years, I will have to follow the `ok_count` variable through the callbacks to understand how it is counted. The error message at the end kind of suggests it could be named `matched_mb_count` or `nmatched_mb` or the like? Off the top of my head I don't have a good suggestion.

imap/mbexamine.c
```diff
@@ -351,19 +368,22 @@ static int do_examine(struct findall_data *data, void *rock __attribute__((unuse
 
     mailbox_close(&mailbox);
 
+    if (!r && ok_count) (*ok_count) ++;
```
rsto, line 368: I think that NULL pointer check on ok_count can be omitted. Neither callback is called with a NULL rock, so better assert(ok_count) at the start of the function?

rsto (commented):
Definitely an improvement, but I have two nits where I would prefer to first read your reply. Would approve as-is, if you do not agree with my comments, though.

imap/mbexamine.c
elliefm: This is the count of things we tried and which were ok (_not_ the count of things we tried) -- the callbacks increment it on success, not on entry.  It's important for the later step, where we determine the final exit status based on a combination of "whether we successfully did anything" and "whether there were errors".

We want to be able to:

* exit success if we successfully did something (even if something else failed)
* exit with an error if we did something but it failed
* exit with a "we didn't do anything" error if we did nothing at all (which would otherwise look like a success, because of no error)

I'm not wedded to the name, but I don't have a better suggestion.

imap/mbexamine.c
elliefm: This is just a good habit -- if your function takes an output pointer, safely allow it to be NULL in case the caller doesn't care.

We have loads of instances where over time we've re-used some prexisting callback in some slightly new context.  Will the same thing happen here?  Probably not.  But if it does, the callback already provides a sane interface, even though its current usage doesn't require it.

An assert would provide the opposite behaviour: mandating the caller either care, or modify the function (and potentially all the other callers) into this shape if they don't.

This isn't a hot path, so the sane interface and good discipline is worth more than the one instruction we'd save by removing it.

(And if we replaced it with an assert, that would be replacing a trivial compare-to-zero with an expensive branch and possible function call, because our asserts are never compiled out!)

elliefm:
@rsto,

> Definitely an improvement, but I have two nits where I would prefer to first read your reply. Would approve as-is, if you do not agree with my comments, though.

I replied to both of your comments a while ago, but haven't heard back.  Can we move this forward please?

imap/mbexamine.c
rsto: thanks for the explanation, I agree

imap/mbexamine.c
rsto: thanks, got it

rsto:
Sorry for the late reply. This all looks good to me.

rsto (approved):
LGTM

elliefm:
Thanks!

imap/mbexamine.c
dilyanpalauzov: Here `r` is always `0`, so `if (!r && …)` is tautology.  Likewise `return r;` on the next line and at the end of `do_quota()`.

PR #4128, opened 2022-06-08
Email/set should ignore expunged messages

ksmurchison:
This solves a problem where a canceled scheduled send message that has been moved back to the Drafts mailbox can't be destroyed.
This happens because the already-expunged message in the Scheduled mailbox is found and we prohibit direct Email/Set on anything in Scheduled.

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -10908,6 +10908,12 @@ static int _email_mboxrecs_read_cb(const conv_guidrec_t *rec, void *_rock)
     /* don't process emails that have this email attached! */
     if (rec->part) goto done;
 
+    /* don't process emails that have been expunged */
+    if ((rec->system_flags & FLAG_DELETED) ||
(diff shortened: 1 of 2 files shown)


imap/jmap_mail.c
```diff
@@ -10908,6 +10908,9 @@ static int _email_mboxrecs_read_cb(const conv_guidrec_t *rec, void *_rock)
     /* don't process emails that have this email attached! */
     if (rec->part) goto done;
 
+    /* don't process emails that have been expunged */
+    if (rec->internal_flags & FLAG_INTERNAL_EXPUNGED) goto done;
```
rsto, line 10912: Also ignore if FLAG_DELETED is set in the system_flags?

rsto (commented):
I think we also want to check the system_flags.

imap/jmap_mail.c
ksmurchison: Agreed

rsto (approved):
LGTM

PR #4130, opened 2022-06-10
Remove iTIP messages on CalendarEvent/set{update,destroy}

ksmurchison:
This blindly removes all iTIP messages having the matching iCalendar UID.  The thought is that if the user updates the calendar via JMAP they probably don't need their CalDAV client to see any of the iTIP for that UID, even if they correspond to single instances.

An alternative to deleting these on CalendarEvent/set would be to just set an expiration time on the Scheudling Inbox (e.g. 1-2 weeks)

### DIFF
--- a/cassandane/Cassandane/Cyrus/Sieve.pm
+++ b/cassandane/Cassandane/Cyrus/Sieve.pm
@@ -4243,6 +4243,156 @@ EOF
     $self->assert_str_equals('2021-09-24T14:00:00', $events->[0]{start});
 }
 
+sub test_remove_itip_on_jmap_update
+    :needs_component_sieve :needs_component_httpd :min_version_3_5
+{
+    my ($self) = @_;
+
+    my $jmap = $self->{jmap};
+
+    xlog $self, "Create calendar user";
+    my $CalDAV = $self->{caldav};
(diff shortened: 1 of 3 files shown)


rsto:
I am not sure the semantics are correct. In JMAP, multiple CalendarEvents can have the same uid, if all of them are separate recurrence instances. The "id" of each CalendarEvent will be unique.

Also: a single row in ical_objs might correspond to multiple rows in the jscal_objs, where each of the rows in the latter have a distinct, non-empty ical_recurid value.

Maybe it is like this:
- if the CalendarEvent does not have a recurrenceId, delete all entries in the Inbox
- if the CalendarEvent does have a recurrenceId and
  -  it is not a standalone instance, then delete all entries in the Inbox
-    it is a standalone instance, only delete the entries in Inbox having that UID and RECURRENCE-ID

A CalendarEvent is a standalone instance, if `caldav_foreach_jscal` with both the `ical_uid` and `ical_recurid` fields set in the  `caldav_jscal_filter` finds a non-tombstone record

ksmurchison:
The standalone instance problem should be fixed.

rsto (approved):
LGTM

Issue #123, opened 2000-08-07
append during select corrupts seen state


brong:
From: [Larry Greenfield](leg+cyrus@andrew.cmu.edu)
Bugzilla-Id: 110
Version: 2.0.x
Owner: [Larry Greenfield](leg+cyrus@andrew.cmu.edu)

PR #4131, opened 2022-06-13
Strip event color from incoming/outgoing iTIP

ksmurchison:
Pretty self-explanatory

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1269,8 +1269,6 @@ nodist_imap_httpd_SOURCES = \
 imap_httpd_SOURCES = \
 	imap/caldav_util.c \
 	imap/caldav_util.h \
-	imap/css3_color.c \
-	imap/css3_color.h \
 	imap/http_admin.c \
 	imap/http_applepush.c \
 	imap/http_caldav.c \
--- a/Makefile.am
+++ b/Makefile.am
@@ -1009,6 +1009,8 @@ imap_libcyrus_imap_la_SOURCES = \
 	imap/conversations.h \
 	imap/convert_code.c \
 	imap/convert_code.h \
+	imap/css3_color.c \
+	imap/css3_color.h \
 	imap/dav_db.c \
 	imap/dav_db.h \
 	imap/dlist.c \
--- a/cassandane/Cassandane/Cyrus/Sieve.pm
+++ b/cassandane/Cassandane/Cyrus/Sieve.pm
@@ -5154,6 +5154,82 @@ EOF
                          $events->[0]{recurrenceOverrides}{'2021-09-23T15:30:00'});
 }
 
+sub test_imip_strip_personal_data
+    :needs_component_sieve :needs_component_httpd :min_version_3_7
+{
+    my ($self) = @_;
+
+    my $IMAP = $self->{store}->get_client();
+    $self->{store}->_select();
+    $self->assert_num_equals(1, $IMAP->uid());
+    $self->{store}->set_fetch_attributes(qw(uid flags));
+
+    xlog $self, "Create calendar user";
+    my $CalDAV = $self->{caldav};
+    my $CalendarId = 'Default';
+    my $uuid = "6de280c9-edff-4019-8ebd-cfebc73f8201";
+
+    xlog $self, "Install a sieve script to process iMIP";
+    $self->{instance}->install_sieve_script(<<EOF
+require ["body", "variables", "imap4flags", "vnd.cyrus.imip"];
+if body :content "text/calendar" :contains "\nMETHOD:" {
+    processimip :outcome "outcome";
+    if string "\${outcome}" "updated" {
+        setflag "\\\\Flagged";
+    }
+}
+EOF
+    );
+
+    my $imip = <<EOF;
+Date: Thu, 23 Sep 2021 09:06:18 -0400
+From: Foo <foo\@example.net>
+To: Cassandane <cassandane\@example.com>
+Message-ID: <$uuid-0\@example.net>
+Content-Type: text/calendar; method=REQUEST; component=VEVENT
+X-Cassandane-Unique: $uuid-0
+
+BEGIN:VCALENDAR
+VERSION:2.0
+PRODID:-//Apple Inc.//Mac OS X 10.10.4//EN
+METHOD:REQUEST
+BEGIN:VEVENT
+CREATED:20210923T034327Z
+UID:$uuid
+DTEND;TZID=America/New_York:20210923T183000
+TRANSP:OPAQUE
+COLOR:red
+SUMMARY:An Event
+DTSTART;TZID=America/New_York:20210923T153000
+DTSTAMP:20210923T034327Z
+SEQUENCE:0
+RRULE:FREQ=WEEKLY
+ORGANIZER;CN=Test User:MAILTO:foo\@example.net
+ATTENDEE;CN=Test User;PARTSTAT=ACCEPTED;RSVP=TRUE:MAILTO:foo\@example.net
+ATTENDEE;PARTSTAT=NEEDS-ACTION;RSVP=TRUE:MAILTO:cassandane\@example.com
+BEGIN:VALARM
+UID:myalarm
+TRIGGER;RELATED=START:PT0S
+ACTION:DISPLAY
+DESCRIPTION:CYR-140
+END:VALARM
+END:VEVENT
+END:VCALENDAR
+EOF
+
+    xlog $self, "Deliver iMIP invite";
+    my $msg = Cassandane::Message->new(raw => $imip);
+    $self->{instance}->deliver($msg);
+
+    xlog $self, "Get the event and verify that personal data has been stripped";
+    my $events = $CalDAV->GetEvents($CalendarId);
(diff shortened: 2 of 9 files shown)


Makefile.am
```diff
@@ -1009,6 +1009,8 @@ imap_libcyrus_imap_la_SOURCES = \
 	imap/conversations.h \
 	imap/convert_code.c \
 	imap/convert_code.h \
+	imap/css3_color.c \
+	imap/css3_color.h \
```
dilyanpalauzov, line 1013: Since imap/css3_color.h and imap/css3_color.c are linked into libcyrus_imap, `httpd`links with libcyrus_imap.so, the explicit dependency of css3_color.[ch] to `httpd` can be removed.

In any case, there is no point to mention the file `css3_color.h` twice in (different) `_SOURCE` variables.

imap/css3_color.h
```diff
@@ -46,4 +46,6 @@
 
 extern const char *css3_color_hex_to_name(const char *hexstr);
 
+extern int is_css3_color(const char *s);
+
```
dilyanpalauzov, line 50: What is the purpose of `extern` here?

imap/itip_support.c
```diff
@@ -1082,20 +1082,35 @@ HIDDEN enum sched_deliver_outcome sched_deliver_local(const char *userid,
...
+            case ICAL_COLOR_PROPERTY:
+            case ICAL_TRANSP_PROPERTY:
+                icalcomponent_remove_property(comp, prop);
+                icalproperty_free(prop);
+                break;
+
+            default:
+                break;
+            }
         }
```
dilyanpalauzov, line 1114: This is a code duplication, in regards to the changes in http_caldav_sched.c:clean_component().

dilyanpalauzov (commented):
I object stripping the COLOR property.   Or rather I generate iTIP messages outside of Cyrus, send them per email with `Content-Type: text/calendar; charset=UTF-8; method=PUBLISH; component=VEVENT` and include a `COLOR` property (depending on the category of the event.  In my use case, the CATEGORY itself is not a color, but a fixed set of human meaningful categories).  Evolution does display the colour, when rendering the iTIP message,  and I the render the title’s background of the event using the COLOR value (using black or white as foreground, depending on which has better contrast).  Once imported, DAVx⁵ permits the user to ignore the per event colors.

I do not see, why  shall TRANSP be cut, but have no compelling reason against removing it.

If something is to be removed, then the default values are best candidates:
* SEQUENCE, if it is 0
* TRANSP, if it is OPAQUE (I do it actually!)
* CLASS, if it is PUBLIC
* CALSCALE, if it is GREGORIAN
* PRIORITY, if it is 0.

ksmurchison:
@dilyanpalauzov The reason to remove VALARM, TRANSP, and COLOR is they are considered per-user meta-data for an event and shouldn't be sent and shouldn't be received (MUST NOT in the case of VALARM, especially with calendar spam).  I think your use-case is not the common case in that you're sending an event to yourself (why not just HTTP PUT it directly on the calendar?).  I certainly don't want the sender of an event setting/updating the alarms, transparency, or color of events on my calendar.

Makefile.am
ksmurchison: Fixed

imap/css3_color.h
ksmurchison: Fixed

imap/ical_support.c
```diff
@@ -2957,4 +2958,20 @@ EXPORTED void icalcomponent_add_required_timezones(icalcomponent *ical)
...
+EXPORTED int ical_categories_is_color(icalproperty *cprop)
+{
+    const char *categories = icalproperty_get_categories(cprop);
+    int i;
+    
+
+    if (categories[0] == '#') {
+        /* Is this #RRGGBB */
+        for (i = 1; i <= 6 && isxdigit(categories[i]); i++);
+
```
dilyanpalauzov, line 2970: If there is a category `#A`, then `categories[4]` does not exist, but it will be read.

dilyanpalauzov: No. it will not be read.  But `#A` is not a colour.

imap/ical_support.h
```diff
@@ -187,6 +187,8 @@ extern void icaltimezone_truncate_vtimezone_advanced(icalcomponent *vtz,
                                                      icaltimetype *last_dtstart,
                                                      int ms_compatible);
 
+extern int ical_categories_is_color(icalproperty *cprop);
```
dilyanpalauzov, line 190: What is the purpose of `extern` here?

dilyanpalauzov:
When I receive per email an iMIP invitation with included VALARM, Evolution asks me, if I want to accept also the reminder.  That said, handling of VALARM in iMIP is up to the CUA.  The same approach could be used for COLOR and TRANSP, however in real life I have never received invitations with COLOR (unless those I send myself) and TRANSP:TRANSPARENT.

> why not just HTTP PUT it directly on the calendar?

I send the METHOD=PUBLISH messages (with COLOR) over a mailing list to a virtually unlimited number of recipients and providers.

For the record, https://calendar.google.com/calendar/ical/aegee.eu_v5mn651imeqpvs87v4ln7hr1f4%40group.calendar.google.com/public/basic.ics is a public calendar and it contains (one) TRANS:TRANSPARENT event.  Whether a transparent event is distributed as webcal feed, or iMIP does not make a big difference.

imap/ical_support.c
rsto: It correctly returns false for `#A`. But it mistakenly returns true if the category value is just `#`. I think the return condition should read `i == 7 && categories[6]...` to make sure it's a legit 6-digit hex color.

rsto (changes requested):
I think the return condition for `is_color` is too lenient. At least it should only accept `#` followed by at least one digit, but maybe it's better to only accept full 6-digit colors.

imap/ical_support.c
elliefm: For whatever it's worth, CSS also supports three-digit hex colours (`#RGB`) -- as a shorthand for colours who values are pairs of identical digits (i.e. `#123` is shorthand for `#112233`), which I think is different from what you would get if you treated it as a 12 bit colour and multiplied it out to 24 bits (`#102030`, right?).  Users/developers who use a lot of CSS might expect the shorthand to work here too.

But `css3_color_hex_to_name()` only supports six-digit (`#RRGGBB`) or eight-digit (`#RRGGBBAA` -- of which it drops the alpha channel) anyway, so I think this function should also require (at least) six digits to consider it a valid colour.

If we wanted to accept the three-digit CSS shorthand, I guess we'd need to extend `css3_color_hex_to_name()` too -- and be careful to do so by duplicating digits rather than just expanding 12-bit to 24-bit.  I don't think it's worth the effort, unless we see it in the wild?

imap/ical_support.h
elliefm: The function `ical_categories_is_color()` is only declared here (or anywhere that includes this file), its definition is elsewhere (i.e. in imap/ical_support.c).  It's usual to use the `extern` keyword like this in header files.

Historically, `extern` was also the default, so old code often omits it, but it's good practice to include it.

dilyanpalauzov: `extern` is extra text to read, which does nothing.  It is obvious, that the definition is not on the line after the prototype.

Extra code, which does nothing, complicates understanding of the code.  Other examples of extra code, which does nothing, are defined, but unused functions.

imap/itip_support.c
ksmurchison: This code has been consolidated

imap/ical_support.c
ksmurchison: Length of the color (7) is now checked

Rik says we won't worry about #123 for now

rsto (changes requested):
Logic looks good to me, just would like to see a test that actually assert CATEGORIES

imap/ical_support.c
```diff
@@ -2957,4 +2958,20 @@ EXPORTED void icalcomponent_add_required_timezones(icalcomponent *ical)
     ptrarray_fini(&tzs);
 }
 
+EXPORTED int ical_categories_is_color(icalproperty *cprop)
+{
+    const char *categories = icalproperty_get_categories(cprop);
+    int i;
+
```
dilyanpalauzov, line 2965: This line is four spaces an a row.

cassandane/Cassandane/Cyrus/Sieve.pm
```diff
@@ -5154,6 +5154,82 @@ EOF
...
+VERSION:2.0
+PRODID:-//Apple Inc.//Mac OS X 10.10.4//EN
+METHOD:REQUEST
+BEGIN:VEVENT
+CREATED:20210923T034327Z
+UID:$uuid
+DTEND;TZID=America/New_York:20210923T183000
+TRANSP:OPAQUE
+COLOR:red
+SUMMARY:An Event
```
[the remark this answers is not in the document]
ksmurchison: Done.

imap/ical_support.c
ksmurchison: Fixed

rsto (approved):
LGTM

PR #4132, opened 2022-06-13
Allow compilation with --disable-http

ksmurchison:


elliefm:
I tested this branch without changing what libraries I had installed, but just by adding `--disable-http --disable-calalarmd --disable-jmap` to my usual configure argument set.

It seems to have compiled the source files okay, but when it gets to linking there are loads of linker errors.  The main culprit seems to be that we're linking against libguesstz but not libical, and libguesstz depends on libical being linked in, and grumbles when it's not.  There might be other linker failures too, but if so I'm overlooking them in the libguesstz noise, which there's a lot of.

I wonder how much effort we want to spend here.  As I understand it, our longer term plan is to make a bunch of features that require httpd (JMAP, et al) non-optional anyway.  So maybe instead of trying to fix --disable-http (again), we just rip that option out now, and always build httpd.

For whatever it's worth, 3.6 already builds okay with --disable-http etc, so I don't need this fixed for backporting to 3.6.  Whatever we do here is forward-facing only.

rsto:
@elliefm can we get rid of this?

PR #4136, opened 2022-06-15
iMIP: Add support for METHOD:PUBLISH (with empty ORGANIZER)

ksmurchison:


rsto (commented):
When I delete the (empty) ORGANIZER property from the iTIP, it gets rejected. Is that on purpose? I wonder what's the practical difference between no ORGANIZER, or an empty-valued one.
I assume that's because of the iTIP standard, but might it make sense to be more lenient here?

ksmurchison:
@rsto libical's restriction checks will cause PUBLISH without ORGANIZER to fail on parsing, unless we override it.

rsto:
OK, not worth the effort then. Thanks.

rsto (approved):
LGTM

PR #4139, opened 2022-06-17
no uniqueid fix for murder frontends (3.4)

elliefm:
uniqueid is not one of the fields synchronised by mupdate, which I think means the mbentries for remote mailboxes (i.e. on murder frontends) won't and can't have uniqueids.

This PR builds upon the work from #4100, such that it won't complain about nor try to "fix" missing uniqueids for remote mailboxes.

This PR is for 3.4 only.  I think master/3.6 will need a similar shape of change in `mboxlist.c:_foreach_cb()` following #4096, but I have not thought about the ramifications there yet.

No tests (yet?), cause I haven't figured out how I would prove to myself that it works.  Am open to ideas!

### DIFF
--- a/imap/ctl_cyrusdb.c
+++ b/imap/ctl_cyrusdb.c
@@ -173,8 +173,8 @@ static int fixmbox(const mbentry_t *mbentry,
                mbentry->name, cyrusdb_strerror(r));
     }
 
-    /* make sure every mbentry has a uniqueid!  */
-    if (!mbentry->uniqueid) {
+    /* make sure every local mbentry has a uniqueid!  */
+    if (!mbentry->uniqueid && mbentry_is_local_mailbox(mbentry)) {
         struct mailbox *mailbox = NULL;
         struct mboxlock *namespacelock = NULL;
         mbentry_t *copy = NULL;
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -1042,7 +1042,8 @@ static int mailbox_open_advanced(const char *name,
         goto done;
     }
 
-    if (!mbentry->uniqueid) {
+    /* XXX can we even get here for remote mbentries? */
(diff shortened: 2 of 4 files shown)


ksmurchison (approved):
This looks reasonable, assuming that entries on frontends actually have the MBTYPE_REMOTE flag set.

If not, a lot of Cyrus code uses something like the following for detecting a frontend config:

`    if (config_mupdate_server && !config_getstring(IMAPOPT_PROXYSERVERS)) {`

elliefm:
> assuming that entries on frontends actually have the MBTYPE_REMOTE flag set

Hmm, I expect they do, but I don't know that they do.  I could spin something up and have a look, but even if I did and they did, I wouldn't feel confident that they always or reliably do.

I've added a belt-and-suspenders `mbentry_is_local_mailbox()` function to mboxlist (and a corresponding `mbentry_is_remote_mailbox()` macro that just negates the result), and updated the existing changes to use the new function.

What do you think, any edge cases I've missed?

If you have any ideas for specific cases you'd like to see tested, please let me know.  I would like to have tests for this, but I keep drawing a blank on what tests to actually write.

ksmurchison (approved):
These changes look good.
I think the only way to test this is to setup a Murder with older Cyrus and then restart it with this new code to make sure that the frontend(s) start up cleanly.

elliefm:
Mmm yeah, which Cassandane can't do anyway.

I think I've convinced myself that the main consequence, if this _doesn't_ work for some reason, is logging a bunch of errors about not being able to open mailboxes.  Which could be annoying, but isn't really a substantial problem, and will be easy enough to track down the cause once someone reports it.

So instead of going down another rabbit hole trying to make testing infrastructure for this, I think I'm just gonna roll with it as is so we can move forward.

PR #4140, opened 2022-06-17
fix assert() definition

### LINKED ISSUE #4123: our assert() is not compatible with the C one, and breaks in edge cases
Reported here as an incompatibility with perl 5.36: https://cyrus.topicbox.com/groups/info/T4dbb95a05bf1581e-Mdc484c415098530c67b67f26

We need to either redefine our `assert()` in a way that's compatible with the C one (losing the `{}` and the `if()`), or stop trying to use it as a general replacement for the C one.

elliefm:
This redefines our `assert()` macro to evaluate as an expression, rather than as a statement.  This makes it syntactically compatible with the C standard one, which means sub-components we link in that expect to have a C standard assert() available do not choke on ours being different.  As far as I know, this is mostly just perl, when building our XS modules.

While in there, I cleaned up an old `#ifdef __STDC__` check, whose else path would have failed to compile anyway, had it ever been hit.

Also had to fix a bunch of miscellaneous lines without semicolons, which the bad macro definition had previously allowed. 😅 

Fixes #4123, I think -- will need someone with perl 5.36 to test and confirm this please.

Once merged, this will be backported to 3.6 and 3.4.

### DIFF
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -4492,7 +4492,7 @@ EXPORTED int mailbox_append_index_record(struct mailbox *mailbox,
     assert(mailbox_index_islocked(mailbox, 1));
 
     /* Append MUST be a higher UID than any we've yet seen */
-    assert(record->uid > mailbox->i.last_uid)
+    assert(record->uid > mailbox->i.last_uid);
 
     /* Append MUST have a message with data */
     assert(record->size);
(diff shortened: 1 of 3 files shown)


robn:
Confirmed to build against 5.36. Patch looks perfectly sensible to my eye. Nice work!

robn (approved):

elliefm:
This has been cherry-picked to the 3.6 and 3.4 branches

PR #4142, opened 2022-06-21
Ignore orphaned dav.db entries in DELETED mailboxes

ksmurchison:
FM support reported a weird issue where an event in a deleted calendar would be returned, and somehow updated after the fact


rsto (approved):

rsto:
Squashed a few commits.

PR #4145, opened 2022-06-24
tiny-tests: ignore Emacs and Vim swap files

rsto:


cassandane/Cassandane/Tiny/Loader.pm
```diff
@@ -15,7 +15,7 @@ sub import {
     Carp::confess(qq{can't find path "$path" for loading tests; Cassandane expects to be run from the ./cyrus-imapd/cassandane directory"});
   }
 
-  my @tests = `find $path -type f`;
+  my @tests = `find $path -type f \! -name "*~" \! -name ".*.swp"`;
```
elliefm, line 18: While we're in here, can we make it ignore any file starting with `.`, rather than only vim swap-files specifically?  I can't remember if `-name pattern` matches pattern anywhere in the filename, or exactly matches the whole filename, but I think maybe something like `\! -path "*/.*"` (i.e. using `-path pattern` so we can match on the `/`) might do the trick?

rsto: I made it ignore any dot file.

elliefm (approved):

PR #4146, opened 2022-06-24
use the correct pattern when altnamespace is on

qyb:
Signed-off-by: qyb <qiuyingbo@gmail.com>

more info: https://cyrus.topicbox.com/groups/info/T19bc8d3e0b073fec-M28948f2e1ef90f0c646d8c6f/cannot-get-shared-mailbox-info-in-jmap-session

### DIFF
--- a/imap/jmap_api.c
+++ b/imap/jmap_api.c
@@ -1018,8 +1018,15 @@ HIDDEN void jmap_accounts(json_t *accounts, json_t *primary_accounts)
 {
     /* Find shared accounts */
     strarray_t patterns = STRARRAY_INITIALIZER;
-    char *userpat = xstrdup("user.*");
-    userpat[4] = jmap_namespace.hier_sep;
+    char *userpat = NULL;
+    if (jmap_namespace.isalt) {
+        userpat = xstrdup("Other Users.*");
+        userpat[11] = jmap_namespace.hier_sep;
+    }
+    else {
+        userpat = xstrdup("user.*");
+        userpat[4] = jmap_namespace.hier_sep;
+    }
     strarray_append(&patterns, userpat);
     struct findaccounts_rock rock = {
         BUF_INITIALIZER, 0, accounts, httpd_userid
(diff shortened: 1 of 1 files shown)


imap/jmap_api.c
```diff
@@ -1018,8 +1018,15 @@ HIDDEN void jmap_accounts(json_t *accounts, json_t *primary_accounts)
 {
     /* Find shared accounts */
     strarray_t patterns = STRARRAY_INITIALIZER;
-    char *userpat = xstrdup("user.*");
-    userpat[4] = jmap_namespace.hier_sep;
+    char *userpat = NULL;
+    if (jmap_namespace.isalt) {
+        userpat = xstrdup("Other Users.*");
+        userpat[11] = jmap_namespace.hier_sep;
+    }
```
elliefm, line 1025: The "Other Users" string comes from the imapd.conf `userprefix` setting, so this should use `config_getstring(IMAPOPT_USERPREFIX)` rather than hard-coding "Other Users".

You might need to use a `struct buf` to assemble the parts, since if you just xstrdup it like the current code it'll be missing the ".*" at the end... maybe something like (untested):

```
if (jmap_namespace.isalt) {
    buf_printf(&buf, "%s%c*", config_getstring(IMAPOPT_USERPREFIX), jmap_namespace.hier_sep);
    userpat = buf_release(&buf);
}
```

That buf_printf line probably needs to be wrapped, but I'm typing in a github comment box, not a code editor 😅

elliefm (changes requested):

elliefm:
@rsto At a glance, looks like master and 3.6 will need this too.  Once this passes review and is merged to 3.4, I can just cherry-pick and push it to those branches too -- but let me know if you'd rather it go through another review first.

imap/jmap_api.c
qyb: I'd like use `xstrndup` to avoid buf_release operation, and I must compute strlen(config_getstring(IMAPOPT_USERPREFIX)) 
 here to replace `jmap_namespace.hier_sep`.

qyb: After comparing their implementation, I'll change `xstrndup` to `struct buf`.

rsto (approved):
Looks good to me. @elliefm I suggest we squash this PR into a single commit? I'm fine with just cherry-picking it across the release branches.

elliefm:
Yeah, I agree -- @qyb can you please squash this into a single commit?

I can just do it myself when I merge it if you prefer, but if you do it then your authorship will remain intact, which might be important to you.

qyb:
I'll request a new PR

PR #4152, opened 2022-06-28
Allow existing oversized vCards to be updated

ksmurchison:


imap/jmap_backup.c
```diff
@@ -809,7 +809,7 @@ static int restore_addressbook_cb(const mbentry_t *mbentry, void *rock)
         /* Store the group vCard of recreated contacts */
         r = carddav_store(*mailboxp, crock->group_vcard, NULL, 0, NULL, NULL, 
                           rrock->req->accountid, rrock->req->authstate,
-                          /*ignorequota*/ 0);
+                          /*ignorequota*/ 0, /*oldsize*/ INT_MAX);
```
rsto, line 812: I don't expect Cyrus to run properly on 32-bit systems anyway, but since the old size type is `uint32_t` I suggest to also `UINT32_MAX` here.

imap/http_carddav.c
```diff
@@ -622,6 +615,15 @@ static int carddav_store_resource(struct transaction_t *txn,
         }
     }
 
+    /* Check size of vCard (allow existing oversized cards to be updated) */
+    struct buf *buf = vcard_as_buf(vcard);
+    if ((buf_len(buf) > (size_t) vcard_max_size) &&
+        !(oldrecord && (oldrecord->size - oldrecord->header_size) > (size_t) vcard_max_size)) {
```
rsto, line 621: I had to wrap my brain around the negated condition. I think it would be simpler to read without the negation, such as ` .. && (!oldrecord || (oldrecord->size - oldrecord->header_size) <= (size_t) vcard_max_size))`. That's also then symmetric with how it is written in carddav_db.c

rsto (changes requested):
Could we please have a test for this? I think there is a similar one in the JMAPCalendar tests already. Other than that I just have two nits.

imap/jmap_backup.c
ksmurchison: Fixed

imap/http_carddav.c
ksmurchison: Fixed

ksmurchison:
@rsto We already have test_contact_set_toolarge.  Are you saying that you want to see a test which allows updating an oversized contact?

rsto:
> @rsto We already have test_contact_set_toolarge. Are you saying that you want to see a test which allows updating an oversized contact?

I think the toolarge test is fine.

rsto (approved):
LGTM

PR #4153, opened 2022-06-29
Quota raw fetch

brong:
This adds the `raw quota` as an annotation fetch item, which can be fetched even if conversation quotas are turned on.


ksmurchison (approved):
LGTM

PR #4156, opened 2022-06-30
Sieve tester fixes

ksmurchison:


elliefm:
I thought I'd seen an open issue about this, and went to cross-link them, but can't find it -- so I guess it must have been on the Fastmail internal tracker rather than here.

One might propose that for every bugfix PR we should also open a GitHub issue describing the problem to link the fix against (if there wasn't already one), so that a future person doing a release, or evaluating whether the fix needs to be backported, knows what the hell the fix achieves.  I don't think this is necessary.  

But I do think that it's necessary that the PR describes the issue being resolved in some way.  If there's an existing issue on GitHub that it can be linked to, great!  But if not, it should be described in the PR description.  We can't assume people looking at the history have access to or awareness of stuff in the Fastmail internal tracker; the context and commits needs to be kept on the same system.

elliefm:
@elliefm when backporting this, also grab 9020b7561, which doesn't seem to have had a PR associated with it (or at least, not that I can find)

elliefm:
@elliefm also 231e7880e...

PR #4159, opened 2022-07-01
fake netstat for WSL1

### LINKED ISSUE #4158: run testrunner.pl under WSL1
Because `netstat` cannot run properly in WSL1 (see: https://github.com/Microsoft/WSL/issues/2249), testrunner.pl wouldn't check master started and fail.

I have written a simple script to fake expected netstat behavior:
```
/mnt/c/Windows/system32/netstat.exe -an | sed -e 's/LISTENING/LISTEN/' -e 's/  TCP/tcp 0 0/'
```

Does anyone work in WSL1 environment? I thought it might be helpful.

qyb:
Signed-off-by: qyb <qiuyingbo@gmail.com>

Fixes #4158

### DIFF
--- a/cassandane/Cassandane/GenericDaemon.pm
+++ b/cassandane/Cassandane/GenericDaemon.pm
@@ -325,7 +325,7 @@ sub _is_listening_af
     my ($self, $af) = @_;
 
     my @cmd = (
-        'netstat',
+        (-e "/mnt/c/Windows/system32/netstat.exe") ? './utils/fakenetstat' : 'netstat',
         '-l',           # listening ports only
         '-n',           # numeric output
         );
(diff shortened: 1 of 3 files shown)


elliefm:
[I've updated the original post to have a descriptive title, and to reference the related issue from the description rather than the title, so that GitHub can link the two together automatically]

elliefm (commented):
I wonder if there's some way we can detect the path of `netstat.exe`, instead of it being hardcoded?

qyb:
> I wonder if there's some way we can detect the path of `netstat.exe`, instead of it being hardcoded?

Microsoft appends PATH variable in a magic place that I can't find. sudo to cyrus will lose the path. Maybe I could build the fakenetstat script when `make` in cassandane. Let me think twice.  🤔

PR #4162, opened 2022-07-05
quota: accept -n flag

cwalther:
The `-n` flag for `quota` was implemented and documented in 23d6ef9, but not actually accepted.

Actual result:
```
$ cyrus/sbin/quota -f -n -q
cyrus/sbin/quota: illegal option -- n
usage: quota [-C <alt_config>] [-d <domain>] [-f] [-q] [-J] [-n] [-u] [mailbox-spec]...
```

Expected result:
```
$ cyrus/sbin/quota -f -n -q
user.test: MESSAGE usage was 0, now 189
…
```

Tested on macOS 12.4.

### DIFF
--- a/imap/quota.c
+++ b/imap/quota.c
@@ -132,7 +132,7 @@ int main(int argc,char **argv)
     int do_report = 1;
     char *alt_config = NULL, *domain = NULL;
 
-    while ((opt = getopt(argc, argv, "C:d:fqJZu")) != EOF) {
+    while ((opt = getopt(argc, argv, "C:d:fqJnZu")) != EOF) {
         switch (opt) {
         case 'C': /* alt config file */
             alt_config = optarg;


elliefm (approved):
Good catch, thanks!

cwalther:
Thanks! Do the 3.2/3.4/3.6 labels mean that the bug is found in these releases, or that the fix will be backported to these, and in the latter case, will you cherry-pick it or am I expected to do anything?

elliefm:
On an issue, those labels would mean that the issue is known to affect those versions.  On a PR I've self-assigned, they're a reminder to me to cherry-pick the fix to those branches, next time I do a bunch of that -- which will be before the next releases from those branches, but not necessarily right away.  I'll drop the labels and the assignment once I've done it.

Nothing left for you to do :)

cwalther:
Got it!

PR #4165, opened 2022-07-07
Sieve: protect against missing/invalid sievedir

ksmurchison:
We had a case where dav_reconstruct tried to create a bytecode file in root

### DIFF
--- a/imap/jmap_sieve.c
+++ b/imap/jmap_sieve.c
@@ -127,6 +127,16 @@ static jmap_method_t jmap_sieve_methods_nonstandard[] = {
 
 HIDDEN void jmap_sieve_init(jmap_settings_t *settings)
 {
+    if (config_getswitch(IMAPOPT_SIEVEUSEHOMEDIR)) {
+        xsyslog(LOG_NOTICE, "%s", "can't use home directories");
+        return;
+    }
+
+    if (!sievedir_valid_path(config_getstring(IMAPOPT_SIEVEDIR))) {
+        xsyslog(LOG_NOTICE, "%s", "sievedir option is not defined/valid");
+        return;
+    }
+
     jmap_method_t *mp;
     for (mp = jmap_sieve_methods_standard; mp->name; mp++) {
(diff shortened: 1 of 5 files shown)


imap/jmap_sieve.c
```diff
@@ -127,6 +127,16 @@ static jmap_method_t jmap_sieve_methods_nonstandard[] = {
 
 HIDDEN void jmap_sieve_init(jmap_settings_t *settings)
 {
+    if (config_getswitch(IMAPOPT_SIEVEUSEHOMEDIR)) {
+        xsyslog(LOG_NOTICE, "%s", "can't use home directories");
+        return;
+    }
+
+    if (!sievedir_valid_path(config_getstring(IMAPOPT_SIEVEDIR))) {
+        xsyslog(LOG_NOTICE, "%s", "sievedir option is not defined/valid");
```
rsto, line 136: Isn't that rather an error than just a notice? Same for the check of sieveusehomedir.

rsto (approved):
LGTM, found just two nits.

timsieved/actions.c
```diff
@@ -91,9 +91,12 @@ int actions_init(void)
 
     if (!sieve_usehomedir) {
         sieve_dir_config = (char *) config_getstring(IMAPOPT_SIEVEDIR);
+        if (!sievedir_valid_path(sieve_dir_config)) {
+            xsyslog(LOG_ERR, "%s", "sievedir option is not defined/valid");
+        }
     } else {
         /* can't use home directories with timsieved */
-        syslog(LOG_ERR, "can't use home directories");
+        xsyslog(LOG_ERR, "%s", "can't use home directories");
```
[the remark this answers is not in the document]
ksmurchison: Yep.  Cut-n-paste error

imap/jmap_sieve.c
rsto: I guess it's the "valid" part in "not defined/valid" that triggered me. If it's invalid, it is more likely and error, but I don't know really.

PR #4168, opened 2022-07-08
jmap_api.c: use json_unpack() in validate_request()

ksmurchison:
Was just staring at some code and I thought this might be a bit cleaner.  If you disagree, feel free to reject this.


rsto (approved):
I would not have written it like this, because it
- splits declaration of the variables with their initialization
- requires a cast to un-const a pointer

But we don't have yet addressed such in the style guide, so I leave it up to you what to do.

ksmurchison:
Don't really need this. Closing

PR #4175, opened 2022-07-12
jmap_mail: reject control characters in Email headers

rsto:
Also checks header contents to be valid UTF-8

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -9027,12 +9027,46 @@ static void _email_fini(struct email *email)
     free(email->body);
 }
 
-static json_t *_header_make(const char *header_name, const char *prop_name, struct buf *val)
+static json_t *_header_make(const char *header_name,
+                            const char *prop_name,
+                            struct buf *hdr,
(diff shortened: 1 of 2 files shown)


imap/jmap_mail.c
```diff
@@ -9048,16 +9082,15 @@ static json_t *_header_from_raw(json_t *jraw,
...
-    }
-    else {
+    if (!json_is_string(jraw)) {
         jmap_parser_invalid(parser, prop_name);
         return NULL;
     }
+
+    struct buf val = BUF_INITIALIZER;
+    buf_setmap(&val, json_string_value(jraw),
+            json_string_length(jraw));
```
ksmurchison, line 9092: I wonder if we can avoid a memcpy() by using buf_init_ro() or buf_init_ro_cstr() instead

ksmurchison (commented):
LGTM other than my one comment

imap/jmap_mail.c
rsto: Done, thanks

ksmurchison (approved):

PR #4177, opened 2022-07-13
jmap_ical.c: read the color from CATEGORIES, if no COLOR property is set

ksmurchison:
Read side, waiting for confirmation that we want to handle the write side


rsto (approved):
LGTM

PR #4180, opened 2022-07-14
Reserve mailbox name (mbentry) before mailbox creation

ksmurchison:
Thinking that this might solve the issue of conv.db being created by name rather than by uuid when replicating a new user INBOX

### DIFF
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -1897,15 +1897,20 @@ EXPORTED int mboxlist_createmailbox(const mbentry_t *mbentry,
 
     newmbentry = mboxlist_entry_create();
     newmbentry->acl = xstrdupnull(acl);
-    newmbentry->mbtype = mbtype | MBTYPE_RESERVE;
+    newmbentry->mbtype = mbtype;
     newmbentry->partition = xstrdupnull(newpartition);
     newmbentry->uniqueid = xstrdup(uniqueid ? uniqueid : makeuuid());
 
     if (!(flags & MBOXLIST_CREATE_DBONLY) && !isremote) {
-        /* Reserve the mailbox - we need this so that mboxname_conf_getpath()
-           during mailbox_create() will find an mbentry with uniqueid */
-        r = mboxlist_update_entry(mboxname, newmbentry, NULL);
-        if (r) goto done;
+        if (mboxname_isusermailbox(mboxname, 1)) {
+            /* Create initial mbentry for new users --
+               the uniqueid in the record is required to open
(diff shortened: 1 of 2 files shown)


imap/mboxlist.c
```diff
@@ -1895,22 +1895,32 @@ EXPORTED int mboxlist_createmailbox(const mbentry_t *mbentry,
     else if (config_getswitch(IMAPOPT_MAILBOX_LEGACY_DIRS))
         mbtype |= MBTYPE_LEGACY_DIRS;
 
+    newmbentry = mboxlist_entry_create();
+    newmbentry->acl = xstrdupnull(acl);
+    newmbentry->mbtype = mbtype | MBTYPE_RESERVE;
```
rsto, line 1900: Does the RESERVE flag later get removed?

rsto (approved):
The change looks good but I am not too much into the mboxlist creation code. Can this PR somehow demonstrate with a test that it's fixing what it is supposed to fix? I guess it's tricky because of replication?

imap/mboxlist.c
ksmurchison: Yes.  We do `newmbentry->mbtype = mbtype` below after the mailbox has been created.

ksmurchison:
Unfortunately, I have been unable to reproduce the issue that robn reported, but I *think* this will fix it

imap/mboxlist.c
```diff
@@ -1895,22 +1895,32 @@ EXPORTED int mboxlist_createmailbox(const mbentry_t *mbentry,
...
+    newmbentry->partition = xstrdupnull(newpartition);
+    newmbentry->uniqueid = xstrdup(uniqueid ? uniqueid : makeuuid());
```
brong, line 1902: this is hoisting the "create the UUID" out of mailbox_create below.  Are there any remaining callers of mailbox_create that don't have a uniqueid?  If not, we should assert for it inside and remove the conditional from there.

ksmurchison: Done

ksmurchison:
I swear I had run Cassandane on the previous code and it worked, but my testing today revealed that using the RESERVED flag for the initial entry breaks stuff because lookup-by-id ignores reserved entries.  Using INTERMEDIATE per the previous attempt at a fix located in mailbox.c

elliefm:
This breaks the cunit annotate tests, because `mailbox_create()` now does `assert(uniqueid)` (instead of just creating a uniqueid itself), but annotate.testc's `set_up()` still calls it with `NULL`.

PR #4185, opened 2022-07-20
http_caldav: add defaults alerts to events fetched with multi-get

rsto:
Several CalDAV clients fetch events with the multi-get REPORT,
 but up until now the returned iCalendar data did not contain
 the default alerts that were set by the user.


ksmurchison (approved):
LGTM

rsto:
This no longer is needed.

PR #4186, opened 2022-07-21
JMAP upload duplicate content fix

ksmurchison:


cassandane/Cassandane/Cyrus/JMAPCore.pm
```diff
@@ -870,4 +870,28 @@ sub test_blob_set_complex
...
+    :min_version_3_7 :needs_component_jmap :JMAPExtensions
+{
+    my $self = shift;
+    my $jmap = $self->{jmap};
+    my $admin = $self->{adminstore}->get_client();
+
+    $jmap->Upload("hello", "application/data");
+
+    my $file = abs_path('data/mime/testEmail.eml');
+    open(FH, "<$file");
```
wolfsage, line 883: nit: 2 arg open and global file handles are generally discouraged these days. Preferably this would be:
```
open(my $fh, '<', $file);
local $/;
my $binary = <$fh>;
close($fh);
```

... though I realize we haven't standardized on this yet so there's a mix all over cassandane

imap/http_jmap.c
```diff
@@ -914,6 +916,9 @@ static int _create_upload_collection(const char *accountid,
         cyrus_acl_set(&newacl, httpd_userid, ACL_MODE_SET,
                       JACL_READITEMS | JACL_WRITE, NULL, NULL);
 
+        xsyslog(LOG_ERR, "reset ACL", "userid=<%s> newacl=<%s>",
```
wolfsage, line 919: is LOG_ERR the right thing to use here or should this be LOG_NOTICE

imap/http_jmap.c
```diff
@@ -1052,11 +1057,12 @@ static int jmap_upload(struct transaction_t *txn)
     const char *type = "application/octet-stream";
     if ((hdr = spool_getheader(hdrcache, "Content-Type"))) {
         type = hdr[0];
+        /* Remove CFWS and encodings from type */
+        normalisedtype = charset_decode_mimeheader(type, CHARSET_KEEPCASE);
```
wolfsage, line 1061: This means "type" in the response object will now be `null` instead of `application/octet-stream` if one isn't provided during upload? Is that expected?

wolfsage (commented):
Just some drive by questions/comments.

I'm not sure I understand commit ed0c26c645901ae6043938c5ae6500d1ff678870 at all. This says it's ignoring user supplied stuff but I don't see how?

cassandane/Cassandane/Cyrus/JMAPCore.pm
rsto: done

imap/http_jmap.c
rsto: That's an error. I updated the code and added a Cassandane test.

imap/http_jmap.c
rsto: LOG_NOTICE indeed is the better choice

rsto:
@wolfsage thanks for the review. I now also added 0d2bb0f2622b584eedbbfe90ca6c3594b54867d3 which got lost during a rebase.

wolfsage:
Can we get reasons added to these commits? The messages don't stand on their own I think:
 * https://github.com/cyrusimap/cyrus-imapd/commit/0d2bb0f2622b584eedbbfe90ca6c3594b54867d3
 * https://github.com/cyrusimap/cyrus-imapd/pull/4186/commits/2da74947a5070f00e082975407f1770da55479c1

rsto:
done

wolfsage:
Thanks. +1 from me

ksmurchison (commented):
LGTM

PR #4188, opened 2022-07-22
ctl_cyrusdb -r ("recover") uniqueid safety

elliefm:
This is a partial forward port of work I did previously for 3.4, as #4100 and #4139, and is part of the solution to #4035 

It contains the `ctl_cyrusdb -r` ("recover") updates, and everything needed to support those updates, and their tests; but it does not contain the `reconstruct` updates, which are more complex and Ken will be pursuing separately.

The updates allow `ctl_cyrusdb -r` to fix up mailboxes.db records for legacy mailboxes (i.e. stored on disk by name) that are missing a uniqueid.  If the mailbox header contains a uniqueid, that value is used; if the header doesn't have one either, then a new one is generated and set in both places.

It does _not_ recover from:

  * mbentry and header have uniqueids that differ
  * mbentry has a uniqueid but the header does not

because detecting either of these cases would require examining every header file on disk, which is beyond what `ctl_cyrusdb -r` does.

`mailbox_open_advanced()` and `mailbox_read_header()` are updated to detect missing uniqueid situations at run time, and syslog a message that the mailbox "needs reconstruct"... though `reconstruct` won't be able to fix it until it's had similar updates.

It also can't recover from any of these situations for uuid-mailboxes, because without already knowing the uniqueid, it can't obtain a namelock.  We'll probably need to fix this in the future, but it's separate work, not just a "simple" forward-port of my 3.4 work.

### DIFF
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -5405,6 +5405,19 @@ static int _check_rec_cb(void *rock,
         }
         break;
     }
+
+    case KEY_TYPE_NAME: {
+        /* Verify that we have a valid N record */
+        mbentry_t *mbentry = NULL;
+
+        r = mboxlist_parse_entry(&mbentry, NULL, 0, data, datalen);
+        if (!r) {
+            *do_upgrade = 0;
+            mboxlist_entry_free(&mbentry);
+            r = CYRUSDB_DONE;
+        }
+        break;
+    }
     }
 
     return r;
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -346,6 +346,26 @@ EXPORTED char *mbentry_archivepath(const struct mboxlist_entry *mbentry, uint32_
                                 uid);
 }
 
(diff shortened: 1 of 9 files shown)


imap/mboxlist.c
```diff
@@ -346,6 +346,26 @@ EXPORTED char *mbentry_archivepath(const struct mboxlist_entry *mbentry, uint32_
...
+             && 0 != strcmpsafe(mbentry->server, config_servername))
+    {
+        /* it's on some server that is not this one */
+        return 0;
+    }
+
+    return 1;
+}
+
 static void mboxlist_dbname_to_key(const char *dbname, size_t len,
```
dilyanpalauzov, line 369: Why do the cyrus-imap-3.2 and cyrus-imap-3.4 braches have already the function mbentry_is_local_mailbox(), while the cyrus-imap-3.6 and master branches do not have it?

elliefm: Because

> This is a partial forward port of work I did previously for 3.4

ksmurchison (approved):
LGTM

PR #4190, opened 2022-07-27
docs: upgrading to 3.6 requires 3.2.10 or 3.4.4 first

elliefm:
This updates the upgrade documentation for 3.6 to explain that one should already be running 3.2.10 or 3.4.4 before attempting to upgrade to 3.6.

I would usually just commit this straight to the 3.6 branch, but I'd like some people other than myself to read it and let me know whether it makes sense, or shake out points of confusion or ambiguity, so here's a PR.  Please have a read, and let me know what you think.

(Every future new version will also require an upgrade stopover at 3.2.10/3.4.4, so once we're happy with the wording and this PR is merged, I'll make similar changes on master so they're there ready for 3.8.)

### DIFF
--- a/docsrc/imap/download/upgrade.rst
+++ b/docsrc/imap/download/upgrade.rst
@@ -22,6 +22,29 @@ Upgrading to 3.6
 
 Things to consider **before** you begin:
 
+Versions to upgrade from
+########################
+
+Before upgrading to 3.6, your deployment should be running either:
+
+* 3.2.10 (or later), or
+* 3.4.4 (or later)
+
+If your existing deployment predates these releases, you should first upgrade
+to one of these versions, let it run for a while, resolve any issues that
+come up, and only then upgrade to 3.6.
+
+3.2.x prior to 3.2.10, 3.4.x prior to 3.4.4, and all 3.0.x and 2.x releases
+have inconsistencies in their storage of an optional metadata field (mailbox
+uniqueids).  This was not previously a problem due to the field being optional.
+
+Architectural changes in 3.6 make mailbox uniqueids required for almost all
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
How about removing from backend.c:backend_version() the detection for versions 3.3, 3.1, 3.0, 2.5, 2.4, and 2.3?

elliefm:
> How about removing from backend.c:backend_version() the detection for versions 3.3, 3.1, 3.0, 2.5, 2.4, and 2.3?

If we did that, then:

* XFER from 3.6 to older versions would lose data unnecessarily
* maintaining the XFER forward-compatibility across the legacy branches would be harder, since we'd start needing two different patches

I don't know why someone would want to XFER from 3.6 to an older version, but that doesn't mean we should just break it for no reason.

dilyanpalauzov:
> > How about removing from backend.c:backend_version() the detection for versions 3.3, 3.1, 3.0, 2.5, 2.4, and 2.3?
> 
> If we did that, then:
> 
>  * XFER from 3.6 to older versions would lose data unnecessarily

I thought removing backend-version detection for 2.4, 2.5, 3.0 on the cyrus/master branch will not break XFER from 3.6 to older versions.  It would prevent the opposite to happen, which anyway cannot work right now.

ksmurchison (approved):
LGTM

rsto (approved):
LGTM

elliefm:
Turns out the documentation on master was the same as the 3.6 documentation, so I've just cherry-picked to master it as-is rather than changing it for "3.7"...  I think that will confuse me less when I'm updating it for 3.8 in the future, because it will turn up when I search for references to 3.6.

So this is now on master too, as 4ec6568860

PR #4193, opened 2022-08-05
Replicate to archive, with tests

### LINKED ISSUE #3830: Add an ENABLE command for sync_client to turn on "spool to archive partition"

elliefm:
This is @robn's #4073, but extended with a couple of Cassandane tests.  If you have already reviewed that, please focus on the new commits from me here; if you haven't, please review the whole thing.

This will want rebasing onto current master before being merged, as the original PR is a few months old, but for now I've kept the history unaltered.

Closes #3830 and #4073

### DIFF
--- a/backup/backupd.c
+++ b/backup/backupd.c
@@ -645,7 +645,7 @@ static void cmdloop(void)
             if (!backupd_userid) goto nologin;
             if (!strcmp(cmd.s, "Apply")) {
                 struct dlist *dl = NULL;
-                c = dlist_parse(&dl, /*parsekeys*/ 1, /*isbackup*/ 1, backupd_in);
+                c = dlist_parse(&dl, /*parsekeys*/ 1, /*isarchive*/ 0, /*isbackup*/ 1, backupd_in);
                 if (c == EOF) goto missingargs;
                 if (c == '\r') c = prot_getc(backupd_in);
                 if (c != '\n') goto extraargs;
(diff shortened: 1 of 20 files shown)


backup/backupd.c
```diff
@@ -645,7 +645,7 @@ static void cmdloop(void)
             if (!backupd_userid) goto nologin;
             if (!strcmp(cmd.s, "Apply")) {
                 struct dlist *dl = NULL;
-                c = dlist_parse(&dl, /*parsekeys*/ 1, /*isbackup*/ 1, backupd_in);
+                c = dlist_parse(&dl, /*parsekeys*/ 1, /*isarchive*/ 0, /*isbackup*/ 1, backupd_in);
```
ksmurchison, line 648: I'd like to see us change the dlist functions to accept a single bitmask of flags rather than a function argument per flag.

imap/sync_support.c
```diff
@@ -8048,9 +8057,31 @@ EXPORTED int sync_connect(struct sync_client_state *sync_cs)
...
+        }
+        else {
+            syslog(LOG_NOTICE, "Destination doesn't support replication to archive");
+        }
+    }
+
+    if (capabilities) {
+        sync_do_enable(sync_cs, capabilities);
+
+        if (capabilities & CAPA_REPLICATION_ARCHIVE && !(sync_cs->flags & SYNC_FLAG_ARCHIVE)) {
```
ksmurchison, line 8081: For clarity, I'd like to see this as:

`if ((capabilities & CAPA_REPLICATION_ARCHIVE) && !(sync_cs->flags & SYNC_FLAG_ARCHIVE)) {`

ksmurchison (approved):
Looks good to me.  I've noted a few nits.  Do with them as you please.

imap/sync_support.c
elliefm: Oh well spotted.  Usually my compiler complains about bitwise-and without parens in a comparison, I wonder why it didn't complain about this one?

Anyway: fixed it, and wrapped the long line too.

backup/backupd.c
elliefm: Yeah, that would be nice.  I've left a comment for next time we're in there.

robn:
Thanks all, much obliged :bow:

PR #4194, opened 2022-08-08
'special_nochildren' mailboxes should have \Noinferiors attribute

ksmurchison:
If we don't allow children to be created, then the mailbox should have \Noinferiors

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -268,6 +268,7 @@ struct list_rock {
     uint32_t last_attributes;
     int last_category;
     hash_table server_table;    /* for proxying */
+    strarray_t *special_nochildren;
 };
 
 /* Information about one mailbox name that LIST returns */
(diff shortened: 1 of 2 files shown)


imap/imapd.c
```diff
@@ -13325,6 +13326,34 @@ static void add_intermediates(const char *extname, struct list_rock *lrock)
     strarray_fini(&inter);
 }
 
+static int is_noinferiors(struct findall_data *data, struct list_rock *rock)
+{
+    int r = 0;
+
+    if (data->mb_category == MBNAME_ALTINBOX)
```
rsto, line 13333: nit: I suggest to also add block-braces here, or handle the else condition first, followed by an else with the single statement. What we really should to do is agree on what to do in that case in the style guide. Note that the style guide already forbids an empty line between if and else blocks.

imap/imapd.c
```diff
@@ -13325,6 +13326,34 @@ static void add_intermediates(const char *extname, struct list_rock *lrock)
...
+        r = 1;
+
+    else if (strarray_size(rock->special_nochildren)) {
+        struct buf attrib = BUF_INITIALIZER;
+
+        annotatemore_lookup(mbname_intname(data->mbname), "/specialuse",
+                            mbname_userid(data->mbname), &attrib);
+
+        if (buf_len(&attrib)) {
+            strarray_t *uses = strarray_split(buf_cstring(&attrib), NULL, 0);
```
rsto, line 13343: nit: in this block I find the empty lines to make it harder to read. Also in the block above. They are not logically separate statements, so let's keep them together.

rsto (changes requested):
The logic looks good to me, but I have remarks on the code style.

imap/imapd.c
ksmurchison: Fixed

imap/imapd.c
ksmurchison: Fixed

rsto (approved):
LGTM

PR #4196, opened 2022-08-10
Add long-options support to command line tools

elliefm:
This adds long-options support to the command line tools, with the exception of `ctl_cyrusdb` and `reconstruct`.  I'll do those separately later, once the uniqueid safety work is finished, to avoid a bunch of rebasing.

This is one of those annoying ones where the lion's share of changes are in strings, which the compiler can't check for spelling or consistency.  We also don't have any tests for the majority of these commands!  So please review with a very close eye.

I have tried to not change any existing behaviour, but simply add long-option synonyms for the existing short options.

I have _not_ added a long-option synonym for `-C altconfig`, because that needs to remain consistent with the service binaries, and this PR does not touch the service binaries.

### DIFF
--- a/docsrc/imap/reference/manpages/systemcommands/deliver.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/deliver.rst
@@ -43,15 +43,15 @@ Options
 
     Ignored for compatibility with **/bin/mail**.
 
-.. option:: -r  address
+.. option:: -r address, --return-path=address
 
     Insert a **Return-Path:** header containing *address*.
 
-.. option:: -f  address
+.. option:: -f address
 
     Insert a **Return-Path:** header containing *address*.
 
-.. option:: -m  mailbox
+.. option:: -m mailbox, --mailbox=mailbox
 
     Deliver to **mailbox**.  If any *userid*\ s are specified, attempts
     to deliver to ``user.``\ *userid*\ ``.mailbox`` for each *userid*\ .
@@ -63,15 +63,15 @@ Options
     . If the ACL on *mailbox* does not grant the sender the "p" right,
     the delivery fails.
 
-.. option:: -a  auth-id
+.. option:: -a auth-id, --auth-id=auth-id
 
     Specify the authorization id of the sender.  Defaults to "anonymous".
 
-.. option:: -q  user-id
+.. option:: -q, --ignore-quota
 
     Deliver message even when receiving mailbox is over quota.
 
-.. option:: -l
+.. option:: -l, --lmtp
 
     Accept messages using the LMTP protocol.
 
--- a/imap/ctl_deliver.c
+++ b/imap/ctl_deliver.c
@@ -75,7 +75,7 @@ int main(int argc, char *argv[])
     char *days = NULL;
     enum { DUMP, PRUNE, NONE } op = NONE;
 
-    while ((opt = getopt(argc, argv, "C:drE:f:")) != EOF) {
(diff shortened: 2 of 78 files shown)


elliefm:
Updated to add a long option for `sync_client -a` (#4193), resolving the merge conflict.

docsrc/imap/reference/manpages/systemcommands/deliver.rst
```diff
@@ -63,15 +63,15 @@ Options
     . If the ACL on *mailbox* does not grant the sender the "p" right,
     the delivery fails.
 
-.. option:: -a  auth-id
+.. option:: -a auth-id, --auth-id=auth-id
 
     Specify the authorization id of the sender.  Defaults to "anonymous".
 
-.. option:: -q  user-id
+.. option:: -q, --ignore-quota
```
rsto, line 70: Was the user-id argument value just an error in the documentation?

imap/ctl_deliver.c
```diff
@@ -75,7 +76,21 @@ int main(int argc, char *argv[])
     char *days = NULL;
     enum { DUMP, PRUNE, NONE } op = NONE;
 
-    while ((opt = getopt(argc, argv, "C:drE:f:")) != EOF) {
+    /* keep this in alphabetical order */
+    static const char *const short_options = "C:E:df:";
```
rsto, line 80: Where did the r option go?

imap/ctl_mboxlist.c
```diff
@@ -1222,7 +1220,30 @@ int main(int argc, char *argv[])
     int dointermediary = 0;
     int undump_legacy = 0;
 
-    while ((opt = getopt(argc, argv, "C:Lawmdurcxf:p:viy")) != EOF) {
+    /* keep this in alphabetical order */
+    static const char *const short_options = "C:Ladf:imp:uvwxy";
```
rsto, line 1224: Also no r option here?

rsto (commented):
That must have been a ton of work! I think this all looks good, but I have a few questions.

docsrc/imap/reference/manpages/systemcommands/deliver.rst
elliefm: Yeah, the `-q` option doesn't take an argument, it just sets a flag.  `userid` is part of the "everything else after getopt has finished parsing", used by all options, rather than specific to `-q`.

I guess whoever wrote the documentation misread or misunderstood something

imap/ctl_deliver.c
elliefm: The commit message for this change explains:

>     ctl_deliver: remove dead '-r' option from getopt string
>     
>     Implementation was removed by 8d6cf338198bae06a7bc7ebb78ba1e5544029e4e

Looking at that commit's date, the "r" in the getopt string had been dangling around with no implementation for nearly twenty years... and it was already commented as "deprecated" at the time it was removed!

So I guess it was waiting all this time for someone to look at it really closely, and I finally did. :)

rsto: Sorry for that. I read the commit message but the only special-handled option that sticked was the C option.

rsto (approved):
LGTM

imap/ctl_mboxlist.c
elliefm: Also no `c` -- the implementations of these were removed in fc296e2bbf, but I guess I forgot to remove them from the getopt string at the time (seems to be a recurring theme!).  Here, I finally noticed, and finished cleaning it up.

Thanks for noticing, I really appreciate the close eye. :)

PR #4198, opened 2022-08-11
mailbox_mbentry_from_path()

ksmurchison:
This allows reconstructing a mailboxes.db record from the cyrus.* files under a given path.
Still need to determine how best to expose this in reconstruct


imap/reconstruct.c
```diff
@@ -152,16 +152,21 @@ int main(int argc, char **argv)
     struct buf buf = BUF_INITIALIZER;
     char *alt_config = NULL;
     char *start_part = NULL;
+    char *path = NULL;
```
elliefm, line 155: nit: I might call this `header_path` or something, so that someone reading the reconstruct source instead of the man page can make a good guess what kind of value they need to supply.  I'll probably call this `--header-path` when I do the long options conversion later.

elliefm (commented):
Yeah nice, this is looking good so far

ksmurchison:
@elliefm I did a rewrite of this to support multiple paths, and to also sanity check data in any existing mailboxes.db records.

I didn't find an option that takes a text file of data elements to process.

imap/reconstruct.c
```diff
@@ -715,3 +731,94 @@ static void do_mboxlist(void)
...
+                   mbentry->name, mbentry->partition, mbentry_byname->partition);
+            goto done;
+        }
+
+        if (strcmpsafe(mbentry->uniqueid, mbentry_byname->uniqueid)) {
+            printf("uniqueid clash with %s - %s vs %s\n",
+                   mbentry->name,
+                   mbentry->uniqueid, mbentry_byname->uniqueid);
+            goto done;
+        }
```
elliefm, line 772: If either (but not both) of the header file or the N record are missing the uniqueid field, this is going to just bail out instead of fixing it.

If both places are missing uniqueid, it won't bail out (`"" == ""`), but I suspect it might crash later somewhere.

imap/mailbox.c
```diff
@@ -8236,3 +8235,81 @@ HIDDEN int mailbox_changequotaroot(struct mailbox *mailbox,
...
+    zeromailbox(mailbox);
+    mailbox.mbentry = mbentry;
+
+    r = mailbox_read_header(&mailbox, path);
+    if (!r) {
+        mbentry->mbtype = mailbox.h.mbtype;
+
+        /* Steal strings from mailbox_header */
+        mbentry->name = mailbox.h.name;
+        mbentry->uniqueid = mailbox.h.uniqueid;
```
elliefm, line 8287: uniqueid might be NULL here, if the header file was missing that field for whatever reason

If `path` is a UUID directory, we should probably get the uniqueid from the path rather than from the header contents, otherwise we might have to rename the directory later instead?

elliefm (commented):
This is looking promising, but it's crucial we have a good uniqueid at the end of it.  Nothing else can work without one, and this is the tool that's supposed to fix that.

imap/reconstruct.c
```diff
@@ -715,3 +732,151 @@ static void do_mboxlist(void)
...
+                mbentry->uniqueid = xstrdup(makeuuid());
+            }
+
+            mbentry->mbtype |= MBTYPE_LEGACY_DIRS;
+        }
+    }
+
+    /* Sanity check the path */
+    int legacy = (mbentry->mbtype & MBTYPE_LEGACY_DIRS) ||
+        config_getswitch(IMAPOPT_MAILBOX_LEGACY_DIRS);
```
elliefm, line 842: I think the IMAPOPT_MAILBOX_LEGACY_DIRS check can be skipped here?

AIUI, that option determines whether _new_ mailboxes are legacy or uuid.  But in the `reconstruct -P` case we know we already have a mailbox on disk cause we have the path to the header file, we're just trying to recreate/repair an mbentry for it.  So all we care about is where it is now, not where it would be created if it were new.

The mbentry->mbtype will already have its MBTYPE_LEGACY_DIRS flag set or not-set correctly from the path examination above.

If the mailbox being reconstructed is already in a uuid path, but the system is set up for new mailboxes to be on a legacy path, then I think this check as is will fall through the `||`, and then accidentally set legacy to 1 because of the IMAPOPT_MAILBOX_LEGACY_DIRS, even though the path says this is a uuid-mailbox.  And then the metapath will be calculated wrong, and the "header ...  is not located in proper location" check will bail out.

imap/reconstruct.c
ksmurchison: Resolved by using cyrus.header, N record, makeuuid(), in that order

imap/mailbox.c
ksmurchison: Done.

imap/reconstruct.c
ksmurchison: Good point.  Fixed

docsrc/imap/reference/manpages/systemcommands/reconstruct.rst
```diff
@@ -162,6 +164,12 @@ Options
 
     Instead of mailbox prefixes, give usernames on the command line
 
+.. option:: -P
+
+    Instead of mailbox prefixes, give paths to cyrus.header files on
+    the command line.  This will ONLY create/repair mailboxes.db
+    records using data in cyrus.header and cyrus.index.
+
```
dilyanpalauzov, line 172: it is unclear, whether paths will be scanned recursively, or `-P` has to be supplied many times, with a different path each time, or many calls to `reconstruct -P` shall be done for each path. 

That is, if I want to repair all mailboxes inder `/X`, can I call `reconstruct -P /X`, or do I have to use instead `for i in $(find /X); do reconstruct -P $i; done`.

Ideally there shall be an example for this.

imap/mailbox.h
```diff
@@ -747,4 +747,6 @@ extern int mailbox_changequotaroot(struct mailbox *mailbox,
 
 extern int mailbox_parse_datafilename(const char *name, uint32_t *uidp);
 
+extern struct mboxlist_entry *mailbox_mbentry_from_path(const char *header_path);
```
dilyanpalauzov, line 750: `extern` here is not necessary.

imap/reconstruct.c
```diff
@@ -346,6 +357,10 @@ int main(int argc, char **argv)
                                   MBOXTREE_TOMBSTONES|MBOXTREE_DELETED);
             continue;
         }
+        else if (dopaths) {
```
dilyanpalauzov, line 360: `else` is not necessary.

imap/reconstruct.c
```diff
@@ -715,3 +732,150 @@ static void do_mboxlist(void)
...
+    }
+    header_path = buf_cstring(&buf);
+
+    /* Create an mbentry from cyrus.header */
+    mbentry = mailbox_mbentry_from_path(header_path);
+    if (!mbentry) {
+        printf("Invalid/missing cyrus.header at '%s'\n", header_path);
+        goto done;
+    }
+    else if (!mbentry->partition) {
```
dilyanpalauzov, line 767: `else` is redundant.

docsrc/imap/reference/manpages/systemcommands/reconstruct.rst
elliefm: You can specify multiple paths to a single `reconstruct -P` and each is dealt with, but it doesn't recurse.

So `reconstruct -P /path/to/mailbox1/cyrus.header /path/to/mailbox2/cyrus.header` would work, as would something like `find /path -name cyrus.header | xargs reconstruct -P`.  But `reconstruct -P /path/to/mailbox1/cyrus.header` will not look for or find `/path/to/mailbox1/child/cyrus.header`.

Since the -P arguments are paths to cyrus.header files, I would be surprised by it trying to recurse.  The use-case here is that things are already damaged enough that it can't just find the cyrus.header by the usual means, so it should do exactly what it's told and no more.

If it did recurse, the recursion would only be able to work for legacy mailboxes, since uuid mailboxes aren't stored according to their mailbox hierarchy.

dilyanpalauzov: > Since the -P arguments are paths to cyrus.header files

This is not spelled clear enough.  My understanding was that the parameter is a directory, containing a cyrus.header file.

elliefm (approved):
Couple of nits in the tests, but this is looking good.

docsrc/imap/reference/manpages/systemcommands/reconstruct.rst
ksmurchison: It can be a directory containing a cyrus.header or an explicit cyrus.header filename

dilyanpalauzov: > It can be a directory containing a cyrus.header or an explicit cyrus.header filename

This needs to be spelled explicitly, otherwise the documentation is ambiguous, leading to mistrust.

cassandane/Cassandane/Cyrus/CyrusDB.pm
```diff
@@ -317,4 +317,256 @@ sub test_recover_create_missing_uniqueid_legacymb
...
+
+    # reconstruct -P should find and fix the missing uniqueid
+    $self->{instance}->getsyslog();
+    $self->{instance}->run_command({ cyrus => 1 },
+                                   'reconstruct', '-P', $cyrus_header);
+    my $syslog = join(q{}, $self->{instance}->getsyslog());
+
+    # should have still existed in cyrus.header
+    $self->assert_matches(
+        qr{mailbox header had no uniqueid, setting from path}, $syslog);
```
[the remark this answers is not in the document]
ksmurchison: Fixed

cassandane/Cassandane/Cyrus/CyrusDB.pm
```diff
@@ -317,4 +317,256 @@ sub test_recover_create_missing_uniqueid_legacymb
     $self->assert_str_equals("user\x1fcassandane", $hash->{N});
 }
 
+sub test_recover_uniqueid_from_header_path_legacymb
```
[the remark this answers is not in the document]
ksmurchison: @elliefm I tried moving them to Reconstrict.pm and they error out immediately.  I added the required 'use' lines and tried with/without :NoStartInstances and then gave up.

If you want to try moving them, fell free.

docsrc/imap/reference/manpages/systemcommands/reconstruct.rst
ksmurchison: Fixed

elliefm (approved):

elliefm:
I've added a commit that moves the `reconstruct -P` tests into Reconstruct.pm.  But I can't review that one myself, since I did it!

ksmurchison:
Cass changes look good.  I can't review my own PR either

ksmurchison:
@elliefm merged into master.  I'll leave it up to you to backport this

elliefm:
Did this accidentally get merged with "rebase and merge" instead of a normal merge?  There's only a single commit on master (1d6d15ee74e11a9bd745e80be69869e5fb8d64d6) which seems to have everything in it, and no link back to the PR in the commit message like there usually would be for a normal merge

ksmurchison:
Yeah, I think my last merge was a squash, and maybe the menu item was stuck there?  Idk.

elliefm:
That sounds about right, I think it "helpfully" remembers the last thing you did, and assumes you want to do that next time as well

dilyanpalauzov:
> Did this accidentally get merged with "rebase and merge" instead of a normal merge?

In fact I like this way more, as it creates simpler/flat history, which is easier to follow.

elliefm:
For the history: this is part of the fix for #4035

PR #4199, opened 2022-08-12
cassandane.ini: don't choke on repeated params

elliefm:
Fixes a bug where if a parameter in the `[config]` section of cassandane.ini was specified more than once, Cassandane would produce junk imapd.conf files, and then all tests would error out due to master aborting.

Now, if a parameter is specified multiple times, the last value read takes precedence, and a warning is logged saying so.  In verbose mode, all values for the offending item are logged.

In practice, this change only affects the `[config]` section; all other cassandane.ini parameters are read individually, and will continue doing whatever they already did.


ksmurchison (approved):
LGTM

PR #4201, opened 2022-08-15
extend itip message format

elliefm:
This extends `imip_send()` to include the JMAP id (as `id:`) and ITIP method (as `method:`) fields in IMIP notifications.  I did this by adding these as individual new fields, though...

Michael said:
> Actually, it would be really useful to just have the same thing we have when we include `calendarEvents` in Email/get, which comes from `jmap_calendar_events_from_msg`.

I had a look at this as an option, and both things end up in `jmapical_tojmap_all()` anyway.  So it's not clear to me why `jmap_calendar_events_from_msg()` would be producing additional fields beyond what `imip_send()` already was?

I'd like to add some tests, I'm concerned that recurid seems to always be NULL in what I've looked at so far... but maybe that's expected for IMIP, I dunno.  I'd appreciate some guidance in identifying which tests I should copy and modify, there's a lot with "itip" or "imip" in the name already, but they all pass, which means none of them cared much about the structure that I've changed.

There's also a bunch of junk debug logging in here, which I'll rebase out once I don't need it anymore.

### DIFF
--- a/cassandane/tiny-tests/JMAPCalendars/calendarevent-encode-imip-uri
+++ b/cassandane/tiny-tests/JMAPCalendars/calendarevent-encode-imip-uri
@@ -6,6 +6,7 @@ sub test_calendarevent_encode_imip_uri
 {
     my ($self) = @_;
     my $jmap = $self->{jmap};
+    my $uid = 'event1uid';
 
     xlog $self, "Install a sieve script to process iMIP";
     $self->{instance}->install_sieve_script(<<EOF
@@ -26,6 +27,7 @@ EOF
                     calendarIds => {
                         Default => JSON::true,
                     },
+                    uid => $uid,
                     title => 'event1',
                     start => '2020-01-01T09:00:00',
                     timeZone => 'Europe/Vienna',
@@ -69,6 +71,9 @@ EOF
     my ($imipnotif) = grep { $_->{METHOD} eq 'imip' } @$data;
     my $payload = decode_json($imipnotif->{MESSAGE});
     $self->assert_str_equals('plus+uri@example.com', $payload->{recipient});
+    my $expect_id = encode_eventid($uid);
+    $self->assert_str_equals($expect_id, $payload->{id});
+    $self->assert_str_equals('REQUEST', $payload->{method});
 
     xlog "Assert the iTIP message has the encoded URI";
     my $itip = $payload->{ical};
--- a/imap/http_caldav_sched.c
+++ b/imap/http_caldav_sched.c
@@ -609,7 +609,7 @@ static int imip_send(const char *userid, struct sched_data *sched_data,
         goto done;
     }
 
-    json_t *jsevent, *patch;
+    json_t *jsevent, *patch, *id;
 
 #ifdef WITH_JMAP
     if (sched_data->oldical) {
@@ -632,9 +632,30 @@ static int imip_send(const char *userid, struct sched_data *sched_data,
         jsevent = json_null();
         patch = jmapical_tojmap(sched_data->newical, NULL, NULL);
     }
+
+    struct jmap_caleventid eid = { 0 };
+    icalproperty *prop = NULL;
+    struct buf buf = BUF_INITIALIZER;
+
+    /* if there's going to be a recurrence-id, it will be a layer down */
+    icalcomponent *comp = icalcomponent_get_first_real_component(sched_data->itip);
+    if (comp) {
+        prop = icalcomponent_get_first_property(comp,
+                                                ICAL_RECURRENCEID_PROPERTY);
+    }
+    if (prop) {
+        eid.ical_recurid = icalproperty_get_value_as_string(prop);
+    }
+
+    /* ditto uid, but icalcomponent_get_uid already does the right thing */
+    eid.ical_uid = icalcomponent_get_uid(sched_data->itip);
+
+    id = json_string(jmap_caleventid_encode(&eid, &buf));
+    buf_free(&buf);
 #else
     jsevent = json_null();
     patch = json_null();
+    id = json_null();
 #endif
 
     /* Don't send a bogus message - check late to not allocate our own copy */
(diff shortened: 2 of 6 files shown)


ksmurchison (approved):
LGTM after debug cruft is removed

elliefm:
@ksmurchison Thanks.  Any comment on `recurid` seeming to always be NULL?  I can't tell if that's expected, or indicative that I'm doing something subtly wrong.  If I knew what I was doing I'd just write some tests, but I don't, and don't know what tests to write.

ksmurchison:
@elliefm The reason the `recurid` is always NULL is that sched_data->itip is actually a VCALENDAR component, not a VEVENT component, so it will never find a RECURRENCE-ID property.  `icalcomponent_get_uid()` has some magic which will dive down inside VCALENDAR to find a VEVENT, VTODO, etc, but `icalcomponent_get_first_property()` only operates on the component given to it.  You're going to want to do something like 

```
prop = icalcomponent_get_first_property(icalcomponent_get_first_real_component(sched_data->itip),
                                        ICAL_RECURRENCEID_PROPERTY)
```
As far as creating a test, one of the few ways that an iTIP message would have a recurrence-id is if an attendee is invited to only a single instance of a recurring event.

elliefm:
Thanks

Looks like `JMAPCalendars.calendarevent_set_standalone_itip` already deals with itip notifications for events where the participant is only invited to a single recurrence, so I'll tinker with that to have it verify the `id` field at the same time.

elliefm:
I was seeing failures from `JMAPCalendars.calendarevent_changes_issue2558` an hour or so ago which seemed to be coming from this branch, but now that I start trying to investigate, they've stopped happening.  Curious.  Will have a look again on Wednesday.

I've extended all the tests I could find that expected to get ITIP notifications, and those now all verify that the new `id` and `method` fields contain sensible values.

ksmurchison (approved):
This all looks good, once the debugging cruft is removed.

elliefm:
@rsto This is now ready for a proper review :)

cassandane/tiny-tests/JMAPCalendars/calendarevent-encode-imip-uri
```diff
@@ -69,6 +71,9 @@ EOF
     my ($imipnotif) = grep { $_->{METHOD} eq 'imip' } @$data;
     my $payload = decode_json($imipnotif->{MESSAGE});
     $self->assert_str_equals('plus+uri@example.com', $payload->{recipient});
+    my $expect_id = "E-$uid";
```
rsto, line 74: Please replace the hardcoded event id strings with `$self->encode_eventid($uid, $recurid)`. The recurid argument is optional.

rsto (changes requested):

cassandane/tiny-tests/JMAPCalendars/calendarevent-encode-imip-uri
elliefm: Oh, great tip, thanks!

elliefm:
```
    my $expect_id = $self->encode_eventid($uid);
    xlog "XXX expect id $uid => $expect_id";
```

```
[4318] =====> Cyrus::JMAPCalendars::test_calendarevent_encode_imip_uri[75] XXX expect id event1uid => ERB-event1uid-Q2Fzc2FuZGFuZTo6Q3lydXM6OkpNQVBDYWxlbmRhcnM9SEFTSCgweDU2MWRlMjhhNjczOCk
```

Weird, `encode_eventid($uid)` is adding a load of (base64?) junk as a recurid when the recurid argument is omitted.

It's strange because I see lots of other tests using `encode_eventid($uid)` without an argument, and they're not failing.  I guess maybe they're comparing eventids where both were generated by Cassandane, so the difference from Cyrus's encoder isn't noticed?  Hmm.

elliefm: Oh I see what's wrong, it's not an object method, it's just a normal function.  So I shouldn't be calling it with `$self->` :)

imap/http_caldav_sched.c
```diff
@@ -632,18 +632,45 @@ static int imip_send(const char *userid, struct sched_data *sched_data,
...
     }
+
+    struct jmap_caleventid eid = { 0 };
+    icalcomponent *comp = NULL;
+    icalproperty *prop = NULL;
+    struct buf buf = BUF_INITIALIZER;
+
+    /* if there's going to be a recurrence-id, it will be a layer down */
+    comp = icalcomponent_get_first_real_component(sched_data->itip);
+    if (comp) {
```
dilyanpalauzov, line 643: The construct
```c
icalcomponent *comp = NULL;
comp = icalcomponent_get_first_real_component(sched_data->itip);
```
leads to dead initialization warning, e.g. in clang static analyzer (kind of useless assignment).  The more concise
```c
icalcomponent *comp = icalcomponent_get_first_real_component(sched_data->itip);
```
produces no such warning.

elliefm: That's interesting, can you paste the warning?  I don't have a recent clang handy

dilyanpalauzov: https://mail.aegee.org/dpa/scan99/2022-08-16-122910-1822492-1/

elliefm:
@rsto I've made the changes you requested, and I've also fixed the dead initialisation that @dilyanpalauzov found.  Can you re-review please?

imap/http_caldav_sched.c
rsto: I would prefer just as @dilyanpalauzov to use single-line declaration and initialization.

rsto (changes requested):

imap/http_caldav_sched.c
elliefm: done

rsto (approved):

PR #4202, opened 2022-08-15
Mailbox/changes should ignore Notes mailbox

ksmurchison:
We ignore Notes in /get but forgot to also ignore it in /changes


rsto (approved):
Does what is says on the tin. LGTM

PR #4203, opened 2022-08-16
Move rawquota

brong:
This PR moves the "userrawquota" field from being a private server metadata item to a shared metadata item on the quotaroot, which means that the admin user can view it for any user rather than having to log in as the individual user.


robn:
The diff looks reasonable. If the previous version was correct (for what it was), this should be fine too.

robn (approved):

brong:
Ken - I'm assigning this to you for a second review - just since you reviewed the original and by making you read the code you'll know what I changed!  It'll take 30 seconds.

Feel free to just merge once you're happy, it's not changing behaviour under anything which knew it was there.

ksmurchison (approved):

PR #4204, opened 2022-08-16
jmap_ical: set managed attachment resource to Link blobId

rsto:
This fixes a regression that got introduced by 59ab8120175a6f3.
Before, the managed attachment resource name matched the blobId
because the JMAP upload handler already set it on the JMAP blob.
Now that the JMAP upload ignores the Content-Disposition header
during upload, it must be set when the attachment is created as
a copy of the JMAP blob.

The JMAPCalendars.calendarevent_set_linkblobid test asserts
that the regression got fixed.


ksmurchison (approved):
LGTM

PR #4205, opened 2022-08-16
jmap_api: reject invalid sinceState with cannotCalculateChanges error

rsto:
Before, an invalid sinceState argument was rejected with an
invalidArguments error.


ksmurchison (approved):
LGTM

PR #4206, opened 2022-08-16
Remove unused ptrarray_takevf()

dilyanpalauzov:
Calling ptrarray_takevf() executes
```c
pa->data = NULL;
memset(pa->data, …);
```
and the first parameter of memset() may not be null.

### DIFF
--- a/cunit/ptrarray.testc
+++ b/cunit/ptrarray.testc
@@ -306,40 +306,6 @@ static void test_remove(void)
     ptrarray_fini(&pa);
 }
 
-static void test_takevf(void)
-{
-    ptrarray_t *pa;
-    void **pp;
-
-    pa = ptrarray_new();
-    CU_ASSERT_PTR_NOT_NULL(pa);
-    ptrarray_append(pa, PTR0);
-    ptrarray_append(pa, PTR1);
-    ptrarray_append(pa, PTR2);
-    ptrarray_append(pa, PTR3);
-    ptrarray_append(pa, PTR4);
-    CU_ASSERT_EQUAL(pa->count, 5);
-    CU_ASSERT(pa->alloc >= pa->count);
-    CU_ASSERT_PTR_NOT_NULL(pa->data);
-    CU_ASSERT_PTR_EQUAL(ptrarray_nth(pa, 0), PTR0);
-    CU_ASSERT_PTR_EQUAL(ptrarray_nth(pa, 1), PTR1);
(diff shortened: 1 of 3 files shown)


rsto:
Right, that memset is just a segfault waiting for someone to actually call ptrarray_takevf. I wouldn't mind removing it from the codebase, but maybe we want to keep it (fixed) to provide the same takevf feature as strarray does. What do you think, @elliefm ?

elliefm:
@rsto It's interesting that the existing test does call the function, and doesn't segfault -- it's even clean under valgrind -- at least with my compiler and libc.  Though I agree that `memset(NULL, ...)` feels like an accident waiting to happen.  I guess this is one of those things we've gotten away with despite it being not quite right.

If we wanted to fix it to make it good, I think the place to do that would be making `ptrarray_fini()` check that `pa->data` is non-NULL before the `memset()`.

I'm quite happy to have functions we don't currently use when they form part of a coherent API, especially when there's a test preventing us accidentally breaking it, and this fits both those criteria.  But I'm also happy to just remove it -- we can always (fix it and) add it back if we need it -- and in that case @dilyanpalauzov has already done the work.

@ksmurchison do you have any opinion?  I haven't really used the ptrarray API much so I don't have skin in this.

I've taken the time to review and test it, and it looks good, so I'm going to put an approve on it.  But I'll hold off on merging it until next week, in case we decide to fix it instead.

elliefm (approved):

ksmurchison:
I don't really care whether we keep ptrarray_takevf() or not, but if if we think that memset(NULL, 0, 0) is problematic, then we better fix ptrarray_fini() because either of the following will also result in such a memset():

```
ptrarray_t pa = PTRARRAY_INITIALIZER;
ptrarray_fini(&pa);

ptrarray_t *pa = ptrarray_new();
ptrarray_free(pa);
```

My guess is that as soon as we remove ptrarray_takevf(), somebody will have a need for it.  So, maybe its best to keep it.

rsto:
I just pushed https://github.com/cyrusimap/cyrus-imapd/pull/4214 for review. I suggest if we agree on the PR I just submitted we can close this PR as I understand we want to keep ptrarray_takevf. Thanks @dilyanpalauzov for the PR and for highlighting was wrong with ptrarray_fini.

rsto:
Thanks again @dilyanpalauzov but now that we merged https://github.com/cyrusimap/cyrus-imapd/pull/4214 we'll keep ptrarray_takevf in the codebase.

PR #4207, opened 2022-08-16
jmap_mail: read Received header during Email/import

rsto:
The spec requires to set the imported email receivedAt value to

- the optional client-set receivedAt value during import
- the optional most recent Received header in the import mail
- the time of import

in that order.

The recency of the Received header is determined to be the first Received header in the top-level headers of the message, regardless of the time value stored in there.


ksmurchison (approved):
LGTM

neilj:
This is correct according to the spec, however I have realised I made an error in the spec – if there are no Received headers, it should fallback to the Date header if present, before finally falling back to the time of import. This is to account for importing "Sent" messages that were not received via SMTP. So it should go:

The receivedAt is set to the first of:
* the optional client-set receivedAt value during import
* the optional most recent Received header in the import mail
* the optional Date header
* the time of import

rsto:
I've now updated the code accordingly and tagged this to include-in-fastmail.

PR #4210, opened 2022-08-17
Allow zero DURATION (P0D and PT0S)

ksmurchison:
Per RFC 5545, Section 3.8.2.5, DURATION > 0, but DURATION == 0 occurs frequently enough in the wild for us to allow it


rsto (approved):
LGTM

PR #4213, opened 2022-08-18
search: improve DNF normalization complexity checks

rsto:
This is a slightly modified version of https://github.com/cyrusimap/cyrus-imapd/pull/4121 which got submitted by @mildred. In contrast to the original pull request it more leniently handles DNF subtrees that double during search expression normalization. In addition, it also aligns evaluation of empty conditions and operator nodes in the Sieve and JMAP handler implementations.

Passes the full Cassandane suite.

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email-query-empty-filter-conds
+++ b/cassandane/tiny-tests/JMAPEmail/email-query-empty-filter-conds
@@ -1,0 +1,47 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_query_empty_filter_conds
+    :min_version_3_7 :needs_component_sieve :needs_component_jmap :JMAPExtensions
+{
+    my ($self) = @_;
+    my $jmap = $self->{jmap};
+    my $imap = $self->{store}->get_client();
+
+    $self->make_message('test');
+
+    xlog $self, 'run squatter';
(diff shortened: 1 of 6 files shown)


elliefm (commented):
Code looks okay, and tests pass here too.

I didn't think very hard about the specifics of the AND/OR/NOT/etc with empty conditions.  Since the tests pass, I figure that they're correct -- though it could mean both the logic and the tests are broken in the same way -- let me know if you'd like me to look closer.

One nit, which I think might be redundant code that just hasn't been cleaned up -- details inline.

elliefm (approved):
This looks okay, with the caveat that I don't completely understand everything happening here.  But it passes all the tests.

One thing I'll note is that the third commit's commit message says:

>     [...]
>     This slightly adapts 0f8de1b39cbb1fd5f8b45aefbe1de19d2464e354
>     to check the complexity of the root tree right after
>     detrivialisation, rather than processing all its children again.

That commit id doesn't exist anymore, I think because of the rebase.  I believe it referred to Mildred's commit?

If you just updated the message with the updated sha1, that would then be at risk of getting stale again if the branch gets rebased again.

Since this commit is only changing the `normalise()` function, maybe this commit message could be rephrased something like "This slightly adapts normalise() to check the...".

rsto:
Thanks for the review. I adapted the commit message.

PR #4214, opened 2022-08-19
ptrarray: do not memset buffer that gets free()d anyway

rsto:
This fixes a potential issue highlighted by @dilyanpalauzov in
https://github.com/cyrusimap/cyrus-imapd/pull/4206

ptrarray_fini calls memset to set the ptrarray data to zero,
irrespective if it points to NULL or an allocated buffer.
This is safe under normal conditions, as a ptrarray with null
data is expected to have a member count. It is unsafe if
this condition does not hold.

This patch removes the call to memset altogether, similar to
how fini is implemented for strarray.

### DIFF
--- a/lib/ptrarray.c
+++ b/lib/ptrarray.c
@@ -57,9 +57,8 @@ EXPORTED void ptrarray_fini(ptrarray_t *pa)
 {
     if (!pa)
         return;
-    memset(pa->data, 0, sizeof(void *) * pa->count);
-    free(pa->data);
-    pa->data = NULL;
+
+    xzfree(pa->data);
     pa->count = 0;
     pa->alloc = 0;
 }


ksmurchison (approved):

elliefm (approved):
Oh yeah, that makes sense.

I think the one thing we lose by not zeroing out the memory region is in the case where something still had a pointer to one of the managed pointers:  if the region had been zero'd out, trying to deref it would be an immediate segfault.  Whereas, without it being zero'd out, that access would just be a use-after-free, which might terminate the program, or might not -- and if it doesn't terminate, it'll get weird.

We don't usually zero buffers out before we free them, so this isn't a protection we're generally reliant on, so I don't think it matters.

rsto:
> We don't usually zero buffers out before we free them, so this isn't a protection we're generally reliant on, so I don't think it matters.

Yes, that was my thinking, too. We might rather want to use a memory protection allocator for all our mallocs and free, but only doing this in ptrarray is too arbitrary.

PR #4215, opened 2022-08-22
Cassandane::Util::Log: highlight xlogs starting with XXX

elliefm:
When I'm writing and/or debugging Cassandane tests, I find myself writing temporary output like this all the time:

```
    xlog "XXX the whatsit is: " . Dumper $whatsit;
```

Though if I'm debugging, I'm also usually running with high verbosity, which is real noisy, so these lines are often still pretty hard to see.

I made this quick patch for xlog, such that logged lines starting with "XXX" are highlighted in colour, and it was an immediate improvement, so here it is for y'all to benefit from too. :)


rsto (approved):
Very useful indeed!

PR #4217, opened 2022-08-22
sieve/ Rationalize calls to strarray_takevf()

dilyanpalauzov:
* free(strarray_takevf(x)) → strarray_free(x)
* free(strarray_safetakevf(x)) → strarray_free(x)


dilyanpalauzov:
This does not work.
```c
free(strarray_takevf(sa));
```
does not call `for (int i = 0 ; i < sa->count ; i++) free(sa->data[i]);`, while `strarray_free(sa);` does call it.

PR #4221, opened 2022-08-29
testrunner.pl: exit early if binary components missing

### LINKED ISSUE #4109: testrunner.pl - check that utils have been built
If you've forgotten to run "make" in the cassandane directory before running testrunner.pl, a bunch of tests will fail due to binary components being missing -- but you won't know that until the run finishes and you get the report.

It'd be nice if testrunner.pl would look for these files at startup and exit right away with an error message if they're missing, instead of running the tests and wasting twenty minutes

elliefm:
If you've forgotten to run "make" in the cassandane directory before running testrunner.pl, a bunch of tests will fail due to binary components being missing -- but you won't know that until the run finishes and you get the report.

This makes testrunner.pl look for these files at startup and exit right away with an error message if they're missing, instead of running the tests and wasting twenty minutes

You can test this by running `make -C utils clean` in the cassandane directory to remove the built components, and then run testrunner.pl as usual to see the new error messages.

Fixes #4109

Once reviewed/approved/merged, I'll cherry pick this to the older branches too.

### DIFF
--- a/cassandane/testrunner.pl
+++ b/cassandane/testrunner.pl
@@ -67,6 +67,29 @@ my $skip_slow = 1;
 my $log_directory;
 my @names;
 
+# Make sure our binary components have been built already
+# -- get their names from the Makefile
+open my $mf, '<', 'utils/Makefile'
+    or die "Can't read utils/Makefile: $!";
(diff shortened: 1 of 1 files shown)


ksmurchison (approved):
I'm not a perl guru, but LGTM

PR #4222, opened 2022-09-05
itip: remove CLASS property from iTIP messages

rsto:
This patch removes the iCalendar CLASS property when receiving and sending iTIP messages. For JMAP Calendars, Cyrus enforces the access classification of the CLASS property for sharees of a calendar mailbox. It should be up to the calendar owner, not the ORGANIZER of an event, to set the desired access level.

One thing to decide before we actually merge this is if Cyrus should always enforce this, or only if configured to do so. Looking at the standards, it seems up to the server implementations how to deal with CLASS. The semantics of the CLASS property only are vaguely described in [RFC 5545, section 3.8.1.3.](https://www.rfc-editor.org/rfc/rfc5545#section-3.8.1.3), and [RFC 5546 ](https://www.rfc-editor.org/rfc/rfc5546) explicitly allows CLASS to exist in iTIP messages, but allows implementation to ignore it as a fallback. We might add an imapd.conf option, or strip the CLASS property if Cyrus was built with JMAP support. My suggestion is to do the latter, as JMAP Calendars requires us to enforce the CLASS property with JMAP semantics.

### DIFF
--- a/cassandane/tiny-tests/JMAPCalendars/itip-remove-class-property
+++ b/cassandane/tiny-tests/JMAPCalendars/itip-remove-class-property
@@ -1,0 +1,151 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_itip_remove_class_property
+    :min_version_3_7 :needs_component_jmap :needs_component_sieve
+{
+    my ($self) = @_;
+    my $jmap = $self->{jmap};
+    my $caldav = $self->{caldav};
+
+    xlog $self, "Install a sieve script to process iMIP";
+    $self->{instance}->install_sieve_script(<<EOF
+require ["body", "variables", "imap4flags", "vnd.cyrus.imip"];
+if body :content "text/calendar" :contains "\nMETHOD:" {
+    processimip :deletecanceled :outcome "outcome";
+}
+EOF
+    );
+
+    xlog $self, "Organizer sends invite with CLASS=PRIVATE";
+
+    my $imip = <<'EOF';
+Date: Thu, 23 Sep 2021 09:06:18 -0400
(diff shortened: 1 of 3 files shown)


dilyanpalauzov:
> It should be up to the calendar owner, not the ORGANIZER of an event, to set the desired access level.

I do not agree.  If the ORGANIZER sets the class to CONFIDENTIAL, then the message is confidential, and this must be communicated to the recipients.  If the recipients can copy the message and remove the class, then then this is intentional and in the accountability of the recipients.

If fact I use the CLASS property for a completely different thing.  Once an iCalendar object is uploaded to a public (or shared) calendar, the system triggers an email notification over a mailing list, only if the CLASS is PUBLIC.  Otherwise the iCalendar object can be retrieved over CalDAV, but not email is sent when the object is uploaded.  I do send the email whenever a file is uploaded over CalDAV based on `notify_external/event_notifier`, so the iTIP trimming does not change very much.  I do think however that for such use cases the CLASS property shall not be stripped by the server.

rsto:
Yeah, I kind of expected that and it helps to learn that you use CLASS differently. I can imagine that there's a couple of different uses of CLASS out there, as the iCalendar RFCs (and from what I know also CalDAV) are very vague on its semantics.

This leaves up the question if Cyrus should strip CLASS if configured to so, or if we should rather map the JSCalendar "privacy" property (on which JMAP Calendars builds on) to a new iCalendar property. I don't have a good answer for this, yet, I will get back to this. Until then, we won't merge this PR.

rsto:
Closing this PR in favor of https://github.com/cyrusimap/cyrus-imapd/pull/4225

PR #4225, opened 2022-09-07
jmap_ical: do not map privacy property to iCalendar CLASS

rsto:
The semantics of the JSCalendar privacy property are well defined and affect how events should be presented to calendar sharees. In contrast, the iCalendar CLASS property only vaguely defines what the various classifications mean and how implementations should deal with it. Using a custom iCalendar extension property to map the privacy property allows both properties to co-exist.


imap/itip_support.c
```diff
@@ -995,6 +1011,37 @@ HIDDEN void itip_strip_personal_data(icalcomponent *comp)
...
+                }
+            }
+            break;
+
+        default:
+            break;
+        }
+    }
+
     for (prop = icalcomponent_get_first_property(comp, ICAL_ANY_PROPERTY);
```
ksmurchison, line 1045: This loop is a duplicate of the above (other than the privacy prop)

ksmurchison (changes requested):
LGTM after removing the duplicate loop

ksmurchison (approved):

imap/caldav_db.c
```diff
@@ -1179,20 +1179,18 @@ EXPORTED int caldav_writeical(struct caldav_db *caldavdb, struct caldav_data *cd
...
-                cdata->comp_flags.privacy = CAL_PRIVACY_SECRET;
-                break;
-            case ICAL_CLASS_PRIVATE:
-            case ICAL_CLASS_X:
-                cdata->comp_flags.privacy = CAL_PRIVACY_PRIVATE;
-                break;
-            default:
-                ; // ignore
+    /* Determine JSCalendar privacy */
+    for (prop = icalcomponent_get_first_property(comp, ICAL_X_PROPERTY);
```
ksmurchison, line 1184: Could probably use icalcomponent_get_x_property_by_name() in ical_support.c for this

ksmurchison (approved):
LGTM other than my one comment

PR #4226, opened 2022-09-09
imapd.conf BYTESIZE smart type

elliefm:
**Note for reviewers**: 

This will be a nuisance to review, because it's lots of small and repetitive changes that need to be correct, and the compiler can't notice semantic mistakes.  I strongly suggest stepping through one commit at a time.  Please pay particular attention to:

* Do the documentation updates match the code updates?  e.g. if the documentation says "the default unit is kibibytes", does the `config_getbytesize()` call specify `'K'`?
* Did I get the magnitude conversions correct?  Lots of these options used to be measured in KB.  Some of them used KB in the config file only, but multiplied it to bytes internally; some used KB internally too.  They're all bytes internally now.  Quota is measured in KB, so the quota-related options now need to convert _backwards_ from bytes, whereas before they didn't.  And so on.
* Does any option names need to be changed?  I have only changed the very obvious ones.
* Please compile this and run all the tests.  I think I have the oldest compiler of the group at the moment, so yours may catch things mine didn't.

-----

This adds a BYTESIZE smart type to the imapd.conf parser.  BYTESIZE options take a suffix like "B", "KB", "MB", "GB", etc to allow the admin to specify readable values at a sensible scale.  Internally, they are all converted to bytes.  It's very similar to the existing DURATION smart type.

The BYTESIZE type's numeric representation is `int64_t` (all other types still use `int` or `long` internally), which allows it to represent values of "2GB" and over.

The units are all powers of two, never multiples of 1000.  Strictly, these are "kibibytes" etc in today's nomenclature.  The parser treats "K", "KB", and "KiB" the same (and case-insensitively) -- as `x * 1024` (and so on for the larger units), and this is documented.

The documentation uses the "kibibytes" etc naming, but I'm not wedded to it.  If we actually want to call them "kilobytes", then I'll do so.

Existing options have been converted to use the BYTESIZE type where appropriate.  In some cases the old option name forced a unit (e.g. `quotawarnkb`) -- these have been renamed (e.g. to `quotawarnsize`), and the old name deprecated.  Note that relevant deprecated names have also been converted to BYTESIZE -- this is the only way the deprecation can work, since it cannot convert between types.

A bunch of these options historically used 0-or-negative values to represent "unlimited", and handled that using `INT_MAX` as either an actual limit or a sentinel value.  I considered changing these to use `INT64_MAX`, but:

* many parts of our code aren't 64-bit clean, especially the older sections, and can't be expected to handle large values even though the config parser now can
* some of these are limits that are published to clients, such as the APPENDLIMIT capabilities response, and it would be inappropriate for us to represent that APPEND can support up to 9000 petabyte uploads when it almost certainly cannot, but that is what would happen if we used `INT64_MAX` and an admin left `maxmessagesize` at its default `"0"`

So instead I have added `#define BYTESIZE_UNLIMITED (INT_MAX)`, and the options that previously used `INT_MAX` to represent "unlimited" now use `BYTESIZE_UNLIMITED` for the same purpose (and with the same internal value).  An admin can choose to set an explicit larger value if they wish (and if it breaks they can keep both parts), but the default behaviour, and the documented "unlimited" behaviour, will continue to work as they have.  The documentation for these options has been updated to reflect that "unlimited" is really a large internally-defined limit.

### DIFF
--- a/docsrc/imap/concepts/overview_and_concepts.rst
+++ b/docsrc/imap/concepts/overview_and_concepts.rst
@@ -321,9 +321,14 @@ the sender.
 Quota Warnings Upon Select When User Has ``d`` Rights
 =====================================================
 
-When a user selects a mailbox whose quota root has usage that is close to or over the limit and the user has ``d`` rights on the mailbox, the server will issue an alert notifying the user that usage is close to or over the limit. The threshold of usage at which the server will issue quota warnings is set by the ``quotawarn`` configuration option.
+When a user selects a mailbox whose quota root has usage that is close to or
+over the limit and the user has ``d`` rights on the mailbox, the server will
+issue an alert notifying the user that usage is close to or over the limit.
+The threshold of usage at which the server will issue quota warnings is set
+by the ``quotawarnpercent`` configuration option.
 
-The server only issues warnings when the user has ``d`` rights because only users with ``d`` rights are capable of correcting the problem.
+The server only issues warnings when the user has ``d`` rights because only
+users with ``d`` rights are capable of correcting the problem.
 
 Quotas and Partitions
 =====================
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -333,9 +333,11 @@ Blank lines and lines beginning with ``#'' are ignored.
    assumed.  */
 */
 
-{ "archive_maxsize", 1024, INT, "3.0.0" }
-/* The size in kilobytes of the largest message that won't be archived
-   immediately.  Default is 1Mb */
+{ "archive_maxsize", "1024 K", BYTESIZE, "UNRELEASED" }
+/* The size of the largest message that won't be archived immediately.
+.PP
+   For backward compatibility, if no unit is specified, kibibytes is
+   assumed.  */
 
 { "archive_keepflagged", 0, SWITCH, "3.0.0" }
 /* If set, messages with the \\Flagged system flag won't be archived,
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -3103,11 +3103,14 @@ product version in the capabilities
    character '.'. Note that with the newnews separator, no dots may
    occur in mailbox names.  The default switched in 3.0 from off to on. */
 
-{ "vcard_max_size", 0, INT, "3.6.0" }
+{ "vcard_max_size", "0", BYTESIZE, "UNRELEASED" }
 /* Maximum allowed vCard size.
    If non-zero, CardDAV and JMAP will reject storage of contacts whose
-   vCard representation is larger than \fIvcard_max_size\fR bytes.
-   If set to 0, this will allow vCards of any size (the default). */
+   vCard representation is larger than \fIvcard_max_size\fR.
+.PP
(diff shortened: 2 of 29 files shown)


ksmurchison (approved):
I went through this slowly, commit-by-commit and this all looks correct.

And it compiles with GCC 11.3.1

elliefm:
Thanks!

lib/imapoptions
```diff
@@ -2129,14 +2173,26 @@ If all partitions are over that limit, this feature is not used anymore.
...
+{ "quotawarnsize", "0", BYTESIZE, "UNRELEASED" }
+/* The maximum amount of free space at which to give a quota warning
+   (if this value is 0, or if the quota is smaller than this
+   amount, then warnings are always given).
+.PP
+   Note that quota has kibibyte granularity.  Values specified here will
+   be truncated to the nearest whole kibibyte.
+.PP
+   For backward compatibility, if no unit is specified, kibibytes is
+   assumed.  */
```
dilyanpalauzov, line 2195: There is no need for backword compatibility, since the option `qoutawarnsize` has not existed in old versions.

docsrc/imap/concepts/overview_and_concepts.rst
```diff
@@ -321,9 +321,14 @@ the sender.
 Quota Warnings Upon Select When User Has ``d`` Rights
 =====================================================
 
-When a user selects a mailbox whose quota root has usage that is close to or over the limit and the user has ``d`` rights on the mailbox, the server will issue an alert notifying the user that usage is close to or over the limit. The threshold of usage at which the server will issue quota warnings is set by the ``quotawarn`` configuration option.
-
-The server only issues warnings when the user has ``d`` rights because only users with ``d`` rights are capable of correcting the problem.
+When a user selects a mailbox whose quota root has usage that is close to or
```
dilyanpalauzov, line 324: According to https://www.cyrusimap.org/imap/reference/admin/access-control/rights-reference.html#d the `d` right is old form, that currently is the join of the rights `x`, `t` and `e`.

Please update the text to use the `x`, `t`, and/or `e` right as appropriate.

lib/imapoptions
```diff
@@ -1666,16 +1702,24 @@ Blank lines and lines beginning with ``#'' are ignored.
...
    zero means no limit */
 
-{ "maxmessagesize", 0, INT, "2.3.17" }
-/* Maximum incoming LMTP message size.  If non-zero, lmtpd will reject
-   messages larger than \fImaxmessagesize\fR bytes.  If set to 0, this
-   will allow messages of any size (the default). */
+{ "maxmessagesize", "0", BYTESIZE, "UNRELEASED" }
+/* Maximum size of messages that will be accepted by Cyrus.  This affects LMTP
+   deliveries, IMAP appends, DAV uploads, etc.  Messages larger than this will
+   be rejected.
```
dilyanpalauzov, line 1708: If this can be overwritten per data type using options like `icalendar_max_size`, `webdav_attachments_max_binary_attach_size`, then this shall be spelled explicitly.

PR #4229, opened 2022-09-12
Sync proto tooling

brong:
This is mostly adding some Perl magic that can be used to dump and undump entire users for building future tests around.  Doesn't touch the C code at all.


brong:
I'm just merging this one since it doesn't touch anything we care about, and it makes the next stage in a separate repo much more sensible

PR #4231, opened 2022-09-13
HTTP: return 503 if a user autheticates but doesn't have an INBOX

ksmurchison:
(this can occur during a user rename if the old auth creds still exist)

This issue has been seen in production and a 503 makes more sense than 500

### DIFF
--- a/cassandane/Cassandane/Cyrus/JMAPCore.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPCore.pm
@@ -926,4 +926,42 @@ sub test_blob_upload_type
     $self->assert_str_equals("application/octet-stream", $res->{type});
 }
 
+sub test_get_session_rename_race
+    :AllowMoves :min_version_3_7 :needs_component_jmap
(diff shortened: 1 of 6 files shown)


rsto (commented):
The patch looks good to me, but: I see this case already handled in jmap_notes and jmap_mail_submission, but not in any of jmap_mail, jmap_calendar, jmap_contacts. Couldn't the same situation also show up there?

ksmurchison:
@rsto The situations that I handled are in the post-HTTP auth callbacks, so we detect them before we actually try to process a request.  It would be a pretty tight race where the client was able to authenticate and execute the callback, and then the INBOX goes away before processing the JMAP request.
Does this make sense, or am I missing something?
Hmm, can we enable JMAP without enabling CalDAV and/or CardDAV?

rsto (approved):

rsto:
We agreed to solve the specific issue at hand, but still need to look more into race conditions with rename.

PR #4232, opened 2022-09-14
Jmap blob 13

brong:
This adds digest support per the latest jmap blob draft


ksmurchison (approved):
@brong I pushed a commit which allows this to work with OpenSSL3, and also uses the SASL MD5 implementation if needed

PR #4234, opened 2022-09-14
http_dav_sharing.c: need to URL decode shared resource URI

ksmurchison:
(and don't try to use a NULL mbentry resulting from a bad URI)

Fix an issue seen on prod


rsto (approved):
LGTM

ksmurchison:
Deployed to staff

PR #2815, opened 2019-06-27
Link libcyrus.so with -lm (3.0)

dilyanpalauzov:
_ because it uses log() and ceil() from bloom.c.

Fixes #2124.

PR #4235, opened 2022-09-15
Fix build with LLD

thesamesam:


dilyanpalauzov:
This is more or less the same as https://github.com/cyrusimap/cyrus-imapd/pull/2815.

dilyanpalauzov:
Does the linking work with the original code, if ld.bfd and ld.gold are used instead of lld?  If it works, then it is a problem with lld and should be reported there (all linkers shall produce the same result, when calls with the same parameters).

If it fails with all linkers, then the linker name shall be removed from the problem description.

elliefm (approved):
Thanks.  This looks good, and works fine with GCC, but I don't have an LLVM setup handy to test that with.  I'll see if someone else does.

elliefm:
@rsto I think maybe you had a Clang/LLVM build environment handy, if so can you test this out please?  Thanks :)

thesamesam:
> Does the linking work with the original code, if ld.bfd and ld.gold are used instead of lld? If it works, then it is a problem with lld and should be reported there (all linkers shall produce the same result, when called with the same parameters).
> 
> If it fails with all linkers, then the linker name shall be removed from the problem description.

I'm not sure what to respond with. It's not an LLD bug, and it works by chance with BFD. This kind of transitive linking issue is common. I haven't bothered trying gold as it's on the way out. But it's a straightforward change and easy-to-verify fix.

elliefm:
@rsto has tried this out in clang, but ran into problems because our master branch doesn't build cleanly under clang at the moment (for reasons unrelated to this PR).

I'm going to merge this anyway, on the basis that it looks correct, and works fine under GCC.  And if it turns out to break something after all, we'll just fix it when we find it.

dilyanpalauzov:
> It's not an LLD bug, and it works by chance with BFD. 

Where is the difference of the behaviour described?

By a chance I use similar changes, mentioned at https://github.com/cyrusimap/cyrus-imapd/pull/2815 .  But as part of my changes, I removed `-lm` from  HTTP_LIBS.  Hence the question:

* If `-lm` is removed from HTTP_LIBS, does linking still work properly?

dilyanpalauzov:
> It's not an LLD bug, and it works by chance with BFD

I disagree.  I had various problematic cases in the past with C and the first thing I had to do is to find out if it is compiler or linker problem.  So I combined gcc+ld.bfd, clang+ld.bfd, gcc+ld.gold, cland+ld.bfd, and maybe at some point I used also LLD.

Using the combinations gcc+ld.bfd, clang+ld.bfd, gcc+ld.gold and cland+ld.bfd, sometimes two cases failed, so the defect component was clear.  But sometimes three cases failed, so it exposed different problems in several components.

The linkers are really meant to be interchangeable.

Please test using gcc+ld.bfd, clang+ld.bfd, gcc+ld.gold, cland+ld.bfd, gcc+ld.lld, and clang+ld.bfd.  If only LLD linking fails, this shall be reported towards LLD and fixed there.  Also I do not think LLD is as mature as ld.bfd or ld.gold.

PR #4237, opened 2022-09-19
Mailbox changes on thread

brong:
It was just about as quick to write it as describe it :)

This was discussed on today's call, there was a bug in Mailbox/changes where it wouldn't notice that counts had changed.  This makes Mailbox/changes a little MORE noisy than it should be, but it's not too bad because this way we notice any mailbox which might have changed (a thread that referenced it changed).


rsto (approved):
Great!

PR #4238, opened 2022-09-20
mboxname.c: log mboxname_lock() errors

ksmurchison:
Getting assert()s in mboxlist_createmailbox() due to failure to get a namespace lock, but we don't know why.

### DIFF
--- a/imap/mboxname.c
+++ b/imap/mboxname.c
@@ -234,10 +234,14 @@ EXPORTED int mboxname_lock(const char *mboxname, struct mboxlock **mboxlockptr,
                      nonblock, fname);
     if (!r) lockitem->l.locktype = locktype;
     else if (errno == EWOULDBLOCK) r = IMAP_MAILBOX_LOCKED;
-    else r = errno;
+    else r = IMAP_IOERROR;
 
 done:
-    if (r) remove_lockitem(lockitem);
+    if (r) {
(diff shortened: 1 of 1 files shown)


imap/mboxname.c
```diff
@@ -234,10 +234,14 @@ EXPORTED int mboxname_lock(const char *mboxname, struct mboxlock **mboxlockptr,
...
     if (!r) lockitem->l.locktype = locktype;
     else if (errno == EWOULDBLOCK) r = IMAP_MAILBOX_LOCKED;
-    else r = errno;
+    else r = IMAP_IOERROR;
 
 done:
-    if (r) remove_lockitem(lockitem);
+    if (r) {
+        xsyslog(LOG_ERR, "can not lock mailbox",
+                "name=%s error=%s (%m)", mboxname, error_message(r));
```
rsto, line 242: The format string in xsyslog should use angle brackets for output arguments, e.g. `name=<%s>`, that's also the format that xsyslog uses for its internal output.
Also, the `%m` format is a glibc-extension, and it doesn't look necessary to me because xsyslog outputs the latest errno in its `syserror` log variable for LOG_ERR anyway.

rsto (changes requested):

imap/mboxname.c
ksmurchison: Fixed

rsto (approved):

PR #4239, opened 2022-09-21
jmap: option to allow relaxed IDs in createdIds

brong:
This was lying around on my desktop - it seems it might be generally useful to merge.

### DIFF
--- a/imap/jmap_api.c
+++ b/imap/jmap_api.c
@@ -330,6 +330,7 @@ static int validate_request(struct transaction_t *txn, const json_t *req,
 HIDDEN int jmap_is_valid_id(const char *id)
 {
     if (!id || *id == '\0') return 0;
+    if (config_getswitch(IMAPOPT_JMAP_RELAXEDIDS)) return 1;
     const char *p;
     for (p = id; *p; p++) {
         if (('0' <= *p && *p <= '9'))
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -1307,6 +1307,9 @@ Blank lines and lines beginning with ``#'' are ignored.
 .PP
    If no unit is specified, seconds is assumed. */
 
+{ "jmap_relaxedids", 0, SWITCH, "3.6.0" }
+/* If enabled, allow invalid characters in JMAP ids */
+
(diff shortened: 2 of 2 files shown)


rsto (commented):
I don't think this is right. The commit message currently tells the truth, because jmap_is_valid_id only is called within jmap_api.c for creation ids. But as soon as we use jmap_valid_id somewhere else it might accept invalid regular JMAP ids (we already have a copy of the jmap_valid_id code in jmap_ical.c for stupid linkage reasons, but we might refactor it to link to jmap_api).
I suggest we either introduce `jmap_is_valid_creation_id` or parametrize `jmap_is_valid_id`.

elliefm (changes requested):
(Changes requested as per @rsto's review)

elliefm:
@brong Do you still think this is useful? Robert thought it needed work, and it no longer applies cleanly so it'll need a rebase at least.

robmueller:
As an FYI, I believe at the moment cyrus uses the uid from the iCal data for the CalendarEvent jmap id and the calendar path component for the Calendar jmap id. Neither of these have to conform to RFC8620 section-1.2, so cyrus can (and does) return invalid jmap id's. It feels problematic that an id returned from cyrus can't then be used in the createdIds map passed back in to cyrus.

rsto:
> As an FYI, I believe at the moment cyrus uses the uid from the iCal data for the CalendarEvent jmap id and the calendar path component for the Calendar jmap id. Neither of these have to conform to RFC8620 section-1.2, so cyrus can (and does) return invalid jmap id's.

That is correct for JMAP calendar ids, it is in general not correct for calendar event ids. For the latter, Cyrus encodes the UID in base64 if the original UID value is not a valid JMAP id. However, an excessively long UID value where either its original or base64-encoded representation exceed the 255 octet limit would render it invalid again.

> It feels problematic that an id returned from cyrus can't then be used in the createdIds map passed back in to cyrus.

I very much agree. I think the answer is to make Cyrus only return valid identifiers for all JMAP object types, rather than relaxing creation ids. This will take time to implement, though, so if there currently is an issue with invalid ids, we might indeed want to relax id validation.

brong:
We're taking a different approach to IDs, this is now not needed

PR #4240, opened 2022-09-21
index: track changes for modseq bump when setting seen on shared folders

brong:
This fixes https://github.com/cyrusimap/cyrus-imapd/issues/3240 properly for new Cyrus


elliefm (approved):
That's really clean, nice.

I've run this through my updated tests from #3829 and it passes them.  Do we want to add them to this PR?  It's two commits that cherry pick easily enough.

PR #4163, opened 2022-07-06
ci: add github workflow

ajaysusarla:
cyrus-imapd has been relying on Travis CI for builds. And Travis CI
has been pretty unreliable for our use-case as we keep running out of
'free OSS credits'.
This commit adds support for Github Actions for builds.

While most of what we did with Travis worked with Actions, a minor
change in the how cassandane stops processes was also needed and is a
part of this commit. The change was to use `stop_command()` method
instead of `_stop_pid()` method directly, making sure that a
`waitpid()` call is made after a `kill()`.

PR #4242, opened 2022-09-21
GitHub actions

wolfsage:
This adds a github action for cyrus that builds cyrus, runs make check, and runs most of the cassandane tests.

The tests take 30m to run, and currently these three fail (seemingly on docker too, so easily reproducible?):
```
Cyrus::Sieve.imip_reply_override                                        [FAILED]
Cyrus::Sieve.imip_reply_override_google                                 [FAILED]
Cyrus::Caldav.invite_switch_implicit_allday_to_dtend                    [FAILED]

```

It collects all of the syslog logs at the end for "easy" debugging of test failures.

If we accept this, or something like it, it's trivial to add other test jobs. I've got branches that add:
 * make check - runs in about 4m
 * clang build + cassandane run
 * address sanitizer run

We could also add a valgrind run, and whatever else we like (linters, etc?)

This would also make it easy-ish to have multiple runs per MR that include different forms of cyrus configurations 

Based off of @ajaysusarla 's work in https://github.com/cyrusimap/cyrus-imapd/pull/4163 (thanks!)


wolfsage:
Here's a sample job: https://github.com/wolfsage/cyrus-imapd/actions/runs/3099174541/jobs/5017995897

.github/workflows/main.yml
```diff
@@ -0,0 +1,49 @@
+name: Cyrus IMAP CI
+
+on:
+  push:
+  pull_request:
+    branches:
+      - master
+      - main
```
dilyanpalauzov, line 8: There is no `main` branch.

rjbs: …yet.

rjbs:
@wolfsage Would you please click the "Details" next to the failed DCO check and take the --signoff steps requested?  Thanks.

elliefm:
> `Caldav.invite_switch_implicit_allday_to_dtend`

Anecdata: this one has been failing on my system for a month or two.  I believe it might be fixed if I update my cyruslibs/libical, but I've forgotten what made me think that.  But since I think that, I've mostly just been ignoring it until I get around to updating libical...

PR #4245, opened 2022-09-22
jmap_calendar: handle jmap_mbentry_from_dav returning NULL

rsto:
jmap_mbentry_from_dav may return NULL if a user attempts to load the mailbox list entry for a shared calendar for which that user does not have the required ACLs. This is a legit condition but the Calendar/Event/set code did not handle it properly.


ksmurchison (approved):
LGTM

PR #4253, opened 2022-09-23
mailbox.c: remove "stale" index records when reconstructing DAV records

ksmurchison:
For dav_reconstruct, iterate from most recent to least recent index records so that if we find an index record with a lower UID than the DAV DB record for that object (the object has been replaced), we can remove the "stale" index record.

The question is, do we want to do this for normal DAV updates or just on reconstruct?  newrecord->uid < cdara->dav.imap_uid is probably always bogus.

### DIFF
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -136,7 +136,7 @@ struct mailbox_iter {
     uint32_t num_records;
     unsigned skipflags;
     seqset_t *uidset;
-    int step_inc;
+    int32_t step_inc;
 };
 
 
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -3520,8 +3520,7 @@ out:
 
 #ifdef WITH_DAV
 static int mailbox_update_carddav(struct mailbox *mailbox,
-                                  const struct index_record *new,
-                                  int is_reconstruct)
+                                  const struct index_record *new)
 {
     struct carddav_db *carddavdb = NULL;
     struct param *param;
@@ -3553,7 +3552,7 @@ static int mailbox_update_carddav(struct mailbox *mailbox,
 
     /* does it still come from this UID? */
     if (cdata->dav.imap_uid > new->uid) {
-        if (is_reconstruct)r = IMAP_NO_MSGGONE;
+        r = IMAP_NO_MSGGONE;
         goto done;
     }
 
@@ -3607,8 +3606,7 @@ done:
 }
 
 static int mailbox_update_caldav(struct mailbox *mailbox,
-                                 const struct index_record *new,
-                                 int is_reconstruct)
+                                 const struct index_record *new)
 {
     struct caldav_db *caldavdb = NULL;
     struct param *param;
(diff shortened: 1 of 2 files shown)


imap/mailbox.c
```diff
@@ -7915,6 +7953,16 @@ EXPORTED struct mailbox_iter *mailbox_iter_init(struct mailbox *mailbox,
     iter->num_records = mailbox->i.num_records;
     iter->msg = message_new();
 
+    /* Are we iterating forward or backward? */
+    if (flags & ITER_STEP_BACKWARD) {
```
brong, line 7957: ooh nice - this is a handy API.  We will need to make sure it's something we can do with any new tech we port this to though

imap/mailbox.c
```diff
@@ -7942,8 +7990,10 @@ EXPORTED const message_t *mailbox_iter_step(struct mailbox_iter *iter)
 {
     if (mailbox_wait_cb) mailbox_wait_cb(mailbox_wait_cb_rock);
 
-    for (iter->recno++; iter->recno <= iter->num_records; iter->recno++) {
```
brong, line 7945: This feels complicated when we could PROBABLY just use a iter_step_backwards API rather than doing two tests every loop

brong (commented):
Generally looks good - I have some comments about design.  It would be nice to have the interfaces work the same for reconstruct and non-reconstruct mode too, and not need to plumb it around so much.  The hard bit is making update_dav return codes be handled either way with a "no change" vs "change applied correctly"

imap/mailbox.c
ksmurchison: I did it this way so we didn't have to replicate code or make function call to run the common code.
We are already doing 6-8 tests per loop as it is, but if you still have a problem with it, I'll rewrite it.

imap/mailbox.c
```diff
@@ -3642,7 +3648,15 @@ static int mailbox_update_caldav(struct mailbox *mailbox,
     caldav_lookup_resource(caldavdb, &mbentry, resource, &cdata, /*tombstones*/1);
 
     /* has this record already been replaced?  Don't write anything */
-    if (cdata->dav.imap_uid > new->uid) goto done;
+    if (cdata->dav.imap_uid > new->uid) {
+        if (is_reconstruct) {
```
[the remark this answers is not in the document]
ksmurchison: I changed the code to always remove stale records, regardless of when we update records.

imap/mailbox.c
```diff
@@ -3550,7 +3552,10 @@ static int mailbox_update_carddav(struct mailbox *mailbox,
     carddav_lookup_resource(carddavdb, &mbentry, resource, &cdata, /*tombstones*/1);
 
     /* does it still come from this UID? */
-    if (cdata->dav.imap_uid > new->uid) goto done;
+    if (cdata->dav.imap_uid > new->uid) {
+        if (is_reconstruct)r = IMAP_NO_MSGGONE;
```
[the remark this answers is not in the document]
ksmurchison: Fixed

imap/mailbox.c
```diff
@@ -136,6 +136,7 @@ struct mailbox_iter {
     uint32_t num_records;
     unsigned skipflags;
     seqset_t *uidset;
+    int step_inc;
```
rsto, line 139: nit: this is added to recno later, which is a uint32_t, so maybe step_inc should be int32_t or int8_t.

ksmurchison: But step_inc can be -1 when iterating backward

rsto: Yes, that's why I suggest a signed integer value that has less or equal width as an unsigned 32-bit.

rsto (approved):
This generally looks good to me, other than the nit I already highlighted in a previous commit.

brong (approved):
Yeah, this is fine :)  It'll do the job.  My only worry is that we may need to rework the ITER_STEP_BACKWARDS if we change the iterator logic later, but that's a problem for another day

PR #4257, opened 2022-09-26
various: mem leak fixes

### LINKED ISSUE #4247: potential memory leak in backup_append()
backup_append() contains:
```c
struct dlist_print_iter *iter = NULL;
iter = dlist_print_iter_new(dlist, 1);
do {
 if (r) goto error;
} while (dlist_print_iter_step(iter, &buf));
dlist_print_iter_free(&iter);
```
When r ≠ 0 the memory pointed by `iter` is not freed.

Assigning twice value to iter at the beginning produces dead initialization warning.

### LINKED ISSUE #4248: Memory leak in write_personal_data()
write_personal_data(() contains:
```c
    struct dlist *dl = dlist_newkvlist(NULL, "CALDATA");
    int ret;

    ret = mailbox_get_annotate_state(mailbox, uid, NULL);
    if (ret) return ret;
```
The memory for `dl` is not freed.

### LINKED ISSUE #4249: Memory leak in notify_parse_path()
notify_parse_path() contains:
```c
done:
    mbname = mbname_from_userid(tgt->userid);
    mbname_push_boxes(mbname, config_getstring(IMAPOPT_DAVNOTIFICATIONSPREFIX));

    /* XXX - hack to allow @domain parts for non-domain-split users */
    if (httpd_extradomain) {
        /* not allowed to be cross domain */
        if (mbname_localpart(mbname) &&
            strcmpsafe(mbname_domain(mbname), httpd_extradomain))
            return HTTP_NOT_FOUND;
    }
```
The memory pointed by `mbname` is not freed on the above `return`.

### LINKED ISSUE #4250: Memory leak in add_address()
http_caldav_sched.c:add_address() contains:
```c
    struct address_t *new = xzmalloc(sizeof(struct address_t));
    const char *address = icalproperty_get_address(prop);
    if (!address) return;
```
`new` can be leaked.

### LINKED ISSUE #4252: Memory leak in index_copy()
index_copy() contains:
```c
ptrarray_t *msgrecs = ptrarray_new();
is_same_user = mboxname_same_userid(index_mboxname(state), name);
if (is_same_user < 0)
    return is_same_user;
r = index_check(state, usinguid, usinguid);
if (r) return r;
```

msgrecs is leaked on any of the returns.

Please consider generally defining and declaring variables just before they are needed, instead of the beginnig of each function.  This simple procedure would avoid this and other memory leaks.

elliefm:
Fixes a bunch of memory leaks found by @dilyanpalauzov 

Fixes #4247
Fixes #4248
Fixes #4249
Fixes #4250
Fixes #4252

### DIFF
--- a/imap/caldav_util.c
+++ b/imap/caldav_util.c
@@ -787,12 +787,13 @@ static int write_personal_data(const char *userid,
     struct message_guid guid;
     struct buf value = BUF_INITIALIZER;
     const char *icalstr = icalcomponent_as_ical_string(vpatch);
-    struct dlist *dl = dlist_newkvlist(NULL, "CALDATA");
+    struct dlist *dl;
     int ret;
 
     ret = mailbox_get_annotate_state(mailbox, uid, NULL);
     if (ret) return ret;
 
+    dl = dlist_newkvlist(NULL, "CALDATA");
     dlist_setdate(dl, "LASTMOD", time(0));
     dlist_setnum64(dl, "MODSEQ", modseq);
     message_guid_generate(&guid, icalstr, strlen(icalstr));
--- a/imap/http_caldav_sched.c
+++ b/imap/http_caldav_sched.c
@@ -189,7 +189,7 @@ struct address_t {
 static void add_address(struct address_t **recipients, icalproperty *prop,
                         const char* (*icalproperty_get_address)(icalproperty *))
 {
-    struct address_t *new = xzmalloc(sizeof(struct address_t));
+    struct address_t *new;
     icalparameter *param;
 
     const char *address = icalproperty_get_address(prop);
@@ -197,6 +197,7 @@ static void add_address(struct address_t **recipients, icalproperty *prop,
     if (!strncasecmp(address, "mailto:", 7))
         address += 7;
 
+    new = xzmalloc(sizeof(struct address_t));
     new->addr = address;
     param = icalproperty_get_first_parameter(prop, ICAL_CN_PARAMETER);
     if (param) {
--- a/imap/index.c
+++ b/imap/index.c
@@ -2989,7 +2989,7 @@ index_copy(struct index_state *state,
     struct mailbox *destmailbox = NULL;
     struct index_map *im;
     int is_same_user;
-    ptrarray_t *msgrecs = ptrarray_new();
+    ptrarray_t *msgrecs = NULL;
 
     *copyuidp = NULL;
 
@@ -3050,6 +3050,7 @@ index_copy(struct index_state *state,
     if (r) goto done;
 
     docopyuid = (appendstate.myrights & ACL_READ);
+    msgrecs = ptrarray_new();
 
     for (i = 0; i < copyargs.nummsg; i++) {
         msgrecord_t *mr = msgrecord_from_index_record(srcmailbox, &copyargs.records[i]);
@@ -3110,11 +3111,13 @@ index_copy(struct index_state *state,
 done:
     free(copyargs.records);
     mailbox_close(&destmailbox);
-    for (i = 0; i < msgrecs->count; i++) {
-        msgrecord_t *mr = ptrarray_nth(msgrecs, i);
-        msgrecord_unref(&mr);
+    if (msgrecs) {
+        for (i = 0; i < msgrecs->count; i++) {
+            msgrecord_t *mr = ptrarray_nth(msgrecs, i);
+            msgrecord_unref(&mr);
+        }
+        ptrarray_free(msgrecs);
     }
-    ptrarray_free(msgrecs);
 
     return r;
 }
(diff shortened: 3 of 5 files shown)


imap/index.c
```diff
@@ -3110,11 +3111,13 @@ index_copy(struct index_state *state,
 done:
     free(copyargs.records);
     mailbox_close(&destmailbox);
-    for (i = 0; i < msgrecs->count; i++) {
-        msgrecord_t *mr = ptrarray_nth(msgrecs, i);
-        msgrecord_unref(&mr);
+    if (msgrecs) {
```
dilyanpalauzov, line 3114: `msgrecs` cannot be NULL here.

Wouldn’t be sufficient to move the `ptrarray_t *msgrecs = ptrarray_new();` from line 2992 to line 3008?

rsto: @dilyanpalauzov the `goto done` jump in line 3023 could jump over the ptrarray_new in line 3053.

imap/http_caldav_sched.c
```diff
@@ -189,14 +189,15 @@ struct address_t {
 static void add_address(struct address_t **recipients, icalproperty *prop,
                         const char* (*icalproperty_get_address)(icalproperty *))
 {
-    struct address_t *new = xzmalloc(sizeof(struct address_t));
+    struct address_t *new;
```
rsto, line 192: couldn't we move this declaration to its initialization in line 200?

imap/caldav_util.c
```diff
@@ -787,12 +787,13 @@ static int write_personal_data(const char *userid,
     struct message_guid guid;
     struct buf value = BUF_INITIALIZER;
     const char *icalstr = icalcomponent_as_ical_string(vpatch);
-    struct dlist *dl = dlist_newkvlist(NULL, "CALDATA");
+    struct dlist *dl;
```
rsto, line 790: move this to line 796?

rsto (changes requested):

imap/http_caldav_sched.c
elliefm: Yes, but I don't want to, because:

> the goto done jump in line 3023 could jump over the ptrarray_new in line

This sort of thing is a pain to keep track of, and we've fixed enough goto-skips-declaration and goto-skips-initialisation bugs in recent years that I very strongly dislike having declarations littered throughout functions in the C++ style.  They're cute to write in the first place but they're a nuisance to maintain over the long term.

Are there gotos used in this particular case?  No.  But a good habit of declaring variables at the top of the nearest appropriate scope, and then cleaning up (if necessary) at the end of that scope, always avoids this problem, and also makes verifying proper memory management easier.

It's also the usual C way of doing things -- this isn't perl, javascript, or C++, and just because C99 copied a feature from those languages, doesn't make it best practice to use that feature.  (It's also a terrible habit in C++: it only makes sense in languages with garbage collection or other automatic memory management, which neither C nor C++ are.)

imap/index.c
elliefm: `msgrecs` can indeed be NULL here, and one of the Cassandane tests even crashes on it without the conditional

These might be tiny changes, but I did actually spend time thinking about them, and I'd appreciate if you respected that.

imap/http_caldav_sched.c
rsto: Fine with me. I'm not entirely sold on the idea that one or the other way is the "usual C way" (is there any, really?), and I see benefit in both. Declaring every variable on top works best for me when there just a few variables in the whole function ([some say](https://www.kernel.org/doc/html/v4.10/process/coding-style.html#functions) if they exceed 5-10 the function is problematic anyway). Unfortunately that doesn't apply to some epic-scale function definitions in the Cyrus code, so at least declaring non-pointer variables at initialization makes sense there. The function at hand is not one of them, though.

rsto (approved):

imap/http_caldav_sched.c
dilyanpalauzov: The proposed changes do correctly resolve the memory leaks.

I prefer having declaration and definition on the same line as this results less code to read, understand and maintain.  The here exposed code deficiencies can be corrected by less code, than currently proposed.  Less code is better. 

The compilers can hint on many places for incorrect usage of `goto`, like reading there not initialized variables.  Adding extra \_\_attiribute__(()) can help the compilers to provide even more hints.  These hints shall be considered more often than currently.

In my opinion the compilers can detect misuse of variables in the goto destinations, thus having extra precautions, which currently solve no problems, is not necessary: initializing `msgrecs` after the second return, and skipping the changes in the goto-destination produces simpler code.

The bottom line is, that the compiler-found deficiencies shall be listened to more often, instead of writing complicated code, which does not solve existing problems.

Considering that my contributions are often ignored, I can also state that my efforts are not respected.  Does claiming for not being respected change anything?  It does not, thus the argument that a unit of efforts is not respected is irrational one.

ksmurchison (approved):

dilyanpalauzov:
This is relevant also for cyrus-imapd-3.6 and cyrus-imapd-3.4.  The latter has  write_personal_data() in http_caldav.c.

PR #4258, opened 2022-09-27
lcb_internal.c:parse_backup_line() last parameter is never null

dilyanpalauzov:
If it is null, then the memory of dl would be leaked.


ksmurchison:
The backup code has been removed.

PR #4261, opened 2022-09-28
http_jmap.c: open conv.db early for JMAP upload

ksmurchison:
Seeing "deadlock avoided" messages during concurrent JMAP upload and API requests.  Hoping this might fix it


rsto (approved):
LGTM

PR #4263, opened 2022-09-28
caldav_alarm.c: skip alarms for cancelled and declined events

ksmurchison:
Split calendar-user-address-set handling into calsched_support.c so it can be used by caldav_alarm.c when checking declined status


rsto (approved):
LGTM

rsto (approved):
LGTM (assuming the business logic is what we actually want)

ksmurchison:
@brong Were you able to look at this?

PR #4266, opened 2022-09-30
Specialuse: test specialuse_nochildren create behaviour

elliefm:
Adds a test for this behaviour:

> **specialuse_nochildren: \Scheduled \Snooze**
> 
> Whitespace separated list of special-use attributes that may not  contain
> child  folders.   If  set, mailboxes with any of these attributes may not
> have child folders created, and these attributes cannot be added to mail‐
> boxes that already have children..

We thought there might be an edge case allowing creation of mailboxes under the `\Trash` mailbox even though `specialuse_nochildren: \Trash` has been configured.  This test shows that there isn't such an edge case, or at least, not one that's simple to reproduce.

### DIFF
--- a/cassandane/Cassandane/Cyrus/JMAPMailbox.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPMailbox.pm
@@ -77,14 +77,25 @@ sub new
     }, @args);
 }
 
-sub set_up
+sub setup_default_using
 {
     my ($self) = @_;
-    $self->SUPER::set_up();
     $self->{jmap}->DefaultUsing([
         'urn:ietf:params:jmap:core',
         'urn:ietf:params:jmap:mail',
     ]);
+}
+
+sub set_up
+{
+    my ($self) = @_;
+    $self->SUPER::set_up();
+
+    if ($self->{jmap}) {
+        $self->setup_default_using();
+    }
+    # n.b. tests that use :NoStartInstances will need to call
+    # $self->setup_default_using() themselves!
 }
 
 sub getinbox
--- a/cassandane/Cassandane/Cyrus/JMAPMailbox.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPMailbox.pm
@@ -54,6 +54,9 @@ use lib '.';
 use base qw(Cassandane::Cyrus::TestCase);
 use Cassandane::Util::Log;
 
+use lib '../perl/imap';
+use Cyrus::DList;
+
 use charnames ':full';
 
 sub new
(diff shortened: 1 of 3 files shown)


elliefm:
I guess we may want to test the same thing with JMAP too?

ksmurchison (approved):

elliefm:
@rsto I assume Mailbox/set create (or whatever it's called) in JMAP uses the same underlying APIs as IMAP create does?  In that case, I don't think we need to test JMAP for this too, just testing IMAP is enough.  But my assumption might be wrong -- can you think of anything that might work differently, or use different options, or in a different order, and which would warrant adding an equivalent test for JMAP?

elliefm:
@rsto Ken's already approved this, but that was before I added the JMAP test.  Can you please re-review it, and let me know whether you think the JMAP test I've added is adequate to show that it behaves correctly?

The JMAP test is simpler than the IMAP one, because the IMAP test also checks for a possible regression in creating an "intermediate" under Trash (i.e. the "grandchild" stuff).  I don't think it was ever possible to create intermediates in JMAP, so I don't think JMAP needs to be tested against that edge case.  Is that correct?

cassandane/Cassandane/Cyrus/JMAPMailbox.pm
```diff
@@ -5282,4 +5296,64 @@ sub test_mailbox_ignore_notes_subfolders
     $self->assert_str_equals($inboxId, $res->[1][1]{list}[0]{id});
 }
 
+sub test_mailbox_set_create_specialuse_nochildren
+    :min_version_3_7 :needs_component_jmap :NoStartInstances
+{
+    my ($self) = @_;
+
+    $self->{instance}->{config}->set('specialuse_nochildren' => '\\Trash');
```
rsto, line 5304: Why not introduce a new test magic for this? Or even put it into the default config? It looks as if allowing sub mailboxes in Trash is the non-standard case, and this would allow us to catch any other tests that break this assumption.

cassandane/Cassandane/Cyrus/JMAPMailbox.pm
```diff
@@ -5282,4 +5296,64 @@ sub test_mailbox_ignore_notes_subfolders
...
+    $self->_start_instances();
+    $self->_setup_http_service_objects();
+    $self->setup_default_using();
+
+    # set up a Trash folder with \Trash special-use annotation
+    my $imaptalk = $self->{store}->get_client();
+    $imaptalk->create("Trash", "(USE (\\Trash))");
```
rsto, line 5311: I think the JMAP test should use JMAP as much as possible, rather than using IMAP. So this should be

    [Mailbox/set, {
      create => {
        trashmbox => {
           name => 'Trash',
           role => 'trash'
        }
      }
    }, 'R1']

and then you can read the trash mailbox id from the response, rather than parsing the IMAP response as a dlist.

cassandane/Cassandane/Cyrus/JMAPMailbox.pm
```diff
@@ -5282,4 +5296,64 @@ sub test_mailbox_ignore_notes_subfolders
...
+
+    # XXX that syslogs an IOERROR, surprisingly -- ignore it
+    $self->{instance}->getsyslog();
+
+    $self->assert_null($res->[0][1]->{created});
+    $self->assert_deep_equals({ 1 => { type => 'forbidden' } },
+                              $res->[0][1]->{notCreated});
+
+    # what if we remove the annotation
+    $imaptalk->setmetadata("Trash", "/private/specialuse", undef);
```
rsto, line 5340: This could be a `Mailbox/set{update}` where you set `role => undef`

cassandane/Cassandane/Cyrus/JMAPMailbox.pm
```diff
@@ -5282,4 +5296,64 @@ sub test_mailbox_ignore_notes_subfolders
...
+        create => {
+            1 => {
+                parentId => $trash_id,
+                name => 'child',
+            },
+        },
+    }, "R1"]]);
+    $self->assert_not_null($res->[0][1]->{created}->{1}->{id});
+
+    # should not be able to add the annotation back
```
rsto, line 5354: This could be a `Mailbox/set{update}` where you set `role => 'trash'`

rsto (changes requested):
The Cyrus change looks good to me but I would prefer the JMAP test to use more of JMAP methods.

cassandane/Cassandane/Cyrus/JMAPMailbox.pm
elliefm: The default value for specialuse_nochildren is `\Scheduled \Snoozed`, which doesn't include \Trash.  If we set \Trash in the default config, then every test that only expects the default set would then need some override to remove \Trash.

I could add a magic for it, but since it's only used by two tests, it doesn't feel worth the effort yet -- my usual threshold is when I find myself doing the same thing the same way a third time, which I haven't yet.  It's tricky anyway, because -- should the magic set "only \Trash" (as these tests do, which don't care about \Scheduled or \Snoozed), or "add  \Trash to the set"?  And will we remember to update the magic in the future if the default set changes?  Until we need it a third time, and know what shape of thing we regularly need, I don't think it's worth the thought.

When we run the tests from in-a-box with the Fastmail config in cassandane.ini, it should have \Trash included in the set because that's the configuration we use, so that will catch any tests that are making bad assumptions.

cassandane/Cassandane/Cyrus/JMAPMailbox.pm
elliefm: Oh great, thanks.  I wasn't sure how to do these operations over JMAP, and I already had IMAP equivalents in the other test, so it was easy enough to just do the same again.  But I'll have a go at changing them to use pure JMAP.

cassandane/Cassandane/Cyrus/JMAPMailbox.pm
elliefm: It looks like this might not be allowed:

```
      'notUpdated' => {
        '21c5642b-63b7-4bd3-8149-22aea5229f23' => {
          'description' => 'The final mailbox state is invalid due to role conflicts:
A mailbox had protected specialuse \\Trash but this specialuse is missing in the new mailbox state.',
          'properties' => [
            'role',
          ],
          'type' => 'invalidProperties',
        },
      },
```

Do I need to do something special to remove the trash role?

cassandane/Cassandane/Cyrus/JMAPMailbox.pm
rsto: My threshold for adding new test magics is considerably lower, but I do not object to the current implementation.

cassandane/Cassandane/Cyrus/JMAPMailbox.pm
rsto: Now that I think of it: that's because the `specialuse_protect` imapd.conf option does not allow to delete `Trash`. I am surprised this works over IMAP.

elliefm: Ohhh that's interesting.

elliefm: We talked about this and decided to just use IMAP for this stage after all, because working around the more restrictive JMAP behaviour would just add complexity that wasn't relevant to the behaviour being tested.

elliefm:
@rsto My most recent update returns to using IMAP to add/remove the \Trash special-use, like we talked about.

Now that I'm not tripping over that, I think I've found an actual problem: at the very end of the test, JMAP seems to be able to add the "trash" role to the Trash mailbox, even though the Trash mailbox has a child mailbox.  In IMAP a few lines earlier, this was forbidden due to specialuse_nochildren, but it seems like JMAP is allowing it?

Though I might be misinterpreting the response object, or maybe the response object is badly formed, I'm not sure.  I was expecting to see `$trash_id` in `...->{notUpdated}`, but instead it's in `...->{updated}`.  If you run this test with `-vvv -fpretty` it'll log (and highlight) the response object.

What do you think?

rsto:
@elliefm yes, this turns out to be a bug in JMAP. Do you want to fix it in the PR or should I have a go at it?

elliefm:
@rsto I don't have any particular insight as to how or where to fix it.  It might be a good learning exercise for me, but I don't know what the urgency is.  If there's urgency, it's probably quicker for you to do it, unless you don't have time.  I guess IETF next week complicates things?

elliefm:
@rsto Great, thanks.  I don't want to merge this PR while it contains tests that will fail, but if we do it as two separate PR's, that's an annoying juggle.

So how about:

1) I'll clean up the remaining WIP commits on this branch so the history is neat
2) You cherry pick these commits onto your own branch where you're fixing the bug
3) I'll then close this PR without merging it
4) The tests and the fix get merged together via your PR

elliefm:
Okay I've done that now, but I've left the JMAP test's commit as "WIP ...", because I couldn't finish off the last few lines of verifying the correct error, since I don't know what the eventual error would be.  So I guess just amend that commit on your branch once the details are known

rsto:
Thanks, I'll work on it soon.

elliefm:
I've rebased this to tidy up the history (which still had a WIP commit), and also rebased it on master since it was pretty out of date

ksmurchison (approved):

rsto (approved):

PR #4269, opened 2022-10-02
http_webdav.c: memory leaks

dilyanpalauzov:
Should my_webdav_auth() call in addition `mboxname_release(&namespacelock);` after the herein inserted `mbname_free(&mbname);`?


ksmurchison:
@dilyanpalauzov Thanks for this.  I'm going to fix my_webdav_auth() a little differently with this https://github.com/cyrusimap/cyrus-imapd/pull/4276

PR #4272, opened 2022-10-03
backup_read_message_data() — memory leak

dilyanpalauzov:
and silence two dead initialization warnings.


ksmurchison:
The backup code has been removed.

PR #4273, opened 2022-10-03
jmap_mail: do not add html body content in plain text preview

rsto:
This only affects installations that do not set email preview with external annotators.

HTML content could go verbatim into a supposedly plain text preview. Now, it only gets returned as HTML text which in turn gets rewritten into plain text for preview generation.


ksmurchison (approved):
LGTM

PR #4274, opened 2022-10-03
append.c: send new message mboxevents before modseq events

ksmurchison:


brong (approved):
Yeah this looks great, thanks.  Seems perfectly safe to do this.

PR #4275, opened 2022-10-04
http_dav.c, mailbox.c: assert that we have a non-NULL resource (for debugging)

ksmurchison:


rsto (approved):
LGTM for debugging. In general, mght want to handle this differently rather than crashing with an assert.

Issue #4216, opened 2022-08-22
httpd killed by SIGSEGV for calendar request


sdalu:
imapd version: 3.4.4
OS: FreeBSD 13.1

config fragment:
~~~
httpmodules: caldav carddav freebusy ischedule jmap
httpprettytelemetry: yes
caldav_realm: foo.com
carddav_realm: foo.com
calendar_user_address_set: foo.com
caldav_allowcalendaradmin: 1
~~~

When trying to access a calendar using:

~~~
curl -u xxx:yyy http://<server>/dav/calendars/user/<user>/793edf22-716-1497341807796-172307
~~~

Backtrace
~~~
root@incal:/usr/ports/devel/libical # lldb -c /var/imap/httpd.core /usr/local/cyrus/libexec/httpd
(lldb) target create "/usr/local/cyrus/libexec/httpd" --core "/var/imap/httpd.core"
Core file '/var/imap/httpd.core' (x86_64) was loaded.
(lldb) bt
This version of LLDB has no plugin for the language "assembler". Inspection of frame variables will be limited.
* thread #1, name = 'httpd', stop reason = signal SIGSEGV
  * frame #0: 0x00000008019fa8f0 libc.so.7`strcmp at strcmp.S:60
    frame #1: 0x0000000801671445 libical.so.3`icalcomponent_isa_component(component=0x0000000800000000) at icalcomponent.c:333:9
    frame #2: 0x0000000801670957 libical.so.3`icalcomponent_add_children(impl=0x000000080244daf0, args=0x00007fffffff9c80) at icalcomponent.c:76:9
    frame #3: 0x0000000801670898 libical.so.3`icalcomponent_vanew(kind=ICAL_VCALENDAR_COMPONENT) at icalcomponent.c:133:5
    frame #4: 0x0000000800b4740f libcyrus_imap.so.0`icalcomponent_new_stream(mailbox=0x0000000802571408, prodid="-//CyrusIMAP.org/Cyrus 3.4.4//EN", name="Work", desc=0x0000000000000000, color="") at ical_support.c:532:12
    frame #5: 0x000000000023ecd9 httpd`begin_icalendar(buf=0x00007fffffffd5f0, mailbox=0x0000000802571408, prodid="-//CyrusIMAP.org/Cyrus 3.4.4//EN", name="Work", desc=0x0000000000000000, color="") at http_caldav.c:1801:12
[...]

PR #4278, opened 2022-10-05
Last parameter of icalcomponent_vanew() must be NULL, not 0

dilyanpalauzov:
https://github.com/libical/libical/issues/585

Close https://github.com/cyrusimap/cyrus-imapd/issues/4216


rsto (approved):
Thanks!

PR #4286, opened 2022-10-07
github actions: get tags from upstream

elliefm:
The build process depends on having access to the release tags to determine the version, and many of our tests need to examine the cyrus version, and will fail in annoying ways if the version is missing or incorrect.  When github actions run from a fork, rather than the main repository, our release tags are probably missing, and so a bunch of tests fail (unless the fork maintainer bothers to keep their fork updated with our release tags for some reason, but they shouldn't have to).

This updates it to fetch git tags directly from the cyrusimap/cyrus-imapd repository, even if the action is running from a fork.

It also updates the multiline run commands to use "shell: bash", which allegedly causes them to be run in bash's fast-fail mode (`set -eo pipefail`).  This should mean that if any command fails the whole job fails, whereas the default is apparently to ignore the exit status of all commands in a `run` except the last one.  Might be useful in the future when debugging github actions stuff.

I'm not sure if we still need the `git fetch --unshallow`, maybe it's redundant and can be removed now?  I read the documentation for unshallow and didn't get any closer to a decision, so I've left it alone for now.

I don't yet know if these changes will break when run from the main repo, but we'll find out after I click "Create pull request"...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,8 @@ jobs:
     - name: configure and build
       run: |
         git fetch --unshallow
-        git fetch --tags
+        git remote add upstream https://github.com/cyrusimap/cyrus-imapd.git
+        git fetch --tags upstream
         ./tools/build-with-cyruslibs.sh
     - name: update jmap test suite
       run: |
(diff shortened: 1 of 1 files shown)


rjbs (approved):
LGTM

PR #4289, opened 2022-10-10
tools/compile_st.pl: strip 'Command-line' from gperf output

elliefm:
We version control the output file from this command (as `imap/rfc822_header.c`), and having the name of the tempfile used to generate it embedded in the output made it show up as different every time it was regenerated, even though usually the only difference was the name of the tempfile.

Now it will only show up as different when something interesting changed.

The file in question is only regenerated in maintainer mode, so you probably don't notice it unless you're doing releases, but it's been annoying me for years...


ksmurchison (approved):
LGTM

Issue #3917, opened 2022-02-15
Sieve enotify (RFC 5435) implementation bugs


MartySvec:
Hello,

I think there're several bugs in sieve "enotify" extension implementation which cause this feature to be **completely broken**. Consider these two rules, one using old "notify" draft and the second one new "enotify" extension:

`if allof (header :contains "subject" "notifytest")
{
        notify :method "mailto" :options ["recipient@somedomain.com"] :message "Notification body";
}`

`if allof (header :contains "subject" "notifytest")
{
        notify :importance "2" :from "notifysender@somedomain.com" :message "Notification body" "mailto:recipient@somedomain.com";
}`

According to `strace`, notification requests based on these rules are submitted to `notifyd` process as follows:

`recvfrom(0, "mailto\0SIEVE\0normal\0mailboxid@somedomain.com\0\0001\0recipient@somedomain.com\0Notification body\0/var/spool/imap44/stage./4400-1644860274-0\0"...`

`recvfrom(0, "mailto:recipient@somedomain.com\0SIEVE\0normal\0mailboxid@somedomain.com\0\0000\0Notification body\0/var/spool/imap44/stage./20883-1644913849-0\0" ...`

Clearly, the second request sends "method" in the form of `method:address` which is not a valid method recognized by `notifyd` and no notification email is sent. Also, the `:from` address seems to be ignored at all.

I looked into the history and guess that these bugs date back to commit [adb4819b3e8c4e51be40741fc523f26bf3e6f095](https://github.com/cyrusimap/cyrus-imapd/commit/adb4819b3e8c4e51be40741fc523f26bf3e6f095):

1. This commit ignores the fact that enotify method string contains recipient email address and saves it to `c->u.n.method` without stripping `:recipient@somedomain.com`.
2. Notifyd code expects recipient address in `options[0]` but `build_notify()` does not handle it in `B_ENOTIFY` branch.
3. Notifyd code doesn't support `:from` address at all. This address is probably ignored somewhere in the call chain before notifyd daemon is called.

IMO sieve "enotify" support should either correctly populate data structures as expected by the old "notify" draft implementation, or the whole notify callchain including notifyd daemon should be rewritten according to the new "enotify" extension standard.

I noticed these issues in 3.4.3 code but master branch seems to be affected to.

Best regards,
Martin

MartySvec:
Nobody cares?

elliefm:
[...]

PR #4290, opened 2022-10-10
Parse Sieve enotify method URI

ksmurchison:
This is a fix for issue #3917

### DIFF
--- a/sieve/bc_eval.c
+++ b/sieve/bc_eval.c
@@ -2036,6 +2036,7 @@ int sieve_eval_bc(sieve_execute_t *exe, int *impl_keep_p, sieve_interp_t *i,
         case B_ENOTIFY:
         case B_NOTIFY:
         {
+            const char *method = cmd.u.n.method;
             const char *message = cmd.u.n.message;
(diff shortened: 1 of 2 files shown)


sieve/bc_eval.c
```diff
@@ -2055,7 +2056,18 @@ int sieve_eval_bc(sieve_execute_t *exe, int *impl_keep_p, sieve_interp_t *i,
                 message = parse_string(message, variables);
             }
 
-            res = do_notify(notify_list, cmd.u.n.id, cmd.u.n.from, cmd.u.n.method,
+            if (op == B_ENOTIFY) {
+                /* Parse/split the method URI */
+                if (!strncmp(method, "mailto:", 7)) {
```
rsto, line 2061: should that be `strcasecmp` or does the Sieve spec require the URI scheme to be lower-case?

ksmurchison: Yeah, case-insensitive makes sense.

rsto (approved):

elliefm (approved):

PR #4294, opened 2022-10-12
Force MIME-encode SUMMARY when creating Subject header field

ksmurchison:
This fixes an issue where embedded newlines in the SUMMARY property would cause us to create a Subject header which would prematurely signal the end of the header section of our constructed message.  This resulted in our message parser never reaching the Content-Disposition header field, thereby NOT parsing our CalDAV resource name and subsequently violating the resource NOT NULL constraint in the ical_objs SQL table.

### DIFF
--- a/lib/charset.c
+++ b/lib/charset.c
@@ -3507,7 +3507,9 @@ EXPORTED char *charset_encode_mimebody(const char *msg_base, size_t len,
 }
 
 
-#define ATOM_SPECIALS  "()<>[]:;@\\,.\" \t"
+#define qp_isspace(c) ((c) == ' ' || (c) == '\t')
+
+#define ATOM_SPECIALS  "()<>[]:;@\\,.\" \t\r\n"
 
 /* Find the first email address (addr-spec) in data */
 static const char *find_addr(const char *data, size_t datalen, size_t *addrlen)
@@ -3532,7 +3534,7 @@ static const char *find_addr(const char *data, size_t datalen, size_t *addrlen)
         /* domain-literal MUST end with ']' */
         if (*e++ != ']') return NULL;
     }
-    else if (!isspace(*e) && !strchr(ATOM_SPECIALS, *e)) {
+    else if (!strchr(ATOM_SPECIALS, *e)) {
         /* find end of dot-atom */
         while (++e < end && (*e == '.' || !strchr(ATOM_SPECIALS, *e)));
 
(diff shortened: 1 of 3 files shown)


lib/charset.c
```diff
@@ -3729,17 +3729,15 @@ static char *qp_encode(const char *data, size_t len, int isheader,
         /* fold header every 78 characters (if possible) */
         size_t i = 0, j = 0, last_wsp = 0;
 
-        while ((len - i > 78) && (j < len)) {
-            j += strcspn(data + j, " \t");
-
+        while ((len - i > 78) && (j += strcspn(data + j, " \t")) < len) {
```
rsto, line 3732: A test in charset.testc would be great for this. I think there already are tests for qpencode that could be extended.

rsto (approved):

lib/charset.c
ksmurchison: The line quoted above was actually incorrect, which was discovered with the added cunit tests

dilyanpalauzov:
Which software does insert new lines in the summary?

I would prefer if the summaries are not truncated, but the new lines are substituted, e.g. with a space, or Down Arrow With Corner Left ↵.

dilyanpalauzov:
If it is only about the Subject: header in the `number.` files, then it does not really matter, if the subject is truncated.

Can the Subject: and MIME-Version lines there also completely disappear?

rsto (approved):

ksmurchison:
Good point about needing to also handle iMIP generation.  I've switched from truncating the SUMMARY to force MIME-encoding the entire thing (minus any trailing CR/LF)

dilyanpalauzov:
The test validates `$self->assert_matches(qr/geo:-37.810551,144.962840/, $newcard);` . It shall validate something about the new lines in the SUMMARY.

imap/http_caldav_sched.c
```diff
@@ -348,7 +348,13 @@ static int imip_send_sendmail(const char *userid, icalcomponent *ical, const cha
...
     buf_printf(&msgbuf, "Subject: %s: ", filename);
     if (summary) {
-        char *mimehdr = charset_encode_mimeheader(summary, 0, 0);
+        size_t len = strlen(summary);
+
+        /* Trim any trailing CR and/or LF */
+        while (strchr("\r\n", summary[len-1])) len--;
+
+        /* force=1 to encode any non-folding LF */
+        char *mimehdr = charset_encode_mimeheader(summary, len, 1 /* force */);
```
rsto, line 357: doesn't force_quote cause all string values to be encoded, even sane short ASCII-only ones?

rsto (commented):
Agree with @dilyanpalauzov that the test needs to get updated. If my comment about force_quote is moot, then consider it approved.

imap/http_caldav_sched.c
ksmurchison: Yes.  It appears as if non-folding LF don't get captured in our checks, which why I used force.  I didn't want to change that check for this specialty case

ksmurchison:
Test has been updated

rsto (approved):
LGTM. Too bad this is such a pain to test.

ksmurchison:
Test added

PR #4297, opened 2022-10-16
Makefile.am: remove duplicate headers

dilyanpalauzov:
The reason to include `.h` files in _SOURCES is to get the .h files in the tarball created by  make dist .  It is sufficient to reference the `.h` files once in any _SOURCES primary.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -852,12 +852,10 @@ backup_backupd_LDADD = backup/libcyrus_backup.la $(LD_SIEVE_ADD) $(LD_SERVER_ADD
 backup_ctl_backups_SOURCES = \
     imap/mutex_fake.c \
     imap/sync_support.c \
-    imap/sync_support.h \
     backup/ctl_backups.c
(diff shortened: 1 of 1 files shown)


elliefm (commented):
This seems fine in theory (though I haven't tested it...)

But -- I don't think people really write Makefile.am rules, or fully understand what they're writing when they do.  I think people mostly find something similar elsewhere in the file, and copy-paste-modify until it seems to work.

When the _SOURCES rules consistently include the header files, someone setting up a new rule with new files is likely to copy that consistency, and get it right.  

When each header file is included in only one _SOURCES rule, that's subtle.  Someone setting up a new rule with new files is likely to not notice a detail like that, and forget to include the header in any rule.  This will work for normal builds, so the problem won't get noticed until someone tries to `make dist` -- probably me.

I'm a little tired of fixing this specific type of problem, so I'd rather not make a change that, while technically correct, will mostly just make this type of problem more common.  This is a situation where redundancy is useful.

Issue #3068, opened 2020-06-16
reconstruct -m is "currently unavailable"


elliefm:
Quoth the man page:

> reconstruct  rebuilds  one  or  more  IMAP  mailboxes.  When invoked with the **-m**
> switch, it rebuilds the master mailboxes file.  It can be used to  recover  from
> almost any sort of data corruption.

and

> **-m**     NOTE: CURRENTLY UNAVAILABLE
> 
> Rebuild the mailboxes file.  Use whatever data in the existing  mailboxes
> file   it   can  scavenge,  then  scans  all  partitions  listed  in  the
> imapd.conf(5) file for additional mailboxes.

It's been like this since at least 2.5.  I don't know what the problem was, and don't know whether it's been fixed (and the documentation neglected) or if it's still broken.

We should probably figure it out, and fix it!

If the problem is that the world has changed so much that the `-m` option is no longer feasible to implement, then we should remove it from the documentation.

elliefm:
Looks like the "CURRENTLY UNAVAILABLE" text was added back in 2002 in f04967583c.  @ksmurchison, this commit has your name on it, I don't suppose there's any chance you remember anything about it?

ksmurchison:
I don't recall if there was a technical reason why the functionality went away or something else.  2002 was before by CMU days so I doubt that I did anything other than document that -m no longer works.

This *may* have coincided with the introduction of Cyrus Murder, but that's just a guess.

elliefm:
Ahh, the previous`-m` implementation was thrown out in 3f96ad229da (in 2000).  Unhelpfully, the commit message only talks about changing the behaviour of `-r`.  Not sure if the loss of `-m` was a mistake or what.

The old implementation was expecting the mailboxes file to be a plain text file, not a cyrusdb like it is now, so there's probably nothing to be gained by trying to bring the old implementation back.

I guess the question now is, is this functionality useful enough to rewrite from scratch?  It's been missing for over 20 years, so if anyone's cared, they haven't raised enough of a stink for anything to happen.

If we do want to rewrite this, I propose postponing it until after mailboxes-by-uuid lands, but then actually do it so that it's done.

If we don't want to rewrite it, then let's just throw the last few shreds in the bin now and forget about it.

rjbs:
To be addressed during CLI overhaul.

PR #4298, opened 2022-10-17
Add long-options support to ctl_cyrusdb and reconstruct

elliefm:
This finishes the work started in #4196, by adding long-options support to the  `ctl_cyrusdb` and `reconstruct` command line tools.

This is one of those annoying ones where the lion's share of changes are in strings, which the compiler can't check for spelling or consistency.  We also don't have any tests for the majority of these commands!  So please review with a very close eye.  I'd suggest stepping through commit-by-commit.

I have tried to not change any existing behaviour, but simply add long-option synonyms for the existing short options.

I have _not_ added a long-option synonym for `-C altconfig`, because that needs to remain consistent with the service binaries, and this work does not touch the service binaries.

### DIFF
--- a/docsrc/imap/reference/manpages/systemcommands/reconstruct.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/reconstruct.rst
@@ -140,8 +140,9 @@ Options
 
 .. option:: -o, --ignore-odd-files
 
-    Ignore odd files in your mailbox disk directories.  Probably useful
-    if you are using some tool which adds additional tracking files.
+    Ignore odd files in your mailbox disk directories, instead of
+    complaining about them.  Probably useful if you are using some
+    tool which adds additional tracking files.
 
 .. option:: -O, --delete-odd-files
 
--- a/docsrc/imap/reference/manpages/systemcommands/reconstruct.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/reconstruct.rst
@@ -65,17 +65,17 @@ Options
 
 .. program:: reconstruct
 
-.. option:: -C  config-file
+.. option:: -C config-file
 
     |cli-dash-c-text|
 
-.. option:: -p  partition
+.. option:: -p partition, --partition=partition
 
     Search for the listed (non-existant) mailboxes on the indicated
     *partition*. Create the mailboxes in the database in addition to
     reconstructing them. (not compatible with the use of wildcards)
 
-.. option:: -x
+.. option:: -x, --ignore-disk-metadata
 
     When processing a mailbox which is not in the mailbox list (e.g.
     via the **-p** or **-f** options), do not import the metadata from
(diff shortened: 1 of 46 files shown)


imap/ctl_cyrusdb.c
```diff
@@ -306,9 +306,21 @@ int main(int argc, char *argv[])
     char *msg = "";
     int i, rotated = 0;
 
-    r = r2 = 0;
+    /* keep this in alphabetical order */
+    static const char *const short_options = "C:crx";
```
dilyanpalauzov, line 310: With `static const char short_options[]` the variable will be put in a read-only segment in the final executable.

docsrc/imap/reference/manpages/systemcommands/reconstruct.rst
```diff
@@ -138,33 +138,33 @@ Options
...
     exhausted other options first!
 
-.. option:: -o
+.. option:: -o, --ignore-odd-files
 
     Ignore odd files in your mailbox disk directories.  Probably useful
     if you are using some tool which adds additional tracking files.
 
-.. option:: -O
+.. option:: -O, --delete-odd-files
```
dilyanpalauzov, line 146: `--ignore-odd-files` and `--delete-odd-files` are mutually exclusive.  Which of them is the default?

docsrc/imap/reference/manpages/systemcommands/reconstruct.rst
```diff
@@ -65,70 +65,70 @@ Options
...
 
-.. option:: -G
+.. option:: -G, --force-reparse
 
     Force re-parsing of the underlying message (checks GUID
     correctness). Reconstruct with -G should fix all possible
     individual message issues, including corrupted data files.
 
-.. option:: -I
+.. option:: -I, --update-uniqueids
```
rsto, line 117: Should this rather be named `dedup-uniqueids`?

imap/reconstruct.c
```diff
@@ -159,7 +157,35 @@ int main(int argc, char **argv)
...
+        { "force-reparse", no_argument, NULL, 'G' },
+        { "update-uniqueids", no_argument, NULL, 'I' },
+        { "prefer-mboxlist", no_argument, NULL, 'M' },
+        { "delete-odd-files", no_argument, NULL, 'O' },
+        { "header-paths", no_argument, NULL, 'P' },
+        { "guid-mismatch-keep", no_argument, NULL, 'R' },
+        { "guid-mismatch-discard", no_argument, NULL, 'U' },
+        { "set-version", required_argument, NULL, 'V' },
+        { "scan-filesystem", no_argument, NULL, 'f' },
+        { "dry-run", no_argument, NULL, 'n' },
```
rsto, line 174: I might have missed the explanation: why does the `m` option not have a long option? Is that for the same reason it's not mentioned in the documentation?

rsto (commented):
This mostly looks good to me, just have 2 questions.

imap/ctl_cyrusdb.c
elliefm: Is that a problem?

docsrc/imap/reference/manpages/systemcommands/reconstruct.rst
elliefm: No idea, maybe neither.  I'm not trying to rewrite the documentation here, but I'll have a look and maybe tweak it a little.

At a guess, I think the default is probably to complain about the odd files, in which case ignore probably shuts up the complaints, whereas delete removes them.  But that's just a guess.

docsrc/imap/reference/manpages/systemcommands/reconstruct.rst
elliefm: Hmmm... it's hard to say.  I think "dedup-uniqueids" might accidentally imply that it's a system-wide deduplication.  But it's not -- it can only deduplicate them for mailboxes it has visited.  The documentation does make that clear, but the risk with adding long options is that people assume they know what it means from the long option's name, and then don't read the documentation.  So I'm trying to find names that don't accidentally imply the wrong thing.

In that sense, I think the "update-uniqueids" is vague enough that someone who doesn't already know what it means needs to look it up, but memorable enough that someone who does know doesn't need to look it up again.  Maybe...

It's also complicated in that duplicate uniqueids should be impossible under uuid-mailboxes, since how can two mailboxes exist in the same location on disk?  Of course, they can still be duplicated for pairs of older mailboxes that haven't been relocate_by_id'd yet (or for systems that are using mailbox_legacy_dirs).

imap/reconstruct.c
elliefm: Huh, I guess I just missed it -- probably because it's not documented.  I'll figure out what it does, and then do something with it.  Thanks!

elliefm: Oh, `-m` is this thing: #3068 -- it's a long-dead option, which we haven't yet quite decided whether to just get rid of it completely, or rewrite the long-missing functionality.

There's no documentation anymore cause that was cleaned up in 56b067a1d65852e7116eb717e94866bb49c0b35f

I'm thinking I'll leave it alone in this PR.  Some future PR will either bring it back properly, with a suitable long option, or discard the remnants forever.

docsrc/imap/reference/manpages/systemcommands/reconstruct.rst
elliefm: My guess was right: by default, it complains about odd files.  Ignore makes it not complain; delete makes it delete them.  I've tweaked the documentation a little to hopefully make this more clear.

imap/ctl_cyrusdb.c
elliefm: Oh, I misread: you were suggesting that I should change it to that.  That's a good suggestion, thanks.

elliefm:
@rsto I've updated this to take into account @dilyan's `short_options[]` suggestion, and applied the same change to all the other tools too, so that they're consistent when we inevitably copy and paste into new tools

I don't think there was anything else required from your last comments?

rsto:
@ellie all good from my side

rsto (approved):

PR #4299, opened 2022-10-19
jmap_mail: limit calendarEvents to at most 8 unique VEVENT UIDs

rsto:
Recent commit 1e13dd7de49257c69df5d158b45b9b94302d012f made Cyrus not return any entry in Email.calendarEvents for calendar attachments that contain VEVENTs with more than one unique UID (as iTIP does not allow multiple UIDs in a single iTIP message).

This has turned out to be an issue for some calendar attachments that are not an iTIP message but contain a small number of VEVENTs, for example, calendar entries for a train trip and return trip.

Until Cyrus can more efficiently handle largish calendar attachments, this limits the number of VEVENT UIDs in a single attachment to 8.


ksmurchison (approved):
LGTM

PR #4300, opened 2022-10-19
mailbox_iter_startuid() needs to initialize recno to 1 rather than 0 (zero)

ksmurchison:
#4253 changed mailbox_iter_inti() to initialize iter->recno to 1 rather than zero, but the same wasn't done for mailbox_iter_startuid()

This was causing messages to not get Xapian indexed (amongst other possible bugs) and leading to Cassandane failures


brong (approved):
The two types of hard problems: naming, cache invalidation, and off by one errors

PR #4302, opened 2022-10-20
Fixed CALDAV PUT requests while proxying

cmollekopf:
txn->req_tgt.namespace->mboxtype is MBTYPE_REMOTE in the proxy.

We have run into issues with proxying caldav requests. This patch allows a PUT request to be successfully proxied, however, we also noticed that autocreate is broken while proxing. Are you interested in getting this fixed, or are there better alternatives to proxying for CALDAV in a murder setup?

### DIFF
--- a/imap/http_dav.c
+++ b/imap/http_dav.c
@@ -7097,7 +7097,7 @@ int meth_put(struct transaction_t *txn, void *params)
     }
 
     /* Make sure mailbox type is correct */
-    if (txn->req_tgt.mbentry->mbtype != txn->req_tgt.namespace->mboxtype)
+    if (txn->req_tgt.mbentry->mbtype != txn->req_tgt.namespace->mboxtype && txn->req_tgt.mbentry->mbtype != MBTYPE_REMOTE)
         return HTTP_FORBIDDEN;
 
     /* Make sure Content-Range isn't specified */


imap/http_dav.c
```diff
@@ -7097,7 +7097,7 @@ int meth_put(struct transaction_t *txn, void *params)
     }
 
     /* Make sure mailbox type is correct */
-    if (txn->req_tgt.mbentry->mbtype != txn->req_tgt.namespace->mboxtype)
+    if (txn->req_tgt.mbentry->mbtype != txn->req_tgt.namespace->mboxtype && txn->req_tgt.mbentry->mbtype != MBTYPE_REMOTE)
         return HTTP_FORBIDDEN;
```
dilyanpalauzov, line 7101: Does this change does still permit PUT in non-proxied setups?

cmollekopf: Yes, it simply relaxes the mbtype check to also allow MBTYPE_REMOTE, which is what we get in the proxy.
This then allows the request to be proxied, but shouldn't affect non-proxied requests at all.

dilyanpalauzov: I do not think this relaxes the mbtype checks.  Before this change requests with `txn->req_tgt.mbentry->mbtype == txn->req_tgt.namespace->mboxtype == MBTYPE_CALENDAR` can be accepted, after the change such requests will get `403 FORBIDDEN` answer.

Also `txn->error.desc` shall be set on failure to return over CalDAV a text explanation why the request was rejected.

dilyanpalauzov: > Before this change requests with `txn->req_tgt.mbentry->mbtype == txn->req_tgt.namespace->mboxtype == MBTYPE_CALENDAR` can be accepted, after the change such requests will get 403 FORBIDDEN answer.

This is not correct.

cmollekopf: Before the change the entry had to match txn->req_tgt.namespace->mboxtype, otherwise a HTTP_FORBIDDEN was returned. After the change the check is effectively disabled if the entries type is MBTYPE_REMOTE.

So the change seems reasonable to me, but there may be a better approach?

ksmurchison:
This PR definitely highlighted a bug.  I took a different path in fixing it in #4437.  Closing this PR in lieu of the alternate fix.

PR #4303, opened 2022-10-21
vcard_support.c: properly translate v4 (uri) <-> v3 (free-form text) UIDs

ksmurchison:


rsto (commented):
Is this conversion necessary? RFC 6350 allows a UID to be both a URI or free-form text value.

If it is necessary: stripping off the NID part of the urn value for v4->v3 conversion is lossy. Can't the UID value for the v3 version stay a full "urn:" URI?

rsto (approved):

PR #4308, opened 2022-10-24
prevent cassandane message id clashes

elliefm:
Apparently Cassandane tests can fail randomly when the message-id generator duplicates a message-id within a single test.

The generator only uses 16 bits of randomness, which isn't much.  I thought about using more random bits, but a) that still wouldn't guarantee no repeats, and b) these only need to be unique within a single test anyway.  Instead, I added a monotonic counter to the generator.  Within a single test, each message-id is now guaranteed unique just by virtue of the counter, even if the random number generator spits out two identical numbers within the same clock second.

A neat side effect of using the counter is that the sequence messages were generated in is now encoded into the message-id, which might be useful for postmortem debugging.

### DIFF
--- a/cassandane/Cassandane/Generator.pm
+++ b/cassandane/Cassandane/Generator.pm
@@ -188,7 +188,11 @@ sub _generate_messageid
 {
     my ($self, $params) = @_;
     my $idsalt = int(rand(65536));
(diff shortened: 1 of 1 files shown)


ksmurchison (commented):
LGTM

rsto (approved):

PR #4309, opened 2022-10-24
Fix invalidResultReference error (as mentioned on IRC!)

brong:
This was posted to IRC, and it's clearly correct.  Doh.


brong:
Interestingly, this fails JMAPTestSuite because it is looking for the wrong thing.  Ouch.

rsto (approved):

PR #4311, opened 2022-10-26
vcard_support.c: can't use reference into ventry->v.value in vparse_set_value()

ksmurchison:
ventry->v.value gets freed before setting the new value, so our reference into the old value is now bogus and subject to segfaults


rsto (approved):

imap/vcard_support.c
```diff
@@ -302,28 +302,34 @@ EXPORTED void vcard_to_v3(struct vparse_card *vcard)
...
-                    if (base64) {
-                        *base64++ = '\0';
-                        vparse_add_param(ventry, "ENCODING", "b");
-                    }
+                    data = base64 + 7;
+                    typelen = base64 - type;
+                }
+                else {
+                    data = strchr(type, ',');
+                    typelen = data - type;
```
rsto, line 318: this can crash if data == NULL

rsto (changes requested):

imap/vcard_support.c
ksmurchison: Fixed

rsto (approved):

PR #4312, opened 2022-10-27
jmap_ical: omit trailing zeros from geo coordinates

rsto:
This isn't strictly necessary, but makes the regression tests run on both the current and upcoming libical release.

### DIFF
--- a/cassandane/tiny-tests/JMAPCalendars/calendarevent-get-locations-geo
+++ b/cassandane/tiny-tests/JMAPCalendars/calendarevent-get-locations-geo
@@ -11,5 +11,5 @@ sub test_calendarevent_get_locations_geo
     my $event = $self->putandget_vevent($id, $ical);
     my @locations = values %{$event->{locations}};
     $self->assert_num_equals(1, scalar @locations);
-    $self->assert_str_equals("geo:37.386013,-122.082930", $locations[0]{coordinates});
+    $self->assert_str_equals("geo:37.386013,-122.08293", $locations[0]{coordinates});
 }
--- a/imap/jmap_ical.c
+++ b/imap/jmap_ical.c
@@ -2909,20 +2909,26 @@ static json_t* location_from_ical(icalproperty *prop, json_t *links,
 static json_t *coordinates_from_ical(icalproperty *prop)
 {
     /* Use verbatim coordinate string, rather than the parsed ical value */
-    const char *p, *val = icalproperty_get_value_as_string(prop);
+    const char *val = icalproperty_get_value_as_string(prop);
     struct buf buf = BUF_INITIALIZER;
     json_t *c;
 
     if (!val) return NULL;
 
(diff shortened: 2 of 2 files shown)


cassandane/tiny-tests/JMAPCalendars/calendarevent-get-locations-geo
```diff
@@ -11,5 +11,5 @@ sub test_calendarevent_get_locations_geo
     my $event = $self->putandget_vevent($id, $ical);
     my @locations = values %{$event->{locations}};
     $self->assert_num_equals(1, scalar @locations);
-    $self->assert_str_equals("geo:37.386013,-122.082930", $locations[0]{coordinates});
+    $self->assert_str_equals("geo:37.386013,-122.08293", $locations[0]{coordinates});
```
dilyanpalauzov, line 14: This is not a real-world example.  Six digits after the decimal point mean precision of 11cm, which precision is not available for mortals.  Five digits after the decimal point are realistic. 

With one digit less `geo:…,-122.08290` means precision of 1.1m, while `geo:…,-122.0829` means precision of 11m.  So the meaning is slightly changed.

rsto: Thanks for the info. I get mixed impressions here: e.g. Open Street Map supports your point about precision

https://wiki.openstreetmap.org/wiki/Precision_of_coordinates#Display_and_measurement_accuracy

But the `geo:` URI RFC 5870 states in [section 3.4](https://datatracker.ietf.org/doc/html/rfc5870#section-3.4.3):

> Note: The number of digits of the values in <coordinates> MUST NOT be interpreted as an indication to the level of uncertainty.

and following section 3.4.4 states

> The term "mathematically identical" used below specifies that some components of the URI MUST be compared as normalized numbers rather than strings to account for the variety in string representations of identical numbers (for example, the strings "43.10" and "43.1" are different, but represent the same number).

Still, I do want to look into the latest libical changes of the GEO property more closely.

dilyanpalauzov: Developers tend to store latitude and longitude as float or double.  When these float/double are converted to text (GEO:…), when no precautions are taken, the values contain many, many, many digits after the decimal point.  More than five digits after the decimal point does not mean, that all digits must be considered meaningful.  For all practical matters only five digits after the decimal point are meaningful here.

rsto: I misunderstood your remark then. I now took note to update this test, but will do so after I did some more testing around geo URI values.

dilyanpalauzov: > I misunderstood your remark then. 

I do not think so, I said different things, which do not exclude each other.

PR #4313, opened 2022-10-27
itip_support.c: prevent partycrashing via iTIP replies

ksmurchison:


rjbs:
This should be configurable.

rsto (approved):

rsto:
Do not merge until it's configurable.

dilyanpalauzov:
What means here partycrashing?

rsto:
> What means here partycrashing?

It means attendees adding themselves to a VEVENT, instead of the organizer adding them.

brong (approved):
This looks good to me - I guess we have to decide if we want to allow party-crashing referrals somehow, but that's a separate thing.

(also - I force pushed the test change just now)

rjbs:
> This should be configurable.

I don't know why I said this.  Do we have any evidence that people want partycrashing?  Obviously we can make this configurable easily enough, but does it matter?  Anyway, would be good to close this out.

PR #4315, opened 2022-10-31
docs: update copyright year

tjwds:


elliefm (approved):
Great, thanks!

This looks obviously fine, but I don't know python well enough to recognise if something is a landmine or not.  So I've built the docs and also `make distcheck`'d with this, and both seem fine and produce sane copyright strings.

PR #4316, opened 2022-11-02
JMAPCalendars: fix calendarevent-get-locations-geo match

elliefm:
Makes the test less sensitive to the number of decimal positions returned by libical, which apparently can vary between versions.

This test was failing on my system, where I haven't updated libical recently.  With this change the test now passes.

**Note to reviewers**: please build and run this test to ensure my change hasn't accidentally broken it in the other direction

### DIFF
--- a/cassandane/tiny-tests/JMAPCalendars/calendarevent-get-locations-geo
+++ b/cassandane/tiny-tests/JMAPCalendars/calendarevent-get-locations-geo
@@ -11,5 +11,6 @@ sub test_calendarevent_get_locations_geo
     my $event = $self->putandget_vevent($id, $ical);
     my @locations = values %{$event->{locations}};
     $self->assert_num_equals(1, scalar @locations);
-    $self->assert_str_equals("geo:37.38601,-122.08290", $locations[0]{coordinates});
+    $self->assert_matches(qr{\Ageo:37\.38601\d*,-122\.08290\d*\Z},
+                          $locations[0]{coordinates});
 }


cassandane/tiny-tests/JMAPCalendars/calendarevent-get-locations-geo
```diff
@@ -11,5 +11,6 @@ sub test_calendarevent_get_locations_geo
     my $event = $self->putandget_vevent($id, $ical);
     my @locations = values %{$event->{locations}};
     $self->assert_num_equals(1, scalar @locations);
-    $self->assert_str_equals("geo:37.38601,-122.08290", $locations[0]{coordinates});
+    $self->assert_matches(qr{\Ageo:37\.38601\d*,-122\.08290\d*\Z},
```
rsto, line 14: All digits, if any, after the expected latitude/longitude values must be zero. So we should not check for `\d` but `0`.

rsto (changes requested):
I'm not particularly happy that we need this just to make tests not fail on oldish libical versions. We really would like to assert that the verbatim latitude/longitude values of the iCalendar GEO property are what gets returned in `coordinates`.
However, I do not have a better idea how to test that properly and still make it work for old libical versions. So let's just make sure that the regex checks for trailing zeros.

cassandane/tiny-tests/JMAPCalendars/calendarevent-get-locations-geo
dilyanpalauzov: As far as I know, libical previously stored internally the coordinates as float.  When these floats were converted to strings, there were many digits, not necessary zero, after the decimal point.  The reason is that saving one number in float and printing that number later to a string, does not reproduce exactly the initially assigned value.

rsto: Right, that particular change is at https://github.com/libical/libical/commit/b0aa3088b8b7aeafba3624c957e8eea3ad72b7a7#diff-1dd2e567bcf8e9ee5fbfc53404fe5907d3bab50954eaafcc0c30cbdbede810fdL1105
But in this test we know that the remaining digits must be zero, if any.

rsto:
@ellie I am going to use this PR as-is but added the `Do Not Merge` until it is approved.

ellie:
> @ellie I am going to use this PR as-is but added the `Do Not Merge` until it is approved.

you meant to @elliefm, not @ellie 😅

rsto:
> > @ellie I am going to use this PR as-is but added the `Do Not Merge` until it is approved.
> 
> you meant to @elliefm, not @ellie 😅

Oops, sorry about that!

cassandane/tiny-tests/JMAPCalendars/calendarevent-get-locations-geo
elliefm: I haven't read up on this in great detail, but from what I have read, my understanding is that there has only ever been 5 significant figures after the decimal anyway, no matter how many digits were actually returned.  In which case, the additional digits being zero-or-otherwise could only be an implementation accident, not any additional precision.  So in my mind the correct thing for the test to do is match the precise digits exactly, and ignore any additional digits.

As long as it remains true that only five of those digits are significant, then implementing the test like this protects us from future underlying implementation changes, and also provides a certain documentation of the reality that we only expect five of these digits to have any meaning.

If at some point we start getting a 6th significant figure, then we'd make a similar change, to exactly match 6 figures and disregard any noise following.

rsto: I'm convinced. Let's merge this as is.

PR #4318, opened 2022-11-02
http_dav: gracefully handle locked dav.db in REPORT

rsto:
Handling a sync REPORT can run into a race where dav.db is locked by another process. This patch makes Cyrus not crash on this error but return a server error. Fixing the conditions that lead to the race is a work in progress.


brong (approved):
Yes, this looks great.  Thanks.

PR #4319, opened 2022-11-02
jmap_api: initialize jmap_changes struct to zero

rsto:
Initialization of the jmap_changes struct is left to the handler function first calling jmap_changes_parse. But in at least one occasion this could get skipped with an early return. Since the compiler did not report this during compilation let's better make sure we zero the structs memory properly.

### DIFF
--- a/imap/jmap_api.h
+++ b/imap/jmap_api.h
@@ -421,6 +421,8 @@ struct jmap_changes {
     json_t *destroyed;
 };
 
+#define JMAP_CHANGES_INITIALIZER {0}
+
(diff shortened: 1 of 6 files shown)


brong (approved):
This looks OK.  There's a common pattern in other code which uses an explicit initialiser like this:

```
struct jmap_changes changes = JMAP_CHANGES_INITIALIZER;
```

Which then allows us to initialise with something other than zeros.

rsto:
> There's a common pattern in other code which uses an explicit initialiser

Yeah, I thought about doing that, then kind of rushed it. But you're right, let's make this properly.

PR #4321, opened 2022-11-04
IMAP Multi-Mailbox Search (RFC 7377)

ksmurchison:
This needs a set of Cassandane tests, but I need to get familiar with IMAPTalk

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -339,7 +339,7 @@ static struct capa_struct base_capabilities[] = {
     /*METADATA-SERVER   RFC 5464.  Sending METADATA implies METADATA-SERVER */
     { "MOVE",                  2 }, /* RFC 6851 */
     { "MULTIAPPEND",           2 }, /* RFC 3502 */
-    /* MULTISEARCH      RFC 7377 is not implemented */
+    { "MULTISEARCH",           2 }, /* RFC 7377 */
     { "NAMESPACE",             2 }, /* RFC 2342 */
     /* NOTIFY           RFC 5465 is not implemented */
     { "OBJECTID",              2 }, /* RFC 8474 */
@@ -410,7 +410,7 @@ static int parse_fetch_args(const char *tag, const char *cmd,
                             struct fetchargs *fa);
 static void cmd_fetch(char *tag, char *sequence, int usinguid);
 static void cmd_store(char *tag, char *sequence, int usinguid);
-static void cmd_search(char *tag, int usinguid);
+static void cmd_search(char *tag, char *cmd);
 static void cmd_sort(char *tag, int usinguid);
 static void cmd_thread(char *tag, int usinguid);
 static void cmd_copy(char *tag, char *sequence, char *name, int usinguid, int ismove);
@@ -1530,6 +1530,13 @@ static void cmdloop(void)
 
                 cmd_enable(tag.s);
             }
+            else if (!strcmp(cmd.s, "Esearch")) {
+                if (c != ' ') goto missingargs;
+
+                cmd_search(tag.s, cmd.s);
+
+                prometheus_increment(CYRUS_IMAP_ESEARCH_TOTAL);
+            }
             else if (!strcmp(cmd.s, "Expunge")) {
                 if (readonly) goto noreadonly;
                 if (!imapd_index && !backend_current) goto nomailbox;
@@ -1976,11 +1983,10 @@ static void cmdloop(void)
             }
             else if (!strcmp(cmd.s, "Search")) {
                 if (!imapd_index && !backend_current) goto nomailbox;
-                usinguid = 0;
                 if (c != ' ') goto missingargs;
             search:
 
-                cmd_search(tag.s, usinguid);
+                cmd_search(tag.s, cmd.s);
 
                 prometheus_increment(CYRUS_IMAP_SEARCH_TOTAL);
             }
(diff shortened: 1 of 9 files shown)


elliefm (commented):
Looking okay so far

ksmurchison:
Tests added

elliefm (approved):
Tests look comprehensive, and I built and ran it and all passed

ksmurchison:
@brong @rsto Do you concur with Ellie that this is good to go?

imap/index.c
```diff
@@ -1913,7 +1913,22 @@ static void begin_esearch_response(struct index_state *state,
...
+
+        if (searchargs->filter) {
+            /* RFC 7377: 2.1
+             * Each ESEARCH response MUST contain the MAILBOX, TAG,
+             * and UIDVALIDITY correlators. */
+            char *extname = mboxname_to_external(index_mboxname(state),
+                                                 searchargs->namespace,
+                                                 searchargs->userid);
+            prot_printf(state->out, " MAILBOX \"%s\" UIDVALIDITY %d",
+
```
rsto, line 1926: nit: empty line

imap/imapd.c
```diff
@@ -6114,21 +6191,121 @@ static void cmd_search(char *tag, int usinguid)
...
+                        mbname_t *mbname = mbname_from_intname(intname);
+                        mrock.root_depth = strarray_size(mbname_boxes(mbname));
+                        mbname_free(&mbname);
+                    }
+
+                    mboxlist_mboxtree(intname, multisearch_cb, &mrock, flags);
+                }
+                break;
+            }
+            }
```
rsto, line 6287: nit: indenting

imap/imapparse.c
```diff
@@ -575,6 +575,131 @@ static int get_search_date(struct protstream *pin, time_t *start, time_t *end)
...
+    }
+
+    do {
+        c = getastring(pin, pout, &extname);
+        if (!buf_len(&extname)) goto bad;
+
+        intname = mboxname_from_external(buf_cstring(&extname),
+                                         searchargs->namespace,
+                                         searchargs->userid);
+        strarray_appendm(mboxes, intname);
```
rsto, line 603: this could crash later if intname is NULL (strarray_appendm does not check for NULL string)

imap/imapd.c
```diff
@@ -6051,27 +6057,98 @@ static void cmd_store(char *tag, char *sequence, int usinguid)
...
+        /* Mailbox is already selected - use current index state */
+        state = imapd_index;
+    }
+    else if ((r = index_open(mbentry->name, &mrock->init, &state))) {
+        return r;
+    }
+
+    /* Use a fresh search_expr_t for each mailbox */
+    if (mrock->args->root) search_expr_free(mrock->args->root);
+    mrock->args->root = search_expr_duplicate(mrock->expr);
```
rsto, line 6110: This leaks
```
[510503] =====> Instance::_check_valgrind_logs[1317] ==510534==    at 0x4849E4C: malloc (vg_replace_malloc.c:307)
[510503] =====> Instance::_check_valgrind_logs[1317] ==510534==    by 0x4BE017F: xmalloc (xmalloc.c:55)
[510503] =====> Instance::_check_valgrind_logs[1317] ==510534==    by 0x4BE01E3: xzmalloc (xmalloc.c:64)
[510503] =====> Instance::_check_valgrind_logs[1317] ==510534==    by 0x4961073: search_expr_new_nnodes (search_expr.c:124)
[510503] =====> Instance::_check_valgrind_logs[1317] ==510534==    by 0x496110F: search_expr_duplicate_nnodes (search_expr.c:241)
[510503] =====> Instance::_check_valgrind_logs[1317] ==510534==    by 0x496466B: search_expr_duplicate (search_expr.c:261)
[510503] =====> Instance::_check_valgrind_logs[1317] ==510534==    by 0x12979B: cmd_search (imapd.c:6200)
[510503] =====> Instance::_check_valgrind_logs[1317] ==510534==    by 0x139313: cmdloop (imapd.c:1536)
[510503] =====> Instance::_check_valgrind_logs[1317] ==510534==    by 0x13C1CF: service_main (imapd.c:1008)
[510503] =====> Instance::_check_valgrind_logs[1317] ==510534==    by 0x151C37: main (service.c:647)
```

rsto (changes requested):
I just looked at RFC 7377 for the first time, but the tests and implementation seem to cover it well.
 
Apart from a memory leak I just found some nits.

imap/index.c
ksmurchison: Hmm.  Must've been a random keystroke.  Fixed

imap/imapparse.c
ksmurchison: Fixed.

imap/imapd.c
ksmurchison: Leak fixed

imap/imapd.c
ksmurchison: @rsto This is an artifact of declaring variables inside a switch clause.  Its looks weird, but this is how Emacs does the indenting.  How would you like this to look>

brong (approved):
This looks good, thanks Ken.

I had some questions about cmd_cancelled but I think it's already called in the inner loop so that's OK.

rsto (approved):
LGTM

imap/imapd.c
elliefm: I don't like the two `}`s at the same indentation level either, but I thought about this myself, and concluded that it's an artefact of our style placing the `case` labels at the same indent level as the `switch` itself.

If our `switch` style was like:

```
switch(foo) {
    case 'a':
        // something simple;
        break;
    case 'b': {
        // something complex
        break;
    }
}
```

-- that is, with the cases at their own indent level -- then this wouldn't be a problem.  But that would use up a lot of horizontal space on whitespace, leaving less for code, meaning more lines need to be wrapped, leading to more complicated wrapping solutions...

I suppose the variable(s) in question could be moved to the containing scope, removing the need for `{}` in the `case`, but as it is their lifetime is clear, even if the `}}` is jarring.  If they were up a level, one would then need to read that whole scope holding them in mind, even though they're only used briefly in that one place.

So while I don't really like it, I didn't have a clearly better suggestion, and still don't.

PR #4322, opened 2022-11-04
mailbox.c: free header ACL before re-reading it

ksmurchison:


elliefm (approved):
Oh, good find, looks good

PR #4324, opened 2022-11-08
Ignore privacy property default values

rsto:
The current implementation of JSCalendar.privacy always writes the privacy level to the custom X-JMAP-PRIVACY property. This isn't necessary and actually broke comparison of normalized iCalendar components. This patch fixes that and also adds a custom normalization method to normalize Cyrus extension property values.


ksmurchison (approved):
LGTM

PR #4331, opened 2022-11-16
http_carddav: capitalize default address name to "Personal"

rsto:
Recent changes in iOS expose this value to users and all the other addressbooks are capitalized.


ksmurchison:
This LGTM but what do we do with existing addressbooks, if anything?

rsto:
We do nothing for existing addressbooks for now.

ksmurchison (approved):

PR #4333, opened 2022-11-19
IMAP4rev2

ksmurchison:
This probably needs another set of eyes on RFC 9501 to make sure that everything that is required is covered in this PR.

Its probably best to review the code commit-by-commit, in order.  Note that #4330 is also part of IMAP4rev2.

### DIFF
--- a/cassandane/Cassandane/Cyrus/IMAP4rev2.pm
+++ b/cassandane/Cassandane/Cyrus/IMAP4rev2.pm
@@ -1,0 +1,286 @@
+#!/usr/bin/perl
+#
+#  Copyright (c) 2011-2022 Fastmail Pty Ltd. All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions
+#  are met:
+#
+#  1. Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#
+#  2. Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in
+#     the documentation and/or other materials provided with the
+#     distribution.
+#
+#  3. The name "Fastmail Pty Ltd" must not be used to
+#     endorse or promote products derived from this software without
+#     prior written permission. For permission or any legal
+#     details, please contact
+#      FastMail Pty Ltd
+#      PO Box 234
+#      Collins St West 8007
+#      Victoria
+#      Australia
+#
+#  4. Redistributions of any form whatsoever must retain the following
+#     acknowledgment:
+#     "This product includes software developed by Fastmail Pty. Ltd."
+#
+#  FASTMAIL PTY LTD DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE,
+#  INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY  AND FITNESS, IN NO
+#  EVENT SHALL OPERA SOFTWARE AUSTRALIA BE LIABLE FOR ANY SPECIAL, INDIRECT
+#  OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF
+#  USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
+#  TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE
+#  OF THIS SOFTWARE.
+#
+
+package Cassandane::Cyrus::IMAP4rev2;
+use strict;
+use warnings;
+use Cwd qw(abs_path);
+use File::Path qw(mkpath);
+use DateTime;
+use Data::Dumper;
+
+use lib '.';
+use base qw(Cassandane::Cyrus::TestCase);
+use Cassandane::Util::Log;
+use Cassandane::Util::NetString;
+
+
+sub new
+{
+    my $class = shift;
+    return $class->SUPER::new({ adminstore => 1, services => ['imap'] }, @_);
+}
+
+sub set_up
+{
+    my ($self) = @_;
+    $self->SUPER::set_up();
+}
+
+sub tear_down
+{
+    my ($self) = @_;
+    $self->SUPER::tear_down();
+}
+
+sub test_basic
+    :NoAltNameSpace :min_version_3_7
+{
+    my ($self) = @_;
+
+    xlog $self, "Make some messages";
+    my $uid = 1;
+    my %msgs;
+    for (1..20)
+    {
+        $msgs{$uid} = $self->make_message("Message $uid");
+        $msgs{$uid}->set_attribute('uid', $uid);
+        $uid++;
+    }
+
+    my $talk = $self->{store}->get_client();
+    $talk->unselect();
+
+    xlog $self, "Create mailbox with mUTF7 encoded name";
+    my $res = $talk->_imap_cmd('CREATE', 0, "", "INBOX.&JgA-");
+    $self->assert_str_equals('ok', $talk->get_last_completion_response());
+
+    xlog $self, "ENABLE IMAP4rev2";
+    $res = $talk->_imap_cmd('ENABLE', 0, "enabled", "IMAP4rev2");
+    $self->assert_num_equals(1, $res->{imap4rev2});
+
+    xlog $self, "Verify that LIST responses use UTF8 mailbox names";
+    $res = $talk->list("", "*");
+    $self->assert_mailbox_structure($res, '.', {
+        'INBOX'    => [qw( \\HasChildren )],
+        "INBOX.☀" => [qw( \\HasNoChildren )],
+    });
+
+    xlog $self, "EXAMINE mailbox with UTF8 mailbox name";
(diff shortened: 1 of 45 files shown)


imap/mboxname.c
```diff
@@ -577,8 +577,33 @@ EXPORTED mbname_t *mbname_from_intname(const char *intname)
     return mbname;
 }
 
-EXPORTED mbname_t *mbname_from_extname(const char *extname, const struct namespace *ns, const char *userid)
+EXPORTED mbname_t *mbname_from_extname(const char *extname,
+                                       const struct namespace *ns,
+                                       const char *userid)
 {
+    char *freeme = NULL;
```
elliefm, line 584: There's some early returns in the unmodified portion of this function, which I think may result in `freeme` being leaked.  Github won't let me comment on them directly though, because the patch doesn't change them... they start around `/* specialuse magic */`

imap/mbpath.c
```diff
@@ -448,6 +448,8 @@ int main(int argc, char **argv)
         fatal(error_message(r), EX_SOFTWARE);
     }
 
+    mbpath_namespace.isutf8 = opts.utf8;
+
```
elliefm, line 452: Could `mboxname_init_namespace()` do this?

elliefm (commented):
This is a partial review -- I've looked at the code, but not the RFC yet.

Mostly looking good so far, though there's possibly a memory leak in one of the combined mboxname external functions.

What are you thinking for testing this?  I guess IMAPTalk probably doesn't know IMAP4rev2 yet, so it might need a similar approach to the ESEARCH stuff, with direct `_imap_cmd()` calls and custom response handlers.

There's also mboxname.testc, which has lots of tests for `mboxname_to_external()`, but none of them exercise the utf-8 behaviour because they don't know it exists.  It'd be nice to have some basic tests for it.

elliefm:
`QResync.qresync_simple` fails on this branch for me... looks like IMAPTalk choking on something unexpected in a command response...

Here's a little bit of the verbose output:
```
[30803] =====> Cyrus::QResync::test_qresync_simple[90] (test_qresync_simple) Mark some messages \Deleted
C: 153 enable qresync
S: 
Perl exception: IMAPTalk: Expected simple atom, got: 
 at /usr/share/perl5/Mail/IMAPTalk.pm line 4311.
```

That empty "S:" line is odd... and then the "expected simple atom, got: [nothing]". I wonder if there's a bug in the `cmd_enable()` output handling?

imap/mbpath.c
ksmurchison: @elliefm  I thought about that.  I think I prefer making the isadmin argument a bitmask of flags rather than adding a third argument.  Your thoughts?

ksmurchison:
@elliefm The QResync test passes on both my desktop and laptop

dilyanpalauzov:
Please update the imap/rfc-support.html file with RFC 9501 as Implemented and create a file under changes/next/ .

For the implemented RFC 5182 (IMAP Saved Search results) there is no changes/next file.

imap/mboxname.c
ksmurchison: Yeah, thanks for catching that.  Should be fixed now.

imap/mbpath.c
elliefm: No strong opinion either way, and I'll accept whatever you choose, but some off the cuff thoughts:

* adding a bitmask is a (small) bunch of extra infrastructure, which is more work than just adding one argument
* but bitmask fields, which have names, are easier to read than a list of ones and zeroes (though we usually comment those, like `foo(..., /*isadmin*/ 0, ...)`, which solves that issue)
* the `mboxname_init_namespace()` function signature has only changed once in as long as it's existed (`isadmin` was originally called `force_std` -- which now that I see that, explains all the "force standard" comments everywhere, hah). I'm not sure we expect it to change much in the future after whatever we do here
* the infrastructure cost of a bitmask makes future changes easier, but if we don't expect any...

Myself, I would probably just add the extra argument, and not worry about  converting to a bitmask unless we started needing a third flag someday.  But that's just me.

ksmurchison:
@elliefm Cass test added.  I think it exercises other than SEARCHRES, which has its own test in Search.pm

elliefm:
Thanks, I'll have a look.  I'm also reading through RFC 9051, which is slow going because I'm trying to read it carefully

elliefm:
I don't see anything implementing OLDNAME:

> [6.3.9.7](https://datatracker.ietf.org/doc/html/rfc9051#section-6.3.9.7).  OLDNAME Extended Data Item
> 
>    The OLDNAME extended data item is included when a mailbox name is
>    created (with the CREATE command), renamed (with the RENAME command),
>    or deleted (with the DELETE command).  (When a mailbox is deleted,
>    the "\NonExistent" attribute is also included.)  IMAP extensions can
>    specify other conditions when the OLDNAME extended data item should
>    be included.
> 
>    If the server allows denormalized mailbox names (see [Section 5.1](https://datatracker.ietf.org/doc/html/rfc9051#section-5.1)) in
>    SELECT/EXAMINE, CREATE, RENAME, or DELETE, it SHOULD return an
>    unsolicited LIST response that includes the OLDNAME extended data
>    item, whenever the supplied mailbox name differs from the resulting
>    normalized mailbox name.  From the client point of view, this is
>    indistinguishable from another user renaming or deleting the mailbox,
>    as specified in the previous paragraph.
> 
>    A deleted mailbox can be announced as follows:
> 
>      S: * LIST (\NonExistent) "." "INBOX.DeletedMailbox"
> 
>    Example of a renamed mailbox:
> 
>      S: * LIST () "/" "NewMailbox" ("OLDNAME" ("OldMailbox"))

ksmurchison:
Hmm.  I swear that I thought I saw existing code for OLDNAME, but now  I can't find it.  I'll created a sub-task for it and work on it.

imap/mbpath.c
ksmurchison: @elliefm Implemented bitmask

cassandane/Cassandane/Cyrus/IMAP4rev2.pm
```diff
@@ -161,4 +161,126 @@ sub test_imap4rev2_simple
...
+    my %handlers =
+    (
+        list => sub
+        {
+            my (undef, $list) = @_;
+            push(@results, $list);
+        },
+    );
+
+    xlog $self, "Create a mailhox with denormalized mailbox name";
```
elliefm, line 196: Recurring "mailhox" typo in this test 😅

ksmurchison: Fixed

ksmurchison:
@elliefm @brong I've merged bugfixes and cleaned up the commit history.  Other than the mboxname_init)namespace() commit, the rest are relatively small and should be easy to understand.

ksmurchison:
@elliefm @brong I think this is ready for a final review.  This PR is currently rebased on imap_quota which gets us STATUS DELETED

imap/imapd.c
```diff
@@ -14859,12 +14862,16 @@ static void cmd_compress(char *tag, char *alg)
...
 {
     static struct buf arg;
-    int c;
+    int c = EOF;
     unsigned new_capa = 0;
 
-    /* RFC 5161 says that enable while selected is actually bogus,
-     * but it's no skin off our nose to support it, so don't
-     * bother checking */
+    if (imapd_index || backend_current) {
```
brong, line 15009: Was this a hard requirement for IMAP4rev2?  We probably allowed this because some dodgy client was doing the wrong thing.

I think this is the one thing I'm concerned about maybe breaking somebody here.

brong (approved):

imap/imap_proxy.c
```diff
@@ -473,16 +473,35 @@ void print_listresponse(unsigned cmd, const char *extname, char hier_sep,
...
-         *     ; the extended LIST command.
-         */
-        if (attributes & MBOX_ATTRIBUTE_CHILDINFO_SUBSCRIBED)
-            prot_puts(imapd_out, "\"SUBSCRIBED\"");
-        prot_puts(imapd_out, "))");
+    if (oldname || (attributes & MBOX_ATTRIBUTE_CHILDINFO_MASK)) {
+        sep = "";
+        prot_puts(imapd_out, " (");
+
+        if (oldname) {
```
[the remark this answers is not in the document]
ksmurchison: @brong OLDNAME is also required for NOTIFY (RFC 5465) which is why I left it ungated.  I don't believe that we would ever set oldname when neither IMAP4rev2 nor NOTIFY are active, but I can add a check if we feels it could be a problem.

imap/imapd.c
ksmurchison: @brong I don't recall if allowing this was an issue with IMAP4rev2.  I'll do some testing and allow it if its not problematic

imap/imapd.c
```diff
@@ -906,6 +906,7 @@ static void imapd_reset(void)
     }
     imapd_userisadmin = 0;
     imapd_userisproxyadmin = 0;
+    imapd_namespace.isutf8 = 0;
```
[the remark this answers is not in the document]
ksmurchison: I removed this because its unneccessary as we do mboxname_init_namespace() in service_main()

imap/imapd.c
ksmurchison: Removed this check and reverted to allowing ENABLE while SELECTEd

PR #4335, opened 2022-11-23
CalendarEvent/participantReply

ksmurchison:


cassandane/tiny-tests/JMAPCalendars/calendarevent-participantreply-simple
```diff
@@ -0,0 +1,128 @@
...
+    $data = $self->{instance}->getnotify();
+    ($imip) = grep { $_->{METHOD} eq 'imip' } @$data;
+    $self->assert_not_null($imip);
+
+    $payload = decode_json($imip->{MESSAGE});
+    $ical = $payload->{ical};
+
+    $self->assert_str_equals("cassandane\@example.com", $payload->{recipient});
+    $self->assert($ical =~ "METHOD:REPLY");
+return;
```
wolfsage, line 108: Early return here - so some of your tests aren't running

imap/jmap_calendar.c
```diff
@@ -7707,6 +7721,274 @@ static int jmap_calendarevent_parse(jmap_req_t *req)
...
+
+    if (!part_id) {
+        r = HTTP_NOT_FOUND;
+        goto done;
+    }
+
+    /* Create patch */
+    struct buf buf = BUF_INITIALIZER;
+    if (update.eid->ical_recurid) {
+        /* XXX  FIX ME */
```
wolfsage, line 7908: Fix what?

imap/jmap_calendar.c
```diff
@@ -5345,6 +5352,31 @@ static void remove_itip_messages(struct caldav_db *db,
     }
 }
 
+static int is_standalone(struct jmap_caleventid *eid,
```
rsto, line 5355: This function name is likely to misunderstand, as it returns a true-ish value on error. Would `check_eventid_exists` be more clear?

imap/jmap_calendar.c
```diff
@@ -7707,6 +7721,276 @@ static int jmap_calendarevent_parse(jmap_req_t *req)
...
+    struct buf buf = BUF_INITIALIZER;
+    if (update.eid->ical_recurid) {
+        /* XXX  FIX ME */
+        struct icaltimetype tt = icaltime_from_string(update.eid->ical_recurid);
+        struct jmapical_datetime dt;
+
+        jmapical_datetime_from_icaltime(tt, &dt);
+        jmapical_localdatetime_as_string(&dt, &buf);
+
+        buf_insertcstr(&buf, 0, "recurrenceOverrides/");
```
rsto, line 7915: This is wrong for stand-alone events. A stand-alone event does not have recurrenceOverrides, rather its top-level participants property should be patched.

imap/jmap_calendar.c
```diff
@@ -7707,6 +7721,276 @@ static int jmap_calendarevent_parse(jmap_req_t *req)
...
+        buf_putc(&buf, '/');
```
rsto, line 7916: If "recurrenceOverrides" is set, then the DateTime-formatted recurrence id also must be set here.

rsto (commented):
The code looks mostly OK but it seems to lack proper support for standalone event instances, so this needs more work.
The whole idea of this new JMAP method looks backward to me (why not just generate the iTIP outside Cyrus), and its empty response body also isn't what I've seen from JMAP specs so far. But this is not the place to further discuss the design.

cassandane/tiny-tests/JMAPCalendars/calendarevent-participantreply-simple
ksmurchison: Fixed

imap/jmap_calendar.c
ksmurchison: Fixed (partially)

imap/jmap_calendar.c
ksmurchison: It is (because my test works).  The recurrence-id is written to the buffer first (because jmapical_localdatetime_as_string() resets the buffer), and then we prepend the string.

imap/jmap_calendar.c
ksmurchison: Hopefully fixed

rsto (approved):
LGTM - as discussed an additional test for standalone instances would be great

brong:
I've rebased this on master and force-pushed.  There was a conflict on the json_pack for imip_send (both branches added a field)

rjbs:
We think this can merge, conceptually.

PR #4337, opened 2022-11-29
Add sync_support for delayed delete mode (3.6)

kanarip:
In `sync_apply_unmailbox()` and `sync_apply_unuser()`, the call to `mboxlist_deletemailbox()` is currently independent of whether `delete_mode: delayed` is configured, creating a risk of data-loss.

Under certain circumstances, in an `A<->B` two-way replication scenario for high availability, where B may have failed to synchronize a mailbox `user/john` to `A`, and `A` has a `user/john` mailbox created, the initial `A->B` "mailbox" replication is refused, but `A` will promote the synchronization to become of the "user" level, which `B` appears to accept; it'd remove the existing `user/john` mailbox, including the file-system hierarchy, resulting in (effectively) data-loss, before recreating the `user/john` mailbox.

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -3630,6 +3630,7 @@ bail:
 int sync_apply_unmailbox(struct dlist *kin, struct sync_state *sstate)
 {
     const char *mboxname = kin->sval;
+    int r;
 
     struct mboxlock *namespacelock = mboxname_usernamespacelock(mboxname);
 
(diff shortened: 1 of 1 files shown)


elliefm:
Can you rebase this PR to be against master please?  It should not land directly into a release branch unless it fixes a problem that doesn't exist on master, or is a backport of an existing fix from master.  I don't believe either is the case here -- but I could be wrong, in which case please say so.  Thanks!

kanarip:
Thanks @elliefm, see PR #4340 for the edition to master -- it's a clean cherry-pick.

elliefm:
Thanks.  I'll mark the other one for backporting, and I'm closing this one to reduce confusion or double-handling.

Once the other one lands on master, I'll cherry-pick it to the 3.6 branch directly.

PR #4340, opened 2022-12-01
Add sync_support for delayed delete mode (master)

kanarip:
In sync_apply_unmailbox() and sync_apply_unuser(), the call to mboxlist_deletemailbox() is currently independent of whether delete_mode: delayed is configured, creating a risk of data-loss.

Under certain circumstances, in an A<->B two-way replication scenario for high availability, where B may have failed to synchronize a mailbox user/john to A, and A has a user/john mailbox created, the initial A->B "mailbox" replication is refused, but A will promote the synchronization to become of the "user" level, which B appears to accept; it'd remove the existing user/john mailbox, including the file-system hierarchy, resulting in (effectively) data-loss, before recreating the user/john mailbox.

From PR #4337, with @elliefm requesting the PR is rebased against master.

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -3811,6 +3811,7 @@ bail:
 int sync_apply_unmailbox(struct dlist *kin, struct sync_state *sstate)
 {
     const char *mboxname = kin->sval;
+    int r;
 
     struct mboxlock *namespacelock = mboxname_usernamespacelock(mboxname);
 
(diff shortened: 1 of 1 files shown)


kanarip:
I believe I could use some help;

In [perl/imap/Cyrus/AccountSync.pm#L240](https://github.com/cyrusimap/cyrus-imapd/blob/master/perl/imap/Cyrus/AccountSync.pm#L240), we would appear to call `Cyrus::Mbname->new_intname("DELETED.user.newuser.63885DC8")`, which appears to fail to set `$self->{is_deleted}`? But [perl/imap/Cyrus/Mbname.pm#L29](https://github.com/cyrusimap/cyrus-imapd/blob/master/perl/imap/Cyrus/Mbname.pm#L29) appears to me should -- what am I missing?

elliefm:
@kanarip Your [most recent comment](https://github.com/cyrusimap/cyrus-imapd/pull/4340#issuecomment-1333701316) makes me think this is incomplete, but I don't know the answer to your question.  The changes apply cleanly, but they don't pass our tests.  Do you expect to keep working on this?  If not I'll close it.

PR #4342, opened 2022-12-05
CaldavAlarm: fix inf loop in test_disable_high_freq

elliefm:
```
# create hourly occurring events with a set of byminute
```

There was a bug in the counter for this loop that caused the test to loop forever if it happened to start at a time whose minutes value was 52, 53, or 54.  This was because the loop condition would end up checking for `$bymin <` 60, 61, or 62, which it always is because of the `% 60` in the loop body.

There was also a bug in how the test results were examined, which was being masked by the test never reaching the assertion that would expose it.

There's 3 minutes each hour when the bug can be reproduced on master, or the fix can be confirmed on this branch.  I've done both myself; you may wish to do so too.

### DIFF
--- a/cassandane/Cassandane/Cyrus/CaldavAlarm.pm
+++ b/cassandane/Cassandane/Cyrus/CaldavAlarm.pm
@@ -3113,9 +3113,9 @@ EOF
     #
     # byminute=startmin and +2 should succeed since there is a 120s interval
     #
-    my $bymin_ok = $startmin + 2;
-    for (my $bymin = $startmin + 1; $bymin < $startmin + 3; $bymin += 1) {
-        $bymin = $bymin % 60;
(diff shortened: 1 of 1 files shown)


elliefm:
Okay, I think I've found a better fix that entirely avoids needing to think about the modulus in the loop management. I'll squash this back to a single correct commit once it's definitely good, but keep the noise and the draft tag until I've confirmed that.

elliefm:
The infinite loop is definitely fixed, and the loop boundaries are all handled correctly now, but there's still something wrong in the body of the test, with the event data not matching.  I think it's to do with `$bymin` being modulo 60 but other things not.  The day is at an end though, will look at it again with fresh eyes next time.

ksmurchison (approved):

PR #4346, opened 2022-12-07
imapd.c: advertise URL-PARTIAL since its already supported (per tests)

ksmurchison:


dilyanpalauzov:
Please move at https://www.cyrusimap.org/dev/imap/rfc-support.html RFC5550 from Wishlist to Implemented, possibly removing RFC 4550 entirely, and create a changes/next file.

ksmurchison:
RFC 4550 includes more than just this one capability, some of which we don't yet support.

dilyanpalauzov:
https://www.iana.org/assignments/imap-capabilities/imap-capabilities.xhtml does not list RFC 4550.  It lists RFC 5550 behind one capability.  RFC 5550 obsoletes RFC 4550.

ksmurchison:
Sorry, 4550 was a typo.  Same story with 5550, more than just URL-PARTIAL

dilyanpalauzov:
Please write in the rfc-supported file what is implemented from RFC 5550 and what is missing.  Like the comment behind RFC 5267 or RFC6352 at https://www.cyrusimap.org/3.4/imap/rfc-support.html .

The comment after RFC 6352 is still at https://www.cyrusimap.org/dev/imap/rfc-support.html .  Wasn’t the server-side conversion between vCard 3.0<->4.0 recently implemented?

rsto (approved):

PR #4347, opened 2022-12-09
Cyrus 3.8: http_caldav: allow CalDAV alarms on calendar home

rsto:
The base for this PR should be the `cyrus-imapd-3.8` branch, but we don't have that, yet. **Do not merge** this in master.

This patch allows setting CalDAV alarm DAV properties on the calendar home. Version 3.7 erroneously broke this and a complete rewrite of default alarm support in Cyrus is in the works.
In order not to include the current broken state in the upcoming 3.8 release, this patch falls back to treating CalDAV alarm DAV properties like any other "dead" DAV property. This is consistent with version 3.6 and earlier.


ksmurchison (approved):
LGTM

elliefm:
This has now been merged onto the 3.8 branch

PR #4348, opened 2022-12-12
github actions: split cass run into several steps

elliefm:
Splits the cassandane part of the github actions workflow into separate steps:

1. build and check all the cassandane components
2. run cassandane quietly (-f prettier)
3. only if the first cass run failed: rerun with --rerun and -f pretty to get noisy output for only the failing tests

Each named step is folded separately in the web interface, so splitting it like this makes it easier to jump to the detail you care about, without having to scroll through all the output for the entire cassandane run.

Re-running the failed tests and only failing the workflow if there's still failures might also reduce false negatives from transient test failures, which are almost never related to the PR being evaluated anyway.

I have no idea whether this will work or not, but I think I need to create a pull request to even find out, so here goes...

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,35 +22,40 @@ jobs:
       with:
         fetch-depth: 0
     - name: configure and build
+      shell: bash
       run: |
         git config --global --add safe.directory /__w/cyrus-imapd/cyrus-imapd
         echo "building cyrus version" $(./tools/git-version.sh)
         ./tools/build-with-cyruslibs.sh
-      shell: bash
     - name: update jmap test suite
+      working-directory: /srv/JMAP-TestSuite.git
+      shell: bash
       run: |
-        cd /srv/JMAP-TestSuite.git
         git fetch
         git checkout origin/master
         git clean -f -x -d
         cpanm --installdeps .
-        cd -
-      shell: bash
     - name: install missing deps
+      shell: bash
       run: |
         cpanm IO::File::fcntl
         cpanm Digest::CRC
+    - name: set up cassandane
+      working-directory: cassandane
       shell: bash
-    - name: run cassandane
       run: |
-        cd cassandane
         cp -af cassandane.ini.dockertests cassandane.ini
         chown cyrus:mail cassandane.ini
         make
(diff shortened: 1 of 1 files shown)


elliefm:
Well, it seems to work, in that the quiet step was successful, and so the noisy one didn't run at all.

Remains to be seen whether the noisy step will correctly fire when the quiet one fails.  I'm concerned that the `continue-on-error: true` in the quiet step might stop the noisy step's `failure()` check from firing, the documentation isn't particularly clear on how these things interact.

Hmm... I guess I could push a temporary commit that deliberately provokes a test failure, and see what happens.

elliefm:
> I'm concerned that the continue-on-error: true in the quiet step might stop the noisy step's failure() check from firing

This seems to have been the case.  Tinkering with it to see if I can find a way to make it work

wolfsage:
Seems like this is still a work in progress? I'm going to wait to review it until you say it's ready!

One comment though that worries me: This might making tracking heisenbugs down hard if the first run fails but the rerun of failed tests succeeds (unless we capture all the output from both runs...)

elliefm:
@wolfsage Okay I've finally seen all possible test/fail combinations work correctly, so it's now ready for review.  It's not actually any different from last time you commented, except that I've dropped the temporary bad test, and squashed the mess out of the history

elliefm:
I might want to extend this to have one more stage at the end that runs with `--slow-only` (once #4485 lands), only if every previous stage passed.  Then we'd get regular CI runs of our slow tests too.

This PR conflicts with #4481, so whichever of these lands first, the other will need to be rebased and then updated accordingly.

This is stale since I forgot about it over EOY, and I haven't yet brought it up to date.  @ksmurchison @rsto can you both please have a look at this and let me know what you think about the general idea -- in particular read the existing discussion too, since @wolfsage had some concerns.  If we like this approach, I'll spend the time getting it up to date and ready to move forward.  If we don't like it, then let's figure out what we do want and then do that.

rsto:
TBH, I haven't cared much about the Github test runner so far, both because of it's out-of-band feedback and because I almost always get error reports for my branches where I can't reproduce the error locally. Instead, always run the full Cassandane test suites on the include-in-fastmail amalgams at least once, often more a week.

That being said, I see how can test runner could improve also my Cyrus development workflows, and any change that makes error reporting for the test runner more sane looks like a plus to me. I don't know if a third step for the slow tests is actually needed, though, why not just run them all the same time?

ksmurchison:
I like the general idea of 2 passes, but I'm not sure we need a third.  As rsto said, why not just always run the slow tests?  Its not like anyone is actively waiting on the results.

elliefm:
Now I'm actually curious how much wallclock time including the slow tests adds to a full run in practice.  Perhaps not much, since we run tests in parallel anyway.  I guess it depends on whether they're slow because they're doing a lot of work, or slow because they spend a lot of time waiting around.  If they're mostly just waiting around, their contribution will be negligible.  I'll do a couple of runs (on the branch from #4485) and find out...

Though I expect the difference will be inconsequential, in which case I'll just enable `--slow` on the existing two stages.

elliefm:
Hah, unscientific statistically insignificant test:

* full run without `--slow`: 1457 wallclock seconds
* full run with `--slow`: 1360 wallclock seconds

Apparently it runs faster with slow tests enabled... so I guess over a full run, the difference between with- and without-slow tests is within the usual variance, and is basically just noise.  So I'll have CI always run the slow tests too, so we don't have to.

elliefm:
This needs to be rebased again after #4501 lands...

elliefm:
@rsto @ksmurchison This has been rebased and updated after all the other recent github actions tinkering, and is now ready for a proper review.  Thanks :)

elliefm:
We don't have github actions on 3.2 or 3.4 (yet?), so can't backport this to those versions (until we do).

PR #4352, opened 2022-12-14
GitHub CI status badge

### LINKED ISSUE #4329: cyrusimap.org still has travis-ci build status links
The website at [cyrusimap.org](https://www.cyrusimap.org) still has build status thingies (bottom left corner) that go to Travis CI, but we've switched to using GitHub Actions for CI.

We should find where these thingies come from, and either update or remove them.

ire4ever1190:
Closes #4329 

Switches the status badge in the readme and doc page from travis to github actions

Currently the badge shows `no status` for 3.4 branch since it seems the actions have never been ran on that branch. Could someone start a manual run for that branch so the status is shown correctly?


elliefm (changes requested):
Thanks, that's really helpful!

Though, the stable version is now 3.6, not 3.4.  I think you submitted this PR around the same time as I was doing the 3.6.0 release, so now we've got a little synchronisation mess to sort out.

Do you mind updating this so that the stable badge references 3.6 instead of 3.4?  It looks like only README.md needs to be changed.

elliefm:
Note to self: when backporting this to the other branches, make sure to update:

```
html_theme_options = { 'github_version': 'master'}
```

to reference the actual branch names, not "master".  Looks like we've been forgetting this, and at least some of the release branches incorrectly call themselves "master".

ire4ever1190:
Updated readme to point to 3.6

elliefm (approved):
Thanks!

elliefm:
The fact that I've pushed to all these branches while backporting this for each should mean that each will also be provoked into running a build, so the "no status" ones should get a real status soon too.

PR #4355, opened 2022-12-15
http_ws.c: proc_register() a WebSocket connection with cmd="WS"

ksmurchison:
For differentiating a simple GET request from a long-running WS connection.  Helps with tooling that might need to terminate long-running connections for a user.


elliefm (approved):
Looks good

PR #4356, opened 2022-12-15
sieve-lex.l: reject non-ASCII chars outside of strings and comments

ksmurchison:
Per RFC 5228:
   With the exceptions of strings and comments, the language is limited
   to US-ASCII characters.


elliefm (approved):
My lex is rusty, but this looks like it does what it says on the tin.  Do we want a cass test to exercise it (and for regression testing)?

ksmurchison:
Cass test added

PR #4358, opened 2022-12-23
Implement sync-collection REPORT on home-set collections

ksmurchison:
Tested by hand but needs Cass tests

### DIFF
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -610,7 +610,13 @@ static struct meth_params caldav_params = {
       { NS_CALDAV, "calendar-data", &caldav_import } },
     { CALDAV_SUPP_DATA, &caldav_put },
     { 0, caldav_props },                        /* Allow infinite depth */
-    caldav_reports
+    caldav_reports,
+    /* Offsets of CalDAV modseqs -- used for report_sync_col() */
+    { offsetof(struct mboxname_counters, caldavmodseq),
+      offsetof(struct mboxname_counters, caldavfoldersmodseq),
+      offsetof(struct mboxname_counters, caldavdeletedmodseq),
+      offsetof(struct mboxname_counters, caldavfoldersdeletedmodseq)
+    }
 };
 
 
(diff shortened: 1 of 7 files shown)


imap/http_caldav.c
```diff
@@ -610,7 +610,12 @@ static struct meth_params caldav_params = {
       { NS_CALDAV, "calendar-data", &caldav_import } },
     { CALDAV_SUPP_DATA, &caldav_put },
     { 0, caldav_props },                        /* Allow infinite depth */
-    caldav_reports
+    caldav_reports,
+    { offsetof(struct mboxname_counters, caldavmodseq),
```
brong, line 614: This is really clever and I hope portable!  It probably needs a comment somewhere explaining how and why it's used,.

imap/http_dav.c
```diff
@@ -7524,6 +7531,110 @@ static int updates_cb(void *rock, void *data)
...
+        len++;
+    }
+
+    mbname_t *mbname = mbname_from_intname(mbentry->name);
+    snprintf(p, MAX_MAILBOX_PATH - len, "%s/",
+             strarray_nth(mbname_boxes(mbname), -1));
+    mbname_free(&mbname);
+
+    /* Skip special calendar collections */
+    if (fctx->req_tgt->namespace->id == URL_NS_CALENDAR &&
```
brong, line 7567: It would be nice to have a flag on these rather than exact name matches, but we have what we have for now!

brong (approved):
This looks fine, it could do with some more comments the modseq pair stuff, but I can understand it all.

And yeah, tests would be nice - particularly tests that it doesn't leak collections that it shouldn't (including ACL checks)

imap/http_caldav.c
ksmurchison: libical uses stuff like this, so I assume its portable

imap/http_dav.c
ksmurchison: Agreed.  We have flags when parsing a method URL, but not when we're iterating over mailboxes.

ksmurchison:
@brong Added a Cass test

brong (approved):
Thanks for the test!  Looks good

PR #4359, opened 2022-12-29
Lock ordering fixes

brong:
Hey Ken,

Here's, I THINK, all the lock ordering fixes that we need.  You can check the logic I used to test them with the top commit on https://github.com/cyrusimap/cyrus-imapd/pull/3179 which I just rebased on top of this, and which does all the LOCKERROR syslogs for cases where we try to lock any index files while not holding a user lock.

I've also rebased this on top of https://github.com/cyrusimap/cyrus-imapd/pull/3859, which already messed with some locking and has been include-in-fastmail for a while.

In my Cassandane tests, this all works with the lock tracking in place.  The big changes are:

1) we take a global blank username lock for shared folders
2) we don't do mailboxes.db locking any more for ACL changes, instead we just lock the owner's user lock for everything - this simplifies the locking considerably for ACL changes - that's the first thing I found in Nottingham
3) we make sure we NEVER try to lock again once we've unlocked an index to stream a fetch response.

There's also a nice little performance improvement, any FETCH command which doesn't have an implicit SETSEEN doesn't ever need to take an exclusive lock, so it's been changed to only take a shared lock if it's just doing a fetch without a SETSEEN.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Expunge.pm
+++ b/cassandane/Cassandane/Cyrus/Expunge.pm
@@ -261,6 +261,9 @@ sub test_ipurge_mboxevent
     );
     my $events = $self->{instance}->getnotify();
 
+    # if it stays selected you see the intermittent state
+    $talk->unselect();
+
     # the messages we just created should've been expunged
     $stat = $talk->status($shared_folder, '(highestmodseq unseen messages)');
     $self->assert_num_equals(0, $stat->{unseen});
--- a/imap/conversations.c
+++ b/imap/conversations.c
@@ -375,15 +375,13 @@ EXPORTED int conversations_open_path(const char *fname, const char *userid, int 
     open_conversations = open;
 
     /* first ensure that the usernamespace is locked */
-    if (userid) {
-        int haslock = user_isnamespacelocked(userid);
-        if (haslock) {
-            if (!shared) assert(haslock != LOCK_SHARED);
-        }
-        else {
-            int locktype = shared ? LOCK_SHARED : LOCK_EXCLUSIVE;
-            open->local_namespacelock = user_namespacelock_full(userid, locktype);
-        }
+    int haslock = user_isnamespacelocked(userid);
+    if (haslock) {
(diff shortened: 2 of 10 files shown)


ksmurchison (commented):
This all looks sane to me

brong:
@ksmurchison happy for me to merge it?  Then we can remove the xconv commands and etc!

ksmurchison:
Yes. But let's check with @ellie first. We may want to fork 3.8 before we merge new work

ellie:
> Yes. But let's check with @ellie first. We may want to fork 3.8 before we merge new work

Hello 👋 

I'm not sure you meant me 🤣 cool project, but I don't really have many opinions on this

ksmurchison:
Sorry, I meant @elliefm

elliefm:
We're in feature freeze until I fork the 3.8 branch, which I plan to do this week.  I'll send round an email to the devel list once I've done so; please don't merge new features or substantial changes to master until that email arrives.  (Trivial bug fixes are fine.)

brong:
@ellie yeah this is definitely not trivial!  THOUGH - it's also something that would be great to have in 3.8 because locking inversions suck.

ellie:
> @ellie yeah this is definitely not trivial!  THOUGH - it's also something that would be great to have in 3.8 because locking inversions suck.

Wow I'm popular on this one 🤣

You meant @elliefm 😊

elliefm:
> it's also something that would be great to have in 3.8 because locking inversions suck.

@brong Yeah I agree -- but it should get into 3.8 by careful consideration, not by luck of sliding under the closing gate.  I see it's tagged for fm builds now, great.  Once we feel like it's good enough to land on master (and has done so), I'll backport it to 3.8 -- assuming 3.8 has not already become a stable release at that point.

elliefm:
I think we'll need this on 3.6 to fix #4611.  I've already done the backport while investigating if it was possible, and that's now sitting in #4613 if I decide to proceed.

PR #4360, opened 2022-12-29
calalarmd: Update nextcheck on floating time events when calendar/principal TZ changes

ksmurchison:
If a user changes the TZ on a floating time event with alarms, we may trigger the (next) alarm too late if the change in TZ has a greater offset from UTC (toward the east).

### DIFF
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -2320,3 +2320,79 @@ static int upgradev4(sqldb_t *db)
 
     return r;
 }
+
+struct floating_rock {
+    struct caldav_db *caldavdb;
+    struct mailbox *mailbox;
+    const char *userid;
+};
+
+static int floating_cb(void *rock, struct caldav_data *cdata)
+{
+    struct floating_rock *frock = (struct floating_rock *) rock;
+
+    return update_alarmdb(mailbox_name(frock->mailbox), cdata->dav.imap_uid,
+                          time(0), ALARM_CALENDAR, 0, 0, 0, NULL);
+}
+
+static int calendar_cb(const mbentry_t *mbentry, void *rock)
+{
+    struct floating_rock *frock = (struct floating_rock *) rock;
+    const char *tzid_annot =
+        DAV_ANNOT_NS "<" XML_NS_CALDAV ">calendar-timezone-id";
+    const char *tz_annot =
+        DAV_ANNOT_NS "<" XML_NS_CALDAV ">calendar-timezone";
+    struct buf attrib = BUF_INITIALIZER;
+
+    if ((annotatemore_lookupmask(mbentry->name,
+                                 tzid_annot, frock->userid, &attrib) == 0 &&
+         buf_len(&attrib)) ||
+        (annotatemore_lookupmask(mbentry->name,
+                                 tz_annot, frock->userid, &attrib) == 0 &&
+         buf_len(&attrib))) {
+        /* calendar has its own time zone for floating events -- skip it */
+        return 0;
+    }
+
+    int r = mailbox_open_irl(mbentry->name, &frock->mailbox);
+    if (!r) {
+        r = caldav_get_floating_events(frock->caldavdb, mbentry,
+                                       &floating_cb, &frock);
+    }
+
+    mailbox_close(&frock->mailbox);
+
+    return r;
+}
+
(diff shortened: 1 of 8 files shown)


imap/caldav_alarm.c
```diff
@@ -2343,3 +2320,92 @@ static int upgradev4(sqldb_t *db)
...
+static int floating_cb(void *rock, struct caldav_data *cdata)
+{
+    struct floating_rock *frock = (struct floating_rock *) rock;
+    struct index_record record = {0};
+    icalcomponent *ical = NULL;
+
+    int r = mailbox_find_index_record(frock->mailbox, cdata->dav.imap_uid, &record);
+    if (!r && (ical = record_to_ical(frock->mailbox, &record, NULL))) {
+        if (has_alarms(ical, frock->mailbox, record.uid, NULL)) {
+            time_t nextcheck = process_alarms(mailbox_name(frock->mailbox),
```
brong, line 2341: A faster option here (just to make this action cheap) might be to update all the nextcheck to `now` and then have calalarmd process the alarms - that way it does all the batching and starvation prevention internally, and avoids this potentially processing thousands or millions of events in a single blocking operation.

imap/caldav_db.c
```diff
@@ -1468,3 +1468,84 @@ EXPORTED const char *caldav_comp_type_as_string(unsigned comp_type)
...
+        }
+    }
+
+    return tz;
+}
+
+#define CMD_SELFLOATING CMD_READFIELDS \
+    " WHERE mailbox = :mailbox AND dtstart NOT LIKE '%T%' ORDER BY imap_uid;"
+
+EXPORTED int caldav_get_floating(struct caldav_db *caldavdb,
```
brong, line 1531: I think this needs a more complete name - I initially thought it was getting the floating timezone, rather than fetching all the UIDs of floating events.

brong (changes requested):
I think what we actually want (from my comments) is to just set nextcheck to now for all floating events in each calendar.  We don't even need to open it, just lookup the annotation to see if it has its own timezone, then run a single SQL statement to update the nextcheck for all matching events.

There's a question here about "should the alert re-fire if it's already fired but the timezone moved".  I'm happy with the answer being no here, since it already won't with this design.

imap/http_caldav.c
```diff
@@ -6190,7 +6190,18 @@ static int proppatch_timezone(xmlNodePtr prop, unsigned set,
...
                 /* Set CALDAV:calendar-timezone */
-                proppatch_todb(prop, set, pctx, propstat, (void *) tz);
+                r = proppatch_todb(prop, set, pctx, propstat, (void *) tz);
+
+                if (!r) {
+                    /* Update alarms on floating time events */
+                    icaltimezone *ical_tz = icaltimezone_new();
+
+                    icaltimezone_set_component(ical_tz, icalcomponent_clone(vtz));
+                    caldav_alarm_update_floating(pctx->mailbox,
```
brong, line 6200: Oh yeah, I almost forgot.  You can't do it at this layer.  This needs to be triggered down in the annotations layer with a string match on the annotation, because otherwise the replica calalarm database won't be updated.

brong (changes requested):

imap/caldav_alarm.c
ksmurchison: Done

imap/caldav_db.c
ksmurchison: Done

imap/http_caldav.c
ksmurchison: I moved the updating into a annotation_set_timezone() callback and used the annotate_state_store() API which uses the callbacks (unlike annotate_state_writemask())

ksmurchison:
@brong GETMETEDATA issue should be resolved.  This is ready for another review

ksmurchison:
@brong I changed to doing the caldav_alarm update at the lowest level which should catch all APIs.  

I tried writing a replication test and I can't figure out how/when calalam.db gets updated on the replica.  It appears that is doesn't get written until and alarm on the master is triggered.  Does this sound correct?

brong (approved):
This looks good - write_entry is a good place to call this.  And the test checks it.  It would be nice to have a replication test too, but that's a general issue with our current tests for annotation stuff.

PR #4361, opened 2022-12-31
IMAP extensions audit

ksmurchison:
Brought RFC 4551/5162 implementation up to RFC 7162

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -363,7 +363,7 @@ static struct capa_struct base_capabilities[] = {
     { "CHILDREN",              CAPA_POSTAUTH,           { 0 } }, /* RFC 3348 */
     { "COMPRESS=DEFLATE",      CAPA_POSTAUTH|CAPA_STATE,         /* RFC 4978 */
       { .statep = &imapd_compress_allowed }                   },
-    { "CONDSTORE",             CAPA_POSTAUTH,           { 0 } }, /* RFC 7162, but the implementation is likely from RFC 4551 */
+    { "CONDSTORE",             CAPA_POSTAUTH,           { 0 } }, /* RFC 7162 */
     /* CONTEXT=SEARCH   RFC 5267 is not fully implemented. Only ESORT is ready */
     /* CONTEXT=SORT     RFC 5267 is not fully implemented. Only ESORT is ready */
     /* CONVERT          RFC 5259 is not implemented */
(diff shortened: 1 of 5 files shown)


imap/index.c
```diff
@@ -1928,7 +1927,7 @@ static int needs_modseq(const struct searchargs *searchargs,
 {
     int i;
 
-    if (search_expr_uses_attr(searchargs->root, "modseq"))
+    if (searchargs->state & GETSEARCH_MODSEQ)
```
elliefm, line 1930: I'm surprised you're not getting a warning here about bitwise-and in a condition being a possible typo?  Usually these want an extra set of parens to convince the compiler that you meant what you wrote, and not `&&`, like this:

```suggestion
    if ((searchargs->state & GETSEARCH_MODSEQ))
```

Though I just tried building this branch and didn't get a warning for it either, so maybe gcc doesn't warn about that anymore, or is now smart enough to not warn when one of the operands is clearly a bitmask.  Curious!

imap/imapd.c
```diff
@@ -302,7 +302,7 @@ struct capa_struct {
 static struct capa_struct base_capabilities[] = {
 /* pre-auth capabilities */
     { "IMAP4rev1",             3 },
-    { "LITERAL+",              3 }, /* This is in RFC 7888, but likely the implementation is for RFC 2088 */
+    { "LITERAL+",              3 }, /* RFC 7888 */
```
elliefm, line 305: I have a real vague memory that there was some detail of `LITERAL+` and/or `LITERAL-` that we were cheating at, which was acceptable with the older RFC but not strictly compliant with the newer one.  But I don't remember what that detail was, and maybe we already fixed it up and handle it properly now, but just neglected to update this comment at the time.  Do you know?

elliefm (approved):
Looks good, couple of conversational comments inline.

I haven't cross-referenced any of the number/section changes to check for typos.  I can if you'd like me to, just let me know.

imap/index.c
ksmurchison: I think I've been using constructs like this for a while and I see numerous uses throughout our code, so I think we're good here

imap/imapd.c
```diff
@@ -333,7 +333,7 @@ static struct capa_struct base_capabilities[] = {
     { "LIST-MYRIGHTS",         2 }, /* RFC 8440 */
     { "LIST-STATUS",           2 }, /* RFC 5819 */
     /* LOGIN-REFERRALS  RFC 2221 is not implemented */
-    /* LOGINDISABLED    RFC 2595/RFC 3591 is announced conditionally in capa_response() */
+    /* LOGINDISABLED    RFC 2595/RFC 3501 is announced conditionally in capa_response() */
```
dilyanpalauzov, line 336: Please add also RFC 9051. https://www.iana.org/assignments/imap-capabilities/imap-capabilities.xhtml mentions two RFCs after LOGINDISABLED.

imap/imapd.c
```diff
@@ -344,7 +344,7 @@ static struct capa_struct base_capabilities[] = {
     /* NOTIFY           RFC 5465 is not implemented */
     { "OBJECTID",              2 }, /* RFC 8474 */
     { "PREVIEW",               2 }, /* RFC 8970 */
-    { "QRESYNC",               2 }, /* RFC 7162, but the implementation is likely from RFC 4551 and RFC 5162 */
+    { "QRESYNC",               2 }, /* RFC 7162 */
```
dilyanpalauzov, line 347: Since RFC 4551 is obsoleted, can it be removed from rfc-support.rst ?

PR #4363, opened 2023-01-04
pre-3.8 fixes

elliefm:
One of the early steps in getting ready for the next major version is making sure a working distribution can be built from the current master branch.  Here's what I had to fix to make that happen


ksmurchison (approved):
LGTM

PR #4364, opened 2023-01-05
Rewrite IMAP proxy_copy() to use a single (reusable) proxy_fetch()

ksmurchison:
Stumbled across the double FETCH response parsing while looking for similar code to steal for REPLACE


elliefm (commented):
The diff for the main commit in this one is really hard to read, because it's smashing the changes together, treating a few random lines as unchanged and rewriting the world around them, instead of recognising and showing that those lines just got moved into new functions, but the world itself didn't change much.

I'm expecting to see that a bunch of logic has been moved out of `proxy_copy()` and into `proxy_fetch()` and/or `copy_cb()`, and that `proxy_copy()` now calls `proxy_fetch()` with `copy_cb` as a callback argument, instead of doing that work itself -- and that would be fine.  But I can't really see that from the diff, much less verify that it's been done well.

I wonder if reordering the functions can produce a usable diff.  If I'm inferring the relationship between these functions correctly, maybe an ordering like

copy_cb()
proxy_copy()
proxy_fetch()

will help diff produce more semantic output?

elliefm:
Getting these errors and failures from this branch:

```
[ERROR ] Cyrus::Metadata.modseq
[FAILED] Cyrus::Metadata.permessage_getset
[FAILED] Cyrus::Metadata.permessage_unknown_allowed
[ERROR ] Cyrus::Metadata.unchangedsince
```

ksmurchison:
@elliefm Moved proxy_fetch() as suggested.  Hopefully is helpful.
The Metadata test failures have been fixed in

imap/imap_proxy.c
```diff
@@ -864,32 +865,210 @@ static char *editflags(char *flags)
...
+    }
+    if (items & FETCH_RFC822) {
+        prot_printf(backend_current->out, "%c%s", sep, "Rfc822.peek");
+        sep = ' ';
+    }
+    if (items & FETCH_UID) {
+        prot_printf(backend_current->out, "%c%s", sep, "Uid");
+        sep = ' ';
+    }
+    prot_puts(backend_current->out, ")\r\n");
```
elliefm, line 1066: I think if `items` was `0` here, then we're going to output a `)` without having output a `(` first.  Looks like the other places using this pattern do something like `if (sep == '(') prot_puts(out, "(");` just before outputting the `)` .

elliefm (commented):
Not finished looking at this, but I really badly need to stop and eat some lunch, and I don't want to accidentally lose what I've already written, so here's a partial review for now where I don't fully understand what's going on yet

imap/imap_proxy.c
ksmurchison: I've added protection from this by bailing if `items == 0`

imap/imap_proxy.c
```diff
@@ -864,32 +865,210 @@ static char *editflags(char *flags)
...
+                free(cur_msg->flags);
+                sep = ' ';
+            }
+            if (cur_msg->idate) {
+                prot_printf(imapd_out, "%cINTERNALDATE %s", sep, cur_msg->idate);
+                free(cur_msg->idate);
+                sep = ' ';
+            }
+            prot_printf(imapd_out, ")\r\n");
+        }
```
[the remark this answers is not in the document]
ksmurchison: I rearranged the if/else if/else to make the logic clearer (I hope)

elliefm (approved):
[Rewriting this because I thought I already submitted it a minute ago, but suddenly there's no evidence I did so??  If you see a similar review from me elsewhere, maybe I put it on the wrong PR somehow...]

I think I finally mostly understand how this works.  The if/else changes helped, so too I think did sleeping on it for a few days.

It's a little awkward, but I think it just needs to be, the work being done is just awkward work.  I know you've already used this as part of the REPLACE implementation (though I haven't looked at that yet), so that proves it's reusable.

PR #4366, opened 2023-01-06
backend: update backend_version() for 3.8 and 3.9

elliefm:
Updates `backend_version()` to recognise versions 3.8 and 3.9


elliefm:
This is now on all release branches back to 2.4

PR #4367, opened 2023-01-06
Fix STORE [+/-]FLAGS.SILENT

ksmurchison:
This has been broken for quite some time, and I stumbled across it while working on a REPLACE implementation #4368

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -158,7 +158,7 @@ static int index_store_annotation(struct index_state *state, uint32_t msgno,
 static int index_fetchreply(struct index_state *state, uint32_t msgno,
                             const struct fetchargs *fetchargs);
 static void index_printflags(struct index_state *state, uint32_t msgno,
-                             int usinguid, int printmodseq);
+                             unsigned tell_flags);
 static char *get_localpart_addr(const char *header);
 static char *get_displayname(const char *header);
 static char *index_extract_subject(const char *subj, size_t len, int *is_refwd);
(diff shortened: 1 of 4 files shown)


imap/index.c
```diff
@@ -4229,7 +4237,7 @@ static void index_fetchflags(struct index_state *state,
     bit32 flagmask = 0;
     struct index_map *im = &state->map[msgno-1];
 
-    prot_printf(state->out, "* %u FETCH (FLAGS ", msgno);
+    prot_puts(state->out, "FLAGS ");
```
elliefm, line 4240: The comment above the function still says "Also sends preceeding * FLAGS if necessary", but that's not quite right anymore.

It also says "Does not send the terminating close paren or CRLF", and it's true that it doesn't send the closing paren or CRLF for the whole `FETCH (...)`, but it does send the closing paren for the `FLAGS (...)` clause, so the second-last line of the function is it sending a `)` -- which makes the comment look wrong at first glance, and it takes some careful reading to figure out what's really going on.

imap/index.c
```diff
@@ -4015,21 +4019,25 @@ EXPORTED void index_tellchanges(struct index_state *state, int canexpunge,
...
     for (msgno = 1; msgno <= state->exists; msgno++) {
         im = &state->map[msgno-1];
 
         /* report if it's changed since last told */
-        if (im->modseq > im->told_modseq)
-            index_printflags(state, msgno, printuid, printmodseq);
+        if (im->modseq > im->told_modseq &&
+            (tell_flags & (TELL_SILENT | TELL_MODSEQ)) != TELL_SILENT)
+            index_printflags(state, msgno, tell_flags);
+        im->told_modseq = im->modseq;
```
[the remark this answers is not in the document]
ksmurchison: That is correct.  I added a comment here

imap/index.c
ksmurchison: Fixed the comment

elliefm (approved):

PR #4368, opened 2023-01-07
IMAP REPLACE

ksmurchison:


elliefm (approved):
Looks good.  The use of proxy_fetch here winds up easy to read, with all that awkwardness from the implementation encapsulated away, nice.

Issue #4369, opened 2023-01-08
ctl_mboxlist -v fails with mailbox names containing a '-'


gshapiro:
After battling it for a while, I discovered ctl_mboxlist -v just can't handle mailboxes with a '-' in the name.  While everything else functions properly with the folder (management, saving/reading messages, etc.), the verification is misleading:

% sbin/ctl_mboxlist -v
'user.gshapiro.Domains.Retsiger-Announce' has a directory '/var/spool/imap/user/gshapiro/Domains/Retsiger-Announce' but no DB entry
'user.gshapiro.Domains.Retsiger-Announce' has a DB entry but no directory on partition 'default'

If I ask it to dump, it sees the folder:

% sbin/ctl_mboxlist -d | grep Retsiger-Announce
"user.gshapiro.Domains.Retsiger-Announce": {"name": "user.gshapiro.Domains.Retsiger-Announce", "mtime": "1673138436", "uidvalidity": "1673138437", "createdmodseq": "1", "foldermodseq": "1", "mbtype": "el", "partition": "default", "acl": {"gshapiro": "lrswipcda"}, "uniqueid": "a8f8164e-bccb-44ee-96cc-48635223dacf", "name_history": []},

Running on FreeBSD 13.1-RELEASE-p5:

% pkg info | grep cyrus
cyrus-imapd36-3.6.0            Cyrus mail server, supporting POP3 and IMAP4 protocols
cyrus-sasl-2.1.28              RFC 2222 SASL (Simple Authentication and Security Layer)

Potentially relevant configuration lines:

altnamespace: yes
unixhierarchysep: no

a-tze:
What is the output of `mbpath user.gshapiro.Domains.Retsiger-Announce`

Just checking if related to #4187 because I cannot reproduce this behaviour if mailbox_legacy_dirs is enabled.

Is it possible that the parent mailbox was created before using cyrus 3.6.x but that specific mail folder was created after the update?

gshapiro:
I started seeing this after upgrading a 3.0 install to 3.2, then 3.4, then 3.6 (running verifications on each step along the way and using the server for a little while to ensure things were working on each jump).  All of the mailboxes existed in the 3.0 days so it doesn't follow the pattern of parent mailbox < 3.6, specifc = 3.6 because specific was also < 3.6.

[...]

Issue #1738, opened 2015-10-09
ctl_mboxlist -v  takes deleted mailboxes into account


brong:
From: [Wolfgang Breyha](wbreyha@gmx.net)
Bugzilla-Id: 3910
Version: 2.5.x (next)
Owner: [Bron Gondwana](brong@fastmail.fm)

a-tze:
Likely fixed in #4371

rsto:
Agreed, this got fixed by @a-tze (thanks!)

PR #4371, opened 2023-01-08
ctl_mboxlist: support uuid mailboxes in ctl_mboxlist -v

### LINKED ISSUE #4187: ctl_mboxlist -v needs to be able to report broken uuid mailboxes
This thing looks at everything in the mailboxes.db and everything on the spool, and reports things that are missing from one or the other.

Ken observes that it doesn't currently seem able to deal with uuid mailboxes.

a-tze:
Fixes #4187

### DIFF
--- a/imap/ctl_mboxlist.c
+++ b/imap/ctl_mboxlist.c
@@ -89,6 +89,7 @@
 /* generated headers are not necessarily in current directory */
 #include "imap/imap_err.h"
 #include "imap/mupdate_err.h"
+#include "lib/ptrarray.h"
 
 enum mboxop { DUMP,
               M_POPULATE,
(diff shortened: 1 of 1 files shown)


a-tze:
Sorry for the force-pushs.. but now I'm really finished and it seems to fix #4369 as well.

a-tze:
Probably fixes #1738 as well.

ksmurchison (approved):
This looks sane to me.
I wonder if we could use a ptrarray or dynarray for found_list rather than have it do its own memory management

a-tze:
Do you mean for the items of found_list or for the strings contained/copied within?

I don't think that the memory management of the found_list is so terrible. The double-capacity realloc logic could use an upper limit so it doesnt blow off the roof in extreme situations, but I think the string copy operations of the things contained in the found_list are more relevant regarding performance and memory footprint.

ksmurchison:
I was talking about found_list. ptrarray was designed for this exact purpose. But I believe the code that you are tweaking predates ptrarray.

a-tze:
Let me have a look into it. If the found_list is not used in the other command implementation of ctl_mboxlist, a refactoring could be a good thing.

a-tze:
Refactoring done. Could someone possibly do this workflow approval thing?

a-tze:
added the Signed-off-by header that the auto-check complained about.

a-tze:
ping :)

ksmurchison (approved):

elliefm:
When backporting this, it'll also need 3be3f1699

a-tze:
@elliefm Has this been a "note for self" or should I ammend that commit here? :)

elliefm:
That one's for me, sorry for confusion :)

PR #4372, opened 2023-01-08
docs: fix typo

elliefm:


ksmurchison (approved):
I don't think that we require a PR for a trivial fix to comments/docs.  At least I wouldn't complain if you just went ahead and pushed this commit without a PR

PR #4374, opened 2023-01-11
lmtp_sieve.c: strip VALARMs from iTIP reply before restriction check

ksmurchison:
Apparently Google sometimes includes VALARMs in replies, and this causes libical restriction checks to fail


elliefm (approved):
Looks good

PR #4375, opened 2023-01-11
jmap_contact.c: properly update a v4 group vCard

ksmurchison:
This fixes a bug where if a v4 group vCard is updated via JMAP, the VERSION is blindly changed to 3 and X-ADDRESSBOOKSERVER-MEMBER properties are added while MEMBER properties are left behind

### DIFF
--- a/imap/jmap_contact.c
+++ b/imap/jmap_contact.c
@@ -850,10 +850,15 @@ static int _add_group_entries(struct jmap_req *req,
                               struct vparse_card *card, json_t *members,
                               json_t *invalid)
 {
-    vparse_delete_entries(card, NULL, "X-ADDRESSBOOKSERVER-MEMBER");
+    const char *group_propname = "X-ADDRESSBOOKSERVER-MEMBER";
+    struct vparse_entry *ventry = vparse_get_entry(card, NULL, "VERSION");
     int r = 0;
     size_t index;
     struct buf buf = BUF_INITIALIZER;
+
+    if (ventry && atof(ventry->v.value) >= 4.0) group_propname =  "MEMBER";
+
+    vparse_delete_entries(card, NULL, group_propname);
 
     for (index = 0; index < json_array_size(members); index++) {
         const char *item = _json_array_get_string(members, index);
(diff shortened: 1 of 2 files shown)


imap/jmap_contact.c
```diff
@@ -850,11 +850,16 @@ static int _add_group_entries(struct jmap_req *req,
...
                               json_t *invalid)
 {
-    vparse_delete_entries(card, NULL, "X-ADDRESSBOOKSERVER-MEMBER");
+    const char *group_propname = "X-ADDRESSBOOKSERVER-MEMBER";
+    struct vparse_entry *ventry = vparse_get_entry(card, NULL, "VERSION");
     int r = 0;
     size_t index;
     struct buf buf = BUF_INITIALIZER;
 
+    if (ventry && ventry->v.value[0] == '4') group_propname =  "MEMBER";
```
elliefm, line 859: In the future, this will treat version 5 (and also all subsequent versions except 40-49, 400-499, etc) as not having the "MEMBER" property, and using "X-ADDRESSBOOKSERVER-MEMBER" instead, which will be embarrassing if we don't catch it in time 😅 

I think it's safe to assume that the property name won't change again now that it's become a real property.  So what about something like this... 

```suggestion
    if (ventry && (isdigit(ventry->v.value[1]) || (isdigit(ventry->v.value[0]) && ventry->v.value[0] >= '4'))) group_propname =  "MEMBER";
```

I haven't tried to wrap it, and I'm not sure the parens match, since I'm typing it live in the github window rather than a proper editor, but you get the idea -- if the version is more than one digit, or if the first digit is four-or-greater, then "MEMBER"

elliefm (changes requested):

imap/jmap_contact.c
ksmurchison: @elliefm Addressed this in a slightly different way. Please re-review

elliefm: Yeah nice, that's a bit more readable than my suggestion

elliefm (approved):

PR #4376, opened 2023-01-12
mailbox_add_dav() write lock

ksmurchison:
Need a write lock so we can delete stale index records


brong:
@ksmurchison there appears to be an extra commit on here?  jscontact thing.

The rest looks good!  Thanks.

brong (approved):

PR #4377, opened 2023-01-13
http_caldav_sched.c: adding an attendee to an override must send iTIP

ksmurchison:
Adding an attendee to an existing override, with no other changes, was failing to send an iTIP request

### DIFF
--- a/cassandane/Cassandane/Cyrus/Caldav.pm
+++ b/cassandane/Cassandane/Cyrus/Caldav.pm
@@ -1320,6 +1320,78 @@ EOF
     );
 }
 
+sub test_invite_add_another_to_override
+    :VirtDomains :min_version_3_0 :needs_component_httpd
+{
+    my ($self) = @_;
+
+    my $service = $self->{instance}->get_service("http");
+    my $CalDAV = Net::CalDAVTalk->new(
+        user => "cassandane%example.com",
+        password => 'pass',
+        host => $service->host(),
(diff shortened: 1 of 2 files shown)


cassandane/Cassandane/Cyrus/Caldav.pm
```diff
@@ -1320,6 +1320,79 @@ EOF
...
+
+    $CalDAV->Request('PUT', $href, $card, 'Content-Type' => 'text/calendar');
+
+    $self->assert_caldav_notified(
+        { recipient => "friend\@example.com", is_update => JSON::false, method => 'REQUEST' },
+    );
+
+    $card =~ s/SEQUENCE:0/SEQUENCE:1/;
+    $card =~ s/RECURRENCE-ID/ATTENDEE;PARTSTAT=NEEDS-ACTION;RSVP=TRUE:MAILTO:friend2\@example.com\nRECURRENCE-ID/;
+warn Dumper($card);
```
elliefm, line 1388: Can you drop this warning please, or convert it an xlog if it's still useful? Thanks

elliefm (approved):
One nit, but doesn't need re-review one it's fixed

cassandane/Cassandane/Cyrus/Caldav.pm
ksmurchison: Good catch.  Removed

PR #4378, opened 2023-01-15
imapd.c: fix list_data_remote() string arrays to match bitmasks

ksmurchison:
Somehow the string arrays and bitmasks got out of sync


elliefm (approved):
Looks good

elliefm:
I tracked down where these got out of sync, and it was only last year, and was never backported.  So this needs to be backported to 3.8, but the older branches never got the bad change and don't need fixing.

PR #4379, opened 2023-01-16
IMAP QUOTA (RFC 9208) Implementation

ksmurchison:

PR #4382, opened 2023-01-18
backend: forward-safe backend_version()

elliefm:
If the backend's banner version looks sensible, but is larger than anything we know about, infer that it supports at least the same MAILBOX_MINOR_VERSION that this server does.

This fixes forever the problem where XFER would downgrade and lose data when sending mailboxes to a backend newer than it recognised; and removes the need to update backend_version(), backport it, and do releases from all the older branches, every time a new Cyrus version is created.

backend_version() still needs to be updated appropriately any time MAILBOX_MINOR_VERSION changes, but only when it changes.

This also simplifies handling of banner versions from 3.0 forward.

This needs to be backported to all the usual branches (including 3.0, 2.5, and 2.4), but that will be the last time it needs to happen.

### DIFF
--- a/changes/next/4380-backend-version-bump-watermark
+++ b/changes/next/4380-backend-version-bump-watermark
@@ -1,0 +1,21 @@
+Description:
+
+XFER to newer backends now assumes at least the current mailbox version,
+rather than the oldest supported mailbox version.
+
+
+Config changes:
+
+None
+
+
+Upgrade instructions:
+
+This is the future absence of an upgrade instruction: a server containing
+this change no longer needs to be upgraded to a version that specifically
+recognises newer backends prior to XFERing mailboxes to them.
+
+
+GitHub issue:
+
+https://github.com/cyrusimap/cyrus-imapd/issues/4380
--- a/cunit/backend.testc
+++ b/cunit/backend.testc
@@ -440,6 +440,7 @@ static void test_unrecognised_backend_version(void)
     struct backend_version_data data[] = {
         { "BOGUS", 6 }, /* completely unrecognised */
         { "Cyrus IMAP v2.3.B (OGUS)", 6 }, /* junk 2.3 revision */
+        { "Cyrus IMAP 10.0.0", MAILBOX_MINOR_VERSION }, /* unknown future version */
     };
 
     default_conditions();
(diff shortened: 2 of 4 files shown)


ksmurchison (approved):
Very nice!

elliefm:
Holding this for a moment while we work out what's going on with #4379 and #4384

Whatever we end up doing there, this will need a rebase-and-update to match before it lands

elliefm:
Actually, this needs more work.  The scanf string matches the banners presented by the cunit tests, but _those are faked_.

Here's an example of a real banner, as seen during one of Cassandane's XFER tests with some extra logging added:

```
* OK [CAPABILITY IMAP4rev1 AUTH=PLAIN ENABLE ID LITERAL+ SASL_IR MUPDATE=mupdate://localhost:9100/] localhost:9103 Cyrus IMAP 3.9.0-alpha0-55-g7804af3998 server ready
```

This passes the Cassandane test, because Cassandane generally runs all instances from the same Cyrus build, so it gets the "identical banner? identical version!" treatment.  But if it weren't, and had fallen through to the actual sscanf, it wouldn't have matched, and would have fallen through to the legacy checks, and wound up downgrading to MMV=6 anyway.  D'oh!

elliefm:
I think this is fixed now, but I'll let my subconscious process it over the weekend, and see if I suddenly think of any other embarrassing holes in it...

elliefm:
So the original code relied only on substring matches, so it was able to handle capabilities in the banner, and pre-release decorators in the version strings, transparently.  But there weren't any tests for those cases.

My first attempt at this made the version parser too specific -- it would have choked on either of these cases.  But since they weren't tested, the tests couldn't detect that I'd broken it.

I've since extended the tests to cover both cases, and then fixed the implementation to pass them.  The first commit is still the same as it was; I think the history is useful here so I won't just squash it all together.

@ksmurchison can you have another look at this please?  Thanks :)

elliefm:
@ksmurchison This is already marked as "approved", but I've made substantial changes to the logic (and improved the tests) since you approved it.  Can you review it again please?  If the interface won't let you mark it as "approved" a second time, can you indicate approval of the changes in a comment or something?  Thanks

We may need to change this code again shortly for the QUOTA/MMV19 changes... but I'm blocked on 3.8 work by the need to do 3.8-compatible XFER releases for the older branches first, for which I need this to have landed so that I can backport it.  So I'd rather just land it, and add MMV19 recognition later, rather than waiting for MMV19 to land.

ksmurchison:
@elliefm This still looks good to me.  And if it does what it says on the tin, consider it approved (again)

elliefm:
This has now been backported all the way back to 2.4.

elliefm:
These fixes are now available in releases 3.6.1, 3.4.5, 3.2.11, and 3.0.18.  If you have one of these releases already, you will no longer need XFER compatibility fixes to accommodate future Cyrus version numbers.

PR #4383, opened 2023-01-18
rolling squatter: compact databases

brong:
Twoskip now delays compacting databases until after the locks are released, but squatter wasn't ever calling this, leading to doing all the compacts on shutdown.  In particular, this meant trying to compact lots of cyrus.indexed.db files


brong:
rebased on master, merging now.

PR #4385, opened 2023-01-23
xunlink: error-logging wrapper around unlink

elliefm:
@brong suspects a weird problem we saw last week may have been due to ZFS rejecting an `unlink()`, and then things being left behind that we assumed were gone.  We can't tell though, because most of our code never even checks whether `unlink()` succeeded or not.  This is bad, but is a big and tricky problem to fix properly.

This PR doesn't fix it, but it does add logging to every `unlink()` call (and every `unlinkat()` call, but we don't currently use that), such that if the result is anything other than "success" or "failed because the file to unlink didn't exist", we'll log an IOERROR.  And if it happens again, it'll be in the logs.

It tries hard to present _exactly the same_ interface as `unlink()` itself, so that if/when we do add proper error handling, the code to do so should be identical regardless

No tests, because cunit can't mock system calls, so I can't usefully fake unlink errors :(

**Note to reviewers**: I'd suggest reviewing this commit-by-commit.  The first commit contains the new code, is interesting, and needs careful review; the second is the mind-numbing update of every extant call to `unlink()`, which also needs careful review, but is boring :)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -802,7 +802,8 @@ include_HEADERS = \
 	lib/tok.h \
 	lib/vparse.h \
 	lib/wildmat.h \
-	lib/xmalloc.h
+	lib/xmalloc.h \
+	lib/xunlink.h
 
 nodist_include_HEADERS = \
 	lib/imapopts.h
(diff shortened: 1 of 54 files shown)


elliefm:
I didn't update the `unlink()` calls in the cunit tests (yet), mostly because I forgot to include *.testc files when searching for occurrences.  Not sure if I should do so or not, will give it some thought.

elliefm:
Decided to update the cunit tests to use xunlink, and have done so

elliefm:
Good catch, fixed!

PR #4387, opened 2023-01-24
IMAP QUOTA Updates (RFC 9208)

ksmurchison:


elliefm (commented):
A few comments inline...

For some reason `JMAPBackup.restore_calendars_all` fails for me on this branch, reproducibly, but it looks like it passed under CI.  Weird.  Does it pass for you?

```
1) JMAPBackup.restore_calendars_all
 expected 'error', got 'Backup/restoreCalendars' at Cassandane/Cyrus/JMAPBackup.pm line 717.
```

"expected 'error'..." is curious, but I haven't looked any closer yet

imap/mailbox.c
```diff
@@ -1924,6 +1926,13 @@ static int mailbox_buf_to_index_header(const char *buf, size_t len,
 
     i->changes_epoch = ntohl(*((bit32 *)(buf+OFFSET_CHANGES_EPOCH)));
 
+    if (i->minor_version < 18) goto crc;
+
+    i->quota_deleted_used = align_ntohll(buf+OFFSET_QUOTA_DELETED_USED);
+    i->quota_expunged_used = align_ntohll(buf+OFFSET_QUOTA_EXPUNGED_USED);
+
+    /* v19 rearranged fields to get modseq_t and quota_t on 8-byte boundaries */
```
brong, line 1934: both these functions are going to have to have a fork placed earlier up so they can read the old formats correctly from the old offsets, since the constants have changed!

brong (changes requested):
The parser won't read and write old index formats correctly at the moment.  We probably need more test cases with those old formats!  And maybe something to back-convert and check for binary correctness of the file.

cassandane/Cassandane/Cyrus/MurderIMAP.pm
```diff
@@ -117,6 +117,18 @@ sub test_frontend_commands
     $self->assert_not_null($result);
     $self->assert_str_equals('ok', $frontend->get_last_completion_response());
 
+    my $res_mailbox = 'MAILBOX';
+    my ($maj, $min) = Cassandane::Instance->get_version();
```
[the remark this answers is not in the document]
ksmurchison: Cassandane::Instance->get_version('murder') seems to get the version of the backend, so I assumed that the default would be the frontend.  If not, I'm not sure we have the plumbing to to get the frontend version.

imap/sync_support.c
```diff
@@ -821,8 +821,13 @@ void sync_encode_quota_limits(struct dlist *kl, const quota_t limits[QUOTA_NUMRE
...
 
     for (res = 0 ; res < QUOTA_NUMRESOURCES ; res++) {
-        if (limits[res] >= 0)
+        if (limits[res] >= 0) {
             dlist_setnum32(kl, quota_names[res], limits[res]);
+
+            /* For backwards compatibility, we also encode the
+             * legacy X-ANNOTATION-STORAGE and X-NUM-FOLDERS limits. */
+            dlist_updatenum32(kl, legacy_quota_names[res], limits[res]);
+        }
```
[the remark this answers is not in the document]
ksmurchison: I am not setup for that.  If you could test, that would be greatly appreciated.

imap/mailbox.c
```diff
@@ -3179,42 +3190,50 @@ static void mailbox_quota_dirty(struct mailbox *mailbox)
...
+#define UPDATE_COUNT(count, is_add)             \
+    if (is_add) count++;                        \
+    else if (count) count--;
+
+#define UPDATE_QUOTA_USED(quota_used, size, is_add)                     \
+    if (is_add) quota_used += size;                                     \
+    /* corruption prevention - check we don't go negative */            \
+    else if (quota_used > size) quota_used -= size;                     \
+    else quota_used = 0;
+
```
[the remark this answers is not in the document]
ksmurchison: Yep, you're right.  Fixed

ksmurchison:
@elliefm The restore test was failing because I redefined OFFSET_CHANGES_EPOCH with the wrong value.  This has been fixed

imap/mailbox.c
ksmurchison: This should now be fixed

elliefm:
Multi-version test results.  In these results, "main" specifies the version of Cyrus that's built and installed to the prefix named in the `[cyrus default]` section of cassandane.ini, "replica" specifies the version installed to the `[cyrus replica]` prefix, "murder" the one in the `[cyrus murder]` prefix, and "cass" the version that's currently checked out in the source directory, from which the Cassandane tests were run.

I ran the Replication, MurderIMAP, and MurderJMAP suites only, because these are almost* the only test suites that make use of the mixed versioning functionality.

*I could also have set a `[cyrus backup]` prefix, installed something there, and run the Backups suite too... except that I forgot about it until just now while writing this up.

**main=imap_quota, replica=3.8, murder=3.8, cass=3.8:**
  all tests pass

**main=imap_quota, replica=3.8, murder=3.8, cass=imap_quota:**
```
1) MurderIMAP.replace_across_backends
 expected 'ok', got 'bad' at Cassandane/Cyrus/MurderIMAP.pm line 1139.
2) MurderIMAP.replace_same_backend
 <undef> unexpected at Cassandane/Cyrus/TestCase.pm line 1072.
 (this is in check_messages)
```

**main=3.8, replica=imap_quota, murder=imap_quota, cass=imap_quota:**
(n.b. this combination might be bogus)
```
1) MurderIMAP.frontend_commands
 <undef> unexpected at Cassandane/Cyrus/MurderIMAP.pm line 129.
```

**main=3.8, replica=imap_quota, murder=imap_quota, cass=3.8:**
  all tests pass

It looks like, as long as the Cassandane running is 3.8, it doesn't matter which arrangement of older or newer versions are installed, everything is happy.  Since in this case the tests don't have the updates, maybe they simply don't exercise anything interesting.

If the Cassandane running is the updated one from imap_quota, there's different MurderIMAP failures depending on which side of the equation is older or newer.

For murder tests, the `[cyrus default]` prefix is used for "backend1" (aka "instance"); the `[cyrus murder]` prefix is used for "backend2" and "frontend" (mupdate runs on frontend).

So when backend1 is imap_quota, and frontend/mupdate and backend2 are 3.8, we get the two "replace..." failures.  I'm not sure what this indicates, but it feels significant.  You should reproduce this setup so you can test it directly.

Whereas, when backend1 is 3.8, and frontend/mupdate and backend2 are imap_quota, we get the "frontend_commands" failure.  This particular case might just be bogus.  It would be nice if it worked, but Cassandane somewhat assumes that the version installed to its `[cyrus default]` prefix is the same version as Cassandane itself, and it might or might not be feasible to address the issue.  Though it could also just be because of the `get_version()` thing I pointed out while reading the diffs.

**To reproduce this setup:**

1) Add `[cyrus murder]` and `[cyrus replica]` sections to your cassandane.ini and configure each with their own prefix, different from your `[cyrus default]` one.  Check cassandane.ini.example for examples with comments.  Do the same for `[cyrus backup]` if you like -- it's not documented like the others, but it works the same way.
2) For each prefix you've configured, check out the Cyrus version you want to run there, do a from-scratch complete build with `./configure --prefix=/the/prefix ...` and install it
3) Do the same for your usual `[cyrus default]` prefix
4) Check out the version whose Cassandane you want to run.  Rebuild Cassandane's binary components with `make -C cassandane/utils`
5) Run Cassandane
6) Rinse and repeat for other combinations

To get back to normal operation when you're done, you can simply delete the extra Cyrus installations -- you can leave the config in cassandane.ini for next time.  Having a prefix configured for a directory that doesn't exist is the same as not having one configured (i.e. it just uses `[cyrus default]` for all processes), so you can leave it configured all the time, and just install/delete replica/murder/backup builds as needed

ksmurchison:
@elliefm Since changes/next/imap_quota states that frontend machines should be upgraded first, I *think* the only combination that I need to look at is 

main=3.8, replica=imap_quota, murder=imap_quota, cass=imap_quota

I will attempt to do so today.

ksmurchison:
@elliefm The Murder Replace tests have been fixed in #4394 

The Murder frontend test has been fixed for 3.9 frontend to 3.8 backend.  We don't support the reverse per the upgrade notes in changes/next/imap_quota

ksmurchison:
@elliefm @brong I think (hope) this is ready for a final review

imap/mailbox.c
```diff
@@ -1829,12 +1829,46 @@ EXPORTED struct entryattlist *mailbox_extract_annots(const struct mailbox *mailb
...
+static const struct index_header_offsets new_offsets = {
+    OFFSET_MAILBOX_OPTIONS,
+    OFFSET_LEAKED_CACHE,
+    OFFSET_HIGHESTMODSEQ,
+    OFFSET_DELETEDMODSEQ,
+    OFFSET_EXISTS,
+    OFFSET_CHANGES_EPOCH,
+    OFFSET_QUOTA_DELETED_USED,
+    OFFSET_QUOTA_EXPUNGED_USED
+};
```
elliefm, line 1863: Hmm... I wonder what this will look like in the future when we rearrange the offsets again, and now need to keep track of n+1 sets

imap/mailbox.h
```diff
@@ -339,36 +343,39 @@ struct mailbox_iter;
...
+#define OFFSET_HEADER_CRC 156
+
+#define ORIG_OFFSET_MAILBOX_OPTIONS 60
+#define ORIG_OFFSET_LEAKED_CACHE 64
+#define ORIG_OFFSET_HIGHESTMODSEQ 68
+#define ORIG_OFFSET_DELETEDMODSEQ 76
+#define ORIG_OFFSET_EXISTS 84
+#define ORIG_OFFSET_CHANGES_EPOCH 136
+#define ORIG_OFFSET_QUOTA_DELETED_USED 140
+#define ORIG_OFFSET_QUOTA_EXPUNGED_USED 148
```
elliefm, line 378: If we called these like `PRE19_OFFSET_...` instead of `ORIG_OFFSET_...`, then that might make it easier to manage next time we need to rearrange them, and a little less ambiguous.

I think a similar thing would work for the `struct index_header_offsets`s too, and then I think the code that switches between them would look a bit nicer then too, like: `if (i.minor_version < 19) offsets = &pre19_offsets;`, or however it went.

Alternatively, we could not worry about it now, and figure it out when we need to.

elliefm (commented):
Couple of inline comments about naming of the offset stuff.

I'm not sure about the mailbox changes specifically -- they look okay to me, but I'm not deeply familiar with this corner of the code, so if something's subtly wrong or missing I'm unlikely to catch it.  I'll be more confident when @brong's looked at it too.

Everything else looks good, and tests are all passing, so that's an approve from me otherwise.

imap/mailbox.c
ksmurchison: Yeah, good thought.

imap/mailbox.h
ksmurchison: Yep, I had the same idea as I was reading your other comment.  Will rename

elliefm (approved):
Yeah, that looks nice!

Approved from me, but you might still want Bron to cast an eye over it too

brong (approved):
This looks fine.  I still would like to have the "happy path" be more optimised code for the parsing and creating index records just because it happens so much, especially parsing!  But it's not a big deal.

brong (approved):

PR #4388, opened 2023-01-25
cyr_buildinfo: add version category with some version details

elliefm:
Add some version details to the data output by cyr_buildinfo

Build wrappers might use this to detect unexpected version changes

### DIFF
--- a/imap/cyr_buildinfo.c
+++ b/imap/cyr_buildinfo.c
@@ -52,11 +52,20 @@
 #include <sysexits.h>
 #include <jansson.h>
 
-#include "global.h"
-#include "proc.h"
-#include "util.h"
-#include "../master/masterconf.h"
-#include "xmalloc.h"
+#include "lib/util.h"
+#include "lib/xmalloc.h"
+
+#include "imap/conversations.h"
+#include "imap/global.h"
+#include "imap/mailbox.h"
+#include "imap/proc.h"
+#include "imap/statuscache.h"
+#include "imap/zoneinfo_db.h"
+
+#include "master/masterconf.h"
+
+#include "sieve/bytecode.h"
+#include "sieve/sieve_interface.h"
 
 /* Make ld happy */
 const char *MASTER_CONFIG_FILENAME = DEFAULT_MASTER_CONFIG_FILENAME;
@@ -84,12 +93,14 @@ static json_t *buildinfo()
     json_t *search = json_object();
     json_t *hardware = json_object();
     json_t *buildconf = json_object();
+    json_t *version = json_object();
 
     json_object_set_new(buildconf, "component", component);
     json_object_set_new(buildconf, "dependency", dependency);
     json_object_set_new(buildconf, "database", database);
     json_object_set_new(buildconf, "search", search);
     json_object_set_new(buildconf, "hardware", hardware);
+    json_object_set_new(buildconf, "version", version);
 
     /* Yikes... */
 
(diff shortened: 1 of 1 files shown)


rjbs:
@elliefm I like the idea, as we discussed.  My C is lousy.  I think I get the gist, but would you be so kind as to show the actual output from this (and maybe its invocation) on a test build?

elliefm:
@rjbs So, `cyr_buildinfo(8)` spits out JSON data containing various pieces of build-time information.  I'll show the full output at the end, but I'll front load the interesting bits.

This PR adds a "misc" category to the output, containing some version fields:

```
$/dev/shm/cyrus/main/sbin/cyr_buildinfo | jq .misc
{
  "CYRUS_VERSION": "3.9.0-alpha0-59-g3dda4713bf",
  "MAILBOX_MINOR_VERSION": 17,
  "MAILBOX_CACHE_MINOR_VERSION": 11
}
```

(Where "/dev/shm/cyrus/main" was the --prefix I configure'd my build with.)

In the specific case we were discussing, we were interested in MAILBOX_MINOR_VERSION specifically.  I added the other two while I was in there anyway, since we might want to monitor these for unexpected changes too.

```
$/dev/shm/cyrus/main/sbin/cyr_buildinfo | jq .misc.MAILBOX_MINOR_VERSION
17
```

Note that this tool is specific to the _build_, so it has no use for the usual `-C /path/to/imapd.conf` option -- though it will accept and ignore it if it's provided.

Full output looks like this:

```
$/dev/shm/cyrus/main/sbin/cyr_buildinfo
{
  "component": {
    "event_notification": true,
    "gssapi": false,
    "autocreate": true,
    "idled": true,
    "httpd": true,
    "kerberos_v4": false,
    "murder": true,
    "nntpd": true,
    "replication": true,
    "sieve": true,
    "calalarmd": true,
    "jmap": true,
    "objectstore": false,
    "backup": true
  },
  "dependency": {
    "ldap": true,
    "openssl": true,
    "zlib": true,
    "jansson": true,
    "pcre": true,
    "clamav": true,
    "openio": false,
    "nghttp2": true,
    "wslay": true,
    "brotli": true,
    "xml2": true,
    "ical": true,
    "icu4c": true,
    "shapelib": false,
    "chardet": true,
    "cld2": true,
    "guesstz": true
  },
  "database": {
    "mysql": false,
    "pgsql": false,
    "sqlite": true
  },
  "search": {
    "squat": true,
    "xapian": true,
    "xapian_cjk_tokens": "words"
  },
  "hardware": {},
  "misc": {
    "CYRUS_VERSION": "3.9.0-alpha0-59-g3dda4713bf",
    "MAILBOX_MINOR_VERSION": 17,
    "MAILBOX_CACHE_MINOR_VERSION": 11
  }
}
```

elliefm:
I think I want to add STATUSCACHE_VERSION (from imap/statuscache.h) to this too

ksmurchison (commented):
This looks sane to me.  Adding STATUSCACHE_VERSION also makes sense

elliefm:
I've updated this to include all the `#define FOO_VERSION` defined in header files that can be included without additional dependencies.

There's a bunch more that might be useful to include.  I skipped the ones that are only defined in C files, and the ones that are defined in header files that themselves need other header files.  If it turns out we find this sort of thing useful, then we can do some refactoring to make those version numbers available too, but I didn't feel like burning the time pre-emptively.

I've renamed the category from "misc" to "version", since this is clearly now a dedicated category for version information, not a dumping ground for miscellany.

ksmurchison (approved):
LGTM

rjbs (approved):
I see I never replied!  The provided output look fit for the purpose we talked about! 👍🏽

PR #4389, opened 2023-01-25
http_dav.c: don't need 'x'/'k' right on cal/abook to advertise [un]bind

ksmurchison:
This fixes a bug where some clients wouldn't allow a sharee to delete events if the sharee didn't also have the right to delete the calendar

### DIFF
--- a/imap/http_dav.c
+++ b/imap/http_dav.c
@@ -349,7 +349,8 @@ enum {
     PRIV_IMPLICIT =             (1<<0),
     PRIV_INBOX =                (1<<1),
     PRIV_OUTBOX =               (1<<2),
-    PRIV_CONTAINED =            (1<<3)
+    PRIV_CONTAINED =            (1<<3),
+    PRIV_NOSUBCOL =             (1<<4),
 };
 
 
@@ -2568,7 +2569,8 @@ static void add_privs(int rights, unsigned flags,
             }
 
             /* DAV:bind */
-            if ((rights & DACL_BIND) == DACL_BIND) {
+            if ((rights & DACL_BIND) == DACL_BIND ||
+                ((flags & PRIV_NOSUBCOL) && (rights & DACL_ADDRSRC))) {
                 priv = xmlNewChild(parent, NULL, BAD_CAST "privilege", NULL);
                 xmlNewChild(priv, NULL, BAD_CAST "bind", NULL);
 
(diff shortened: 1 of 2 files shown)


imap/http_dav.c
```diff
@@ -2724,8 +2727,14 @@ int propfind_curprivset(const xmlChar *name, xmlNsPtr ns,
...
                 else if (!strcmp(fctx->req_tgt->collection, SCHED_OUTBOX))
                     flags = PRIV_OUTBOX;
+
+                flags |= PRIV_NOSUBCOL;
             }
         }
+        else if (fctx->req_tgt->collection &&
+                 fctx->req_tgt->namespace->id == URL_NS_ADDRESSBOOK) {
+            flags |= PRIV_NOSUBCOL;
+        }
```
elliefm, line 2737: It's not touched as part of this PR so I can't comment on it directly, but just below this block is:

```
flags += PRIV_CONTAINED;
```

Note that it's `+=` rather than `|=`... I think that could have bad effects if the PRIV_CONTAINED bit was already set?  Looks like it can't be already set at the moment, because `flags` was initialised to 0 at the top of the function and this is the first time PRIV_CONTAINED is mentioned.  But if we ever need to do something with PRIV_CONTAINED earlier, this line could bite us

elliefm (approved):
Looks okay, with the caveat that I don't fully understand calendar ACLs/privileges, so if there's something subtly wrong I might not see it.

One future tripping hazard identified, though it's not actually part of this PR, just nearby

imap/http_dav.c
ksmurchison: Good catch.  Fixed.

ksmurchison:
@brong Do you want to give this a look as well?

imap/http_dav.c
```diff
@@ -2724,10 +2727,16 @@ int propfind_curprivset(const xmlChar *name, xmlNsPtr ns,
...
 
-        flags += PRIV_CONTAINED;
+        flags |= PRIV_CONTAINED;
```
brong, line 2739: Nice that this just works unless you accidentally do it twice.  Ouch!  And even then it still works just fine so long as you don't have a bigger key :p

brong (approved):
This all looks fine - though I'd like to see a test case for it.

ksmurchison:
Added Caldav.dav_bind

PR #4391, opened 2023-01-26
Rename.pm: test renaming a user who has had mailboxes shared to them

ksmurchison:
This test confirms that we rewrite ACLs of shared mailboxes when renaming a user (when reverseacls are enabled)

### DIFF
--- a/cassandane/Cassandane/Cyrus/Rename.pm
+++ b/cassandane/Cassandane/Cyrus/Rename.pm
@@ -924,4 +924,53 @@ sub test_intermediate_cleanup
     _match_intermediates($self, %set);
 }
 
+sub test_rename_user_sharee
+    :AllowMoves :NoAltNameSpace :ReverseACLs :min_version_3_6
+{
+    my ($self) = @_;
+
+    my $admintalk = $self->{adminstore}->get_client();
+
+    xlog $self, "Test user rename with shares";
+
+    my %exp;
+
+    $admintalk->create("user.foo") || die;
+    $admintalk->create("user.foo.shared") || die;
+    $admintalk->setacl("user.foo.shared", 'cassandane' => 'lrs') || die;
+
+    $admintalk->create("user.foo.#calendars") || die;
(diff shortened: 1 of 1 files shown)


cassandane/Cassandane/Cyrus/Rename.pm
```diff
@@ -924,4 +924,53 @@ sub test_intermediate_cleanup
...
+        'user.foo.shared'  => [ '\\Subscribed', '\\HasNoChildren' ],
+    };
+
+
+    my $list = $imaptalk->list([qw( vendor.cmu-dav )], '', '*', 'return', ['subscribed']);
+
+    $self->assert_mailbox_structure($list, '.', $structure);
+
+    my $res = $admintalk->rename('user.cassandane', 'user.newuser');
+    $self->assert(not $admintalk->get_last_error());
```
elliefm, line 965: Plain boolean `assert()` kinda sucks, because if it fails the default message is low information.

I think I usually use something like `$self->assert_str_equals('ok', $admintalk->get_last_completion_response())` in cases like this, which is long-winded but produces a better message when it fails

elliefm (approved):
Looks good; one nit about use of plain `assert()`

cassandane/Cassandane/Cyrus/Rename.pm
ksmurchison: Replaced.  Thanks.

elliefm:
This test actually fails when backported to 3.6... so I guess we fixed something between 3.6 and 3.8, which this proves the fix for.  Need to track down what that was, so I can backport it too

elliefm:
The oldest 3.7 commit I can still easily build is 4e27c5ef0, and that commit already passes this test when it's cherry-picked in, so I can't bisect to find the fix.  At least, not without figuring out why the older commits don't build, which at a guess is probably that they require an older libical than the one I have.

Leaving this tagged to backport for now, but I might just give up remove the tag next time.

Issue #4328, opened 2022-11-15
index_finduid returns the previous record if the UID isn't exactly there


brong:
There are three cases of callees:

1) want an exact match.  They need to remember to `if (index_getuid(index, msgno) == uid) {}`

2) want to start a range - they need to remember to do `if (!msgno || index_getuid(index, msgno) != uid) msgno++;`

3) want to end a range - the value is always correct for that.

This also applies to `mailbox_finduid`.

We should create separate APIs for each of the three cases above and have callers use them as appropriate, rather than open coding the post-call checks in every caller.

PR #4393, opened 2023-01-27
index_finduid(): new mode param determines return value for nonexistent UID

ksmurchison:
Fix for #4328

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -1125,11 +1125,11 @@ EXPORTED void index_fetchresponses(struct index_state *state,
 
             if (usinguid) {
                 if (first > 1)
-                    start = index_finduid(state, first);
+                    start = index_finduid(state, first, FIND_LE);
                 if (first == last)
                     end = start;
                 else if (last < state->last_uid)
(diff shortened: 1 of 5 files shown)


imap/index.c
```diff
@@ -3391,6 +3392,9 @@ EXPORTED uint32_t index_finduid(struct index_state *state, uint32_t uid)
         else
             low = mid + 1;
     }
+
+    if (mode == FIND_EQ) return 0;
+    if (mode == FIND_GE) return high+1;
```
brong, line 3397: Can this mode lead to an off-by-one overflow?

brong (approved):
This looks good - my only issue is that maybe it can FIND_GE can point AFTER the final record in theory. But I think that's fine in all the callers.

imap/index.c
ksmurchison: The only place that FIND_GE is used is in nntpd, and its handled properly there.  If we returned state->exists rather than state->exists+1, nntpd would incorrectly access the last message if a UID greater than actually exists is requested.

PR #4394, opened 2023-01-30
MurderIMAP.pm: skip IMAP REPLACE tests if frontend version < 3.9

ksmurchison:


elliefm (approved):
Oh -- I'd misremembered how `get_version()` for non-default installs works, but you figured it out, great. :)

PR #4395, opened 2023-01-31
Fix iter-startuid off by one issue

brong:
This was completely breaking indexing mailboxes that didn't have UID 1 in them any more, because iter_startuid being called with `1` led to setting iter->recno to zero, and it mailbox_iter_step would return NULL immediately!

Also, if the called UID wasn't in the mailbox, it would replay the previous message, which isn't what we ever want.

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -3367,8 +3367,10 @@ EXPORTED int index_copy_remote(struct index_state *state, char *sequence,
 /*
  * Returns the msgno of the message with UID corresponding to 'uid', based on 'mode'
  * If mode is EQ, msgno with             UID == 'uid', otherwise 0
- * If mode is GE, msgno with the lowest  UID >= 'uid'
- * If mode is LE, msgno with the highest UID <= 'uid'
+ * If mode is GE, msgno with the lowest  UID >= 'uid', which is never zero,
+ *                                                     but may be past the end
+ * If mode is LE, msgno with the highest UID <= 'uid', which may be zero,
+ *                                                     but never past the end
  */
 EXPORTED uint32_t index_finduid(struct index_state *state, uint32_t uid, int mode)
 {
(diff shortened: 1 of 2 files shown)


rsto (approved):
Ouch! LGTM

ksmurchison (approved):
LGTM

brong:
@elliefm I haven't tagged for earlier but it may be worth backporting to earlier.  And my apologies, the commit to index.c won't backport, and is not needed - it's just a comment on a new API to detail that it has the same edge condition.

elliefm:
@brong Do you know how far back the bug goes?  I usually use a combination of `git blame`, `git branch --contains`, and `git log` to find which commit introduced the bug, and which branches have that commit id (or a commit with the same message, i.e. if the bug itself was cherry-picked/backported)

I can and will do all this myself, but since you already know the bug, you'd be one step ahead of me if you did it...

elliefm:
mailbox_iter_startuid was created in b59ec23a1 (before 3.0), and modified by 724a576c5 (before 3.8)

So depending on which of those introduced the bug, this either affects all 3.0+ versions, or only 3.8 and master

I've just backported it to 3.8 because I'm in the middle of a batch of 3.8 backports anyway, but I'll look at the others some other time

elliefm:
When backported to 3.6, this causes a couple of SearchFuzzy tests to fail.  Not sure if it's because this patch isn't suitable for 3.6, or if it's exposing some other bug whose fix also needs to be backported.

elliefm:
Trying to backport this to 3.6 again, and in addition to SearchFuzzy, it also causes some JMAPTestSuite tests to fail.  There's definitely something wrong that this fix is exposing, but I don't know what yet.  I thought maybe 724a576c5 ("uid-1") contained the answer, but actually it was a one-line change that's completely subsumed by this PR anyway.

Here's the list of failures, in case it's ever useful:

```
[FAILED] Cyrus::JMAPTestSuite.t:Email:query:filtering
[FAILED] Cyrus::JMAPTestSuiteWS.t:Email:query:filtering
[FAILED] Cyrus::SearchFuzzy.dedup_part_compact
[FAILED] Cyrus::SearchFuzzy.dedup_part_index
[FAILED] Cyrus::JMAPTestSuite.t:Email:queryChanges:simple
[FAILED] Cyrus::JMAPTestSuiteWS.t:Email:queryChanges:simple
```

PR #4396, opened 2023-01-31
Html tag colon

brong:
The HTML parser wasn't handling : in tag names, which is legit because you can have namespaces in Outlook-generated emails.  This was found by a user who had an email very much like the test case I added as part of this.  Confirmed fixed by this test.

### DIFF
--- a/cassandane/Cassandane/Cyrus/SearchFuzzy.pm
+++ b/cassandane/Cassandane/Cyrus/SearchFuzzy.pm
@@ -1464,7 +1464,7 @@ sub test_html_only
         mime_type => "text/html",
         body => ""
           . "<html xmlns:o=\"urn:schemas-microsoft-com:office:office\">\r\n"
-          . "<div>This is an html <o:p>LL123</o:p> body.</div>\r\n"
+          . "<div>This is an html <o:p>LL123</o:p> <h11>xyzzy</h11> body.</div>\r\n"
           . "</html"
 
     ) || die;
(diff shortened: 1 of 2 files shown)


cassandane/Cassandane/Cyrus/SearchFuzzy.pm
```diff
@@ -1453,6 +1453,45 @@ sub test_striphtml_alternative
...
+    xlog "Assert that HTML in plain text is stripped";
+    my $uids = $talk->search('fuzzy', 'body', 'html') || die;
+    $self->assert_deep_equals([1], $uids);
+
+    $uids = $talk->search('fuzzy', 'body', 'div') || die;
+    $self->assert_deep_equals([], $uids);
+
+    # make sure the "p" doesn't leak into a token
+    $uids = $talk->search('fuzzy', 'body', 'LL123p') || die;
+    $self->assert_deep_equals([], $uids);
```
elliefm, line 1482: Do we need a negative check for 'pLL123p' too?  Since it would have choked on the opening tag too...

I assume this check is based on what you saw happening in practice, that the dregs from the opening tag probably wind up as their own word, since there's a space before the opening tag, and it's only the dregs from the closing tag that get wrongly concatenated onto the previous word -- in which case it's fine as it is.

cassandane/Cassandane/Cyrus/SearchFuzzy.pm
```diff
@@ -1453,6 +1453,45 @@ sub test_striphtml_alternative
...
+
+    # make sure the real token gets indexed
+    $uids = $talk->search('fuzzy', 'body', 'LL123') || die;
+    $self->assert_deep_equals([1], $uids);
+
+    # make sure the h11 doesn't leak
+    $uids = $talk->search('fuzzy', 'body', 'xyzzy1') || die;
+    $self->assert_deep_equals([], $uids);
```
elliefm, line 1490: Same question as above for "1xyzzy1"

elliefm (approved):
Looks good.

Couple of places that may or may not want additional negative checks, but I don't think I need to re-review if you add them.

brong:
Answered this in slack as well - I think there's a '>' in between still, so it would  be indexing 'p>LL123p>' which becomes `p` and `LL123` terms.  This is consistent with the original record for the user reporting the bug, which had `p` indexed as well.

brong:
I probably wouldn't bother backporting further, but this feels like a candidate for 3.8

PR #4397, opened 2023-02-01
JMAP Calendars: always guess IANA timezone id

rsto:
JSCalendar RFC8984 allows custom timezones to be set in the the "CalendarEvent.timeZones" property. This property is only going to be allowed for JSCalendar Group objects in jscalendarbis. JMAP Calendars clients have no way to fetch that custom timezone definition in lack of the Group object and consequently can't deal with custom timezone identifiers in an event "timeZone" property.

This set of patch removes the "timeZones" property from events returned in JMAP Calendars and makes Cyrus always attempt an IANA timezone identifier for a non-IANA timezone. The custom VTIMEZONE component is preserved in the iCalendar data.


ksmurchison (approved):
LGTM

brong:
Just rebased it to merge

PR #4400, opened 2023-02-02
jmap_ical.c: don't add RECURRENCE-ID property for recurrenceId:null

ksmurchison:
This fixes a bug where JSCalendar event id would end up being something like id:"ER-00000000T000000-abcd"


brong (approved):
Yep this looks sensible - I sure did have to scroll back a LONG way to see where this gets set, but it looks right to me.

PR #4401, opened 2023-02-05
Jmap move deleted parent

brong:
Just tagging all for whoever feels like reviewing :)  This was triggered by a user breaking replication horribly by the faulty "rename under the DELETED namespace".  We should maybe actually plumb a "kill this with fire" check down a level too in mboxlist_renamemailbox(), but this fix certainly seems right anyway.

My only issue would be whether it somehow breaks "Mailbox/changes", but the other existing tests seem to indicate it's OK!

### DIFF
--- a/cassandane/Cassandane/Cyrus/JMAPMailbox.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPMailbox.pm
@@ -1375,6 +1375,79 @@ sub test_mailbox_set_order
     $self->assert_null($res->[0][1]{notDestroyed});
 }
 
+sub test_mailbox_move_to_deleted_parent
+    :min_version_3_8 :needs_component_jmap
+{
+    my ($self) = @_;
+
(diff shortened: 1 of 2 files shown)


cassandane/Cassandane/Cyrus/JMAPMailbox.pm
```diff
@@ -1375,6 +1375,79 @@ sub test_mailbox_set_order
     $self->assert_null($res->[0][1]{notDestroyed});
 }
 
+sub test_mailbox_move_to_deleted_parent
+    :min_version_3_8 :needs_component_jmap
```
elliefm, line 1379: I think 3.6 will want a backport of this too, so can you make the test `:min_version_3_6` please?

```suggestion
    :min_version_3_6 :needs_component_jmap
```

brong: I've done this.  Thanks ;)

PR #4402, opened 2023-02-07
Append fixes

brong:
As mentioned in this morning's call, here's three sanity-check fixes:

1) use a distinct filename per message for all appends

2) check the munmap call result

3) check the GUID of the found file and link it directly if found rather than overwriting the existing spool location

### DIFF
--- a/imap/append.c
+++ b/imap/append.c
@@ -54,6 +54,7 @@
 #include <sys/un.h>
 #include <sys/wait.h>
 #include <sys/poll.h>
+#include <inttypes.h>
 
 #include "acl.h"
 #include "assert.h"
@@ -94,7 +95,7 @@ struct stagemsg {
     struct message_guid guid;
 };
 
-uint64_t append_counter;
+static uint64_t append_counter;
 
 static int append_addseen(struct mailbox *mailbox, const char *userid,
                           seqset_t *newseen);
@@ -333,8 +334,8 @@ EXPORTED FILE *append_newstage_full(const char *mailboxname, time_t internaldate
     stage = xmalloc(sizeof(struct stagemsg));
     strarray_init(&stage->parts);
 
-    snprintf(stage->fname, sizeof(stage->fname), "%d-%llu",
-             (int) getpid(), (long long unsigned)append_counter++);
+    snprintf(stage->fname, sizeof(stage->fname), "%d-%" PRIu64,
+             (int) getpid(), append_counter++);
 
     r = mboxlist_findstage(mailboxname, stagedir, sizeof(stagedir));
     if (r) {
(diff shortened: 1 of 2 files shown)


imap/append.c
```diff
@@ -1063,7 +1079,7 @@ EXPORTED int append_fromstage_full(struct appendstate *as, struct body **body,
     r = msgrecord_get_fname(msgrec, &fname);
     if (r) goto out;
 
-    r = mailbox_copyfile(stagefile, fname, nolink);
+    r = mailbox_copyfile(linkfile ? linkfile : stagefile, fname, nolink);
     if (r) goto out;
```
ksmurchison, line 1083: Unless I'm missing something, it looks like we are leaking `linkfile`.

ksmurchison (changes requested):
Need to plug the linkfile leak

ksmurchison (changes requested):
Need to plug the `linkfile` leak.  Otherwise, looks good.

imap/append.c
```diff
@@ -94,6 +94,8 @@ struct stagemsg {
     struct message_guid guid;
 };
 
+uint64_t append_counter;
```
elliefm, line 97: This wants to be `static`.  Among other things, that will guarantee it's initialised to zero when left uninitialised like this...

imap/append.c
```diff
@@ -331,8 +333,8 @@ EXPORTED FILE *append_newstage_full(const char *mailboxname, time_t internaldate
     stage = xmalloc(sizeof(struct stagemsg));
     strarray_init(&stage->parts);
 
-    snprintf(stage->fname, sizeof(stage->fname), "%d-%d-%d",
-             (int) getpid(), (int) internaldate, msgnum);
+    snprintf(stage->fname, sizeof(stage->fname), "%d-%llu",
+             (int) getpid(), (long long unsigned)append_counter++);
```
elliefm, line 337: Add `#include <inttypes.h>`, and then use the proper uint64_t formatter, instead of the longwinded cast

```suggestion
    snprintf(stage->fname, sizeof(stage->fname), "%d-%" PRIu64,
             (int) getpid(), append_counter++);
```

(Check this before integrating! I'm typing in a github window, not my editor...)

brong:
I've fixed both those - thanks for the reviews!

imap/append.c
brong: Fixed, ta

imap/append.c
brong: Yep, done.

imap/append.c
brong: I've added a free()

elliefm (approved):

ksmurchison (approved):

PR #4403, opened 2023-02-10
IMAP IDLE & NOTIFY

ksmurchison:
This rewrites IMAP IDLE, and uses those changes to implement unsolicited LIST responses for IMAP4rev2 and IMAP NOTIFY (RFC 5465).

Its probably best to review commit-by-commit.


elliefm:
This is failing a load of cassandane tests on CI.  Do you expect that, and want me to get started reviewing anyway, or should I hold off until you've had a chance to sort it out?

ksmurchison:
Hold off. I haven't run Cass yet. It's probably something stupid or a bad test

elliefm (commented):
Did a partial read of this, found one nit but generally looks okay so far.  Stepping away cause I'm losing focus, need to come back to it again later.

imap/idled.c
```diff
@@ -87,170 +114,211 @@ EXPORTED void fatal(const char *msg, int err)
...
+    json_t *msg;
+    arrayu64_t *failed_pids;
+};
+
+static int notify_cb(sqlite3_stmt *stmt, void *rock)
 {
-    struct ientry *t, *p = NULL;
+    struct notify_rock *nrock = rock;
+    const char *mboxid =
+        json_string_value(json_object_get(nrock->msg, "mailboxID"));
```
brong, line 139: we should probably make this all lowercase - that's what JSON API users are used to, and maybe there will be pluggable IDLE engines

brong (changes requested):
I think we should lowercase `mailboxid` over the wire.  Otherwise looks great.

lib/sqldb.c
```diff
@@ -371,6 +371,10 @@ static void buf_replace_bindvals(struct buf *cmd, struct sqldb_bindval bval[])
             else
                 buf_setcstr(&buf, "NULL");
             break;
+
+        case SQLITE_BLOB:
+            buf_printf(&buf, "<%lu bytes>", buf_len(&bval->val.b));
```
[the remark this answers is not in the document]
ksmurchison: Yep.  Done

imap/idled.c
brong: Legit

brong (approved):
Yeah, I approve of this

PR #4405, opened 2023-02-13
mailbox offset alignment tests

elliefm:
This adds a new test suite, mailbox.testc, containing tests that assert that the various OFFSET_FOO for `struct index_header` and `struct index_record` fields are aligned correctly according to the requirements of the type used to read/write into the file.

This was inspired by someone observing that some `modseq_t` and `quota_t` fields weren't correctly aligned to 8 byte boundaries in version 17 and 18 indexes.  When this branch is based on `master` today, it correctly detects these misaligned fields.  Those fields are already fixed in index version 19, which will land with #4387, so this PR will remain a draft until that lands, at which point I'll rebase and update this for version 19 indexes. [Edit: this has now been done.]

The tests depends on the C11 `alignof` macro (from `stdalign.h`), and are conditionally defined so that if this feature isn't available, they'll still compile, but do nothing (and pass).  Our unit testing infrastructure has no way to mark tests as skipped, nor can individual tests be compiled out, so this will have to do.

Reviewers, please check carefully that the OFFSETs in the tests match up with the correct struct fields.  I think I got it right, but the eyes glaze over while reading it closely.  If you have clang available, please give that a shot too; I've only tested it with gcc.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -635,6 +635,7 @@ cunit_TESTS = \
 	cunit/imapurl.testc \
 	cunit/imparse.testc \
 	cunit/libconfig.testc \
+	cunit/mailbox.testc \
 	cunit/mboxname.testc \
 	cunit/md5.testc \
 	cunit/message.testc \
--- a/configure.ac
+++ b/configure.ac
@@ -202,7 +202,7 @@ AC_CHECK_ALIGNOF(uint32_t)
 dnl check for -R, etc. switch
 CMU_GUESS_RUNPATH_SWITCH
 
-AC_CHECK_HEADERS(unistd.h sys/select.h sys/param.h stdarg.h)
+AC_CHECK_HEADERS(unistd.h sys/select.h sys/param.h stdalign.h stdarg.h)
 AC_REPLACE_FUNCS(memmove strcasecmp ftruncate strerror posix_fadvise strsep memmem memrchr)
 AC_CHECK_FUNCS(strlcat strlcpy strnchr getgrouplist fmemopen pselect futimens futimes getline)
 AC_HEADER_DIRENT
--- a/configure.ac
+++ b/configure.ac
@@ -2230,6 +2230,44 @@ AS_IF(
         [Do we have support for optimize __attribute__?])]
 )
 
+AC_CACHE_CHECK(
+    [whether the C compiler supports alignof(expression)],
+    [cyrus_cv_gnu_alignof_expression],
+    [
+        AC_LANG_PUSH([C])
+        saved_CFLAGS="$CFLAGS"
+        saved_CPPFLAGS="$CPPFLAGS"
+        CFLAGS="$CFLAGS -Wall -Werror -Wno-unused"
+        CPPFLAGS="$CPPFLAGS -Wall -Werror -Wno-unused"
+        AC_COMPILE_IFELSE(
+            [AC_LANG_PROGRAM(
+                [
+                    #include <stdalign.h>
+                    #include <stddef.h>
+                ],
+                [
+                    int x;
+                    struct { float y; } y;
+                    size_t a, b;
+
+                    a = alignof(x);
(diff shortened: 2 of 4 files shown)


elliefm:
Rebased and updated for v18 and v19

rsto:
I just compiled this with clang 

    Debian clang version 11.0.1-2
    Target: aarch64-unknown-linux-gnu

and got

    ./cunit/mailbox.testc:60:51: error: '_Alignof' applied to an expression is a GNU extension [-Werror,-Wgnu-alignof-expression]
    CU_ASSERT_EQUAL(0, OFFSET_ANSWERED          % alignof(h.answered));

ksmurchison (dismissed):
Works for me, but I'm using GCC 12.2.1

I can get clang 15.0.7 to work if I remove -Werror and add  -Wgnu-alignof-expression to CPPFLAGS

rsto (approved):
Checked offsets and looks good to me.

rsto:
I recompiled successfully with clang `-Wgnu-alignof-expression ` now, too.

elliefm:
I'm confused about how adding `-Wgnu-alignof-expression` helps, it seems like that would enable the warning, which was presumably already enabled.  Do you mean -W**no**-gnu-alignof-expression, perhaps?

I hadn't realised that only `alignof(type)` is standard, and `alignof(expression)` is a GNU extension.  But the whole point of using it on expressions here is so that the tests are always testing the actual struct fields.  If I change it to their types, then there'd be a mess of hardcoded types in the test which we would forget to update when the structure changes, and the tests would continue to pass without saying anything meaningful about whether the new changes are correct, which would be useless.

I guess I could add another configure check for whether `alignof(expression)` is supported, and make these tests conditional on that too, like how they're already conditional on stdalign.h existing.  Not sure if it's worth the effort though, what do you think?

elliefm:
@ksmurchison @rsto any response to my last comment?  Do you want me to add a configure check to shush or work around the clang warning while I'm in here anyway, or is it not worth the trouble and I should merge this as-is?

rsto:
@elliefm I'd prefer for configure to handle that, rather than clang builds aborting for people that don't know of `-Wgnu-alignof-expression`. Could we update autoconf to add that option to CLANG/CXXLANG if required?

ksmurchison:
Adding an autoconf option is a good idea

elliefm:
I'm struggling to find any documentation of whether clang "supports this extension, but warns about it", or "doesn't support this extension, and warns about it".  The only clang documentation I've found that even mentions it only states that the warning is enabled by default (it doesn't even explain what it's warning about... I guess you're supposed to just know?).

I think I'm going to make a check for whether `alignof(expression)` compiles okay, and if it doesn't, then the tests that use that behaviour won't be compiled.  That'll solve the general case of "this is a gnu extension that might not be available", and if it turns out clang does provide it, we can make some clang-specific provision in the future if it seems important enough.

elliefm:
I've rebased this on current master, and added two new commits -- a configure.ac check that detects whether the GNU `alignof(expression)` extension is available, and an update to mailbox.testc that makes the tests that require the extension do nothing if it's missing

The "yes" case works for me with GCC, and the "no" case also works if I deliberately break the configure.ac check to misdetect.

When your CC is clang, I expect configure should be like `checking whether the C compiler supports alignof(expression)... no`, and after building everything, if you run the cunit mailbox tests verbosely (`make -j4 check-mailbox` will do it), you should see the `no alignof(expression), can't do anything useful` message in the output

If you arrange locally for clang to suppress the extension warning, with something like `-Wno-gnu-alignof-expression` in your CFLAGS, then I think that might suppress the error during the AC_COMPILE_IFELSE, and allow configure to detect that the feature is supported, and then compile it in.  I don't know if clang actually supports the extension, so if you do this and it breaks you can keep both parts, but the option might be there.

ksmurchison (approved):

PR #4406, opened 2023-02-14
search_engines: do not return search builder for null mailbox

rsto:
The search builder API requires a mailbox for initialization but did not do a proper NULL check of the argument. This could cause the Xapian and Squat search builders to crash.


ksmurchison (approved):
Simple enough

PR #4407, opened 2023-02-15
JMAP Calendars: sanitize geo: coordinates when read from iCalendar

rsto:
Bogus clients may TEXT-escape geo: URI values in properties such
    as X-APPLE-STRUCTURED-LOCATION. Until now Cyrus returned these
    bogus values in the Location "coordinates" property. This resulted
    in Cyrus rejecting an update on such calendar events, because
    the geo: uri was validated during write.

    This patch changes Cyrus to sanitize such bogus geo: uri values
    and return the sanitized value as Location coordinates. It still
    rejects such values on write.


ksmurchison (approved):
LGTM

PR #4408, opened 2023-02-16
index: keep angle-bracketed URLs in plain text bodies

rsto:
This fixes a regression introduced in 0790eaf65b75f74fbeafc5a883fd401a458d0206

Search indexing attempts to strip HTML tags also from plain text bodies if they are part of a multipart/alternative. This is because ill-behaving email implementations tend to put HTML in these, too.

Unfortunately, legit email clients might enclose URLs in angle-brackets when they convert anchor hrefs in HTML bodies to plain text alternatives. This caused Cyrus to also strip such URLs from plain text, instead of indexing them for text search.


ksmurchison (approved):
LGTM

PR #4410, opened 2023-02-17
Add test showing query cache isn't invalidated by squatter

wolfsage:
That is... if you manage to query for something before squatter has a chance to process the latest changes to your mailbox, that query will be cached and you won't see the results until the cache is invalidated (by the next mailbox stage change or by the cache record expiring).

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email-query-cached-negative
+++ b/cassandane/tiny-tests/JMAPEmail/email-query-cached-negative
@@ -1,0 +1,101 @@
+#!perl
+use Cassandane::Tiny;
+
+# If an email arrives and we manage to query for it *before* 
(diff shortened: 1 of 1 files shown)


elliefm (approved):
Ah, I see this adds a test that fails, proving the existence of a known bug.  But in the FM tracker, the bug is marked "maybe someday", so we aren't likely to fix it ourselves anytime soon.

The test looks good, but it'll be annoying to have a test that fails all the time.  So I'm  gonna approve this because it's good and correct, but also label it "do not merge" for the time being.  If we or anyone else ever gets around to fixing the bug, the test will be right here waiting, and can be merged then.

wolfsage:
ah right, we can close this then, thanks

PR #4411, opened 2023-02-20
notify_log: use struct buf

elliefm:
fixes theoretical buffer overrun from too many options

### DIFF
--- a/notifyd/notify_log.c
+++ b/notifyd/notify_log.c
@@ -48,31 +48,29 @@
 #include <string.h>
 #include <stdio.h>
 
+#include "lib/util.h"
+
 char* notify_log(const char *class, const char *priority,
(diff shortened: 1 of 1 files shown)


rsto (approved):
Ack, notify_log doesn't look to be on any hot path so doing the safer heap-allocation here is a good choice.

ksmurchison (approved):

elliefm:
> Ack, notify_log doesn't look to be on any hot path so doing the safer heap-allocation here is a good choice.

Yeah. I thought about making the struct buf static and reseting it each time, so that it only reallocs when it needs a longer buffer, and avoids heap fragmentation from repetitive malloc/free... but then it didn't seem worth it in this use case anyway.  I figure we can do that later if it turns out it's needed.

PR #4412, opened 2023-02-20
fix bare configure build

### LINKED ISSUE #4409: bare ./configure doesn't build
On master,

```
$ git clean -xfd
$ autoreconf -is
$ ./configure
$ make -j4 -k
```

... results in a handful of errors.  Skimming, it looks like the main culprits are jmap_util.h and httpd.h.  It might be a simple matter of adding some judicious `#ifdef`s in the right places...

elliefm:
This re-enables Cyrus builds configured with just `./configure`, mostly by detangling some header mess and adding some judicious `#ifdef`s where necessary.

Also annotates a bunch of Cassandane tests with their dependencies so that they won't try to run if the feature they need wasn't compiled in.  I did this in the simplest way possible because my goal right now is something backportable to 3.8, but I don't really like it.  (For example, lots of tests needed to be individually annotated ":needs_component_httpd", even though that test doesn't need it, just because the suite itself starts the httpd service.  In the future it would be nice to handle this better, but that's a non-trivial refactor.)

Fixes #4409

**Reviewers:**  please build and test this with both:
* your usual configure arguments, to make sure I haven't broken anything any of us rely on
* a minimal configure (I used `./configure --enable-silent-rules --prefix=/my/cass/prefix`), to make sure the fixes are adequate with your compiler/environment and not just mine

If you run Cassandane on the minimal build, you'll (probably) find a bunch of Foo.relocate_* and RelocateByID.* tests have errors.  I don't know why, but I've put details in #4409.  Since this PR is a prerequisite for even building like that at all, it seems sensible to get the build problem fixed first, and then investigate the relocate failures.


ksmurchison:
@elliefm My normal build still works fine.  

I tried a bare build, but mbpath.c failed to compile due to an unused function complaint.  I fixed this on master.
The next problem is that master.c uses pow() and therefore needs to link with libmath.  We should probably just add this to LIBS and be done with it.

ksmurchison (approved):
This looks good after the pow() compile issue is fixed.

elliefm:
For my future reference: the mbpath fix Ken mentioned was 9034a5949 -- I'll need to grab this when I backport to 3.8

elliefm:
@ksmurchison 

> The next problem is that master.c uses pow() and therefore needs to link with libm. We should probably just add this to LIBS unconditionally and be done with it.

It already links with libm if required, thanks to 4830785d3, using `LT_LIB_M`...

But `LT_LIB_M` fails when compiling with CFLAGS=-Werror, and as a result doesn't detect/set `LIBM='-lm'`.

The HTTPD section handles this differently, with a comment describing a similar problem.  So I guess your minimal build still has CFLAGS=-Werror, and that's why it fails to detect LIBM, and fails to build master.  Whereas my minimal build doesn't have CFLAGS set, because that's usually set by my wrapper script, which I'm not running.

But that means in both of our _usual_ builds, we don't get LIBM, and yet master builds fine in that case.  That's weird.

I have an idea for making this work (save CFLAGS, set `CFLAGS=$CFLAGS -Wno-error`, run LT_LIB_M, restore CFLAGS) which I'll tinker with in a bit.  If that works, we can clean up the HTTPD section too

elliefm:
@ksmurchison: I've rebased on master, added your suggested fixes for the RelocateById and Sieve test failures, and tidied up our libm handling in the way I proposed in my previous comment.  Seems to work properly for me now, in both full builds and minimal, and with and without -Werror.  Can you try it out again please, and confirm whether it's fixed on your system now too?  Especially the pow() bit.  Thanks

ksmurchison:
@elliefm The LIBM hack didn't work for me with -Werror in either config.  If I remove -Werror, its good in both.

ksmurchison (changes requested):
LIBM still not linked in for me with -Werror.  I wish I had an idea how to fix.

ksmurchison (approved):
I approve once we add save/set/restore magic for CPPFLAGS around LT_LIB_M

elliefm:
Ahh yeah, right!

elliefm:
@elliefm 3.6 needs at least 7b8c76f99f23e861450f26d801133ca9fc34a500, but also check the other commits from this PR

PR #4415, opened 2023-02-23
user_sieve_path(): never return "" -- always return a sensible path

ksmurchison:
This should fix an issue where mailbox_open_sieve() via dav_reconstruct would try to write to root

### DIFF
--- a/imap/user.c
+++ b/imap/user.c
@@ -201,8 +201,7 @@ EXPORTED const char *user_sieve_path(const char *inuser)
         int r = mboxlist_lookup(inboxname, &mbentry, NULL);
         free(inboxname);
 
-        if (r) sieve_path = "";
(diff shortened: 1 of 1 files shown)


elliefm:
Do you know which commit introduced the bug, or which version was first affected?  I'll want to backport this

elliefm (approved):

rsto (approved):

ksmurchison:
@elliefm This occurred as part of the UUID mailboxes work:

https://github.com/cyrusimap/cyrus-imapd/commit/4ca939366e8c76e48270fc309bd4d93e702df7ed#diff-09238fdff26df10fe2205827617598a66ea1c9246bbf420962ff439ca13e9271

elliefm:
Thanks, that tells me 3.6 will need the fix too

PR #4416, opened 2023-02-27
Cassandane: Test that we can bind to ports before using them

wolfsage:
This will make tests more resilient when running on servers where other services are using ports cassandane overlaps with.

### DIFF
--- a/cassandane/Cassandane/PortManager.pm
+++ b/cassandane/Cassandane/PortManager.pm
@@ -43,6 +43,8 @@ use warnings;
 
 use lib '.';
 use Cassandane::Cassini;
+use IO::Socket::IP;
+use POSIX qw(EADDRINUSE);
 
 my $base_port;
 my $max_ports = 20;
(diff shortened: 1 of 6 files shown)


cassandane/Cassandane/PortManager.pm
```diff
@@ -74,6 +78,31 @@ sub alloc
...
+    my $host = shift;
+    my $port = shift;
+
+    # If we can bind to the port no one else is currently using it
+    my $socket = IO::Socket::IP->new(
+        LocalAddr => $host,
+        LocalPort => $port,
+        Proto     => 'tcp',
+        ReuseAddr => 1,
+    );
```
elliefm, line 92: Oh, I guess this gets unbound again automatically when `$socket` goes out of scope as `port_is_free()` returns.  I was staring at it for a while wondering how binding the port here wouldn't create the exact problem we were trying to avoid...  Spend a lot of time writing C, forget other languages have destructors. 😅

elliefm (approved):
Nice!

cassandane/Cassandane/PortManager.pm
wolfsage: :)

wolfsage:
Thanks!

PR #4417, opened 2023-02-27
Squash some clang complaints

ksmurchison:


rsto (approved):
Ran OK with `-Wall` on Debian bullseye default clang
```
Debian clang version 11.0.1-2
Target: aarch64-unknown-linux-gnu
```

PR #4420, opened 2023-02-28
Add support for HAProxy protocol

ksmurchison:
https://github.com/haproxy/haproxy/blob/master/doc/proxy-protocol.txt

Currently just plumbed into httpd since it can be tested with curl.

I'm also not sure if -P is the best choice for the command line switch, but don't have any other ideas.

### DIFF
--- a/imtest/imtest.c
+++ b/imtest/imtest.c
@@ -837,12 +837,11 @@ static sasl_security_properties_t *make_secprops(int min,int max)
  * Initialize SASL and set necessary options
  */
 static int init_sasl(char *service, char *serverFQDN, int minssf, int maxssf,
-                     unsigned flags)
+                     unsigned flags, char localip[], char remoteip[])
 {
     int saslresult;
     sasl_security_properties_t *secprops=NULL;
     socklen_t addrsize;
-    char localip[60], remoteip[60];
     struct sockaddr_storage saddr_l;
     struct sockaddr_storage saddr_r;
 
@@ -1433,9 +1432,10 @@ static void interactive(struct protocol_t *protocol, char *filename)
 
                 if (unauth) {
                     /* Reset auth and connection state (other than TLS) */
+                    char localip[60], remoteip[60];
                     sasl_dispose(&conn);
                     if (init_sasl(protocol->service, NULL,
-                                  0, 128, 0) != IMTEST_OK) {
+                                  0, 128, 0, localip, remoteip) != IMTEST_OK) {
                         imtest_fatal("SASL initialization");
                     }
                     unauth = 0;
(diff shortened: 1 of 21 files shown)


elliefm (commented):
That's quite neat, nice.

Maybe we want to also add an implementation of the client side of this to imtest?  I guess it'd need to take a host:port specifying where to pretend the connection was proxied from, and maybe an option for whether to use v1 or v2.

It'd probably help us test the implementation of non-curlable protocols, without needing to stick a real proxy in the middle.  And sysadmins too, if they're having problems with their setup and want to remove the proxy from the equation while still testing the proxied behaviour.

bamthomas:
are there some news about this PR?
Greatly interested by it as we are using SNI termination with HAproxy.

bamthomas:
Sorry to come back on this PR: we have attackers (probably from Russia) that are relentlessly trying to connect on our IMAP server and as there is a front [haproxy](https://www.haproxy.org/) we only have the local IP. So we have no way to use fail2ban and abuseIPDB notifications, or any automated process to get rid of these. It would be of a great help for us.

bamthomas:
thank you very much @ksmurchison

elliefm (approved):
Looks good

I kinda wonder if it's worth putting the header parsing function into its own source+header file.  Since it's more or less copied from the reference implementation, having it in a separate file might make it easier to patch in the future if the reference implementation gets updated.  Though that would be an easy enough change to make then, if that situation arises, so I'm not really worried about it.

ksmurchison:
@elliefm I split the header parsing function into its own source file and I simplified the global.c changes by having haproxy_read_hdr() **only** get the sockaddrs

elliefm (commented):
Hmm, maybe we should leave the CMU copyright boilerplate off the haproxy.[ch] files?  I don't like having to think about copyright stuff much, but claiming copyright on code we mostly copy-and-pasted seems a bit on the nose at best 😅

As far as I can tell the example code in the proxy-protocol.txt file is GPL, since it doesn't say otherwise, and the main haproxy repository license says everything is GPL unless it says otherwise.

Perhaps we could ask them to clarify, and/or leave this PR as a do-not-merge one that people can manually patch in if they want it, but not ship it by default.

elliefm:
> My only outstanding question is whether we should use -H for the services rather than -P. I chose -H for imtest because -P is already being used for specifying the underlying protocol (IMAP, POP3, etc).

Are there any services that _can't_ use `-H`?  I think the services must all be the same as each other (even if imtest is different), and we know `-P` is available.  But if `-H` is available for all of them, making them consistent with imtest sounds good.  I also see you've started calling it "HAProxy" rather than "PROXY", and that's another good reason to change it from `-P` to `-H` if possible, as well as update the "PROXY" references in the man pages.

ksmurchison:
@elliefm I changed all of the services to use -H (including lmtpd, which I had forgotten) and the docs to use HAProxy throughout.

imtest/imtest.c
```diff
@@ -2880,13 +2890,14 @@ int main(int argc, char **argv)
 
     /* keep this in alphabetical order */
     static const char short_options[] =
-        "?I:P:X:a:cf:hik:l:m:n:o:p:qr:st:u:vw:x:z";
+        "?I:P:X:H:a:cf:hik:l:m:n:o:p:qr:st:u:vw:x:z";
```
elliefm, line 2893: The comment above says "keep this in alphabetical order", so `H:` should go between `?` and `I:` 😅

imtest/imtest.c
```diff
@@ -2880,13 +2890,14 @@ int main(int argc, char **argv)
...
 
     static const struct option long_options[] = {
         /* n.b. -? is duplicated as -h */
         { "pidfile", required_argument, NULL, 'I' },
         { "protocol", required_argument, NULL, 'P' },
         /* n.b. -X is duplicated as -x */
+        { "haproxy-clientip", required_argument, NULL, 'H' },
         { "authname", required_argument, NULL, 'a' },
         { "do-challenge", no_argument, NULL, 'c' },
         { "input-filename", required_argument, NULL, 'f' },
```
elliefm, line 2903: When you fix the ordering in short_options, fix it here too.  This struct is a hassle to read if it's not sorted by the short opt.

imtest/imtest.c
ksmurchison: Done

imtest/imtest.c
ksmurchison: Done

elliefm (approved):

PR #4422, opened 2023-03-03
lmtp_sieve.c: strip phrase from redirect address (addr-spec only)

ksmurchison:
Can't use addresses like "Foo <foo@example.com>" in SMTP RCPT TO

This is a simple fix, but how to test?  I don't know how to check for SMTP pass/fail in Cassandane

### DIFF
--- a/cassandane/Cassandane/Cyrus/Sieve.pm
+++ b/cassandane/Cassandane/Cyrus/Sieve.pm
@@ -7133,4 +7133,31 @@ EOF
     $self->check_messages({ 1 => $msg }, check_guid => 0);
 }
 
+sub test_redirect_address_with_phrase
+    :needs_component_sieve
+{
+    my ($self) = @_;
+
+    xlog $self, "Install a script";
+    $self->{instance}->install_sieve_script(<<EOF
+redirect "Foo <foo\@example.com>";
+EOF
+    );
+
+    xlog $self, "Deliver a message";
+    my $msg = $self->{gen}->generate(subject => "Message 1");
+    $self->{instance}->deliver($msg);
+
+    # Verify that message was redirected (no RCPT TO error)
+    if ($self->{instance}->{have_syslog_replacement}) {
+        my @lines = $self->{instance}->getsyslog();
+        $self->assert_does_not_match(qr/RCPT TO: code=553 text=5.1.1/, "@lines");
+    }
+
+    xlog $self, "Make sure that message is NOT in INBOX (due to runtime error)";
+    my $talk = $self->{store}->get_client();
+    $talk->select("INBOX");
+    $self->assert_num_equals(0, $talk->get_response_code('exists'));
+}
+
 1;
--- a/cassandane/Cassandane/Net/SMTPServer.pm
+++ b/cassandane/Cassandane/Net/SMTPServer.pm
@@ -74,7 +74,7 @@ sub rcpt_to {
     $Self->{_rcpt_to_count}++;
     if ($Self->{_rcpt_to_count} > 10) {
         $Self->send_client_resp(550, "5.5.3 Too many recipients");
-    } elsif ($To =~ /\@fail\.to\.deliver$/i) {
+    } elsif ($To =~ /[<>]/ || $To =~ /\@fail\.to\.deliver$/i) {
         $Self->send_client_resp(553, "5.1.1 Bad destination mailbox address");
         $Self->mylog("SMTP: 553 5.1.1");
     } else {
--- a/imap/lmtp_sieve.c
+++ b/imap/lmtp_sieve.c
@@ -785,7 +785,22 @@ static int send_forward(sieve_redirect_context_t *rc,
 #endif
     }
     else {
-        smtp_envelope_add_rcpt(&sm_env, rc->addr);
+        struct address *addr = NULL;
+        char *rcpt = NULL;
+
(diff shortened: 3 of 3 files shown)


imap/lmtp_sieve.c
```diff
@@ -785,7 +785,13 @@ static int send_forward(sieve_redirect_context_t *rc,
...
     }
     else {
-        smtp_envelope_add_rcpt(&sm_env, rc->addr);
+        struct address *addr = NULL;
+        char *rcpt = NULL;
+
+        parseaddr_list(rc->addr, &addr);
+        rcpt = address_get_all(addr, 1);
+        smtp_envelope_add_rcpt(&sm_env, rcpt);
+        free(rcpt);
```
elliefm, line 794: Think you want a `parseaddr_free(addr)` in here too, that `struct address` type looks like a dynamically-allocated linked list

elliefm (commented):
Memory leak maybe?

I'll have a think about testing... I know we've got some kind of SMTP mocking in Cassandane -- have a look at `Cassandane::Net::SMTPServer` and `_start_smtpd()` in `Cassandane::Instance` -- but I'm not sure I've ever used it before (maybe @rsto has?), and I'm not sure how it works or whether it's useful in this case.

imap/lmtp_sieve.c
ksmurchison: Yep.  Good catch!.  FIxed

ksmurchison:
@elliefm I added a check of syslog, which makes sure that RCPT TO doesn't cause an error.  The absence of the error should be good enough to prove the fix works

imap/lmtp_sieve.c
```diff
@@ -785,7 +785,14 @@ static int send_forward(sieve_redirect_context_t *rc,
 #endif
     }
     else {
-        smtp_envelope_add_rcpt(&sm_env, rc->addr);
+        struct address *addr = NULL;
+        char *rcpt = NULL;
+
+        parseaddr_list(rc->addr, &addr);
+        rcpt = address_get_all(addr, 1);
```
rsto, line 792: Is this guaranteed to only ever be called with valid addr-spec values? I know that the redirect rule sieve parser validates the addr-spec, but could this be called from anywhere else where validation is not guaranteed?
If so, then a NULL check here should fix this.

rsto (commented):
I am unsure if a NULL check is required (it never hurts, though!). Otherwise this looks good.

imap/lmtp_sieve.c
ksmurchison: It shouldn't be called from anywhere else.  I'm not really using the parser here to validate the address (as you said, the script parser has already determined that its not junk), I'm just using it to break it down into its component parts to strip the phrase part.

cassandane/Cassandane/Net/SMTPServer.pm
```diff
@@ -74,7 +74,7 @@ sub rcpt_to {
     $Self->{_rcpt_to_count}++;
     if ($Self->{_rcpt_to_count} > 10) {
         $Self->send_client_resp(550, "5.5.3 Too many recipients");
-    } elsif ($To =~ /\@fail\.to\.deliver$/i) {
+    } elsif ($To =~ /[<>]/ || $To =~ /\@fail\.to\.deliver$/i) {
```
elliefm, line 77: Ah yep -- make the mock reject the bad rcpt like a real server would, and then if it's no longer getting rejected, we're no longer sending bad rcpt's, nice

cassandane/Cassandane/Cyrus/Sieve.pm
```diff
@@ -7133,4 +7133,28 @@ EOF
...
+
+    xlog $self, "Deliver a message";
+    my $msg = $self->{gen}->generate(subject => "Message 1");
+    $self->{instance}->deliver($msg);
+
+    # Verify that message was sent
+    if ($self->{instance}->{have_syslog_replacement}) {
+        my @lines = $self->{instance}->getsyslog();
+        $self->assert_does_not_match(qr/RCPT TO: code=553 text=5.1.1/, "@lines");
+    }
```
elliefm, line 7157: For a belt-and-suspenders approach, you could also use `check_messages()` to make sure the message actually arrived.  That way, it would keep the test valid even when run in environments where cassandane can't capture syslog.  I guess you'd need to add a "foo@example.com" user to the instance for that to work... there's examples around of that sort of thing, have a look at `test_plus_address_partial_virtdom()` in `Cassandane::Cyrus::Delivery` for one.

elliefm (approved):

cassandane/Cassandane/Cyrus/Sieve.pm
ksmurchison: @elliefm I don't think SMTPServer.pm actually does message delivery, so there would be no redirected message to look for.

However, you suggestion made me think: If the redirect is successful, the original recipient would NOT have the message delivered to their INBOX (unless :copy is used).  It would only end up in their INBOX if the script encounters a runtime error (from the redirect action).  So, I added a check for that.

rsto (approved):

cassandane/Cassandane/Cyrus/Sieve.pm
elliefm: Ahh yeah nice!

PR #4423, opened 2023-03-06
global.c: leniently support message-ids without brackets

rsto:
RFC 5322 is clear: A valid message-id must be enclosed in "<" and ">" brackets.

Unfortunately, some email implementations do not get this right and dump message-ids without brackets in their Message-Id and References headers.

This patch updates the find_msgid function to accept such malformed message-ids iff they are of the form

    WSP dot-atom-text '@' dot-atom-text [WSP|EOF]

and return them with proper closing and end brackets.

This does not apply to the In-Reply-To header, because broken MUAs tend to dump email addresses in that header.

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_get_threadid_bogus_messageid
+++ b/cassandane/tiny-tests/JMAPEmail/email_get_threadid_bogus_messageid
@@ -1,0 +1,143 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_get_threadid_nobracket_messageid
+    :min_version_3_7 :needs_component_jmap
+{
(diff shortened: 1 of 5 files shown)


ksmurchison (approved):
Looks reasonable

rsto:
@ksmurchison added https://github.com/cyrusimap/cyrus-imapd/pull/4423/commits/661adb52325efc8e170f6108e5dcd44885a478c1

ksmurchison (approved):
Update looks good

rsto:
This needs more work.

rsto:
We are not going to follow this idea.

PR #4426, opened 2023-03-07
mboxlist.c: don't orphan I records when deleting a mailbox

ksmurchison:


elliefm:
backporting: not sure how far back this needs to go, we'll find out when we get there.  Off the top of my head, I think I records were introduced with 3.6, which that being the case: that far.

ksmurchison:
@elliefm The commit that broke this was 
https://github.com/cyrusimap/cyrus-imapd/commit/8541ae90b3d4bc99df0a3a0ba19388ee409a861e

PR #4428, opened 2023-03-10
JMAP Calendar: preserve iCalendar x-properties in updated event

rsto:
This changes the CalendarEvent/set{update} to preserve unknown iCalendar x-properties of the original iCalendar resource.

While implemented as a Cyrus extension property in the JSCalendar object, this property only is exposed Cyrus-internally.

### DIFF
--- a/imap/jmap_ical.c
+++ b/imap/jmap_ical.c
@@ -71,6 +71,7 @@
 #include "http_proxy.h"
 #include "ical_support.h"
 #include "icu_wrap.h"
+#include "jcal.h"
 #include "json_support.h"
 #include "mailbox.h"
 #include "mboxlist.h"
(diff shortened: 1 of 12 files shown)


imap/jmap_ical.c
```diff
@@ -1627,6 +1628,57 @@ static int identity_int(int i) {
...
+    if (strncasecmpsafe(name, "X-", 2))
+        return 1;
+
+    // all JMAP extension property names are reserved
+    if (!strncasecmpsafe(name, "X-JMAP", 6))
+        return 1;
+
+    // we used this for useDefaultAlerts on VEVENTs
+    if (!strcasecmpsafe(name, "X-APPLE-DEFAULT-ALARM") &&
+            kind == ICAL_VEVENT_COMPONENT)
```
ksmurchison, line 1643: Probably a little nit-picky, but I would do the component type check first since its cheaper than a string comparison.  No need to do the string check if we're going to ignore it.

imap/jmap_ical.c
```diff
@@ -1627,6 +1628,57 @@ static int identity_int(int i) {
...
+        return 1;
+
+    // this is converted as a Location object
+    if (!strcasecmpsafe(name, "X-APPLE-STRUCTURED-LOCATION") &&
+            kind == ICAL_VEVENT_COMPONENT)
```
ksmurchison, line 1648: Same comment as above

ksmurchison (approved):
Other than my 2 comments, LGTM

imap/jmap_ical.c
rsto: I know and actually had written this first the way you suggested. I figured the way it's currently written is more consistent to read with the other conditionals, and the cost of the string comparison is negligible. But since you noticed, I'll rewrite it the previous variant.

PR #4429, opened 2023-03-10
jmap_backup.c: don't setup for next batch if no remaining records

ksmurchison:


cassandane/Cassandane/Cyrus/JMAPBackup.pm
```diff
@@ -967,6 +969,95 @@ sub test_restore_calendars_batch_size_bug_slow
...
+            onDestroyRemoveEvents => JSON::true,
+         }, "R2."]
+    ]);
+    $self->assert_str_equals($calid, $res->[0][1]{destroyed}[0]);
+
+    xlog "restore calendar";
+    $res = $jmap->CallMethods([
+        ['Backup/restoreCalendars', {
+            performDryRun => JSON::true,
+            undoPeriod => "PT3S",
```
wolfsage, line 1048: I'm surprised 3 seconds works here, we only do a sleep 1, I'd expect this to fail randomly when 3 seconds is before we even created the calendar because tests ran fast...

ksmurchison: The previously-existing test uses the same undoPeriod, so I kept the same.  We can adjust if we detect an issue.  Because these tests are marked as slow, the don't run by default anyways.

wolfsage: Huh! Thanks

rsto (approved):

PR #4430, opened 2023-03-10
index.c: short-circuit _fetch_setseen() on expunged records

ksmurchison:
This fixes the unseen underflow issues found by auditslot.

The only issue here would be if the record is unexpunged, the \Seen flag WON'T be set.

If we think this matters, this PR can be altered to actually update the record on disk, but still skip updating state->numunseen.


brong:
Yeah of course, that makes sense!  I think it's better to NOT write the \Seen flag to disk, since if we unexpunged, the user would prefer to see it still unseen.

brong (approved):

PR #4431, opened 2023-03-14
http_caldav.c: Strip overrides that occur before start of RRULE

ksmurchison:
This is a bugfix for Fantastical when splitting a recurring event with existing overrides prior to the split.
Tested with Fantastical 3.7.8

### DIFF
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -3674,6 +3674,26 @@ static int validate_dtend_duration(icalcomponent *comp, struct error_t *error)
     return 0;
 }
 
+struct override_rock {
+    icalcomponent *ical;
+    uint64_t start;
+    unsigned *stripped;
+};
+
+static void strip_past_override(uint64_t recurid, void *data, void *rock)
+{
+    struct override_rock *orock = (struct override_rock *) rock;
+
+    if (recurid < orock->start) {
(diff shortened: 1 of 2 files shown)


imap/http_caldav.c
```diff
@@ -3730,11 +3754,9 @@ static int caldav_put(struct transaction_t *txn, void *obj,
         goto done;
     }
 
+    construct_hashu64_table(&overrides, 10, 0);
```
rsto, line 3757: This magic number seems unnecessarily small. If we can't bother to count the number of overrides in a linear scan of the icalcomponent, then just use some "safer" higher limit like 128 or even 512 here. We've seen Cyrus choke on the rare large size iCalendar choke before.

rsto (changes requested):
Please add a Cassandane test.

imap/http_caldav.c
ksmurchison: Done.  I split the diff and went with 256.

rsto:
@ksmurchison are you still writing that test?

ksmurchison:
Got distracted.  Writing a test now.

ksmurchison:
@rsto Cass test added

rsto (approved):
LGTM

PR #4432, opened 2023-03-14
DAV sharing invite-reply common-name

ksmurchison:
This allows Fantastical to show the sharee name in the reply notifications


rsto (approved):
LGTM, but can we please get a test for this?

PR #4436, opened 2023-03-16
imapurl: refactor toURL to take dynamically growing buf argument

rsto:
This fixes a potential memory corruption issue that was highlighted by https://github.com/cyrusimap/cyrus-imapd/pull/4433

### DIFF
--- a/imap/mboxevent.c
+++ b/imap/mboxevent.c
@@ -872,7 +872,6 @@ EXPORTED void mboxevent_set_access(struct mboxevent *event,
                                    const char *userid, const char *mailboxname,
                                    const int ext_name __attribute__((unused)))
 {
-    char url[MAX_MAILBOX_PATH+1];
     struct imapurl imapurl;
     int r;
 
@@ -896,12 +895,12 @@ EXPORTED void mboxevent_set_access(struct mboxevent *event,
     imapurl.mailbox = extname;
     mbname_free(&mbname);
 
-    imapurl_toURL(url, &imapurl);
-
     // All events want a URI parameter, which in the case of Login/Logout
     // might be useful if it took in to account TLS SNI for example.
     if (!event->params[EVENT_URI].filled) {
-        FILL_STRING_PARAM(event, EVENT_URI, xstrdup(url));
+        struct buf url = BUF_INITIALIZER;
+        imapurl_toURL(&url, &imapurl);
+        FILL_STRING_PARAM(event, EVENT_URI, buf_release(&url));
     }
 
     // Login and Logout events do not have a mailboxname, so avoid looking that up...
@@ -1537,7 +1536,6 @@ void mboxevent_extract_quota(struct mboxevent *event, const struct quota *quota,
                              enum quota_resource res)
 {
     struct imapurl imapurl;
-    char url[MAX_MAILBOX_PATH+1];
 
     if (!event)
         return;
@@ -1577,13 +1575,12 @@ void mboxevent_extract_quota(struct mboxevent *event, const struct quota *quota,
         /* translate internal mailbox name to external */
         char *extname = mboxname_to_external(quota->root, &namespace, NULL);
         imapurl.mailbox = extname;
-
-        imapurl_toURL(url, &imapurl);
-
         free(extname);
 
         if (!event->params[EVENT_URI].filled) {
-            FILL_STRING_PARAM(event, EVENT_URI, xstrdup(url));
+            struct buf url = BUF_INITIALIZER;
+            imapurl_toURL(&url, &imapurl);
+            FILL_STRING_PARAM(event, EVENT_URI, buf_release(&url));
         }
 
         /* Note that userbuf for shared folders is NULL, and xstrdup
@@ -1652,7 +1649,6 @@ EXPORTED void mboxevent_extract_mailbox(struct mboxevent *event,
                                         struct mailbox *mailbox)
 {
     struct imapurl imapurl;
-    char url[MAX_MAILBOX_PATH+1];
 
     if (!event)
         return;
@@ -1685,9 +1681,11 @@ EXPORTED void mboxevent_extract_mailbox(struct mboxevent *event,
     }
 
     /* all events needs uri parameter */
-    imapurl_toURL(url, &imapurl);
-    FILL_STRING_PARAM(event, EVENT_URI, xstrdup(url));
+    struct buf url = BUF_INITIALIZER;
+    imapurl_toURL(&url, &imapurl);
+    FILL_STRING_PARAM(event, EVENT_URI, buf_release(&url));
 
+    buf_free(&url);
     free(extname);
 
     FILL_STRING_PARAM(event, EVENT_MBTYPE,
@@ -1781,7 +1779,6 @@ void mboxevent_extract_old_mailbox(struct mboxevent *event,
                                    const struct mailbox *mailbox)
 {
     struct imapurl imapurl;
-    char url[MAX_MAILBOX_PATH+1];
 
     if (!event)
         return;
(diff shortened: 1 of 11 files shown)


imap/mboxevent.c
```diff
@@ -1665,8 +1665,9 @@ EXPORTED void mboxevent_extract_mailbox(struct mboxevent *event,
     }
 
     /* all events needs uri parameter */
-    imapurl_toURL(url, &imapurl);
-    FILL_STRING_PARAM(event, EVENT_URI, xstrdup(url));
+    struct buf url = BUF_INITIALIZER;
+    imapurl_toURL(&url, &imapurl);
+    FILL_STRING_PARAM(event, EVENT_URI, buf_release(&url));
```
ksmurchison, line 1671: Do we want buf_free() here for completeness?  I know nothing gets leaked right now, but it might in the future.

imap/mboxevent.c
```diff
@@ -1774,8 +1774,9 @@ void mboxevent_extract_old_mailbox(struct mboxevent *event,
     char *extname = mboxname_to_external(mailbox_name(mailbox), &namespace, NULL);
     imapurl.mailbox = extname;
 
-    imapurl_toURL(url, &imapurl);
-    FILL_STRING_PARAM(event, EVENT_OLD_MAILBOX_ID, xstrdup(url));
+    struct buf url = BUF_INITIALIZER;
+    imapurl_toURL(&url, &imapurl);
+    FILL_STRING_PARAM(event, EVENT_OLD_MAILBOX_ID, buf_release(&url));
```
ksmurchison, line 1780: Same as above.

lib/imapurl.c
```diff
@@ -449,43 +448,43 @@ EXPORTED int imapurl_fromURL(struct imapurl *url, const char *s)
...
         }
     }
     if (url->urlauth.access) {
         if (url->urlauth.expire) {
-            strcpy(dst, ";EXPIRE=");
-            dst += strlen(dst);
-            dst += time_to_iso8601(url->urlauth.expire, dst, INT_MAX, 1);
+            buf_appendcstr(dst, ";EXPIRE=");
+            char buf[RFC3339_DATETIME_MAX+1] = { 0 };
+            time_to_iso8601(url->urlauth.expire, buf, RFC3339_DATETIME_MAX, 1);
```
ksmurchison, line 481: Since time_to_iso8601() returns the length written, we *could* expand the dst buf by RFC3339_DATETIME_MAX, write directly into it at the current offset, and then truncate it to the correct length.

ksmurchison (approved):
LGTM other than my non-blocking comments

lib/imapurl.c
rsto: I thought to keep this part of the code to be as simple as possible.

perl/imap/IMAP.xs
```diff
@@ -717,26 +717,23 @@ imclient_toURL(client,server,box)
...
 
-        if(!out_buf[0]) {
-                safefree(out_buf);
+        if(!buf_len(&buf)) {
+                buf_free(&buf);
                 XSRETURN_UNDEF;
         }
 
-        XPUSHs(sv_2mortal(newSVpv(out_buf, 0)));
+        XPUSHs(sv_2mortal(newSVpv(buf_release(&buf), 0)));
```
elliefm, line 734: I'm not sure if this is correct, but I'm also not sure how to find out.  Maybe Rik or Rob M might know?

The original `out_buf` was managed with `safemalloc()`/`safefree()`, but the new one in `buf` will be from ordinary `malloc()` or `realloc()`.  `newSVpv()` looks like it might care about the difference... I'm guessing its name might stand for something like "new safe value from pointer value", in which case it might not be valid to pass it memory from an ordinary malloc.

perl/imap/IMAP.xs
```diff
@@ -717,26 +717,23 @@ imclient_toURL(client,server,box)
...
 
         /* newSVpv copies this */
-        safefree(out_buf);
+        buf_free(&buf);
```
elliefm, line 737: Though actually, the comment here says that `newSVpv` _copies_ its argument, and the original `out_buf` was then freed, supporting that claim.

In which case, I think it's fine to use `newSVpv()` with the malloced memory from `struct buf`, because it doesn't take ownership -- but then the argument should be `buf_cstring(&buf)`, not `buf_release(&buf)`, otherwise the buffer will be ownerless and leaked.

imap/mboxevent.c
```diff
@@ -876,13 +875,15 @@ EXPORTED void mboxevent_set_access(struct mboxevent *event,
...
+    struct buf url = BUF_INITIALIZER;
+    imapurl_toURL(&url, &imapurl);
 
     // All events want a URI parameter, which in the case of Login/Logout
     // might be useful if it took in to account TLS SNI for example.
     if (!event->params[EVENT_URI].filled) {
-        FILL_STRING_PARAM(event, EVENT_URI, xstrdup(url));
+        FILL_STRING_PARAM(event, EVENT_URI, buf_release(&url));
     }
+    buf_free(&url);
```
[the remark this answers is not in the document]
rsto: Did not see that, thanks!

lib/util.c
```diff
@@ -1818,6 +1839,23 @@ EXPORTED int hex_to_bin(const char *hex, size_t hexlen, void *bin)
...
 }
 
+EXPORTED int buf_hex_to_bin(struct buf *bin, const char *hex, size_t hexlen)
+{
+    if (hex == NULL)
+        return -1;
+    if (hexlen == 0)
+        hexlen = strlen(hex);
+
+    size_t newlen = bin->len + hexlen / 2;
```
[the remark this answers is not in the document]
rsto: I did only add the checks that are relevant for buf_hex_to_bin to work properly until it calls hex_to_bin. But I now duplicated the check.

perl/imap/IMAP.xs
rsto: right, newSVpv takes a const pointer. I've changed buf_release to buf_cstring here. thanks.

lib/util.c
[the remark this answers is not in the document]
elliefm: I was confused because I couldn't see the changes, but I was looking at the "util: ..." commit.  Now I see you've applied these changes, but in the "imapurl: ..." commit instead.  Which is surprising, did you accidentally squash them into the wrong commit?

ksmurchison (changes requested):
This PR causes imapd (and maybe httpd) to crash when running the yath tests.  
I spent so much time digging into the tests thinking that my IMAP QUOTA changes were the issue, that when I finally figured out that this PR was the issue, I was too frustrated to dig into the core dump.  Sorry

lib/util.c
```diff
@@ -1780,6 +1780,17 @@ EXPORTED int bin_to_hex(const void *bin, size_t binlen, char *hex, int flags)
     return p-hex;
 }
 
+EXPORTED int buf_bin_to_hex(struct buf *hex, const void *bin, size_t binlen, int flags)
+{
+    size_t seplen = _BH_GETSEP(flags) && binlen ? binlen - 1 : 0;
```
ksmurchison, line 1785: I think the above is incorrect.  it should ne binlen **+** 1, correct?

ksmurchison (changes requested):

lib/util.c
rsto: Turns out adding +1 was correct, but not in this line. We want `newlen` to only be set to the actual length of the current buffer content length plus the length of the new hex-encoded data. But we need to allocate one more byte in `buf_ensure` two lines later, because the underlying `bin_to_hex` function writes a finishing zero byte.

Classic case of "fix a buffer overflow" and accidentally introduce a new 1-byte buffer overflow...

I added an according test to cunit (`cunit/binhex.testc:test_bin_to_hex_realloc`) and ran the updated branch on all our available test suites.

elliefm:
This was backported as #4482

PR #4438, opened 2023-03-16
search_xapian: return error rather than abort on mailbox renames

rsto:
A race condition in mailbox renames could cause squatter to abort on an unmet assertion. This changes it log and return an error rather.

Writing an automated for this is not trivial, as we can't get the timing in Cassandane right to trigger the race condition.


rsto:
We definitely need to handle pendingdelete user renames better. But not in this pull request.

ksmurchison (approved):

PR #4448, opened 2023-03-21
bc_eval.c: Properly handle system tzoffset for [CURRENT]DATE actions

ksmurchison:


brong (approved):
I'm not sure why we're using localtime and a localtime offset rather than gmtime... but it looks sensible otherwise.

PR #4449, opened 2023-03-21
idled.c: squash comparison of integer expressions of different signedness

ksmurchison:
My local compiler (GCC 12.2.1 20221121) didn't complain, but this failed to build on Debian 10.2.1-6


brong (approved):
Yep, that's good

PR #4450, opened 2023-03-21
cass: now depends on DateTime::Format::Mail

elliefm:
Adds the missing dependency to Cassandane's README.deps, and installs it during the github actions CI run.

Cassandane has a lot more dependencies than those which are installed from the workflow script, which I'm not sure where they come from, but I guess they get bundled into the docker image.  Adding DateTime::Format::Mail like this is a workaround to get CI working again.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,6 +39,7 @@ jobs:
       run: |
         cpanm IO::File::fcntl
         cpanm Digest::CRC
+        cpanm DateTime::Format::Mail
       shell: bash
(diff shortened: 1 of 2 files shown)


ksmurchison (commented):
Looks fine to me, but do we want to do this or rewrite the test to use str2time() as Rik suggested?

elliefm:
I don't have an opinion one way or another really, but if you wanna do that I can review a PR, and merge one or the other.  I guess it would be nice to drop the dependency that's only used once.

PR #4451, opened 2023-03-22
Sieve.pm: remove dependencies on DateTime::TimeZone and DateTime::Format::Mail

ksmurchison:


rjbs (approved):
LGTM

cassandane/Cassandane/Cyrus/Sieve.pm
```diff
@@ -3875,22 +3874,18 @@ EOF
 
 sub test_date_local_zone
     :needs_component_sieve :min_version_3_9
-    :needs_component_httpd
```
elliefm, line 3878: Every test in this file needs `:needs_component_httpd` (or something http-dependent, like `:needs_component_jmap`), whether or not it actually uses any HTTP functions, because the `new()` constructor asks for `services => [ ..., 'http']`.  I just added all these a few weeks ago so it would stop trying to run these tests on bare configure builds, because they were erroring out due to being unable to start httpd

ksmurchison: @elliefm I'm beyond using my brain for anything useful tonight. Can you elaborate on what you would like to see changed?

elliefm: no worries, I'll just fix it myself, expect a review request next time you're working :)

PR #4452, opened 2023-03-22
bc_eval.c another local time fix (local offset changes with DST - duh!)

ksmurchison:
@brong do you have a machine currently using Australia/Melbourne that you can test this on (Sieve.date* tests)?

### DIFF
--- a/cassandane/Cassandane/Cyrus/Sieve.pm
+++ b/cassandane/Cassandane/Cyrus/Sieve.pm
@@ -3786,7 +3786,7 @@ sub test_date
     $self->{instance}->install_sieve_script(<<'EOF'
 require ["date", "variables", "imap4flags", "regex", "relational"];
 
-if date :originalzone "date" "date" "2018-05-16" {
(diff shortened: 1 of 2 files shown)


cassandane/Cassandane/Cyrus/Sieve.pm
```diff
@@ -3786,7 +3786,7 @@ sub test_date
     $self->{instance}->install_sieve_script(<<'EOF'
 require ["date", "variables", "imap4flags", "regex", "relational"];
 
-if date :originalzone "date" "date" "2018-05-16" {
+if date :originalzone "date" "date" [ "2018-05-16", "2018-12-16" ] {
```
elliefm, line 3789: Ahh yeah -- always test both summer and winter times, good thinking.

ksmurchison:
@elliefm Can you test this in your zone? I think I finally have it correct. Using December as you suggested revealed the problem on my end

elliefm (approved):
That seems to have fixed it -- Sieve.date and Sieve.date_iana_tzid are both passing in Melbourne DST now

PR #4455, opened 2023-03-24
Sieve.pm: make test_date_local_zone immune to Cass/Cyrus diff in seconds

ksmurchison:


cassandane/Cassandane/Cyrus/Sieve.pm
```diff
@@ -3905,10 +3905,10 @@ sub test_date_local_zone
 
     my $cur_hour = $now->strftime("%H");
     my $cur_zone = $now->strftime("%z");
-    my $cur_std11 = $now->strftime("%a, %d %b %Y %T %z");
+    my $cur_std11 = $now->strftime("%a, %d %b %Y %H:[0-9]{2}:[0-9]{2} %z");
```
elliefm, line 3908: This fails in Australia because the `%z` in strftime expands to "+1100", resulting in a regex like this:

`Mon, 27 Mar 2023 11:[0-9]{2}:[0-9]{2} +1100`

and then the match ends up looks for "one or more spaces, followed by the digits 1, 1, 0, 0", but instead finds "one space, one plus, followed by digits 1, 1, 0, 0", and the match fails.

Mangling it like this seems to fix it for me, though it's inelegant:

```suggestion
    my $cur_std11 = $now->strftime("%a, %d %b %Y %H:[0-9]{2}:[0-9]{2} %z");
    $cur_std11 =~ s/\+/[+]/g;  # escape any '+' from %z
```

elliefm (changes requested):

cassandane/Cassandane/Cyrus/Sieve.pm
ksmurchison: Fixed with your suggestion

rsto (approved):

PR #4457, opened 2023-03-24
jmap_mail.c: add support for state tokens to Email/import

ksmurchison:


rsto (approved):

elliefm:
Should this be backported to 3.8?  The 3.8 branch is currently failing CI runs because JMAPTestSuite has been updated to expect the new fields, but 3.8 doesn't have them.  

The change itself looks tiny, but suspiciously so... is it building upon something else that's on master only?

elliefm:
I guess it's building on top of something else... the commit cherry-picks to 3.8 cleanly, but the build fails.  Hmm

Can we make the import tests conditional on the Cyrus version?  Yes we can, by editing https://github.com/cyrusimap/cyrus-imapd/blob/master/cassandane/Cassandane/Cyrus/JMAPTestSuite.pm#L60-L67 and https://github.com/cyrusimap/cyrus-imapd/blob/master/cassandane/Cassandane/Cyrus/JMAPTestSuiteWS.pm#L61-L68

So the question is, do we just make these tests conditional on 3.9+, or do we try to identify everything that needs to be backported to 3.8 to make it work?  I guess even if we did that, it'd still fail on 3.6, so we'd still need to make these tests conditional on some version.

So I think I've convinced myself to not bother backporting, but to just make these tests version-dependent.  PR incoming...

elliefm:
@ksmurchison Ahh -- it's dependent on `jmap_modseq()`, which was added by #4345

PR #4458, opened 2023-03-24
append.c: add username and accountId to annotator dlist

ksmurchison:
Per a request from @rjbs 

Is there a way to do any testing for these fields in Cass?

### DIFF
--- a/imap/append.c
+++ b/imap/append.c
@@ -595,6 +595,8 @@ out:
  */
 static void callout_encode_args(struct buf *args,
                                 const char *fname,
+                                const char *username,
+                                const char *accountid,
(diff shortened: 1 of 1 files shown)


elliefm:
Looks like Annotator.pm uses this feature... have a look at `new()` (sets up `annotation_callout` option), and `start_my_instances()` (arranges for `utils/annotator.pl` to run as a daemon service -- so have a look at that too)

rsto:
@ksmurchison this PR is still a draft, do you already want my review?

rsto:
@ksmurchison please add reviewers again once it's clear if to continue with this

PR #4459, opened 2023-03-24
jmap_quota.c: Implement JMAP quotas per RFC9425

ksmurchison:
The tin says it all

### DIFF
--- a/imap/quota_db.c
+++ b/imap/quota_db.c
@@ -370,6 +370,7 @@ struct quota_foreach_t {
     quotaproc_t *proc;
     void *rock;
     struct txn **tid;
+    unsigned use_conv : 1;
 };
 
 static int do_onequota(void *rock,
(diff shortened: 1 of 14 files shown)


imap/quota_db.c
```diff
@@ -409,13 +419,27 @@ EXPORTED int quota_foreach(const char *prefix, quotaproc_t *proc,
...
     return r;
 }
 
+EXPORTED int quota_foreach(const char *prefix, quotaproc_t *proc,
+                           void *rock, struct txn **tid)
+{
+    return _quota_foreach(prefix, proc, rock, tid, 0);
+}
+
+EXPORTED int quota_foreach_withconversations(const char *prefix,
```
rsto, line 436: nit: I would have added a flag to `quota_foreach` rather than a new function here, but we have no hard rules for either. Is there a chance that we'll want to extend `quota_foreach` in the future with additional options? Did you weigh both variants against each before you chose a new function, if so what was the rationale?

rsto (commented):
This looks good to me, but I haven't checked if it covers the spec in all detail. Just have one question before I hit 'Approve'.

imap/quota_db.c
ksmurchison: Done

rsto (approved):

PR #4463, opened 2023-03-29
JMAPTestSuite: Email:import behaviour changed in 3.8+

elliefm:
These tests now expect the new behaviour (#4457), so suppress them when Cyrus is older

This will fix the failing CI on the 3.8 branch once it's backported.  It'll need to be backported to every branch whose Cassandane runs JMAPTestSuite.

### DIFF
--- a/cassandane/Cassandane/Cyrus/JMAPTestSuite.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPTestSuite.pm
@@ -64,6 +64,8 @@ my %suppressed;
 # dot-separated.  See skip_before() below for implementation details.
 my %notbefore = (
     't:Email:get:header-header-field-name' => '3.5',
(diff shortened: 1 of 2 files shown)


ksmurchison (dismissed):
This looks sane, but I'm wondering if we want to backport the Email/import state token changes to whatever version(s) that we think we claim conformance with RFC 8621.

elliefm:
@ksmurchison Yeah -- I might drop this one back to Draft for now, until we work out if we're backporting the other one or not.

If we do backport the other one, then I'll change "3.9+" in this one to be however far we backported it.

elliefm:
I think the other PR will be going back to 3.8, but no farther, so I've updated this to mark the tests "not before 3.8".

ksmurchison (approved):

PR #4465, opened 2023-03-30
mbdump: call mailbox_iter_done on several places

dilyanpalauzov:


rsto (approved):
Thanks alot.

elliefm:
I'm not sure how far this wants to be backported, but I'm tagging them all for now, and I'll find out when I get there.

PR #4468, opened 2023-03-31
relocate_by_id:main() frees statically allocated data

dilyanpalauzov:
By just reading the code I found out, that user_sieve_path() returns pointer to statically allocated data.  In relocate_by_id.c:main() this data is appended into strarray_t *oldpaths, and finally strarray_free(oldpaths); is called.  The latter free()s that statically allocated data.

### DIFF
--- a/imap/relocate_by_id.c
+++ b/imap/relocate_by_id.c
@@ -298,7 +298,7 @@ int main(int argc, char **argv)
                 /* Add sieve path */
                 path = user_sieve_path(userid);
                 if (*path) {
-                    strarray_append(oldpaths, path);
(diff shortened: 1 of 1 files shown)


elliefm (approved):
Good catch, thanks

ksmurchison:
Unless I'm missing something, there is nothing wrong with the existing code.  strarray_append() already does xstrdupnull() on the string passed to it.  So all this PR has done is create a memory leak.

There WOULD be a problem had we been using strarray_appendm(), but we are not.

elliefm:
Oh... good catch, thanks 😅

PR #4469, opened 2023-03-31
cassandane: support custom cassandane.ini path in Makefile

rsto:
This updates the Cassini object to also look for its initialization file at the path defined in the optional CASSINI_FILENAME environment variable.

It has precedence over the standard cassandane.ini locations, but lower precedence than the `--config` command line argument.

Was: _This adds the CASSANDANE_INI variable to the cassandane Makefile. Before, cassandane builds only supported cassandane.ini file locations at one of the standard directories. Now, the location of a custom cassandane.ini file path can be defined before the build._

### DIFF
--- a/cassandane/Cassandane/Cassini.pm
+++ b/cassandane/Cassandane/Cassini.pm
@@ -69,6 +69,11 @@ sub new
         # explicitly requested filename: just use it
         $filename = $params{filename};
     }
+    elsif (defined $ENV{CASSINI_FILENAME}) {
+        xlog "Using ini file from environment:"
+             . " filename=\"$ENV{CASSINI_FILENAME}\"";
+        $filename = $ENV{CASSINI_FILENAME};
+    }
     else {
         # check some likely places, in order
         foreach my $dir (q{.},


cassandane/Cassandane/Cassini.pm
```diff
@@ -131,11 +131,11 @@ sub new
 
 sub instance
 {
-    my ($class) = @_;
+    my ($class, %params) = @_;
 
     if (!defined $instance)
     {
-        $instance = Cassandane::Cassini->new();
+        $instance = Cassandane::Cassini->new(%params);
```
elliefm, line 138: This is kind of awkward, because if you call `Cassandane::Cassini->instance(%myparams)` you would reasonably expect to get back a Cassini object initialised with those params, but the actual result would depend on whether or not something else already initialised the singleton, which smells like an annoying race condition waiting to happen

cassandane/Makefile
```diff
@@ -60,7 +60,10 @@ MODULES := $(shell find . -type f -name '*.pm' | sort)
 # checks to proceed when no Cyrus install exists
 ifndef NOCYRUS
     SCRIPTS += $(shell find . -type f -name 'annotator.pl')
-    CYRUS_PERL_PATHS := $(shell $(PERL) utils/cyrus-perl-paths.pl)
+    ifdef CASSANDANE_INI
+	CYRUS_PERL_PATHS_ARGS := --config $(CASSANDANE_INI)
+    endif
+    CYRUS_PERL_PATHS := $(shell $(PERL) utils/cyrus-perl-paths.pl $(CYRUS_PERL_PATHS_ARGS))
```
elliefm, line 66: If this behaviour is going to be based around an environment variable being set anyway, would it not be simpler to just have Cassini itself look for a cassandane.ini path defined in an environment variable, rather than all this work to pass it down?  And then anything that uses Cassini (such as cyrus-perl-paths.pl, but also all of cassandane) doesn't need any special handling of its own, it can just ask for a Cassini instance as usual and get something coherent.

Cassini can already read any or all of its config entirely from the environment -- with potentially no cassandane.ini required at all (see `sub val` for the main implementation, and the end of `sub new` for the handling when no cassandane.ini file was found, where a `CASSINI_CASSANDANE_ALLOW_NOINIFILE` can suppress the error).  The only configuration detail it can't currently read from the environment is a path to a cassandane.ini file, and that seems trivial enough to add?  (Probably in `sub new` somewhere, but I can't just add a suggestion because a review won't let me comment on lines that weren't edited already.  I would probably call this environment variable something like `CASSINI_FILENAME`, for consistency with the other environment variables it looks for.)

cassandane/utils/cyrus-perl-paths.pl
```diff
@@ -32,7 +32,27 @@ sub _cyrus_perl_search_path
...
+    {
+        $cassini_params{filename} = shift;
+    }
+    else
+    {
+        usage;
+    }
+}
+
+my $cassini = Cassandane::Cassini->instance(%cassini_params);
```
elliefm, line 55: Since this is a single process script which only ever calls methods directly on its own `$cassini` object, it could bypass `instance()` and call `new()` directly with the params it wants, and then we wouldn't need the awkward racey change to `instance()`

But I think the better solution is just making Cassini->new() able to read a cassandane.ini path from the environment as per my other comment, and then nothing else needs to change at all

elliefm (changes requested):
This could be a lot simpler and a lot more robust as a result -- details in inline comments

rsto:
@elliefm I'm happy with using an environment variable in contrast to all this. I did not do this, because I think to remember that you had deliberately not supported a CASSANDANE_INI variable, because it would make it harder to reason instead of passing command line arguments. I'll change the PR to just use an environment variable.

elliefm:
@rsto Hmm, that seems like a thing I might have said, but I've forgotten having said it, heh.  I think as long as Cassini->new() xlogs that it's getting the filename from the environment, that's probably fine.

elliefm (approved):

Issue #4461, opened 2023-03-26
cassandane :want_foo magic


elliefm:
Maybe we can add a new magic attribute set for cassandane tests so that individual tests can specify which services/etc they need, rather than them being set on a whole-suite basis in the constructor.  This would remove the need for junk like #4454 

I imagine this being called `:want_` and working a bit like `:needs_`.  The latter looks for any `:needs_*` attribute, parses it into tokens, and then looks for a suitable true value in the `cyr_buildinfo` data.  If such an entry doesn't exist or contains a false value, the test isn't run.  We could make a similar parser for `:want_`, and use it to inject stuff into the TestCase object's `want` structure before the instances are created (e.g. `:want_imapmurder` would be the same as `want => { imapmurder => 1 }`, `:want_services_http` would be the same as `want => { services => [ 'http' ] }`, and so on).

As a first pass this would probably be distinct from :needs, and therefore tests that want an http service would need both `:needs_component_httpd` and `:want_services_http` attributes.  But we can probably refine it further, once we've gotten a feel for what sort of repetitive patterns we use a lot.

rjbs:
Fixed by #4470

PR #4470, opened 2023-04-03
cassandane: add :want_foo magic for per-test configuration of test properties

elliefm:
This PR adds `:want_foo` and `:want_service_foo` magic to Cassandane::Cyrus::TestCase.  These allow properties controlled by the `want` hash to be set easily on a test-by-test basis.  Previously, you had to either set them in the constructor (affecting the whole suite); or mark individual tests with `:NoStartInstances`, and then tinker around with the options prior to calling `$self->_start_instances()`.

The list of wantable things (and their defaults) is defined as `$want` in `sub new` (unchanged).  The list of recognised service names is defined in Cassandane::ServiceFactory (also unchanged).

I have updated the Sieve test suite to make use of the new behaviour.  This suite particularly needs a http service for some, but not all, of its tests.  With the new attribute magic, the tests that need a http service can arrange to have one (or be skipped), and the tests that don't need a http service aren't skipped when Cyrus was built without http support for whatever reason.

This feature is not available for test modules that subclass directly from the Cassandane::Unit::TestCase superclass (n.b. "Unit" vs "Cyrus"), as that class has no concept of "want", and is mostly only used directly by Cassandane's self-tests anyway.

First part of #4461

### DIFF
--- a/cassandane/Cassandane/Cyrus/Sieve.pm
+++ b/cassandane/Cassandane/Cyrus/Sieve.pm
@@ -91,7 +91,7 @@ sub new
             config => $config,
             deliver => 1,
             jmap => 1,
-            services => [ 'imap', 'sieve', 'http' ],
+            services => [ 'imap', 'sieve' ],
             adminstore => 1,
     }, @_);
 }
@@ -257,7 +257,6 @@ sub compile_sieve_script
 
 sub test_vacation_with_following_rules
     :needs_component_sieve :min_version_3_0
-    :needs_component_httpd
 {
     my ($self) = @_;
 
@@ -299,7 +298,7 @@ EOF
 }
 
 sub test_deliver
-    :needs_component_sieve :needs_component_httpd
+    :needs_component_sieve
 {
     my ($self) = @_;
 
@@ -341,7 +340,6 @@ sub test_deliver_specialuse
     :min_version_3_0
     :needs_component_sieve
     :NoAltNameSpace
-    :needs_component_httpd
 {
     my ($self) = @_;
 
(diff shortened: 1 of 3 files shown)


cassandane/Cassandane/Cyrus/Sieve.pm
```diff
@@ -3958,7 +3891,7 @@ EOF
 
 sub test_imip_publish
     :needs_component_sieve :needs_component_httpd :min_version_3_7
```
rsto, line 3893: Maybe I don't understand the new logic, yet, but isn't `needs_component_httpd` unnecessary here and in the following imip_tests? and it it isn't, wouldn't `component_httpd` imply that `service_httpd` is required?

rsto (commented):
Looks good to me, but I either haven't grokked the change fully or there are redundant magic attributes for some tests.

cassandane/Cassandane/Cyrus/Sieve.pm
elliefm: This is addressed by the big XXX comment in the filter function, where the magic happens -- for the time being, if you `:want_service_http` (to have a http service started), you must also `:need_component_httpd` (to ensure the test is skipped if the httpd binary doesn't exist).

There isn't yet any mapping between the two, though I'd like there to be eventually.  The names are different (n.b. "http" vs "httpd") and inherently unrelated (:needs_x_y comes from cyr_buildinfo output, service names come from Cassandane::ServiceFactory), so even though this particular example seems obvious, the general case isn't.  So I want to see if and how we actually use this stuff before I go spending time on deeper infrastructure that we might not use.

rsto (approved):

elliefm:
Would like to backport this cass functionality 3.6, but the accompanying Sieve updates don't apply cleanly

PR #4479, opened 2023-04-13
Update JMAP contacts support to IETF standards

ksmurchison:
[https://datatracker.ietf.org/doc/draft-ietf-jmap-contacts/](url)
[https://datatracker.ietf.org/doc/draft-ietf-calext-jscontact/](url)
[https://datatracker.ietf.org/doc/draft-ietf-calext-jscontact-vcard/](url)
[https://datatracker.ietf.org/doc/draft-ietf-calext-vcard-jscontact-extensions/](url)


elliefm:
This was merged with failing CI checks 👎

ksmurchison:
What's failing?

elliefm:
Should be fixed by #4684 -- will find out when its checks finish...

elliefm:
It introduced a dependency on XML::Simple, which neither the docker image nor the workflow script was installing

PR #4481, opened 2023-04-17
main.yml: update imaptest before running Cassandane

ksmurchison:
This should fix the failure in the notify test (which has been updated upstream), and run any other new/updated tests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,6 +39,8 @@ jobs:
       run: |
         cd /srv/imaptest.git
         git fetch
+        git remote -v
+        git status
         git checkout origin/main
         git clean -f -x -d
         cd -
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,6 +35,14 @@ jobs:
         cpanm --installdeps .
         cd -
       shell: bash
+    - name: update imap test suite
+      run: |
+        cd /srv/imaptest.git
+        git fetch
+        git checkout origin/main
+        git clean -f -x -d
+        cd -
+      shell: bash
     - name: install missing deps
       run: |
         cpanm IO::File::fcntl


wolfsage (commented):
lgtm!

wolfsage (approved):

.github/workflows/main.yml
```diff
@@ -35,6 +35,16 @@ jobs:
...
+    - name: update imap test suite
+      run: |
+        cd /srv/imaptest.git
+        git fetch
+        git remote -v
+        git status
+        git checkout origin/main
+        git clean -f -x -d
+        cd -
+      shell: bash
```
elliefm, line 47: Do you need to run `make` here?  After the `git clean -f -x -d`, all the compiled components will be gone.  The comment in the example cassandane.ini for the `[imaptest]` section says it wants the path to a "built but not installed ImapTest", but this one will just be a bare source tree...

(Did you copy from the JMAP-TestSuite job above?  That one doesn't have any build line because it's perl, there's nothing to compile.)

I don't update ImapTest locally very often myself... I'm still running the nightly release from 2019-05-04. 😅 I suppose I probably should make more effort to run a current one, but I don't remember why I don't.

.github/workflows/main.yml
```diff
@@ -39,6 +39,8 @@ jobs:
       run: |
         cd /srv/imaptest.git
         git fetch
+        git remote -v
```
elliefm, line 42: I see it's failing because of missing 'origin/main'...  maybe add `git branch -vv` here too to see what branches you actually have present.  `git remote -vv` (two v's) can be useful too, I usually use both these commands with two v's all the time, don't remember what the difference between one v and two  was though.

ksmurchison: Matt seems to think that we're using a local repository and not the upstream one, so we aren't able to fetch the updates which fix the failure in the notify test.  I don't know why we have our own local copy, and I don't really have the knowledge to dig in and fix it.

wolfsage: I think the solution for now is - if you have access to [cyrusimap/imaptest](https://github.com/cyrusimap/imaptest), then you can update its master branch to be the upstreams latest main branch, and change this code to use master instead of main, and you're done

elliefm: Might be enough to click this "sync fork" button? (see image).  Idk if there's ramifications though.  Do we know who set up our fork?  Maybe @brong?

Screenshot 2023-04-24 at 9 49 05 am

rsto:
This is obsolete

PR #4485, opened 2023-04-21
cassandane: slow test improvements

elliefm:
Got curious and discovered many of our "_slow" tests run in ~5 seconds each; this doesn't seem slow enough to be worth the loss of regular regression coverage, so I've removed the "_slow" marker from the fast ones.

"calendarevent_get_standalone_instances_slow" got renamed to "calendarevent_get_standalone_instances_multi" along the way, because it turned out there was already a "calendarevent_get_standalone_instances" (which did the same thing, but only once).  The rest just lost their "_slow" marker.

While I was in there, I got frustrated at the difficulty of running just the slow tests, without everything else too, so I added a --slow-only option for testrunner.pl and plumbed it through.

Bug3072.copy_longset_slow (112 seconds) and ClamAV.remove_infected_slow (21 seconds) are still marked slow because they are slow, and are of low-importance IMO.  JMAPBackup.restore_mail_twice_slow is also still marked slow -- but because it fails (quickly), and I didn't want the distraction; I don't know how long it would take if it succeeded.

### DIFF
--- a/cassandane/Cassandane/Cyrus/JMAPBackup.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPBackup.pm
@@ -880,7 +880,7 @@ sub test_restore_calendars_all_dryrun
     $self->assert_null($imip);
 }
 
-sub test_restore_calendars_batch_size_bug1_slow
+sub test_restore_calendars_batch_size_bug1
     :min_version_3_7 :needs_component_jmap
 {
     my ($self) = @_;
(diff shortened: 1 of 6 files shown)


rsto (approved):
Agree, the less `_slow` the better, even at the cost of a minute or more or overall cassandane test runs.

elliefm:
JMAPBackup.restore_mail_twice_slow has since been fixed by #4487, so I'll rebase this on master, see how long that test takes in practice, and if it's fast, I'll remove its `_slow` too.

elliefm:
Looks like that one also only takes ~5 seconds wallclock time, so I've dropped the `_slow` off it too.  Idk if this needs "re-review" per se, but by the time I ask and get back an answer, it would have taken just as long, so I'll just request a re-review rather than asking informally 😅

ksmurchison (approved):
LGTM

elliefm:
The CI failure is just `ImapTest.notify`, which we know about, so I'm ignoring it and merging anyway.

elliefm:
This doesn't apply cleanly to 3.6, I think mostly because of the tiny tests conversion

PR #4486, opened 2023-04-24
Separate JMAP calendar default alarms from Apple default alarms

rsto:
This replaces the current JMAP default alarm implementation and its related PR https://github.com/cyrusimap/cyrus-imapd/pull/4185 .

Before, JMAP default alarms and Apple clients used the same DAV property to store default alarms (as VALARMs). In addition, Cyrus rejected DAV client PROPPATCH requests that attempted to set these properties on the calendar home or principal, which broke default alarms for Apple clients.

Now, Apple default alarms and JMAP default alarms are stored in two separate mailbox annotations. JMAP default alarms only are accessible over JMAP and are hidden in CalDAV. Apple default alarm annotations are treated as any other DAV "dead" properties and can be set on any CalDAV collection or resource.

Both Apple and JMAP default alarms may now co-exist in a VEVENT, where Apple default alarms have the X-APPLE-DEFAULT-ALARM property set (as used by Apple clients) and JMAP default alarms X-JMAP-DEFAULT-ALARM. In addition, a CalDAV PUT on an existing calendar event resource disables useDefaultAlerts for this event if one or all JMAP default alarms are removed or changed.


ksmurchison (approved):
This is a pretty complex PR and I'm not sure where we are with what should/shouldn't be done with default alarms, but I didn't spot any code that looks troublesome.  If it works as expected, I'm fine with it (even though I'm not a fan of default alarms).

PR #4494, opened 2023-05-02
JMAPBackup: increase restore time window length

rsto:
A too short window length makes the test indeterministic.


ksmurchison (approved):
I'm fine with this, but should we try sleep 4 and PT2S to see if that also works reliably?

rsto:
IDK, is the 4 seconds increased runtime of the test worth it? I'd say if 12 seconds gets us to a deterministic test I'd keep it like that.

PR #4495, opened 2023-05-02
Prevent deadlock between SEARCH and squatter

rsto:
This aligns the lock acquisition order of the SEARCH command and squatter to prevent a deadlock between them running concurrently on the same user and mailbox.

The deadlock occurred when squatter had acquired a lock on conversations.db and wanted to acquire an exclusive lock on xapianactive.db. Meanwhile, SEARCH had acquired a lock on xapianactive.db and wanted to acquire an exclusive lock conversations.db.

This patch aligns their lock acquisition order to prevent this deadlock.

Testing this currently is only done manually. A brute-force test is to runs hundreds of SEARCHes and squatters on the same user to hit the deadlock. A better test is done by use of special-purpose tool that I built that traces the acquired locks of a Cyrus process across its lifetime. I do plan to integrate this tool into our automated tests but I don't want this patch to wait until then.

### DIFF
--- a/imap/search_engines.c
+++ b/imap/search_engines.c
@@ -164,16 +164,12 @@ static int search_batch_size(void)
  * Returns an IMAP error code or 0 on success.
  */
 static int flush_batch(search_text_receiver_t *rx,
-                       struct mailbox *mailbox,
                        int flags,
                        ptrarray_t *batch)
 {
     int i;
     int r = 0;
     int indexflags = 0;
-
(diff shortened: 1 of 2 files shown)


elliefm (approved):
I haven't done anything in particular to test this other than the usual cunit/cassandane run, but this does look like it does what it says on the tin, and it sounds like your own testing has been thorough. :)

PR #4496, opened 2023-05-02
lmtp_sieve.c: allow PUBLISH with no ORGANIZER

ksmurchison:
Fix for an existing Cass test.  Not sure when this broke

### DIFF
--- a/imap/lmtp_sieve.c
+++ b/imap/lmtp_sieve.c
@@ -1499,8 +1499,9 @@ static int sieve_imip(void *ac, void *ic, void *sc, void *mc,
 
     cyrus_icalrestriction_check(itip);
     if ((errstr = get_icalcomponent_errstr(itip)) &&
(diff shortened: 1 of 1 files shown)


elliefm (approved):
This looks okay, but what test was failing that this fixed?  I didn't have any failures on latest master this morning... so I wonder if there's something dependent on config differences that shouldn't be

rsto (commented):
Same here, the only Cassandane tests for PUBLISH are in the Cassandane Sieve suite, they all pass for me.
Also: https://www.rfc-editor.org/rfc/rfc5546.html#section-3.2.1 states that ORGANIZER is mandatory for PUBLISH, so I assume this change accepts bogus data seen in the wild?

ksmurchison:
Interesting. Perhaps the Sieve.imip_publish test started failing for me because of a change in libical. I will hold off on merging this and will investigate further.

PR #4500, opened 2023-05-05
V39/cass local month race bug

elliefm:


elliefm:
Need to rebase this once #4548 lands

elliefm:
THIS IS NOT MERGED!

But the assertion changes that were originally on this branch were merged as #4548, and this branch has now been rebased on top, ready for continuing the work on the actual bug fix.  Github seems to be assuming that with all the commits so far merged, the task is now complete, when actually, it means this task hasn't really been started yet.

Not sure if it will figure itself out once I push some new commits to the branch, or if I'll need to make a whole new PR for the actual fix, but I guess I'll find out when I get there.

PR #4501, opened 2023-05-08
github actions: workaround "dubious ownership" failures

elliefm:
We recently built a new cyrus-buster docker image, with updated dovecot and imaptest builds, to resolve the problem with ImapTest.notify failing every CI run.  The new docker image seems to have gotten an updated git that contains a security fix which complains when the directory ownership looks sketchy -- which, it does when running github actions on docker images! https://github.com/actions/runner/issues/2033

Initially, this was failing on the post-checkout fiddling we were doing to make the git tags available to the build (which needs them for versioning), so the first commit here removes the post-checkout fiddling and instead configures the checkout step to get everything we need in the first place.  But we need the `git describe` command during the build to determine the version number, which fails for the same reason, so the second commit adds the workaround anyway.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,10 @@ jobs:
       with:
         fetch-depth: 0
     - name: configure and build
-      run: ./tools/build-with-cyruslibs.sh
+      run: |
+        git config --global --add safe.directory /__w/cyrus-imapd/cyrus-imapd
+        echo "building cyrus version" $(./tools/git-version.sh)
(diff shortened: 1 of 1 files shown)


rsto (approved):
This looks sane to me, but I did not test this. Given that this only is for running test instances, I say that's good enough.

wolfsage:
Nice, thanks

elliefm:
We don't have github actions on 3.2 or 3.4 (yet?), so can't backport this to those versions (until we do).

PR #4507, opened 2023-05-15
developer-testing.rst: explain CI imaptest setup

elliefm:
Updates the developer testing instructions, which advises building Dovecot and IMAPTest from nightly build tarballs, to explain how you can also build these from git by copying what the docker image build does, which will get you the same tests CI runs.

I didn't provide copy-and-pastable commands here, because it relies on checking out a known commit id on the Dovecot repository, and I don't want to have to try to keep this documentation synchronised.  Instead, I just link to the docker image build script, and advise the reader to copy what it does if they want to do the same thing.

Really, my main purpose in writing this is so that in a few years when we next need to update IMAPTest for CI, I don't get confused by our documentation saying one thing, but our process doing another.  I still think just using a matched pair of nightlies is the simplest approach, unless you're specifically trying to reproduce what CI does.


ksmurchison (approved):
LGTM

PR #4512, opened 2023-05-22
docsrc/conf.py: use absolute dev release notes link

elliefm:
so that older branches (and future older branches) don't end up with a dead link to release notes they don't have, and don't need to be kept up to date with current dev release notes either

(3.6 and earlier already got this change during release of 3.8.0; this is a forward port from there)

### DIFF
--- a/docsrc/conf.py
+++ b/docsrc/conf.py
@@ -453,7 +453,7 @@ rst_prolog = """
 
 .. |imap_development_release_notes| raw:: html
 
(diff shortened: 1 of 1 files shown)


elliefm:
@rsto This has slipped through the cracks, can you review please? It's trivial. Thanks :)

elliefm:
@rsto @ksmurchison Can someone please just review this already? It'll take three seconds.

rsto (approved):

PR #4513, opened 2023-05-23
docs: remove references to jmap_guesstz_fname

elliefm:
This was an option that existed briefly on a feature branch, but was rebased out before the feature landed.  A reference to it accidentally hung around in documentation, and wound up in release notes.


rsto (approved):
Thanks

Issue #4071, opened 2022-05-04
Should depend on libical >= 3.0.10


robn:
Compiling master today:

```
imap/itip_support.c: In function ‘deliver_merge_add’:
imap/itip_support.c:958:9: error: ‘ICALTIMETYPE_INITIALIZER’ undeclared (first use in this function)
  958 |         ICALTIMETYPE_INITIALIZER,
      |         ^~~~~~~~~~~~~~~~~~~~~~~~
imap/itip_support.c:958:9: note: each undeclared identifier is reported only once for each function it appears in
imap/itip_support.c:959:9: error: ‘ICALPERIODTYPE_INITIALIZER’ undeclared (first use in this function)
```

I have libical 3.0.9, but the initializer constants in libical didn't appear until 3.0.10 (libical/libical@45ad8c0a).

configure should do a version check for libical.

elliefm:
https://github.com/cyrusimap/cyrus-imapd/blob/d3eb596bccbc834fc24aa3073506669a8354dda0/configure.ac#L1635-L1640

We have such a check, but it thinks 3.0.0 is adequate.  Looks like the most recent release was 3.0.14.  @ksmurchison shall we bump this to 3.0.10, or straight to 3.0.14?  Probably better check what version the cyruslibs fork thinks it is...

The cyrus-imapd commit that starts using `ICALTIMETYPE_INITIALIZER` (c287d863c7) isn't on the 3.6 branch, so we don't need to worry about 3.6 here, just master.

ksmurchison:
If 3.0.10 is the minimum requires, I'd got with that

philfry:
Hi all,
since cyrus-imapd 3.8.0 uses `icalproperty_get_datetime_with_component()` which came with libical-3.0.5, the minimum requirement should be raised accordingly.

elliefm:
Thanks, will update

dilyanpalauzov:
master brach requires libical 3.0.10 since 96ea872427aeba1289f12a93.  3.8 branch requires libical 3.0.10 since bedb08b1f8eb2c2 .

Can this be closed?

elliefm:
Yeah I think so, though the 3.8 change hasn't been released yet.  Guess I should put together a 3.8.1 soon

PR #4514, opened 2023-05-24
Require libical >= 3.0.10

ksmurchison:


imap/ical_support.h
```diff
@@ -203,14 +203,6 @@ extern void icalcomponent_normalize_x(icalcomponent *ical);
...
     icalcomponent_get_first_property(comp, ICAL_ACKNOWLEDGED_PROPERTY)
 
-#ifndef HAVE_RFC7986_COLOR
-
-/* Replacement for missing function in 3.0.0 <= libical < 3.0.5 */
-
-extern icalproperty *icalproperty_new_color(const char *v);
-
-#endif /* HAVE_RFC7986_COLOR */
-
```
elliefm, line 213: Should we also remove the implementation of this from [ical_support.c](https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/ical_support.c#L2121-L2132), or does it need to remain for some reason I don't understand?

elliefm (commented):
Thanks! One question...

elliefm:
The CI is failing on these tests:

* JMAPCalendars.calendarevent_set_fullblown                                            
* JMAPCalendars.calendarevent_get_privacy_shared

and both are failing because they're missing an expected "color" property.  Hmm.

The build succeeded, so the docker image must have a new enough libical.  I wonder if something else is conditional on that `HAVE_RFC7986_COLOR` which is no longer being defined during configure?  But when I search for that string, the only occurrences are the two in ical_support.c that I already identified.  So I'm puzzled by why those two tests should suddenly not get their expected "color" fields anymore.

Another hmm: I can reproduce those two cass failures locally, so that'll make it easier to figure out what's up.  I don't expect a response from you till you're back from leave, so I'll tinker in the meantime, and if I don't figure it out myself, it'll wait.

elliefm:
Oh curious: in both cases the failure looks something like this:

```
 Structures begin differing at:
  $a->{recurrenceOverrides}{2021-01-02T01:00:00}{color} = undef
  $b->{recurrenceOverrides}{2021-01-02T01:00:00}{color} = Does not exist
 at tiny-tests/JMAPCalendars/calendarevent-get-privacy-shared line 199.
```

That is, one of the deep structures being compared has the "color" key with an undefined value, and the other one doesn't have the "color" key at all.  It's not clear to me at a glance which of these is the "actual" vs the "expected".

elliefm:
And it's specifically inside "reccurranceOverrides", not at the higher level.

elliefm:
In JMAPCalendars.calendarevent_set_fullblown, we create an event that has this:

```
        recurrenceOverrides => {
            '2021-02-02T02:00:00' => {
                title => 'recurrenceOverrideTitle',
            },
        },
```

and get this back in the response:

```
  'recurrenceOverrides' => {
    '2021-02-02T02:00:00' => {
      'color' => undef,
      'title' => 'recurrenceOverrideTitle',
    },
  },
```

And then the test compares the original with the result, and complains because it didn't expect a `'color'` key to be present.

But I don't understand why this would have changed from this change...

elliefm:
Oh wild: if I remove the old fallback `icalproperty_new_color()` implementation from ical_support.c and rebuild, like this:

```
diff --git a/imap/ical_support.c b/imap/ical_support.c
index 5003c2ade..104de592d 100644
--- a/imap/ical_support.c
+++ b/imap/ical_support.c
@@ -2117,20 +2117,6 @@ EXPORTED int icalcomponent_apply_vpatch(icalcomponent *ical,
     return 0;
 }
 
-
-#ifndef HAVE_RFC7986_COLOR
-
-/* Replacement for missing function in 3.0.0 <= libical < 3.0.5 */
-
-EXPORTED icalproperty *icalproperty_new_color(const char *v)
-{
-    icalproperty *prop = icalproperty_new_x(v);
-    icalproperty_set_x_name(prop, "COLOR");
-    return prop;
-}
-
-#endif /* HAVE_RFC7986_COLOR */
-
 EXPORTED const char *icaltimezone_get_location_tzid(const icaltimezone *zone)
 {
     const char *v = icaltimezone_get_location((icaltimezone*) zone);
```

then both tests start passing.  So now I understand:

* the fallback `icalproperty_new_color()` had a different behaviour from the libical one when its argument was NULL, but we never noticed because we always have new enough libical that we don't use the fallback
* when this PR removed the libical check and the `HAVE_RFC7986_COLOR`, the fallback implementation started being compiled in, because `HAVE_RFC7986_COLOR` was no longer defined, and shadowed the real library function
* we started using the fallback instead of the real library function, and started seeing the different behaviour
* which means the fix really is just cleaning up the old fallback function, like I noticed in my original review

I'll just push a commit that does that, since I'm already this deep anyway...

imap/ical_support.h
elliefm: I wound up removing it myself!

elliefm:
@rsto I guess you'd better review this now, since I've got commits on the branch now myself

elliefm:
This fixes #4071

rsto (approved):

PR #4520, opened 2023-06-02
docs: remove post-upgrade survey link

elliefm:
we're not doing anything useful with the responses anyway


elliefm:
note to self: will need to backport this all the way to 3.0, maybe even 2.5, for the link to completely disappear from the web

rsto (approved):

PR #4521, opened 2023-06-05
JMAPTestSuite: don't leave behind defunct squatter procs

elliefm:
Noticed while investigating something else that JMAPTestSuite (and JMAPTestSuiteWS) create a lot of defunct squatter processes that don't get cleaned up until cassandane itself exits.  Turns out we were killing the process after each test, but not calling waitpid on it.

This arranges for it to be cleaned up properly, using the already existing mechanism.  Rolling squatter exits with EX_TEMPFAIL when shut down with SIGTERM, so it also needs the custom exited_abnormally that ignores EX_TEMPFAIL (but still dies for any other non-zero exit code).


rsto (approved):

wolfsage:
lgtm,thanks!

wolfsage (approved):

PR #4522, opened 2023-06-05
register DAEMON and EVENTS processes so cyr_info proc can list them

elliefm:
**reviewers:** This will be easiest to review commit-by-commit.

`cyr_info proc` lists all the running processes for cyrus.conf SERVICES entries.  This PR extends this to also include processes for EVENTS and DAEMON entries.  With this change, `cyr_info proc` lists all* running processes managed by `master`.

The existing SERVICES processes register themselves using the API functions provided by imap/proc.c.  EVENT and DAEMON processes are expected to often be external scripts, so their registration is taken care of directly by master.  This required relocating the proc.c API functions from libcyrus_imap to libcyrus_min (so that master can call them), and refactoring to make them re-entrant (because master needs to register its many children, not itself).

Some Cassandane internals needed a small refactor to make this testable, and I found and fixed a bug in our cunit test logging too.

Processes from the START section are not registered, and thus not listed by `cyr_info proc`.  Since these are by definition short-lived processes that only run once during startup -- and if they fail, startup fails -- it didn't seem worth it.  I could be persuaded otherwise.


ksmurchison (approved):
LGTM

PR #4524, opened 2023-06-06
jmap_calendar: support FilterOperator in dav.db queries

rsto:
CalendarEvent/query filters using any of the AND/OR/NOT operators were not supported using dav.db. This now supports such queries, allowing both for faster lookup and for previously unsupported queries such as OR(uid1,uid2,uid2).


ksmurchison (approved):
LGTM

rsto:
Now updated with support to

- optimize for OR(uid, uid, ..) queries
- return a distinct sqldb error if SQL errors are too large
- return `unsupportedFilter` for too large queries

@ksmurchison could you please review the latest three commits, especially the sqldb changes?

ksmurchison (approved):
LGTM

Issue #4528, opened 2023-06-09
If user's IMAP identity doesn't specify domain, allow using a default domain


pprindeville:
I have a small server with a handful of accounts hosted on it.  I use SASL for authentication and the users don't have accounts on this machine.  I don't use virtual domains because there's no point.  Simple credentials based on `username` are sufficient.

The SASL database is used to authenticate multiple services, so I for the realm I don't use a virtual domain but instead a service category (`mail`, `http`, etc).

Things worked as expected when the hostname of the server was `mail` (unqualified).

I migrated the service to a new server called `mail2` and authentication broke.

I looked through the code, and figured out that in the absence of a realm copied from the domain in the user's credential, that the serverFQDN was being used instead in [sasldb_auxprop_lookup()](https://github.com/cyrusimap/cyrus-sasl/blob/master/plugins/sasldb.c#L87-L91).

That seems like policy decisions being made in a low-level library, which I don't agree with, but I'm not going to address that here.

Examining the IMAP code, it seems there are hooks for a default-realm, it's just never used (the parameter gets passed as `NULL`).

elliefm:
Are you certain you don't need to just set the existing "defaultdomain" option appropriately?

pprindeville:
> Are you certain you don't need to just set the existing "defaultdomain" option appropriately?

`defaultdomain` only works when `virtdomains` is also enabled, and that brings with it a bunch of undesirable side-effects.

But I tried it with:

```
virtdomains: on
defaultdomain: mail
```

and I sill see:

```
Jun 14 10:49:21 mail cyrus/imaps[36407]: client id sessionid=<cyrus-1686761361-36407-1-4953568428733999007>: "name" "Mac OS X Mail" "version" "16.0 (3731.600.7)" "os" "Mac OS X" "os-version" "13.4 (22F66)" "vendor" "Apple Inc."
Jun 14 10:49:21 mail cyrus/imaps[36407]: badlogin: macbook3-3.redfish-solutions.com [192.168.8.12] CRAM-MD5 (-notset-) [SASL(-13): user not found: user: joeblow@redfish-solutions.com property: cmusaslsecretCRAM-MD5 not found in /etc/sasl2/sasldb2]
[...]

PR #4529, opened 2023-06-09
imap: support trivial credentials via default domain

pprindeville:
Addresses issue #4528


elliefm (changes requested):
One comment inline.  Sorry I'm not qualified to review this fully; I don't know anything about SASL realms.

Looks like the CI tests got choked up, so something in here isn't right.  But it's hard to tell what, since they weren't able to run to completion and report their results.

PR #4533, opened 2023-06-14
cyr_expire: add noexpire_until annotation

rsto:
This PR includes fixes and enhancements for cyr_expire:

- It fixes a bug in which the `expire/delete/archive` mailbox annotations were only read for a given mailbox, rather than looking its value up in all the parent mailboxes.
- It removes a noisy and unhelpful debug message from logging
- It adds support for non-day durations in the `expire/delete/archive` annotations by use of the Cyrus-config supported duration format. cyr_expire already had included support to read such values, but they could only be set as command line flags.
- It adds the `noexpire_until` annotation that allows to block expiring and deleting until a given future date, or indefinitely.

This PR best is read commit by commit.

### DIFF
--- a/docsrc/imap/reference/manpages/systemcommands/cyr_expire.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/cyr_expire.rst
@@ -43,8 +43,9 @@ There are various annotations that **cyr_expire** respects:
 - ``/vendor/cmu/cyrus-imapd/delete`` which controls the deletion of
   messages
 
-These mailbox annotations specify the age(in days) of messages in the
-given mailbox that should be expired/archived/deleted.
+These mailbox annotations specify the age of messages in the
+given mailbox that should be expired/archived/deleted. The
+age is specified as a duration, the default unit are days.
 
 The value of the ``/vendor/cmu/cyrus-imapd/expire`` annotation is
 inherited by all children of the mailbox on which it is set, so an
--- a/docsrc/imap/reference/manpages/systemcommands/cyr_expire.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/cyr_expire.rst
@@ -48,6 +48,7 @@ There are various annotations that **cyr_expire** respects:
 The first three mailbox annotations specify the age of messages in the
 given mailbox that should be expired/archived/deleted. The
 age is specified as a duration, the default unit are days.
+The duration format is defined in :cyrusman:`imapd.conf(5)`.
 
 The last mailbox annotation specifies the UNIX epoch time in seconds
 until which expiring messages or removing deleted mailboxes is blocked.
@@ -101,6 +102,8 @@ Options
     ``archivepartition-*`` has been set in your config.
     This value is only used for entries which do not have a
     corresponding ``/vendonr/cmu/cyrus-imapd/archive`` mailbox annotation.
+    The duration format is defined in :cyrusman:`imapd.conf(5)`. The default
+    unit are days.
 
     |v3-new-feature|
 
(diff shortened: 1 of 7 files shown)


ksmurchison (approved):
LGTM

elliefm:
CI shows this failing these two tests:

```
[FAILED] Cyrus::Metadata.private
[FAILED] Cyrus::Metadata.shared
```

docsrc/imap/reference/manpages/systemcommands/cyr_expire.rst
```diff
@@ -42,9 +42,18 @@ There are various annotations that **cyr_expire** respects:
...
   messages
-
-These mailbox annotations specify the age(in days) of messages in the
-given mailbox that should be expired/archived/deleted.
+- ``/vendor/cmu/cyrus-imapd/noexpire_until`` which disables the expire
+  and delete operations per user
+
+The first three mailbox annotations specify the age of messages in the
+given mailbox that should be expired/archived/deleted. The
+age is specified as a duration, the default unit are days.
```
elliefm, line 50: Need to describe the valid annotation durations here (or refer them to imapd.conf(5), which already describes them), and also need to be explicit that these are _different from the durations supported in the command line arguments_ (which are described under the heading for `-D, --delete-duration`).  

The command line durations are parsed by `parse_duration()`, a static function in cyr_expire.c, which allows and understands decimal places (e.g. `1.5d`) for whatever reason, though it reduces to integer seconds in the end.  `config_parseduration()` doesn't accept decimals because we didn't think it necessary -- the same thing can be expressed as `36h` or `1d12h`, for example (the latter of which `parse_duration()` can't handle).  So as presently implemented, valid command line durations and valid annotation/config durations are superficially similar, but aren't strictly interchangeable.

An approach here might be to just get rid of `parse_duration()`, and switch to using `config_parseduration()` for the command-line options too.  We'd need to document the change loudly when we release 3.10, in case anyone had been relying on the decimal parsing, but at least then all our "durations" will be consistent (unless there's others hiding somewhere...).  If we decide decimals are important enough to keep after all, we could probably extend `config_parseduration()` to support them.  (Though that could get messy... what should `1.5h12m` do?)

The description in imapd.conf(5) comes from here:  https://github.com/cyrusimap/cyrus-imapd/blob/master/lib/imapoptions#L87-L94.  But you can't just copy and paste that into an RST document, it's in the wrong format (for silly historical reasons).  If you want to copy and paste this text, first `make man` so that `docsrc/imap/reference/manpages/configs/imapd.conf.rst` gets generated, then copy and paste from that.

The RST syntax for linking to one of our man pages is:

```
:cyrusman:`imapd.conf(5)`
```

elliefm (changes requested):
Two failing tests under CI, though they might just be stale tests following these changes.

Of greater concern to me is the syntax incompatibility within cyr_expire between durations as received from the command line vs durations read from annotations.  This needs to be carefully documented, or perhaps homogenised... and then carefully documented.  Details within. :)

imap/cyr_expire.c
```diff
@@ -282,43 +283,114 @@ static int parse_duration(const char *s, int *secondsp)
...
+            mbname_free(&mbname);
+            mbname = tmp;
+        }
+        mbname_truncate_boxes(mbname, 0);
+
+        // Lookup annotation
+        time_t until;
+        if (get_time_annotation(mbname_intname(mbname),
+                    IMAP_ANNOT_NS "noexpire_until", &until, false))
+            ret = !until || progtime < until;
```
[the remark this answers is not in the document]
rsto: These conditionals should be fine. Quoting [GNU libc documentation](https://www.gnu.org/software/libc/manual/html_node/Time-Types.html):

> On POSIX-conformant systems, time_t is an integer type and its values represent the number of seconds elapsed since the epoch, which is 00:00:00 on January 1, 1970, Coordinated Universal Time. 

Also, the annotation value type in annotate.c is defined be an unsigned integer, so any negative value will be rejected on write.

I'm not sure if there's a need to break up the lines, still?

docsrc/imap/reference/manpages/systemcommands/cyr_expire.rst
rsto: Argh, thanks for this. I was totally not aware of this.

elliefm: cyr_expire is a _mess_ and it feels everything we try to clean up in it shakes out more chaos 😅

imap/cyr_expire.c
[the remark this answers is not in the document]
elliefm: In that case it might be worth putting a comment there, so the next reader doesn't need to wrack their brain wondering whether there's a bad edge case for pre-1970 dates (negative time_t)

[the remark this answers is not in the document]
rsto: Fixed with https://github.com/cyrusimap/cyrus-imapd/pull/4533/commits/e9ec21a79394289d7280ed7d657c810b224eaefb#diff-e2b440985f7fec0ccf9584ee3e7606debbfd031652c5e1188af23d7079091659R380-R384

docsrc/imap/reference/manpages/systemcommands/cyr_expire.rst
rsto: I now replaced custom cyr_expire durations with config_parseduration: https://github.com/cyrusimap/cyrus-imapd/pull/4533/commits/cf1043174822d21a130ec9fbde9dd6c679c20e11

elliefm (commented):
That looks good, thanks.

The two Metadata tests are still failing in CI.  I haven't looked closely at them; I assume they're probably just out of date wrt these changes and need to be updated, but they might be complaining about something real.  They failed for me locally last time I reviewed, and I assume they would do the same again, but I haven't actually pulled the branch and built/tested it this time.

rsto:
> The two Metadata tests are still failing in CI. 

Fixed now.

elliefm (approved):

PR #4534, opened 2023-06-20
libconfig: reject durations longer than INT32_MAX seconds

rsto:
We currently use these just for parsing imapd.conf but also want to use it for annotation values.

### DIFF
--- a/lib/libconfig.c
+++ b/lib/libconfig.c
@@ -196,8 +196,8 @@ EXPORTED int config_parseduration(const char *str, int defunit, int *out_duratio
 
     const size_t len = strlen(str);
     const char *p;
-    int accum = 0, duration = 0, neg = 0, sawdigit = 0, r = 0;
-    char *copy = NULL;
+    int duration = 0, neg = 0, r = 0;
+    char *copy = NULL, *end;
 
     /* the default default unit is seconds */
     if (!defunit) defunit = 's';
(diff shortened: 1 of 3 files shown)


lib/libconfig.c
```diff
@@ -211,53 +211,65 @@ EXPORTED int config_parseduration(const char *str, int defunit, int *out_duratio
...
+            val *= 24;
+            /* fall through */
+        case 'h':
+            val *= 60;
+            /* fall through */
+        case 'm':
+            val *= 60;
+            /* fall through */
+        case 's':
+            if (duration + val > INT32_MAX) {
```
elliefm, line 244: I can see that you used a similar implementation to cyr_expire's `parse_duration()`, letting `strtoul()` handle the digit parsing instead of accumulating them one at a time.  But this approach only works here because the type being used is `int`, which is smaller than `uint64_t`.  If we did the same thing in `config_parsebytesize()`, which computes an `int64_t` value, the overflow check would itself overflow -- so we couldn't use the same strategy in that one, and I think keeping the strategies similar is important for maintainability.

I was anticipating implementing these fixes similar to `getint32()` in imap/imapparse.c -- which each time it sees a new digit, it checks the current value, and if it's already too big to accept an additional digit, then it bails out.  For `config_parseduration()` and `config_parsebytesize()` it would be a little more complicated than that, because those functions don't only multiply by ten.  But the point is it's important to detect the potential overflow _before_ it happens, because we can't necessarily detect it after.

elliefm (approved):
We chatted about this out of band... I want to rewrite this so that config_parseduration and config_parsebytesize can use the same overflow strategy, which this implementation doesn't allow.

In the meantime, the tests show that it does at least provide the overflow protection we want in this case, and we want to deploy it as-is at Fastmail so that we can safely deploy #4533.

So I'm approving this so that it can be marked include-in-fastmail and do-not-merge, and we'll re-do this fix in a more general way soon.

PR #4535, opened 2023-06-23
Processimip rrule complete

brong:
Hi Ken, as promised, here's the test case and also a really basic possible solution to it, just to double check my logic.


ksmurchison (approved):

ksmurchison:
@brong As requested, I updated this to NOT run past the end of the recurrence set

brong:
hah oops, I made basically the same changes and force pushed them on top too, since I still had it checked out.

PR #4539, opened 2023-06-29
master: free memory before early return

rsto:


ksmurchison (approved):

elliefm:
Oh, good find.  Looks like that leak has been there for ages, but it was code that Cassandane couldn't test until this week, but now it does.

Issue #4532, opened 2023-06-14
autocreate perhaps doesn't recognise legacy mailboxes; creates new ones on login


elliefm:
Originally reported here: https://github.com/cyrusimap/cyrus-imapd/issues/4035#issuecomment-1586303928

elliefm:
Skimming over autocreate.c and imapd.c, if there's a bug of this shape here it isn't obvious on a code read.

`autocreate_inbox()` (in imapd.c) hasn't changed significantly in years, and maybe the bug is that it needs an update it hasn't received, but what's there appears to be sensible (that is: only try to autocreate if it specifically got IMAP_MAILBOX_NONEXISTANT back from a lookup).  Perhaps it's doing the lookup wrong, in a way that compiles but fails?

We only have positive tests for autocreate, that is, we only have tests that set up an environment where we expect autocreate to do something, and then check that it does.  We don't have any tests that set up an environment where we expect autocreate to do nothing, and check that it does nothing.  We probably should.

elliefm:
Hmm... put together a cassandane test to try to reproduce this, but it doesn't seem to occur (unless I'm testing it wrong, which is still highly likely at this point)

elliefm:
WIP: https://github.com/cyrusimap/cyrus-imapd/compare/master...elliefm:cyrus-imapd:v39/4532-autocreate-uuid-mailboxes

xkl:
corresponding report in debian bts: [1037346](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1037346)

elliefm:
As per #4540, the suspected autocreate problem does not appear to exist, unless it only exists in some configuration or circumstance that the test missed.

If you can show it happening and help us reproduce it, we'll be able to fix it, but as far as we can see at the moment, there isn't a problem with autocreate.

elliefm:
> When the services are restarted after an upgrade, if the MTA attempts a delivery before the cyrus is started up

How can this happen?  If Cyrus isn't fully started up yet, there should be no lmtpd running to accept the message and invoke autocreate.  Unless there was a stale lmtpd process still hanging around from the previous run which somehow hadn't shut down properly, and the MTA was talking to that?  But that doesn't make any sense either, because a stale lmtpd process would still be running the not-yet-upgraded code, and would not be confused by the not-yet-upgraded mailboxes.db format.  Unless it raced against the upgrade, I suppose?  Hmm

PR #4540, opened 2023-07-03
cass: test that autocreate doesn't replace legacy mailboxes

elliefm:
This adds a test to verify that autocreate doesn't replace legacy mailboxes, which seems to show that the problem I suspected in #4532 does not exist.


rsto (approved):

elliefm:
This can't be easily backported because it depends on Cassandane internal changes from #4522

PR #4542, opened 2023-07-03
Allow cyrus to compile Sphinx documentation with Sphinx version 4..0 …

lol-catz:
…or greater

When compiling cyrus-imap 3.8.0 on debian 12 with Sphinx documentation  cyrus  fails to compile with the following  error:

Exception occurred:
|   File "/usr/lib/python3/dist-packages/docutils/writers/manpage.py", line 171, in __init__
|     self.settings = settings = document.settings
| AttributeError: 'CyrusManualPageBuilder' object has no attribute 'settings'

Cause: 

CyrusManualPageTranslator inherits from Sphinx' ManualPageTranslator and calls the super-class constructor as follows:

 BaseTranslator.__init__(self, builder, *args, **kwds)

However, starting with Sphinx 2.0 it should be called with document as the first argument and builder as the second one. The old way of calling it raised a warning for some time and was completely removed in Sphinx 4.0.

### DIFF
--- a/docsrc/exts/sphinxlocal/writers/manpage.py
+++ b/docsrc/exts/sphinxlocal/writers/manpage.py
@@ -38,7 +38,7 @@ class CyrusManualPageWriter(ManualPageWriter):
         self.builder = builder
 
     def translate(self):
-        visitor = CyrusManualPageTranslator(self.builder, self.document)
+        visitor = CyrusManualPageTranslator(self.document, self.builder)
         self.visitor = visitor
         self.document.walkabout(visitor)
         self.output = visitor.astext()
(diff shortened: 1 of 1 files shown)


elliefm:
Thanks!

From what I understand from the PR description, the fix is only backward compatible as far as sphinx-build 2.0.  I've just checked the system that automatically rebuilds the cyrusimap.org website, and it's currently running sphinx-build 1.4.9 -- so if we merge this PR at the moment, the website will cease to update.

We ought to get this merged, but we'll need to update sphinx-build on the system that rebuilds the website first!

elliefm:
Apparently, even this fix is broken for sphinx 5.0 😅 

The custom CyrusManualPageBuilder/CyrusManualPageTranslator stuff didn't seem to get us anything but maintenance headaches, so we've retired it as of #4561, and we're just using the built-in man stuff instead.  So I think this PR is now obsolete -- thanks, though! :)

PR #4544, opened 2023-07-05
fixes for leaks found by valgrinding cunit

elliefm:
This fixes a handful of leaks that `make -j4 VG=1 check` detects.  Most of these are trivial but there's a couple of complicated ones:

* search_expr.testc was leaking internal state that's usually cleaned up by `cyrus_done()`, but which is missed in the tests because the way `set_up()` initialises it bypasses installing the done callback.  To fix this one, I tweaked some things to expose search_expr.c's internals to cunit, so that `tear_down()` could then reach in and clean up itself.  But perhaps it would have been better to just add a comment and a suppression rule instead?
* backend.c was leaking an `SSL_SESSION` object when disconnecting from a backend that we had used starttls when connecting to.  This one's a real leak in Cyrus, not just a leak in the tests.  The fix seems simple, and our tests all pass, but I'm not totally certain it's correct.  When reviewing this one, please look around, not just at the diff.  @ksmurchison recently fixed a similar SSL_SESSION leak in a3523d406.

Please pull this branch and valgrind the tests yourself if you can -- I expect there to be a bunch of "still reachable" leaks, but every time valgrind says anything, it should also say `ERROR SUMMARY: 0 errors from 0 contexts`, indicating that there's not anything that really needs fixing.  If your valgrind finds things mine doesn't, that would be useful to know about!

If you think I should dive in and track down the "still reachable" ones too, please say so -- I'm on the fence myself.  On the one hand, valgrind doesn't consider them to be errors; on the other hand, they're noise that makes real errors harder to see.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -674,7 +674,7 @@ cunit_TESTS += cunit/ical_support.testc
 endif
 
 cunit_unit_SOURCES = $(cunit_FRAMEWORK) $(cunit_TESTS) \
-		imap/mutex_fake.c imap/spool.c
+		imap/mutex_fake.c imap/spool.c imap/search_expr.c
 cunit_unit_LDADD = $(LD_SIEVE_ADD) $(LD_UTILITY_ADD) -lcunit
 
 CUNIT_PL = $(top_srcdir)/cunit/cunit.pl --project $(CUNIT_PROJECT)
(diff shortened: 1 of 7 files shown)


elliefm:
Marked for backporting because of the backend.c leak fix

ksmurchison (approved):

elliefm:
Skipped the search_expr.test commit for 3.6.  It cherry-picked cleanly but didn't compile, looks like search_expr internals probably changed at some point.

Issue #3861, opened 2022-01-10
cyrus-imapd: depends on obsolete pcre library


guimard:
Ref: https://bugs.debian.org/1000034

Hi,

when using pcre2, all test passed except this one:
```
  TESTCASE("re: no re: foobar", "nore:foobar");
```
which gives:
```
  Test: subject_normalise ...FAILED
    1. ./cunit/conversations.testc:1353  -
       CU_ASSERT_EQUAL(b.len=14,sizeof(_exp)-1=11)
    2. ./cunit/conversations.testc:1353  -
       CU_ASSERT_STRING_EQUAL(b.s="re:nore:foobar",_exp="nore:foobar")
```
Cheers,
Yadd

elliefm:
This is deeply confusing but I think I have finally understood...

* The OLD "pcre" is called "pcre3" for some reason
* The NEW "pcre" is called "pcre2"
* Cyrus works fine with the old pcre
* Debian want to remove the old one, and only package the new one, but this will require Cyrus to start supporting the new one
* Cyrus builds fine with the new one, but there is at least one test failure when built like this -- and maybe other problems we've yet to discover for lack of a test...

Is that a correct summation of the situation?

There's more detail in the linked Debian bug, but I want to make sure I'm clear on the basics before I try reading any deeper

guimard:
Hi @elliefm, thanks for the translation and sorry for this : I'm not fluent in English. You are totally right here

guimard:
Sure, sorry. Is this just a test issue or a core issue ?

elliefm:
No idea yet.  But I think it's reasonable to assume that if our tests are doing something the new library doesn't like, then our real code is probably also doing it and will also break.

elliefm:
@brong does this sound like the same issue you mentioned on the Cyrus conf call the other day?

robert-scheck:
Just as a heads up: Fedora also tries to get rid of the old pcre, which is meanwhile even end-of-life at upstream, in favor of pcre2; see also: https://bugzilla.redhat.com/show_bug.cgi?id=2128286

ArchangeGabriel:
Same for ArchLinux: https://bugs.archlinux.org/task/77058#comment214212
Building with the pcre2 patch in this comment leads to the same test failure.

loqs:
```patch
diff --git a/configure.ac b/configure.ac
index 8fe7d78..ac601a3 100644
--- a/configure.ac
[...]

PR #4545, opened 2023-07-11
Update to pcre2

yselkowitz:
Classic pcre is unmaintained and being dropped from distributions.

Fixes: #3861


yselkowitz:
v2: implement idempotent guards for `<pcre2posix.h>` to compile with current versions; reported upstream as https://github.com/PCRE2Project/pcre2/pull/274

elliefm:
I've extended your patch to support pcre2 while also maintaining support for legacy pcre.  It's now in #4736.  Thanks!

PR #4547, opened 2023-07-14
cassandane: add slurp_file() function

elliefm:
... and dedup the many many times we've hand-rolled (or copy-pasted) this

**reviewers:** this has lots of small, similar (but not always identical) changes, might be easiest to review commit-by-commit


rsto (approved):
LGTM. This will come in handy! I ran all Cassandane tests successfully.

ksmurchison (approved):

PR #4548, opened 2023-07-17
cassandane: improve file test assertions

elliefm:
This adds two new assertion functions to Cassandane: `assert_file_test()` and `assert_not_file_test()`, for assertions based on the `perldoc -f -X` filename tests (i.e. `-e`, `-f`, `-x`, etc) with helpful messages, and then updates a bunch of our tests to use them.

There's a handful of tests that consistently fail when the local month is different from the UTC month, and some of those were using bare file test assertions without descriptive messages, making it hard to track down the cause of the failures.  This PR makes it so those tests will fail with better messages, so that next time I might finally be able to track down and fix whatever is actually causing the failure.

I started working on the test failures as #4500, but this chunk stands on its own, so I'm splitting it out to get it landed early.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Archive.pm
+++ b/cassandane/Cassandane/Cyrus/Archive.pm
@@ -104,36 +104,35 @@ sub test_archive_messages
     my $datadir = $data->{data};
     my $archivedir = $data->{archive};
 
-    $self->assert(-f "$datadir/1.");
-    $self->assert(-f "$datadir/2.");
-    $self->assert(-f "$datadir/3.");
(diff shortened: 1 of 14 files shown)


rsto (approved):

elliefm:
Not backporting this any further, too many conflicts

elliefm:
Okay I am backporting this to 3.6 after all, because #4558 needs it.  On closer inspection, the conflicts were mostly just where changes had been made to tests that 3.6 doesn't have, which are easy enough to deal with.

PR #4550, opened 2023-07-20
configure.ac: detect xxd, error if needed and missing

elliefm:
We weren't checking whether `xxd` existed, just crossing our fingers and hoping.  If it wasn't available, `configure` would still succeed, but then `make` would fail.  This PR fixes that.

The files we produce with xxd are included pre-generated in our release tarballs, so someone building from a release tarball could have technically not required xxd (as long as they didn't do anything that would cause those files to need to be rebuilt).  I initially tried a more complicated implementation so that such people wouldn't be forced to install xxd, but it was messy and didn't offer much real improvement, so I stripped it back out again.  It's not a difficult dependency to meet.


rsto (approved):
I removed `xxd` from my system for testing. Works as intended.

PR #4552, opened 2023-07-25
Fix IMAP SCAN on selected mailbox crasher

ksmurchison:
imapd.c: needs an open mailbox for index_scan(), imapd_index is released before every command is parsed, so we need to reopen the mailbox

### DIFF
--- a/cassandane/Cassandane/Cyrus/List.pm
+++ b/cassandane/Cassandane/Cyrus/List.pm
@@ -2343,7 +2343,6 @@ sub test_scan
 
     xlog $self, "listing...";
     my $res = $imaptalk->list("", "*");
-warn Dumper($res);
 
     $self->assert_mailbox_structure($res, '.', {
         'INBOX'      => [qw( \\HasChildren )],
--- a/cassandane/Cassandane/Cyrus/List.pm
+++ b/cassandane/Cassandane/Cyrus/List.pm
@@ -2329,4 +2329,55 @@ sub test_list_return_subscribed_trash_nochildren
     });
 }
 
+sub test_scan
+    :min_version_3_9
+{
+    my ($self) = @_;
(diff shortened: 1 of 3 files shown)


elliefm (changes requested):
The fix looks fine on a code read, but can you please add a test to prove it (and to prevent us breaking it again in the future)?  Thanks

rsto (changes requested):
Agree with @elliefm, code looks good but please add a test.

ksmurchison:
Actually, SCAN was broken worse than just the crasher.  It always returned nothing due to a logic error in the ACL test.  And then it returned EXISTS, RECENT, and FLAGS responses fro every mailbox that is scanned.  All issues have been fixed and tested - SCAN with/without a previously SELECTed mailbox

elliefm (approved):
Looks good, thanks

rsto (approved):

rsto:
I fixed a memory leak in the new code and pushed an update.

cassandane/Cassandane/Cyrus/List.pm
```diff
@@ -2329,4 +2329,55 @@ sub test_list_return_subscribed_trash_nochildren
...
+    my $store = $self->{store};
+    my $imaptalk = $store->get_client();
+
+    $self->setup_mailbox_structure($imaptalk, [
+        [ 'create' => [qw( INBOX.a INBOX.b INBOX.c )] ],
+    ]);
+
+    xlog $self, "listing...";
+    my $res = $imaptalk->list("", "*");
+warn Dumper($res);
```
elliefm, line 2346: can you remove this warning though please :)

rsto: done

PR #4553, opened 2023-07-26
github actions: report cyrus version

elliefm:
What it says on the tin.  Requested by @wolfsage


elliefm:
@wolfsage,

> and any other useful info

I can't think of any other detail I'm yearning for, so I've left it at just the cyrus version for now, but lmk if there's other details you had in mind.

wolfsage (commented):
Awesome, thanks

wolfsage (approved):

wolfsage:
> @wolfsage,
> 
> > and any other useful info
> 
> I can't think of any other detail I'm yearning for, so I've left it at just the cyrus version for now, but lmk if there's other details you had in mind.

I can't either. Ah well!

PR #4555, opened 2023-07-26
doc generation: stop using custom manpage builder

rjbs:
Why is the current one broken with newer libraries?

I don't know!  But we also no longer know why we had a customer builder, either.  At first glance, the output without the custom code seems fine. We should inspect it more, and then we can decide if we can go forward with this and also delete the custom code!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1978,7 +1978,7 @@ clean-docsrc:
 
 ## XXX doesn't detect if other rst sources are updated...
 man/.sphinx-build.stamp: docsrc/.sphinx-build.stamp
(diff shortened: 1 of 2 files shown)


ksmurchison:
This solves the problem for me.

elliefm:
The only differences I'm seeing between the man pages generated by the custom formatter and those generated by the stock formatter are some indentation levels and the date format.  I don't think these difference are worth continuing to carry around the custom formatter for.

ksmurchison:
@elliefm I agree

elliefm:
This landed as part of #4561.  Thanks!

PR #4557, opened 2023-07-27
ical_support: mark multiple standalone overrides as recurring

rsto:


ksmurchison (approved):
LGTM.  Glad I was correct on the source of the problem and the fix was simple

PR #4558, opened 2023-08-04
better cyrusdb / ctl_cyrusdb -r UX

elliefm:
**Reviewers:** this has grown into a collection of small related things rather than a single thing.  Probably easiest to review commit-by-commit.

If `ctl_cyrusdb -r` was not run during Cyrus startup (or failed in a particular way, see further on), then your logs will be full of errors like this from every single Cyrus process:

> DBERROR: read failed, assuming the worst: filename=</path/to/confdir/db/skipstamp> [...]

Would you be able to guess from that that the problem was that you'd forgotten to add `ctl_cyrusdb -r` to your cyrus.conf START section?  Me either!

This PR adjusts the logging such that if the reason the skipstamp file is unreadable is that it doesn't exist, then an informational message is also logged advising the need for `ctl_cyrusdb -r`.  I've kept the new log message concise, but hopefully mentioning ctl_cyrusdb is enough to prompt someone to read the man page and learn that they need to add it to their startup.  I've also tidied up the errno handling so as to not leak noise into xsyslog later.

I've also added a test that confirms the behaviour, and logging about, the missing skipstamp file.  Next time someone asks about these errors, it'll be easy to see from the test how it's supposed to work, rather than needing to read the source.

I've also updated the handling of the "skipcleanshutdown" file.  When this file is present,  `ctl_cyrusdb -r` did not regenerate the skipstamp file -- which, if it somehow didn't already exist, could lead to it never being created.  It now also checks for the existence of the skipstamp file, and if it's missing, then skipcleanshutdown is ignored and skipstamp is regenerated as usual.  And there's a new test for this behaviour too.

In the end, it turned out the person complaining about all the "assuming the worst" errors had forgotten or not realised that they need to manually create configdir/db when provisioning a system (and so had I, and the log line that reports the missing directory only happens once when `ctl_cyrusdb -r` runs at startup, and is drowned out by the spew of "assuming the worst" errors).  Manually creating this directory seems like a ridiculous requirement when a computer could do such a trivial task itself, so I've made cyrusdb just create the damn directory itself if it's missing, and made Cassandane and cunit _not_ pre-create it (so if cyrusdb ever starts failing to create it, lots of tests will fail due to syslog errors from it not being there).

### DIFF
--- a/lib/cyrusdb_skiplist.c
+++ b/lib/cyrusdb_skiplist.c
@@ -250,7 +250,7 @@ static int myinit(const char *dbdir, int myflags)
 
         if (r != -1) r = ftruncate(fd, 0);
         net32_time = htonl(global_recovery);
-        if (r != -1) r = write(fd, &net32_time, 4);
+        if (r != -1) r = retry_write(fd, &net32_time, 4);
         xclose(fd);
 
         if (r == -1) {
@@ -262,19 +262,31 @@ static int myinit(const char *dbdir, int myflags)
     } else {
 normal:
         /* read the global recovery timestamp */
+        errno = 0;
 
-        fd = open(sfile, O_RDONLY, 0644);
-        if (fd == -1) r = -1;
-        if (r != -1) r = read(fd, &net32_time, 4);
-        xclose(fd);
+        r = fd = open(sfile, O_RDONLY, 0644);
+        if (r == -1 && errno == ENOENT) {
+            /* tell the admin what they need to do! */
+            xsyslog(LOG_INFO, "skipstamp is missing, have you run `ctl_cyrusdb -r`?",
+                              "filename=<%s>", sfile);
+        }
+
+        if (r != -1) r = retry_read(fd, &net32_time, 4);
 
         if (r == -1) {
-            xsyslog(LOG_ERR, "DBERROR: read failed, assuming the worst",
+            xsyslog(LOG_ERR, "DBERROR: skipstamp unreadable, assuming the worst",
                              "filename=<%s>", sfile);
+            /* "assuming the worst" means recovery will run for every skiplist
+             * database every time it's opened, because we can't determine that
+             * it doesn't need it.
+             */
             global_recovery = 0;
         } else {
             global_recovery = ntohl(net32_time);
         }
+
+        xclose(fd);
+        errno = 0;
     }
 
     srand(time(NULL) * getpid());
--- a/lib/cyrusdb_skiplist.c
+++ b/lib/cyrusdb_skiplist.c
@@ -230,14 +230,21 @@ static int myinit(const char *dbdir, int myflags)
         struct stat sbuf;
         char cleanfile[1024];
 
+        /* n.b. nothing in cyrus creates this file, but presumably some
+         * init.d or systemd script does
+         */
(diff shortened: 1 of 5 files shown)


elliefm:
Got another related thing to fix/test, and I'll do it in this PR since it's in the same section of code, so back to draft while I work on that.

elliefm:
Fixed handling of skipcleanshutdown file, and edited original description.

elliefm:
I've got another small QoL improvement to make in the same code, so back to draft again while I do that

elliefm:
Updated cyrusdb_init to create the directory before telling the backends to use it, so that nobody ever needs to remember to manually create it again.  Original description updated.

lib/cyrusdb_skiplist.c
```diff
@@ -250,7 +250,7 @@ static int myinit(const char *dbdir, int myflags)
 
         if (r != -1) r = ftruncate(fd, 0);
         net32_time = htonl(global_recovery);
-        if (r != -1) r = write(fd, &net32_time, 4);
+        if (r != -1) r = retry_write(fd, &net32_time, 4);
```
rsto, line 260: I am going to approve this PR but I find use of the `r` variable in this function quite confusing.  I know you just built on what's already there (except for storing a file descriptor in `r`), but maybe you are up to clean that variable up, too.

Some points I find confusing:

- typically, we use `r` as the return value for a function. But this function hardcodes its return values to either be 0 or CYRUSDB_IOERROR, so `r` might already be a misnomer.
- for the read() and retry_write() calls, `r` is meant to store a `ssize_t`, whereas we know that exactly 4 bytes must be read or written or otherwise it's an error
- `r` now also stores the file descriptor `fd` value, whereas the code could just as well check `fd !- -1` which might make it even more clear on what's the error condition.
- in case of ftruncate() `r` stores an (inversed) boolean, but at least we know that false is -1 here.

rsto (approved):
Works as designed. I find the init code of skiplist confusing, but that's a deficiency of the code this PR builds on.

lib/cyrusdb_skiplist.c
elliefm: I did think about and tinker with a lot of the things you raise, but they mostly wound up cluttering things up and making it harder, not easier, to read.

We do often use `r` like this, to keep track of an internal success or failure state but eventually returning some other success or failure value (or just void), so I don't think this use of the name `r` is unusual.  We do also use it the way you describe, and probably some other ways too.  I think of `r` as a place for storing the result of some function call (but not necessarily the eventual result of _this_ function call).  Sometimes we also return it up to our caller, but not always.

For what it's worth, `retry_read(..., n)` and `retry_write(..., n)` only ever return `-1` or `n`.  Internally, they repeat the syscall until they've successfully read or written the expected number of bytes.  If anything goes wrong, they return `-1`, _not_ the count-so-far.  When dealing with nontrivial files, that can make it tricky to reason about the current file offset.  But for the kind of thing we're doing here, the binary it-either-completely-succeeded-or-it-failed behaviour is the point, and is why I don't need to also check for `4`.

elliefm:
The tests for this depend on the `assert_file_test` stuff from #4548, which I had already decided not to backport to 3.6...

elliefm:
#4548 has been backported and now so has this one

Issue #3215, opened 2020-10-02
improved_mboxlist_sort: yes is the only sane setting


elliefm:
It's nonsensical to leave this option turned off, but it's off by default, and we keep it like that because changing it requires downing the server to rebuild mailboxes.db.  We're basically catering to people with old deployments who want to upgrade in place with minimal effort, but at the same time we're ensuring that nearly every new deployment ends up stuck with the same bogus default...

Nowadays we don't recommend upgrade-in-place for major releases anyway, instead we tell people to build a new server on the new version with their new configuration and then replicate their data to it.

We've already done big old-world-breaking default changes for unixhierarchysep and altnamespace in 3.0.0, and we plan to do a similar one for virtdomains (#2847), so let's also change this once and for all.

elliefm:
We should probably verify that a server with "improved_mboxlist_sort: no" can successfully replicate to one with "improved_mboxlist_sort: yes" before we go ahead, if that's how we're gonna tell people to upgrade...

elliefm:
Should also verify that the documentation for changing this setting is correct and complete... it lives [here](https://www.cyrusimap.org/imap/developer/thoughts/improved_mboxlist_sort.html), and also somewhat in imapd.conf.5

dilyanpalauzov:
Since `ctl_mboxlist -d -y` provides more output than `ctl_mboxlisd -d`, https://www.cyrusimap.org/imap/developer/thoughts/improved_mboxlist_sort.html shall explicitly state, whether -y shall be passed or not.

dilyanpalauzov:
Note: as far as I remember (see my mail on cyrus-devel from 5th October 2020), exporting the list with ctl_mboxlist -d and importing the list with ctl_mboxlist -u does not work, as suggested in the link above, as the fields `I`, `V` and `F` are missing after the conversion.

dilyanpalauzov:
I have [extended](https://github.com/cyrusimap/cyrus-imapd/pull/3384) ctl_mboxlist to dump/undump all fields.  Following the described procedure: dump and undump on a “dead system” has lead to mboxlist.db doubling its size.  https://www.cyrusimap.org/imap/developer/thoughts/improved_mboxlist_sort.html might need to say, that between the dump and undump mboxlist.db shall be deleted (renamed).

hagedose:
[...]

PR #4559, opened 2023-08-06
Remove improved_mboxlist_sort leftovers

dilyanpalauzov:
My understanding is that after 3.4 and 3.6 the option `improved_mboxlist_sort` was removed.  This removes some leftovers.


elliefm:
Interesting, I hadn't realised that had changed.  Looks like we stopped using it in 916b8f339, which was 3.6 as you say.

lib/imapoptions
```diff
@@ -1206,15 +1206,7 @@ Blank lines and lines beginning with ``#'' are ignored.
    is prepended. */
 
 { "improved_mboxlist_sort", 0, SWITCH, "2.3.17" }
```
elliefm, line 1208: Also need to mechanically mark the option as deprecated, like this:

```suggestion
{ "improved_mboxlist_sort", 0, SWITCH, "UNRELEASED", "UNRELEASED" }
```

The first occurrence of "UNRELEASED" here replaces the "2.3.17" as being the most recent version in which something about this option changed (the fact it is being deprecated is a change).  The second occurrence marks the option as being deprecated (as of the specified version).  The magic word "UNRELEASED" is how we refer to a release that doesn't yet exist without trying to predict what its version number will be.  It'll be replaced with a real version number when a real version is released.

cassandane/Cassandane/Cyrus/Quota.pm
```diff
@@ -1134,7 +1134,6 @@ sub test_quota_f_no_improved_mboxlist_sort
         'user2@example.com',
     );
 
-    $self->{instance}->{config}->set('improved_mboxlist_sort', 'no');
```
elliefm, line 1137: Same for `sub test_quota_f_no_improved_mboxlist_sort`, no special behaviour for it to test anymore, so we don't need the test.

cassandane/Cassandane/Cyrus/Prometheus.pm
```diff
@@ -263,7 +263,6 @@ sub test_quota_commitments_no_improved_mboxlist_sort
 {
     my ($self) = @_;
 
-    $self->{instance}->{config}->set('improved_mboxlist_sort', 'no');
```
elliefm, line 266: I think you can remove the whole `sub test_quota_commitments_no_improved_mboxlist_sort`, since there's no special behaviour left for it to test.

elliefm (dismissed):
Thanks!  This looks good.

I've requested one change to lib/imapoptions to properly mark the option as deprecated.

I've also identified a couple of cassandane tests that I think you could just remove entirely, but they're less important than getting lib/imapoptions right.

elliefm:
Note to self: marking this for backport to 3.6 and 3.8, but only the 3.6.0.rst should be backported.  The option will remain present-but-ignored in those releases, and will be formally deprecated in the next major release

elliefm:
This closes #3215

dilyanpalauzov:
I realized that without Improved-Mboxlist the option CYRUSDB_MBOXSORT is unused, and this permits deleting even more code.

In cunit/aaa-db.testc I do not know if after line 802 the code can be deleted until the end of the function.

elliefm:
I'm not comfortable with carving up the CyrusDB API so suddenly.  Maybe I'd prefer to just rename `CYRUSDB_MBOXSORT` to `CYRUSDB_DEPRECATED_MBOXSORT`, and add some code to `lib/cyrusdb.c:_myopen()` that logs and clears that bit if it's set before calling the backend's open function.  But not actually remove it outright until we're confident that that much didn't break things.

@brong: you probably have opinions here, let's hear em! :)

dilyanpalauzov:
Please split this changeset in two parts at your discretion.  Apply the one part now and the other later.

elliefm (approved):

elliefm:
Fixed up and merged manually as cec19a093

PR #4561, opened 2023-08-07
retire custom manpage generator

elliefm:
Previously: #4555

Switches our docs build to use the stock sphinx manpage generator, rather than the custom one that keeps getting broken by new sphinx releases (thanks @rjbs), and then removes the parts we're no longer using.

I think I've removed everything that needed removing, but it shouldn't hurt if I've missed stuff.

**Reviewers:** please test this -- it works fine for me, but does it work for you?

1) Assuming you usually have all the dependencies for docs generation, run configure as you usually would
2) `make man` -- should exit success.  There will be a bunch of warnings ("duplicate term", "glob pattern didn't match", "document isn't included in any doctree") but these can be ignored, unless they're different from what I've listed.
3) It should have created the man pages in the `man/` directory.  `ls -l man/* | wc -l` gives 84 for me.  Check that they're not empty!  Also check a couple of them to ensure they look okay: `man man/imapd.conf.5` etc.
4) Clean your repo however you usually would (I use `git clean -xfd`, though n.b. this will trash uncommitted files)
5) Change into the `docsrc` directory
6) Run `make man` in here -- you do not need to have run configure for this!
7) Expect slightly different terminal output, but same warnings as before
8) Generated man pages will be `docsrc/build/man`, check those too as per step 3.


rjbs (approved):
I didn't find any nasty traces!

PR #4562, opened 2023-08-07
caldav_alarm.c: don't skip other alarms if one has an absolute trigger

ksmurchison:
We need to continue to inspect other alarms even if we suppress an alarm with an absolute trigger.


elliefm (approved):
Looks good

Issue #4370, opened 2023-01-08
XFER between murder backends does not delete source mailbox in filesystem after successful transfer


a-tze:
This issue might be related to #2193

Version affected: 3.6.0-rc2 (Debian bullseye-backports package)

Environment: 
- cyrus murder consisting of 1 dedicated frontend server (cyrus-1) and 2 backends (cyrus-2, cyrus-3)
- test cluster installed from scratch, no migrated data

Actions performed:
- create mailbox user.test1 on cyrus-2
- xfer it to cyrus-3

Expected result:
- mailbox user.test1 resides on cyrus-3
- cyrus-2 has nothing left of user.test1 (besides tombstone if needed), or remains are removed with cyr_expire

Actual result:
- xfer works
- user.test1 is marked as deleted on cyrus-2
- But: mailbox directery is not scrubbed from filesystem on cyrus-2. It is not scrubbed by using cyr_expire. Behaviour is not affected by delete_mode or mailbox_legacy_dirs config settings. Manual scrubbing is difficult because even mbpath does not report mailbox details.

Notes:
mailbox.db on cyrus-2 has mbtype "edl" on user.test1 after xfer. Invoking ctl_mboxlist -v does not complain about user.test1 directory still being present.
After digging in the code for some time, I think this is what happens: after the first transfer, the mailbox gets the deleted-flag and after final transfer, when it should really be removed, the call into the mboxlist-functions fails because the mailbox is already marked deleted.

Detailled logs and information:

Source backend server log:

```
Jan  8 15:55:38 cyrus-2 cyrus/imap[937]: command: 5 Xfer
Jan  8 15:55:38 cyrus-2 cyrus/imap[937]: requested partition directory for unknown partition 'user.test1'
Jan  8 15:55:38 cyrus-2 cyrus/imap[937]: XFER: connecting to server 'cyrus-3'
Jan  8 15:55:38 cyrus-2 cyrus/imap[937]: auditlog: proxy  sessionid=<cyrus-1673189717-937-1-1756795977903769820> remote=<cyrus-1673189737-960-1-7936834340323250578>
Jan  8 15:55:38 cyrus-2 cyrus/imap[937]: XFER: destination supports replication
Jan  8 15:55:38 cyrus-2 cyrus/imap[937]: XFER: destination supports #sieve mailbox
Jan  8 15:55:38 cyrus-2 cyrus/imap[937]: ENABLE 0x800
[...]

PR #4563, opened 2023-08-08
imapd.c: make sure we delete XFERed mailboxes from source (issue #4370)

ksmurchison:
this commit also removes an extra write to mailboxes.db, since mboxlist_deletemailbox() already sets MBTYPE_DELETED

### DIFF
--- a/cassandane/Cassandane/Cyrus/MurderIMAP.pm
+++ b/cassandane/Cassandane/Cyrus/MurderIMAP.pm
@@ -53,6 +53,11 @@ $Data::Dumper::Sortkeys = 1;
 sub new
 {
     my $class = shift;
+
+    my $config = Cassandane::Config->default()->clone();
+
+    $config->set(conversations => 'yes');
+
     return $class->SUPER::new({
         imapmurder => 1, adminstore => 1, deliver => 1,
     }, @_);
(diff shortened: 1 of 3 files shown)


elliefm (commented):
Looks okay to me, and it's passing the tests...

That said, our XFER tests don't presently check whether the mailbox was actually removed from its original filesystem location after the XFER succeeds (if they did, we would've caught the bug a lot earlier).  It would be good to extend these tests to also check that, though off the top of my head I'm not sure how hard that would be.  Might need a `cyr_expire` run?

I don't think we need to add any new tests here, but making the ones we have look a bit deeper at the post-XFER system state would be good.

ksmurchison:
@elliefm I agree that testing that the dirs makes sense.  We have plenty of places where we check for mailbox/metadata dirs in Cass, but I don't think that the underlying run_mbpath() is setup to work with different instances/stores.

elliefm:
`run_mbpath()` is a method on Cassandane::Instance, so I think it should work fine as long as you call `$self->{foo}->run_mbpath(...)` on the correct `foo`.  That'll be `$self->{instance}` for the first/default backend, `$self->{backend2}` for the second backend, and `$self->{frontend}` for the frontend/mupdate server

At a quick glance, the xfer tests all seem to start with the mailboxes on `$self->{instance}` and xfer them to `$self->{backend2}`, so to make sure the source was cleaned up, you'd mostly be looking at `$self->{instance}`

elliefm:
Ditto for `cyr_expire` if it's needed -- `run_command()` is also a method on Cassandane::Instance, so as long as you `$self->{foo}->run_command(...)` on the right `foo` (as above), it'll run in the right place.

ksmurchison:
@elliefm Added a test which checks the directories

elliefm (approved):

a-tze:
@elliefm Is there any chance to backport this to 3.6/3.8?

elliefm:
@a-tze thanks for the prompt.  That seems sensible, and I'm not sure why I didn't already mark it for backporting.

I've now marked it for backporting, and if it is backportable, it'll be in the next releases.

PR #4570, opened 2023-08-13
3.6 ctl_mboxlist don't free static data 

dilyanpalauzov:
backport of 3be3f16999d724b44c1f009f2eb9c3fe06458825

### DIFF
--- a/imap/ctl_mboxlist.c
+++ b/imap/ctl_mboxlist.c
@@ -993,8 +993,6 @@ static void add_part(ptrarray_t *found,
             /* found it */
             if (override) {
                 /* replace the path with the one containing cyrus.header */
-                if (entry->path)
-                    free(entry->path);
                 strcpy(entry->path, path);
             }
 


dilyanpalauzov:
Why are the tests failing?

elliefm:
In this case, those two tests are failing because a bug in JMAPTestSuite was _fixed_.  But this PR is against cyrus-imapd-3.6, which doesn't have the corresponding Cyrus fix that JMAPTestSuite now exists.  Those tests will stop failing once I backport #4309 to the 3.6 branch.  I'll do that today since it's clearly getting in the way.

elliefm:
Hang on... I had already backported that to 3.6 (but hadn't removed the backport tag yet), so that's not it.

Ahh: 10a72d845 is on cyrus-imapd-3.6, but it's not on your PR branch, so you must have created your branch before it was backported.  To resolve it, you can pull the latest cyrus-imapd-3.6 from upstream, rebase your branch on top of it, then force push your branch.

PR #4578, opened 2023-08-17
conversations: sort thread by createdmodseq if same internaldate

brong:
This gives us a stable sort when multiple external messages are appended very quickly, giving them the same internaldate.

This is being added without a new version number, as it extends a data structure in a backwards and forwards compatible way.  If not set, the CREATEDMODSEQ field will be zero, which is fine.  This will still lead to correct sorting for newly added emails on any thread.  The only slight weirdness might be if an existing message in a thread gets moved or copied into a new folder, which will lead to it getting a createdmodseq value, and hence sorting to the end for that internaldate. This will be very rare, and a convdb rebuild will fix it.

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_get_createdmodseq
+++ b/cassandane/tiny-tests/JMAPEmail/email_get_createdmodseq
@@ -1,0 +1,51 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_get_createdmodseq
+    :min_version_3_9 :needs_component_jmap :JMAPExtensions
+{
+    my ($self) = @_;
+    my $jmap = $self->{jmap};
+    my $imap = $self->{store}->get_client();
+    $jmap->AddUsing('https://cyrusimap.org/ns/jmap/mail');
+
+    xlog $self, "Append duplicate messages";
+    $mimeMessage = <<'EOF';
+From: <from@local>
+To: to@local
+Subject: test
+Date: Mon, 13 Apr 2020 15:34:03 +0200
+MIME-Version: 1.0
+Content-Type: text/plain
+
+test
+EOF
+    $mimeMessage =~ s/\r?\n/\r\n/gs;
+    $imap->append('INBOX', '17-Aug-2023 15:13:54 +0200', $mimeMessage) || die $@;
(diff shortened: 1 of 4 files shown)


rsto (commented):
This looks good. Can you add a test for that?

rsto:
I will add createdmodseq support for Email/get on top of this PR. I will then add a test for Thread/get, too.

rsto:
Now with support for createdModseq in Email/get. Removing myself as reviewer, as I wrote the Email/get part of this. I did not come up with a good test for Thread/get so this will need to do.

cassandane/tiny-tests/JMAPEmail/email_get_createdmodseq
```diff
@@ -0,0 +1,106 @@
...
+
+test
+EOF
+    $mimeMessage =~ s/\r?\n/\r\n/gs;
+
+    # Sleep until the start of a second. This should result
+    # in the duplicates having the same internal date,
+    # unless the test code is running really slow.
+    my ($now_secs, $now_microsecs) = gettimeofday;
+    usleep(2 * 1000000 - $now_microsecs);
```
ksmurchison, line 41: Why not just set the internaldate as part of the APPEND?

ksmurchison (commented):
LGTM other than my comment.  Perhaps IMAPTalk doesn't support setting internaldate in APPEND?

cassandane/tiny-tests/JMAPEmail/email_get_createdmodseq
rsto: Because I was not aware that this is an option. I updated the test code.

ksmurchison (approved):

brong:
> LGTM other than my comment. Perhaps IMAPTalk doesn't support setting internaldate in APPEND?

I'm pretty sure it does!

```
my @res = $Talk->append($name, $flagtext, { Quote => $InternalDate }, $fh);
my @res = $Talk->append($name, $flagtext, { Quote => $InternalDate }, { Literal => $content });
```

rsto:
Yep, the test got updated to use that in APPEND.

PR #4579, opened 2023-08-17
jmap_mail: do not dereference NULL pointer in Email/parse

### LINKED ISSUE #4574: NULL dereference in jmap_mail.c:jmap_email_parse()
In the code below is executed first `part=NULL` and then `part->encoding`.  This is the same as `NULL->encoding` and invalid.

https://github.com/cyrusimap/cyrus-imapd/blob/c53219dd47ce40b7677406e22b298b40a10a6dfb/imap/jmap_mail.c#L8441-L8447

rsto:
Fixes #4574

Thanks to @dilyanpalauzov

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_parse_inmemory_blob
+++ b/cassandane/tiny-tests/JMAPEmail/email_parse_inmemory_blob
@@ -1,0 +1,45 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_parse_inmemory_blob
+    :min_version_3_9 :needs_component_jmap
+{
+    my ($self) = @_;
+    my $jmap = $self->{jmap};
+
+    # XXX: replace with the upstream one once RFC 9404 is finished
+    $jmap->AddUsing('https://cyrusimap.org/ns/jmap/blob');
+
+    my $mimeMsg = <<'EOF';
+From: <from@local>
+To: to@local
+Bcc: bcc@local
+Subject: test
+Date: Mon, 13 Apr 2020 15:34:03 +0200
+MIME-Version: 1.0
+Content-Type: text/plain
+
+hello
+EOF
+    $mimeMsg =~ s/\r?\n/\r\n/gs;
+
+    # XXX: can't use a result reference in array
+    my $blobId = 'G67501cd2e1eaaf65d25e6f3b49554d2193f06ee8';
+
+    $res = $jmap->CallMethods([
+        ['Blob/upload', {
+            create => {
+                b1 => {
+                    data => [{'data:asText' => $mimeMsg}]
+                },
+            },
+        }, 'R1'],
+        ['Email/parse', {
+            blobIds => [$blobId],
+            properties => ['subject', 'bodyStructure'],
+        }, 'R2'],
+    ]);
+    $self->assert_str_equals('Blob/upload', $res->[0][0]);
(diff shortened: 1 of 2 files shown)


cassandane/tiny-tests/JMAPEmail/email_parse_inmemory_blob
```diff
@@ -0,0 +1,45 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_parse_inmemory_blob
+    :min_version_3_9 :needs_component_jmap
```
dilyanpalauzov, line 5: Version 3.4 is also affected.

rsto: @elliefm is 3.4 still supported? I tagged this to backport to 3.8.

elliefm: @rsto I've tagged it for backporting as far as 3.4, and I'll see if it cherry-picks cleanly next time I'm doing a 3.4 release.  I wouldn't do a new 3.4 release just for this fix, but if at some point I'm doing a new 3.4 release for some other reason, this might as well be fixed too.

dilyanpalauzov: Why isn’t this copied to the 3.4, 3.6 and 3.8 branches now, irrespective of whether there will be a release?

elliefm:
Had to discard the test when backporting to 3.6, because it requires a capability 3.6 doesn't support.

If this goes to 3.4, cherry-pick the 3.6 version (98b9b2380) to save doing the same work twice.

dilyanpalauzov:
For me it is unclear why this is not backported to 3.4 irrespective of whether there will be a release.  The same applies to all other backports: if something is subject to backporting to any branch, why is it not backported immediately.

elliefm:
Backporting takes time and energy, and more so the further back it needs to go.  It's somewhat easier to do all the backporting for a particular version in one batch, write the release notes at the same time, and then do a release.

As far as I know, people mostly aren't using the git branches directly, and aren't likely to upgrade until there's a real release, so it doesn't make a lot of difference when they land on the git branch, as long as they do so before the release.

Are you still running 3.4?  I thought you might have upgraded to 3.6, since you've raised a bunch of issues and pull requests for 3.6 lately.

dilyanpalauzov:
When I upgrade to a new version, I adapt all the zillion patches I have from the previous version.  For some of them I create pull request in the hope that they will be integrated, so that in the future I will have to forward-port less changesets.  Then I run static analysis on cyrus-imap.  It does find real and not so real problems, this thread is about NULL dereference.  Then I look on the master branch to see if a problem found by the static analysis is already fixed.  When it is fixed I might create a pull request for backporting it.  https://github.com/cyrusimap/cyrus-imapd/pull/4570 is such an example, which is not handled yet.  I do not create other backporting pull requests, for errors fixed upstream, since it is unclear if creating the pull requests will be lost of time.

Sometimes it turns out that the fix is planned for backporting and is backported at a later time, but when I look for problems, the fix is not yet integrated so it still costs my time and efforts to tackle this.

That is why I prefer if things are backported promptly, irrespective of how many people might be using git to get the source.  In the current situation one of the reasons why new backports are not submitted is because already submitted backports are not handled.

PR #4580, opened 2023-08-17
jmap_calendar.c: don't send updates to attendees when one changes their partstat

ksmurchison:
Solves the problem of attendees getting mail bombed when one changes their partstat via CalendarEvent/participantReply


rsto (approved):
LGTM

PR #4592, opened 2023-08-19
jmap_contact.c:_contacts_set() memory leak

dilyanpalauzov:
• account in  imap/http_dblookup.c that mboxname_abook() cannot return NULL.


ksmurchison:
This will become obsolete once #4479 gets merged (next week?)

dilyanpalauzov:
This is relevant for the 3.6 and 3.8 branches.

dilyanpalauzov:
This is in no way obsolete.  The herein made observation, that `mboxname_abook()` cannot return `NULL` is still correct.  The code still does check, if the returned value from `mboxname_abook()` is `NULL`.

PR #4593, opened 2023-08-19
xapian_wrap.cpp:parse_doclangs() check for val.empty() only once

dilyanpalauzov:
It was checked before and within the function call, whether val is empty.


rsto (approved):
LGTM. Thanks.

PR #4596, opened 2023-08-19
Note on specialstrings and annotations in concept documentation

mfriedr2:
Add a concept documentation note on specialstrings and the (relatively new) Access Right "annotate" in these special strings.

### DIFF
--- a/docsrc/imap/concepts/overview_and_concepts.rst
+++ b/docsrc/imap/concepts/overview_and_concepts.rst
@@ -161,8 +161,8 @@ These are referred to as *negative rights*.
 Special Strings
 ===============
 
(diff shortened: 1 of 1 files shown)


elliefm (commented):
The overview is the wrong place for this detail.  The special strings cited are convenience aliases that only exist within the cyradm tool.  The cyradm tool recognises and translates them into valid ACL strings, but nothing else recognises them -- you will make a mess if you try to use these special strings anywhere else.

They're already documented in the cyradm man page, so I don't really think there's a useful place to move this section to, and even it it were rephrased to make clear that it's specific to cyradm, it would be too specific for the overview.  So I'm going to just close this rather than requesting changes.

Thanks, though!

mfriedr2:
I agree, @elliefm , and I have one personal question that lead to my original pull request and why I splitted in 2 commits :
Do you know any cyradm man page that references the imapd Access Right `n` ("annotatemsg") ?
Previous versions of cyrus-imapd were implementing annotatemsg by the `w` flag, but this is not the case any more. 
Any hint where to find this fact in the ACL concept ?

Issue #2213, opened 2017-12-11
static analysis finding


dilyanpalauzov:
The clang static analyzer reports:
```C
lib/imclient.c: line 696, column 17
695	            } else {
696	                parsed += avail; // Value stored to 'parsed' is never read
697	                imclient->replyliteralleft -= avail;
698	                return;
699	            }
700	        }

imap/ical_support.c:	line 311, column 26
311	            icaltimetype myend = rdate.time; // Value stored to 'myend' during its initialization is never read
312	            if (icalperiodtype_is_null_period(rdate.period)) {
313	                myend = icaltime_add(mystart, event_length);
314	            }
315	            else {
316	                mystart = rdate.period.start;
317	                if (icaldurationtype_is_null_duration(rdate.period.duration))
318	                    myend = rdate.period.end;
319	                else
320	                    myend = icaltime_add(mystart, rdate.period.duration);
321	            }

imap/search_squat.c: 	line 390, column 21
388	static int add_unindexed(SquatBuilderData *bb)
389	{
390	    struct opstack *top = opstack_top(bb); // Value stored to 'top' during its initialization is never read
391	    int r = 0;
392	 
393	    top = opstack_push(bb, /*doesn't matter*/0);

imap/message.c:	line 3020, column 17
3020	    const char *p = *sect; // Value stored to 'p' during its initialization is never read
3021	    struct body *subpart;
3022	    size_t len;
3023	    uint32_t cte;
3024	 
3025	    n = CACHE_ITEM_BIT32(*sect);
3026	    p = *sect += CACHE_ITEM_SIZE_SKIP;

imap/caldav_alarm.c:	lines 342 and 345, column 22
342	        icaltimetype alarmtime = icaltime_null_time(); // Value stored to 'alarmtime' during its initialization is never read
343	        if (icalvalue_isa(val) == ICAL_DURATION_VALUE) {
344	            icalparameter *param = icalproperty_get_first_parameter(prop, ICAL_RELATED_PARAMETER);
345	            icaltimetype base = icaltime_null_time();  // Value stored to 'base' during its initialization is never read
346	            if (param && icalparameter_get_related(param) == ICAL_RELATED_END) {
347	                base = end;
348	            }
349	            else {
350	                base = start;
351	            }
352	            base.is_date = 0; /* need an actual time for triggers */
353	            alarmtime = icaltime_add(base, trigger.duration);
354	        }
355	        else {
356	            /* absolute */
[...]

PR #4600, opened 2023-08-20
Correct values set, but unused

bogdro:
Should fix most of:
- https://github.com/cyrusimap/cyrus-imapd/issues/2213
- https://github.com/cyrusimap/cyrus-imapd/issues/2212

I chose not to correct quota.c:errmsg(), because it looks like an easy copy-paste pattern to allow adding new if-conditions in the future.

I chose to comment-out the assignments in case someone would add some default values again in the future, causing the same warnings. You are free to delete the commented-out code instead.


elliefm (changes requested):
Thanks! Sorry for the late review.

There's no need to leave old code lying around in comments, so please just remove it if you're removing it.  If we want it back later, it's in the git history.

This no longer merges cleanly.  If you're still around, can you rebase on current master to fix the conflicts and then force push please?

bogdro:
Hi!
Yes, I'm still around. I'll try to do all that when I have a bit of time.

elliefm:
Something has gone wrong with your rebase.  The only commits on the PR should be yours.  Probably your local master branch is not up to date with the upstream master branch, and you have rebased against your local one rather than the upstream one; or you have merged a master branch into your PR branch, instead of rebasing the PR branch onto the master branch.

You should first get your local master branch to the same point as the upstream one, and then rebase your PR branch onto that.  If you have not made changes on your own master branch, then you can update it something like below, assuming your git remote for this repository is called "upstream" (run `git remote -vv` to see the names).

**n.b. If you _have_ made changes on your local master branch, they will be lost if you proceed!**

```
# switch to your local master branch
git checkout master
# update your refs from the "upstream" remote
git fetch upstream
# force the local master to match the one on "upstream"
git reset --hard upstream/master
# your local master will now match the upstream master
# switch to your PR branch
git checkout codeqa
# redo the rebase against the local master
git rebase master
# visually confirm that the only new commits are now your ones
git log upstream/master..HEAD
# and then test and push your PR branch however you usually do
git push
```

bogdro:
Thanks for the help.
I did that, looked correct, but pushing failed due to the remote branch being forward.
After syncing all with master and attempting another rebase, I got conflicts during the rebase, again on the same files, but also brand-new conflicts, and quite many of those (don't know why, since everything should be synced).
To make it easier, can we close this PR and use https://github.com/cyrusimap/cyrus-imapd/pull/5131 instead?

PR #4601, opened 2023-08-20
Fix build with disabled sieve

bogdro:
When building without sieve (`./configure --disable-sieve`), compilation fails on imap/cyr_buildinfo.c because of missing headers and, in turn, missing symbols.
This simple patch (a bunch of #ifdefs) fixes the issue.
(Perhaps it's nonsense to build without sieve, I don't know, but I wanted actually just the CalDAV functionality + the minimum necessary requirements).

### DIFF
--- a/imap/cyr_buildinfo.c
+++ b/imap/cyr_buildinfo.c
@@ -64,8 +64,10 @@
 
 #include "master/masterconf.h"
 
+#ifdef USE_SIEVE
 #include "sieve/bytecode.h"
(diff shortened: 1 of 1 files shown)


ksmurchison (approved):

dilyanpalauzov:
For NGHTTP2, Jansson, Wslay, Sqlite and other cyr_buildinfo prints explicitly, if the feature is enabled or disabled at compile time.

For Sieve it seems this is currently not the case.

PR #4603, opened 2023-08-21
JMAP Email: ignore NUL bytes in Email/get bodyValues

rsto:
Sometimes we see HTML emails containing body parts with NUL bytes, e.g. 

    Content-Transfer-Encoding: quoted-printable
    Content-Type: text/html;charset=utf-8
    
    <html><body>hello=00 world</body></html>

This currently makes Cyrus truncate this body value at the first occurrence of a NUL byte.

I'm not entirely convinced NUL is legit in HTML, looking at its current definition of the [PLAINTEXT state](https://html.spec.whatwg.org/multipage/parsing.html#plaintext-state) and the ilk. But in any case we should do what everyone else is doing, which is omitting these bytes from the decoded body.

This PR just does that.

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_get_bodyvalues_crlf
+++ b/cassandane/tiny-tests/JMAPEmail/email_get_bodyvalues_crlf
@@ -1,0 +1,44 @@
+#!perl
+use Cassandane::Tiny;
+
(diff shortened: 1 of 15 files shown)


ksmurchison (approved):
LGTM

Issue #4611, opened 2023-08-22
shared mailbox - assertion failed when setting sharedseen


MrTfromVIE:
OS: debian 12
cyrus-imapd    3.6.1-4  - cyrus databases are twoskip

reproducing steps:
cyradm --user cyrus localhost
localhost> setmd xxx@xxx.local sharedseen true

results in 
cyrus/imap: Fatal error: Internal error: assertion failed: imap/mboxlist.c: 1088: user_isnamespacelocked(userid)

however. with a user mailbox it's functioning without any issue, f.e.: > setmd user/x1@xxx.local sharedseen true

testet with other shared mailbox - same issue -> assertion failed

elliefm:
I've thrown together a Cassandane test that reproduces this on cyrus-imapd-3.6 but not on master, so it looks like we already fixed the problem, but the fix hasn't been backported.  I'll track down what the fix was and see about backporting it to 3.6 (and 3.8, if it needs it too).

elliefm:
3.8 isn't affected either, looks like just 3.6

MrTfromVIE:
can confirm: just tested debian 12 with 3.8 from testing -> no assertion failure on shared mailbox

elliefm:
git bisect with my regression test says this was fixed on master by a164795dccb72b947d13f6f2555df2b9b6d16ea4

elliefm:
And that came from #4359, which contains a huge amount of other lock ordering fixes too.  I'll guess I'll see if the whole thing backports cleanly, or if I need to be selective about it...

MrTfromVIE:
thx for the links ...
if it helps:  tested with debian 11 and bullseye-backports version 3.6.1-2~bpo11+1 -> same issue, assertion failed ...

rjbs:
The affected versions are no longer supported, so I am closing this issue.

PR #4612, opened 2023-08-23
Flags: add test_seen_sharedmb_sharedseen

elliefm:
Regression test for #4611

(It's just a copy of `test_seen_sharedmb_nosharedseen` above it, but with `sharedseen => true` instead of `false`)


elliefm:
For 3.6, this goes with #4613

rsto (approved):

PR #4615, opened 2023-08-23
sitemap.py: sort work list for a stable sitemap.xml

elliefm:
The default ordering of `env.found_docs` seems to be unstable in newer Sphinx versions, resulting in sitemap.xml changes every single time the website builder runs, even if nothing actually changed.

With this list now processed in sorted order, sitemap.xml shouldn't change unless we've actually added, removed, or renamed documents.


elliefm:
@elliefm: this needs to be backported all the way to 2.5

rsto (approved):

elliefm:
This is now backported as far as 3.0.  Turns out 2.5 doesn't have the sitemap builder, so doesn't need the fix.

PR #4618, opened 2023-08-24
sieve-lex.l: fix warning for using function without prototype

dilyanpalauzov:
I cannot get the warning disappear by any other way.

### DIFF
--- a/sieve/sieve-lex.l
+++ b/sieve/sieve-lex.l
@@ -409,7 +409,7 @@ ws              [ \t]+
                                   /* RFC 5228: With the exceptions of strings
                                      and comments, the language is limited to
                                      US-ASCII characters. */
-                                  if (!isascii(yytext[0])) {
+                                  if (yytext[0] & ~0x7f) {
                                       sieveerror(sscript, "non-ASCII character");
                                   }
                                   return yytext[0];


ksmurchison:
I'm not getting a warning on my dev. system.

Does adding #include <ctype.h> fix it?

dilyanpalauzov:
The exact warning is
```
sieve/sieve-lex.l: In function ‘sievelex’:
sieve/sieve-lex.l:406:40: warning: implicit declaration of function ‘isascii’ [-Wimplicit-function-declaration]
  406 |                                       sieveerror(sscript, "non-ASCII character");
      |                                        ^~~~~~~
```
I use flex 2.6.4 with libc 2.37.  If I include in sieve-lex.l on line 49 `#include <ctype.h>`, then the generated sieve-lex.c contains:
```c
#if !(defined(_POSIX_C_SOURCE) || defined(_XOPEN_SOURCE) || \
    defined(_POSIX_SOURCE))
# define _POSIX_C_SOURCE 1 /* Required for fileno() */
# define _POSIX_SOURCE 1
#endif
…
#include <ctype.h>
```

ctype.h contains
```c
#if defined __USE_MISC || defined __USE_XOPEN

/* Return nonzero iff C is in the ASCII set
   (i.e., is no more than 7 bits wide).  */
extern int isascii (int __c) __THROW;

#endif /* Use X/Open or use misc.  */
```
When the above is preprocessed, __USE_MISC and __USE_XOPEN are not defined, and this has something to do with the _POSIX #defines inserted by flex.

ksmurchison:
Now I'm wondering if isascii() is what we want anyways.  I think we want isgraph()  (32 < yytext[0] < 127).  
I'd be fine with just using the manual check of yytext[0] value rather than a ctype.h function/macro

PR #4619, opened 2023-08-24
track the behavior seen by clients

rjbs:
The goal is, during an IMAP session, to keep track of interesting behavior, and then log it all at the end.  This reduces the need to put syslog lines in all over the place, and also reduces the need to correlate the lines to get a total picture of the session activity.

The real driver here is answering "Which IMAP clients use feature XYZ?"  If they identify with `ID` and we log that along with the behaviors seen, we can then:

* gather all the logs over a week or two
* group them by client id (by "name")
* union all the did-see-feature-X flags
* the end!

This means:
* making a new struct to track seen behaviors
* populating it as the session continues
* log it at session end
* reset it before new session begins

### DIFF
--- a/cassandane/Cassandane/Cyrus/ID.pm
+++ b/cassandane/Cassandane/Cyrus/ID.pm
@@ -69,6 +69,9 @@ sub test_cmd_id
 {
     my ($self) = @_;
 
+    # Purge any syslog lines before this test runs.
+    $self->{instance}->getsyslog();
+
     my $imaptalk = $self->{store}->get_client();
 
     return if not $imaptalk->capability()->{id};
@@ -77,6 +80,22 @@ sub test_cmd_id
     xlog $self, Dumper $res;
 
     $self->assert_str_equals('ok', $imaptalk->get_last_completion_response());
+
+    # should have logged some timer output, which should include the sess id,
+    # and since we sent a client id via IMAP ID, we should get that, too!
+    if ($self->{instance}->{have_syslog_replacement}) {
+        # make sure that the connection is ended so that imapd reset happens
+        $imaptalk->logout();
+        undef $imaptalk;
+
+        my @lines = $self->{instance}->getsyslog();
+
+        my (@behavior_lines) = grep { /session ended/ } @lines;
+
+        $self->assert_num_gte(1, scalar @behavior_lines);
+
+        $self->assert_matches(qr/\bid\.name=<cassandane>/, $_) for @behavior_lines;
+    }
 }
 
 1;
--- a/cassandane/Cassandane/Cyrus/Notify.pm
+++ b/cassandane/Cassandane/Cyrus/Notify.pm
@@ -280,6 +280,10 @@ sub test_message
 
     # we enabled NOTIFY, so we should see it in client behaviors
     if ($self->{instance}->{have_syslog_replacement}) {
+        # make sure that the connection is ended so that imapd reset happens
+        $talk->logout();
+        undef $talk;
+
         my (@lines) = grep { /session ended/ && /notify=<1>/ }
                       $self->{instance}->getsyslog();
 
--- a/cassandane/Cassandane/Cyrus/Notify.pm
+++ b/cassandane/Cassandane/Cyrus/Notify.pm
@@ -277,6 +277,14 @@ sub test_message
     # Should get no unsolicited responses
     $res = $store->idle_response({}, 1);
     $self->assert(!$res, "no unsolicited responses");
+
+    # we enabled NOTIFY, so we should see it in client behaviors
+    if ($self->{instance}->{have_syslog_replacement}) {
+        my (@lines) = grep { /session ended/ && /notify=<1>/ }
+                      $self->{instance}->getsyslog();
+
+        $self->assert_num_gte(1, scalar @lines);
+    }
 }
 
 sub test_mailbox
--- a/cassandane/Cassandane/Cyrus/Simple.pm
+++ b/cassandane/Cassandane/Cyrus/Simple.pm
@@ -195,11 +195,18 @@ sub test_cmdtimer_sessionid
 
     # should have logged some timer output, which should include the sess id
     if ($self->{instance}->{have_syslog_replacement}) {
-        my @lines = grep { m/\bcmdtimer:/ } $self->{instance}->getsyslog();
-        $self->assert_num_gte(1, scalar @lines);
-        foreach my $line (@lines) {
+        my @lines = $self->{instance}->getsyslog();
+
+        my @timer_lines = grep { m/\bcmdtimer:/ } @lines;
+        $self->assert_num_gte(1, scalar @timer_lines);
+        foreach my $line (@timer_lines) {
             $self->assert_matches(qr/sessionid=<[^ >]+>/, $line);
         }
+
+        my (@behavior_lines) = grep { /session ended/ && /notify=<[01]>/ }
+                               @lines;
+
+        $self->assert_num_gte(1, scalar @behavior_lines);
     }
 }
 
--- a/cassandane/Cassandane/Cyrus/Simple.pm
+++ b/cassandane/Cassandane/Cyrus/Simple.pm
@@ -195,6 +195,10 @@ sub test_cmdtimer_sessionid
 
     # should have logged some timer output, which should include the sess id
     if ($self->{instance}->{have_syslog_replacement}) {
+        # make sure that the connection is ended so that imapd reset happens
+        $imaptalk->logout();
+        undef $imaptalk;
+
         my @lines = $self->{instance}->getsyslog();
 
         my @timer_lines = grep { m/\bcmdtimer:/ } @lines;
--- a/cassandane/Cassandane/Cyrus/Simple.pm
+++ b/cassandane/Cassandane/Cyrus/Simple.pm
@@ -203,8 +203,7 @@ sub test_cmdtimer_sessionid
             $self->assert_matches(qr/sessionid=<[^ >]+>/, $line);
         }
 
-        my (@behavior_lines) = grep { /session ended/ && /notify=<[01]>/ }
-                               @lines;
+        my (@behavior_lines) = grep { /session ended/ } @lines;
 
         $self->assert_num_gte(1, scalar @behavior_lines);
     }
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -11070,6 +11070,8 @@ static void cmd_getmetadata(const char *tag)
     struct getmetadata_options opts = OPTS_INITIALIZER;
     annotate_state_t *astate = NULL;
 
+    client_behavior.did_metadata = 1;
+
     while (nlists < 3)
     {
         c = parse_metadata_string_or_list(tag, &lists[nlists], &is_list[nlists]);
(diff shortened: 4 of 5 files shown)


imap/imapd.h
```diff
@@ -431,4 +431,9 @@ enum {
 
 extern struct protstream *imapd_out, *imapd_in;
 
+struct client_behavior_registry {
+    int did_compress : 1;   /* client started COMPRESS */
+    int did_notify   : 1;   /* client used NOTIFY */
```
elliefm, line 436: Someone said elsewhere recently that we should be using `unsigned` rather than `int` for single-bit struct fields, I don't remember why exactly but it was something about sign extension ambiguity.  I also notice you're logging these fields with `%u` so maybe you meant to do that and forgot to go back.

That said, there's only ever one of these per process, so I don't think the extra instructions to bit shift these values into place are worth the space savings.  I'd just make them both full-size ints, signedness irrelevant.  We're C programmers; we're used to using ints for bools. :)

imap/imapd.c
```diff
@@ -141,6 +141,13 @@ static int nosaslpasswdcheck = 0;
 static int apns_enabled = 0;
 static int64_t maxsize = 0;
 
+const struct client_behavior_registry client_behavior_null = {
+  .did_notify = 0,
+  .did_compress = 0,
+};
+
+struct client_behavior_registry client_behavior = client_behavior_null;
+
```
elliefm, line 150: You don't need `client_behaviour_null`

`client_behaviour` should be `static` (it's not used outside this module), and you can initialise it without `client_behaviour_null` here in a few ways:

```
static struct client_behaviour_registry client_behaviour;  // 'static' globals are implicitly initialized with zeroes
static struct client_behaviour_registry client_behaviour = { 0 };   // or do it semi-explicitly
static struct client_behaviour_registry client_behaviour = { 0, 0 };   // or really explicitly
```

imap/imapd.c
```diff
@@ -868,6 +875,16 @@ static void imapd_reset(void)
...
 
+    /* log the client behaviors */
+    xsyslog(LOG_NOTICE, "session ended",
+                        "sessionid=<%s> userid=<%s>"
+                        "compress=<%u> notify=<%u>",
+                        session_id(),
+                        imapd_userid ? imapd_userid : "",
+                        client_behavior.did_compress,
+                        client_behavior.did_notify);
+    client_behavior = client_behavior_null;
```
elliefm, line 886: Without `client_behaviour_null`, you do this like this:

```suggestion
    memset(&client_behaviour, 0, sizeof(client_behaviour));
```

C99 and later add "compound literals", but we don't really use them in Cyrus (yet).  Using one of those would look something like this, and would be more in the spirit of your original choice (i.e. avoiding a function call):

```suggestion
    client_behavior = (struct client_behaviour_registry){ 0 };
```

imap/imapd.c
rjbs: Amended, thank you very much!

imap/imapd.c
rjbs: Thanks, amended!

imap/imapd.h
rjbs: I added the `: 1` at rsto's suggestion, and following some other code we have, so I'm going to stick with that.  I did add the unsigned in an amendment.  I just hadn't thought about it.

imap/imapd.c
rjbs: Hilariously, you wrote `behaviour` and I just copied that, so the code then had mismatched spellings.  I fixed it quickly, but it definitely made me smile. 😄

imap/imapd.h
dilyanpalauzov: This program prints -1.  Clang prints also 
```
a.c:9:7: warning: implicit truncation from 'int' to a one-bit wide bit-field changes value from 1 to -1 [-Wsingle-bit-bitfield-constant-conversion]
  b.a = 1;
      ^ ~
1 warning generated.
```
```c
#include <stdio.h>

struct f {
  int a: 1;
};

int main(void) {
  struct f b;
  b.a = 1;
  printf("%i\n", b.a);
  return 0;
}
```

rjbs:
@elliefm Thanks for your previous review, it was helpful.  Here's the current status:

* I believe the code should all work and not be a total mess.  Your feedback on that is welcomed. 😉
* The tests are not passing in automation, and I'm not sure why.  Can you tell me?
* I believe there are going to be more places we should instrument, and I would like to fix those over time.  For example, providing the VANISHED modifier to FETCH should turn on `did_qresync`, but I am happy with the code as it stands for our initial run.

I would love to merge this ASAP, if and only if it is *suitable* for merging, which at least means that I fix the tests.

Thanks again!

rjbs:
Oh, one more thing.  I want to log the "name" (case-insensitive) parameter from the ID we got, if any.  We should know this from `id_data.params`, but I don't understand how to find a specific key.  Seems like I need to iterate over them and stricmp the `av->value`, but I don't quite know how to get that going.

Advice appreciated.

rjbs:
I think this is as ready for review as it was before I touched it.  I'm on a train and reading RFCs (as one does) and realized I should add tracking for the ESEARCH command, which is the MULTISEARCH capability.

elliefm:
> imapd: add client_behavior, to track what **cliends** do

doh ;)

(proper review to follow)

imap/imapd.c
```diff
@@ -873,11 +873,27 @@ static void imapd_reset(void)
...
+
+                        client_behavior.did_metadata  ? " metadata=<1>"   : "",
+                        client_behavior.did_move      ? " move=<1>"       : "",
+                        client_behavior.did_notify    ? " notify=<1>"     : "",
+                        client_behavior.did_preview   ? " preview=<1>"    : "",
+
+                        client_behavior.did_qresync   ? " qresync=<1>"    : "",
+                        client_behavior.did_replace   ? " replace=<1>"    : "",
+                        client_behavior.did_savedate  ? " savedate=<1>"   : "",
+                        client_behavior.did_searchres ? " searchres=<1>"  : "");
```
elliefm: So, I assume there's a reason you're omitting fields that are zero, rather than just logging a lot of `"field=<%u>"` which will get zeroes or ones.  I might like to see a comment explaining the complexity.

If you don't mind rearchitecting the whole thing, you could have the computer compute the log string by hardcoding more metadata.  Something like (untested, written directly in the github web ui, probably full of typos, so think of this as pseudocode...):

```
enum client_behaviour {
    CLIENT_DID_ANNOTATE,
    // [...]
    CLIENT_DID_SEARCHRES,

    NUM_CLIENT_BEHAVIOURS /* leave last! */
};

static uint8_t client_behavior_registry[NUM_CLIENT_BEHAVIOURS];

static const char const *client_behavior_name[NUM_CLIENT_BEHAVIOURS] = {
    "annotate",
    // [...]
    "searchres",
};

static void get_client_behaviors_string(struct buf *buf)
{
    const char *sep = "";
    unsigned i;

    for (i = 0; i < NUM_CLIENT_BEHAVIOURS; i++) {
        // XXX decide whether to always log foo=<%u>, or only log when foo=<1>
        buf_printf(buf, "%s%s=<%u>",
            sep,
            client_behaviour_name[i],
            client_behaviour_registry[i]);
        sep = " ";
    }
}

```

And then each time a client does a thing, you'd do something like:

```
   client_behaviour_registry[CLIENT_DID_ANNOTATE] = 1;
```

And in `imapd_reset()`, you'd do something like:

```
    struct buf behaviors = BUF_INITIALIZER;
    // [...]

    get_client_behaviors_string(&behaviors);

    xsyslog(LOG_NOTICE, "session ended",
                        "sessionid=<%s> userid=<%s> %s"
                        session_id(),
                        imapd_userid ? imapd_userid : "",
                        buf_cstring(&behaviors));

    // [...]
    memset(client_behavior_registry, 0, sizeof client_behavior_registry);
    buf_free(&behaviours);
```

This would come at the cost of each behaviour flag taking a byte rather than a bit to store, but I still don't think that's an optimisation worth chasing.  If you wanted to chase that anyway, you could make client_behaviour_registry a large unsigned integer, and if the client performs behaviour n (according to the enum), set the nth bit.  You could also do the same with a `struct { foo :1; }`, but that would be brittle, because then you'd have three lists to have to keep coordinated (behaviour ids, behaviour names, struct field names).

cassandane/Cassandane/Cyrus/Notify.pm
```diff
@@ -277,6 +277,14 @@ sub test_message
...
     $res = $store->idle_response({}, 1);
     $self->assert(!$res, "no unsolicited responses");
+
+    # we enabled NOTIFY, so we should see it in client behaviors
+    if ($self->{instance}->{have_syslog_replacement}) {
+        my (@lines) = grep { /session ended/ && /notify=<1>/ }
+                      $self->{instance}->getsyslog();
+
+        $self->assert_num_gte(1, scalar @lines);
+    }
```
elliefm, line 287: Same thing as the other failing test -- we enabled NOTIFY on the client `$talk`, and that client is still connected, so we haven't logged its "session ended" yet

elliefm:
> Oh, one more thing. I want to log the "name" (case-insensitive) parameter from the ID we got, if any. We should know this from id_data.params, but I don't understand how to find a specific key. Seems like I need to iterate over them and stricmp the av->value, but I don't quite know how to get that going.

I haven't really looked at attvaluelist before, but it looks like it's an ordinary linked list traversal... so what about something like:

```
struct attvaluelist *p = imapd_id.params;
char *id_name = NULL;

for (; p; p = p->next) {
    if (0 == strcasecmp(p->attrib, "name")) {
        id_name = xstrdup(buf_cstring(&p->value));
        break;
    }
}

xsyslog(..., 
    "... idname=<%s> ...",
    ...,
    id_name,
    ...);
xfree(id_name);
```

If this all happens together in the one scope then it's not necessary to dup/free the value, just use the pointer you get from buf_cstring().  But if you were going to cache this value early but not print it till later for some reason, better to take a proper copy of the value.

cassandane/Cassandane/Cyrus/ID.pm
```diff
@@ -69,6 +69,11 @@ sub test_cmd_id
 {
     my ($self) = @_;
 
+    if ($self->{instance}->{have_syslog_replacement}) {
+        # Purge any syslog lines before this test runs.
+        $self->{instance}->getsyslog();
+    }
```
elliefm, line 75: FYI This part doesn't need to be wrapped in the have_syslog_replacement check, since we don't care about the logged lines, so it doesn't matter if there weren't any due to a missing syslog replacement :)

ksmurchison (approved):
I haven't verified that we caught ALL places where an extension has been implicitly/explicitly enabled, but otherwise LGTM

cassandane/Cassandane/Cyrus/Simple.pm
```diff
@@ -195,11 +195,18 @@ sub test_cmdtimer_sessionid
...
+        my @timer_lines = grep { m/\bcmdtimer:/ } @lines;
+        $self->assert_num_gte(1, scalar @timer_lines);
+        foreach my $line (@timer_lines) {
             $self->assert_matches(qr/sessionid=<[^ >]+>/, $line);
         }
+
+        my (@behavior_lines) = grep { /session ended/ && /notify=<[01]>/ }
+                               @lines;
+
+        $self->assert_num_gte(1, scalar @behavior_lines);
```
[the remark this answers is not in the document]
rjbs: Thanks, that was it!

imap/imapd.c
rjbs: Thanks, I have added a comment.  Your provided code was interesting and provided some implicit pointers (ha) on C idiom for me.  I think for now, I'll leave things as they are, but may come back to this in the future if it seems like the trade-offs changes…

cassandane/Cassandane/Cyrus/ID.pm
rjbs: Legit, I amended things.

PR #4625, opened 2023-08-26
upgrade.rst: remove instructions for upgrades before 3.2

dilyanpalauzov:
The only way to upgrade to 3.6 is to have first version 3.2 installed. That said, it can be assumed that users will first read the upgrade instructions to version 3.2 from https://www.cyrusimap.org/3.2/imap/download/upgrade.html .

Since there is stated already that BDB is gone in Cyrus IMAP 3.0 and the handling of special-use flags after 2.4 is discussed, there is no need to have this information in the upgrade notes for 3.6, 3.8 and master/dev .

### DIFF
--- a/docsrc/imap/download/upgrade.rst
+++ b/docsrc/imap/download/upgrade.rst
@@ -132,59 +132,15 @@ of the instructions are assuming a sandboxed 3.8 installation.
 Upgrade by replicating
 ~~~~~~~~~~~~~~~~~~~~~~
 
-If you're familiar with replication, and your current installation is 2.4 or
-newer, you can set up your existing installation to replicate data to a new
-3.8 installation and failover to the new installation when you're ready. The
-replication protocol has been kept backwards compatible.
-
-If your old installation contains mailboxes or messages that are older than
-2.4, they may not have GUID fields in their indexes (index version too old),
-or they may have their GUID field set to zero.  3.8 will not accept message
-replications without valid matching GUIDs, so you need to fix this on your
-old installation first.
-
-You can check for affected mailboxes by examining the output from the
-:cyrusman:`mbexamine(8)` tool:
-
-* mailboxes that report a 'Minor Version:' less than 10 will need to have
-  their index upgraded using :cyrusman:`reconstruct(8)` with the
-  `-V <version>` parameter to be at least 10.
-* mailboxes containing messages that report 'GUID:0' will need to have
-  their GUIDs recalculated using :cyrusman:`reconstruct(8)` with the `-G`
-  parameter.
-
-If you have a large amount of data, these reconstructs will take a long time,
-so it's better to identify the mailboxes needing attention and target them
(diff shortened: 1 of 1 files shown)


rsto:
@elliefm This looks good to me, other than it should refer to version 3.10 rather than 3.8 in the upgrade instructions. Do you agree?

elliefm:
@rsto This is the upgrade instructions on the master branch, and the master branch documentation is for the eventual next release, so "3.10" wouldn't be strictly correct here either.  But I don't think it really matters anyway, since I'll review and update the document when preparing next year's stable release and fix the number then.  I suppose we could use the magic word "UNRELEASED" here, like we already do in imapoptions.

rsto:
@elliefm I would apply the following changes, e.g. keep mentioning that BDB is not supported anymore: 

```
Upgrade by replicating
~~~~~~~~~~~~~~~~~~~~~~

If you're familiar with replication, you can set up your existing installation to
replicate data to a new UNRELEASED installation and failover to the new installation
when you're ready. The replication protocol has been kept backwards compatible.

Upgrade in place
~~~~~~~~~~~~~~~~

If you are upgrading in place, you will need to shut down Cyrus
entirely while you install the new package.  If your old installation
was using Berkeley DB format databases, you will need to convert or
upgrade the databases **before** you upgrade.  Cyrus UNRELEASED does not
support Berkeley DB at all.
```

dilyanpalauzov:
> keep mentioning that BDB is not supported anymore

Why?

elliefm:
I wound up making similar changes myself as part of the 3.12 release, they're in 8ac92541f

dilyanpalauzov:
Currently the text at https://www.cyrusimap.org/3.12/imap/download/upgrade.html states: “you should first upgrade to 3.10.1 … Upgrade by replicating … If you're familiar with replication, and your current installation is 2.4 or newer, you can set up your existing installation to replicate data to a new 3.12 installation”

This text still suggests at the same time that before upgrading to 3.12 the underlying system must be 3.10, but states also that upgrade over replacation from 2.4 can be achieved.

If upgrades from 2.4 are still possible, the sentence that upgrade first upgrade to 3.10 should be made, is confusing.

elliefm:
Yeah I left that section alone, because I don't know what's actually true -- as far as I know you can still replicate from 2.4 to 3.12.

This document is a bit of a mess, and these kind of piecemeal updates of individual details don't really resolve that, but it's what we've got to work with.  I might mess around with the heading levels, and see if I can get a more clear separation between the "upgrade in place" vs "install a new server and replicate your data into it" sections.

elliefm:
Actually, I just found another thing (#4967) that reckons that replication from 3.0-or-earlier to 3.2-or-later are broken for mailboxes that have annotations.  So replication from 2.4->3.12 _might_ work if no mailboxes have annotations, but will break if they do.

PR #4629, opened 2023-08-29
Add .mailmap

tjwds:


tjwds:
cc @brong

brong (approved):
This looks great as a first pass - we can always review it again later.  Thanks for doing this!

PR #4633, opened 2023-09-01
fix FETCH BINARY response for unknown CTE

### LINKED ISSUE #4567: BINARY FETCH and x-uuencode
I'm not sure Cyrus does not support uuencoded parts at all, but the behavior I see is wrong anyway.

```
C: A0008 UID FETCH 31 (BINARY.PEEK[2])
S: * 2 FETCH (UID 31)
S: A0008 OK Completed (0.000 sec)
```

So, it does not return the body nor an error. Dovecot when it can't decode a content returns `NO [UNKNOWN-CTE]` or `NO [PARSE]`. I suppose, if I don't see `BINARY[2]` in the response I could consider this an error condition and fallback to use non-binary FETCH, but a `NO` response would be better. Of course, it would be even better to make it just work.

And another issue with a non-UID FETCH.
```
C: A0008 FETCH 2 (BINARY.PEEK[2])
S: A0008 NO No matching messages (0.000 sec)
C: A0009 FETCH 2 (BODY.PEEK[2])
S: * 2 FETCH (BODY[2] {4272}
...
```
So the response is wrong, again.

Here's the sample message [test_message_uuencode_bug.txt](https://github.com/cyrusimap/cyrus-imapd/files/12326844/test_message_uuencode_bug.txt)

Tested with Cyrus IMAP 3.6.0.

elliefm:
FETCH of BINARY, BINARY.PEEK, or BINARY.SIZE are supposed to fail and report `[UNKNOWN-CTE]` if the requested message section has an unknown Content-Transfer-Encoding.  For FETCH, we were failing and reporting "No matching messages".  For UID FETCH, we were succeeding(!) but not outputting the requested section.

We already detected and returned the `IMAP_NO_UNKNOWN_CTE` error condition in `index_fetchsection()`, but the result was discarded by `void index_fetchresponses()` on the way back up the stack, and never sent to the client.

This PR changes `index_fetchresponses()` to return `int`, and plumbs the error through  so that `cmd_fetch()` can eventually send the correct response back to the client.  It also ensures that if some part of the fetch has failed, other parts succeeding do not conceal the error.  A new test verifies the correct behaviour for BINARY.PEEK and BINARY.SIZE.  I didn't bother testing plain BINARY, because it's the same as BINARY.PEEK except that it also sets the \Seen flag, which leads to an unsolicited flags response, which would complicate the test in an uninteresting way.

Fixes #4567

### DIFF
--- a/cassandane/Cassandane/Cyrus/Fetch.pm
+++ b/cassandane/Cassandane/Cyrus/Fetch.pm
@@ -1007,4 +1007,63 @@ sub test_preview_args
     $self->assert_str_equals('bad', $imaptalk->get_last_completion_response());
 }
 
+sub test_unknown_cte
+{
+    my ($self) = @_;
+
+    my $imaptalk = $self->{store}->get_client();
+
+    my $ct = "multipart/mixed";
+    my $boundary = "mitomycin";
+
+    # Subpart 1
+    my $body = ""
+    . "--$boundary\r\n"
+    . "Content-Type: text/plain; charset=\"utf8\"\r\n"
+    . "\r\n"
+    . "The attachment to this email is meaningless gibberish\r\n";
+
+    # Subpart 2
+    $body .= ""
+    . "--$boundary\r\n"
+    . "Content-Disposition: attachment; filename=\"gibberish.dat\"\r\n"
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):
LGTM

PR #4640, opened 2023-09-05
Sunset cyruslibs in favor of third_party directory

rsto:
This moves the contents of the https://github.com/cyrusimap/cyruslibs repository into version control of the cyrus-imapd project. Doing so allows to define and build the cyruslibs dependencies of a given branch. For example, the current https://github.com/cyrusimap/cyrus-imapd/tree/jmap_contacts_ietf tree requires a specific libical branch for building, but nowhere is that made clear during Cyrus builds.

This PR does nothing to further integrate building cyruslibs. It still is required to first run the `build.sh` script before running autoconf for the cyrus-imapd source.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,40 +1,40 @@
 [submodule "jansson"]
-	path = jansson
+	path = third_party/jansson
 	url = https://github.com/cyrusimap/jansson.git
 	branch = cyrus
 [submodule "xapian"]
-	path = xapian
+	path = third_party/xapian
 	url = https://github.com/cyrusimap/xapian.git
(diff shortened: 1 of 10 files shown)


dilyanpalauzov:
This something FastMail specific.  I do not think it is necessary in the general case.

rsto:
>  I do not think it is necessary in the general case.

It indeed isn't. But it's very useful for using https://github.com/cyrusimap/cyruslibs

dilyanpalauzov:
Since this is only relevant for FastMail I am against including it in the cyrus-imapd repository.

PR #4641, opened 2023-09-07
Conversations correct basecid

brong:
OK - this turned out to be surprisingly complex!

Rob M reported to me that a user was seeing blank messages in the Fastmail UI.  Symptom was "Email/query returns emailIds that Email/get says notFound for".

I discovered that there were messages with a missing basecid - the same GUID was in two separate folders, and in one it had a basecid, in the other not.  This was for a split conversation.  I quickly realised that the fix I put in a while back for copying the CID for the same GUID wasn't also copying the basecid.

So, I fixed that - but realised to get the basecid, we needed to store it in every G key.  So I added that (only if you have uuid folders, I figured that was better than doing some odd/even version magic).

Then I found more issues when testing.  Turns out, if there was more than one copy, the "B" keys weren't being cleaned up correctly, because the exists count kept finding the other zeroed out record.

Also, when you zero, we should also nuke all the newcid records on the `#splitconversations` folder.  There was an XXX comment to that effect, but it contained incorrect code, so I fixed that too.

In all, I'd say I've spent a solid 8 hours on this.  Damn.  But it's looking really solid now, and the test case includes copying the same message to two folders and the CIDs and BASECIDs matching across the lot now.

When we update servers, the G key parser is actually backwards compatible, so it's going to be safe to downgrade, though not safe to upgrade again if we change the G key datastructure format AND reuse the version number, so let's don't do that.

I also discovered that `ctl_conversationsdb -R` writes the G keys a few times, but I believe it's because it's working on two copies of the database so I'm not too stressed about that (yep, I scatted syslogs in places)

This MR also adds cid=<> logging to append and expunge.  We already log it for touched, so it just adds a tiny bit more logging and you can see CID changes more easily.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Conversations.pm
+++ b/cassandane/Cassandane/Cyrus/Conversations.pm
@@ -344,8 +344,10 @@ sub test_reconstruct_splitconv
     my ($self) = @_;
     my %exp;
 
+    my $talk = $self->{store}->get_client();
+
     # check IMAP server has the XCONVERSATIONS capability
-    $self->assert($self->{store}->get_client()->capability()->{xconversations});
+    $self->assert($talk->capability()->{xconversations});
 
     xlog $self, "generating message A";
     $exp{A} = $self->make_message("Message A");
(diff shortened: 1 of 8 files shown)


cassandane/Cassandane/Cyrus/Conversations.pm
```diff
@@ -380,6 +388,22 @@ sub test_reconstruct_splitconv
...
 
     $self->check_messages(\%exp, keyed_on => 'uid');
+    $talk->select("foo");
+    $self->check_messages(\%exp, keyed_on => 'uid');
+    $talk->select("INBOX");
+
+    # zero everything out
+    $self->{instance}->run_command({ cyrus => 1 }, 'ctl_conversationsdb', '-z', 'cassandane');
+
+    $self->{instance}->run_command({ cyrus => 1 }, 'ctl_conversationsdb', '-d', 'cassandane');
```
rsto, line 398: no need to dump here?

rsto (approved):
LGTM minus one nit in the test code

cassandane/Cassandane/Cyrus/Conversations.pm
brong: Oh yeah, that was just so I could read the output in `-v -v`, I'll remove it.

brong:
Documentation added :)  Woohoo

PR #4642, opened 2023-09-07
JMAP Calendars: report unshared calendars as destroyed

rsto:
This changes the Calendar/changes code to report any unshared calendar ids in the current modseq window as destroyed, rather than omitting them from the response.
This is required so that clients can remove calendar ids from their shared account cache, in case such calendars get unshared for the authenticated user.
As for Mailbox/changes, this has the drawback that calendar ids of calendars hidden to the authenticated user get reported.


ksmurchison (approved):
As discussed offline, this is an acceptable interim solution until we have something better.

PR #4644, opened 2023-09-07
jmap_mail: capitalize MIME in _email_to_mime

tjwds:


rsto (approved):
MIME header names are case-insensitive, but I'm OK with making things simpler for implementations.

PR #4646, opened 2023-09-10
configure.ac: add AM_SILENT_RULES([yes])

dilyanpalauzov:
This will make compiler warnings more visible by default.


elliefm (approved):

elliefm:
This was already possible by passing `--enable-silent-rules` to configure, but I have no issue with making that the default

PR #4647, opened 2023-09-10
seen_db:seen_merge_cb() memory leak

dilyanpalauzov:


ksmurchison (changes requested):
This looks partially correct.  Looks like we also need to free newsd.seenuids.
And we should probably just use seen_freedata() in both places

ksmurchison (approved):

PR #4651, opened 2023-09-15
cyr_info BYTESIZE suppport

### LINKED ISSUE #4650: core dump while checking config with cyr_info
cyr_info 3.8.1 aborts while checking config.

cyr_info conf-lint -C <path to imapd.conf> -M <path to cyrus.conf> works, but conf-default / conf_all aborts with core dump:

# cyr_info conf-default -C /etc/imapd.conf -M /etc/cyrus.conf
acl_admin_implies_write: no
..
archive_after: 7d
Aborted (core dumped)

either -C or -M or both specified, conf-default/conf-all aborts always on my system.

elliefm:
`cyr_info` was overlooked when adding BYTESIZE support, leading to `abort()` on the unrecognised option type when the `conf`, `conf-all`, or `conf-default` sub-commands were executed.

The fix itself is trivial -- `OPT_BYTESIZE` just needs to be handled the same as `OPT_DURATION` in these functions (`do_conf()` and `do_defconf()`).  The greater part of this PR is in updating our tests to detect this class of problem in the future.

We used to have a test for `cyr_info conf` that would have exposed the crash, but our config handling changed at some point, the test was disabled rather than updated, and so we no longer had such a test.  The changes here bring that test back and update it for current reality, as well as add similar tests for `cyr_info conf-all` and `cyr_info conf-default`.

If we add another new option type in the future, and forget to update cyr_info to know about it, these tests will fail due to the core files left behind when it aborts, and we'll know we've missed it.

Fixes #4650 

**Reviewers:** This is most easily read commit-by-commit, the single diff winds up a shambles.

### DIFF
--- a/cassandane/Cassandane/Config.pm
+++ b/cassandane/Cassandane/Config.pm
@@ -103,9 +103,6 @@ sub default
             'partition-default' => '@basedir@/data',
             sasl_mech_list => 'PLAIN LOGIN',
             allowplaintext => 'yes',
-            # config options used at FastMail - may as well be testing our stuff
-            expunge_mode => 'delayed',
-            delete_mode => 'delayed',
             # for debugging - see cassandane.ini.example
             debug_command => '@prefix@/utils/gdbtramp %s %d',
             # everyone should be running this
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):
LGTM

elliefm:
@elliefm once this is backported to 3.8, #4723 can be closed

PR #4657, opened 2023-09-20
enable CI for 3.4 branch

elliefm:
This is basically just a cherry-pick of all the commits to `.github/` from the master branch.  Submitting it as a PR in the hope that CI picks it up and runs, which will show me whether I've done it right.


elliefm:
That last WIP commit is aka #4658 -- integrate it properly when that's been merged.

elliefm:
@wolfsage Do you might casting an eye over this, in case I've somehow missed something?

My process was approximately:

1. identify all the commits that ever altered `.github/` on the master branch
2. identify what PRs they came from (using `git log --graph ...` and searching the output)
3. rebase each of those PR commit sets `--onto` this branch*
4. push it all, and let CI run

There were one or two PR's that also changed unrelated things, and in those cases at step 3 I only grabbed the changes that were relevant.  I tackled the PRs in the same order they were merged to master, so the commit dates are somewhat out of order, but the sequence of changes should match.

The first time I tried this I just took the commits that updated `.github/` files, but in doing so I missed a corresponding change to `tools/build-with-cyruslibs.sh`, so CI wasn't running the JMAP tests.  At that point I switched to doing whole-PRs rather than single commits, and I think I've got everything now (and CI seems to have run about the right number of tests).  But a second set of eyes on it would be appreciated!

wolfsage:
This looks right to me!

There are two commits in master that are *not* in this MR, should they be?

a0da42f0f95
01f73d6cbe1

Cheers

wolfsage (approved):

elliefm:
Oooh good question...

* **tools/build-with-cyruslibs.sh: add -Werror to CFLAGS and CXXFLAGS**
  we don't really do new development on 3.4, so we're unlikely to be introducing new warnings that we didn't already find and squash on master
* **cassandane.ini.dockertests: no longer suppress urlauth**
  probably need to test this one out -- it might be that imaptest.urlauth now works on master due to changes on master, in which case it would still need to be suppressed on 3.4 if those changes weren't backported

They don't seem particularly important to me right now, and they don't change the same .github/workflows/main.yml as the rest do, so it's not so important that they be landed in a historically-consistent order.  So I think I'll omit them for now, and just grab them later if it turns out they're useful.

PR #4661, opened 2023-09-22
small cyradm fixes

elliefm:
This fixes a couple of problems that were making `cyradm` not work on my system:

IMAP.so (for Cyrus::IMAP) needs to be linked with the regex library, but wasn't.

The boilerplate added to perl scripts by `tools/fixsearchpath.pl` (to detect and compensate when Cyrus has been installed with a `DESTDIR`) was assuming `$bindir` contains the full path, when it actually just contains the path relative to `$prefix`.  As a result, it was mistaking `$prefix` for `DESTDIR`, trying to compensate for it, and then being unable to find its dependencies.  The fix is to use the full `"$prefix$bindir"` path where the full path is expected.

I guess we've never really noticed this because we don't use `cyradm`, and others are probably installing to a standard prefix that perl already knows to look in, in which case their `cyradm` is able to find its dependencies despite the bogus path compensation.  Though I have no idea how it works for them when the regex library isn't linked properly.


ksmurchison (approved):
If it works, it works

PR #4663, opened 2023-09-23
lmtp_sieve.c:sieve_imip() call `zoneinfo_open(NULL)` during initialization

dilyanpalauzov:
Otherwise when `processimip` calls `caldav_store_resource()` → `strip_vtimezones()` → `zoneinfo_lookup()` the timezones are never found.  Then the known timezones are not stripped from the iCalendar object.


brong (approved):
This looks good, thanks

dilyanpalauzov:
This should be backported.

PR #4669, opened 2023-09-24
Replace `xzfree()` with `free()`

dilyanpalauzov:
when the former has no added value.  That is, when a new value is assigned to the parameter on the next line.

In `strarray_fini()` the data is never read again, so there is need to zero it.


brong:
No, we don't want to change this - it's not a serious performance concern and we're creating a habit of using zxfree so we always zero out things that we're freeing.  It's safer for the next author to add something in between.

PR #4670, opened 2023-09-24
jmap_query_parse() and jmap_querychanges_parse(): parameter err is never NULL

dilyanpalauzov:
In all invocations of `jmap_querychanges_parse()` and `jmap_query_parse()` the last parameter err is never `NULL`.  When `err` is compared with NULL in the code the clang static analyzer assumes that err can be `NULL`.  On later `*err=` assignments the analyzer thinks that err can be NULL and reports

> Dereference of null pointer (loaded from variable 'err')

### DIFF
--- a/imap/jmap_api.c
+++ b/imap/jmap_api.c
@@ -2374,7 +2374,7 @@ HIDDEN void jmap_query_parse(jmap_req_t *req, struct jmap_parser *parser,
                                   filter_cb, filter_rock, err);
                 jmap_parser_pop(parser);
                 query->filter = arg;
-                if (err && *err) {
+                if (*err) {
                     goto done;
                 }
             }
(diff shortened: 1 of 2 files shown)


brong:
No, the APIs are designed so they can be called with null.  If this is an issue, we should fix the code to always test `if (err)` before assigning instead.

dilyanpalauzov:
The modified functions are very similar to this one
```c
void z(char** y, int a) {
  if (y && a) *y = "abc";
  if (a) *y = "123"; else *y = "xyz";
}
```
If the function `z()` is put in a file a.c, so that a.c  does not contain anything else, calling then `clang --analyze -c a.c` prints:
```
a.c:3:13: warning: Dereference of null pointer (loaded from variable 'y') [core.NullDereference]
  if (a) *y = "123"; else *y = "xyz";
          ~ ^
a.c:3:30: warning: Dereference of null pointer (loaded from variable 'y') [core.NullDereference]
  if (a) *y = "123"; else *y = "xyz";
                           ~ ^
2 warnings generated.
```
That is: the function `z()` first is not sure, whether `y` can be NULL and checks if `y` is NULL.  But later the function `z()` assumes that `y` is not NULL and assigns a value to it.

By closing this ticket the problem reported by the static analyzer remains a valid concern.

The way how the APIs are designed in respect to `NULL` parameters is to my knowledge not documented. 

Moreover, if a function is designed to accept NULL as parameter, but the code never passes NULL on that position, when calling the function, then the code is unnecessary complicated.

PR #4671, opened 2023-09-25
JMAP Sieve: do not decode valid UTF-8 strings in MIME headers

rsto:
This fixes an issue with malformed MIME headers where implementations store arbitrary UTF-8 strings without encoding them as defined in RFC 2047. This impacts JMAP Email/query filters during Sieve evaluation.

Such headers already were handled leniently for stored messages, e.g. ones having a cache record. Now messages without cache record also allow undecoded UTF-8 values.


ksmurchison (approved):
LGTM

elliefm:
Not sure if this affects 3.6 or 3.8, but I've added the backport tags just in case.

PR #4672, opened 2023-09-26
lmtp_sieve: ignore VTIMEZONE errors for IANA timezones

rsto:
Sometimes a client send iTIP requests containing invalid VTIMEZONE definitions such as

   BEGIN:VTIMEZONE
   TZID:UTC
   END:VTIMEZONE

Cyrus is perfectly capable of making sense of this VTIMEZONE, so let's leniently accept bogus VTIMEZONEs in iTIP messages if they can be mapped to an IANA timezone. This only applies for iTIP, such iCalendar data still gets rejected during PUT and import.

### DIFF
--- a/imap/lmtp_sieve.c
+++ b/imap/lmtp_sieve.c
@@ -1502,7 +1502,8 @@ static int sieve_imip(void *ac, void *ic, void *sc, void *mc,
     }
 
     cyrus_icalrestriction_check(itip);
-    if ((errstr = get_icalcomponent_errstr(itip)) &&
+    if ((errstr = get_icalcomponent_errstr(itip,
+                    ICAL_SUPPORT_ALLOW_INVALID_IANA_TIMEZONE)) &&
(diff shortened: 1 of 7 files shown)


ksmurchison (approved):
LGTM

imap/lmtp_sieve.c
```diff
@@ -1501,8 +1501,9 @@ static int sieve_imip(void *ac, void *ic, void *sc, void *mc,
         itip_strip_personal_data(comp);
     }
 
+    unsigned ical_support_flags = ICAL_SUPPORT_ALLOW_INVALID_IANA_TIMEZONE;
     cyrus_icalrestriction_check(itip);
-    if ((errstr = get_icalcomponent_errstr(itip)) &&
+    if ((errstr = get_icalcomponent_errstr(itip, ical_support_flags)) &&
```
dilyanpalauzov, line 1506: Why is here an indirection over a variable done, rather than passing a consant as second parameter?

rsto: Right, there isn't really a good reason other than it's a left-over from refactoring. Removed.

PR #4677, opened 2023-10-03
jmap_calendar.c: skip sending iMIP if /participantReply{participationStatus} isn't changing

ksmurchison:


cassandane/tiny-tests/JMAPCalendars/calendarevent-participantreply-simple
```diff
@@ -160,8 +160,33 @@ sub test_calendarevent_participantreply_simple
...
+            participantEmail => "bugs\@looneytunes.com",
+            updates => {
+                participationStatus => "accepted"
+            }
+        }, "R2"]]);
+        $self->assert_str_equals("2.0", $res->[0][1]{scheduleStatus});
+
+        xlog $self, "verify that NO iMIP messages are sent to organizer/attendee";
+        $data = $self->{instance}->getnotify();
+        (@imips) = grep { $_->{METHOD} eq 'imip' } @$data;
```
rsto, line 184: The parentheses around `@imips` are redundant here AFAIK. It's testing OK without them.

imap/jmap_calendar.c
```diff
@@ -8108,6 +8109,19 @@ static int jmap_calendarevent_participantreply(struct jmap_req *req)
         goto done;
     }
 
+    /* Compare partStat */
+    param = icalproperty_get_first_parameter(attendee, ICAL_PARTSTAT_PARAMETER);
```
rsto, line 8113: Calling `icalproperty_get_first_parameter` with a NULL attendee won't cause a crash, but still is considered an error by libical, e.g. it will set libical's errno to `ICAL_BADARG_ERROR`. I think this whole conditional block could be moved up in the loop that sets that `attendee`, right before the `break`.

rsto (changes requested):

cassandane/tiny-tests/JMAPCalendars/calendarevent-participantreply-simple
ksmurchison: Yeah, not sure where that came from.  Fixed in all 3 test files.

imap/jmap_calendar.c
ksmurchison: Good call. Fixed

rsto (approved):

PR #4678, opened 2023-10-06
various squat fixes

### LINKED ISSUE #4660: squatter fatal error, assertion failed
When testing a new server, I've hit the same bug as in issue #4336, but as there was no real solution, I started debugging.

After many hours digging through the code and adding debugging fprintf's, I found the reason for the failed assertion.

The error happens when the existing squat index is read and then used to build the new index. Specifically, when doc_ID_map_lookup() is called with a docID that is equal to doc_ID_map->max, doc_ID_map_lookup() returns the data after the last initialised member of the array.  This memory spot is properly allocated, but contains random data.

This patch  fixes the issue:
```
diff -up cyrus-imapd-3.4.1/imap/squat_build.c.maplookup cyrus-imapd-3.4.1/imap/squat_build.c
--- cyrus-imapd-3.4.1/imap/squat_build.c.maplookup	2021-05-05 05:21:59.000000000 +0200
+++ cyrus-imapd-3.4.1/imap/squat_build.c	2023-09-21 14:17:12.851281278 +0200
@@ -355,7 +355,7 @@ static void doc_ID_map_add(struct doc_ID
             xrealloc(doc_ID_map->map, doc_ID_map->alloc * sizeof(int));
     }
     if (exists) {
-        doc_ID_map->map[doc_ID_map->max++] = doc_ID_map->new++;
+        doc_ID_map->map[doc_ID_map->max++] = ++doc_ID_map->new;
     } else {
         doc_ID_map->map[doc_ID_map->max++] = 0; /* Does not exist in new index */
     }
@@ -366,7 +366,7 @@ static int doc_ID_map_lookup(struct doc_
     if ((docID < 1) || (docID > doc_ID_map->max))
         return (0);
 
-    return (doc_ID_map->map[docID]);
+    return (doc_ID_map->map[--docID]);
 }
 
 static int copy_docIDs(void *closure, SquatListDoc const *doc)
```

[edited to wrap patch in ```'s -- ellie]

elliefm:
Fixes three bugs found in the Squat search backend while investigating #4660  

1) Reindexes were always incremental, regardless of whether or not the `-i` option was passed to squatter.  The `flags` argument handling has been fixed.
2) Messages parts that weren't supposed to be indexed were indexed anyway due to state not being set/checked/cleared properly in `begin_part()`/`append_text()`/`end_part()`.  The affected text was indexed into the wrong document if you were lucky, or corrupted the index if you weren't.  The state management has been fixed by making sure to check and clear `d->doc_name`.
3) `doc_ID_map` started at zero, but its callers and bounds checks expected it to start at 1, leading to accesses past the end of the valid data, and the assertion failure from the original report, when attempting an incremental reindex of a mailbox containing corruption as per 2.  `doc_ID_map` now starts at one, and the functions for managing it assert their invariants.

**Reviewers:** this will make the most sense when reviewed commit-by-commit

Closes #4660

### DIFF
--- a/cassandane/Cassandane/Cyrus/SearchSquat.pm
+++ b/cassandane/Cassandane/Cyrus/SearchSquat.pm
@@ -162,6 +162,12 @@ sub test_one_doc_per_message
                             body => $body) || die;
     }
 
+    # make enough other messages such that an incremental reindex
+    # will need to realloc doc_ID_map
+    for (1..50) {
+        $self->make_message() || die;
+    }
+
     $self->run_squatter();
 
     my @tests = ({
@@ -202,16 +208,16 @@ sub test_one_doc_per_message
 
     push @tests, {
         search => ['body', 'term5'],
-        wantUids => [5],
+        wantUids => [55],
     }, {
         search => ['body', 'term6'],
-        wantUids => [6, 7],
+        wantUids => [56, 57],
     }, {
         search => ['body', 'term7'],
         wantUids => [],
     }, {
         search => ['body', 'term8'],
-        wantUids => [8],
+        wantUids => [58],
     };
 
     # better not be any off-by-one errors in search results!
@@ -286,8 +292,9 @@ sub test_incremental
     my $imap = $self->{store}->get_client();
     my $err;
 
-    # some initial messages
-    for (1..5) {
+    # some initial messages - enough to definitely force a doc_ID_map realloc
+    # when incrementally reindexing later
+    for (1..50) {
         $self->make_message();
     }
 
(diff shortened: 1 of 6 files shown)


rsto (approved):

elliefm:
Just realised I failed to update a detail in the incremental reindex test: one of the bugs I fixed was only reproduceable in mailboxes that contained enough messages to need the doc_ID_map allocation to grow.  I verified that it was good during development by reducing the initial allocation from 50 to 1, such that it could be exercised with only a few messages.  But when I restored the initial allocation back to its previous value of 50, I didn't make corresponding changes to the test loop's to force it to cross the threshold.  So the bug is fixed, but the tests might not detect a future regression.  I'll fix that up.

elliefm:
@rsto I've rebased on current master, but the only real change since your last review is the last commit -- can you review that too please?  Thanks

rsto (approved):

elliefm:
Depends on #3316, which can't be backported because it in turn depends on other significant search refactors.  So it looks like these (and later) Squat fixes can't be backported to 3.2.

PR #4680, opened 2023-10-07
imap/httpd: remove DIGEST-MD5

dilyanpalauzov:
DIGEST-MD5 is declared Historic by RFC 6331 and is removed from Cyrus SASL by https://github.com/cyrusimap/cyrus-sasl/commit/2ce03b618ad60338e8f9094f5b7ba7ee0aef40bf.


dilyanpalauzov:
In docsrc/imap/rfc-support.rst I have replaced RFC 7615 with RFC 9110 under “RFC 2617 updated by…“, as the latter replaces the former.

docsrc/assets/man-imtest.rst does still mention the paramters `--realm` and `--reauth-attempts` but does not mention mechanisms, which use these parameters.

imap/httpd.h
```diff
@@ -185,7 +183,6 @@ struct auth_scheme_t {
 /* Auth scheme identifiers */
 enum {
     AUTH_BASIC        = (1<<0),
-    AUTH_DIGEST       = (1<<1),
     AUTH_SPNEGO       = (1<<2),
```
ksmurchison, line 186: Can you update this PR to either add a comment that (1<<1) was deprecated, or adjust the values for AUTH_SPNEGO, etc?  Having values of the enum jump from (1<<0) to (1<<2) with now explanation would look weird.

ksmurchison (changes requested):
This looks complete and correct.  I have just one comment, and once fixed, I will approve.

ksmurchison (approved):

PR #4681, opened 2023-10-08
cassandane: skip tests that JMAP-TestSuite will have skipped anyway

elliefm:
There's a few tests that JMAP-TestSuite skips when it detects the backend is Cyrus, but Cassandane still has to set up and then tear down a Cyrus instance for these tests because it doesn't know they'll be skipped.

This PR gives knowledge of that to Cassandane, so it too can skip them.


ksmurchison (approved):
LGTM

rsto (approved):

PR #4682, opened 2023-10-09
add xzrealloc

elliefm:
Adds an `xzrealloc()` function that's like `xrealloc()` but also zeroes out the newly allocated bytes if the allocation grew in size.  n.b. It needs one more argument than regular `xrealloc()`, so it's not a drop-in replacement.

### DIFF
--- a/imap/message.c
+++ b/imap/message.c
@@ -3134,10 +3134,9 @@ static int message_read_body(struct protstream *strm, struct body *body, const c
 
         body->type = xstrdup("MULTIPART");
         do {
-            body->subpart =
-                (struct body *)xrealloc((char *)body->subpart,
-                                        (body->numparts+1)*sizeof(struct body));
-            memset(&body->subpart[body->numparts], 0, sizeof(struct body));
+            body->subpart = xzrealloc(body->subpart,
+                                      body->numparts * sizeof(struct body),
+                                      (body->numparts + 1) * sizeof(struct body));
             buf_reset(&buf);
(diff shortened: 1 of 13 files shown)


imap/message.c
```diff
@@ -3134,9 +3134,9 @@ static int message_read_body(struct protstream *strm, struct body *body, const c
 
         body->type = xstrdup("MULTIPART");
         do {
-            body->subpart =
-                (struct body *)xrealloc((char *)body->subpart,
-                                        (body->numparts+1)*sizeof(struct body));
+            body->subpart = xzrealloc(body->subpart,
+                                      body->numparts * sizeof(struct body),
+                                      (body->numparts + 1) * sizeof(struct body));
             memset(&body->subpart[body->numparts], 0, sizeof(struct body));
```
rsto, line 3140: That's not required anymore, right?

rsto (approved):
Just found one nit. And this PR again higlights how much code still uses `int` at places where we really want to use `size_t` 😔

imap/message.c
ksmurchison: Right

ksmurchison (dismissed):
Need to remove the stray memset() as pointed out by rsto.  Once done, I approve without a re-review

imap/message.c
elliefm: Hah, yeah, forgot to delete it after changing the line above.  The simpler or more repetitive the change, the dumber the mistakes 😅

PR #4684, opened 2023-10-10
cassandane needs XML::Simple

elliefm:
Quick fix to get CI working again.  The better fix would be to have the docker image pull libxml-simple-perl from apt.


ksmurchison (approved):
Whatever works

PR #4686, opened 2023-10-15
bookworm fixes

elliefm:
This PR contains all the miscellaneous fixes I need to debug and then fix the build/test issues under Debian Bookworm on x86_64.  Best reviewed commit-by-commit.

@rsto IIRC you're on the new Apple platform rather than x86_64.  Can you build and test this on your Bookworm VM and see if it works there too?  Thanks!


elliefm:
When building on bookworm, you won't get "pcre" support for now, because Debian has switched to a newer pcre that Cyrus doesn't recognise.  I'll address that later in a separate PR.

ksmurchison (approved):
This doesn't break my Fedora build.

mosvald:
Thanks for this commit and for pointing out the related Perl change behind it https://github.com/cyrusimap/cyrus-imapd/pull/4686/commits/77cbcc29872442195c583a06c27a8e9caa2b5244

All syslog-dependent tests are passing now!

PR #4687, opened 2023-10-16
jmap_calendar: fix bug erasing default alerts on update

rsto:
This fixes a bug where updating just one of the Calendar defaultAlertsWithTime and defaultAlertsWithoutTime properties resulted in erasing the contents of the other.


ksmurchison (approved):
LGTM

PR #4689, opened 2023-10-17
<libgen.h> is required for basename(3)

obache:
From missing a header,  `basename()` may be recognized as return type `int`.

### DIFF
--- a/imap/relocate_by_id.c
+++ b/imap/relocate_by_id.c
@@ -54,6 +54,7 @@
 #include <syslog.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <libgen.h>
 
 #include "dav_db.h"
 #include "global.h"


elliefm:
My man page for basename(3) says:

> There  are  two  different  versions of basename() - the POSIX version described
> above, and the GNU version, which one gets after
> 
>                #define _GNU_SOURCE         /* See feature_test_macros(7) */
>                #include <string.h>
> 
> The GNU version never modifies its argument, and returns the empty  string  when
> path  has  a trailing slash, and in particular also when it is "/".  There is no
> GNU version of dirname().
> 
> With glibc, one gets the POSIX version of  basename()  when  <libgen.h>  is  in‐
> cluded, and the GNU version otherwise.

On Linux, my config.h has `#define _GNU_SOURCE 1`, and relocate_by_id.c includes both config.h and string.h, so it will currently be using the GNU version of the function.  Including libgen.h will switch Linux to using the POSIX version of the function.

Looking at some other sources where we use `progname = basename(argv[0])`:

* imap/relocate_by_id.c
* imap/jmap_expire.c
* imap/reconstruct.c
* imap/cyrdump.c
* imap/cyr_expire.c

reconstruct, cyrdump, and cyr_expire all include libgen.h, and get the POSIX behaviour.  relocate_by_id and jmap_expire do not, and get the GNU behaviour.  Based on the older tools including libgen.h, I think we intend to have the POSIX behaviour here, and the libgen detail was just missed in the newer tools.  So libgen should be added to both relocate_by_id and jmap_expire.

@ksmurchison @rsto What do you think?

ksmurchison:
@elliefm yes, I would expect that we want the POSIX variant. I had no intention of pulling in any GNU-specific variant.

ksmurchison (approved):

PR #4691, opened 2023-10-17
cyr_cd.sh: change to POSIX sh style

obache:
`function` keyword is not required to declare functions.
Additionally, prevent to overwrite global scope variables.


ksmurchison (approved):
I'm not a shell programming expert, but LGTM

PR #4693, opened 2023-10-18
master ready file

elliefm:
This makes master(8) write out a ready file at the same time as it logs its "ready for work" message.  For things that want to know when master is ready to start servicing clients, monitoring the ready file (perhaps with inotify) should be a bit cleaner than tailing syslog looking for the message.

Adds `master_ready_file` to imapd.conf, defaulting to `{configdirectory}/master.ready`.  Also adds a `-r ready_file` option to master to override the configured path.

Adds cassandane tests for the case where the ready file does not yet exist, and the case where it already exists.  No test for the `-r ready_file` option, because it would take a lot to plumb that through.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Master.pm
+++ b/cassandane/Cassandane/Cyrus/Master.pm
@@ -40,6 +40,7 @@
 package Cassandane::Cyrus::Master;
 use strict;
 use warnings;
+use File::stat;
 use POSIX qw(getcwd);
 use DateTime;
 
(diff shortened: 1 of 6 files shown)


elliefm:
I'm kind of tempted to tidy up the pidfile handling while I'm in here.  Currently, the pidfile path is a compile-time constant controlled by the `--with-pidfile` configure option, and overridden with a `-p` command line argument to master.  It doesn't make much sense for it to be a compile-time constant, since the pid is a property of the running instance, not the installed executable.  Its documentation is inconsistent too, and nobody's complaining about that, so I wonder how many people even use it.

I would add a `master_pid_file` option paralleling `master_ready_file`, default it to `{configdirectory}/master.pid`, keep the `-p pidfile` override, and discard the `--with-pidfile` configure option.  What do we think?

Since we normally run multiple instances per installation, presumably we're always using the `-p pidfile` option to master anyway, so the default changing won't affect us.  But it would give us a future option to simplify our setup, by setting the pidfile in imapd.conf instead of in the init script that invokes master.

ksmurchison (approved):
LGTM.

I agree that having pidfile a compile-time option is silly.  Making config option makes sense to me.

rsto (approved):

brong (approved):
This looks fine - should we just merge it and do a separate PR for the pidfile changes?  (I don't mind either way, and the pidfile being a config option makes sense to me)

elliefm:
If we want it urgently we can land it as is, and do the pidfile stuff separately.  But if we don't want it urgently, I'd prefer to hold it open for now and do the pidfile stuff in the same PR, to keep the related work together.

elliefm:
I've made the pidfile changes I proposed, can this be re-reviewed please?  Only the last two commits are new.

rsto (approved):
Looks good to me. Up to @robmueller if it meets the requirements, though.

robmueller:
This looks good to me, and IMHO it will be nice to remove the current hacky tail the logfile until we see /ready for work/

PR #4698, opened 2023-10-24
search_query: fix filtering unindexed messages in squat backend

### LINKED ISSUE #4692: Not yet squattered emails are always returned on any search (3.4 and later)
Hi, I've been digging this problem using gdb but I hardly can figure out what's happening here.

Using squatter mode in imapd.conf, when recent emails arrive and are not yet been squattered, any search will return the correct result of the squattered emails + all of these recent emails, which are not filtered considering the query.
This can be easily simulated by commenting out the rolling squatter from cyrus.conf, then sending a couple of new emails to the test account: try to do queries and you will always receive these new emails as a result, even if they don't match the query expression.

I discovered that the problem is during subquery_run_indexed() (search_query.c), after running the squatter query via bx->run() it will run the subquery_post_enginesearch callback for any folder. Here it will run index_search_evaluate(), where at the beginning runs search_expr_always_same: this returns 1 because e->op is SEOP_TRUE, causing the code to not consider the original query and going on adding the uids as a match.

This search expression is sent to index_search_evaluate from sub->expr, which is the one containing SEOP_TRUE, while I've seen previous code checking the "op" inside sub->indexed, I don't know what this means actually.
I also can't find why this expr is SEOP_TRUE when the search is a SEOP_MATCH (justs a search "FROM abcd"), and infact sub->indexed reflects it.

@rsto @elliefm you may already have checked something in the past with me about it, but maybe with these new considerations we may find where the bug is?

Here is  the previous issue: #4224 

Gabriele

rsto:
The squat backend returns false positives for unindexed messages, but filtering these messages has been broken.

This patch fixes that based on a contribution of @gbulfon. Thanks!

Fixes #4692

### DIFF
--- a/cassandane/Cassandane/Cyrus/SearchSquat.pm
+++ b/cassandane/Cassandane/Cyrus/SearchSquat.pm
@@ -438,4 +438,28 @@ sub test_relocate_legacy_nosearchdb
     $self->assert_equals(0, scalar @files);
 }
 
+sub test_unindexed
+    :SearchEngineSquat :min_version_3_4
+{
+    my ($self) = @_;
+    my $imap = $self->{store}->get_client();
+
+    $self->make_message("needle 1", body => "needle") || die;
+    $self->make_message("xxxxxx 2", body => "xxxxxx") || die;
+
+    $self->run_squatter;
+
+    my $uids = $imap->search('fuzzy', 'body', 'needle') || die;
+    $self->assert_deep_equals([1], $uids);
+
+    $self->make_message("needle 3", body => "needle") || die;
+    $self->make_message("xxxxxx 4", body => "xxxxxx") || die;
+
+    # Do not rerun squatter. Make sure search only returns
+    # a matching unindexed message.
+
+    $uids = $imap->search('fuzzy', 'body', 'needle') || die;
+    $self->assert_deep_equals([1,3], $uids);
+}
+
 1;
--- a/imap/search_query.c
+++ b/imap/search_query.c
@@ -137,6 +137,11 @@ EXPORTED void search_query_free(search_query_t *query)
 
 /* ====================================================================== */
 
+static int subquery_run_one_folder(search_query_t *query,
+                                   const char *mboxname,
+                                   search_expr_t *e);
+
+/* ---------------------------------------------------------------------- */
 
 /*
  * Find the named folder folder.  Returns NULL if there are no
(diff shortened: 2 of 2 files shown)


cassandane/Cassandane/Cyrus/SearchSquat.pm
```diff
@@ -438,4 +438,28 @@ sub test_relocate_legacy_nosearchdb
...
+sub test_unindexed
+    :SearchEngineSquat :min_version_3_4
+{
+    my ($self) = @_;
+    my $imap = $self->{store}->get_client();
+
+    $self->make_message("needle 1", body => "needle") || die;
+    $self->make_message("xxxxxx 2", body => "xxxxxx") || die;
+
+    $self->{instance}->run_command({cyrus => 1}, 'squatter');
```
elliefm, line 450: You could use `$self->run_squatter();` here now ([implementation](https://github.com/cyrusimap/cyrus-imapd/blob/master/cassandane/Cassandane/Cyrus/SearchSquat.pm#L72-L91)).  I added it just recently for the other squat fixes.

imap/search_query.c
```diff
@@ -724,7 +748,9 @@ static void subquery_run_indexed(const char *key __attribute__((unused)),
 
     bx = search_begin_search(query->state->mailbox, opts);
     if (!bx) {
-        r = IMAP_INTERNAL;
+        r = config_getenum(IMAPOPT_SEARCH_ENGINE) == IMAP_ENUM_SEARCH_ENGINE_SQUAT ?
+            subquery_run_one_folder(query, index_mboxname(query->state), sub->indexed) :
+            IMAP_INTERNAL;
```
elliefm, line 753: The ternary is hard to read here -- there's a lot of text to read for a single statement.  What about something like:

```suggestion
        if (config_getenum(IMAPOPT_SEARCH_ENGINE) == IMAP_ENUM_SEARCH_ENGINE_SQUAT) {
            r = subquery_run_one_folder(query, index_mboxname(query->state), sub->indexed);
        }
        else {
            r = IMAP_INTERNAL;
        }
```

or maybe even just:

```suggestion
        r = IMAP_INTERNAL;
        if (config_getenum(IMAPOPT_SEARCH_ENGINE) == IMAP_ENUM_SEARCH_ENGINE_SQUAT)
            r = subquery_run_one_folder(query, index_mboxname(query->state), sub->indexed);
```

elliefm (approved):
Looks good!  Couple of nits/suggestions, but it doesn't need re-review as long as CI passes.

cassandane/Cassandane/Cyrus/SearchSquat.pm
rsto: I saw, just didn't bother. But I updated the code now to use it.

PR #4699, opened 2023-10-24
http_carddav.c: CardDAV:addressbook_query using libicalvcard

ksmurchison:
also concatenate all NICKNAMEs for carddav_db record


rsto (approved):
LGTM. I realized this requires the latest libicalvcard version, otherwise it will core dump miserably.

PR #4701, opened 2023-10-25
fix print format for timespec and timeval

obache:
Change to use `TIME_T_FMT` for `tv_sec` (time_t), and fix to zero prefix for `tv_nsec` and `tv_usec`.
Use `unsigned long long` for HEX format of `time_t`, `int` may be short.


elliefm (approved):

elliefm:
Thanks!

PR #4705, opened 2023-10-30
search_engines: fix batching

brong:
This was lost in ba7fc3ab14f0a843df3f95ec1249115fa7744c54 when refactoring to allow better lock handling.


rsto (approved):
Obviously the right thing to do!

PR #4706, opened 2023-10-30
Convdb zero singlecid

brong:
This allows us to zero out just a single CID, rather than everything - it's to help fix up the "same email, multiple CIDs" bug that we currently have - we can just nuke each of those cids.

Ideally you'd give a list of CIDs to zero, but that's a bunch more work to build, and this will be fast enough because usually it's just a couple of CIDs for each user.  Zeroing EVERYTHING is much slower and breaks the mailbox for a bit.

### DIFF
--- a/imap/ctl_conversationsdb.c
+++ b/imap/ctl_conversationsdb.c
@@ -196,14 +196,12 @@ static int delannot_cb(const char *mboxname,
 	remove = 0;
 
         // is the value this CID, remove this
-        if (value->len == 16) {
-            cid = strtoull(value->s, 0, 16);
-            if (zerocid == cid) remove = 1;
-        }
+        parsehex(value->s, NULL, 16, &cid);
+        if (zerocid == cid) remove = 1;
 
         // is the key this CID, remove this, 
-        if (conversation_id_decode(&cid, entry + strlen(IMAP_ANNOT_NS) + strlen("newcid/")))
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):

imap/ctl_conversationsdb.c
```diff
@@ -854,7 +870,7 @@ int main(int argc, char **argv)
     int recursive = 0;
 
     /* keep in alphabetical order */
-    static const char short_options[] = "AC:FRST:bdruvz";
+    static const char short_options[] = "AC:FRST:bdruvzZ:";
```
elliefm, line 873: please, this list _was_ ascii sorted!

imap/ctl_conversationsdb.c
```diff
@@ -869,6 +885,7 @@ int main(int argc, char **argv)
         { "undump", no_argument, NULL, 'u' },
         { "verbose", no_argument, NULL, 'v' },
         { "clear", no_argument, NULL, 'z' },
+        { "clearcid", required_argument, NULL, 'Z' },
```
elliefm, line 888: as was this struct!

imap/ctl_conversationsdb.c
elliefm: there's even a `/* keep in alphabetical order */` right above it

PR #4707, opened 2023-10-31
lmtpd: support JMAP attachment body search in Sieve

rsto:
This patch enables JMAP Email attachmentBody search in `jmapquery` Sieve filters.


ksmurchison (approved):

rsto:
Now rewrote this to only call the attachment extractor if the jmapquery filter actually requires it.

imap/index.c
```diff
@@ -5820,32 +5820,42 @@ static int getsearchtext_cb(int isbody, charset_t charset, int encoding,
...
+                    "attachment %s (%s/%s)",
                     message_guid_encode(content_guid),
-                    type, subtype, error_message(r));
+                    type, subtype);
 
-            if (str->flags & INDEX_GETSEARCHTEXT_PARTIALS) {
-                /* mark message as partially indexed and continue */
-                str->indexlevel |= SEARCH_INDEXLEVEL_PARTIAL;
-                r = 0;
+            str->indexlevel |= SEARCH_INDEXLEVEL_PARTIAL;
```
elliefm, line 5829: Is it okay that we're setting it as partial here without checking that partials are allowed?  Ahh, I see a comment later explaining that it is okay.

imap/index.c
```diff
@@ -6056,7 +6066,8 @@ EXPORTED int index_getsearchtext(message_t *msg, const strarray_t *partids,
     if (r == IMAP_OK_COMPLETED) r = 0;
 
     /* Log erroneous or partially indexed message */
-    if (r || (str.indexlevel & SEARCH_INDEXLEVEL_PARTIAL)) {
+    if ((r || (str.indexlevel & SEARCH_INDEXLEVEL_PARTIAL)) &&
+        !(flags & INDEX_GETSEARCHTEXT_NOLOG_PARTIALS)) {
```
elliefm, line 6070: This looks like turning off logging of partials will also turn off logging of errors.  Is that intended?

imap/index.h
```diff
@@ -349,8 +349,19 @@ extern int index_search_evaluate(struct index_state *state, const search_expr_t
...
+ * returned an error. If this flag is not set, then the first attachment
+ * extractor error causes getsearchtext to return with an error. */
+#define INDEX_GETSEARCHTEXT_ALLOW_PARTIALS (1<<1)
+/* Do not log a warning if messages could only be indexed partially
+ * (implies allowing partial message indexes) */
+#define INDEX_GETSEARCHTEXT_NOLOG_PARTIALS \
+    (INDEX_GETSEARCHTEXT_ALLOW_PARTIALS | (1<<2))
+/* Disable calling the attachment extractor. The message is marked as
+ * partially indexed, regardless of the ALLOW_PARTIALs flag. */
+#define INDEX_GETSEARCHTEXT_NOCALL_ATTACHEXTRACT (1<<3)
```
elliefm, line 364: Ahh. I see that this explains that allow_partials is ignored when the extractor is disabled

PR #4709, opened 2023-10-31
Dblookup email2uids regression

ksmurchison:


brong (approved):
Looks good to me! (spoke in person too)

PR #4710, opened 2023-11-01
fix userflags corruption crash

elliefm:
This fixes a crash that can occur during cleanup up after copy/append failures when the mailbox on which the copy/append failed happened to have custom userflags.  Seen in the wild a couple of times at Fastmail.  @wolfsage found a way to reproduce it, which the Cassandane test is based on.  The cause of the crash turned out to be memory corruption resulting from calling the `xzfree()` macro -- which evaluates its `ptr` argument multiple times -- with an argument containing side effects.

It also hardens up the `xzfree()` macro somewhat to hopefully prevent this sort of thing in the future.  With the compiler options I use, the original misuse is now detected at compile time as a `sequence-point` warning, due to the comparison in the assert expression having an undefined evaluation order.  If the compiler misses it, it would still be caught at runtime due to the two evaluations not matching and the assertion failing, and hopefully the developer making the mistake would notice some test failing.

The new test reproduces the crash if run without the other fixes, and succeeds with them.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Flags.pm
+++ b/cassandane/Cassandane/Cyrus/Flags.pm
@@ -1551,4 +1551,62 @@ sub test_setseen_notify
     $self->assert($payload2->{modseq} > $payload1->{modseq}, "modseq has increased: $payload2->{modseq} > $payload1->{modseq}");
 }
 
+sub test_userflags_crash
+    :Conversations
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):
Great work on this!

rsto (approved):

PR #4712, opened 2023-11-02
Subscriptions bump folderstate

brong:
This is actually two fixes!

First, I discovered that an entire important test was being killed with a `return;` at the top!  Ouch.  I discovered this when trying to add to it.

I had to fix that test, during which I discovered a bug with isSubscribed == false, so there's a fix for that.

Then the actual fix, which was discovered when testing with the TMail client - our server wasn't bumping the state string for the mailbox when changing the isSubscribed value.


rsto (approved):
LGTM

elliefm:
CI failed for this.  Please pay attention to CI failures, they aren't there for nothing!

PR #4713, opened 2023-11-02
jmap_vacation.c: restore this to functional

ksmurchison:


rsto (approved):
LGTM, but the _(don't know when it broke)_ isn't really necessary in the commit message

PR #4715, opened 2023-11-03
Revert folder modseq bump

brong:
This broke the tests- sorry @elliefm.  Reverting the offending commits


elliefm (commented):
This has #4714's change embedded in it too, is that deliberate or an accident?

elliefm (approved):

PR #4718, opened 2023-11-03
imapd.c: move LIST-METADATA into the "standard" section of extensions

ksmurchison:
Expecting that this will become a standard since its a WG doc


elliefm (approved):
Seems fine.  Does rfc-support.rst need updating too?  It looks like it has an `IETF RFC Drafts` section for this sort of thing.

For whatever it's worth, we have a magic linkifier for drafts -- search for `:draft:` under docsrc for examples of how to use it (and it's implemented under `extlinks` in docsrc/conf.py).  For some reason rfc-support.rst doesn't use it, but maybe it should.

rsto (approved):

PR #4720, opened 2023-11-04
Add a 'replicaonly' config option and directory

brong:
If the config is set, or there is a file with the username in the directory, then abort if any attempt is made to call `nextmodseq` or `nextuidvalidity`, instead only sync_server writes are allowed.

This branch also fixes two bugs found by this test already.

### DIFF
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -2217,6 +2217,10 @@ If all partitions are over that limit, this feature is not used anymore.
 /* If enabled, all IMAP, POP and JMAP connections are read-only,
  * no writes allowed. */
 
+{ "replicaonly", 0, SWITCH, "3.9.0" }
+/* If enabled, nothing is allowed to increase modseq or uidvalidity,
+ * no writes allowed at all. */
+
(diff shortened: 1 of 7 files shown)


ksmurchison (approved):
LGTM.
Perhps the config option description should saythat no changes are allowed other than via syncserver?

brong:
Yeah, that seems reasonable.  I'll be extending calalarmd with it as well.

lib/imapoptions
```diff
@@ -2217,6 +2217,10 @@ If all partitions are over that limit, this feature is not used anymore.
 /* If enabled, all IMAP, POP and JMAP connections are read-only,
  * no writes allowed. */
 
+{ "replicaonly", 0, SWITCH, "3.9.0" }
+/* If enabled, nothing is allowed to increase modseq or uidvalidity,
+ * no writes allowed at all. */
+
```
elliefm, line 2223: Use `"UNRELEASED"`, don't guess and hardcode a version number

elliefm (commented):
This needs a file under changes/next describing the new config option and behaviour for the release notes

PR #4721, opened 2023-11-05
Calalarmd replicaonly

brong:
This adds both tests for replicaonly mode (disabled for the tests which MUST actually screw with the replica to test splitbrain recovery!), and support to calalarmd to skip processing for users who are in replicaonly mode.

### DIFF
--- a/cassandane/Cassandane/Cyrus/CaldavAlarm.pm
+++ b/cassandane/Cassandane/Cyrus/CaldavAlarm.pm
@@ -1378,7 +1378,7 @@ EOF
 }
 
 sub test_replication_at1
-    :min_version_3_0 :needs_component_calalarmd
+    :min_version_3_0 :needs_component_calalarmd :NoReplicaOnly
 {
     my ($self) = @_;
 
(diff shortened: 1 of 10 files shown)


elliefm (changes requested):
At least these three cass tests are failing under CI:

```
[FAILED] Cyrus::CaldavAlarm.replication_at1
[FAILED] Cyrus::CaldavAlarm.replication_withalarms_in_tz_with_dst
[FAILED] Cyrus::CaldavAlarm.replication_withalarms_in_tz_without_dst
```

This looks like it extends #4720, so it would be good to also catch up on the things that were missed when that was reviewed:

* imapoptions contains a bogus hardcoded release version number (use `"UNRELEASED"` for changes that are not yet in a stable release; don't guess)
* needs a changes/next file based on 0000-TEMPLATE for future inclusion in release notes

brong:
@elliefm I've added a fix to those three tests (basically: they intentionally run alarms on the replica, so of course they need `:NoReplicaOnly` config.  And I've corrected the imapoptions version (sorry, the wrong one is actually already on master), and added upgrade notes.

elliefm:
@elliefm "replicaonly" is actually already in 3.10.0-beta1, but completely unacknowledged because of the missed changes file and hardcoded imapoptions version from the original PR.  This PR fixes those issues retroactively, but it will take some finesse getting it onto the 3.10 branch.  And we'll need to backport any future replicaonly fixes to 3.10 too.

elliefm (approved):

PR #4724, opened 2023-11-06
fixes found by gcc static analyzer

elliefm:
This contains a bunch of fixes for things found by the GCC static analyzer.  It fixes a bunch of things that were definitely wrong, and also adds a few assertions where we assume things.

It does not address everything the static analyzer complained about.  The rest is in #4730


rsto (approved):
Excellent!

PR #4725, opened 2023-11-06
defaultalarms: ignore user alerts if useDefaultAlerts=true

rsto:
This PR got rewritten for new requirements: https://github.com/cyrusimap/cyrus-imapd/pull/4725#issuecomment-1824398864

~~By default, Cyrus assumes CalDAV clients do not know about the semantics of our custom X-JMAP-USEDEFAULTALERTS property. This patch adds a header for PUT to allow indicating them they do.~~

~~CalDAV clients could attempt to store a user-defined VALARM with PUT in an event, but being good clients and preserving x-properties, they might keep our X-property set to TRUE. This would result in Cyrus ignoring that user alarm. To mitigate this, Cyrus by default attempts to determine if the CalDAV client did change VALARMs in an event, and if so, it disables default alerts for this event (by setting X-JMAP-USEDEFAULTALERTS to false).~~

~~This patch adds a HTTP header "X-Cyrus-rewrite-default-alarms" which allows to suppress rewriting X-JMAP-USEDEFAULTALERTS on PUT. Its value must be set to "false" or "f", or otherwise is ignored.~~


ksmurchison (approved):
LGTM

rsto:
This PR got rewritten for new requirements:

Before this patch, user alerts triggered if default alerts were enabled for a calendar event, but the calendar of that event had no default alerts set. Now, user alerts always are ignored if useDefaultAlerts=true.

In addition, this patch now also rewrites useDefaultAlerts in PUT for new events. This makes Cyrus rewrite the useDefaultAlerts property in CalDAV. PUT for both new and updated events. A newly created event with useDefaultAlerts=true gets rewritten to useDefaultAlert=false if the calendar of this event does not have default alerts set. The rewriting for updated events is kept as-is.

elliefm:
> A newly created event with useDefaultAlerts=true gets rewritten to useDefaultAlert=false if the calendar of this event does not have default alerts set. 

If the user later sets default alerts on the calendar, what happens to existing events in that calendar which have had useDefaultAlerts=false set?

rsto:
> If the user later sets default alerts on the calendar, what happens to existing events in that calendar which have had useDefaultAlerts=false set?

They keep having useDefaultAlerts=false set.

PR #4728, opened 2023-11-07
imapd.c: log client use of BINARY and CATENATE

ksmurchison:
Over lunch with a few implementers it was asked if any clients use these extensions.  I told them that we can and will gather some data.


rsto (approved):

rjbs:
Nice one!

PR #4733, opened 2023-11-09
IMAP: add support for UIDONLY (draft-ietf-extra-imap-uidonly)

ksmurchison:


cassandane/tiny-tests/UIDonly/copy-move
```diff
@@ -0,0 +1,49 @@
...
+
+    xlog $self, "ENABLE UIDONLY";
+    $res = $imaptalk->_imap_cmd('ENABLE', 0, 'enabled', 'UIDONLY');
+    $self->assert_num_equals(1, $res->{uidonly});
+
+    xlog $self, "attempt COPY";
+    $res = $imaptalk->copy(1, $folder);
+    $self->assert_str_equals('bad', $imaptalk->get_last_completion_response());
+    # get_response_code() doesn't (yet) handle [UIDREQUIRED]
+    $self->assert_matches(qr/[UIDREQUIRED]/, $imaptalk->get_last_error());
```
elliefm, line 35: You need to escape those `[]` in the regular expression.  It looks like you intend to match on the string `"[UIDREQUIRED]"`, but at the moment you're matching on any single character from the set `UIDREQ` (with some dups in the spec).

Same comment applies in multiple places, I won't repeat it, but search and fix all of them :)

elliefm (changes requested):
Looks good, but there's one bug that's repeated a few times in the tests

cassandane/tiny-tests/UIDonly/copy-move
ksmurchison: Fixed

elliefm, line 35: Whoops, missed this one!

ksmurchison: Fixed (again)

elliefm (approved):

PR #4734, opened 2023-11-09
jmap_contact: rename Title.organization property organizationId

rsto:


ksmurchison (approved):
LGTM (assuming that you grepped everywhere that 'organization' was used)

rsto:
I did. Passes JSContact tests and the jscontact test suite on jmapio.

PR #4737, opened 2023-11-15
github-actions: report debian version

elliefm:
because I don't remember what version this runs, and this way nobody has to, it's right there in the report


elliefm:
@elliefm This needs to be backported to all the branches we run CI for.  I don't remember what those are, but I'll figure it out when I get there.

rsto (approved):

elliefm:
Backported to 3.8, 3.6, and 3.4.  We don't currently run CI for 3.2

PR #4739, opened 2023-11-17
improve cassandane getsyslog

elliefm:
This reduces the amount of boilerplate needed to check for syslog messages from Cassandane tests.

* Adds an optional pattern parameter to `getsyslog()`, so you can filter the lines without an extra step
* Adds `assert_syslog_matches()` and `assert_syslog_does_not_match()` for the common case where you just want to check whether syslog does or does not contain any lines matching some pattern.
* Updates existing tests to use the new features where it makes sense to

Note that the new assertions call `getsyslog()` themselves, so they're _not_ useful if you want to check for multiple patterns at once, because the first call will read all the available lines, and the subsequent calls won't find any.  For that sort of thing you should still use `getsyslog()` directly to read the lines, then match those lines in whatever way seems appropriate.

**Reviewers:** lots of repetitive small changes in here.  Probably easiest to review commit by commit.


ksmurchison (approved):
LGTM

PR #4740, opened 2023-11-22
fix configure.ac problems under -Werror

elliefm:
This fixes up a bunch of configure details that were misdetected due to test programs with warnings failing to compile because of `-Werror` (rather than because of the feature being unavailable)


elliefm:
@rsto Can you test this also when you get a chance please?  I don't expect anything substantial to actually change (otherwise we would have noticed this stuff earlier), but it always feels a little precarious to be messing around in the darker corners of configure.ac

PR #4741, opened 2023-11-22
fix pcre2 crash when regex_t uninitialised

elliefm:
This fixes a crash seen in the wild at Fastmail when Cyrus is built with PCRE2 rather than PCRE.

Looks like the PCRE2 `pcre2_regcomp()` has different behaviour from the POSIX or PCRE1 `regcomp()` when the pattern fails to compile.  In this case the PCRE2 version doesn't touch its regex_t argument.  If it wasn't already initialized otherwise, it will remain uninitialized, and a later call to `pcre2_regfree()` will crash.

That means we need to be sure to initialise regex_t objects, not rely on `regcomp()` to do it.

### DIFF
--- a/lib/glob.c
+++ b/lib/glob.c
@@ -58,6 +58,7 @@
 EXPORTED glob *glob_init(const char *str, char sep)
 {
     struct buf buf = BUF_INITIALIZER;
+    int r;
 
     buf_appendcstr(&buf, "(^");
     while (*str) {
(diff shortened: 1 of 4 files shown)


rsto (approved):

brong (approved):
Thanks ellie - this looks good and tests are good

lib/glob.c
```diff
@@ -109,7 +110,9 @@ EXPORTED glob *glob_init(const char *str, char sep)
     buf_appendcstr(&buf, "]|$)");
 
     glob *g = xmalloc(sizeof(glob));
-    regcomp(&g->regex, buf_cstring(&buf), REG_EXTENDED);
+    r = regcomp(&g->regex, buf_cstring(&buf), REG_EXTENDED);
+    /* XXX handle regex compilation failure properly! */
+    assert(r == 0);
```
yselkowitz, line 115:
```
lib/glob.c: In function 'glob_init':
lib/glob.c:115:5: error: implicit declaration of function 'assert' [-Werror=implicit-function-declaration]
  115 |     assert(r == 0);
      |     ^~~~~~
lib/glob.c:51:1: note: 'assert' is defined in header '<assert.h>'; did you forget to '#include <assert.h>'?
```

elliefm: Oh good catch, thanks!  I'll get a fix for that up shortly

elliefm: Fixed in #4749

PR #4742, opened 2023-11-23
Remove jmap_expire and automatically prune calendareventnotifs

rsto:
This removes the experimental `jmap_expire` executable.

Instead, it updates Cyrus to mark all JMAP CalendarEventNotification objects but the last 200 items as expunged, each time a new one is created. The maximum count can be set in the `jmap_max_calendareventnotifs` config option.


elliefm:
This one is interesting because jmap_expire only landed on master this year, and therefore hasn't yet existed in a stable release, so if we land it on master before the next stable release, then it's correct for all mention of it to just disappear.  I think this PR will need a _new_ changes/next file describing the _new_ difference (that is, that calendareventnotifs are automatically pruned rather than hanging around forever).  You might have already done this and I haven't noticed yet, I'm reacting early...

If for some unlikely reason we delay this until after 3.10 is released (with jmap_expire in it), then it will need to add a new changes/next file saying that jmap_expire has been removed, rather than just deleting the old changes/next file.  But hopefully we can avoid this case!

All that said, I think it's fine if this one lands on master during the December feature freeze, cause it'll save double-handling later.  I'm tagging this backport-to-3.10, but that tag can just be removed if it lands on master before the 3.10 branch is forked.

lib/imapoptions
```diff
@@ -1293,6 +1293,12 @@ Blank lines and lines beginning with ``#'' are ignored.
    For backward compatibility, if no unit is specified, kibibytes is assumed.
    */
 
+{ "jmap_max_calendareventnotifs", 200, INT, "UNRELEASED" }
+/* The maximum count of CalendarEventNotification objects to keep per account.
+   Any notifications exceeding this count are deleted to make room for new
+   ones. Zero or any negative number disables this limit.
+    */
+
```
elliefm, line 1301:
```suggestion
{ "jmap_max_calendareventnotifs", 200, INT, "UNRELEASED" }
/* The maximum count of CalendarEventNotification objects to keep per account.
   Any notifications exceeding this count are expunged to make room for new
   ones. Zero or any negative number disables this limit.
    */

```

I think it's better to say "expunged" rather than "deleted" here to be technically correct about what happens, since "deleted" is kind of a loaded word wrt email.

If I understand this change correctly, all we're doing here is marking the messages as expunged, and therefore in most cases cyr_expire will still need to run at some point for them to really disappear?  I don't think it's necessary to document that detail here, as long as we say "expunged", I think we can trust that the admin remembers (or can look up) what `expunge_mode` they've configured and what ramifications it has...

elliefm (commented):
Looks good.  Can we get a test for it somehow?

lib/imapoptions
rsto: done

rsto:
Updated now to include a Cassandane test

elliefm (approved):

imap/jmap_notif.c
```diff
@@ -128,12 +149,31 @@ static int append_eventnotif(const char *from,
...
+        struct mailbox_iter *iter = mailbox_iter_init(notifmbox, 0,
+                ITER_STEP_BACKWARD|ITER_SKIP_UNLINKED|
+                ITER_SKIP_EXPUNGED|ITER_SKIP_DELETED);
+        size_t n_notifications = 0;
+
         message_t *msg;
         while ((msg = (message_t *) mailbox_iter_step(iter))) {
+
+            // Notifications exceeding notifications_max must
+            // have been pruned already, no need to check.
```
brong, line 171: This seems pointless - since they've been pruned, they're going to be skipped by ITER_SKIP_EXPUNGED - so we have changed the direction of iteration for no reason

brong: (I guess it does stop reading potentially thousands if something wrote a ton of notifications in the cyr_expire window

elliefm: I also had a good hard think about this part while I was reviewing it, and eventually concluded that it was sensible even though on the surface it seemed redundant.  I've forgotten my reasoning though... probably should have written it down! 😅

elliefm:
@rsto This should land on master ASAP, because it removes a new feature which we want to pretend never existed.  Unless there's something else you're waiting for?  I see it's already been in our builds for a couple of weeks.  Not sure if I should just merge it myself....  If it's fine to merge, please do so; otherwise please tag it do-not-merge so that I know not to just do it myself, and in that case we can chat about what else is needed.

I'd like it to be merged before EOY if possible, because it'll save some double-handling once 3.10 forks in the near year.

rsto:
I kept it outside the main branch in case we want to quickly undo the change. But it's fine to merge now, so I'll do that.

Issue #4717, opened 2023-11-03
3.4.6 POP3 RETR failure for some messages


yaonos:
Hi,

Since our upgrade from v2.4.22 to v3.4.x there are some POP3 sessions that fail during the processing of a RETR command.

This issue has been observed with Thunderbird as a POP3 client.

While triyng to analyse this issue, it seems that Cyrus IMAP ends the mail with a line containing only ".\r" instead of ".\r\n" : the "\n" seems to be missing so the client does not send the remaining commands and the connection is closed by the server after a timeout.

This issue seems to only occur for some particular messages and is reproducible although the client may sometimes go on to and retrieve the next mail.

I was able to collect some session logs both on my lab servers and thunderbird with a mail sample that triggers the issue. We have a MURDER setup so I collected the logs on both the frontend and the backend, but it seems that the issue is located on the backend.

Seeing that the blat() function in imap/pop3c calls prot_printf() in order to send ".\r\n" at the end of the message, I tried to add a call to "prot_flush(popd_out);" after this prot_printf in case the "\n" was kept in a buffer and not send, but this does not change anything.

The blat() function does not appear to have been modified between 2.4.22 and 3.4.6 so the issue may come from a change in the lib/prot.c library.

Regards

[lab-backend-pop3-4036621.txt](https://github.com/cyrusimap/cyrus-imapd/files/13251804/lab-backend-pop3-4036621.txt)
[lab-frontend-pop3-3342595.txt](https://github.com/cyrusimap/cyrus-imapd/files/13251810/lab-frontend-pop3-3342595.txt)
[lab-thunderbird.txt](https://github.com/cyrusimap/cyrus-imapd/files/13251812/lab-thunderbird.txt)
[4.txt](https://github.com/cyrusimap/cyrus-imapd/files/13251815/4.txt)

yaonos:
Hello again,

Since my last message, I foud out that I misinterpreted Cyrus IMAP debug log and that the "\n" is not missing but is not sent on the wire in the same packet as the ".\r". This may not be correctly handled by thunderbird.

I tried another patch where I added a call to prot_flush() before sending the final ".\n\r" and this seems to resolve my issue. I will deploy this patched version on our production servers next week in order to confirm the resolution.

I have attached my patch to this message. Given that is is a workaround for a thunderbird's bug, this patch may not have its place Cyrus IMAP's official sources.

I'll leave a follow up later to let you know if the fix is confirmed.

Regards

[...]

PR #4743, opened 2023-11-23
pop3d: try to not split ".\r\n" across packet boundary

elliefm:
Works around some Thunderbird bug.  Based on patch provided and tested in #4717 by @yaonos

Fixes: #4717

### DIFF
--- a/imap/pop3d.c
+++ b/imap/pop3d.c
@@ -2073,6 +2073,9 @@ static int blat(int msgno, int lines)
     /* Protect against messages not ending in CRLF */
     if (buf[strlen(buf)-1] != '\n') prot_printf(popd_out, "\r\n");
 
+    /* Clients may become confused when the .\r\n sequence gets split
+     * up across packets.  Flush first to prevent that. */
+    prot_flush(popd_out);
     prot_printf(popd_out, ".\r\n");
 
     /* Reset inactivity timer in case we spend a long time


ksmurchison (commented):
Not sure I understand why a flush before and after is needed

elliefm:
As I understand it, the problem occurs when the protstream's buffer was already within a byte or two of a packet boundary, such that when the ".\r\n" sequence is written, part of it is at the end of one packet and the rest is at the start of the next.  Apparently Thunderbird gets confused when that specific sequence is split up.

I think the first flush causes whatever was already in the buffer to be sent as is, rather than waiting for more bytes to fill out the packet.  Then the ".\r\n" becomes the start of the next packet, and stays intact.  The unconditional flush probably results in more packets than strictly necessary, which might be preventable by examining the pending buffer length, knowing the packet size, and only flushing if it were actually necessary to prevent a split.  But I don't think anyone wants to write and test that, and this works.

I suspect the second flush might be redundant, but it was in the patch that OP provided and tested, so it's in mine too.

yaonos:
The second flush may Indeed bé redundant.
As explained by @elliefm the goal of the first flush is to avoid spliting ".\r\n" on 2 packets.
Regards

ksmurchison:
Yes, I believe it is redundant since flushonread has been set on the input stream and we iterate back to the top of cmdloop() as soon as we exit blat()

elliefm:
Since we all agree it's probably redundant, I've dropped the second flush

ksmurchison (approved):

Issue #1763, opened 2016-12-27
Backups for SMB (lock entire server for a moment while taking a snapshot)


brong:
Via email from Anatoli <me@anatoli.ws>

1. Small (SMB) deployments with a single server and somehow limited physical resources (e.g. disk space).

Here as an example comes the excellent backup mechanism Ellie implemented that suits well the needs of medium to large deployments, but IMO that's not the best approach for small deployments, as it requires a separate server or, if ran at the same server just for the safe data-to-disk synchronization, twice the disk space.

A better approach for small deployments, as I see it (and I believe it's highly demanded by the community), would be to have an executable that would instruct Cyrus daemon to synchronize to disk all the internal structures and lock (stop writing to disk) for a defined period. The lock could be implemented by hanging on network write requests or by writing them to temporary files, or by accumulating the changes in memory (the latter approach has a potential for data loss).

Once the flush is performed and the lock is applied, a (custom) backup script could create a snapshot of the partition that would hold the Cyrus data in a safe-to-backup state. Immediately after creating the snapshot, the lock would be released and the daemon would continue its normal operation. Then the backup script would be able to safely backup the data, e.g. create an incremental backup and upload it to some external storage, then destroy the snapshot.

Usage example: cyrus_sync2disk --lock=5 -> returns 0 when the data is synced and a lock for 5 seconds is obtained. cyrus_sync2disk --unlock -> returns 0 if the lock has been released and 1 if there was no active lock (e.g a previous lock has expired), so the backup script knows if it performed the required operations with the lock still in place or if it should perform the lock-snapshot-unlock operation again. The short timeout is to protect the daemon from an infinite lock if a backup script fails to unlock it.

brong:
response from Vladislav Bogdanov <bubble@hoster-ok.com>

The better technique is to run snapshot-creating command from inside the 
utility which makes the flush and lock:

cyrus_sync2disk --lock-over 'lvcreate -s ...'

That way no race conditions apply. Also, lock is held during the minimal 
possible time.

brong:
I understand what you want here, and I mostly like the idea.

The one thing that gives me pause is that it requires a single lock against ALL cyrus processes.  Right now, there's no global lock that processes take while making changes, and we'd need to add one.  I would want to make it be something that needs to be turned on in config so that sites which DON'T need it don't have to pay the extra locking cost.

[...]

PR #4744, opened 2023-11-26
Global lock

brong:
This addresses https://github.com/cyrusimap/cyrus-imapd/issues/1763, which has been outstanding for a LONG time, but I'm finally getting to it because I was working on mailbox locking recently, and the user locks and mailbox locks are now consistent enough that it's possible to put this structure into place.

Also, I finally figured out how to make this work sensibly - it's pretty cool - because new locks ALWAYS go on the front of open_mboxlocks, and you can't change a lock from shared to exclusive - it's safe to have shared locks while running this kind of snapshot command - so I could insert the GLOBAL lock before the first exclusive lock of any type, and remove it if it's the front lock, and it guarantees coverage of all writes.  Super simple.  And reads can keep running.

One could also make it so even reads can't run, by just removing the restriction that this only applies to EXCLUSIVE locks.  Or have two locks, one that goes before any reads, and one that only goes before writes.  If we did this, we'd add another flag to cyr_withlock_run to say which one it needed, so you could even lock and run modifying commands!

But that's an enhancement for later, this is good for now!

### DIFF
--- a/docsrc/imap/reference/manpages/systemcommands/cyr_withlock_run.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/cyr_withlock_run.rst
@@ -1,0 +1,85 @@
+.. cyrusman:: cyr_withlock_run(8)
+
+.. author: Bron Gondwana
+
+.. _imap-reference-manpages-systemcommands-cyr_withlock_run:
+
+====================
+**cyr_withlock_run**
+====================
+
+is used to run system command with a lock held
+
+..  warning::
+
+    This command runs as the Cyrus user, so if you need to call something
+    as root, the cyrus user may need sudo privileges.
+
+Synopsis
+========
+
+.. parsed-literal::
+
+    **cyr_withlock_run** [ **-C** *config-file* ] [ **-u** *userid* ] cmd args...
+
+Description
+===========
+
+**cyr_withlock_run** will run the command and arguments provided on the
+command line, either locking the entire server, or a particular user, so
+no changes can be made to it while this command is running.
+
+WARNING: since this takes an exclusive lock, it will deadlock if the command tries
+to connect to Cyrus via IMAP/JMAP/etc and run commands that take exclusive locks.
+You can run cyrus commandline tools so long as the environment is passed through.
+
+It is most useful for running an external filesystem snapshot command, which can
+be run safely, knowing that files aren't in a partial state.
+
+
+**cyr_withlock_run** |default-conf-text|
+
+Options
+=======
+
+.. program:: cyr_withlock_run
+
+.. option:: -C config-file
+
+    |cli-dash-c-text|
+
+.. option:: -u, --user <userid>
+
+   Lock just the specified userid rather than the global lock.
+
+   NOTE: if you run for the global lock, then your server must have the
+   `global_lock` config option set, or this command will fail with an
+   error as it can't get a guaranteed lock.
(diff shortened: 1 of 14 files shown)


imap/cyr_withlock_run.c
```diff
@@ -0,0 +1,138 @@
+/* cyr_pwd.c -- current working directory within a spool dir
```
elliefm, line 1: whoops ;)

docsrc/imap/reference/manpages/systemcommands/cyr_withlock_run.rst
```diff
@@ -0,0 +1,77 @@
+.. cyrusman:: cyr_withlock_run(8)
+
+.. author: Bron Gondwana
+
+.. _imap-reference-manpages-systemcommands-cyr_withlock_run:
+
+================
+**cyr_withlock_run**
+================
```
elliefm, line 9: Sphinx will complain about the overline and underline not being the same length as the heading

```suggestion
====================
**cyr_withlock_run**
====================
```

docsrc/imap/reference/manpages/systemcommands/cyr_withlock_run.rst
```diff
@@ -0,0 +1,77 @@
...
+    **cyr_withlock_run** [ **-C** *config-file* ] [ **-u** *userid* ] cmd args...
+
+Description
+===========
+
+**cyr_withlock_run** will run the command and arguments provided on the
+command line, either locking the entire server, or a particular user, so
+no changes can be made to it while this command is running.
+
+WARNING: since this takes an exclusive lock, it MUST NOT be used to run an cyrus
```
elliefm, line 32:
```suggestion
WARNING: since this takes an exclusive lock, it MUST NOT be used to run a cyrus
```

elliefm (changes requested):
This needs a changes/next file.  Also, couple of nits in the documentation.  Implementation looks okay!  I wonder if there's some way this could be tested

brong:
Have added some clever environment magic to allow calling Cyrus commandline tools within the executed command, as well as the requested documentation updates

imap/cyr_withlock_run.c
ksmurchison: This looks like it still needs to be fixed

ksmurchison (approved):
Other than the one comment nit pointed out by @elliefm , this LGTM.  No need for another review after fixing the comment

docsrc/imap/reference/manpages/systemcommands/cyr_withlock_run.rst
```diff
@@ -0,0 +1,85 @@
...
+        Pause all writes on the server for 300 seconds (stops any writes from getting locks)
+
+Files
+=====
+
+/etc/imapd.conf
+
+
+Environment
+===
```
elliefm, line 77: Sphinx is picky about this crap so I have to be too:

```suggestion
Environment
===========
```

Please just squash these trivial suggestions back into the commits they repair, they clutter up the history for no benefit otherwise

imap/cyr_withlock_run.c
brong: what?  I swear I fixed that

brong:
Per @elliefm comments - have merged trivial changes back and compacted this into sensible commits.

PR #4746, opened 2023-11-27
DAV propfind: don't crash if we fail to transform iCal/vCard data

ksmurchison:
respond with 422 (Unprocessable Content)


brong (approved):
This looks good too

PR #4748, opened 2023-11-28
vCard: strip any X-LIC-ERR properties

ksmurchison:
Also, don't return ETag on CardDAV PUT if we stripped

This PR is required when compiled against libicalvcard with:
https://github.com/libical/libical/pull/584/commits/771beabb5f28fa72b84b1b5dd6b11b3494fff00a
which now inserts X-LIC-ERR for non-fatal parse errors and continues to parse.

This allows us to parse and accept vCards with bogus properties such as REV:REV even if they already exist on disk.

### DIFF
--- a/imap/http_carddav.c
+++ b/imap/http_carddav.c
@@ -1366,6 +1366,7 @@ static int carddav_put(struct transaction_t *txn, void *obj,
     char *type = NULL, *subtype = NULL;
     struct param *params = NULL;
     const char *want_ver = NULL;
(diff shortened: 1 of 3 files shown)


elliefm (approved):
Looks okay to me, with the caveat that if there's deeper ramifications to beware of I don't know enough to see them

brong (approved):
This looks good.  I've merged the vcard changes into the cyrus branch in libical builds for cyruslibs and build v48 with it.

PR #4750, opened 2023-11-29
runtime toggleable debug logging

elliefm:
This adds an API such that `config_debug` (aka "debug:" in imapd.conf) can be toggled on and off at runtime, and hooks up a signal handler such that SIGUSR1 will toggle it.  The test case tests this with `imapd`, but it should work the same for any Cyrus process (specifically, any process that calls both `cyrus_init()` and `signals_add_handlers()`).

You can already crudely toggle debug logging by changing debug: in imapd.conf and sending master a SIGHUP, but that a) affects all Cyrus process, b) only after they restart and reread the config, and c) also picks up other unrelated config changes.  This PR allows toggling debug logging on or off for a specific process, while that process is already running, and without reloading the config generally.

SIGUSR1 is already used by cunit (for test timeout handling) and Cassandane (as part of hooking a debugger to a running test process), but isn't used by Cyrus itself.  This usage shouldn't clash with the test systems' usages.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Simple.pm
+++ b/cassandane/Cassandane/Cyrus/Simple.pm
@@ -213,4 +213,71 @@ sub test_cmdtimer_sessionid
     }
 }
 
+sub test_toggleable_debug_logging
+    :min_version_3_9
+{
+    my ($self) = @_;
+
+    my $config_debug = $self->{instance}->{config}->get_bool('debug', 'no');
+    my $imaptalk = $self->{store}->get_client();
+
+    # can't do anything without captured syslog
+    if (!$self->{instance}->{have_syslog_replacement}) {
+        xlog $self, "can't examine syslog, test is useless";
+        return;
+    }
+
+    # find our imapd pid from syslog
+    my $loginpat = qr{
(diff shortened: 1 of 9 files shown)


elliefm:
Pre-emptive "do not merge" cause by the time this is reviewed and I see it again we'll be in feature freeze.  Will tag it for FM once approved.

ksmurchison (approved):
LGTM.

If you don't want to have a special XNOOP command for testing, SELECT/EXAMINE already has a simple DEBUG log line when it completes

elliefm:
I've discarded the xnoop implementation, rewritten the test to use select instead, and added a comment to imapd.c explaining that cmd_select's debug log line is now load bearing, so that anyone trying to change it in the future doesn't get a surprise

wolfsage:
This is amazing, thanks! I haven't played with it yet but would love to see it in action

ksmurchison (approved):

elliefm:
Won't merge this till after the feature freeze, but I've tagged it include-in-fastmail for now so we can get it into our builds for @wolfsage to play with.  Any issues, let me know, or just untag it.

PR #4751, opened 2023-11-29
sync_support: stronger logging in sync_mailbox_compare_update

elliefm:
We might abort if this returns non-zero, so let's make sure to loudly log why we're returning non-zero...


brong (approved):
These all look good - though some of these are legit "there was a split brain", and maybe should change to LOG_NOTICE and be SYNC_NOTICE still.

rsto:
@brong you re-requested review for this, is there anything that you'd like to see addressed before merging?

brong (approved):
Nope, I'm happy with this!

rjbs:

PR #4756, opened 2023-12-01
message: fix uninitialized access in extract_convdata

elliefm:
Fixes an uninitialized access crash in extract_convdata.


elliefm:
I've reproduced this and confirmed the fix, but I'm still polishing the test into something usable, so it's a draft for now

PR #4758, opened 2023-12-02
Rename between users

brong:
We found an issue in Fastmail production where a user had done:

1) `rename user.A.foo user.B.foo`

followed by

2) `rename user.B.foo user.A.foo`

Which had caused conversations to be all broken because we weren't correctly detecting this case in the mailbox_rename_nocopy (uuid paths) case.

This PR adds a test for the case, then code to make it work (and yes, I did tdd this!)

### DIFF
--- a/cassandane/Cassandane/Cyrus/Conversations.pm
+++ b/cassandane/Cassandane/Cyrus/Conversations.pm
@@ -1095,4 +1095,88 @@ sub test_x_me_message_id_nomatch_threading
     $self->check_messages(\%exp);
 }
 
+sub test_rename_between_users
+	:NoAltNameSpace
+{
+    my ($self) = @_;
+    my $admintalk = $self->{adminstore}->get_client();
+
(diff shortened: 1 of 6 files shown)


ksmurchison (changes requested):
The code looks sane to me, but its failing CI

brong:
CI is fully passing now on my machine (including JMAP-Tester, I added that)

ksmurchison (approved):
This all looks sane to me, and if it works, it works

brong:
@elliefm I've added backport tags to this - it fixes an otherwise bad bug that came in with uuid mailboxes

elliefm:
This doesn't backport well to 3.6, because the test Quota.storage_convquota_immediate relies on the `userrawquota` annotation, which doesn't exist yet.  I guess that's why @brong didn't tag this for 3.6 initially?

If we do want to backport this to 3.6, we'll either also need #4203 and #4209 (at least), or need to drop this test

PR #4762, opened 2023-12-05
Minor vCard cleanup

ksmurchison:
One small commit

### DIFF
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -3693,6 +3693,14 @@ static int mailbox_update_carddav(struct mailbox *mailbox,
 #ifdef HAVE_LIBICALVCARD
         vcardcomponent *vcard = record_to_vcard_x(mailbox, new);
 
+        if (!vcard) {
+            xsyslog(LOG_ERR, "record_to_vcard_x() failed",
+                    "record=<%u> mailbox=<%s>",
+                    cdata->dav.imap_uid, mailbox_name(mailbox));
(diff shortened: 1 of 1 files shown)


imap/mailbox.c
```diff
@@ -3693,6 +3693,13 @@ static int mailbox_update_carddav(struct mailbox *mailbox,
 #ifdef HAVE_LIBICALVCARD
         vcardcomponent *vcard = record_to_vcard_x(mailbox, new);
 
+        if (!vcard) {
+            syslog(LOG_ERR, "record_to_vcard_x failed for record %u:%s",
+                   cdata->dav.imap_uid, mailbox_name(mailbox));
```
elliefm, line 3698: I see that this block was copied from the `#else` chunk below, which already used plain syslog.  But perhaps you could convert both to xsyslog too while you're in here anyway, something like:

```suggestion
            xsyslog(LOG_ERR, "record_to_vcard_x failed",
                             "record=<%u> mailbox=<%s>",
                             cdata->dav.imap_uid, mailbox_name(mailbox));
```

(n.b. I wrote this suggestion on the web, not in my editor.  I think it's right, but it might not be.)

elliefm (approved):
Looks good.  Minor suggestion re syslog vs xsyslog, but up to you whether you take or leave it

imap/mailbox.c
ksmurchison: Fixed

elliefm (approved):

PR #4767, opened 2023-12-08
Allow limiting Sieve :regex execution time

ksmurchison:
This will about from regexec(), and return an error code so that execution of the script will fail (message still delivered to INBOX)
The lmtpd will be terminated after the message has been delivered to all recipients.

### DIFF
--- a/sieve/comparator.c
+++ b/sieve/comparator.c
@@ -342,19 +342,96 @@ static int octet_matches(const char *text, size_t tlen, const char *pat,
 
 
 #ifdef ENABLE_REGEX
+#include <errno.h>
+#include <setjmp.h>
+#include <signal.h>
+#include <syslog.h>
+
+#include "libconfig.h"
+
+static sigjmp_buf jmpbuf;
+static volatile sig_atomic_t canjump;
+
+static void sig_alrm(int sig __attribute__((unused)))
+{
+    if (!canjump) return;  // unexpected signal, ignore
+
+    canjump = 0;
+    siglongjmp(jmpbuf, 1); // jump back to octet_regex()
+}
+
 static int octet_regex(const char *text, size_t tlen, const char *pat,
                        strarray_t *match_vars,
                        void *rock __attribute__((unused)))
 {
+    static struct sigaction action;
+    static double timeout;
+    struct itimerval it = { 0 };
+    struct timeval start, end;
     regmatch_t pm[MAX_MATCH_VARS+1];
     size_t nmatch = 0;
-    int r;
+    static int r;
 
     if (match_vars) {
         strarray_fini(match_vars);
         nmatch = MAX_MATCH_VARS+1;
         memset(&pm, 0, sizeof(pm));
     }
+
+    if (!action.sa_handler) {
+        const char *timeoutstr = config_getstring(IMAPOPT_SIEVE_REGEX_TIMEOUT);
+
+        if (timeoutstr) timeout = atof(timeoutstr);
+        if (timeout < 0) timeout = 0;
+
+        /*
+         * signal() on some platforms may set SA_RESTART by default.
+         *
+         * Therefore, we use sigaction().
+         */
+        action.sa_handler = &sig_alrm;
+        sigemptyset(&action.sa_mask);
+#ifdef SA_INTERRUPT
+        action.sa_flags |= SA_INTERRUPT;
+#endif
+
+        if (timeout > 0 && sigaction(SIGALRM, &action, NULL) < 0) {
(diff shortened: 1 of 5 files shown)


sieve/comparator.c
```diff
@@ -334,20 +334,97 @@ static int octet_matches(const char *text, size_t tlen, const char *pat,
...
+
+    canjump = 0;
+    siglongjmp(jmpbuf, 1); // jump back to octet_regex()
+}
+
 static int octet_regex(const char *text, size_t tlen, const char *pat,
                        strarray_t *match_vars,
                        void *rock __attribute__((unused)))
 {
+    struct sigaction action;
```
elliefm, line 359: I think you meant for this to be `static struct sigaction action;`?  Later on, it looks like we're trying to skip initialising it if we already did that, but without being static, that previous initialisation won't be preserved

sieve/comparator.c
```diff
@@ -334,20 +334,97 @@ static int octet_matches(const char *text, size_t tlen, const char *pat,
...
-    int r;
+    static int r;
 
     if (match_vars) {
         strarray_fini(match_vars);
         nmatch = MAX_MATCH_VARS+1;
         memset(&pm, 0, sizeof(pm));
     }
 
+    if (!action.sa_handler) {
```
elliefm, line 373: ... and this also looks like it expects `action` to be `static`, because otherwise this is an uninitialised access.  If it has some junk in it already, we won't initialise the signal handler.  This might be why the new test fails in CI?

elliefm (commented):
My setjmp(3) man page says this about calling longjmp and siglongjmp from a signal handler, like we're doing here.

> POSIX.1-2008 Technical Corrigendum 2 adds longjmp() and siglongjmp() to the list of  async-signal-safe  functions.  However, the standard recommends avoiding the use of these functions from signal handlers and goes on to  point  out  that  if these  functions  are  called from a signal handler that interrupted a call to a non-async-signal-safe function (or some equivalent, such as the steps equivalent to exit(3) that occur upon a return from the initial call to main()), the behavior is undefined if the program subsequently makes a call to a non-async-signal-safe  function.  The only way of avoiding undefined behavior is to ensure one of the following:
> 
> •  After long jumping from the signal handler, the program  does  not  call  any non-async-signal-safe  functions and does not return from the initial call to main().
> 
> •  Any signal whose handler performs a long jump must be  blocked  during  every call  to  a non-async-signal-safe function and no non-async-signal-safe functions are called after returning from the initial call to main().

It's not clear to me whether regcomp is a non-async-signal-safe function, and it probably varies depending on which regex library Cyrus is linked against anyway.  If we suppose that regcomp is non-async-signal-safe, then it might not be safe to allow lmtpd to finish processing the rest of the recipients after one of them trips a regex timeout -- especially if any of the subsequent recipients also have regex in their sieve, which would definitely call the non-async-signal-safe regcomp again, which this documentation explicitly calls out as undefined.

sieve/comparator.c
ksmurchison: yeo

sieve/comparator.c
ksmurchison: yep

elliefm:
https://datatracker.ietf.org/doc/html/rfc2033.html#section-5:
>    The server SHOULD send each reply as soon as possible.  If it is
>    going to spend a nontrivial amount of time handling delivery for the
>    next recipient, it SHOULD flush any outgoing LMTP buffer, so the
>    reply may be quickly received by the client.
> 
>    The client SHOULD process the replies as they come in, instead of
>    waiting for all of the replies to arrive before processing any of
>    them.  If the connection closes after replies for some, but not all,
>    recipients have arrived, the client MUST process the replies that
>    arrived and treat the rest as temporary failures.

I think this means that, as long as lmtpd makes sure to prot_flush() after writing each recipient reply, then at any time it should be okay for it to exit early, and anything it didn't get to will just queue and retry later.

I don't remember the lmtpd architecture in detail though.  At the time we're processing some user's sieve regex, have we already sent the replies for any earlier recipients?  If we have, then in the case of regexec timeout, it should be fine to just send (and flush) the failed reply for this recipient, and then `shut_down()`.  That would save us worrying about any undefined behaviour while processing the remaining recipients.

It still leaves open a possibility for a user with a pathological regex in their sieve to occasionally delay mail for other users (if those other users are the queued-and-retried later recipients).  We might want to recommend that deployments that allow user-supplied sieve and have the "regex" extension enabled should configure their MTA to send one recipient per message, like FM does.  Though I'm not sure where in the documentation this recommendation would best fit.

rsto:
@ksmurchison This currently fails on CI and it seems unclear if this is the approach we'd should be taking. I have converted this back to draft.

PR #4768, opened 2023-12-10
Add libxml/parser.h include (2)

mosvald:
This is required for libxml2 2.12.0 and above.

Missing piece in PR #4745.

Revealed when compiled with --enable-jmap.


elliefm (approved):
Good catch, thanks!

PR #4774, opened 2023-12-18
Message parse received date invalid read

rsto:
This is the follow-up PR to https://github.com/cyrusimap/cyrus-imapd/pull/4771

### DIFF
--- a/cunit/message.testc
+++ b/cunit/message.testc
@@ -1391,4 +1391,17 @@ static void test_parse_bogus_charset_params(void)
 #undef TESTCASE
 }
 
+/*
+ * Verifies that message_parse_received_date() does not read
+ * uninitialized data in the second call to message_parse_string()
+ */
+static void test_parse_received_semicolon(void)
+{
+   static const char msg[] = "Received: abc;\r\n\r\nd";
+   struct body body;
+   memset(&body, 0x45, sizeof(body));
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):

dilyanpalauzov:
>  A non-existent date-time part in the header value should result in a NULL received_date value. The code building on body->received_date in lmtpd and jmap_mail is already prepared to handle NULL received_dates.

`message_parse_received_date()` contains as comment

>  /* Didn't find ; - fill in hdrp so we don't look at next received header */

My understanding is, that `message_parse_received_date()` has always to store in `hdrp` NON-NULL value, when parsing the first `Received:` header.  This ensures, that `body.received_date` is never filled from the second `Received:` header, if the first one contains no text after the semicolon, or contains no semicolon.  For me the new implementaion of `message_parse_received_date()` does not align with my understing of this comment.

I looked again in message.testc and I see no example for first `Received:` header without semicolon.

rsto:
Right, my bad. Fixed in https://github.com/cyrusimap/cyrus-imapd/pull/4776

dilyanpalauzov:
This should be backported, together with #4776.

elliefm:
Thanks for the prompt.  I intend to backport your original fixes from #4771 with my fix for test (i.e. just the fix, but no behaviour change), but I had forgotten to label that PR for backporting.  I've now done so.

Issue #4491, opened 2023-04-27
fix for sphinx-build 6.2.1


BlackZork:
sphinx-build 6.2.1 fails. Build log attached.

[build.log](https://github.com/cyrusimap/cyrus-imapd/files/11344370/build.log)

A patch:

```diff
--- package.pristine/docsrc/conf.py	2022-12-12 03:05:16.000000000 +0100
+++ package.patched/docsrc/conf.py	2023-04-27 15:49:52.022327117 +0200
@@ -527,8 +527,8 @@
 # XXX handlers here, except that these urls require expanding the version string
 # XXX twice, and the sphinx extlinks thingy currently only supports a single %s
 extlinks = {
-        'draft':('https://tools.ietf.org/html/%s', ''),
-        'issue':('https://github.com/cyrusimap/cyrus-imapd/issues/%s', 'Issue #'),
+        'draft':('https://tools.ietf.org/html/%s', None),
+        'issue':('https://github.com/cyrusimap/cyrus-imapd/issues/%s', 'Issue #%s'),
         'cyrus-2.5':('https://www.cyrusimap.org/2.5%s',None),
         'cyrus-3.0':('https://www.cyrusimap.org/3.0%s',None),
         'cyrus-3.2':('https://www.cyrusimap.org/3.2%s',None),
```

elliefm:
Thanks!

Reading the sphinx docs, it looks like support for `%s` in the caption parameter was introduced in version 4.0.  There wasn't any mention of version 6.x, but I guess that version has deprecated or removed the original behaviour so that `%s` is mandatory.

For the time being, we need to still support older versions of sphinx-build for the website builder, so if we start using `%s` in captions today, the website will break.  But someday, when the website builder is using 4.0+, we'll be able to apply this patch, and in the meantime it's here for anyone else who needs it.

I'll note that the fix for `draft` appears to be wrong, though.  In older versions, the substituted text is appended to the given caption, so the old behaviour would make the caption the name of the draft.  Replacing it with `None` will have the effect of making the caption the full URL.  Without being able to test and confirm, I think the correct future caption for `draft` will be `'%s'`.

dilyanpalauzov:
When will the website builder be updated?

Currently to contribute to Cyrus Imap one needs very new version of libical and very old version of Sphinx-build?  It does cost time to figure out how to build the documentation with Sphinx.

elliefm:
[...]

PR #4777, opened 2023-12-19
Fix compilation failure due to -Werror flag

Thor-x86:
This is an attempt to get this commands working for building the cyrus-imapd:

```bash
autoreconf -i
./configure CFLAGS="-W -Wno-unused-parameter -g -O0 -Wall -Wextra -Werror -fPIC --std=gnu99"   \
    --enable-srs --enable-afs --enable-autocreate --enable-idled --enable-nntp --enable-murder \
    --enable-http --enable-calalarmd --enable-jmap --enable-xapian --enable-replication        \
    --enable-backup --enable-cmulocal --enable-oldsievename --enable-unit-tests                \
    --enable-cyrusdb-benchmarks
make lex-fix
cd perl/annotator; perl Makefile.PL
cd ../imap; perl Makefile.PL
cd ../sieve/managesieve; perl Makefile.PL
cd ../../..
make -j8
make check
```

Without error caused by warning and `-Werror` flag. Also notes that currently I'm working with upstream's latest stable version of toolchains and dependencies. Thus, the fix might not quite relevant on Debian but definitely needed in rolling release distributions like Arch.

### DIFF
--- a/docsrc/imap/concepts/deployment/storage.rst
+++ b/docsrc/imap/concepts/deployment/storage.rst
@@ -969,11 +969,3 @@ chunks, which may not be what you require at that moment.
 .. _Cluster LVM: https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Logical_Volume_Manager_Administration/LVM_Cluster_Overview.html
 .. _dm-cache: http://en.wikipedia.org/wiki/Dm-cache
 .. _Kolab Systems AG: https://kolabsys.com
-
-.. glossary::
-        MTBF
-            Mean Time Between Failures
-
-.. glossary::
-        HBA
-            Host Bus Adapter - connects a host system (computer) to other network and storage devices
--- a/docsrc/imap/concepts/features/server-aggregation.rst
+++ b/docsrc/imap/concepts/features/server-aggregation.rst
@@ -387,13 +387,3 @@ Back to :ref:`imap-features`
 
     Including but not limited to ``SELECT``, ``UID MOVE``, ``RENAME``,
     etc.
-
-.. glossary::
-
-    frontend
-	  the user interface
-
-.. glossary::
-
-	backend
-	  the server components
--- a/imap/http_dav.c
+++ b/imap/http_dav.c
@@ -3816,7 +3816,7 @@ int parse_xml_body(struct transaction_t *txn, xmlNodePtr *root,
         return HTTP_BAD_REQUEST;
     }
     else {
-        xmlErrorPtr err = xmlCtxtGetLastError(txn->conn->xml);
+        const xmlError *err = xmlCtxtGetLastError(txn->conn->xml);
         if (err) {
             xmlFreeDoc(doc);
             txn->error.desc = err->message;


Thor-x86 (commented):
LGTM, CUnit and Cassandane shows no problem.

elliefm:
> cd perl/annotator; perl Makefile.PL
> cd ../imap; perl Makefile.PL
> cd ../sieve/managesieve; perl Makefile.PL
> cd ../../..

These steps shouldn't be necessary, the top level `make` takes care of all that

docsrc/conf.py
```diff
@@ -527,8 +527,8 @@
 # XXX handlers here, except that these urls require expanding the version string
 # XXX twice, and the sphinx extlinks thingy currently only supports a single %s
 extlinks = {
-        'draft':('https://tools.ietf.org/html/%s', ''),
-        'issue':('https://github.com/cyrusimap/cyrus-imapd/issues/%s', 'Issue #'),
+        'draft':('https://tools.ietf.org/html/%s', '%s'),
+        'issue':('https://github.com/cyrusimap/cyrus-imapd/issues/%s', 'Issue #%s'),
```
elliefm, line 531: We will need to make this change eventually to support Sphinx 6, but we can't yet, because it breaks backward compatibility with Sphinx 3, which is what's used to build the cyrusimap.org website.  More discussion here: https://github.com/cyrusimap/cyrus-imapd/issues/4491

In the meantime, we support up to Sphinx 5, and ignore this warning

docsrc/conf.py
```diff
@@ -539,7 +539,7 @@
         'cyrus-stable': ('https://www.cyrusimap.org%s',None),
         'github-release':
             ('https://github.com/cyrusimap/cyrus-imapd/releases/tag/cyrus-imapd-%s',
-             'cyrus-imapd-'),
+             'cyrus-imapd-%s'),
```
elliefm, line 543: Same comment as the earlier '%s' changes.  Please just drop this commit from your PR.

docsrc/imap/download/release-notes/3.3/index.rst
```diff
@@ -16,4 +16,3 @@ Cyrus IMAP 3.3 Tags
     :glob:
 
     x/?.?.?
-    x/?.?.??
```
elliefm, line 19: All these globs are required to collect release notes _in the correct order_ (including for future releases that don't yet exist).  We ignore the warnings they produce when such releases don't exist yet.

`-Werror` is a C compiler flag, it doesn't affect sphinx-build.  These warnings are not why your build stopped.  Please drop this whole commit from the PR.

docsrc/conf.py
```diff
@@ -535,6 +535,7 @@
         'cyrus-3.4':('https://www.cyrusimap.org/3.4%s',None),
         'cyrus-3.6':('https://www.cyrusimap.org/3.6%s',None),
         'cyrus-3.8':('https://www.cyrusimap.org/3.8%s',None),
+        'cyrus-3.9':('https://www.cyrusimap.org/3.9%s',None),
```
elliefm, line 538: This section is only for stable releases, which since 3.0 are only even-numbered.  3.9 does not belong here; there is no `/3.9` tree on the website.

The documentation for the current development version is at `/dev`, which is taken care of by `'cyrus-dev'` a line below.

docsrc/imap/download/release-notes/index.rst
```diff
@@ -31,6 +31,16 @@ Latest development snapshot is |imap_development_release_notes|. Documentation a
...
+Series 3.9
+----------
+
+Documentation at :cyrus-3.9:`/`.
+
+.. toctree::
+    :maxdepth: 1
+
+    3.9/index
+
```
elliefm, line 43: 3.9 is the current unstable version, not a "Supported Product Series".  There is no `3.9/` tree on the website.  Please just drop this whole commit.

imap/http_dav.c
```diff
@@ -3816,7 +3816,7 @@ int parse_xml_body(struct transaction_t *txn, xmlNodePtr *root,
         return HTTP_BAD_REQUEST;
     }
     else {
-        xmlErrorPtr err = xmlCtxtGetLastError(txn->conn->xml);
+        const xmlError *err = xmlCtxtGetLastError(txn->conn->xml);
```
elliefm, line 3819: @ksmurchison can you have a look at this one please, I'm not familiar with the XML library.

In particular, a few lines later we save `err->message` in `txn->error.desc`.  That might not be safe if we don't have lifetime ownership of the `err` object, which the constness might imply.

Since we aren't getting this warning/error outselves, perhaps this is an API change in the XML library that we haven't stumbled upon yet.

docsrc/imap/reference/admin/sop/deleting.rst
```diff
@@ -15,7 +15,7 @@ Message context
 
 .. glossary::
 
-    Delete
+    Deleted
```
elliefm, line 18: These terms exist twice because they mean different things in different contexts, and both meanings need to be defined.  Changing the word silences the warning, but makes the documentation worse.

docsrc/imap/reference/admin/sop/deleting.rst
```diff
@@ -38,7 +38,7 @@ Message context
 
 .. glossary::
 
-    Undelete
+    Undeleted
```
elliefm, line 41: Same comment as for 'Delete' above.  Please just drop this commit from the PR.

docsrc/imap/concepts/deployment/storage.rst
```diff
@@ -969,11 +969,3 @@ chunks, which may not be what you require at that moment.
...
 .. _dm-cache: http://en.wikipedia.org/wiki/Dm-cache
 .. _Kolab Systems AG: https://kolabsys.com
-
-.. glossary::
-        MTBF
-            Mean Time Between Failures
-
-.. glossary::
-        HBA
-            Host Bus Adapter - connects a host system (computer) to other network and storage devices
```
elliefm, line 979: Looks good, thanks

docsrc/imap/concepts/features/server-aggregation.rst
```diff
@@ -387,13 +387,3 @@ Back to :ref:`imap-features`
...
-
-.. glossary::
-
-    frontend
-	  the user interface
-
-.. glossary::
-
-	backend
-	  the server components
```
elliefm, line 399: Looks good, thanks

elliefm (changes requested):
Thanks.  I appreciate the enthusiasm for cleaning up the documentation warnings, but most of your proposed changes make things worse.  I've addressed each with individual comments.

The documentation warnings are not what's causing your build to fail. `-Werror` only applies to the C compiler, not to sphinx-build.

It looks like your build fails because of the `const` warning from http_dav.c.  I've asked Ken to have a closer look at this one, because if the object returned by the libxml call is const (or is const in newer versions of libxml), the way we use it might be unsafe.

docsrc/conf.py
Thor-x86: Okay, may I drop the commit?

imap/http_dav.c
Thor-x86: As an information, the Arch distribution uses [libxml2 v2.12.3][1]. I see the changes at [libxml2 GitLab repository][2] which related to [this issue][3].  This is what maintainer's say:

> This is a slight break of the API, but users really shouldn't modify the global error struct. The goal is to make xmlLastError use static buffers for its strings eventually. This should warn people if they're abusing the struct.

[1]: https://gitlab.gnome.org/GNOME/libxml2/-/commits/v2.12.3
[2]: https://gitlab.gnome.org/GNOME/libxml2/-/commit/61034116d0a3c8b295c6137956adc3ae55720711
[3]: https://gitlab.gnome.org/GNOME/libxml2/-/commit/45470611b047db78106dcb2fdbd4164163c15ab7

Thor-x86:
Drop commit requests are fulfilled, please take a look

> These steps shouldn't be necessary, the top level `make` takes care of all that

I tried, but it yells "No build target" error. There is no explanation or verbose message about the problem unfortunately, so I have to do that manually.

> The documentation warnings are not what's causing your build to fail. -Werror only applies to the C compiler, not to sphinx-build.

The commit 6a9d6170444372201015368f35555e7637552d06 went first, but `-Werror` causes GNU Make complaining on Sphinx. So the compilation process refused to carry on. Initially I thought `-Werror` only for C/C++ not the other stuffs. Re-clone and applying changes again still results the same error. Perhaps I missed something on configuration step?

imap/http_dav.c
Thor-x86: > Adding const seems correct as that is what the API specifies. Since we're getting the error from the XML parser context, and there is a xmlCtxtResetLastError() function, I assume its safe to reference the error until we free the parser context.

Alright, thanks for the clarification. Is there anything I can do?

Thor-x86 (commented):
I've reviewed the changes and ran the CUnit again. @elliefm let me know if you agree with modification on `imap/http_dav.c`. Thanks

pgnd:
can this get a review/merge?
currently, still causing build fails with sphinx-build enabled ...

rsto:
Fixed by 03028deac7777b1c476fc33d34281514ac85991d

PR #4773, opened 2023-12-18
fix typo in promstatsd error message

laf0rge:
The message so far read

cyrus/promstatsd[17284]: Prometheus metrics are not being tracked.  Set prometheus_enable in imapd.conf

and I added that setting to my imapd.conf, which didn't solve the problem. Further investigation resulted in detecting that the parameter actually is called "prometheus_enabled" and not "prometheus_enable".

Let's fix the error message to avoid others falling into the same trap.

PR #4778, opened 2023-12-20
fix typo in promstatsd error message

laf0rge:
The message so far read

cyrus/promstatsd[17284]: Prometheus metrics are not being tracked.  Set prometheus_enable in imapd.conf

and I added that setting to my imapd.conf, which didn't solve the problem. Further investigation resulted in detecting that the parameter actually is called "prometheus_enabled" and not "prometheus_enable".

Let's fix the error message to avoid others falling into the same trap.


laf0rge:
This is a rebase of #4773 as that one had just been approved but wasn't merged.  Please re-approve and merge so I can delete my repo+branch.

ksmurchison (approved):

elliefm:
Thanks!

PR #4780, opened 2024-01-03
SearchFuzzy.pm: do not use non-standard XSNIPPETS command

rsto:
The XSNIPPETS and XCONVMULTISTANDARD commands in Cyrus got deprecated, so don't keep our test using it.


ksmurchison (approved):
LGTM

PR #4782, opened 2024-01-04
xapian_wrap.cpp: query verbatim for terms using fullwidth form

rsto:
This works around an issue with Xapian 1.5 when indexing and querying for text containing codepoints in the Halfwidth and Fullwidth Forms Unicode block.

It does so by querying for both the terms as generated by the query parser and the form as generated by the term generator during indexing. This only is applied if the query actually contains at least on codepoint in that Unicode block.

The issue has been reported to the Xapian mailing list at https://lists.xapian.org/pipermail/xapian-devel/2024-January/003410.html


ksmurchison (approved):
I don't know anything about Xapian, so I'll trust that this fixes the problem

PR #4595, opened 2023-08-19
imapd.c:cmd_create() - remove dead code

dilyanpalauzov:

PR #4783, opened 2024-01-05
imapd.c: cleanup sending CREATE args to backend

ksmurchison:
(inspired by #4595)

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -7708,25 +7708,23 @@ static void cmd_create(char *tag, char *name, struct dlist *extargs, int localon
 
                     // special use needs extended support, so pass through extargs
                     if (specialuse.len || uniqueid) {
+                        const char *sep = "";
+
                         prot_printf(s_conn->out, " (");
-                        int printsp = 0;
+
                         if (specialuse.len) {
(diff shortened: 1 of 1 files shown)


imap/imapd.c
```diff
@@ -7708,25 +7708,24 @@ static void cmd_create(char *tag, char *name, struct dlist *extargs, int localon
...
+
                         prot_printf(s_conn->out, " (");
-                        int printsp = 0;
+
                         if (specialuse.len) {
-                            if (printsp) prot_putc(' ', s_conn->out);
-                            printsp = 1;
-                            prot_printf(s_conn->out, "USE (%s)", buf_cstring(&specialuse));
+                            prot_printf(s_conn->out, "%sUSE (%s)",
+                                        sep, buf_cstring(&specialuse));
```
dilyanpalauzov, line 7717: As `sep` here is the empty string, it can be removed completely from `prot_printf("%s", sep);`.

rsto (changes requested):
Agree with @dilyanpalauzov's analysis. Otherwise this LGTM.

imap/imapd.c
ksmurchison: Fixed

rsto (approved):

PR #4786, opened 2024-01-09
sieve.pm: add test_vacation_multiple

ksmurchison:


cassandane/Cassandane/Cyrus/Sieve.pm
```diff
@@ -1964,6 +1964,40 @@ EOF
...
+    sleep 1;
+
+    xlog $self, "Deliver another message";
+    my $msg2 = $self->{gen}->generate(subject => "Message 2",
+                                      to => Cassandane::Address->new(localpart => 'cassandane', domain => 'example.com'));
+    $self->{instance}->deliver($msg2);
+
+    # Make sure that we only sent one response
+    my @resp = $self->{instance}->getsyslog(qr/smtpclient_open:/);
+    $self->assert_num_equals(1, scalar @resp);
```
elliefm, line 1998: This test will false-negative on systems where we were unable to inject our syslog replacement for whatever reason, because there won't be any saved lines for `getsyslog()` to read, and so even if Cyrus behaved correctly, `@resp` will be empty and the `assert_num_equals()` will fail.

Since the test hinges on being able to check for that log line, if the syslog replacement isn't available, we might as well short-circuit the whole test.  We don't currently have a magic function attribute to skip tests like that (... 🤔), but we can approximate it by checking for `have_syslog_replacement` near the top of the function and just returning early if it's false -- for an example of this, see `test_toggleable_debug_logging` in Simple.pm.

elliefm (changes requested):
Looks good, but there's one improvement to make it more robust if we couldn't inject our syslog replacement for some reason

cassandane/Cassandane/Cyrus/Sieve.pm
ksmurchison: Good call.  Check added.

elliefm (approved):

PR #4790, opened 2024-01-18
Fix missing troff format for some manpages in dist tarballs

J5lx:
I noticed that some commands do not have their manpages pre-generated in troff (i.e. the `man` format) in the dist tarballs, even though their HTML version is included. This PR adjusts the Makefile so the missing manpages are also included in troff and not just HTML.


elliefm (approved):
Oh, well spotted, thanks!

elliefm:
@elliefm when backporting, be careful to prune lines for tools that didn't exist yet

PR #4793, opened 2024-01-19
charset.c: Add 'len' argument to unicode_casemap()

ksmurchison:


rsto (approved):
`-1` isn't that common in the Cyrus codebase as sentinel value for zero-terminated strings but I think it's indeed better than using `0`.

ksmurchison:
I used -1 because that's what ICU4C is using and I can just pass that to u_strFromUTF8()

rsto:
Ah, even better then.

PR #4794, opened 2024-01-19
bc_eval.c: prepare (regcomp, etc) patterns once, not per value string

ksmurchison:
Should help performance and will be also leveraged so we don't have do i;unicode-casemap transformation on patterns multiple times

### DIFF
--- a/sieve/bc_eval.c
+++ b/sieve/bc_eval.c
@@ -59,6 +59,7 @@
 
 #include "gmtoff.h"
 #include "charset.h"
+#include "dynarray.h"
 #include "xmalloc.h"
 #include "xstrlcpy.h"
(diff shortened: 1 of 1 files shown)


rsto (changes requested):
Running this with valgrind shows memory leaks for the `editheader_basic` and `editheader_complex` tests, due to the reallocated needles.

Since this is all about performance: using dynarray_t instead of ptrarray_t might make this more cache-friendly.

ksmurchison:
Yeah, the leak was a minor typo which for some reason wasn't found by my original valgrind run.  That has been fixed along with switching to dynarray_t.

rsto (approved):

PR #4795, opened 2024-01-19
Sieve: add support for comparator-i;unicode-casemap (RFC 5051)

ksmurchison:


changes/next/sieve-unicode-casemap
```diff
@@ -0,0 +1,14 @@
+
+Description:
+
+Adds support for comparator-i;unicode-casemap (RFC 9394) to Sieve
+
```
dilyanpalauzov, line 5: As far as I see, RFC 9394 is not about Sieve and not about Unicode.

docsrc/imap/rfc-support.rst
```diff
@@ -390,6 +390,10 @@ The following is an inventory of RFCs supported by Cyrus IMAP.
 
     IMAP URL Scheme, updated by :rfc:`5593`.
 
+:rfc:`5051`
+
+    i;unicode-casemap - Simple Unicode Collation Algorithm
+
```
dilyanpalauzov, line 396: If RFC 5051 can be supported in the context of Sieve and in the context of IMAP, I suggest adding a note, that narrows the scope to Sieve.

changes/next/sieve-unicode-casemap
rsto: Agreed, that looks like it should be 5051 ?

rsto (changes requested):

changes/next/sieve-unicode-casemap
ksmurchison: Fixed

docsrc/imap/rfc-support.rst
ksmurchison: Fixed

rsto (approved):

Issue #3044, opened 2020-05-15
remove '-I.', and include headers using the relative path


elliefm:
I started this a long time ago, in 39dab7c8071, because our generated headers need their directory to be explicit to be correctly found during certain types of builds (which `make distcheck` checks, and therefore we need to succeed).

And it's fine having that comment in files that need those headers, but every so often we need to add a generated header to a file that doesn't already have any, and this detail gets missed and breaks distcheck again.

We can avoid tripping over this repeatedly by removing -I. completely, and updating every file to name its includes explicitly.

elliefm:
We have ~50 (by crude search) source files that `#include <assert.h>` (i.e. using angle brackets) instead of `#include "assert.h"` (double quotes).

At the moment, they get the correct assertions (ours) because of our abuse of `-I`.  When we fix this issue, those source files will start using the wrong assertions (stdc) instead of ours.

So, as part of fixing all the `#include "foo.h"` lines to be `#include "path/foo.h"`, we'll also need to fix up the `#include <assert.h>` lines to be `#include "lib/assert.h"` too.

elliefm:
WIP branch here (nothing to see yet): https://github.com/cyrusimap/cyrus-imapd/compare/master..elliefm:cyrus-imapd:v39/3044-include-cleanup

elliefm:
This gets more complicated, because some headers get installed so that other code can be linked against libcyrus et al.  These headers get installed to wherever they get installed to, with no extra `lib/` subdirectory.  If the installed headers include other installed headers by spelling it like `#include "lib/foo.h"`, included header will not be found -- since it is not in a `lib/` subdirectory relative to the first one.

So, even though spelling these includes as `#include "lib/foo.h"` works during a normal Cyrus build, it makes the headers that use that spelling unusable to other projects.

PR #4798, opened 2024-01-23
installed headers don't compile

elliefm:
Cyrus installs a bunch of its headers so that other software can compile against the various libcyrus* libraries.  But if you try that it doesn't work, because of various problems in the headers that come of us assuming our code is always compiled from within the cyrus-imapd source tree.

This PR fixes some, but not all, of those issues.  It also adds example source files for calling the working parts of the Cyrus libraries, and adds a LibCyrus test suite to Cassandane to make sure they do actually compile and link correctly.

There's a lot more to fix in this space, but nobody's really asking for it anymore, and I ran out of steam months ago.  Here's a cleaned-up version of the former work in progress so we at least get some benefit from it, and future efforts to fix this stuff will already have test infrastructure ready to go.

Original description:
The Debian package maintainer privately reported a problem from the abi-compatibility-checker not being able to run due to the installed headers not compiling.  This is part of Debian's work to switch to 64-bit time_t on some 32-bit platforms.  Further details in https://cyrus.topicbox.com/groups/devel/T7b15dfc105949d4e-Meae606200efd965cc932318f

I'm using this PR to accumulate the fixes that we need to get the abi-compatibility-checker to actually run.

### DIFF
--- a/lib/cyr_qsort_r.c
+++ b/lib/cyr_qsort_r.c
@@ -1,3 +1,5 @@
+#include <config.h>
+
 #include "cyr_qsort_r.h"
 
 #ifndef cyr_qsort_r
--- a/.gitignore
+++ b/.gitignore
@@ -57,6 +57,7 @@ cunit/unit
 debian/
 depcomp
 doc/examples/imapd_conf/imapd.conf.sample
+doc/examples/libcyrus/example_libcyrus_min
 doc/html/
 doc/legacy/murder.png
 doc/legacy/netnews.png
--- a/Makefile.am
+++ b/Makefile.am
@@ -808,6 +808,7 @@ include_HEADERS = \
     lib/vparse.h \
     lib/wildmat.h \
     lib/xmalloc.h \
+    lib/xsha1.h \
     lib/xunlink.h
 
 nodist_include_HEADERS = \
@@ -825,7 +826,6 @@ noinst_HEADERS += \
     lib/ptrarray.h \
     lib/slowio.h \
     lib/util.h \
-    lib/xsha1.h \
     lib/xstrlcat.h \
     lib/xstrlcpy.h \
     lib/xstrnchr.h
--- a/Makefile.am
+++ b/Makefile.am
@@ -764,7 +764,6 @@ include_HEADERS = \
     lib/bitvector.h \
     lib/bloom.h \
     lib/bsearch.h \
-    lib/bufarray.h \
     lib/charset.h \
     lib/chartable.h \
     lib/command.h \
@@ -805,7 +804,6 @@ include_HEADERS = \
     lib/stristr.h \
     lib/times.h \
     lib/tok.h \
-    lib/vparse.h \
     lib/wildmat.h \
     lib/xmalloc.h \
     lib/xsha1.h \
@@ -818,6 +816,7 @@ nobase_include_HEADERS = sieve/sieve_interface.h
 nobase_nodist_include_HEADERS = sieve/sieve_err.h
 
 noinst_HEADERS += \
+    lib/bufarray.h \
     lib/byteorder.h \
     lib/gai.h \
     lib/libconfig.h \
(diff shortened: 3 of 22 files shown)


dilyanpalauzov:
I thought your plan is to prefix in all `#include`s with a `lib/`, when appropriate.  `assert.h` without `lib/` can get very confusing and by tweaking `CFLAGS` replaced by assert from libc.

elliefm:
That was my plan (https://github.com/cyrusimap/cyrus-imapd/issues/3044), but until now I hadn't properly considered if or how these headers work when they're installed for use by third-party software that wants to link against libcyrus, and it turns out that's been broken by some of my changes in this direction.

Hopefully, cleaning it up enough for the abi-compatibility-checker to work will shake out the issues, and reveal a clear path forward.

lib/cyr_qsort_r.c
```diff
@@ -1,3 +1,5 @@
+#include <config.h>
```
guimard, line 1: Hi, from where comes this config.h file ?

elliefm: It's one of the outputs from running `./configure`.  It's not installed, and it can't be installed, so it's a mistake when the headers that are installed depend on it.

dilyanpalauzov:
> That was my plan (https://github.com/cyrusimap/cyrus-imapd/issues/3044), but until now I hadn't properly considered if or how these headers work when they're installed for use by third-party software that wants to link against libcyrus, and it turns out that's been broken by some of my changes in this direction.
> 
> Hopefully, cleaning it up enough for the abi-compatibility-checker to work will shake out the issues, and reveal a clear path forward.

I do not understand how installing or not a header somewhere under `/usr/include/cyrus/` is related to the Binary Interface compatibility.  How do third party software stop working correctly, if Cyrus installs a `.h` file?

elliefm:
> I do not understand how installing or not a header somewhere under /usr/include/cyrus/ is related to the Binary Interface compatibility.

It's not, really.

The Debian people needed to know whether the ABI would be compatible when they change some of their 32 bit platforms to use 64-bit time_t.  They have a tool to check for this automatically, but it wants to compile the installed header files, which fails because some of them rely on stuff like config.h which isn't installed.

We've since established that the packages won't be ABI compatible, and so the Debian people will set their packages up to accommodate that.  The tool no longer matters particularly... but the installed headers are still broken.

So it's not that the headers are a problem for ABI compatibility; they're a problem because they don't compile.  It just so happens that it was the ABI compatibility checker that made us notice they don't compile.

dilyanpalauzov:
My understanding now is, that cyrus-imap installs header files, which in some way depend on `conifg.h`.

rsto (approved):

dilyanpalauzov:
How is this in xsha1.h supposed to work?  Including xsha1.h either works nicely, when config.h is not available, or it fails.  But it cannot work good with or without config.h, but still include config.h.
```c
#if HAVE_CONFIG_H
#include <config.h>
#endif
```
My understaning is that this change is supposed to install only headers, which do not use config.h.  util.h does use config.h, util.h is not installed.  mappedfile.h, sqldb.h, imapurl.h, crc32.h, charset.h, rfc822tok.h are installed and do include util.h and thus depend on config.h.  I addressed “util.h in charset.h” at https://github.com/cyrusimap/cyrus-imapd/pull/5108.

elliefm:
We need config.h so we can check HAVE_SSL, which determines whether cyrus-imapd was configured to use openssl's sha1 implementation, or the internal libcyrus one.

If xsha1.h is being used outside the cyrus-imapd source tree, and therefore config.h is unavailable, the guards make sure we don't `#include` it.  In this case, HAVE_SSL will also be undefined, so outside users of libcyrus will always get the libcyrus sha1 implementation, not the openssl one.

This goes hand in hand with the other change in this file, which makes sure the libcyrus sha1 implementation is always compiled into the library, even if cyrus-imapd is using the openssl one.

PR #4799, opened 2024-01-24
Sieve: a STOP action should not be an error (exit the processing loop directly)

ksmurchison:
An error will skip filing into a custom implicit keep target


rsto (approved):

wolfsage (approved):
lgtm, thanks

PR #4801, opened 2024-01-28
github actions: increase cassandane job count

elliefm:
GH actions now run on twice as many virtual CPUs (thanks @dilyanpalauzov):
https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/

If we take the opportunity to double our cassandane job count, do our CI runs get noticeably faster? [Looks like they do](https://github.com/cyrusimap/cyrus-imapd/actions) -- our runs usually take ~27 minutes, but the ones for this PR are consistently taking ~20.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -54,16 +54,16 @@ jobs:
       run: |
         cp -af cassandane.ini.dockertests cassandane.ini
         chown cyrus:mail cassandane.ini
-        make -j4
+        make -j8
     - name: run cassandane quietly
       id: cass1
       continue-on-error: true
       working-directory: cassandane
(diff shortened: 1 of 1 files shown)


elliefm:
First run looks good -- 20 minutes rather than the usual 27.  Force pushed with a tweak to the commit message, so it'll re-run it.  Let's see if the speedup is repeatable...

elliefm:
@elliefm backport this to all the branches we have CI for

wolfsage:
Awesome, sounds good to me.

wolfsage (approved):

elliefm:
We don't have CI for 3.2.  Leaving the tag there to make this easier to find if we go back and set it up...

PR #4802, opened 2024-01-30
sync_support.c: validate Sieve in first mailbox_compare_update() pass

ksmurchison:
Avoid an abort() in sync_server if we have Sieve with deprecated extensions on the master

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -1007,6 +1007,43 @@ int sync_sieve_delete(const char *userid, const char *fname)
 
     return r;
 }
+
+static int sync_sieve_validate(struct index_record *record,
+                               struct sync_msgid_list *part_list)
+{
+    struct sync_msgid *item = sync_msgid_lookup(part_list, &record->guid);
+    FILE *f = NULL;
+    struct body *body = NULL;
+    const char *msg_base = NULL;
+    size_t msg_len;
+    int r = SIEVE_OK;
+    
+    if (!(item && item->fname && (f = fopen(item->fname, "r")))) return r;
+
+    if (!message_parse_file(f, &msg_base, &msg_len, &body, item->fname)) {
+        sieve_script_t *s = NULL;
+        char *err = NULL;
+
+        r = sieve_script_parse_string(NULL,
+                                      msg_base + body->header_size, &err, &s);
+        sieve_script_free(&s);
+
+        if (r == SIEVE_OK) {
+            // create the record now to avoid a re-parse in sync_append_copyfile
+            message_create_record(record, body);
+        }
+    }
+
+    if (body) {
+        message_free_body(body);
+        free(body);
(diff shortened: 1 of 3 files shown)


elliefm (commented):
Code looks good visually, but I wonder if it's possible to make a test for this somehow.  We'd need to be able to install a known-bad script on the master and then confirm that replication doesn't crash.  But _can_ we install a known-bad script without it just being rejected?

There's a bunch of sieve tests in Replication.pm but I don't see any of them testing deliberately-bad scripts.

ksmurchison:
Right. I think we need an old sievec that only supports the old notify

elliefm:
Maybe we could compile with an old sievec offline, commit the script and bytecode into Cassandane's data directory, and have the test copy them into place on the master instance then try to replicate

imap/sync_support.c
```diff
@@ -1019,6 +1056,12 @@ int sync_sieve_delete(const char *userid __attribute__((unused)),
 {
     return 0;
 }
+
+static int sync_sieve_validate(struct index_record *record,
```
brong, line 1060: This will need `__attribute__((unused))` on its parameters or it will fail to compile

imap/sync_support.c
```diff
@@ -2162,7 +2205,7 @@ int sync_append_copyfile(struct mailbox *mailbox,
 
     if (!item || !item->fname)
         r = IMAP_IOERROR;
-    else
+    else if (!record->cache_version)
```
brong, line 2208: This seems an unrelated optimisation to be in this commit!

imap/sync_support.c
```diff
@@ -2604,7 +2647,17 @@ static int sync_mailbox_compare_update(struct mailbox *mailbox,
         /* after LAST_UID, it's an append, that's OK */
         else {
             /* skip out on the first pass */
-            if (!doupdate) continue;
+            if (!doupdate) {
```
brong, line 2650: I get the slight optimisation here (don't do it on both passes) but generally this code repeats the checks on the doupdate path as well... should this just do it both times and move the doupdate check to after?

brong (changes requested):
Putting _request changes_ because the non-sieve-enabled compile will be broken with strict parameter use checks.

imap/sync_support.c
ksmurchison: It is related in that sync_sieve_validate() has parsed the message during the first pass, so we avoid parsing it a second time

imap/sync_support.c
ksmurchison: Actually, this stub can be removed entirely.

brong (approved):

elliefm:
@elliefm this isn't presently marked for backporting, but if it is ever backported, it'll also need #4826

PR #4804, opened 2024-02-07
mailbox: fix mailbox_maxmessages checks

elliefm:
mbtype isn't a simple bitfield!  Need to mask out the flag bits and then compare the type for equality

The fix here is tiny, the larger part of this PR is in testing it.  The tests verify that the limits affect what we expect them to affect, and also that they don't affect things we don't expect them to affect.

If testing this locally, you can prove that the tests provide coverage of the problem (and therefore can prove the solution) by reverting the mailbox.c commit and watching the new calendar/addressbook tests fail.

These tests will stall when run in verbose mode, due to some weird edge case in Cassandane's SMTP server.  `JMAPEmailSubmission.emailsubmission_set_futurerelease` (and probably others from that suite) have the same problem.  You can work around it in a pinch by increasing `max_servers` in `Instance::_start_smtpd()`, but that increases the memory consumption per test and makes the whole test suite bog down for non-verbose runs, so it's only a small scale workaround.  I'll try to fix the SMTP problem later in a separate PR.


ksmurchison (approved):
LGTM

elliefm:
@elliefm when backporting this, also need #4808

elliefm:
@elliefm and https://github.com/cyrusimap/cyrus-imapd/pull/4424 because the tests expect the specific error, and e4b8fb379 for `assert_syslog_matches()`

PR #4805, opened 2024-02-10
chk_cyrus: Fix fatal error on nonexistent mailboxes

caldwell:
I have these mailboxes on my system:
```
user.me.Junk (\HasChildren)
user.me.Junk.Learn (\NonExistent \HasChildren)
user.me.Junk.Learn.Ham (\HasNoChildren)
user.me.Junk.Learn.Spam (\HasNoChildren)
```
As far as I understand it's perfectly valid to have a nonexistent parent in imap.

However, when I run `chk_cyrus` it gets a fatal error on `user.me.Junk.Learn`:

```
$ chk_cyrus -M user.me.Junk.Learn
Examining mailbox: user.me.Junk.Learn
bad mailbox user.me.Junk.Learn in chkmbox
fatal error: fatal error
```

I investigated and found that the fatal error was `IMAP_MAILBOX_NONEXISTENT`. It seems to me that shouldn't be a fatal error in `chk_cyrus`.

This PR has 2 small commits:

1. One of them prints the error out. With just this commit the output of `chk_cyrus` becomes:
   ```
   $ chk_cyrus -M user.me.Junk.Learn
   Examining mailbox: user.me.Junk.Learn
   bad mailbox user.me.Junk.Learn in chkmbox: Mailbox does not exist
   fatal error: fatal error
   ```
2. One them changes `chk_cyrus` to just skip over nonexistent mailboxes:
   ```
   $  chk_cyrus -M user.me.Junk.Learn
   Examining mailbox: user.me.Junk.Learn
   $
   ```

Does this seem reasonable?

### DIFF
--- a/imap/chk_cyrus.c
+++ b/imap/chk_cyrus.c
@@ -63,6 +63,7 @@
 #include "mailbox.h"
 #include "map.h"
 #include "xmalloc.h"
+#include "imap_err.h"
 
 static void usage(void)
 {
@@ -81,6 +82,9 @@ static int chkmbox(struct findall_data *data, void *rock __attribute__((unused))
     const char *name = mbname_intname(data->mbname);
 
     r = mboxlist_lookup(name, &mbentry, NULL);
+
+    if (r == IMAP_MAILBOX_NONEXISTENT)
+       return 0;
 
     /* xxx reserved mailboxes? */
 
--- a/imap/chk_cyrus.c
+++ b/imap/chk_cyrus.c
@@ -89,7 +89,7 @@ static int chkmbox(struct findall_data *data, void *rock __attribute__((unused))
     /* xxx reserved mailboxes? */
 
     if (r) {
-        fprintf(stderr, "bad mailbox %s in chkmbox\n", name);
+        fprintf(stderr, "bad mailbox %s in chkmbox: %s\n", name, error_message(r));
         fatal("fatal error",EX_TEMPFAIL);
     }
 


elliefm:
What version of Cyrus are you using?  As of a few versions ago (I'm not sure which, perhaps 3.4 or 3.6), Cyrus no longer produces these nonexistent intermediate mailboxes, intermediate path tokens are now always created as full mailboxes.  I don't remember why exactly.  That might be why chk_cyrus is complaining about them.

On servers that have been around for a while and upgraded over the years, they can of course still exist, but in that case you should probably create them as real mailboxes.  So perhaps chk_cyrus should report them as missing, but not fatal out.

The change to make the fatal error log the error message looks good.  The other change, to ignore the nonexistent error code, looks fine for what it is, but I'm not sure what behaviour we actually want here.

rsto:
Adding the actual error in the error message (de520b263023165447e5a78abcb2804b2c4e92b2) looks good as-is. Skipping silently over mailbox names which have no entry in mailboxes.db (4c45f113bbe68537aac90245cfdfceb72e3e00ba) only seems to me to be the right approach if

1. mailboxes.db is known to still allow non-existent intermediaries and
2. the `chkmbox`  function got called as callback to a pattern-based mailbox search.

@ksmurchison is there any way we could determine during runtime if point 1 is the case? I had thought that mailboxes.db would upgrade on the fly from non-existent intermediaries to mboxlist entries, but I might be mistaken. E.g. `mboxlist_update_intermediaries` only gets called during IMAP RENAME and JMAP Mailbox renames currently.

rsto:
This and #5086 both show that handling legacy mailboxes.db with intermediate mailboxes is broken.

ksmurchison (approved):

PR #4806, opened 2024-02-14
re-fix filtering unindexed messages in squat backend

### LINKED ISSUE #4692: Not yet squattered emails are always returned on any search (3.4 and later)

elliefm:
This was originally fixed in #4698, except it turns out there was additional complexity required that was missed because our tests were inadequate.

The search is split into indexed and unindexed subexpressions.  The indexed subexpression is run through the search engine backend, and the unindexed subexpression is then applied as a filter over the returned results.

The squat engine returns all unindexed messages in its matched set, expecting the post-processing phase to filter them further.  This has been broken (#4692) for a while, because nowadays the post-processing phase doesn't apply the indexed subexpression that squat expects it to, only the unindexed subexpression.

The original fix in #4698 failed to recognise the search having previously been split into separate subexpressions.  It copied the indexed subexpression over the unindexed subexpression and used that for filtering, which had the desired effect when the whole search could be completed from the search index, but broke if there were also unindexed terms (e.g. `SINCE date`) to filter on which had now been lost.

This new attempt at a fix is similar, but recombines the indexed and unindexed subexpressions instead of replacing one with the other.  It also adds a bunch of tests!

Fixes #4692

### DIFF
--- a/cassandane/Cassandane/Cyrus/SearchSquat.pm
+++ b/cassandane/Cassandane/Cyrus/SearchSquat.pm
@@ -500,6 +500,9 @@ sub test_unindexed_since
     my $uids = $imap->search('text', 'needle', 'since', '1-Feb-2024');
     $self->assert_deep_equals([1], $uids);
 
+    $uids = $imap->search('text', 'needle', 'not', 'since', '1-Feb-2024');
+    $self->assert_deep_equals([3], $uids);
+
     $self->make_message("needle 4", body => "needle") || die;
     $self->make_message("xxxxxx 5", body => "xxxxxx") || die;
     $self->make_message("old 6", date => $past_dt, body => "needle") || die;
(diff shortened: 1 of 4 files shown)


elliefm:
@ksmurchison Can I get a review on this please?  I'm blocked on some things until this lands.  Thanks

ksmurchison (approved):

PR #4807, opened 2024-02-14
jmap_api.c: ignore rights on \Scheduled when adjusting #jmap

ksmurchison:
\Scheduled has '-anyone' identifier to remove r/w rights which when inherited by #jmap prevents setting \Expunged flag


brong (commented):
Looks good - I wonder if we want to be even more specific about what should and shouldn't lead to shared rights, but this is a fine start.

brong (approved):

PR #4808, opened 2024-02-15
itip_support.c: fix pick_scheddefault_cb()

ksmurchison:
mbtype isn't a simple bitfield!


elliefm (approved):
Oh, I thought I'd audited for other examples of this in my PR, but maybe I forgot.  Good catch!  I just had a quick look myself and this does look like the only one I left behind, whew.

elliefm:
No need to backport for 3.6 or earlier

PR #4809, opened 2024-02-16
IMAP4rev1 uses 32-bit unsigned numbers (not 31-bit)

ksmurchison:


elliefm (commented):
Code looks good, but can you add a changes file please?  I think we'll want to mention this one in release notes

ksmurchison:
> Code looks good, but can you add a changes file please? I think we'll want to mention this one in release notes

Done

elliefm (approved):

PR #4810, opened 2024-02-16
imapparse.c: consolidate getXintXX() code and simplfy older/younger

ksmurchison:


imap/imapparse.c
```diff
@@ -259,160 +259,112 @@ EXPORTED int getxstring(struct protstream *pin, struct protstream *pout,
...
+    uint8_t ones = max % 10;
     int c = EOF;
     int gotchar = 0;
 
-    /* INT_MAX == 2147483647 */
     while ((c = prot_getc(pin)) != EOF && cyrus_isdigit(c)) {
-        if (result > 214748364 || (result == 214748364 && (c > '7')))
+        uint8_t digit = c - '0';
+
+        if (result > tens || (result == tens && (digit > ones)))
```
brong, line 274: Why not just test if it's over INT32_MAX and ignore the tens/ones logic entirely?  We have TONS of headroom with the 64 bits!

brong (approved):
Approving, but I do wonder if we want to change that tens/ones thing.

imap/imapparse.c
ksmurchison: This is just a straight port of what was there. And there is no headroom for getuint64()

imap/imapparse.c
```diff
@@ -259,160 +259,112 @@ EXPORTED int getxstring(struct protstream *pin, struct protstream *pout,
...
+    int c = _getint(pin, LLONG_MAX, &result);
 
     *num = result;
 
     return c;
 }
 
+EXPORTED int getsint64(struct protstream *pin, int64_t *num)
+{
+    return _getsint(pin, LLONG_MAX, num);
```
elliefm, line 360: Since we're using `int64_t` and `uint64_t`, we should also use `INT64_MAX` and `UINT64_MAX` (rather than guessing what native types these are typedefs of).  You'll need `#include <stdint.h>`

imap/imapparse.c
```diff
@@ -259,160 +259,112 @@ EXPORTED int getxstring(struct protstream *pin, struct protstream *pout,
...
+}
+
+EXPORTED int getuint64(struct protstream *pin, uint64_t *num)
+{
+    return _getint(pin, ULLONG_MAX, num);
```
elliefm, line 365: ditto re UINT64_MAX

imap/imapparse.c
```diff
@@ -259,160 +259,112 @@ EXPORTED int getxstring(struct protstream *pin, struct protstream *pout,
...
-    }
-
-    if (!gotchar) {
-        if (c != EOF) prot_ungetc(c, pin);
-        return EOF;
-    }
+    uint64_t result;
+    int c = _getint(pin, INT_MAX, &result);
 
     *num = result;
```
elliefm, line 323: The really big thing for me here is these `*num = result` assignments, when num is of a narrower type with a different signedness.  Does it work?  I dunno, probably, mostly.  But this sort of conversion is where C can be tricky to reason about, and even if it's technically good by the book, does the compiler get it right? Does the optimiser? Etc.

This is where, for the sake of avoiding having to think about (and second-guess) the integer conversion, I'd kinda rather just keep the separate implementations.

imap/imapparse.c
ksmurchison: Rewritten to use strtoull() as we were doing in getmodseq()

ksmurchison:
@brong @elliefm Do you want to give this another look since I rewrote it?

rsto (approved):

PR #4811, opened 2024-02-17
replicaonly: add silent RACL updates and batch them to sync

brong:
This fixes the bumping of RACLMODSEQ on replicas when changes are synced in.  Without this, replicaonly breaks on mailbox creation and deletion if there's shared folders.

We do sync the RACLMODSEQ value along with the mailbox when it syncs, so it's correct not to bump it separately in the update_entry code.

Note: this code also includes and assert that it always gets called with `silent` if replicaonly, which is a nice extra protection and helped with debugging.

### DIFF
--- a/imap/cyr_expire.c
+++ b/imap/cyr_expire.c
@@ -75,6 +75,7 @@
 #include "mboxevent.h"
 #include "mboxlist.h"
 #include "conversations.h"
+#include "user.h"
 #include "util.h"
 #include "xmalloc.h"
 #include "strarray.h"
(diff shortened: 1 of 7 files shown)


brong:
Fixed sync_client and I _think_ cyr_expire as well with two more changes.  One more thing to do is neuter cyr_expire's "delete messages" codepath.

brong:
I think that's it :)

ksmurchison (approved):
If it works, it works.  LGTM

PR #4815, opened 2024-02-21
cassandane bitfield fixes

elliefm:
While investigating an error (which turned out to be not a Cyrus problem), I found a bunch of shortcomings in how we test bitfields in imapd.conf.


ksmurchison (approved):
LGTM

PR #4816, opened 2024-02-21
Cassandane testrunner.pl: Fix XML formatter

wolfsage:
It needs to filter on enable_wanted_properties too, otherwise things like :want_service_http aren't processed, so $self->{caldav} won't be set up leading to this error:

     Perl exception: Can't call method "GetEvents" on an undefined value
     at Cassandane/Cyrus/Sieve.pm line 5560.

... when testrunner.pl is run with -f xml


lerlacher-fm (approved):
LGTM

elliefm (approved):
Good catch, thanks!

PR #4818, opened 2024-02-25
imapoptions: do not spell explicitly the default values

dilyanpalauzov:
As these are anyway rendered as such in the `man` and `html` output for `imapd.conf`.


ksmurchison (approved):
This seems sane to me.  @elliefm do you agree?

elliefm (approved):
Thanks

Issue #4785, opened 2024-01-06
imapd/pop3d crashing on TLS shutdown


citrus-it:
I recently upgraded the OS on my mail servers from a version that shipped Cyrus IMAP 3.6.1 and OpenSSL 3.0 so one that incoudes Cyrus IMAP 3.8.1 and OpenSSL 3.1.
I immediately started seeing crashes of both imapd and pop3d during TLS session shutdown.

The following analysis is from a core file created while the application was running under the illumos libumem memory allocator, with debugging turned on.

The crash occurs when attempting to free an `SSL` structure via `SSL_free()`

```
fffffaffffdf8a60 libcrypto.so.3`CRYPTO_free_ex_data+0x4c()
fffffaffffdf8a80 libssl.so.3`SSL_SESSION_free+0x44()
fffffaffffdf8ab0 libssl.so.3`SSL_free+0x14e()
fffffaffffdf8ae0 libcyrus_imap.so.0.0.0`tls_reset_servertls+0x2b(46fcd0)
fffffaffffdf8b60 service_main+0x438(2, a34f20, fffffaffffdfb8a0)
fffffaffffdfb830 main+0x71c(2, fffffaffffdfb888, fffffaffffdfb8a0)
fffffaffffdfb860 _start_crt+0x87()
fffffaffffdfb870 _start+0x18()
```

The `SSL` object concerned was originally allocated by `SSL_new()`, as can be seen by looking at the allocation stack trace:

```
> 46fcd0/K | ::whatis
d87010 is d87000+10, allocated from umem_alloc_8192:
            ADDR          BUFADDR        TIMESTAMP           THREAD
                            CACHE          LASTLOG         CONTENTS
          d8aee0           d87000  10cdb04daa6b9b7                1
                           9d7028           8f3c00                0
                 libumem.so.1`umem_cache_alloc_debug+0x213
                 libumem.so.1`umem_cache_alloc+0xe3
                 libumem.so.1`umem_alloc+0x9f
                 libumem.so.1`umem_malloc+0x46
                 libcrypto.so.3`CRYPTO_zalloc+0x11
                 libssl.so.3`SSL_new+0x3b
                 libcyrus_imap.so.0.0.0`tls_start_servertls+0xaf
                 cmd_starttls+0x82
                 service_main+0x684
                 main+0x71c
                 _start_crt+0x87
                 _start+0x18
```

The crash occurs because `SSL->session` was previously freed in `ssl3_send_alert()`:

[...]

PR #4819, opened 2024-02-25
Revert "tls.c: need to free SSL_SESSION when removing it from cache"

dilyanpalauzov:
This reverts commit a3523d4067c502a52cb7ee859c1493d1a980b7e0.

imapd does crash sometimes, when it handles the TLS connection.  https://github.com/cyrusimap/cyrus-imapd/issues/4785 describes such reports from different users.

In such cases Cyrus can log “no shared cipher in SSL_accept() -> fail” or “unexpected eof while reading in SSL_accept() -> fail”.

The recommendation at https://github.com/openssl/openssl/issues/23031#issuecomment-1879632721 also states, than doing `SSL_SESSION_free()` during remove_session_cb() is wrong.  If there were leaks, there is probably another case where there is a missing SSL_SESSION_free() call but this place is not right.

• backport to 3.10 and 3.8

### DIFF
--- a/imap/tls.c
+++ b/imap/tls.c
@@ -637,7 +637,6 @@ static void remove_session_cb(SSL_CTX *ctx __attribute__((unused)),
     session_id = SSL_SESSION_get_id(sess, &session_id_length);
 
     remove_session(session_id, session_id_length);
-    SSL_SESSION_free(sess);
 }
 
 /*


dilyanpalauzov:
Closes https://github.com/cyrusimap/cyrus-imapd/issues/4785.

elliefm (approved):

elliefm:
@ksmurchison I'm happy with this, but can I get a second opinion?  Thanks

I have no insight about where the leak really is (I haven't looked though), but there's enough crash reports to suggest that this wasn't the fix for it.

ksmurchison:
Hmm. Let me think about it. My patch definitely made valgrind happy. Where are the crash reports?

elliefm:
In #4785

ksmurchison (approved):

PR #4821, opened 2024-02-27
http_jmap.c: create Problem Details Object on JMAP upload failure

ksmurchison:
This originated from a request to response with 507 and a Problem Details for over quota rather than a 500.
Over quota can no longer happen due to unrelated bugfixes, but it seemed like a good idea to output Problem Details for any upload failure.


rsto (changes requested):
Could you please add a test for this in the Cassandane JMAPCore suite?

rsto (approved):

PR #4823, opened 2024-02-29
UIDonly.pm: account for IMAPTalk +/- v4.06 differences

ksmurchison:
The UIDFETCH data gets returned to the callback as a hash in v4.06 but is an array in v4.04


elliefm (approved):
Looks good.  CI still uses 4.04, so that tells us this is backward compatible

PR #4824, opened 2024-02-29
Idle.pm: pull SESSIONID from procinfo rather than LOGIN response

ksmurchison:
IMAPTalk v4.06 now send CAPABILITY command immediately after LOGIN so we lose the LOGIN response

### DIFF
--- a/cassandane/Cassandane/Cyrus/Idle.pm
+++ b/cassandane/Cassandane/Cyrus/Idle.pm
@@ -428,23 +428,20 @@ sub test_sigterm
     my $store = $svc->create_store(folder => 'INBOX');
     my $talk = $store->get_client();
 
-    # User logged in SESSIONID=<0604061-1337148251-29539-1>
-    my $rem = $talk->get_response_code('remainder');
-    my (undef, $start, $imapd_pid, undef) =
-        ($rem =~ m/SESSIONID=<([^-]+)-(\d+)-(\d+)-(\d+)/);
-    # cyrus switched pid and start at one point - the start will ALWAYS
-    # be larger than the pid, so....
-    ($imapd_pid, $start) = ($start, $imapd_pid) if ($start < $imapd_pid);
-    $self->assert_not_null($imapd_pid);
-    $imapd_pid = 0 + $imapd_pid;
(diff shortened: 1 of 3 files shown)


cassandane/Cassandane/Cyrus/Idle.pm
```diff
@@ -429,9 +429,9 @@ sub test_sigterm
     my $talk = $store->get_client();
 
     # User logged in SESSIONID=<0604061-1337148251-29539-1>
-    my $rem = $talk->get_response_code('remainder');
+    my $syslog = join "\n", $self->{instance}->getsyslog();
```
elliefm, line 432: This won't work if we couldn't inject our syslog sniffer for some reason, and then the test will fail a few lines later on `assert_not_null($imapd_pid)`

One possibility is to check for `have_syslog_replacement` and exit early if it's not available -- see `test_toggleable_debug_logging` in Simple.pm for an example of this sort of thing.  But that's not ideal here, the test has nothing to do with syslog other than we're using it to get the pid, so skipping it entirely when syslog capture isn't available isn't quite right.

A better way might be by running `cyr_info proc` and parsing the imapd pid from the output.  We don't have anything already lying around to do this, but we've got parts of it.  Have a look at `run_cyr_info` and the various `test_proc_...` tests in Info.pm.  It would be easy to move `run_cyr_info` into Instance.pm so that Idle.pm tests can call it too -- the only reason it isn't already there is that I didn't want to clutter up Instance.pm if nothing else was gonna use it.

The `cyr_info proc` output format is defined by `print_procinfo` in cyr_info.c.  Skimming that, I think you'd be looking for a line like `^([0-9]+) imap localhost cassandane.*$` in the output to tell you the pid of the process servicing cassandane's imap session.

I don't think this test actually uses `$start` except for swapping the SESSIONID fields if necessary, so once you get the pid from cyr_info instead, you don't need `$start` anymore, and it doesn't matter that this method doesn't provide it.

ksmurchison: @elliefm Thanks for the procinfo suggestion.  That's way better

PR #4825, opened 2024-02-29
search_query.c: min_uid might start at zero

ksmurchison:
_subquery_run_one_folder() can be called with an existing folder so we don't have the chance to set min_uid to last_uid + 1

This bug only appeared with using Xapian for search


rsto (approved):

brong:
Yep, that's good.  I'm putting some backport markers on this, it's a protocol compliance bug.

ksmurchison:
@elliefm I don't think this will need to be backported, as the change that created the bug was part of implementing IMAP PARTIAL, which was done fairly recently.

PR #4826, opened 2024-03-01
sync_support: duplicate the sieve script

brong:
This fixes mmap size issues with just using the script as a string

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -1024,13 +1024,19 @@ static int sync_sieve_validate(struct index_record *record,
         sieve_script_t *s = NULL;
         char *err = NULL;
 
-        r = sieve_script_parse_string(NULL,
-                                      msg_base + body->header_size, &err, &s);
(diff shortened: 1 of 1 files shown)


elliefm (approved):

elliefm:
@elliefm probably need to backport this as far back as the #sieve mailbox goes.  I think that's 3.6, but I'll find out as I go

imap/sync_support.c
```diff
@@ -1024,14 +1024,20 @@ static int sync_sieve_validate(struct index_record *record,
         sieve_script_t *s = NULL;
         char *err = NULL;
 
-        r = sieve_script_parse_string(NULL,
-                                      msg_base + body->header_size, &err, &s);
+        // XXX: if mmap is a multiple of 4096 bytes, we can read past the
+	// end of the mmap - the string won't be correctly NULL terminated
```
dilyanpalauzov, line 1028: This line has a leading TAB instead of leading spaces, like 207 other lines.

PR #4828, opened 2024-03-01
http_carddav.c: Don't allow the Default addressbook to be deleted

ksmurchison:


cassandane/Cassandane/Cyrus/Carddav.pm
```diff
@@ -1450,4 +1450,27 @@ EOF
     $self->assert_matches(qr/2500 members/, $value);
 }
 
+sub test_delete_default_addressbook
+    :min_version_3_9 :needs_component_httpd
```
elliefm, line 1454: You've marked this for backporting as far as 3.6, so could you make this

```suggestion
    :min_version_3_6 :needs_component_httpd
```

... so that the test will run once it's backported?  Thanks

elliefm (approved):
Looks good.  One nit, but doesn't need re-review after it's fixed

rsto (approved):
LGTM in addition to what @elliefm said

PR #4829, opened 2024-03-02
Cassandane: convert many tests to tiny tests

rjbs:
This converts all `Cassandane/Cyrus/*.pm` files that were over 100k in size to use the tiny test system.

This was largely done mechanically, with human review.  My human review found some problems, so I updated the tooling to do better and the tests to be more tooling friendly.

I think we have general agreement that tiny tests are better, but I'm not entirely sure, so we should get general consensus.  That said, I believe this code is ready for review.  Any commit that says "split X tests into tiny tests" was mechanical and should be reviewed as such.  Any other commit was made by my manual work.


ksmurchison (approved):
@rjbs Can you rebase this on master?  Two of the renamed JMAPCalendar files were updated this morning

rjbs:
@ksmurchison rebased.  I found one new test added to Carddav, which was a "Tiny" test in this branch, so I split it out and amended that change into the other work in this branch.

PR #4831, opened 2024-03-05
Parse email addresses with non-ASCII UTF8 domain parts

rsto:
Email addresses having non-ASCII UTF8 domain parts up until now were parsed incorrectly, leading to truncated email addresses being stored in the mailbox cache and returned in JMAP and IMAP.

This fixes the email address parser to parse such domains and bumps the mailbox minor cache version number.

### DIFF
--- a/lib/parseaddr.c
+++ b/lib/parseaddr.c
@@ -314,7 +314,7 @@ static int parseaddr_domain(char **inp, char **domainp, char **commentp, int *in
 
     for (;;) {
         c = *src++;
-        if (Uisalnum(c) || c == '-' || c == '[' || c == ']' || c == ':') {
+        if (Uisalnum(c) || c == '-' || c == '[' || c == ']' || c == ':' || c > 127) {
             *dst++ = c;
             if (commentp) *commentp = 0;
         }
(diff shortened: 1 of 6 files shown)


rsto:
Huh, that success locally but fails in CI. Changing this to Draft until I fixed that.

lib/parseaddr.c
```diff
@@ -314,7 +314,7 @@ static int parseaddr_domain(char **inp, char **domainp, char **commentp, int *in
 
     for (;;) {
         c = *src++;
-        if (Uisalnum(c) || c == '-' || c == '[' || c == ']' || c == ':') {
+        if (Uisalnum(c) || c == '-' || c == '[' || c == ']' || c == ':' || c > 127) {
             *dst++ = c;
```
ksmurchison, line 318: Do we care to make sure that the char is in the (0x80 <= c <= 0xF4)  range?

rsto: I'd say not to. Unless we change the parseaddr to check if the byte sequence is truly valid UTF-8, we should leave that to the client.

rsto:
Thanks for the `int` vs `u_char` fix. Interesting that this didn't fail on my setup.

ksmurchison (approved):

PR #4837, opened 2024-03-09
Simplify multi-column to single column , fixes #4184

### LINKED ISSUE #4184: cyradm perl warnings on perl v5.34.1
list mailbox with wildcard e.g. (lm user/foobar/* or lm user/foo*) will produce this error at seemingly random intervals:

`Use of uninitialized value $w in numeric gt (>) at /usr/lib64/perl5/vendor_perl/Cyrus/IMAP/Shell.pm line 676.`

This is on Fedora 36 - perl-interpreter-5.34.1-486.fc36.x86_64

jeffgoh:
Fixes #4184  - Negative repeat count does nothing at /usr/lib64/perl5/vendor_perl/Cyrus/IMAP/Shell.pm line 688

### DIFF
--- a/perl/imap/IMAP/Shell.pm
+++ b/perl/imap/IMAP/Shell.pm
@@ -685,13 +685,8 @@ sub _sc_list {
   $w = $ll if $w > $ll;
   my $n = int($ll / $w);
   my $c;
-  for ($l = 0; $l < int((@l + $n - 1) / $n); $l++) {
-    for ($c = 0; $c < @l; $c += int((@l + $n - 1) / $n)) {
-      if ($l + $c < @l) {
-        $lfh->[1]->print($l[$l + $c], ' ' x ($w + 1 - length($l[$l + $c])));
-      }
-    }
-    $lfh->[1]->print("\n");
+  for (@l) {
+    $lfh->[1]->print("$_","\n");
   }
   0;
 }


elliefm (changes requested):
Thanks for submitting this!  I've been re-reading the original issue to refresh my memory, and in a comment there, I said this:

> I think all the calculation of `$w`, `$ll`, `$n`, `$c` are unnecessary once the column formatting is discarded, so you should also remove the parts that calculate those values. `$l` and `@l` look like they're still needed though.

You haven't removed those parts though, do you mind doing so (and test it out)?  I could do it myself, but I don't actually use cyradm, so it would be hard for me to check whether I've accidentally broken something.  I think the changes you need are like this:

* line 662 -> remove `$w, `
* line 668 -> remove line (no longer need to initialise $w)
* lines 677-679 -> remove lines (no longer need to compute $w)
* lines 684-687 -> remove lines (no longer need $ll, $w, $n, $c)

You might want to do these in the reverse order so the line numbers don't change as you make the changes... :)

Also, this PR branch has a merge commit on it.  Could you please rebase to get rid of the merge, and then force-push?  If you don't know how to do this, that's okay, I'll land the PR manually and do it myself in the process.  But if you do, it would make life easier.  Thanks!

elliefm:
@rsto This PR hasn't been updated since I requested changes, other than an automated force push from Rik a few weeks ago to make the CI checks run.  Looks like Rik's force push at least got rid of the merge commit, but the other requested changes haven't been addressed.

I'm not sure what to do with this.  It would be easy enough to just merge it and then make the additional changes ourselves, but since we don't use (or test) cyradm, it's hard for us to be sure we haven't accidentally broken something.  I was hoping the submitter would take care of that part, but I guess they're not around anymore.

PR #4838, opened 2024-03-11
mboxlist.c: add 'R' field (rootid, a.k.a. inboxid) to mbentry_t

ksmurchison:


imap/ctl_cyrusdb.c
```diff
@@ -189,35 +191,55 @@ static int fixmbox(const mbentry_t *mbentry,
...
+
+    if (mboxname_isusermailbox(mbentry->name, 1)) {
+        free(frock->inboxname);
+        free(frock->inboxid);
+        frock->inboxname = xstrdup(mbentry->name);
+        frock->inboxid = xstrdup(mbentry->uniqueid);
+    }
+    if (!mbentry->inboxid && frock->inboxname &&
+        mboxname_is_prefix(mbentry->name, frock->inboxname)) {
+        xsyslog(LOG_INFO, "mbentry had no inboxid, setting from",
```
brong, line 212: This is going to be very noisy on upgrade, and doesn't log the mbentry name.  It should probably at least do that!

brong (approved):
Looks good - my only comment is about the syslog during upgrading being a bit meaningless without the mboxname in it.

ksmurchison:
@brong I added 'inboxid' code to mailbox_create() which fixes the CaldavAlarm replication tests failing on the v5 branch if you want to re-review.
I also added code to ctl_mboxlist.c to dump the 'U' field

ksmurchison:
@brong Does this look ready?  If so, I'll flag it as include-in-fastmail

brong (approved):

ksmurchison:
@brong This PR was more complete than my other one so I closed it.  However, the old one used an 'R' tag for 'root_mailboxid'.  And this one uses 'U' for 'user_inboxid'.  Do you have a preference?  I'd like to just use 'inboxid' but 'I' is already used for 'uniqueid'.  We could use 'B' or 'X'.

rsto:
@ksmurchison Is this still work in progress? Should this get closed or reverted to Draft?

ksmurchison:
I think we still want this.  Let's discuss on Monday.

ksmurchison:
@brong Can you review this again?  I now have a use for this: JMAP mailbox ids based on createdmodseq.  

I'll be adding a new J record to mailboxes.db to be able to lookup mbentries by JMAP ID, however most (all?) INBOXen will have createdmodseq == 1, so to resolve these conflicts, I will need to make the key for these records 'J' + inbox_uniqueid + '.' + encoded_createdmodseq

dilyanpalauzov:
For me it seems that `ctl_mboxlist.c:do_undump()` was forgotten.

ksmurchison:
@brong I now have a use for this: compactids for contacts will have to store a userid along with the encoded cmodseq in dav.db.  We obviously don't want to use the user's name, which isn't rename resistant, so we will use the uniqueid of the inbox.  We don't want to have to extract the username from the addressbook name and lookup the INBOX every time we create a contact, so having the inboxid in the mbentry for the addressbook is a win.

brong:
@ksmurchison should we revisit and maybe finally ship this soon?

PR #4840, opened 2024-03-12
Mmap null

brong:
This makes sure we always have NULL at the end of any mmap.


ksmurchison (approved):

brong:
yeah, turns out there's code assuming things about map_len!  I'm going to rewrite

brong:
OK that's better - it still replies saying the mapped in size, but adds an extra char for the onceonly case.

PR #4842, opened 2024-03-12
caldav_alarm.c: switch from using mboxname to mboxid + inboxid

ksmurchison:
Requires mboxlist_inboxid


imap/user.c
```diff
@@ -743,3 +743,18 @@ EXPORTED int user_isreplicaonly(const char *userid)
...
+    mbentry_t *mbentry = NULL;
+    char *inbox = mboxname_user_mbox(userid, NULL);
+    char *inboxid = NULL;
+
+    mboxlist_lookup_allow_all(inbox, &mbentry, NULL);
+    if (mbentry) {
+        inboxid = xstrdup(mbentry->uniqueid);
+        mboxlist_entry_free(&mbentry);
+    }
+
```
brong, line 758: do we need to free `index` here?

brong (changes requested):
Looks like there's a memory leak...

rsto (changes requested):
Couldn't the upgrade code have race conditions with a concurrent user rename? Or maybe I just don't see how the code prevents that.

imap/user.c
ksmurchison: fixed

imap/caldav_alarm.c
```diff
@@ -148,61 +148,82 @@ EXPORTED int caldav_alarm_done(void)
     " PRIMARY KEY (mboxname, imap_uid)"                 \
     ");"                                                \
 
+#define CMD_CREATE_INDEXES                                        \
+    CMD_CREATEv4_INDEXES                                          \
+    "CREATE INDEX IF NOT EXISTS idx_inboxid ON events (inboxid);"
+
+#define CMD_CREATE_TABLE(name)                          \
+    "CREATE TABLE IF NOT EXISTS " name " ("             \
+    " inboxid TEXT,"                                    \
```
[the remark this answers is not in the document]
ksmurchison: If there is an event with an alarm in a calendar in the shared namespace.  We don't try to process them in calalarmd anyways

ksmurchison:
@brong Regarding @rsto's question about a race.  The upgrade code runs the first time caldav_alarm_open() is called.
So, we could either have the upgrade code grab an exclusive lock on all of mailboxes.db, or we could enable wait daemon mode for calalarmd and only signal to master that its ready after the first  call to caldav_alarm_process(), or we could tell admins to run `calalarmd` -U' prior to starting up Cyrus.
Or, any other ideas?

elliefm:
> So, we could either have the upgrade code grab an exclusive lock on all of mailboxes.db, or we could enable wait daemon mode for calalarmd and only signal to master that its ready after the first call to caldav_alarm_process()

We'd probably need to do both anyway, because wait daemon mode doesn't guarantee the daemon is _only_ run during startup, only that when it is run during startup, it runs before services and other daemons start.  If the daemon crashes or otherwise exits once services are already running, master will just restart it like a normal daemon.  (Though presumably at that point the upgrade has already run, so there's no problem... as long as the upgrade isn't the cause of the crash/exit.)

We'd need to instruct people to add wait=y to their calalarmd entry, and some won't read or follow that instruction, and potentially trip over something later.

> or we could tell admins to run calalarmd -U prior to starting up Cyrus.

Same potential issue wrt people upgrading without reading instructions.

Can we make `ctl_cyrusdb -r` take care of it, like we did with the uuid mailboxes upgrade?  Everyone _should_ already be running that from START, so it runs and finishes before even waitdaemons.

ksmurchison:
> Can we make `ctl_cyrusdb -r` take care of it, like we did with the uuid mailboxes upgrade? Everyone _should_ already be running that from START, so it runs and finishes before even waitdaemons.

Yes.  We can probably just have it call caldav_alarm_upgrade().

@brong What do you think?

rsto:
@ksmurchison I'm going to unassign my review status from this PR. Please add me for review again when this PR is ready for re-review.

PR #4843, opened 2024-03-13
Fix valgrind issues in CardDAV

rsto:


ksmurchison (approved):

dilyanpalauzov:
Do the memory leaks on PUT in `carddav_store_resource()` happen, when a vCard contains two VERSION, UID or FN properties?  In such a case the PUT should be rejected as having invalid data.

rsto:
We've only seen multiple FN properties so far. I agree that having multiple VERSION or UID properties is pretty bogus and we might reject that unconditionally.

For FN I'm not so sure, because as Fastmail we tend to accept invalid data as long as we can make some sense out of it. But I agree that this shouldn't have to be the same for other Cyrus installations. Maybe we introduce a new config option that indicates if Cyrus should accept this and similar vCard violations?

What do you think, @ksmurchison ?

ksmurchison:
Which code path?  Both vparse.c and libicalvcard should enforce min/max for UID and VERSION and reject the resource.  FN is allowed multiple times

rsto:
Yeah, right! My bad, so we're good as-is it looks like.

dilyanpalauzov:
If a single UID and VERSION are enforced, when have the memory leaks, fixed here, happened?

rsto:
They haven't. But the current way the code is written they could happen. We might just as well rewrite that code to not introduce a memory leak should there be multiple UID and VERSION properties, which I agree we should reject anyway earlier in the code path.

PR #4845, opened 2024-03-15
sync_support: fix logging multiple flags

elliefm:
Fixes a bug found by @wolfsage whereby make_flags would just keep overwriting its own buffer if there were multiple flags to log.  Unlike `snprintf`, `buf_printf` _does_ append rather than overwrite.

If this function is ever called, the `static struct buf`'s internal buffer will be "still reachable" at process exit.  There isn't a good way to clean this up.  I think valgrind doesn't complain about "still reachable" leaks with the configuration we use.  

I've run cunit and the cassandane Replication tests under valgrind without complaint, so maybe it doesn't mind, or maybe no tests exercise this code path. That being the case, better review this one carefully 😅

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -5576,32 +5576,34 @@ static int renumber_one_record(const struct index_record *mp,
 
 static const char *make_flags(struct mailbox *mailbox, struct index_record *record)
 {
-    static char buf[4096];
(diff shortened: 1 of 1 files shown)


ksmurchison (approved):

wolfsage:
Awesome, thanks

PR #4846, opened 2024-03-15
annotate_state_writemdata: only called from sync_server, make silent

brong:


ksmurchison (approved):
Looks good, assuming CI Passes

PR #4849, opened 2024-03-17
jmap_mail_query: gracefully handle NULL cstate

rsto:


ksmurchison (approved):
I was there when this was debugged and it solves the problem

PR #4850, opened 2024-03-18
message: return newcid annotation values in message_extract_cids

rsto:
This fixes a bug that has shown up when evaluating JMAP Email queries in Sieve for split conversations.

The some/all/noneInThreadHaveKeyword filter criteria failed, because message_extract_cids returned the conversation id of the first split of all splits in the conversation. This patch changes it to return the very last split which is going to be the one that this message will end up in (unless the message starts a new split).


rsto:
Removed the function call to encode the Sieve script as UTF-8. This particular test doesn't require it to include non-ASCII characters anyway.

PR #4851, opened 2024-03-18
conversations: don't update xconvmodseq on replicas

brong:
It will be synced anyway; we sync_log the affected mailboxes; and this stops it being accidentally bumped by the replica if the message changes are synced out of order.

### DIFF
--- a/cunit/conversations.testc
+++ b/cunit/conversations.testc
@@ -412,7 +412,8 @@ static void test_folder_rename(void)
                         &ecounts1,
                         /*size*/10, /*counts*/NULL,
                         /*modseq*/1,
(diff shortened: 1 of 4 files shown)


ksmurchison:
@brong this CI failure looks like nonsense, but can we force it to run again?

ksmurchison (approved):
Other than the CI failure this LGTM.  Just a simple plumbing the silent flag through.

We should probably understand the CI failure before merging.

rsto:
The build was failing because the unit calling `conversation_update` were broken, now that the function signature changed. I pushed a fix for that and it now tests cleanly.

PR #4852, opened 2024-03-18
Sync user repair

brong:
This set of commits is a test and the code - it adds all the listed mailboxes on both master and replica.  It also fixes any multiple-mailbox sync to always include all the mentioned users, and then check for and either rename or remove any mailbox based on tombstone.

It doesn't solve the XXX case of "mailbox exists on replica and is entirely unknown to master", but that's not a regression.


ksmurchison (approved):
LGTM

PR #4855, opened 2024-03-20
cassandane: configurable core pattern

elliefm:
Improves cassandane's core file detection a little, and also makes the core pattern it searches for configurable so that we don't need to handle every possible eventuality programmatically.

### DIFF
--- a/cassandane/Cassandane/Cassini.pm
+++ b/cassandane/Cassandane/Cassini.pm
@@ -129,6 +129,11 @@ sub new
         die "couldn't find a cassandane.ini file";
     }
 
+    # pre-validate cassandane.core_pattern early -- if the configured
+    # pattern is invalid the qr// will crash out
+    my $core_pattern = $self->val('cassandane', 'core_pattern');
+    $core_pattern = qr{$core_pattern} if $core_pattern;
+
     $instance = $self
         unless defined $instance;
     return $self;
--- a/cassandane/Cassandane/Cassini.pm
+++ b/cassandane/Cassandane/Cassini.pm
@@ -243,4 +243,13 @@ sub get_section
     return \%params;
 }
 
+sub get_core_pattern
+{
+    my ($self) = @_;
+
+    my $core_pattern = $self->val('cassandane', 'core_pattern',
+                                  '^core.*?(?:\.(\d+))?$');
+    return qr{$core_pattern};
+}
+
 1;
--- a/cassandane/Cassandane/Instance.pm
+++ b/cassandane/Cassandane/Instance.pm
@@ -1408,12 +1408,18 @@ sub _detect_core_program
     my $lines = 0;
     my $prog;
 
+    my $bindir_pattern = qr{
+        \/
+        (?:bin|sbin|libexec)
+        \/
+    }x;
+
     open STRINGS, '-|', ('strings', '-a', $core)
         or die "Cannot run strings on $core: $!";
     while (<STRINGS>)
     {
         chomp;
-        if (m/\/bin\//)
+        if (m/$bindir_pattern/)
         {
             $prog = $_;
             last;
--- a/cassandane/Cassandane/Instance.pm
+++ b/cassandane/Cassandane/Instance.pm
@@ -1438,9 +1438,7 @@ sub find_cores
     my $coredir = $self->{basedir} . '/conf/cores';
 
     my $cassini = Cassandane::Cassini->instance();
-    my $core_pattern = $cassini->val('cassandane', 'core_pattern',
-                                     '^core.*?(?:\.\d+)?$');
-    $core_pattern = qr{$core_pattern};
+    my $core_pattern = $cassini->get_core_pattern();
 
     my @cores;
 
--- a/cassandane/Cassandane/Instance.pm
+++ b/cassandane/Cassandane/Instance.pm
@@ -1426,12 +1426,17 @@ sub _detect_core_program
     return $prog;
 }
 
-sub _check_cores
+sub find_cores
 {
     my ($self) = @_;
-
     my $coredir = $self->{basedir} . '/conf/cores';
-    my $ncores = 0;
+
+    my $cassini = Cassandane::Cassini->instance();
+    my $core_pattern = $cassini->val('cassandane', 'core_pattern',
+                                     '^core.*?(?:\.\d+)?$');
+    $core_pattern = qr{$core_pattern};
+
+    my @cores;
 
     return unless -d $coredir;
     opendir CORES, $coredir
@@ -1439,11 +1444,11 @@ sub _check_cores
     while ($_ = readdir CORES)
     {
         next if m/^\./;
-        next unless m/^core(\.\d+)?$/;
+        next unless m/$core_pattern/;
         my $core = "$coredir/$_";
         next if -z $core;
         chmod(0644, $core);
-        $ncores++;
+        push @cores, $core;
 
         my $prog = _detect_core_program($core);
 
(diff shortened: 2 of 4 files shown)


rsto (approved):
LGTM

cassandane/Cassandane/Cassini.pm
```diff
@@ -129,6 +129,11 @@ sub new
         die "couldn't find a cassandane.ini file";
     }
 
+    # pre-validate cassandane.core_pattern early -- if the configured
+    # pattern is invalid the qr// will crash out
+    my $core_pattern = $self->val('cassandane', 'core_pattern');
+    $core_pattern = qr{$core_pattern} if $core_pattern;
```
wolfsage, line 135: Compiling user supplied regex can be dangerous, but in this case... it's someone configuring their own setup, so I guess that's okay

cassandane/Cassandane/Instance.pm
```diff
@@ -1408,12 +1408,18 @@ sub _detect_core_program
     my $lines = 0;
     my $prog;
 
+    my $bindir_pattern = qr{
+        \/
+        (?:bin|sbin|libexec)
+        \/
+    }x;
+
```
wolfsage, line 1416: It'd be nice if the commit message that added this explain why this was changing (as it seems unrelated to core patterns)

cassandane/Cassandane/Instance.pm
```diff
@@ -1426,24 +1432,29 @@ sub _detect_core_program
...
+sub find_cores
 {
     my ($self) = @_;
-
     my $coredir = $self->{basedir} . '/conf/cores';
-    my $ncores = 0;
+
+    my $cassini = Cassandane::Cassini->instance();
+    my $core_pattern = $cassini->val('cassandane', 'core_pattern',
+                                     '^core.*?(?:\.\d+)?$');
```
wolfsage, line 1442: This pattern is significantly different than the old one. Before we'd strictly match core.1234 (period here being the actual '.' character). Now we'll match anything that starts with the word 'core'.
Since (?:\.\d+) is followed by ?, it can match 0 times, which means the pattern is really just /^core.*$/

cassandane/cassandane.ini.example
```diff
@@ -95,6 +95,11 @@
 # to run Cassandane using only configuration from environment
 # variables (or defaults)
 ##allow_noinifile = no
+# Perl regular expression to match core file names on your system.
+# The default should match typical default setups, but if you have
+# configured something unusual in sysctl kernel.core_pattern, you
+# should configure this to a regex that will match it.
+##core_pattern = ^core.*?(?:\.\d+)?$
```
wolfsage, line 102: Same complaint here

wolfsage (commented):
A few comments/nits, but thanks for this!

cassandane/Cassandane/Instance.pm
elliefm: done, how's that?

cassandane/Cassandane/Cassini.pm
elliefm: yeah, if they break it they can keep the pieces

cassandane/Cassandane/Instance.pm
wolfsage: Thanks

cassandane/Cassandane/Instance.pm
elliefm: This is the detail that I most wanted your feedback on, so thanks :)

We used to match either exactly "core" or "core.%p".  This all came up because Fastmail's test system has "core.%E.%t.%s.%p", so my goal was to also be able to match that in the default pattern, without getting too specific about the inner fields.

I thought about looking specifically for dot-separated fields, but while I was thinking about that it occurred to me that if I were configuring my own system, I would probably use hyphens for the non-pid fields, so I ought to match those too, and then I decided to not sweat the details too much and just match anything between "core" and ".%p".  But you're right, with the ".%p" being optional, it becomes redundant because the `.*?` would match it anyway.

Right now, I'm torn between simplifying it to just `^core.*$` (which it effectively is already), or changing it to something like `^core(?:.*?\.\d+)?$` (which maybe better captures my original intent, but reduces its generality).  What do you think?

cassandane/cassandane.ini.example
elliefm: This is just documentation, so it'll be whatever the actual pattern was above

cassandane/Cassandane/Instance.pm
elliefm: I also keep going both ways on whether that `.*?` needs to be non-greedy or if `.*` would be fine.  It's a relatively simple big-picture problem so I'm getting stuck on the overthinking the details 😅

wolfsage: I think there's probably three goals
 * Don't change behaviour and cause breakage where there was none before
 * Don't increase false positives
 * Don't increase false negatives

It's hard to know about the latter 2 goals. In the interest of not having to think too hard about it, I wouldn't change the default pattern from the one in master at this point in time, but that's probably just me being lazy :)

elliefm: The other detail I guess is that the Test::Core tests examine the filename of the core file they found, and if it contains a series of digits, we assert that it matches the pid of our crasher.  If someone has a complicated system core format where pid is not the first numeric term, perhaps say "core.timestamp.pid", even if they configure cassandane's core_pattern to detect those files, the Test::Core will mistake the timestamp part of the filename for a pid and then complain that it doesn't match the crashing process (because it's possible that something crashed other than the thing we intended to crash, invalidating the test).

I could change Test::Core to extract _all_ numeric sections from the filename and assert that if any are present, at least one of them matches... but that would still be a problem for core formats that have numeric fields but the pid isn't one of them.

Maybe instead Test::Core should also just apply the configured core_pattern, and _iff there were a capture group that matched_, compare its contents to the pid.  That way someone configuring a custom core_pattern can control if or where the pid field should be found.  And in that case I'd make the default core_pattern something like `/^core.*?(?:\.(\d+))?$/`, intending to still match anything starting with "core" as we do already, but if it ends with .pid then the pid can be captured and checked.  If someone's pid field is present but not at the end, they can configure their core_pattern with the capture group in the appropriate place.

wolfsage: Yeah... that's an option. Another would be to use a named capture that the user must use in their core pattern:
/^core.*?(?<pid>\d+)\.pid/

And then you can check `$+{pid}`... but maybe that's more complicated.

elliefm: Yeah I haven't used named captures myself before.  Saw them in the docs while finding the correct vocabulary for "capture group", but if I haven't used them as someone who spends roughly half my programming time writing perl, the chances some sysadmin has are slim!  If we were trying to capture multiple fields that might be in any order, they'd be perfect, but since we're only after one field, I think this is a case of do-the-simpler-thing.

I mean, hopefully most people never need to use the core_pattern setting anyway, especially now that the matching is more general than it initially was.  But if they do need to change it, at this point with the capture group for pid I think they'll be able to do what they need without needing further code changes.

elliefm:
@rsto @ksmurchison @wolfsage: This was originally approved by @rsto a while ago, but it's had some changes since then based on discussion with @wolfsage.  I think it's fine now and ready to go, but it's stalled out waiting for an up to date approval.  Can I get a tick from someone please, or further discussion so this can move forward again?  Thanks

ksmurchison (approved):

PR #4862, opened 2024-03-23
docsrc: Replace CRAM-MD5 and DIGEST-MD5 with SCRAM

dilyanpalauzov:
… as the former are removed from upstream Cyrus SASL.

https://github.com/cyrusimap/cyrus-sasl/commit/5436909b1d47142ba0961de31bc984bbd544bd80
https://github.com/cyrusimap/cyrus-sasl/commit/2ce03b618ad60338e8f9094f5b7ba7ee0aef40bf

In fact I am not sure if this substitution is correct - if SCRAM or any other current  mechanism requires storing password in clear text - and if the paragraphs are relevant.


dilyanpalauzov:
https://www.cyrusimap.org/sasl/sasl/faqs/plaintextpasswords.html says for SCRAM the password must be present on the server in clear text.

ksmurchison (approved):

PR #4864, opened 2024-03-25
sync_support: more silent

brong:


imap/sieve_db.c
```diff
@@ -935,6 +937,7 @@ EXPORTED int sieve_ensure_folder(const char *userid, struct mailbox **mailboxptr
                    mboxname, error_message(r));
             goto done;
         }
+	(*mailboxptr)->silentchanges = 1;
```
elliefm, line 940: Is this a typo? I would expect it to be `... = silent;` rather than `... = 1;`, but maybe I've missed a detail earlier

ksmurchison: I have the same question/concern

brong: This here is responsible for the CRC madness!  Yes, you're right.

brong:
@elliefm @ksmurchison - sorry about not fixing this straight away - it caused all the CRC madness too!

PR #4865, opened 2024-03-25
Group racl

brong:
This is a little skanky, but "PTS" is basically a key to list lookup protocol, so I'm using it in reverse to map a group to its members.

(This pairs with an MR in Fastmail hm to add the same thing to our lookup)

### DIFF
--- a/cassandane/Cassandane/Cyrus/HTTPPTS.pm
+++ b/cassandane/Cassandane/Cyrus/HTTPPTS.pm
@@ -92,7 +92,28 @@ sub set_up
                 $res = Plack::Response->new(200);
                 $res->content_type('application/json');
                 $res->body(encode_json({ otheruser => [ "group:group co",
-                                                        "group:group c" ] }));
+                                                        "group:group o" ] }));
+            } elsif ($req->query_parameters->{id} eq 'group:group c') {
+                $res = Plack::Response->new(200);
+                $res->content_type('application/json');
+                $res->body(encode_json({ 'group:group c' => [ "cassandane" ] }));
+            } elsif ($req->query_parameters->{id} eq 'group:group co') {
+                $res = Plack::Response->new(200);
+                $res->content_type('application/json');
+                $res->body(encode_json({ 'group:group co' => [ "cassandane",
+                                                               "otheruser" ] }));
+            } elsif ($req->query_parameters->{id} eq 'group:group o') {
+                $res = Plack::Response->new(200);
+                $res->content_type('application/json');
+                $res->body(encode_json({ 'group:group o' => [ "otheruser" ] }));
+            } elsif ($req->query_parameters->{id} eq 'group:foo') {
+                $res = Plack::Response->new(200);
+                $res->content_type('application/json');
+                $res->body(encode_json({ 'group:foo' => [ ] }));
+            } elsif ($req->query_parameters->{id} eq 'group:this group name has spaces') {
+                $res = Plack::Response->new(200);
+                $res->content_type('application/json');
+                $res->body(encode_json({ 'group:this group name has spaces' => [ ] }));
             } else {
                 $res = Plack::Response->new(404);
             }
--- a/cassandane/Cassandane/Cyrus/HTTPPTS.pm
+++ b/cassandane/Cassandane/Cyrus/HTTPPTS.pm
@@ -241,7 +241,7 @@ sub test_list_groupaccess_noracl
 }
 
 sub test_list_groupaccess_racl
-    :ReverseACLs :min_version_3_7 :NoAltNamespace
+    :ReverseACLs :min_version_3_7 :NoAltNamespace :Conversations
 {
     my ($self) = @_;
 
@@ -252,10 +252,15 @@ sub test_list_groupaccess_racl
     $self->assert_str_equals('ok',
         $admintalk->get_last_completion_response());
 
+    my $precounters = $self->{store}->get_counters();
+
     $admintalk->setacl("user.otheruser.groupaccess",
                        "group:group co", "lrswipkxtecdn");
     $self->assert_str_equals('ok',
         $admintalk->get_last_completion_response());
+
+    my $postcounters = $self->{store}->get_counters();
+    $self->assert_num_not_equals($precounters->{raclmodseq}, $postcounters->{raclmodseq}, "RACL modseq changed");
 
     if (get_verbose()) {
         $self->{instance}->run_command(
--- a/cassandane/data/directory.ldif
+++ b/cassandane/data/directory.ldif
@@ -38,15 +38,19 @@ objectclass: group
 cn: group co
 ou: groups
 o: cyrus
+member: uid=cassandane,dc=internal
+member: uid=otheruser,dc=internal
 
 dn: cn=group c,ou=groups,o=cyrus
 objectclass: group
 cn: group c
 ou: groups
 o: cyrus
+member: uid=cassandane,dc=internal
 
 dn: cn=group o,ou=groups,o=cyrus
 objectclass: group
 cn: group o
 ou: groups
 o: cyrus
+member: uid=otheruser,dc=internal
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -545,6 +545,7 @@ static void cmd_undump(char *tag, char *name);
 static void cmd_xfer(const char *tag, const char *name,
                      const char *toserver, const char *topart);
 static void cmd_rename(char *tag, char *oldname, char *newname, char *partition, int noisy);
+static void cmd_raclmodseq(const char *tag, const char *userid);
 static void cmd_reconstruct(const char *tag, const char *name, int recursive);
 static void getlistargs(char *tag, struct listargs *listargs);
 static void cmd_list(char *tag, struct listargs *listargs);
@@ -2161,6 +2162,13 @@ static void cmdloop(void)
                 cmd_rename(tag.s, arg1.s, arg2.s, havepartition ? arg3.s : 0, /*noisy*/0);
 
                 /* XXX prometheus_increment(CYRUS_IMAP_RENAME_TOTAL); */
+            } else if(!strcmp(cmd.s, "Raclmodseq")) {
+                if (readonly) goto noreadonly;
+                if (c != ' ') goto missingargs;
+                c = getastring(imapd_in, imapd_out, &arg1);
+                if (!IS_EOL(c, imapd_in)) goto extraargs;
+                cmd_raclmodseq(tag.s, arg1.s);
+
             } else if(!strcmp(cmd.s, "Reconstruct")) {
                 if (readonly) goto noreadonly;
                 recursive = 0;
@@ -8235,6 +8243,25 @@ done:
     free(newextname);
     free(olduser);
     free(newuser);
+}
+
+static void cmd_raclmodseq(const char *tag, const char *userid)
+{
+    int r = 0;
+
+    /* administrators only please */
+    if (!imapd_userisadmin)
+        r = IMAP_PERMISSION_DENIED;
+
+    /* touch the modseq */
+    if (!r) r = mboxlist_update_raclmodseq(userid);
+
+    if (r) {
+        prot_printf(imapd_out, "%s NO %s\r\n", tag, error_message(r));
+    } else {
+        prot_printf(imapd_out, "%s OK %s\r\n", tag,
+                    error_message(IMAP_OK_COMPLETED));
+    }
 }
 
 /*
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -1018,19 +1018,14 @@ static int user_can_read(const strarray_t *aclbits, const char *user)
     return 0;
 }
 
-static int mboxlist_update_raclmodseq(const char *acluser, strarray_t *touched_users)
+EXPORTED int mboxlist_update_raclmodseq(const char *userid)
 {
-    if (strarray_find(touched_users, acluser, 0) >= 0) return 0;
-    strarray_add(touched_users, acluser);
-
-    char *acluserinbox = mboxname_user_mbox(acluser, NULL);
-    mbentry_t *raclmbentry = NULL;
-    if (mboxlist_lookup(acluserinbox, &raclmbentry, NULL) == 0) {
-        mboxname_nextraclmodseq(acluserinbox, 0);
-        sync_log_mailbox(acluserinbox);
+    char *inbox = mboxname_user_mbox(userid, NULL);
+    if (mboxlist_lookup(inbox, NULL, NULL) == 0) {
+        mboxname_nextraclmodseq(inbox, 0);
+        sync_log_mailbox(inbox);
     }
-    mboxlist_entry_free(&raclmbentry);
-    free(acluserinbox);
+    free(inbox);
     return 0;
 }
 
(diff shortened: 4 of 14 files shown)


cassandane/Cassandane/Cyrus/HTTPPTS.pm
```diff
@@ -93,6 +93,16 @@ sub set_up
                 $res->content_type('application/json');
                 $res->body(encode_json({ otheruser => [ "group:group co",
                                                         "group:group c" ] }));
+            } elsif ($req->query_parameters->{id} eq 'group:group c') {
+                $res = Plack::Response->new(200);
+                $res->content_type('application/json');
+                $res->body(encode_json({ 'group:group co' => [ "cassandane",
+                                                               "otheruser" ] }));
```
elliefm, line 100: This looks like a copy/paste error?  "group c" should only contain "cassandane" (the "c", "co" etc in the group names are bits describing whether **c**assandane and/or **o**theruser are members)

elliefm (commented):
I can't tell just from looking whether this affects all PTS or just HTTPPTS.  If it affects all PTS, might want to update LDAP.pm too

cassandane/Cassandane/Cyrus/HTTPPTS.pm
brong: Yeah that makes sense except; if you look at the line immediately above, "otheruser" is in both groups!  So I just mirrored that.

cassandane/data/directory.ldif
```diff
@@ -21,6 +21,20 @@ dc: internal
...
+memberof: cn=cassandane,ou=groups,o=cyrus
+memberof: cn=otheruser,ou=groups,o=cyrus
+
+dn: uid=group co,dc=internal
+objectclass: user
+uid: group co
+dc: internal
+memberof: cn=cassandane,ou=groups,o=cyrus
+memberof: cn=otheruser,ou=groups,o=cyrus
+
```
elliefm, line 37: Hmm.  This is actually really weird from an LDAP point of view.  In a real LDAP system, groups are real, you wouldn't fake it like this.

If you have the directory schema this file uses (where group membership is a property of the user object, which is what Cyrus's LDAP implementation seems to assume), then you would index your user objectclass on memberof, and expand group memberships by searching on `(&(objectclass=user)(memberof=[the group you're interested in]))` and collect the user dn's from that.

On the other hand, if you have first class group objects (the schema I'm used to from prior job), you would have a group objectclass with a member attribute that contains its members' dns, and your group expansions are `(&(objectclass=group)(dn=[the group you're interested in]))` and you collect the user dn's from the returned member values.

The data is also wrong.  The original groups-by-user data says "cassandane" is a member of "c" and "co", "otheruser" is a member of "o" and "co".  You're missing "group o" altogether, and have "otheruser" in "group c", where it doesn't belong.  The point of this test data is that both users should be in multiple groups, but not in _all_ groups.  The groups have those bits in their names as a redundancy for detecting whether a weird result is from bad test data or bad code.

With this bad data, if the tests are passing, then either:
* LDAP PTS doesn't use this data, in which case it's useless, or,
* the LDAP tests don't exercise the new users-by-group code, or
* the tests don't adequately verify that groups-by-user and users-by-group results are consistent

The HTTPPTS tests are based on the LDAP ones.  If the group memberships there are different, that was a mistake when they were converted... turns out "c" and "o" are visually very similar, easy to misread 😓

brong: I changed all the LDAP code to make more sense in standard LDAP terms.  The ldap pts module really is very awful code :(  I removed some of the most egregious mis-holding of APIs but it could use some more cleanup.

I'm also not entirely sure if what I'm doing for group membership is "right" but it works fine for test purposes.  We could change to using a filtered scan instead.

brong: And both modules now test for this case :)

ksmurchison: I believe that the ldap pts module was written by the OpenLDAP folks.

brong:
I've re-requested review!  This now includes the two extra bits; a "raclmodseq" command that only admins can use to touch a user's raclmodseq (needed when group membership changes); and a mode to ptexpire which takes a username or group and nukes the cached lookup immediately.

plus of course fixes to LDAP and to tests.

imap/imapd.c
```diff
@@ -2130,6 +2131,13 @@ static void cmdloop(void)
                 cmd_rename(tag.s, arg1.s, arg2.s, havepartition ? arg3.s : 0, /*noisy*/0);
 
                 /* XXX prometheus_increment(CYRUS_IMAP_RENAME_TOTAL); */
+            } else if(!strcmp(cmd.s, "Raclmodseq")) {
```
ksmurchison, line 2134: Wouldn't this be better placed in ctl_mboxlist.c, or do we need/want to do this over the wire?

brong: It's kind of nice over the wire, because you don't need root access on the server - in the Fastmail code we're calling this from a web session over on a separate server.

It could be a HTTP endpoint instead, but we have lots of other admin-only endpoints in imapd.

cassandane/data/directory.ldif
brong: yep; it's more whether what I've done for the test case is idiomatic ldap group with username members.  Yeah, OK it looks like this:

https://www.linkedin.com/advice/0/what-best-practices-managing-ldap-groups-roles

```

dn: cn=Admins,ou=Groups,dc=example,dc=com

changetype: modify

add: member

member: uid=charlie,ou=Users,dc=example,dc=com
```

So I want to add 'member' as the field, and extract a "uid" component from it for each user

lib/imapoptions
```diff
@@ -1457,6 +1457,9 @@ Blank lines and lines beginning with ``#'' are ignored.
 { "ldap_group_base", "", STRING, "2.3.17" }
 /* LDAP base dn for ldap_group_filter. */
 
+{ "ldap_group_hasmember_attribute", NULL, STRING, "UNRELEASED" }
+/* Specify a field within groups with a list of members. */
+
```
elliefm, line 1462: Is this not already present as "ldap_member_attribute"?

>            ldap_member_method: attribute
>               Specify  a  group method.  The "attribute" method retrieves groups from a
>               multi-valued attribute specified in ldap_member_attribute.
> 
>               The "filter" method uses a filter, specified  by  ldap_member_filter,  to
>               find  groups;  ldap_member_attribute  is  a  single-value attribute group
>               name.  Allowed values: attribute, filter

cassandane/data/directory.ldif
```diff
@@ -38,12 +38,16 @@ objectclass: group
 cn: group co
 ou: groups
 o: cyrus
+hasmember: cassandane
+hasmember: otheruser
```
elliefm, line 42: Usually these would be dns, not just uids, because the directory wants to be able to enforce foreign key constraints and so on.  Compare for example the memberof attribute within the user objects, whose values are the dns of the groups.  Also, I'd expect this attribute to typically be called just "member" too, or perhaps be named to denote a specific kind of member (real world examples from memory: "uniquemember", "rfc822mailmember", etc), but not "hasmember".

We can make directory.ldif say whatever the hell we want, but it's important to remember that if it isn't representative of a plausible real world directory layout, then testing against it doesn't prove anything about whether Cyrus's LDAP support works, and a test that doesn't prove anything is a waste of time

cassandane/data/directory.ldif
elliefm: See, this makes sense, I don't know how you got to the "hasmember" stuff you implemented... though this comment was an hour ago, the "hasmember" stuff might have been before that

imap/imapd.c
brong: Reconstruct (for example) doesn't have.  https://datatracker.ietf.org/doc/html/rfc6648 says nah.

lib/imapoptions
brong: yep it is :)  I'll fix this up.

brong: actually, 'ldap_member_attribute' is 'memberof'; which is different:

```
dn: uid=cassandane,dc=internal
objectclass: user
uid: cassandane
dc: internal
memberof: cn=group co,ou=groups,o=cyrus
memberof: cn=group c,ou=groups,o=cyrus
```

as opposed to:

```
dn: cn=group co,ou=groups,o=cyrus
objectclass: group
cn: group co
ou: groups
o: cyrus
member: uid=cassandane,dc=internal
member: uid=otheruser,dc=internal
```

brong:
Hi again @ksmurchison and @elliefm !

I've updated the ldap module to work the way I think it should from my reading of the internets - with "memberof" for the group list in a user, and "member" for the user entry in a group.  I've also fixed the HTTPPTS tests with correct group names and memberships, and added the other groups referenced in the tests to the auth daemon, since it's now mandatory for a group to exists (return 200 to a lookup) in order to be valid (and we check it when setting ACLs)

ptclient/ldap.c
```diff
@@ -1387,11 +1386,29 @@ static int ptsmodule_make_authstate_group(
...
+    *dsize = sizeof(struct auth_state) + (numvals * sizeof(struct auth_ident));
     *newstate = xzmalloc(*dsize);
+    (*newstate)->ngroups = numvals;
     strcpy((*newstate)->userid.id, canon_id);
     (*newstate)->userid.hash = strhash(canon_id);
     (*newstate)->mark = time(0);
+    int i;
+    for (i = 0; i < numvals; i++) {
+        char **rdn = ldap_explode_rdn(vals[i], 1);
+        strlcat((*newstate)->groups[i].id, rdn[0], sizeof((*newstate)->groups[i].id));
```
elliefm, line 1408: This is looking better but it still feels non-LDAPpy.  Looks like it's assuming the first component of the DN is a usable unique key?  But the whole point of a dn is that _it_ is the unique key, and so its components don't have to be.

In my experience, real world user dns are things like `cn=Ellie Timoney,ou=Cyrus Developers,o=Fastmail,c=au`.  I'm under the impression that the schemas I'm familiar with were pretty standard at the time, but I don't know that with certainty, and even if they were, I don't know if they still are.  In the faked up directory.ldif data, the first dn component is the uid, and so it is coincidentally unique in the way you're trying to use it, but that's mostly because I didn't feel like making up common names or a complicated schema for the entries I was creating.  That's my bad I guess, cause in taking that shortcut I've implied a simplicity that doesn't necessarily exist.  

I think the goal here is to get a list of uids back.  To turn a list of dns into a list of uid's, you would do a scope=base search (which returns exactly 1 result or a no-such-object error) for each of those dn's, fetching just the uid attribute from each.  I think our fakeldapd should be able to handle these requests already.  The attribute to fetch should be configurable, I think, because it might not always be called "uid" (e.g. it might be called "mail").  This might be what `ldap_user_attribute` is for, in which case it can be re-used.  If not, I think it's fine to hard-code "uid" for now, and someone who uses something else in the real world can tell us all about it so we can fix it properly. :)

Interestingly, expanding the group in this way will also mean that if the group contains dangling member entries for people that are no longer in the organisation, those dn searches will return zero results, and the final result will be automatically filtered to only contain uids that really exist -- nice!

brong: Yeah - I've done exactly that!  It still passes tests :)

ptclient/ldap.c
```diff
@@ -1403,11 +1404,37 @@ static int ptsmodule_make_authstate_group(
...
+
+        entry = ldap_first_entry(ptsm->ld, res2);
+        char **uvals = ldap_get_values(ptsm->ld, entry, (char *)ptsm->user_attribute);
+        int unumvals = ldap_count_values(uvals);
+        if (!unumvals) continue;  // no user attribute on the DN
+
+        // create a group item
+        strlcat((*newstate)->groups[j].id, uvals[0], sizeof((*newstate)->groups[j].id));
+        (*newstate)->groups[j].hash = strhash((*newstate)->groups[j].id);
+        j++;
```
elliefm, line 1437: Looks like if the `user_attribute` occurs multiple times on this record, we'll ignore all but the first one.

I've forgotten the expected shape of the data here.  Does it make any sense for there to be more than one? If not, maybe we should log a warning or something if `unumvals > 1` and we're ignoring the trailing ones.  If so, maybe we want to be in a loop over `uvals[]` too?

elliefm (commented):
Looking good, but one nit/issue still maybe

imap/mboxlist.c
```diff
@@ -1049,7 +1047,10 @@ static int mboxlist_update_raclmodseq_wrapper(const char *acluser,
     int r = 0;
     int i;
     for (i = 0; i < strarray_size(members); i++) {
-        r = mboxlist_update_raclmodseq(strarray_nth(members, i), touched_users);
+	const char *member = strarray_nth(members, i);
+        if (strarray_find(touched_users, member, 0) >= 0) return 0;
```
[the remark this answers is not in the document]
brong: You're quite right.  Also can use strarray_append rather than strarray_add since we've literally just done the test that starray_add will run in a second.  Fixed.

ptclient/ldap.c
brong: Unlikely, this will be the `uid` property on a DN for a user.  In `ptsmodule_make_authstate_attribute` we require exactly one of them or we don't copy it into the userid part.  I guess I could make it only work if there's exactly one, but I'm not sure if that is more or less surprising.

brong: I made it just iterate and add them all; it's simple enough

brong:
OK, I've re-requested review @elliefm - added fixes for the two nits you have proposed.  Still happily passing all the LDAP tests.

ptclient/ldap.c
```diff
@@ -1429,12 +1429,13 @@ static int ptsmodule_make_authstate_group(
...
-        if (!unumvals) continue;  // no user attribute on the DN
-
-        // create a group item
-        strlcat((*newstate)->groups[j].id, uvals[0], sizeof((*newstate)->groups[j].id));
-        (*newstate)->groups[j].hash = strhash((*newstate)->groups[j].id);
-        j++;
+        int k;
+        for (k = 0; k < unumvals; k++) {
+            // create a group item
+            strlcat((*newstate)->groups[j].id, uvals[k], sizeof((*newstate)->groups[j].id));
```
elliefm, line 1435: Hmm.  Do we need a separator of some sort between uvals here?  I don't follow exactly what the generated id is used for, but it smells like there could be a problem if "foo", "bar", and "foobar" are all valid user_attribute values and the record contains both "foo" and "bar".

Perhaps it makes more sense to mandate that there's only one.  I don't think that would be surprising in the LDAP world, it's normal for a schema to constrain particular attributes to be single-valued, multi-valued, etc.

brong: Ok, this is getting a bit into the very very weeds.  if you look a couple of lines later, there's a `j++`, so maybe this should be strncpy instead but no, that's not an issue.  We're adding all the items.  This does mean that a user might be added more than once but that's already an issue from this half-arsed underlying structure.

elliefm: Oh right, yeah, I missed how `j` was being used here, my bad

brong: @elliefm are you happy to approve now?  Hoping to get this one merged so we can build on top :)

dilyanpalauzov:
Should the documentation be updated?

elliefm (commented):
One last thing: can you add a changes/next/* file please -- copy 0000-TEMPLATE and fill in the fields.  I think we're thinking of this as experimental for the time being, so please describe it as such.  I'm happy to approve it once that's done.

@dilyanpalauzov asked about documentation.  If we're thinking of this as experimental then I don't think that's strictly necessary yet, as long as we're willing to help out anyone else who wants to experiment with it.

brong:
Thanks @elliefm - I've added changes/next file and rebased.  This is still an old "in upstream" branch :)

elliefm (approved):
Thanks!

PR #4871, opened 2024-03-28
Caldav: test changing organiser and overwriting including removing recurrences

brong:
This tests a bug we suspected might have existed, but can't reproduce!


ksmurchison (approved):
Looks good, but per @elliefm previous request, do you want to remove the :min_version ?

Issue #4878, opened 2024-04-03
Wrong sieve script is applied when mail is delivered to multiple users


stbr4:
I'm running Cyrus on Debian Bookworm. Cyrus version: 3.6.1-4+deb12u1 (I recently upgraded from version 3.2.6-2~bpo10+1)

 * A mail is deliverd via lmtp to users U1 U2 and U3
 * U1 has a sieve script which redirects to X
 * U2 has no sieve script
 * U3 has a sieve script which does something

It seems that U1's sieve script is also applied to U2 and U2 doesn't receive the email. U3 receives the mail, even if the script doesn't match.

elliefm:
Interesting.  @ksmurchison, any thoughts?  If this is something we've already fixed in later versions, I might be able to backport it, but I don't remember seeing anything along these lines.  I think Fastmail's MTAs probably send one-mail-per-recipient to lmtpd so we don't see this sort of thing ourselves.

@stbr4 In the meantime, I think you should be able to work around this by configuring your MTA (i.e. postfix, sendmail, exim, etc) to always send one-mail-per-recipient to lmptd.  I don't know how to do this for any particular MTA, but presumably your MTA's documentation will say.

stbr4:
Thank you @elliefm!

We mitigated the problem for now by putting a dummy sieve script in every mailbox. I will look into configuring postfix to deliver mails one at a time.

ksmurchison:
I created #4882 to test this and it passes just fine against master.  I also don't see any relevant changes between 3.6.1 and master, but maybe a second set of eyes will.

stbr4:
Thank you for looking into it, @ksmurchison!

I will try to reproduce it next week. Maybe there is some weird misconfiguration in my setup. Can you suggest anything that i can try to find the cause of this problem?

ksmurchison:
Does this occur across all users or does it only occur if U1 is one of the recipients?

Are the bytecode scripts/directories for the users somehow linked in your sievedir?

stbr4:
U1 has to be one of the recipients. The sieve scripts/bytecode/directories were definitely not linked before the Upgrade. The way sieve scripts are stored changed with this upgrade. I am fairly certain that i checked the directories when i was investigating this problem and didn't find anything unusual. Now there are sieve scripts for every user, so i cannot check now.

[...]

PR #4882, opened 2024-04-05
Add Sieve.deliver_multiple_users

ksmurchison:
Test for #4878


rsto (approved):

elliefm:
I fiddled this test onto 3.6 (which doesn't have tiny tests, so I had to copy and paste the new test function into the main Sieve.pm) and it passes for 3.6.  So whatever the bug in #4878 is, either this test doesn't reproduce it, or it was already fixed sometime between 3.6.1 and 3.6.4 (though I don't see anything in the commit log that looks relevant)

PR #4884, opened 2024-04-11
JMAP Email: update Email/import logic for receivedAt property

rsto:
This updates Email/import to set the receivedAt property to:

- the Email/import receivedAt argument, if present
- the value of the X-Deliveredinternaldate header, if present
- the value of the first valid Received header
- the value of the Date header, if present
- the current time, if none of the above are present or valid

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -13236,6 +13236,57 @@ static int msgimport_checkacl_cb(const mbentry_t *mbentry, void *xrock)
     return 0;
 }
 
+static time_t _email_import_parse_received_at(const char *blob, size_t blob_len)
+{
+    message_t *msg = message_new_from_data(blob, blob_len);
+    struct buf buf = BUF_INITIALIZER;
+    time_t received_at = 0;
+    enum message_format format =  MESSAGE_DECODED | MESSAGE_TRIM;
+
+    if (!msg)
+        goto done;
+
+    if (!message_get_field(msg, "X-Deliveredinternaldate", format, &buf)) {
+        buf_appendcstr(&buf, "\r\n");
+        const char *hdr = buf_cstring(&buf);
+        char *val = NULL;
+        message_parse_string(hdr, &val);
+        if (time_from_rfc822(val, &received_at) < 0)
+            received_at = 0;
+        xzfree(val);
+    }
+
+    if (!received_at && !message_get_field(msg, "Received", format, &buf)) {
+        buf_appendcstr(&buf, "\r\n");
+        const char *hdr = buf_cstring(&buf);
+        char *val = NULL;
+        do {
+            message_parse_received_date(hdr, &val);
+            if (time_from_rfc822(val, &received_at) < 0)
+                received_at = 0;
+            hdr = strchr(hdr, '\n');
+            if (hdr)
+                hdr = strchr(hdr + 1, ':');
+            xzfree(val);
+        } while (!received_at && hdr++);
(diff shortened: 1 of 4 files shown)


imap/jmap_mail.c
```diff
@@ -13236,6 +13236,63 @@ static int msgimport_checkacl_cb(const mbentry_t *mbentry, void *xrock)
...
+        char *val = NULL;
+        message_parse_string(hdr, &val);
+        if (time_from_rfc822(val, &received_at) < 0)
+            received_at = 0;
+        xzfree(val);
+    }
+
+    if (received_at)
+        goto done;
+
```
ksmurchison, line 13261: Just a nit, but rather than using the goto above, the conditional below code be

if (!received_at && !message_get_field(...

imap/jmap_mail.c
```diff
@@ -13236,6 +13236,63 @@ static int msgimport_checkacl_cb(const mbentry_t *mbentry, void *xrock)
...
+            hdr = strchr(hdr, '\n');
+            if (hdr)
+                hdr = strchr(hdr + 1, ':');
+            xzfree(val);
+        } while (!received_at && hdr++);
+    }
+
+    if (received_at)
+        goto done;
+
```
ksmurchison, line 13279: Same comment as above regarding goto.

Also, since parsing X-Deliveredinternaldate and Date are the same, I wonder if we could factor this out into a separate function that accept the header name as an argument.

And if we decide to only check the first Received header, we *may* be able to roll that into the new function as well.

imap/jmap_mail.c
rsto: It figured it's more obvious that we return early this way, but I can change that if you prefer.

imap/jmap_mail.c
rsto: We will always want to check all Received headers, since otherwise we could use the logic from message.c.

I figured it's not worth it to refactor the Date and X-Deliveredinternaldate code into a function since it's just two occasions where we need it. But I can refactor if you prefer.

imap/jmap_mail.c
ksmurchison: Its not a huge deal.  Or maybe put that check inside the if {} after the xzfree()

imap/jmap_mail.c
ksmurchison: OK, that's fine.  But we should change the description to state that we check all Received headers.

rsto: Which description?

ksmurchison: From the PR description:

This updates Email/import to set the receivedAt property to:

    the Email/import receivedAt argument, if present
    the value of the X-Deliveredinternaldate header, if present
    the value of the **first** valid Received header, if present

imap/jmap_mail.c
rsto: I now rewrote this to use `if (!received_at && !message_get_field(..`

imap/jmap_mail.c
rsto: I rephrased this to
```
- the value of the first valid Received header, of all Received headers
```
to make this more clear

ksmurchison (approved):
Actually, your original description was better.  I just missed "first valid Received" and read it as "first Received".  I think what you had is more clear and less redundant than the new description.  Sorry.

rsto:
Thanks, updated to old description.

PR #4887, opened 2024-04-14
potential fixes for debian time_t transition

### LINKED ISSUE #4876: Build/test fail on architectures affected by the time_t transition
From [Debian #1068223](https://bugs.debian.org/1068223).

> cyrus-imapd is failing to build on the architectures affected by the time_t transition _(armel, armhf, several debian-ports architectures)_ with the following error.
>
> > unit: fatal(Internal error: assertion failed: cunit/timeofday.c: 113: tv.tv_usec != 0xffffffff)
>
> A quick look at the code, suggests that the testsuite is trying to intercept gettimeofday, and this interception is broken by the time64 changes.
>
> The specific error appears to be cause by calling the non-time64 gettimeofday function from glibc with the time64 definition of struct timeval, but I suspect there are other issues with the interception code that will rear their ugly head if that one is fixed.

I can launch 3.8.2 build on such architecture if you give me a patch to test.

Best regards,

elliefm:
This draft PR is for producing patches so that @guimard can check for whether they fix  #4876 or not.

### DIFF
--- a/backup/lcb_compact.c
+++ b/backup/lcb_compact.c
@@ -142,8 +142,8 @@ static int compact_closerename(struct backup **originalp,
     struct buf ts_index_fname = BUF_INITIALIZER;
     int r;
 
-    buf_printf(&ts_data_fname, "%s.%ld", original->data_fname, now);
-    buf_printf(&ts_index_fname, "%s.%ld", original->index_fname, now);
+    buf_printf(&ts_data_fname, "%s." TIME_T_FMT, original->data_fname, now);
+    buf_printf(&ts_index_fname, "%s." TIME_T_FMT, original->index_fname, now);
 
     /* link original files into timestamped names */
     r = link(original->data_fname, buf_cstring(&ts_data_fname));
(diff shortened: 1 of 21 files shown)


dilyanpalauzov:
https://wiki.debian.org/ReleaseGoals/64bit-time contains:

> Dpkg enabling -D_FILE_OFFSET_BITS=64 -D_TIME_BITS=64
> Because you have to have LFS if you have 64-bit time_t (glibc enforces this), this goal is a superset of [ReleaseGoals/LFS](https://wiki.debian.org/ReleaseGoals/LFS)

The latter includes

>  For projects using autoconf, by adding the AC_SYS_LARGEFILE m4 macro.

This likely means, that Debian compiles with `-D_FILE_OFFSET_BITS=64 -D_TIME_BITS=64`, and irrespective of this AC_SYS_LARGEFILE must be called. But autoconf definition of AC_SYS_LARGEFILE also states “Define ‘_FILE_OFFSET_BITS’ and ‘_LARGE_FILES’ if necessary.” - I do not know if this is necessary.

All that said, I think adjusting to time_t means calling explicitly AC_SYS_LARGEFILE, which is currently not done.

elliefm:
This is now #5009

PR #4888, opened 2024-04-15
JMAP: store uploaded blobs as application/octet-stream

rsto:
JMAP blobs are untyped and storing the client-supplied media type may result in Cyrus message blobs having a wrong Content-Type header. This changes the JMAP blob upload handler to store blobs as `application/octet-stream` content, unless the uploaded blob is of type `message/rfc822` and the blob passes our checks for valid messages of that type.

### DIFF
--- a/imap/http_jmap.c
+++ b/imap/http_jmap.c
@@ -1158,7 +1158,7 @@ static int jmap_upload(struct transaction_t *txn)
         fprintf(f, "Message-ID: %s\r\n", hdr[0]);
     }
 
-    fprintf(f, "Content-Type: %s\r\n", normalisedtype);
+    fputs("Content-Type: application/octet-stream\r\n", f);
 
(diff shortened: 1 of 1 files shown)


ksmurchison (commented):
The problem here is that if a user uploads a JPEG to use as a contact avatar, we lose the type when creating the PHOTO property.

rsto:
> The problem here is that if a user uploads a JPEG to use as a contact avatar, we lose the type when creating the PHOTO property.

Right, that applies to the current non-standard JMAP Addressbook implementation. In JMAP, it would be up to the client to define the media type of the blob that they are embedding in the contact. Making this as Draft until that's resolved.

rsto:
Turns out the current experimental API already provides the `type` property.

ksmurchison (approved):

PR #4889, opened 2024-04-15
More vCard v3 <-> v4 translations

ksmurchison:
This PR will fail CI until we update to the latest libical vCard branch

### DIFF
--- a/cassandane/tiny-tests/Carddav/put_get_v3_v4
+++ b/cassandane/tiny-tests/Carddav/put_get_v3_v4
@@ -34,13 +34,25 @@ ORG:Bubba Gump Shrimp Co.
 TITLE:Shrimp Man
 REV:2008-04-24T19:52:43Z
 GEO:$lat;$lon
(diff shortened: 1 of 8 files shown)


rsto:
The `vcard` branch doesn't merge cleanly with `libical/master`, so we need to sort out these conflicts first.

rsto (changes requested):
The `vcard` branch doesn't merge cleanly with `libical/master`, so we need to sort out these conflicts first.

ksmurchison:
> The `vcard` branch doesn't merge cleanly with `libical/master`, so we need to sort out these conflicts first.

This has been fixed.

rsto:
@ksmurchison I'm not on top of this here anymore. Is CI failing because it requires new cyruslibs? But Github CI doesn't make use of libicalvcard anyway.

ksmurchison:
@rsto I was able to fix the vparse side of things so that CI passes.

rsto (approved):

PR #4890, opened 2024-04-16
xapian_wrap: ignore punct-only terms only in unstructed search parts

rsto:
This fixes a regression in 0c282379154a776099a83f120311a973c628eb55 Punctuation-only terms only should be ignored in unstructured search parts, such as bodies and subjects.


ksmurchison (approved):
I don't know anything about Xapian, so if it works...

PR #4893, opened 2024-04-19
optional I/O throttling for testing

elliefm:
This PR:

* Adds a configure option `--enable-debug-slowio` and related imapd.conf `debug_slowio` option.  If _both_ are enabled, I/O operations in Cyrus are throttled to about 50Mbps.  This is mainly useful for testing behaviours that only happen when Cyrus is running slowly, such as INPROGRESS response codes.
* Adds a component:slowio => true/false to the cyr_buildinfo output
* Adds `:SlowIO` magic for tests to request this functionality.  If you're using this, you should also add `:needs_component_slowio` so the test can be skipped when the feature hasn't been compiled in
* Updates the InProgress tests to use this feature, rather than the old 15k_messages.tar.gz data.
* Updates tools/build-with-cyruslibs.sh to enable this feature, so that CI can run these tests

**all:** You probably want to add `--enable-debug-slowio` to your local build scripts so that the new tests aren't just skipped in your local Cassandane runs


rsto (approved):
Very neat!

ksmurchison (approved):
LGTM

PR #4894, opened 2024-04-22
deprecate cyrus backups

elliefm:
Updates documentation to mark the experimental Cyrus Backups system as deprecated.

This one will need to be backported, so it only contains the documentation changes.  There will be a separate PR later to remove the feature from master.

### DIFF
--- a/changes/next/deprecate-backups
+++ b/changes/next/deprecate-backups
@@ -1,0 +1,25 @@
+Description:
+
+Deprecates the experimental Cyrus Backups feature
+
+Config changes:
+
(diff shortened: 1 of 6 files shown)


elliefm:
@elliefm When backporting this to the pre-3.10 branches, omit the imapoptions change.

elliefm:
Need to deprecate the `xbackup_enabled` config option too...

elliefm:
Can't deprecate the options in imapoptions until they're actually no longer used!

ksmurchison (approved):

PR #4895, opened 2024-04-22
auth_mboxgroups: a new in-cyrus group mechanism

brong:
This takes groups into Cyrus.  I want to add a JMAP-style management interface as well, but that's for v2.

It also adds additional pusher info to push a list of all users who can see the change, for handy pusher/notify extensions.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Mboxgroups.pm
+++ b/cassandane/Cassandane/Cyrus/Mboxgroups.pm
@@ -1,0 +1,304 @@
+#!/usr/bin/perl
+#
+#  Copyright (c) 2024 Fastmail Pty Ltd. All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions
+#  are met:
+#
+#  1. Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#
+#  2. Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in
+#     the documentation and/or other materials provided with the
+#     distribution.
+#
+#  3. The name "Fastmail Pty Ltd" must not be used to
+#     endorse or promote products derived from this software without
+#     prior written permission. For permission or any legal
+#     details, please contact
+#      FastMail Pty Ltd
+#      PO Box 234
+#      Collins St West 8007
+#      Victoria
+#      Australia
+#
+#  4. Redistributions of any form whatsoever must retain the following
+#     acknowledgment:
+#     "This product includes software developed by Fastmail Pty. Ltd."
+#
+#  FASTMAIL PTY LTD DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE,
+#  INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY  AND FITNESS, IN NO
+#  EVENT SHALL OPERA SOFTWARE AUSTRALIA BE LIABLE FOR ANY SPECIAL, INDIRECT
+#  OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF
+#  USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
+#  TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE
+#  OF THIS SOFTWARE.
+#
+
+package Cassandane::Cyrus::Mboxgroups;
+use strict;
+use warnings;
+use Cwd qw(realpath);
+use JSON;
+use Data::Dumper;
+
+use lib '.';
+use base qw(Cassandane::Cyrus::TestCase);
+use base qw(Cassandane::Unit::TestCase);
+use Cassandane::Util::Log;
+
+sub new
+{
+    my ($class, @args) = @_;
+
+    my $config = Cassandane::Config->default()->clone();
+    $config->set(
+        auth_mech => 'mboxgroups',
+    );
+
+    my $self = $class->SUPER::new({
+        config => $config,
+        adminstore => 1,
+        services => [qw( imap )],
+        start_instances => 0,
+    }, @args);
+
+    return $self;
+}
+
+sub set_up
+{
+    my ($self) = @_;
+
+    $self->SUPER::set_up();
+
+    $self->_start_instances();
+
+    $self->{instance}->create_user("otheruser");
+
+    my $admintalk = $self->{adminstore}->get_client();
+
+    $admintalk->_imap_cmd('SETUSERGROUP', 0, '', 'cassandane', 'group:group c');
+    $admintalk->_imap_cmd('SETUSERGROUP', 0, '', 'cassandane', 'group:group co');
+    $admintalk->_imap_cmd('SETUSERGROUP', 0, '', 'otheruser', 'group:group co');
+    $admintalk->_imap_cmd('SETUSERGROUP', 0, '', 'otheruser', 'group:group o');
+}
+
+sub tear_down
+{
+    my ($self) = @_;
+
+    # clean this up as soon as we're done with it, cause it's holding a
+    # port open!
+    delete $self->{server};
+
+    $self->SUPER::tear_down();
+}
+
+sub test_setacl_groupid
+    :min_version_3_7
+{
+    my ($self) = @_;
+
+    my $admintalk = $self->{adminstore}->get_client();
+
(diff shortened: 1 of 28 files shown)


imap/imapd.c
```diff
@@ -9183,6 +9206,53 @@ static void cmd_list(char *tag, struct listargs *listargs)
     }
 }
 
+/*
+ * Perform a SUBSCRIBE (add is nonzero) or
+ * UNSUBSCRIBE (add is zero) command
```
brong, line 9228: comment is wrong

lib/auth_mboxgroups.c
```diff
@@ -0,0 +1,236 @@
+/* auth_mboxgroups.c -- Mailboxes.db groups authentication
+ *
+ * Copyright (c) 1994-2008 Carnegie Mellon University.  All rights reserved.
```
brong, line 3: Copyright dates need fixing

brong (commented):
Missing replication for usergroups - we need a way to do that before this can be shipped!

brong:
> Missing replication for usergroups - we need a way to do that before this can be shipped!

Did that :)

brong:
damn I need to do some rebasery to make this merge correctly with other branches

ksmurchison (approved):
This all looks sane to me

lib/imapoptions
```diff
@@ -364,7 +364,7 @@ Blank lines and lines beginning with ``#'' are ignored.
    file or mailboxes list entry?  It's noisy so disabled by default, but
    can be very useful for tracking down what happened if things look strange. */
 
-{ "auth_mech", "unix", STRINGLIST("unix", "pts", "krb", "krb5"), "2.3.17" }
+{ "auth_mech", "unix", STRINGLIST("unix", "pts", "krb", "krb5", "mboxgroups"), "2.3.17" }
```
elliefm, line 367: Option has changed, need to bump last updated field to UNRELEASED

```suggestion
{ "auth_mech", "unix", STRINGLIST("unix", "pts", "krb", "krb5", "mboxgroups"), "UNRELEASED" }
```

lib/imapoptions
```diff
@@ -959,7 +959,7 @@ Blank lines and lines beginning with ``#'' are ignored.
...
+{ "event_extra_params", "timestamp", BITFIELD("bodyStructure", "clientAddress", "diskUsed", "flagNames", "messageContent", "messageSize", "messages", "modseq", "service", "timestamp", "uidnext", "vnd.cmu.midset", "vnd.cmu.unseenMessages", "vnd.cmu.envelope", "vnd.cmu.sessionId", "vnd.cmu.mailboxACL", "vnd.cmu.mbtype", "vnd.cmu.davFilename", "vnd.cmu.davUid", "vnd.fastmail.clientId", "vnd.fastmail.sessionId", "vnd.fastmail.convExists", "vnd.fastmail.convUnseen", "vnd.fastmail.cid", "vnd.fastmail.counters", "vnd.fastmail.jmapEmail", "vnd.fastmail.jmapStates", "vnd.cmu.emailid", "vnd.cmu.threadid", "vnd.cmu.visibleUsers"), "3.6.0" }
```
elliefm, line 962: Ditto

```suggestion
{ "event_extra_params", "timestamp", BITFIELD("bodyStructure", "clientAddress", "diskUsed", "flagNames", "messageContent", "messageSize", "messages", "modseq", "service", "timestamp", "uidnext", "vnd.cmu.midset", "vnd.cmu.unseenMessages", "vnd.cmu.envelope", "vnd.cmu.sessionId", "vnd.cmu.mailboxACL", "vnd.cmu.mbtype", "vnd.cmu.davFilename", "vnd.cmu.davUid", "vnd.fastmail.clientId", "vnd.fastmail.sessionId", "vnd.fastmail.convExists", "vnd.fastmail.convUnseen", "vnd.fastmail.cid", "vnd.fastmail.counters", "vnd.fastmail.jmapEmail", "vnd.fastmail.jmapStates", "vnd.cmu.emailid", "vnd.cmu.threadid", "vnd.cmu.visibleUsers"), "UNRELEASED" }
```

elliefm (changes requested):
Looks good generally, but I have a bunch of nits of varying severity, details inline.

Also, I'd especially like to see tests for the replication functionality, probably in Replication.pm.  I assume this code is already minimally exercised by the existing tests just because the new key/value will be in their MAILBOX dlists.  But it would be good to see some tests that tinker with the user's groups and then verify that the expected changes actually appear on the replica

lib/auth_mboxgroups.c
```diff
@@ -0,0 +1,244 @@
...
+#include "auth.h"
+#include "libcyr_cfg.h"
+#include "xmalloc.h"
+#include "strarray.h"
+#include "util.h"
+
+static int (*our_mboxlookup)(const char *userid, strarray_t *sa);
+
+struct auth_state {
+    char userid[81];
```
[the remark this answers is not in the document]
brong: Yeah it's from auth_unix.  No change from previous behaviour.

lib/auth_mboxgroups.c
```diff
@@ -0,0 +1,244 @@
...
+        }
+    }
+
+    return retbuf;
+}
+
+/*
+ * Set the current user to 'identifier'.  'cacheid', if non-null,
+ * points to a 16-byte binary key to cache identifier's information
+ * with.
```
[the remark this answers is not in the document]
brong: ahh yeah that's rubbish from auth_unix too, I'll remove it

cassandane/Cassandane/Cyrus/Mboxgroups.pm
```diff
@@ -0,0 +1,304 @@
...
+
+    # clean this up as soon as we're done with it, cause it's holding a
+    # port open!
+    delete $self->{server};
+
+    $self->SUPER::tear_down();
+}
+
+sub test_setacl_groupid
+    :min_version_3_7
```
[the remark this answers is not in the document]
brong: yep!  Done.

brong:
@elliefm I've addressed the things you mentioned, and also add the Replication test (which raised that in some configs we won't send the groups due to missing RACLs, so I've made it that we always send them.

cassandane/tiny-tests/Replication/mboxgroups
```diff
@@ -0,0 +1,116 @@
...
+    $admintalk->setacl("user.cassandane", "group:shared", "lrs");
+
+    $mastertalk->create("INBOX.Test") || die;
+    $mastertalk->create("INBOX.Test.Sub") || die;
+    $mastertalk->create("INBOX.Test Foo") || die;
+
+    my $ldata = $mastertalk->list("", "*");
+    $self->assert_deep_equals($ldata, [
+          [
+            [
```
elliefm, line 31: You could use `setup_mailbox_structure()` and `assert_mailbox_structure()` here to remove a lot of clutter.  See the List tests for lots of examples.

cassandane/tiny-tests/Replication/mboxgroups
```diff
@@ -0,0 +1,116 @@
...
+    my $replicatalk = $replicastore->get_client();
+
+    my $rdata = $replicatalk->list("", "*");
+    $self->assert_deep_equals($rdata, [
+          [
+            [
+              '\\HasChildren'
+            ],
+            '.',
+            'INBOX'
```
elliefm, line 85: Likewise wrt `assert_mailbox_structure()`.  It would make it easier to see at a glance that the test expects to see brandnew's own mailboxes, plus cassandane's Inbox, without those details getting lost in all the lines of `[[[]]]`

elliefm (commented):
Looks good, couple of details in the new tests that could be streamlined (details inline), but it's not crucial.

But!  This is failing the `Backups.xbackup_shared` and `Backups.shared_mailbox` tests in CI, look like the sync_client side is crashing.  It might be that backupd doesn't recognise the changed protocol and rejects it, and sync_client doesn't know how to proceed after that?  It's not real clear from the CI output, would need to be looked at closely in action.

I might be tempted to say just ignore these failures, since we're getting rid of all the backups code and tests soon anyway (#4896).  But I notice that both of the failing tests are something to do with shared mailboxes, and we have a bad habit of accidentally breaking shared mailboxes code, so I think in this case it warrants a deeper look.  If it is really broken, it would be good to get a regression test for it added to Replication, so we don't lose that coverage when we throw out the Backups stuff.

brong:
@elliefm yes - you're right - it was broken for shared!  I've added a test - Replication.shared_folder so we catch this if it happens again, and fixed it.

elliefm (approved):
Looks good :)

dilyanpalauzov:
Can this get some documentation, please, e.g. at https://www.cyrusimap.org/imap/concepts/overview_and_concepts.html#kerberos-vs-unix-authorization?  How can groups be viewed and edited in mailboxes.db?

PR #4899, opened 2024-04-26
typo:(overview_and_concepts.rst) fix typo

poc-sm:
change listaclmamilbox to listaclmailbox which is the correct command in the exemple


elliefm (approved):
Thanks!

PR #4900, opened 2024-04-29
JMAP Calendars: support vendor-extension properties in VirtualLocations

rsto:
This persists vendor-extensions properties such as "example.com:foo" for VirtualLocation objects. It uses the same approach as the vCard JSPROP property defined for JSContact, but only supports JSCalendar properties having names with vendor-extension prefix.

This PR also adds support for the standard VirtualLocation.features property which hadn't been implemented.


ksmurchison (approved):
LGTM

PR #4902, opened 2024-04-29
JMAP Email: rewrite JMAP Email to MIME conversion

rsto:
This updates how MIME messages are created from JMAP Email objects for Email/set{create}. Changes include:

- Inline plain text is rewritten before conversion regardless if its content is sourced from an EmailBodyValue or Blob. Non-plain text content now never is rewritten, even if it is defined with in a EmailBodyValue.
- If the charset property for inline plain text declares the content to be ASCII or UTF-8, then the EmailBodyPart is rejected for invalid characters.
- The content transfer encoding for "message" media types is chosen by the media subtype, e.g. handling "message/rfc822" and "message/global" differently.
- Bare carriage returns in inline plain text now get expanded to CRLF, rather than omitted from the body part contents.
- Transfer encoding names are written in lowercase, e.g. "8bit" rather than "8BIT", as recommended by

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_blob_set_singlecommand
+++ b/cassandane/tiny-tests/JMAPEmail/email_blob_set_singlecommand
@@ -82,6 +82,7 @@ EOF
                     attachments => [{
                         blobId => '#img',
                         name => "logo.gif",
+                        type => 'image/gif',
                     }],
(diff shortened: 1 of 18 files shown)


rsto:
This is now part of #4926

PR #4904, opened 2024-05-03
docsrc/setup-postfix: fix for socketmap

dilyanpalauzov:
https://www.postfix.org/socketmap_table.5.html says after socketmap:unix:PATH follow colon and a map name.  The map name is sent in the request, as described in the prelude of imap/smmapd.c, but skipped by smmapd.


elliefm (approved):

elliefm:
Thanks!

PR #4906, opened 2024-05-09
vCard: add support for libicalvcard vcardcomponent_transform()

ksmurchison:
This will require a new cyruslibs to be built before the gets merged.


rsto (changes requested):
This generally looks good to me, but if I read this correctly, we now use three different implementations how to convert v3/v4 cards:
1. the lib/vparse one
2. the Cyrus-implementation that builds on top of libicalvcard
3. the vcardcomponent_transform one that's baked into the most recent libicalvcard

We really should have only one implementation, but at least we should not increase the number of implementations. I suggest we get rid of implementation 2, anyone using libicalvcard at this point can also be expected to upgrade to its latest version.

ksmurchison:
> This generally looks good to me, but if I read this correctly, we now use three different implementations how to convert v3/v4 cards:
> 
>     1. the lib/vparse one
> 
>     2. the Cyrus-implementation that builds on top of libicalvcard
> 
>     3. the vcardcomponent_transform one that's baked into the most recent libicalvcard
> 
> 
> We really should have only one implementation, but at least we should not increase the number of implementations. I suggest we get rid of implementation 2, anyone using libicalvcard at this point can also be expected to upgrade to its latest version.

That's the plan, but we can't do that until we update cyruslibs to use the latest libical.  I'd also like you to review vcardcomponent_transform() to make sure that it looks sane.

ksmurchison:
Removed vcard_to_v[3|4]_x() and use vcardcomponent_transform() exclusively

rsto (approved):
LGTM

PR #4908, opened 2024-05-09
imap/smmapd: utilize the userdeny database and possibly skip plus addressing

dilyanpalauzov:
Does lmtp with `autocreate_post:1` create mailboxes, different from `INBOX`, when the envelope recipient is a plus-address?


ksmurchison (approved):
This looks ok to me.

@brong I don't believe that either one these changes will require a change to FM config, but can you confirm?

brong (approved):
Yep, looks fine to me as well.  Might cause more userdeny database lookups but they should be fast.

PR #4909, opened 2024-05-10
jmap_mail: convert Unicode email domains to ASCII in Email/get

rsto:
26a11165390c9cb25c6c72dd08bea24d2c9d2c1d fixed the Cyrus email address parser to accept non-ASCII characters in email address domain names, as required by RFC 6532.

However, until it is clear how to present Internationalized Domain Names (RFC 5890) to JMAP clients, Cyrus will stick to only returning their ASCII representation in the EmailAddress object. JMAP clients still can obtain the Unicode representation by fetching the raw headers of the Email.


ksmurchison (approved):
LGTM

PR #4910, opened 2024-05-10
Fix invalid read for BINARY content-transfer-encoding header

rsto:
This fixes an off-by-one read when monkey-patching a `Content-Transfer-Encoding` header value from `BINARY` to `BASE64`.

Once approved, this should get backported this to at least 3.10 but I'm not sure I need to open a new PR for that?

### DIFF
--- a/cassandane/Cassandane/Cyrus/Simple.pm
+++ b/cassandane/Cassandane/Cyrus/Simple.pm
@@ -280,4 +280,26 @@ sub test_toggleable_debug_logging
     }
 }
 
+sub test_append_binary
+{
+    my ($self) = @_;
+    my $imap = $self->{store}->get_client();
+
+    my $mime = <<'EOF' =~ s/\n/\r\n/gr;
+To: to@local
+From: from@local
+Subject: test
+Content-Transfer-Encoding:binary
+
+test
+EOF
+
+    $imap->append("INBOX", { Binary => $mime });
+    $self->assert_str_equals('ok', $imap->get_last_completion_response());
+
+    $imap->select('INBOX');
+    my $res = $imap->fetch('1', '(BINARY[1])');
+    $self->assert_str_equals("test\r\n", $res->{1}{binary});
+}
+
 1;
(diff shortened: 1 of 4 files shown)


rsto:
CI fails because it uses an outdated IMAPTalk. It seems to use the Debian package, rather than the latest version from CPAN.

ksmurchison (approved):

elliefm:
> CI fails because it uses an outdated IMAPTalk. It seems to use the Debian package, rather than the latest version from CPAN.

You should add a `sub skip_check` to Simple.pm that arranges for `test_append_binary` to be skipped if `$Mail::IMAPTalk::VERSION` is too old.  It wouldn't let the test run, but it would at least not result in a failure when the dependency is too old, and the test will start running without further handholding in the future once there's a new enough Mail::IMAPTalk installed.

Look at JMAPEmailSubmission.pm for a practical working example of `skip_check`, and https://github.com/cyrusimap/cyrus-imapd/blob/master/cassandane/Cassandane/Unit/TestCase.pm#L175-L190 for how it works under the hood

> Once approved, this should get backported this to at least 3.10 but I'm not sure I need to open a new PR for that?

Nah, just stick the backport-to-3.10 label on it (which I've just done), and I'll take care of it next time I'm prepping a release.

If it were a really complicated PR that would be tricky to backport, a separate PR against the release branch with the hard parts already done would be appreciated, but in this case it's trivial, and a separate PR would just complicate things.

elliefm:
> CI fails because it uses an outdated IMAPTalk. It seems to use the Debian package, rather than the latest version from CPAN.

Turns out it does use the version from CPAN, but it uses the version that was on CPAN at the time the docker image was built, and the docker image was last updated last year.

I've created #4912 to make CI update Mail::IMAPTalk from CPAN each time it runs, rather than relying on what's in the docker image.  So this problem will go away forever once that lands.

rsto:
I rebased my PR but it looks like IMAPTalk still isn't the right version?

PR #4912, opened 2024-05-13
github actions: fetch Mail::IMAPTalk each time

elliefm:
This updates our CI workflow to fetch the latest copy of Mail::IMAPTalk each time it runs, instead of relying on the version embedded in the docker image.

It also reports the Mail::IMAPTalk version in the "report version information" step (which now occurs after the module has been updated)


rsto (approved):
Great, thanks.

PR #4915, opened 2024-05-13
JMAP: treat IMAP_NO_OVERQUOTA the same as IMAP_QUOTA_EXCEEDED

ksmurchison:


rsto (approved):

ksmurchison:
Actually, this PR is crap.  It turns out that mailbox_append_index_record() is using the wrong return code for that check that it is making.

PR #4916, opened 2024-05-14
JMAPBackup: use more deterministic timing

ksmurchison:


rsto (approved):

rsto:
The `InProgress.sort` test failed on CI, is that related to this change?

ksmurchison:
Tested all of the JMAPBackup tests on inabox three times both with and without --valgrind.  All 

> The `InProgress.sort` test failed on CI, is that related to this change?

It better not be.  I only touched the JMAPBackup tests and nothing else

rsto (approved):
I'm going to approve these tests and they hopefully will make them more stable. But the failing `Inprogress` test is a hint that overall some cassandane tests still are indeterministic.

elliefm:
```
1) InProgress.sort
 expected '[INPROGRESS', got 'Completed' at Cassandane/Cyrus/InProgress.pm line 296.
```

Interesting: it ran too fast, even with :SlowIO on.

Even more interesting: if it failed in GitHub CI, that means it failed twice in a row!  Because the way GitHub CI runs is, first it runs all of cassandane in quiet mode, and if everything passes, it's done.  But if any tests fail, it re-runs those tests (with `--rerun`) verbosely.  If the originally failing tests pass on a re-run, then CI passes, but if any fail again, then CI fails with the verbose errors.  We do it this way to avoid spurious failures from the non-deterministic tests, and to avoid cluttering the test logs up with verbose output from tests that passed.

I'm going to tell this CI job to re-run, and we'll see if it happens again... If it's consistently happening only on this PR, it would seem to suggest something about this PR is causing it, though I can't for the life of me think why.

elliefm:
They passed this time.  I also built this branch locally and ran the InProgress tests 10 times back to back with no failures.  So I guess InProgress is a little non-deterministic but hasn't gotten any _worse_ from this PR.

>  And Ci doesn't even run the JMAPBackup tests

Hmm.  Perhaps it's worth having it do so.  Add a commit that modifies `cassandane/cassandane.ini.dockertests` to remove JMAPBackup from the `suppress =` line and push that, and then CI should run these tests too.  We can always drop that commit before we merge the PR if we think they're still too unreliable for general use.

PR #4917, opened 2024-05-14
caldav_alarm: suppress duplicate alarms

rsto:
Suppress duplicate calendar alarms in calalarmd. At most one alarm notification is sent for any given alarm action, point in time and calendar event.

Some broken calendar client implementations seem to append the exact same alarm every time they update the event, resulting in duplicate alarms being triggered.

This feature is optional and enabled by default.

This PR also includes duplicate suppression in recurrence rules. This is an internal fix only.

### DIFF
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -56,6 +56,7 @@
 #include "caldav_util.h"
 #include "cyrusdb.h"
 #include "defaultalarms.h"
+#include "hashset.h"
 #include "httpd.h"
 #include "http_dav.h"
 #include "ical_support.h"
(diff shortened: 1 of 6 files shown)


imap/ical_support.c
```diff
@@ -454,6 +528,18 @@ static icaltimetype multirrule_iterator_next(struct multirrule_iterator *iter)
...
     iter->entries[min].next = icalrecur_iterator_next(iter->entries[min].icaliter);
+
+    // now forward all iterators that are at exactly the same recurid
+    for (i = 0; i < iter->nentries; i++) {
+        struct multirrule_iterator_entry *entry = iter->entries + i;
+        if (icaltime_is_null_time(entry->next)) {
+            continue;
+        }
+        if (min != i && !icaltime_compare(next, entry->next)) {
+            iter->entries[i].next = icalrecur_iterator_next(iter->entries[i].icaliter);
```
ksmurchison, line 539: I wonder if rather than having a second loop, this could be part of the first loop as:

```
else if (!icaltime_compare(next, entry->next)) {
            iter->entries[i].next = icalrecur_iterator_next(iter->entries[i].icaliter);
}

This would also bump duplicates of future recurrences until we find the actual minimum next
```

ksmurchison (approved):
LGTM with a comment

imap/caldav_alarm.c
```diff
@@ -614,6 +630,12 @@ static int process_alarm_cb(icalcomponent *comp,
                        icaltime_as_ical_string(start), alarmno,
                        summary);
             }
+            else if (!hashset_add(seen_alarms, &alarm_id) && suppress_duplicates) {
+                syslog(LOG_ERR, "suppressing duplicate alarm "
```
wolfsage, line 634: Do we want to syslog each time we suppress an alarm? What if you have say 1,000,000 suppressions.

Would it be better to count the number of suppressed alarms and log that after?

ksmurchison: I agree.  Keeping a count and logging at the end is probably the best.

rsto: Agreed and done.

rsto: Also demoted the log level from LOG_ERR to LOG_INFO

PR #4918, opened 2024-05-15
replicate user-defined flags on MAILBOX

elliefm:
This means that even unused ones will be replicated, which will make for a more consistent user experience when a replica gets promoted to primary

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -1944,9 +1944,11 @@ static int sync_prepare_dlists(struct mailbox *mailbox,
 {
     struct sync_annot_list *annots = NULL;
     struct mailbox_iter *iter = NULL;
+    struct dlist *ufl = NULL;
     modseq_t xconvmodseq = 0;
     int r = 0;
     int ispartial = local ? local->ispartial : 0;
+    int userflag;
 
     if (!topart) topart = mailbox_partition(mailbox);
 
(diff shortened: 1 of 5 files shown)


imap/sync_support.c
```diff
@@ -2011,6 +2013,13 @@ static int sync_prepare_dlists(struct mailbox *mailbox,
     encode_annotations(kl, mailbox, NULL, annots);
     sync_annot_list_free(&annots);
 
+    /* always send mailbox userflags */
+    ufl = dlist_newlist(kl, "USERFLAGS");
+    for (userflag = 0; userflag < MAX_USER_FLAGS; userflag++) {
+        if (mailbox->h.flagname[userflag])
+            dlist_setatom(ufl, "USERFLAG", mailbox->h.flagname[userflag]);
```
brong, line 2020: Should use `dlist_set_flag` here maybe - it allows things which are legal flags but not legal atoms (yeah, really) to go on the wire naturally.

imap/sync_support.c
```diff
@@ -3072,6 +3083,27 @@ int sync_apply_mailbox(struct dlist *kin,
...
+    /* take all userflags from the master */
+    if (dlist_isatomlist(userflags)) {
+        struct dlist *di;
+
+        for (di = userflags->head; di; di = di->next) {
+            const char *userflag;
+
+            if (di->type != DL_ATOM || !di->sval) continue; /* XXX ??? */
+
+            userflag = di->sval;
```
brong, line 3095: why not use `dlist_toatom` here instead of inspecting inside?

brong (approved):
Ok like this, but better with using the dlist API more idiomatically.

imap/sync_support.c
elliefm: Oh, I missed that `dlist_set_flag` exists.  Yeah I'll use that!

imap/sync_support.c
elliefm: Yeah, didn't realise that one existed either!  Nearby code was just looking at sval directly, and I was uncomfortable enough while copying that to add a paranoid type check, but didn't see a better way.  Thanks for the tip! :)

brong (approved):
Yes, this is all very clear and looks correct.  Thanks!

PR #4919, opened 2024-05-15
mailbox.c: use the correct error code for over message quota

ksmurchison:
IMAP_NO_OVERQUOTA is used for IMAPOPT_QUOTAWARNPERCENT

This will replace #4915 which will be reverted


rsto (approved):

rsto (commented):
I mistakenly approved this and only then saw that CI failed

ksmurchison:
> I mistakenly approved this and only then saw that CI failed

Forced pushed a fix for the test

rsto (approved):

PR #4927, opened 2024-05-28
jmap_ical.c: also repair broken VALARMs in repair_broken_ical

rsto:
Removes VALARMs with no TRIGGER and sets ACTION to DISPLAY if missing.


ksmurchison (approved):
LGTM

PR #4928, opened 2024-06-04
charset.c: also insert whitespace after "phrasing" HTML elements

rsto:
This updates charset_extract to also insert whitespace for any text following HTML elements that until now were considered "phrasing", such as `<em>`, `<span>`, etc. Inserting whitespace emulates content between these elements being separated by CSS.

For example, before this change HTML text like `foo<em>bar</em>` got extracted as "foobar", now it results in "foo bar".

### DIFF
--- a/cunit/charset.testc
+++ b/cunit/charset.testc
@@ -952,7 +952,7 @@ static void test_extract(void)
     /* HTML: OMITTAG tags with and without end tags */
     TESTCASE("<hr>Terry<hr> <hr>Richardson</hr>",
         "us-ascii", ENCODING_NONE, "HTML",
-        " TERRY RICHARDSON ");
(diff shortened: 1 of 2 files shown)


brong (approved):
Looks like it does what it says on the tin and the design is good - I don't see a reason WHY we're making the change, is there a related bug report?

rsto:
The reason is kind of hidden in this: "Inserting whitespace emulates content between these elements being separated by CSS."

We are seeing HTML email text using elements such as `<span>` and `<em>` in a way where they visually separate the content between such elements using CSS. Since our main use case for HTML extraction is in search indexing we want to make sure we do index separate terms for content enclosed in what we so far considered "phrasing".

PR #4929, opened 2024-06-07
github actions: finesse triggers, update actions/checkout

### LINKED ISSUE #4922: CI always fails
On each PR github sends an email with subject “Run failed: Cyrus IMAP CI  - branch (commit-id)” and content: “
Cyrus IMAP CI: All jobs have failed… “View workflow run” - with a hyperlink e.g. to https://github.com/dilyanpalauzov/cyrus-imapd/actions/runs/9147022674 .”

There is nothing, which can be done, once the email arrives.

• Do something so these emails are not sent anymore after each PR.

elliefm:
Few small things here:
* Upgrade to `actions/checkout@v4`, which should fix a warning about "Node.js 16" being deprecated
* Use a wildcard for specifying the release branches, so we don't need to update the list every new release
* Rationalise the `push` and `pull_request` triggers

The last point should prevent:
* double CI runs when pushing to PR branch on main repo (due to "push" and "pull_request" events both firing)
* CI runs for unpublished branches on forks with actions enabled (Fixes #4922)

But still allow:
* CI runs for PRs against official branches (drafts included)
* CI runs for pushes directly to official branches

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,6 +2,10 @@ name: Cyrus IMAP CI
 
 on:
   push:
+    branches:
+      - master
+      - main
+      - 'cyrus-imapd-*'
   pull_request:
     branches:
       - master
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,12 +6,7 @@ on:
     branches:
       - master
       - main
-      - cyrus-imapd-3.10
-      - cyrus-imapd-3.8
-      - cyrus-imapd-3.6
-      - cyrus-imapd-3.4
-      - cyrus-imapd-3.2
-      - cyrus-imapd-3.0
+      - 'cyrus-imapd-*'
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,7 @@ jobs:
         image: cyrusimapdocker/cyrus-buster:latest
         options: --sysctl net.ipv6.conf.all.disable_ipv6=0 --init
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         fetch-depth: 0
     - name: install missing or frequently-updated deps
--- a/docsrc/imap/developer/major-releasing.rst
+++ b/docsrc/imap/developer/major-releasing.rst
@@ -150,12 +150,7 @@ series will be 3.8 (and then master will become 3.9).
    The generated documentation will be under the `doc/html/` directory --
    examine it in your browser to make sure all your formatting and such makes
    sense.
-9. Tell Github Actions about the new branch: edit `.github/workflows/main.yml`
-   and add the new branch to the list in the obvious spot.  Does anything else
-   need to be done for this step?  Unknown... figure it out and document it!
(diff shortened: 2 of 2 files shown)


rsto (approved):

dilyanpalauzov:
> CI runs for unpublished branches on forks with actions enabled

I do not think unpublished branch is a correct term, as my branches are public.

In any case, the problem at https://github.com/cyrusimap/cyrus-imapd/issues/4922 is that the checks fail.  The current change seems to disable the checks, but it should instead resolve the reason for the failures.

elliefm:
@dilyanpalauzov The checks pass for the upstream branches, so when they fail on your branch it indicates a problem with your branch.

In some cases it could just be that there was a problem upstream and we already fixed it, but your branch doesn't have those fixes.  In these cases you can probably resolve it just by rebasing your branch on the current upstream.

In the other cases, it's either a bug in your code that the tests are detecting, or a change in behaviour that the tests need to be updated to accommodate.  I'm happy help you with debugging and/or updating the tests for these ones, just ping me.  But please rebase first and see if that solves it, so that we don't waste time chasing stuff that's already been fixed.

Now that the triggers have been fixed, you'll need to create a PR to have the checks run.    

Also, if you don't already run Cassandane on your dev machine, I'd suggest doing so.  It's near impossible to debug test failures using CI only.

dilyanpalauzov:
I nearly always first get the latest tip of the master branch and then make changes.  So the problem cannot be, that my tip (branch) is too old.  As I always get these errors, whatever the problem cannot be that there is always error in my code.

Can you try to submit a PR from a fork, reverting the last changes, to see if the involvement of a fork repository does constitute triggering emails about failed checks?

elliefm:
Interesting, I guess that means the errors must be something to do with the actions running directly on the fork rather.  I can't think of anything that would cause it to fail in that case, but I'll turn on github actions on my fork and create a branch with the trigger updates reverted, and see what happens

elliefm:
Okay, I understand the problem.  Researching/thinking about a solution...

The problem is that our personal forks don't have up to date release tags.  On my fork, the most recent release tag is cyrus-imapd-3.0.0-beta3 (`git ls-remote --tags elliefm | less`).  What this means is the compiled cyrus detects and reports its version as 3.0, and the Cassandane tests adjust themselves accordingly.  In 3.0, the way to get semi-delayed expunge was with "expunge_mode: default".  Since 3.2, it's "expunge_mode: semidelayed", and "default" is invalid and causes master to exit with EX_CONFIG.  But since Cassandane thinks Cyrus is 3.0, it sets "expunge_mode: default" for the tests that want to use semi-delayed expunge, which then fail because master exits during startup.  I haven't looked at your fork, but I expect it's a similar situation.

We can't push release tags to arbitrary personal forks when doing releases, but having individual contributors needing to remember to fetch and push new release tags to their fork doesn't seem workable either.

I remember there having been something that made the CI always fetch the latest tags from the real upstream, regardless of which source was being built, so that the build would be able to correctly detect and report its own version.  This might have been in the old Travis CI scripts, or it might have been on some Fastmail-internal build system...  I'll look around and see what I can find, and see whether it can be adapted for Github Actions.

elliefm:
@dilyanpalauzov Should be sorted out by #4940

PR #4930, opened 2024-06-10
No relica intermediates

brong:
This puts some hard protections around replicas creating intermediate folders on their own behalf!


ksmurchison (approved):
LGTM

PR #4931, opened 2024-06-12
jmap_api: reject non-array query filter conditions

elliefm:
This fixes `jmap_filter_parse()` to reject filters whose "conditions" is not an array.

### DIFF
--- a/cassandane/tiny-tests/JMAPCalendars/calendarevent_query
+++ b/cassandane/tiny-tests/JMAPCalendars/calendarevent_query
@@ -206,6 +206,48 @@ sub test_calendarevent_query
         wantTotal => 3,
         wantIds => [$eventB1{id}],
         wantFastPath => JSON::false,
+    }, {
+        filter => {
+            operator => 'NOT',
+            conditions => [{
+                blarg => 'foo', # invalid - blarg is not a calevent property
+            }],
+        },
+        wantErr => {
+            type => 'invalidArguments',
+            arguments => [ 'filter/conditions[0]/blarg' ],
+        },
+    }, {
+        filter => {
+            operator => 'NOT',
+            conditions => {  # invalid - object rather than array
+                blarg => 'foo',
+            },
+        },
+        wantErr => {
+            type => 'invalidArguments',
+            arguments => [ 'filter/conditions' ],
+        },
+    }, {
+        filter => {
+            operator => 'NOT',
+            # invalid - no conditions
+        },
+        wantErr => {
+            type => 'invalidArguments',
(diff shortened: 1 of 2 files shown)


elliefm:
@rsto can you take a look at this even though it's still a draft? I'd appreciate some guidance.  Thanks!

It's not clear to me whether it should be an error when "conditions" is completely missing, and if so, if it should be the same kind of error or something different.  For now, I'm treating it like an error.

There's a similar function `jmap_buildfilter()` in this file which also processes query filters, but it doesn't have any error handling at all.  I'm not sure how it's used, nor whether I should leave it alone, or add error handling here too.

imap/jmap_api.c
```diff
@@ -2231,10 +2231,20 @@ HIDDEN void jmap_filter_parse(jmap_req_t *req, struct jmap_parser *parser,
...
+                jmap_filter_parse(req, parser, val, unsupported,
+                                  parse_condition, cond_rock, err);
+                jmap_parser_pop(parser);
+            }
+        }
+        else if (arg) {
+            jmap_parser_invalid(parser, "conditions");
+        }
+        else {
+            /* XXX no conditions... is this an error or not? */
```
rsto, line 2246: Looking at https://jmap.io/spec-core.html#query I think this should report missing `conditions` as invalid. The type-specific query evaluators later are more lenient, e.g. the JMAP Email/query code certainly is. But it's good to filter out non-spec compliant queries.

cassandane/tiny-tests/JMAPCalendars/calendarevent_query
```diff
@@ -206,6 +206,37 @@ sub test_calendarevent_query
         wantTotal => 3,
         wantIds => [$eventB1{id}],
         wantFastPath => JSON::false,
+    }, {
```
rsto, line 209: While you are it, maybe also add a test for an invalid operator? Don't know if we already test this somewhere else.

rsto (commented):
@elliefm updated with my review comments

rsto:
> There's a similar function `jmap_buildfilter()` in this file which also processes query filters, but it doesn't have any error handling at all. I'm not sure how it's used, nor whether I should leave it alone, or add error handling here too.

That function is supposed only to be used after the query arguments got validated. It wouldn't hurt to add error handling there, but it would require to plumb through a `jmap_parser` argument that keeps track of where argument parse tree. I don't think it's worth the effort.

elliefm:
@rsto Thanks! I've tidied it up, and added a test for an invalid operator.

rsto (approved):
LGTM

PR #4934, opened 2024-06-17
sync_support; use read-only lock for first pass

brong:
This is a big ugly, but if we're not in is_repeat for sync_client's update_mailbox_once, we shouldn't take a write lock at all, that way we can keep syncing from a read-only filesystem.


ksmurchison (changes requested):
Looks good, but can we create a test for this?

brong:
@ksmurchison hmm... I dunno, not easily.  The only way I can think that's practical is to take the global lock and then run a sync_client that SHOULD run cleanly and see if it blocks.

ksmurchison (approved):

PR #4936, opened 2024-06-18
http_caldav_sched.c: do not send REPLY for EXDATEs on event creation

rsto:
Fixes a bug where creating a JMAP calendar event having an EXDATE as well as the user as an attendee causes an iTIP REPLY to be sent, even if the PARTSTAT of that attendee is NEEDS-ACTION.


ksmurchison (approved):
LGTM

PR #4937, opened 2024-06-19
imapparse: accept client LITERAL+ for mupdate

elliefm:
The mupdate protocol uses LITERAL+ like syntax in both directions.  This got broken when we made `getxstring()` strict about only accepting it from clients (per RFC 7888: "The non-synchronizing literal form MUST NOT be sent from server to client."), resulting in murder backends being unable to synchronise their mailboxes.db with the mupdate server.  `ctl_mboxlist -m` would fail with an error like "couldn't do LIST command on mupdate server" when parsing the mailboxes from the LIST response.  Our tests didn't detect this because they only run `ctl_mboxlist -m` at startup, when there aren't any mailboxes yet, and the LIST response is empty.

* Adds an additional `ctl_mboxlist -m` run to the post-test shutdown checks for any instances that are configured as a murder backend.
* Adds an mupdate-specific flag to `getxstring()`, and arranges for mupdate-client.c to use it.  When this flag is present, the LITERAL+ syntax is accepted in any direction.

### DIFF
--- a/cassandane/Cassandane/Instance.pm
+++ b/cassandane/Cassandane/Instance.pm
@@ -1476,6 +1476,38 @@ sub _check_cores
     return "Core files found in $coredir" if scalar $self->find_cores();
 }
 
+sub _check_mupdate
+{
+    my ($self) = @_;
+
+    my $mupdate_server = $self->{config}->get('mupdate_server');
+    return if not $mupdate_server; # not in a murder
+
+    my $serverlist = $self->{config}->get('serverlist');
+    return if $serverlist; # don't sync mboxlist on frontends
+
+    # Run ctl_mboxlist -m to sync backend mailboxes with mupdate.
+    #
+    # You typically run this from START, and we do, but at test start
+    # there's no mailboxes yet, so there's nothing to sync, and if
+    # something is broken it probably won't be detected.
+    my $basedir = $self->{basedir};
+    eval {
+        $self->run_command({
+                redirects => { stdout => "$basedir/ctl_mboxlist.out",
+                               stderr => "$basedir/ctl_mboxlist.err",
+                             },
+                cyrus => 1,
+            }, 'ctl_mboxlist', '-m');
+    };
+    if ($@) {
+        my @err = slurp_file("$basedir/ctl_mboxlist.err");
+        chomp for @err;
+        xlog "ctl_mboxlist -m failed: " . Dumper \@err;
+        return "unable to sync local mailboxes with mupdate";
+    }
+}
+
 sub _check_sanity
 {
     my ($self) = @_;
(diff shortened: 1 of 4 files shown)


ksmurchison:
I wonder if a simpler solution is to just do this (untested):

`diff --git a/imap/mupdate-client.c b/imap/mupdate-client.c
index 0934a9db1..920fb87df 100644
--- a/imap/mupdate-client.c
+++ b/imap/mupdate-client.c
@@ -133,6 +133,8 @@ EXPORTED int mupdate_connect(const char *server,
         return MUPDATE_NOCONN;
     }
 
+    prot_setisclient(h->conn->in, 1);
+
     h->saslcompleted = 1;
 
     /* SUCCESS */
`

elliefm:
Maybe, but that would bring us into a world of lying about which prot objects are clients for the side effect of altering literal handling, which is confusing, and will be annoying if we later need to hook other behaviours onto client/not-client status.  It would also hide the reality that the mupdate protocol is slightly different from the IMAP protocol, and so we'd trip over the difference again next time we need to modify getxstring.

This is also why I've called the new flag "GXS_MUPDATE", rather than something more specifically about accepting literal+ from a server.  While the literal+ handling is the only variation mupdate requires (as far as we yet know), if it turns out there's more, or if future IMAP extensions would change the parser in ways that aren't compatible with mupdate, all the mupdate-specific overrides will have a single flag to hang off whose purpose is clear.  And it will be clear to future maintainers of getxstring that a) this function is used for both IMAP _and_ mupdate (despite being in "imapparse.c"), and b) that even though mupdate is generally similar to IMAP, it _is_ different in ways that matter.

I thought about changing the mupdate protocol to behave more like IMAP, so that the parser could remain the same.  But changing it in a backward-compatible way (and testing it...) seemed like its own nightmare, so instead the mupdate protocol remains unchanged, and the parser now knows about the difference(s) from IMAP.

I also thought about duplicating getxstring specifically for mupdate, but decided that having two sets of code to (forget to) test and maintain was worse than just having an extra flag.

elliefm:
Hmm, yeah.  I'll give that a try and see what I think when I see it in my editor.

I originally just replaced all the macros, because I didn't want to have to think about which ones are or aren't used, and it could be a problem if I only redefined the ones that are used, and then someone updated the mupdate code to use other ones that were missing the GXS_MUPDATE flag.  But then, given that I don't think mupdate is likely to ever change, that doesn't seem like a real concern.  It does mean actually changing mupdate though (to use getmstring instead of getstring), ironically!

ksmurchison (approved):

guimard:
Hi @elliefm,

can I apply these changes to Debian packages to fix CVE regression or is it better to wait for next release ?

Best regards,
Xavier

elliefm:
@guimard There will be new releases with the mupdate fixes soon, once I'm confident that the issues introduced by the CVE fix are actually fixed.  There's one thing I'm still checking into first.  Please don't release this patch for Debian yet, because if it turns out that it's not the whole fix, it will just make things more confusing.

PR #4939, opened 2024-06-20
imapd.c: don't split conversations when renaming a user

ksmurchison:


brong (approved):
As already reviewed on slack!  Looks good, this solves the issue by ensuring we don't bump modseqs or recalculate CIDs while renaming a user.

PR #4940, opened 2024-06-21
github actions on forks

elliefm:
Updates github actions to fetch the release tags from the upstream repository if the repository the action is running on is not the upstream repository.

This solves Cassandane test failures on forks that have actions enabled but do not have up to date release tags.  It could also be solved on the fork by updating the release tags, but the cause and solution aren't obvious, so just automate it to avoid confusion.


ksmurchison (approved):
I'm not a GitHub expert, but if this does what it says on the tin, the LGTM

wolfsage (approved):

rjbs (approved):

PR #4943, opened 2024-06-24
configurable abort on fatal errors

elliefm:
When a Cyrus process detects a fatal error, it logs an error message and shuts down with a non-zero exit code.  This PR adds a new imapd.conf option `fatals_abort`, which when enabled, will cause Cyrus to call `abort()` instead of exiting non-zero.  Depending on system configuration, this will usually result in a core dump.

**Fastmail:**
This will conflict with the "Fastmail ONLY - make assertion failures and fatal errors into coredumps" commit on our capstone branch.  To deploy this change at Fastmail, I think we need to do something like:
1. Remove that commit from the capstone branch.  New builds will lack the core dump on fatal error behaviour.
2. Label this PR include-in-fastmail. New builds will have the configurable behaviour, but it won't be enabled.
3. Deploy new builds across our environment.
4. Deploy an imapd.conf change to add `fatals_abort: yes`.  We start getting core dumps on fatal errors again.

If we deploy the imapd.conf update before our Cyrus builds include this PR, we might get errors about the unrecognised option.  So I think it's better to get the code deployed first, and then turn it on afterwards, though this leaves a window where we won't get core dumps.

We should not land this PR on the master branch until our own deployment process is finished, so that we can use include-in-fastmail to control the timing.

### DIFF
--- a/backup/backupd.c
+++ b/backup/backupd.c
@@ -173,6 +173,9 @@ EXPORTED void fatal(const char* s, int code)
         prot_flush(backupd_out);
     }
     syslog(LOG_ERR, "Fatal error: %s", s);
+
+    if (code != EX_PROTOCOL && config_fatals_abort) abort();
+
     shut_down(code);
 }
 
--- a/backup/ctl_backups.c
+++ b/backup/ctl_backups.c
@@ -72,6 +72,9 @@ EXPORTED void fatal(const char *error, int code)
 {
     fprintf(stderr, "fatal error: %s\n", error);
     cyrus_done();
+
+    if (code != EX_PROTOCOL && config_fatals_abort) abort();
+
     exit(code);
 }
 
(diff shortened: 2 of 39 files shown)


rsto (commented):
This does not seem to produce a core dump for failed assertions. Is that on purpose?

elliefm:
> This does not seem to produce a core dump for failed assertions. Is that on purpose?

It ought to, `assertionfailed()` in lib/assert.h has the update to call `abort()`.  Though even if it didn't have that update, it calls `fatal()` and all the fatals now have the abort anyway.

Where are you seeing it not dump core?  Perhaps I should add a test for assertion failures too.

rsto:
Sorry, my error. I must have overlooked the changes in `lib/assert.h` during code review.

rsto (approved):

brong (approved):
This looks great, thanks for moving it from a local patch to a useful configurable option.

PR #4945, opened 2024-06-24
tiny tests: fix propagation of syntax errors

rjbs:
A bug in Cassandane::Tiny::Loader was eating syntax errors.  I have fixed that, then fixed the two tests that were not being run because of that.


rsto (approved):
thanks!

PR #4952, opened 2024-06-29
Remove cmulocal/telnet.m4

dilyanpalauzov:
As far as I see this was never used.


ksmurchison (approved):
LGTM.  @elliefm do you agree?

elliefm:
Yeah, I also don't see anything using these. Thanks!

PR #4953, opened 2024-06-29
compiling.rst: remove RPM packages for SASL DIGEST-MD5

dilyanpalauzov:
As DIGEST-MD5 is gone, installing its libsasl2 plugin, sasl2/libdigestmd5.so, should not be necessary any more.

On Debian libsasl2-modules installs apart from libdigestmd5.so and libcrammd5.so also liblogin.so.  However libsasl2-modules is still required in the documentation on the line before the deleted herein line.

On Fedora cyrus-sasl-md5 does include only libcrammd5.so and libdigestmd5.so.

If there are no github-side RPM-based tests, this change cannot be tried out.


elliefm (approved):
Thanks

PR #4957, opened 2024-07-03
jmap_ical.c: set sanitized CN parameter value in Participant.name

rsto:


imap/ical_support.c
```diff
@@ -3241,6 +3241,44 @@ EXPORTED void icalcomponent_set_jmapid(icalcomponent *comp, const char *id)
...
+    int has_specials = 0;
+    for (const char *c = cn; *c; c++) {
+        if (strchr(MIME_TSPECIALS, *c)) {
+            has_specials = 1;
+            break;
+        }
+    }
+
+    if (has_specials) {
+        struct address *a = NULL;
```
ksmurchison, line 3261: The has_specials loop is correct, but it just occurred to me that the loop and has_specials variable could all be replaced by just:

`if (strcspn(cn, MIME_SPECIALS) < strlen(cn))`

ksmurchison (approved):
LGTM other than my comment, but feel free to ignore if you prefer your code.

imap/ical_support.c
rsto: I like it, thanks :) Updated accordingly.

ksmurchison (approved):

PR #4959, opened 2024-07-03
http_dav.c: squash GCC 14.1.1 warning

ksmurchison:


rsto (approved):

dilyanpalauzov:
Why not prepending instead a `const` at the beginning of the line?

This warning appeared also with gcc 13.2.1 20230731.

ksmurchison:
> Why not prepending instead a `const` at the beginning of the line?

I tried that first and the warning persisted.

dilyanpalauzov:
This is correct.  I have overseen that my actual change back from January is to replace `xmlErrorPtr err = ` with  `const xmlError * err =`.  This does help, but I do not understand right now why does this help.

PR #4961, opened 2024-07-04
jmap_calendar.c: ignore cache errors in CalendarEvent/get

rsto:
Errors during writing the CalendarEvent cache in /get always were meant to be ignored, but an error when starting the write transaction still led to aborting the JMAP request.

This patch updates that to also ignore, but report, such errors.

### DIFF
--- a/imap/jmap_calendar.c
+++ b/imap/jmap_calendar.c
@@ -4092,11 +4092,17 @@ static int jmap_calendarevent_get(struct jmap_req *req)
 
     if (hashu64_count(&rock.cache_jsevents)) {
         r = caldav_begin(db);
-        if (!r) hashu64_enumerate(&rock.cache_jsevents,
+        if (!r) {
+            hashu64_enumerate(&rock.cache_jsevents,
                 cachecalendarevents_cb, &rock);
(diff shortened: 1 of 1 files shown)


imap/jmap_calendar.c
```diff
@@ -4092,11 +4092,17 @@ static int jmap_calendarevent_get(struct jmap_req *req)
...
         r = caldav_begin(db);
-        if (!r) hashu64_enumerate(&rock.cache_jsevents,
+        if (!r) {
+            hashu64_enumerate(&rock.cache_jsevents,
                 cachecalendarevents_cb, &rock);
-        if (r) caldav_abort(db);
-        else r = caldav_commit(db);
-        if (r) goto done;
+            r = caldav_commit(db);
+        }
```
ksmurchison, line 4099: Is there no case in which we want to call caldav_abort()?

rsto: No.

rsto: At least none that I could see to be relevant. We could plumb any errors from the callback up, though.

ksmurchison (approved):

PR #4962, opened 2024-07-04
jmap_contact.c: support X-PHONETIC extension properties

rsto:
Apple macOS Contacts.app allows to store phonetic information for First Name, Middle Name, Last Name and Company. These are stored in the vCard with the X-PHONETIC prefix.

This patch adds the "phoneticFirstName, "phoneticMiddleName", "phoneticLastName" and "phoneticCompany" properties to the Contact type to support these vCard properties in JMAP.

### DIFF
--- a/cassandane/tiny-tests/JMAPContacts/contact_copy_phonetic
+++ b/cassandane/tiny-tests/JMAPContacts/contact_copy_phonetic
@@ -1,0 +1,104 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_contact_copy_phonetic
+  : needs_component_jmap {
+    my ($self)    = @_;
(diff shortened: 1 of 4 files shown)


rsto:
Now updated with "phoneticMiddleName" as Apple's iOS Contacts.app supports setting X-PHONETIC-MIDDLE-NAME.

ksmurchison (commented):
Do we also need to do this for the libicalvcard side of things?

rsto:
libicalvcard only is used for JSContact, right? RFC 9554 adds the standard `PHONETIC` property, so we'll need to handle both properties being present.

ksmurchison (approved):

PR #4965, opened 2024-07-08
cassandane: remove a literal NUL from a test

rjbs:
This is just bizarre!  Why was there a NUL here?  Why does it compile without issue?  Well, apparently NUL is just ignored in some number of places in Perl source, and this is news to me.

Anyway, it isn't a load-bearing NUL.  It isn't part of a string.  But it *did* confuse perltidy, so let's yank it out.

Because GitHub thinks of this as a binary file, here's a diff, with `^@` in place of the NUL:

```diff
--- a/cassandane/tiny-tests/JMAPContacts/card_set_create_media_blob
+++ b/cassandane/tiny-tests/JMAPContacts/card_set_create_media_blob
@@ -5,7 +5,7 @@ sub test_card_set_create_media_blob
     :min_version_3_9 :needs_component_jmap :needs_dependency_icalvcard
 {
     my ($self) = @_;
-    my $jmap = $self->{jmap};^@
+    my $jmap = $self->{jmap};
```


rsto (approved):
Weird.

PR #4966, opened 2024-07-09
smtpclient: add smtp_ehlo_hostname option for EHLO command

### LINKED ISSUE #4960: smtpclient.c hardcodes EHLO=localhost (RFC 2821)
During testing for #4955, I attempted to move from `smtp_backend: sendmail`  to `smtp_backend: host` and ran into an [RFC 2821](https://datatracker.ietf.org/doc/html/rfc2821#section-4.1.1.1) compliance issue as https://github.com/cyrusimap/cyrus-imapd/blob/d42c2c36965b303d753e2b5b7193ff4162ca3717/imap/smtpclient.c#L108 and https://github.com/cyrusimap/cyrus-imapd/blob/d42c2c36965b303d753e2b5b7193ff4162ca3717/imap/smtpclient.c#L369 hardcode `EHLO=localhost`.  Perhaps this could default to `servername` or become a configurable option.

My Postfix instance issued the error `504 5.5.2 <localhost>: Helo command rejected: need fully-qualified hostname` which is how I came across it.

rsto:
Fixes #4960

Testing this manually is easy, but automated testing in Cassandane requires similar changes to the ones for https://github.com/cyrusimap/cyrus-imapd/pull/4956.

### DIFF
--- a/cassandane/Cassandane/Instance.pm
+++ b/cassandane/Cassandane/Instance.pm
@@ -729,6 +729,11 @@ sub _generate_imapd_conf
                 smtp_backend => 'host',
                 smtp_host => $self->{smtphost},
             );
+            if (not defined $config->get("servername")) {
+                # smtpclient needs client_bind_name or servername,
+                # falls back to servername if client_bind isn't set.
+                $config->set(servername => "127.0.0.1");
+            }
         }
     }
     else {
--- a/imap/smtpclient.c
+++ b/imap/smtpclient.c
@@ -102,10 +102,13 @@ enum {
     SMTPCLIENT_CAPA_JMAPID    = (1 << 10),
 };
 
+static const char *smtpclient_ehlo_hostname = NULL;
+
 static struct protocol_t smtp_protocol =
 { "smtp", "smtp", TYPE_STD,
   { { { 0, "220 " },
-      { "EHLO", "localhost", "250 ", NULL,
+        // EHLO hostname will be set in smtpclient_open().
+      { "EHLO", NULL, "250 ", NULL,
         CAPAF_ONE_PER_LINE|CAPAF_SKIP_FIRST_WORD|CAPAF_DASH_STUFFING,
         { { "AUTH", CAPA_AUTH },
           { "STARTTLS", CAPA_STARTTLS },
@@ -143,6 +146,31 @@ static void smtpclient_logerror(smtpclient_t *sm, const char *cmd, int r);
 
 EXPORTED int smtpclient_open(smtpclient_t **smp)
 {
+    if (!smtpclient_ehlo_hostname) {
+        if (config_getswitch(IMAPOPT_CLIENT_BIND)) {
+            smtpclient_ehlo_hostname =
+                config_getstring(IMAPOPT_CLIENT_BIND_NAME);
+        }
+        if (!smtpclient_ehlo_hostname) {
+            smtpclient_ehlo_hostname = config_getstring(IMAPOPT_SERVERNAME);
+        }
+        if (!smtpclient_ehlo_hostname) {
+            xsyslog(LOG_ERR,
+                    "can not determine EHLO hostname, "
+                    "either client_bind/client_bind_name or servername option must be set",
+                    NULL);
+            return IMAP_INTERNAL;
+        }
+        if (smtp_protocol.type == TYPE_STD &&
+            !strcasecmpsafe(smtp_protocol.u.std.capa_cmd.cmd, "EHLO")) {
+            smtp_protocol.u.std.capa_cmd.arg = smtpclient_ehlo_hostname;
+        }
+        else {
+            xsyslog(LOG_ERR, "unexpected smtp_protocol value", NULL);
+            return IMAP_INTERNAL;
+        }
+    }
+
     int r = 0;
     const char *backend = config_getstring(IMAPOPT_SMTP_BACKEND);
 
@@ -366,7 +394,7 @@ static int smtpclient_ehlo(smtpclient_t *sm)
     int r = 0;
 
     /* Say EHLO */
-    buf_setcstr(&sm->buf, "EHLO localhost\r\n");
+    buf_printf(&sm->buf, "EHLO %s\r\n", smtpclient_ehlo_hostname);
     r = smtpclient_writebuf(sm, &sm->buf, 1);
     if (r) goto done;
     buf_reset(&sm->buf);
--- a/cassandane/Cassandane/Net/SMTPServer.pm
+++ b/cassandane/Cassandane/Net/SMTPServer.pm
@@ -46,8 +46,8 @@ sub new_connection {
 }
 
 sub helo {
-    my ($Self) = @_;
-    $Self->mylog("SMTP: HELO");
+    my ($Self, $Host) = @_;
+    $Self->mylog("SMTP: HELO $Host");
     return if $Self->override('helo');
     $Self->send_client_resp(250, "localhost",
                             "AUTH", "DSN", "SIZE 10000", "ENHANCEDSTATUSCODES");
--- a/changes/next/smtpclient_ehlo_hostname
+++ b/changes/next/smtpclient_ehlo_hostname
@@ -1,0 +1,16 @@
+Description:
+
+Changes the SMTP client EHLO hostname from "localhost" to the `client_bind_name` config option, if set and enabled, or `servername` otherwise.
+
+
+Config changes:
+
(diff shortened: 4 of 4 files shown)


imap/smtpclient.c
```diff
@@ -143,6 +146,21 @@ static void smtpclient_logerror(smtpclient_t *sm, const char *cmd, int r);
 
 EXPORTED int smtpclient_open(smtpclient_t **smp)
 {
+    if (!smtpclient_ehlo_hostname) {
+        smtpclient_ehlo_hostname = config_getstring(IMAPOPT_SMTP_EHLO_HOSTNAME);
+        if (!smtpclient_ehlo_hostname) {
+            smtpclient_ehlo_hostname = "localhost";
```
ksmurchison, line 152: I wonder two things:

1. Do we actually need yet another option for a client/hostname?  Couldn't we just use the `servername` ?
2. If we DO want a separate option, should we fallback to `servername` rather than `localhost` ?

rsto: 1. I chose to add a new one because using `servername` would be a breaking change to the current state. All installations that expect localhost in their EHLO would have to change their config.
2. Same reason as for 1, as this is the least intrusive change.

We *can* make this a breaking change, I just chose the less intrusive approach first.

ksmurchison: Perhaps we should ask on the list if anyone cares if we break this.  I'd personally prefer to not add a new option unless its required.  But maybe I'm on an island and others don't mind additional options.

rsto: @elliefm what's your take? If we choose to use an existing option rather than introducing a new one, then we might either want to use `servername` or `client_bind_name`.

ksmurchison: We could use `client_bind_name` and then fall back to `servername`

rsto: I'm fine with all of this, but I'm going to wait for @elliefm to chime in.

rsto: BTW, @amessina who raised this issue asks if we could make it use `servername` if `smtp_ehlo_hostname` is not set. I guess we should decide for one of the following options:

1. Introduce `smtp_ehlo_hostname` and use this and this one only.
2. Introduce `smtp_ehlo_hostname` but if this isn't set, use `client_bind_name`, if that isn't set either use `servername`.
3. Do not introduce `smtp_ehlo_hostname` and use `client_bind_name` if that is set, fall back to `servername` otherwise.
4. Always use `servername`.
5. Always use `client_bind_name`.

ksmurchison: Obviously my vote is for #3 but I won't stand in the way if we reach consensus on one of the other options.

amessina: I think option 3 makes sense.

rsto: Same. So let's do that then.

elliefm: I had to look up the documentation for `client_bind_name` because I wasn't familiar with it...

> **client_bind: 0**
> If enabled, a specific IP will be bound when performing a client  connection.   client_bind_name  is  used  if it is set, otherwise servername is used.  This is useful on multi-homed servers where Cyrus should  not  use other services' interfaces.
> 
> If  not enabled (the default), no bind will be performed.  Client connections will use an IP chosen by the operating system.
> 
> **client_bind_name: <none>**
> IPv4, IPv6 address or  hostname  to  bind  for  client  connections  when client_bind  is  enabled.   If  not set (the default), servername will be used.
> 
> **servername: <none>**
> This is the hostname visible in the greeting messages of  the  POP,  IMAP and  LMTP daemons. If it is unset, then the result returned from gethostname(2) is used.  This is also the value used by murder clusters to identify  the host name.  It should be resolvable by DNS to the correct host, and unique within an active cluster.  If you are using low level replication (e.g. drbd) then it should be the same on each copy and the DNS name should also be moved to the new master on failover.

I don't have a _strong_ opinion, but to the extent that I do have an opinion, it's something like this:

```
if (client_bind) {
    if (client_bind_name) {
        EHLO as client_bind_name
    }
    else {
        EHLO as servername
    }
}
else {
    // we could consider using servername here
    // instead of localhost, but localhost
    // preserves the existing behaviour for
    // trivial deployments (if we care)
    EHLO as localhost
}
```

That is: presumably smtpclient is already using the client_bind options for configuring its outgoing socket, and I think it would make the most sense to EHLO using the same name as was chosen for the socket.  This might be an IP address!  See sections [4.1.1.1 ](https://www.rfc-editor.org/rfc/rfc5321.html#section-4.1.1.1) and [4.1.3](https://www.rfc-editor.org/rfc/rfc5321.html#section-4.1.3) of RFC5321 for the EHLO syntax details when it's not just "localhost".

If smtpclient is currently ignoring client_bind in its socket setup, then we should either address that too, or leave EHLO as "localhost" until we implement client_bind properly.

Regardless, as long as documentation is updated to match, and there's a changes file highlighting the change, then the release notes will highlight the change, and people upgrading will know they might need to adjust other parts of their environment to match.

elliefm: The other option I like is just using `servername` unconditionally (option 4).  My thinking here is that the `client_bind` + `client_bind_name` combo is for overriding `servername` _when choosing a client IP_, but `servername` is still the server's name even in that case.  Without `client_bind`, `client_bind_name` is meaningless, and `servername` is still the server's name.  So in all cases `servername` is the server's name.

I'm about evenly split on whether I think EHLO should honour client_bind or ignore it.

I also read the RFC a bit closer and "localhost" for EHLO is invalid -- it should either be a fully-qualified domain name, or an IP4/6 address with special syntax.  So in my previous comment where I proposed keeping localhost as the non-client-bind fallback, I think we should change to servername.  Which reduces my previous suggestion to this, which is a variation on option 3:

```
if (client_bind && client_bind_name) {
    EHLO as client_bind_name
}
else {
    EHLO as servername
}
```

I no longer think it's useful to preserve the existing "EHLO localhost" behaviour.  It's broken and needs to be fixed, and if someone's stuff is dependent on it, that's broken too.

If `servername` is not a fully-qualified domain name, the EHLO will be broken again (since if it's not a FQDN, EHLO should use an IP address).  But at that point it's a configuration problem, and the server admin can fix their config if it's causing them problems.  So I don't think we need any complicated "fall back to IP if servername isn't a FQDN" logic.

rsto: Done

ksmurchison (approved):

imap/smtpclient.c
```diff
@@ -143,6 +146,31 @@ static void smtpclient_logerror(smtpclient_t *sm, const char *cmd, int r);
...
+        if (!smtpclient_ehlo_hostname) {
+            smtpclient_ehlo_hostname = config_getstring(IMAPOPT_SERVERNAME);
+        }
+        if (!smtpclient_ehlo_hostname) {
+            xsyslog(LOG_ERR,
+                    "can not determine EHLO hostname, "
+                    "either client_bind/client_bind_name or servername option must be set",
+                    NULL);
+            return IMAP_INTERNAL;
+        }
```
elliefm, line 163: don't use `config_getstring(IMAPOPT_SERVERNAME)`, use `config_servername`, which is always set to something because it falls back on `gethostname(2)` (as documented)

cassandane/Cassandane/Instance.pm
```diff
@@ -729,6 +729,11 @@ sub _generate_imapd_conf
                 smtp_backend => 'host',
                 smtp_host => $self->{smtphost},
             );
+            if (not defined $config->get("servername")) {
+                # smtpclient needs client_bind_name or servername,
+                # falls back to servername if client_bind isn't set.
+                $config->set(servername => "127.0.0.1");
+            }
```
elliefm, line 736: And this isn't necessary because `config_servername` is always set, even when that option is blank.

This is also the wrong place to add imapd.conf defaults, because tests cannot override anything that is set here (i.e. the same problem the other PR was having with "imipnotifier: imip").  The things that are configured here are (or should be) only details that Cassandane's own infrastructure depends on, where Cassandane itself would break if a test were to configure something different.

If you need something configured for one test or one suite, add it for that test or suite.  If you need it for every (or most) tests, add it to `sub default` in `Cassandane::Config`.  But in this case it only seems necessary because the implementation in smtpclient is wrong.

imap/smtpclient.c
rsto: I now created https://github.com/cyrusimap/cyrus-imapd/pull/4971

PR #4969, opened 2024-07-10
Update nginx-proxy.rst

luca-rei:
an example of nginx as imapd proxy, with two different streams for encrypt connections and chose imap server based on account name


luca-rei:
Could someone merge this? is only an example on how to use nginx as imapdproxy with imap encryption via stream
thanks

dilyanpalauzov:
Thanks for your contribution.  At your place, if I were uncertain with the English grammar, I would put the text in some free online grammar checkers, and consider the suggestions.

PR #4970, opened 2024-07-12
cassandane: support multiple output formats per run

elliefm:
This adds support to Cassandane for producing multiple formats of output from a single run, as long as they won't be fighting over stdout.  In practice, this means "tap", "pretty", and "prettier" still cannot be used together, but "xml" can be used alongside any of them.  This will become more useful if we add more non-stdout formats in the future (perhaps "json"), but this PR doesn't do that.

There's a bunch of internal architectural changes here, but the main outwardly-visible one should be that you can now specify the `-f <format>` argument to testrunner.pl multiple times.  The other is that the `--rerun` behaviour after test failures now works independently of output format(s).

**Reviewers**: Please pull this branch and run Cassandane in any of the ways you usually would.  If you can test this in an FM inabox or build system setup, even better.  I'm confident that I haven't broken any of my own use cases, but less confident about use cases I don't use myself.  The diff probably makes the most sense when read commit-by-commit.

### DIFF
--- a/cassandane/Cassandane/Unit/Runner.pm
+++ b/cassandane/Cassandane/Unit/Runner.pm
@@ -78,6 +78,19 @@ sub add_formatter
     push @{$self->{formatters}}, $formatter;
 }
 
+# this is very similar to Test::Unit::Result's tell_listeners(), except
+# without the annoying crash when the listener doesn't care about the event
+sub tell_formatters
+{
+    my ($self, $method, @args) = @_;
+
+    foreach my $formatter (@{$self->{formatters}}) {
+        if ($formatter->can($method)) {
+            $formatter->$method(@args);
+        }
+    }
+}
+
 sub do_run
 {
     my ($self, $suite) = @_;
--- a/cassandane/Cassandane/Unit/RunnerXML.pm
+++ b/cassandane/Cassandane/Unit/RunnerXML.pm
@@ -38,39 +38,37 @@
 #
 
 package Cassandane::Unit::RunnerXML;
+use strict;
+use warnings;
+use vars qw($VERSION);
 
 use XML::Generator;
 use Time::HiRes qw(time);
 use Sys::Hostname;
 use POSIX qw(strftime);
 
-use strict;
-use warnings;
-use vars qw($VERSION);
-# XXX should this inherit from our own Cassandane::Unit::Runner?
-use base qw(Test::Unit::Runner);
+use lib '.';
+use base qw(Cassandane::Unit::Runner);
 
 # $Id: XML.pm 27 2004-08-24 11:22:24Z andrew $
 $VERSION = '0.1';
 
 sub new {
-    my ($class, $directory, $generator) = @_;
+    my ($class, $params, @args) = @_;
 
-    $generator ||= XML::Generator->new(escape => 'always', pretty => 2);
+    my $self = $class->SUPER::new(@args);
 
-    return bless({directory => $directory, gen => $generator,
-                  all_tests_passed => 1,
-                  classrecs => {}},
-                 $class);
+    $params->{generator} ||= XML::Generator->new(escape => 'always',
+                                                 pretty => 2);
+
+    $self->{directory} = $params->{directory};
+    $self->{gen} = $params->{generator};
+    $self->{classrecs} = {};
+
+    return $self;
 }
 
-sub all_tests_passed {
-    my ($self) = @_;
-
-    return $self->{all_tests_passed};
-}
-
-sub start {
+sub do_run {
     my ($self, $suite) = @_;
 
     my $result = $self->create_test_result();
@@ -78,6 +76,7 @@ sub start {
     my $start_time = time();
     $suite->run($result, $self);
     $self->_emit_xml();
+    return $result->was_successful;
 }
 
 sub _classrec {
@@ -115,7 +114,6 @@ sub add_failure {
     my $cr = $self->_classrec($test);
     my $tr = $self->_testrec($test);
     $cr->{failures}++;
-    $self->{all_tests_passed} = 0;
     push(@{$tr->{child_nodes}},
          $self->{gen}->failure({type => _extype($exception),
                                 message => $exception->get_message()},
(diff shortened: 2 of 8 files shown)


cassandane/Cassandane/Unit/RunnerXML.pm
```diff
@@ -38,46 +38,45 @@
...
 use Time::HiRes qw(time);
 use Sys::Hostname;
 use POSIX qw(strftime);
 
-use strict;
-use warnings;
-use vars qw($VERSION);
-# XXX should this inherit from our own Cassandane::Unit::Runner?
-use base qw(Test::Unit::Runner);
+use lib '.';
```
rjbs, line 50: Why is this being added?  Feels like a code smell.  The contents of @INC should probably be correct by now, unless I don't know how this works — which is possible!

elliefm: Cassandane expects to run out of its own source tree, which means none of its modules are installed to anywhere `@INC` looks, so all of Cassandane needs to `use lib '.';` before the use lines for its own stuff.  See d03d49befde7

RunnerXML didn't already have this line because it was an oddball that didn't use any Cassandane modules, but now it does so it does.

cassandane/Cassandane/Unit/Runner.pm
```diff
@@ -40,26 +40,28 @@
 package Cassandane::Unit::Runner;
 use strict;
 use warnings;
-use base qw(Test::Unit::TestRunner);
+use base qw(Test::Unit::Runner);
```
rjbs, line 43: This change is kind of weird to me.  This class is called Runner, but used to inherit from TestRunner… which itself inherits from Runner.  The docs are not great here.

I think the deal is that a TestRunner is a more concrete kind of runner, and it used to be that our (Cass's) Runner was just specializing that.  Now that we're building in this multiple formatter jobby, we're shifing our parent one level up toward the abstract.  Fine!

Actually, even better, because we're no longer inhering from something and then changing the signature on its constructor, so I'll sleep easier!

cassandane/testrunner.pl
```diff
@@ -249,13 +246,15 @@ sub usage
...
+        usage unless defined $formatters{$format};
+        $want_formats{$format} = 1;
     }
     elsif ($a =~ m/^-f(\w+)$/)
     {
-        $format = $1;
-        usage unless defined $runners{$format};
+        my $format = $1;
+        usage unless defined $formatters{$format};
+        $want_formats{$format} = 1;
```
rjbs, line 229: I would love to overhaul this code... but your changes are clearly improvements.  They're just building on a mess. :)

cassandane/testrunner.pl
```diff
@@ -384,10 +383,30 @@ sub usage
 {
     # Build the schedule per commandline
     $plan->schedule(@names);
+
     # Run the schedule
-    open my $fh, '>&', \*STDOUT
-        or die "Cannot save STDOUT as a runner print stream: $!";
-    exit(! $runners{$format}->($plan, $fh));
+    $want_formats{tap} = 1 if not scalar keys %want_formats;
```
rjbs: You don't need that `scalar`, it's implied by `not`.

cassandane/testrunner.pl
```diff
@@ -384,10 +383,30 @@ sub usage
...
+    my @writes_to_stdout = grep {
+        $formatters{$_}->{writes_to_stdout}
+    } keys %want_formats;
+    if (scalar @writes_to_stdout > 1) {
```
rjbs, line 364: Don't need that `scalar` either, implied by `>` operator.

cassandane/testrunner.pl
```diff
@@ -385,7 +371,7 @@ sub usage
     $plan->schedule(@names);
 
     # Run the schedule
-    $want_formats{tap} = 1 if not scalar keys %want_formats;
+    $want_formats{prettier} = 1 if not scalar keys %want_formats;
```
rjbs, line 360: (scalar)

rjbs (approved):
Like many heroic acts, I'm glad you did this and sorry it was required. 🤪

This looks like a nice clean refactoring and addition.  I still really dislike the underlying code, but we gotta make do with what we have sometimes!  This cleaned it up a little, which is nice.

I have not yet attempted to run this, only read it.  I will give it a run tomorrowish.

cassandane/testrunner.pl
elliefm: Yeah, I try to be explicit in my intent because it shuts up the "wait, is that a bug???" voice and subsequent rabbit hole dive next time I look at the code after having forgotten writing it -- I'm not in perl often or deeply enough to keep the context rules and exceptions swapped in at all times.  Also theoretically* communicates to the next reader that I considered this and made a deliberate choice, I wasn't merely fumbling pieces together until they worked. ;)

[* It could also communicate the exact opposite, depending what assumptions the reader starts from. Heh /sigh]

cassandane/Cassandane/Unit/Runner.pm
elliefm: Oh yeah this whole thing is a mess, I've slightly vacuumed one corner of one rug at best.  And yeah your interpretation of what's going on here is correct, Cass's Runner has become more abstract.  Specialising TestRunner used to get us output formatting (which we then mostly overrode anyway).  Now our FormatFoo's take care of output formatting, and Runner can just be a Runner (albeit one that needs to know about FormatFoos, because the T:U:Runner doesn't)

cassandane/Cassandane/Unit/RunnerXML.pm
rjbs: This (`use lib '.'`) is the kind of thing I'd expect in the program that uses the modules, not the modules themselves.  For example, `testrunner.pl`.  Then you could delete it from many other places.

cassandane/testrunner.pl
rjbs: My advice is "never use `scalar`".  It's usually confusing for an everyday Perl programmer.  "Why was this needed?  What thing here is special?"  Generally you only need it when you do this:

```perl
%hash = (
  key => EXPR,
  xyz => $foo,
);
```

…and EXPR might evaluate to more than one thing.  But that you could address by assigning `EXPR` to a scalar, and using it.  Anyway, this isn't a big deal, but generally "using scalar will make a Perl code reviewer more confused than not". :)

PR #4974, opened 2024-07-14
developer docs: remove note about Test::Unit problems

rjbs:
It is no longer busted on recent perls, as of 0.026.

### DIFF
--- a/docsrc/imap/developer/developer-testing.rst
+++ b/docsrc/imap/developer/developer-testing.rst
@@ -59,10 +59,7 @@ Install and configure Cassandane
         sudo cpan -i Net::CalDAVTalk
         sudo cpan -i Mail::JMAPTalk
         sudo cpan -i Math::Int64
-
-   With recent perl versions ``Test::Unit::TestRunner`` cannot be installed with ``cpan``.  Download
-   the package and apply the changes from https://sources.debian.org/patches/libtest-unit-perl/0.25-7/
-   before installing it.
+        sudo cpan -i Test::Unit
 
 3. Build Cassandane's binary components
 


docsrc/imap/developer/developer-testing.rst
```diff
@@ -60,10 +60,6 @@ Install and configure Cassandane
         sudo cpan -i Mail::JMAPTalk
         sudo cpan -i Math::Int64
 
-   With recent perl versions ``Test::Unit::TestRunner`` cannot be installed with ``cpan``.  Download
```
dilyanpalauzov, line 63: There is missing a `cpan -i` command above, which installs Test::Unit::TestRunner, or is this somehow an indirect, implicit dependency?

dilyanpalauzov (commented):
.

docsrc/imap/developer/developer-testing.rst
elliefm: It's part of `libtest-unit-perl` which is generally expected to have been installed by apt-get (or similar) in the previous step.  If you haven't used a system package for whatever reason, then it's implied that you'd need to install those dependencies by some other means (such as CPAN).  But this list of CPAN modules is the extra stuff that's not available via apt-get, so it's correct that Test::Unit is missing from it.

dilyanpalauzov: Last time I added cpan-packages in the documentation I was not aware, that first some packages need to be installed over apt-get.  In fact I had no apt-get, so I wrote what I have installed with `cpan -i`.

rjbs:
I've amended the commit to mention installing Test::Unit with cpan.  After all, it shows `libanyevent-perl` and installing AnyEvent via CPAN.

brong (approved):
Yep, this is good - thanks for adding the extra cpan item.

PR #4975, opened 2024-07-15
misc: fix warnings with -O3

elliefm:
There's a bunch of compiler warnings that aren't detected unless certain optimisations are being performed.  This PR fixes the warnings that turn up for me using GCC 12.2.0-14 on Debian Bookworm when compiling with `-O3` in CFLAGS and CXXFLAGS.  If you have a newer compiler to try with, it might find more.

### DIFF
--- a/cassandane/Cassandane/Cyrus/CyrusDB.pm
+++ b/cassandane/Cassandane/Cyrus/CyrusDB.pm
@@ -177,15 +177,17 @@ sub test_recover_uniqueid_from_header_legacymb
     # ctl_cyrusdb -r should find and fix the missing uniqueid
     $self->{instance}->getsyslog();
     $self->{instance}->start();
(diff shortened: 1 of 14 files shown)


elliefm:
Back to draft, gonna fix some that turn up with `-D_FORTIFY_SOURCE` too while I'm in here

rsto:
I'm using the same compiler version.

ksmurchison:
I will test against GCC 13.3.1 and 14.1.1

ksmurchison (changes requested):
Need to add the diff that I supplied

elliefm:
Thanks!

elliefm:
I thought I was going to fix these warnings and then work on the actual problem separately, but the two wound up entangled and it no longer makes sense to keep them separate, so I'm closing this one off.

This work, as well as the actual bug fix, are now in #4980

PR #4976, opened 2024-07-17
message.c: decode Subject header before comparing with conversations

rsto:
Fixes a bug where the subject of a message without cache record is compared in its word-encoded form to the normalized conversation subject. This primarily impacted Sieve jmapquery filters using one of the thread keyword filter conditions.


ksmurchison (approved):
LGTM other than the no-op change to jmap_mail_query.c

rsto:
> LGTM other than the no-op change to jmap_mail_query.c

Thanks, didn't catch this. Fixed.

PR #4977, opened 2024-07-17
Index Message-ID, References and In-Reply-To headers in Xapian

rsto:
This indexes the `Message-ID`, `References` and `In-Reply-To` headers in Xapian and makes them available for JMAP Email queries search with the new `messageId`, `references` and `inReplyTo` JMAP filter conditions.

This pull request builds on the changes of https://github.com/cyrusimap/cyrus-imapd/pull/4926


ksmurchison (approved):
Xapian is not my strong suit, but loosk sane to me.

PR #4979, opened 2024-07-18
sync_support: copy_remote speedup

brong:
Avoid fully parsing and allocating memory for all the records we don't need during copy_remote


ksmurchison (approved):
LGTM.  What a time saver this should be!

Issue #1981, opened 2017-05-22
lmtpd aborts when compiled with optimisation and FORTIFY_SOURCE flags


ajaysusarla:
When I was testing by build(on `Ubuntu 17.04/Zesty`), with the `CFLAGS` containing `-O1 -g -DFORTIFY_SOURCE=1`, I find that `lmtpd` crashes with the following message in the syslog (when I try to send an email):
```
cyrus/master[29136]: process type:SERVICE name:lmtpunix path:/opt/cyrus/libexec/lmtpd age:600.078s pid:18501 signaled to death by signal 6 (Aborted, core dumped)
```
The stack trace from `lmtpd`:
```
(gdb) thread apply all bt

Thread 1 (Thread 0x7fed34043c00 (LWP 29493)):
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:58
#1  0x00007fed30a0d37a in __GI_abort () at abort.c:89
#2  0x00007fed30a4f090 in __libc_message (do_abort=do_abort@entry=1, fmt=fmt@entry=0x7fed30b62796 "%s") at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007fed30a4f0d2 in __GI___libc_fatal (message=0x7fed30b63bd8 "*** invalid %N$ use detected ***\n") at ../sysdeps/posix/libc_fatal.c:185
#4  0x00007fed30a21a3c in printf_positional (s=s@entry=0x7ffc4c4498f0, format=format@entry=0x564b2774dbb6 "250 2.1.5 Ok SESSIONID=<%2$s>", readonly_format=readonly_format@entry=0, ap=ap@entry=0x7ffc4c449b10, 
    ap_savep=ap_savep@entry=0x7ffc4c449488, done=<optimised out>, nspecs_done=<optimised out>, lead_str_end=<optimised out>, work_buffer=<optimised out>, save_errno=<optimised out>, grouping=<optimised out>, 
    thousands_sep=<optimised out>) at vfprintf.c:1898
#5  0x00007fed30a23d1f in _IO_vfprintf_internal (s=s@entry=0x7ffc4c4498f0, format=<optimised out>, format@entry=0x564b2774dbb6 "250 2.1.5 Ok SESSIONID=<%2$s>", ap=ap@entry=0x7ffc4c449b10) at vfprintf.c:1682
[...]

PR #4980, opened 2024-07-19
fix imapd capabilities crash when built with source fortification

elliefm:
This fixes a SIGABRT during the imapd capabilities response when Cyrus was compiled with [glibc source fortification](https://www.gnu.org/software/libc/manual/html_node/Source-Fortification.html) enabled, due to a bad printf format string.  This is not our first time introducing and having to fix this kind of problem (e.g. #1981)

Also fixes a bunch of warnings that are detected when compiled with -O3, because I initially mistook optimisations for the cause and needed to get it building to test it.  Thanks @ksmurchison for a few extra fixes here for warnings that newer GCC's report but mine didn't.

Also smooths out some Cassandane issues with our syslog replacement: we can't reliably inject our syslog replacement when source fortification is enabled, but nor can we reliably detect whether the injection failed in that case.  So now if source fortification was enabled in the Cyrus build, Cassandane just doesn't bother with the syslog replacement.  There were also a bunch of tests that weren't checking `have_syslog_replacement` correctly, so I fixed those too.

When I build this with `-O3` and `-D_FORTIFY_SOURCE=3`, random Cassandane tests fail with "Found some stray processes" in `post_tear_down`.  This seems to be Cassandane's SMTP implementation not reliably killing off all its child processes during shutdown, and they get left behind for Cassandane to clean up (which it does, and then complains about).  I have no idea why this would start happening with these Cyrus build options, since it's all perl.  It might be as dumb as the tests running faster because Cyrus is optimised, and something racing against that.  It doesn't seem worth chasing further right now, since the Cassandane SMTP implementation is already known to be a mess in a bunch of other ways and needs a big refactor.

I expect this will pass CI just fine, because CI doesn't build with optimisations or source fortification.  This will show that I haven't broken anything that already worked!  But if you want to verify the fixes fix things, please build yourself with optimisations and source fortification.

### DIFF
--- a/cassandane/Cassandane/Cyrus/CyrusDB.pm
+++ b/cassandane/Cassandane/Cyrus/CyrusDB.pm
@@ -177,15 +177,17 @@ sub test_recover_uniqueid_from_header_legacymb
     # ctl_cyrusdb -r should find and fix the missing uniqueid
     $self->{instance}->getsyslog();
     $self->{instance}->start();
-    my $syslog = join(q{}, $self->{instance}->getsyslog());
+    if ($self->{instance}->{have_syslog_replacement}) {
+        my $syslog = join(q{}, $self->{instance}->getsyslog());
 
-    # should have still existed in cyrus.header
-    $self->assert_does_not_match(
-        qr{mailbox header had no uniqueid, creating one}, $syslog);
+        # should have still existed in cyrus.header
+        $self->assert_does_not_match(
+            qr{mailbox header had no uniqueid, creating one}, $syslog);
 
-    # expect to find the log line
-    $self->assert_matches(qr{mbentry had no uniqueid, setting from header},
-                          $syslog);
+        # expect to find the log line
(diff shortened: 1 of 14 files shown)


rsto:
>  I have no idea why this would start happening with these Cyrus build options, since it's all perl. It might be as dumb as the tests running faster because Cyrus is optimised, and something racing against that.

I see these errors regularly on my development machine without these build options. I don't think they are related.

rsto (approved):

elliefm:
> > I have no idea why this would start happening with these Cyrus build options, since it's all perl. It might be as dumb as the tests running faster because Cyrus is optimised, and something racing against that.
> 
> I see these errors regularly on my development machine without these build options. I don't think they are related.

Oh that's very interesting, it suggests that it is just due to the tests running faster.  Your dev machine is much newer (and presumably faster) than mine so I guess it hits the issue normally, whereas mine doesn't hit it unless Cyrus was optimised.

PR #4981, opened 2024-07-19
Annotate speedup

brong:
Another one from the list of "things that are expensive in replication split brain"


ksmurchison:
@brong see #4982

ksmurchison (commented):
@brong I think #4982 might be simpler

brong:
Agree,let's work over there.

PR #4982, opened 2024-07-19
annotate: don't re-read mbentry every time if mailbox is open

ksmurchison:
I think this is a simpler/cleaner solution than #4981

### DIFF
--- a/imap/annotate.c
+++ b/imap/annotate.c
@@ -1176,7 +1176,7 @@ static int _findall(struct findall_data *data, void *rock)
 
 static int annotatemore_findall_full(const char *pattern, /* internal */
                          const struct mailbox *mailbox,
-                         const char *mboxname,
+                         const mbentry_t *mbentry,
                          unsigned int uid,
                          const char *entry,
                          modseq_t since_modseq,
@@ -1186,11 +1186,9 @@ static int annotatemore_findall_full(const char *pattern, /* internal */
 {
     int r;
     struct find_rock frock;
-    mbentry_t *mbentry = NULL;
 
     init_internal();
 
-    assert(pattern || mailbox || mboxname);
     assert(entry);
 
     frock.pattern = pattern;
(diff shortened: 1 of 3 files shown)


imap/annotate.c
```diff
@@ -1229,7 +1215,7 @@ static int annotatemore_findall_full(const char *pattern, /* internal */
         struct findall_data data = { .mbentry = mbentry, .is_exactmatch = 1 };
         r = _findall(&data, &frock);
     }
-    else if (!*pattern) {
+    else if (!pattern) {
```
brong, line 1218: I'm not sure if we should be looking for BOTH here - NULL or blank string.

imap/annotate.c
```diff
@@ -1204,19 +1202,7 @@ static int annotatemore_findall_full(const char *pattern, /* internal */
...
-        mboxname = NULL;
-        pattern = "";
-    }
-
-    if (mailbox || mboxname) {
-        if (mailbox) mboxname = mailbox_name(mailbox);
-        r = mboxlist_lookup_allow_all(mboxname, &mbentry, NULL);
-        if (r) goto out;
-    }
+    if (mailbox) mbentry = mailbox_mbentry(mailbox);
```
brong, line 1205: Should we lift this into all the callers as well?

brong (commented):
A couple comments - I'm not sure if this goes quite far enough to normalising things, but looks generally good :)

imap/annotate.c
ksmurchison: Empty string in the previous code was used to signal server entries, but that signaling has changed.  Having said that I have to issue with checking for both.

imap/annotate.c
ksmurchison: I think I had in annotatemore _findall_mailbox() and moved it out.  I don't recall why and it probably wasn't a good reason.  It probably should be there.

rsto (approved):

PR #4984, opened 2024-07-22
prometheus: per-report update frequencies

elliefm:
We found that the Prometheus "usage" report (which reports user and mailbox counts, and quota commitments) is relatively expensive to produce, as it needs to iterate the mailboxes and quota databases.  It is not appropriate to run at the default 10s frequency used by the other reports.

This PR splits the `prometheus_update_freq` imapd.conf option into three new ones: `prometheus_service_update_freq`, `prometheus_master_update_freq`, and `prometheus_usage_update_freq`, so that these reports can be updated at different frequencies.

* `prometheus_service_update_freq` defaults to 10s, like `prometheus_update_freq` did
* `prometheus_master_update_freq` defaults to 0, in which case `prometheus_service_update_freq` will be used instead
* `prometheus_usage_update_freq` defaults to 0, in which case the usage report will not be computed
* The usual deprecated option handling deals with imapd.conf files that still contain the old `prometheus_update_freq` setting

The `/metrics.txt` file that is served by the http_prometheus module remains essentially unchanged: it concatenates the latest version of the reports into a single file whenever it's queried.  Metrics are individually timestamped anyway, so changes to update frequencies will be visible on graphs, but the behaviour doesn't change.

It would be nice to set a useful default for `prometheus_usage_update_freq` instead of disabling it, but we'll need timing data from a production deployment to determine a suitable value.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Prometheus.pm
+++ b/cassandane/Cassandane/Cyrus/Prometheus.pm
@@ -61,6 +61,7 @@ sub new
     $config->set(prometheus_need_auth => "none");
     $config->set(prometheus_service_update_freq => 2);
     $config->set(prometheus_master_update_freq => 2);
+    $config->set(prometheus_usage_update_freq => 2);
 
     return $class->SUPER::new(
         { adminstore => 1,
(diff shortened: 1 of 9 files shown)


rsto (approved):
Implementation and docs look good to me.

PR #4986, opened 2024-07-24
sync: add support for a skipuser touchfile

brong:
This allows skipping replication on one particular user, both rolling and by hand (while using the channel name - you can bypass it by syncing with the IP instead)

This came up after a server failure at Fastmail - it would have been very handy to keep rolling replication going without futzing with the one broken user.  This gives us a tool for this.


ksmurchison (approved):
LGTM, provided CI succeeds, which I assume that it will

PR #4988, opened 2024-07-30
message.c: fix reading cached bodystructure for "message/global"

rsto:
Fixes a bug in which cached bodystructures having a cache version less than the current version 13 failed to parse.

### DIFF
--- a/imap/message.c
+++ b/imap/message.c
@@ -4425,7 +4425,8 @@ err:
     return EOF;
 }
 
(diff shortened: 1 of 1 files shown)


rsto:
Testing this in Cassandane is not feasible since this only impacts parsing cache files older than the Cyrus version being tested. Likewise, mangling the cache file in a Cassandane test file is likely to produce bogus results in the future.
I tested this manually.

ksmurchison (approved):
If it works in testing LGTM

PR #4990, opened 2024-08-02
Create a strarray_contains API and migrate to using it for contains checks

brong:
We had tons of `>= 0` or `< 0` checks for strarray_find and a few bugs where the zero check wasn't done right, so create a more sensible boolean API for the actual question we want to ask!

I also replaced all the uses of strarray_find that are purely of this kind (not using the index for anything else), and grouped the changes into one commit per style of usage (postive vs negative, and comparision or test for -1) for easy reading.

Finally, I have bugfixes for two of the bugs found, and left one complex bug in jmap_mailbox for later examination.


rjbs (approved):
Great change!

Issue #4985, opened 2024-07-24
WebDAV REPORT sync-collection fails with empty sync-token


dilyanpalauzov:
DAVx⁵ 4.4 can do the communication below to synchronize collections.  Indeed, `report_sync_col()` contains: 
```c
r = sscanf((char *) str, SYNC_TOKEN_URL_SCHEME
  "%u-" MODSEQ_FMT "-" MODSEQ_FMT "%1s",
  &uidvalidity, &syncmodseq, &basemodseq,
  tokenuri /* test for trailing junk */);
if (r < 2 || r > 3 …) fctx->txn->error.desc = "Invalid sync-token";
else if (r == 3) {
  if (basemodseq > highestmodseq)
    fctx->txn->error.desc = "Invalid sync-token";
  else if (basemodseq < mailbox->i.deletedmodseq)
    fctx->txn->error.desc = "Stale sync-token";
} 
```
and I think the content of `str` is the content of &lt;sync-token /&gt; - so the empty string or NULL.

https://www.rfc-editor.org/rfc/rfc6578#section-3.4: 

> When the DAV:sync-collection request contains an empty DAV:sync-token element, the server MUST return all member URLs of the collection (taking account of the DAV:sync-level XML element value as per Section 3.3, and optional truncation of the result set as per Section 3.6) and it MUST NOT return any removed member URLs.  All types of member (collection or non-collection) MUST be reported.

For empty &lt;sync-token&gt; Cyrus IMAP returns “Invalid token” instead of all member URLs of the collection.
```
[sync.SyncManager] Starting initial sync
[sync.SyncManager] Number of local non-dirty entries: 0
[sync.SyncManager] Listing changes since null
--> REPORT https://cal.aegee.org/dav/calendars/user/cal/agenda/ h2
Depth: 0
User-Agent: DAVx5/4.4.1.1-ose (2024/07/13; dav4jvm; okhttp/4.12.0) Android/10
Accept-Language: bg-BG, bg;q=0.7, *;q=0.5
Content-Type: application/xml; charset=utf-8
Content-Length: 151
Host: cal.aegee.org
Connection: Keep-Alive
Accept-Encoding: gzip

<?xml version='1.0' encoding='UTF-8' ?><sync-collection xmlns="DAV:"><sync-token /><sync-level>1</sync-level><prop><getetag /></prop></sync-collection>
--> END REPORT (151-byte body)

[...]

PR #4992, opened 2024-08-04
report_sync_col: handle empty <sync-token/>

dilyanpalauzov:
I do not why, but this fixes: https://github.com/cyrusimap/cyrus-imapd/issues/4985.

https://www.rfc-editor.org/rfc/rfc6578#section-3.4:

>  When the DAV:sync-collection request contains an empty DAV:sync-token element, the server MUST return all member URLs of the collection (taking account of the DAV:sync-level XML element value as per Section 3.3, and optional truncation of the result set as per Section 3.6) and it MUST NOT return any removed member URLs. All types of member (collection or non-collection) MUST be reported.

### DIFF
--- a/imap/http_dav.c
+++ b/imap/http_dav.c
@@ -7635,7 +7635,9 @@ int report_sync_col(struct transaction_t *txn, struct meth_params *rparams,
                 syslog(LOG_DEBUG, "scanned token %s to %d %u %llu %llu",
                        str, r, uidvalidity, syncmodseq, basemodseq);
                 /* Sanity check the token components */
-                if (r < 2 || r > 3 ||
+                if (*str == 0) {
+                    /* Initial sync - empty sync-token */
+                } else if (r < 2 || r > 3 ||
                     (uidvalidity != mailbox->i.uidvalidity) ||
                     (syncmodseq > highestmodseq)) {
                     fctx->txn->error.desc = "Invalid sync-token";


ksmurchison (commented):
I'm inclined to do this instead:

```
diff --git a/imap/http_dav.c b/imap/http_dav.c
index 9e7625e99..98eec2032 100644
--- a/imap/http_dav.c
+++ b/imap/http_dav.c
@@ -7621,7 +7621,8 @@ int report_sync_col(struct transaction_t *txn, struct meth_params *rparams,
         xmlChar *str = NULL;
         if (node->type == XML_ELEMENT_NODE) {
             if (!xmlStrcmp(node->name, BAD_CAST "sync-token") &&
-                (str = xmlNodeListGetString(inroot->doc, node->children, 1))) {
+                (str = xmlNodeListGetString(inroot->doc, node->children, 1)) &&
+                xmlStrlen(str) > 0 /* empty token */) {
                 /* Add sync-token to our header cache */
                 spool_cache_header(xstrdup(":token"),
                                    xstrdup((const char *) str), txn->req_hdrs);

```

dilyanpalauzov:
I am fine with any fix.  This should go on all places, where xmlNodeListGetString() is used.  For me it is still unclear, why  xmlNodeListGetString() does not return consistent result, when the xml element is empty.

dilyanpalauzov:
As with libxml2 2.13.4 the problem will disappear, I am inclined not to fix this here.

ksmurchison:
libxml2 regression.  Won't fix here

PR #4993, opened 2024-08-06
imap/telemetry: log rusage at LOG_INFO

svmhdvn:
The syslog RFC 5424 states that LOG_NOTICE should be used for "normal but significant conditions". USAGE logs aren't significant conditions on the system since they aren't actually conditions at all; they are a way to view numerical metrics on resource usage on a particular action.


elliefm (approved):
This seems reasonable, thanks

PR #4703, opened 2023-10-26
Sieve: processimip -> processcalendar, :deletecanceled -> :deletecancelled, add :allowpublic

ksmurchison:
Minimal changes to update to https://www.ietf.org/archive/id/draft-ietf-extra-processimip-03.html

Eventually will rename all "imip" to "cal" or something

PR #4995, opened 2024-08-09
Sieve: update to latest processcalendar spec

ksmurchison:
- will continue to execute processimp bytecode
- upgrades processimip to processcalendar at parse/compile time

This supersedes #4703


elliefm (commented):
Looks good, though a few nits with suggestions, and maybe a bug in sieve_listextensions?

rsto (commented):
Generally looks good to me. In addition to what @elliefm found I wonder if this PR also could switch all existing `processimip` tests to use `processcalendar`? A separate test could then assert that the `processimip` capability still is supported.

elliefm (approved):
Nice

rsto (approved):
great!

PR #5002, opened 2024-08-19
FastMail: test that rename a->c, b->a, c->b works

brong:
Even if replication didn't catch each step of the way - we need this to work.  This adds two case tests, and fixes replication for this case


ksmurchison (approved):
LGTM

Issue #4489, opened 2023-04-27
calendar-color "changes" namespace


WhyNotHugo:
I'm setting a property on a calendar collection:

```xml
<propertyupdate xmlns="DAV:" xmlns:ICAL="http://apple.com/ns/ical/">
  <set>
    <prop>
      <ICAL:calendar-color>#ff00ff</ICAL:calendar-color>
    </prop>
  </set>
</propertyupdate>
```

When I then read the property, it is missing from the response, but a same-named property with a different namespace is present.

```xml
<?xml version="1.0" encoding="utf-8"?>
<multistatus xmlns="DAV:">
  <response>
    <href>/dav/calendars/user/vdirsyncer@fastmail.com/JdvI8QAUmqEEcdvQ/</href>
    <propstat>
      <prop>
        <calendar-color><![CDATA[#ff00ff]]></calendar-color>
      </prop>
      <status>HTTP/1.1 200 OK</status>
    </propstat>
  </response>
```

As a result, a client application will assume that the property is "unset".

WhyNotHugo:
This is still an issue. The server is rewriting the namespace for this property into the wrong namespace.

WhyNotHugo:
The same issue is also present for the `calendar-order` property.

dilyanpalauzov:
The semantics both properites is the same in both namespaces.  Why don’t you set and read the property in the `DAV:` namespace?

WhyNotHugo:
Other clients and servers give special meaning to `{http://apple.com/ns/ical/}calendar-color`, but not to `{DAV:}calendar-color`. For example:

- **Nextcloud** uses the value of `{http://apple.com/ns/ical/}calendar-color` when rendering calendar colours. If I write to `{DAV:}calendar-color`, then Nextcloud won't show a colour for that calendar.

- **Tasks (client)** reads `{http://apple.com/ns/ical/}calendar-color` for calendar colours.

- **SailfishOS's caldav plugin** also uses `{http://apple.com/ns/ical/}calendar-color` for calendar colours.

- [A quick search shows a lot of other clients follow the same behaviour](https://github.com/search?q=apple.com%2Fns%2Fical+calendar-color&type=code).

If I start writing to `{DAV:}calendar-color`, I break interoperability with the rest of the world.

---

[...]

PR #5003, opened 2024-08-19
xml_add_ns: try a different namespace prefix, if the current one is in use

dilyanpalauzov:
The problem is with XML requests, which utilize in a child node a namespace with a prefix, which prefix was already added in the XML response to the root node.

Closes: https://github.com/cyrusimap/cyrus-imapd/issues/4489


ksmurchison:
@dilyanpalauzov Nice find!  But I believe my slightly altered patch is more correct  #5007

ksmurchison (commented):
Can you extend the Caldav:calendar_setcolor test to include the set of problematic PROPFINDs?  And then I will approve and merge.  Thank you

dilyanpalauzov:
I have taken the modified test from https://github.com/cyrusimap/cyrus-imapd/pull/5007.

The very right correction would be to wrap all calls of `xmlNewNs()` in a function, which retries calling `xmlNewNs()` with different prefixes, until `xmlNewNs()` returns non-NULL.

ksmurchison (approved):

dilyanpalauzov:
This should be backported.

PR #5005, opened 2024-08-21
cass smtpd refactor

elliefm:
Refactors Cassandane's smtpd setup to address a few different issue.  This probably makes most sense when reviewed commit by commit.

1) Breaks the smtpd thing out into a standalone script (`utils/fakesmtpd`) that's fork-exec'd rather than just forked.  This means:
    * the smtpd processes no longer share memory with Cassandane, and therefore no longer try to prematurely destruct Cassandane's state when they exit
    * Cassandane's verbose handling no longer conflicts with Net::Server mapping the client socket to stdin/stdout
2) Cassandane now sets up its Cyrus instances with broken smtp configuration, because Cyrus's defaults would send mail to the world if the system's sendmail can do so, and that's a whoopsie we don't need.  Now, if Cyrus tries to send mail from a test that hasn't specifically overridden the broken configuration, then Cyrus's `smtpclient_open()` will fail due to the bad config and no mail will go out.
3) Adds a `want` parameter for tests to request it if they need it, and we don't configure or start it unless they request it.  This shaves a decent amount of time vs starting it for every test like we used to.
4) Marks up all the tests that actually need it so that they get it.
5) Fixes a few other small bugs that were in the way


elliefm:
@ksmurchison The `:needs_foo` stuff is a different kind of magic from this (automatic identification of Cyrus features based on cyr_buildinfo output).

The `:SMTP` magic is pretty difficult to hold correctly, regardless of what it's called.  A test author needs to know whether Cyrus is going to try to make smtpclient_open calls during their test, which it does in some non-obvious places (e.g. JMAP `Identity/get`).  If a test fails because Cyrus made an smtpclient_open call when the test wasn't set up for it, the error is pretty opaque too -- it's not really obvious that the solution is to turn on SMTP in the test unless you already know and are thinking about this problem.  I would like to make it easier to hold, but the name isn't really the problem here.

elliefm:
@ksmurchison I started adding some comments explaining how to use `:SMTP` and how to detect when it was necessary, and in the process of trying to explain it I think I've figured out some things I can rename to make it all a bit clearer.  So, thanks for the prompt about the name!

elliefm:
Also remembered that we already have a general purpose `:want_foo` thing for adding stuff to `want` per-test, so we don't need a separate `:SMTP` magic word for it anyway

elliefm:
@rsto This might also fix your "Found some stray processes" failures we were talking about [here](https://github.com/cyrusimap/cyrus-imapd/pull/4980#issuecomment-2246626558), as those errors no longer occur for me on this branch with optimisations on

rsto (approved):
This looks real great to me. Looking forward to these fixes.

PR #5009, opened 2024-08-28
misc fixes from building on 32-bit architectures with 64-bit time_t

elliefm:
I spent some time a few months ago trying to track down a variety of build failures on armel, armhf, and some others platforms affected by Debian's transition to providing a 64-bit time_t on these 32-bit platforms (#4876, #4887), but eventually stalled out on a stack smashing problem that seemed to be coming from within libical.  @rjbs and I decided that it wasn't worth spending more time chasing the problem around since it makes no sense to run an IMAP service on these sort of devices anyway.

This PR is a distillation of the useful changes I made while investigating.

### DIFF
--- a/backup/lcb_compact.c
+++ b/backup/lcb_compact.c
@@ -143,8 +143,8 @@ static int compact_closerename(struct backup **originalp,
     struct buf ts_index_fname = BUF_INITIALIZER;
     int r;
 
-    buf_printf(&ts_data_fname, "%s.%ld", original->data_fname, now);
-    buf_printf(&ts_index_fname, "%s.%ld", original->index_fname, now);
+    buf_printf(&ts_data_fname, "%s." TIME_T_FMT, original->data_fname, now);
+    buf_printf(&ts_index_fname, "%s." TIME_T_FMT, original->index_fname, now);
 
     /* link original files into timestamped names */
     r = link(original->data_fname, buf_cstring(&ts_data_fname));
--- a/backup/lcb_indexw.c
+++ b/backup/lcb_indexw.c
@@ -609,7 +609,7 @@ static int _index_sub(struct backup *backup, struct dlist *dl,
 
     /* set the unsubscribed time if this is an UNSUB */
     if (!strcmp(dl->name, "UNSUB")) {
-        syslog(LOG_DEBUG, "setting unsubscribed to %ld for %s", ts, mboxname);
+        syslog(LOG_DEBUG, "setting unsubscribed to " TIME_T_FMT " for %s", ts, mboxname);
         struct sqldb_bindval *unsubscribed_bval = &bval[2];
         assert(strcmp(unsubscribed_bval->name, ":unsubscribed") == 0);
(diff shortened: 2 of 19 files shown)


dilyanpalauzov:
As mentioned at https://github.com/cyrusimap/cyrus-imapd/pull/4887#issuecomment-2170631099 , I think configure.ac should call `AC_SYS_LARGEFILE`.

ksmurchison (commented):
Initial testing:

- compiles on Fedora 39 with GCC 13.3.1
- does NOT compile on Fedora 40 with GCC 14.2.1
```
  CCLD     cunit/unit
/usr/bin/ld: cunit/timeofday.o: in function `real_gettimeofday':
/home/ken/cyrus-imapd-upstream/cunit/timeofday.c:203:(.text+0x44d): undefined reference to `__gettimeofday64'

```

elliefm:
Huh, curious.  I wonder what that's about

elliefm:
@ksmurchison can you do something like `ack __gettimeofday64 /usr/include/` on Fedora 40 to see which (if any) of the system headers mention `__gettimeofday64` anywhere, and then email me those header files?  I want to study the defines and stuff.

@rsto Can you try building this PR too please? I'm curious whether it works ok for you, has the same issues, or has different ones.

rsto:
@elliefm it builds fine for me on Debian 12.6 (bookworm)

gcc (Debian 12.2.0-14) 12.2.0
aarch64

CFLAGS="-g -Og -fPIC -W -Wall -Werror -fstack-protector-all -fno-semantic-interposition -Bsymbolic -fno-omit-frame-pointer"
CXXFLAGS="-g -O0 -Wunused -Wall -Wextra -Werror"

elliefm:
@ksmurchison also, what version of libc does Fedora 40 have?  You can find the version with `ldd --version`

ksmurchison:
> @ksmurchison also, what version of libc does Fedora 40 have? You can find the version with `ldd --version`

ldd (GNU libc) 2.39

ksmurchison:
> @ksmurchison can you do something like `ack __gettimeofday64 /usr/include/` on Fedora 40 to see which (if any) of the system headers mention `__gettimeofday64` anywhere, and then email me those header files? I want to study the defines and stuff.

Emailed

elliefm:
@ksmurchison Thanks, that was helpful.  Looks like this is a libc change rather than a Debian/Fedora thing, turns out the macros for dealing with 64-bit time on 32-bit hardware were [changed in libc 2.39](https://sourceware.org/git/?p=glibc.git;a=commit;h=dd535f4f19ef2b5c367a362af445ecadcf45401e).

I've updated cunit/timeofday.c again to accommodate this... hopefully!  It still works for me with libc 2.36, so I don't think I've broken anything that already worked, but I can't test whether I've actually fixed it for 2.39 or not.  Can you try it out again please?  Thanks

ksmurchison:
@elliefm Looks good on my end

ksmurchison (approved):

PR #5011, opened 2024-08-28
TestPlan: check for bad/orphaned tiny-tests

elliefm:
Makes Cassandane check for various bad tiny-tests conditions at startup.  This is mostly useful for detecting mistakes while backporting changes to older versions that don't have tiny tests at all, or don't use them as much as as master does.  It might also be helpful when implementing new tiny-tests suites if you're not just converting an existing suite using the tool.

It complains about:
    * test suites that try to Cassandane::Tiny::Loader a nonexistent directory
    * multiple test suites that Cassandane::Tiny::Loader the same directory
    * tiny-tests directories that no test suite calls Cassandane::Tiny::Loader on
    * tiny-tests directories that don't contain any tests

It's implemented as a sanity check within Cassandane::Unit::TestPlan, not within Cassandane::Tiny, because its main purpose is to be run on the release branches where Cassandane::Tiny doesn't necessarily exist, in order to detect accidentally created tiny-tests files that will never get executed

### DIFF
--- a/cassandane/Cassandane/Unit/TestPlan.pm
+++ b/cassandane/Cassandane/Unit/TestPlan.pm
@@ -481,6 +481,7 @@ sub add_pass
 }
 
 package Cassandane::Unit::TestPlan;
+use File::Find;
 use File::Temp qw(tempfile);
 use File::Path qw(mkpath);
 use Data::Dumper;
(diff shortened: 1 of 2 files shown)


rsto (approved):
I did not run into the issues this PR aims to resolve but I'm not much doing backporting work.
Code looks good to me. If it works, it works.

PR #5015, opened 2024-09-02
caldav_alarm: support suppressing alarms for VEVENT or VTODO

rsto:
This adds the `caldav_alarm_support_components` imapd.conf option to define if alarms should be sent for VEVENT, VTODO, or both. The default is to send alarms for both.

Before this change, VALARMs were processed for any iCalendar component type that libical deems to be a "real" component, that is one of: VEVENT, VTODO, VJOURNAL, VFREEBUSY, VAVAILABILITY, VPOLL, VPATCH, VQUERY, VAGENDA. In practice, RFC 5545 only defines VEVENT and VTODO to contain VALARM components and the temporal semantics for relative alarm triggers would be undefined for the other component types anyway.

Restricting supported components explicitly to VEVENT and VTODO seems to make configuration more clear than some general scheme.

### DIFF
--- a/cassandane/Cassandane/Cyrus/TestCase.pm
+++ b/cassandane/Cassandane/Cyrus/TestCase.pm
@@ -479,6 +479,11 @@ magic(Mboxgroups => sub {
     my $self = shift;
     $self->config_set('auth_mech' => 'mboxgroups');
 });
+magic(CaldavAlarmOnlyVevent => sub {
+    my $self = shift;
+    $self->config_set('caldav_alarm_support_components' => 'VEVENT');
+});
+
 
 # Run any magic handlers indicated by the test name or attributes
 sub _run_magic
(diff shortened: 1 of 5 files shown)


rsto:
This is meant to replace https://github.com/cyrusimap/cyrus-imapd/pull/3284

ksmurchison (commented):
It seems to me that rather than skipping over VTODO entries in caldav_alarm.db, that we shouldn't write them into the DB at all, e.g. skip caldav_alarm_add_record() in mailbox_update_caldav().  Does that make sense, or am I missing something?

rsto:
> It seems to me that rather than skipping over VTODO entries in caldav_alarm.db, that we shouldn't write them into the DB at all, e.g. skip caldav_alarm_add_record() in mailbox_update_caldav(). Does that make sense, or am I missing something?

Right, that will work for new records, I will add that code. I do think we still need to also check in process_alarms for already existing entries, though.

rsto:
@ksmurchison ready for re-review

rsto:
Renamed option name `caldav_alarm_supported_components` to `caldav_alarm_support_components`

PR #5016, opened 2024-09-03
message.c: only read message/global envelope if cache version >= 13

rsto:
Fixes a bug introduced by a2684ec50c63b2e7a965702d4ef83a5387c7a2d5 which made the message cache parser errouneously expect a cached envelope for message/global, causing to an invalid memory read.

Also see 22efff1aa2aa48a71350edd9f926ddff9e109f9a


rsto:
For the same reasons as https://github.com/cyrusimap/cyrus-imapd/pull/4988 this is not feasible to test in Cassandane. I tested it manually.

ksmurchison (approved):

PR #5017, opened 2024-09-04
Resolve clang-tidy warnings for Xapian wrapper

rsto:
The current Xapian wrapper code in `imap/xapian_wrap.cpp` is anything but idiomatic or modern C++ code. https://github.com/cyrusimap/cyrus-imapd/pull/2818 highlights some of that, but we didn't act on it timely and the code now is stale. It also didn't help that the code changes are all in a single commit.

This PR attempts to sanitize the C++ code in smaller steps. As a baseline of what constitutes "sane C++" code, we'll start with the [clang-tidy](https://clang.llvm.org/extra/clang-tidy/) linter with the `bugprone` check category.

I am using this with Debian LLVM version 14.0.6.

### DIFF
--- a/imap/xapian_wrap.cpp
+++ b/imap/xapian_wrap.cpp
@@ -1003,7 +1003,7 @@ static int add_email_part(xapian_dbw_t *dbw, const struct buf *part, int partnum
             dbw->term_generator->set_stopper(NULL);
             dbw->term_generator->index_text(Xapian::Utf8Iterator(val), 1, prefix);
         }
-        if (addr->domain && strcmp(addr->domain, "unspecified-domain")) {
+        if (addr->domain && strcmp(addr->domain, "unspecified-domain") != 0) {
             char *utf8_domain = charset_idna_to_utf8(addr->domain);
             char *puny_domain = charset_idna_to_ascii(addr->domain);
 
@@ -1522,7 +1522,8 @@ EXPORTED int xapian_db_langstats(xapian_db_t *db, ptrarray_t* lstats,
 
 EXPORTED void xapian_query_add_stemmer(xapian_db_t *db, const char *iso_lang)
 {
-    if (strcmp(iso_lang, "en")) db->stem_languages->insert(iso_lang);
+    if (strcmp(iso_lang, "en") != 0)
+        db->stem_languages->insert(iso_lang);
 }
 
 static Xapian::Query* query_new_textmatch(const xapian_db_t *db,
--- a/imap/xapian_wrap.cpp
+++ b/imap/xapian_wrap.cpp
@@ -553,7 +553,7 @@ EXPORTED int xapian_compact_dbs(const char *dest, const char **sources)
 
 /* ====================================================================== */
 
-static std::string get_term_prefix(int partnum)
+static std::string get_term_prefix(enum search_part partnum)
 {
     assert(partnum > SEARCH_PART_ANY && partnum < SEARCH_NUM_PARTS);
 
@@ -590,7 +590,7 @@ static std::string get_term_prefix(int partnum)
     return term_prefixes[partnum];
 }
 
-static Xapian::TermGenerator::stem_strategy get_stem_strategy(int partnum)
+static Xapian::TermGenerator::stem_strategy get_stem_strategy(enum search_part partnum)
 {
     static Xapian::TermGenerator::stem_strategy stem_strategy[SEARCH_NUM_PARTS] = {
         // Version 2 and higher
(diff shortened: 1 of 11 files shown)


rsto:
Once we are done with the clang-tidy changes, let's get back to https://github.com/cyrusimap/cyrus-imapd/pull/2818 and see if there is anything in there that's not addressed already.

imap/xapian_wrap.cpp
```diff
@@ -700,7 +699,9 @@ EXPORTED int xapian_dbw_open(const char **paths, xapian_dbw_t **dbwp,
     std::set<int> db_versions;
     try {
         int flags = Xapian::DB_BACKEND_GLASS|Xapian::DB_RETRY_LOCK;
-        if (nosync) flags |= Xapian::DB_DANGEROUS|Xapian::DB_NO_SYNC;
+        if (mode & XAPIAN_DBW_NOSYNC) {
+            flags |= Xapian::DB_DANGEROUS|Xapian::DB_NO_SYNC;
```
elliefm, line 703: Do we also need something like `mode &= ~XAPIAN_DBW_NOSYNC;` in here?  I notice later on around line 740 we do a `mode == ...` comparison, and that would now fail if this flag has been or'd in.

imap/xapian_wrap.cpp
```diff
@@ -392,7 +392,9 @@ static std::string detect_language(const struct buf *part)
 {
     std::string iso_lang;
     bool reliable = false;
-    CLD2::Language lang = CLD2::DetectLanguage(part->s, part->len, 1, &reliable);
+    CLD2::Language lang = CLD2::DetectLanguage(
+        part->s, part->len < INT_MAX ? static_cast<int>(part->len) : INT_MAX, 1,
+        &reliable);
```
elliefm, line 397: The `, 1,` at the end of that middle line is a little lost, visually.  I've been finding lately that if there isn't logical groupings to wrap function parameters into, or if those logical groupings are still too long, then one parameter per line ends up being the most readable wrapping for me.  And especially so when some of the parameters are complex expressions in their own right, like the ternary there.  What about something like this:

```suggestion
    CLD2::Language lang = CLD2::DetectLanguage(
        part->s,
        part->len < INT_MAX ? static_cast<int>(part->len) : INT_MAX,
        1,
        &reliable);
```

rsto: ok, done.

rsto:
I have updated this PR description, this PR only is meant to fix issues found by the `bugprone` clang-tidy category. I will submit separate PRs for further categories.

ksmurchison (approved):
If it compiles and runs...

imap/xapian_wrap.cpp
```diff
@@ -2089,10 +2090,7 @@ xapian_query_new_match(const xapian_db_t *db, int partnum, const char *str)
         return NULL;
     }
 
-    const char *prefix = get_term_prefix(partnum);
-    if (!prefix) {
-        return NULL;
-    }
+    std::string(get_term_prefix(partnum));
```
elliefm, line 2093: Something fishy here, what's this line do?  The only thing I can see is the side effect of `get_term_prefix()` asserting if partnum is out of range.  If that's the intention, a comment to explain it would be good, because otherwise it looks like dead code that just constructs and discards two std::strings.  It's actually surprising to me that there isn't a warning!

imap/xapian_wrap.cpp
```diff
@@ -1621,17 +1626,17 @@ static Xapian::Query *query_new_listid(const xapian_db_t *db,
 }
 
 static Xapian::Query *query_new_messageid(const xapian_db_t *db __attribute__((unused)),
-                                          const char *prefix,
+                                          enum search_part partnum,
                                           const char *str)
 {
-    return new Xapian::Query(std::string(prefix) + str);
+    return new Xapian::Query(std::string(get_term_prefix(partnum)) + str);
```
elliefm, line 1632: `get_term_prefix()` now returns a `std::string` so I think there's no need to construct another one here?

```suggestion
    return new Xapian::Query(get_term_prefix(partnum) + str);
```

elliefm (commented):
I like these changes, couple of nits though

imap/xapian_wrap.cpp
rsto: Thanks, that was unintentional. I'm surprised there isn't a warning, too.

imap/xapian_wrap.cpp
rsto: Right. Fixed.

elliefm (approved):

PR #5021, opened 2024-09-05
jmap_mail.c: batch Email/copy by source mailbox

ksmurchison:
@rsto I just realized that this PR doesn't aggregate flags from ALL copies of the message in the source account.

I also started wondering if we can leverage the bulk_update code to do most of the work.


ksmurchison:
Superseded by #5023

PR #5022, opened 2024-09-06
cassandane: better handling of feature requirements

elliefm:
This improves the handling of test feature requirements somewhat, making it so that "needs foo bar" can be set for entire suites, not only individual tests.  It then updates test suites for which every test needs some feature to use the new mechanism, and removes the noise from the individual tests.

The diff here is kind of insane because of all the tiny-tests files that have been updated.  I'd suggest reviewing commit by commit.  The first three commits are interesting, the rest are mostly find-replace stuff.

I have more work I want to do in this space -- ideally cassandane would be able to infer needs from wants/magic -- but this is already getting unwieldy, so I'll split the logic changes here and pick up the rest in a separate PR later.


ksmurchison:
I love where you're going with this!

elliefm:
This will need to be updated after #5031 lands

rsto (approved):
Ran this locally, too.

ksmurchison (approved):

PR #5023, opened 2024-09-07
jmap_email.c: refactor Email/copy to use _email_bulkupdate

ksmurchison:
As it says on the tin.  

An alternative to #5021 

I'm sure I missed something and/or we need more tests.

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -11921,11 +11921,15 @@ static void _email_updateplan_error(struct email_updateplan *plan, int errcode, 
     json_t *err = jmap_server_error(errcode);
     int i;
     for (i = 0; i < ptrarray_size(&plan->copy); i++) {
-        struct email_uidrec *uidrec = ptrarray_nth(&plan->copy, i);
-        if (json_object_get(set_errors, uidrec->email_id)) {
-            continue;
+        ptrarray_t *uidrecs = ptrarray_nth(&plan->copy, i);
+        int j;
+        for (j = 0; j < ptrarray_size(uidrecs); j++) {
+            struct email_uidrec *uidrec = ptrarray_nth(uidrecs, j);
+            if (json_object_get(set_errors, uidrec->email_id)) {
+                continue;
+            }
+            json_object_set(set_errors, uidrec->email_id, err);
         }
-        json_object_set(set_errors, uidrec->email_id, err);
     }
     for (i = 0; i < ptrarray_size(&plan->setflags); i++) {
         struct email_uidrec *uidrec = ptrarray_nth(&plan->setflags, i);
@@ -12684,7 +12688,7 @@ static int _email_bulkupdate_plan(struct email_bulkupdate *bulk, ptrarray_t *upd
         hash_iter_reset(plan_iter);
         while (hash_iter_next(plan_iter)) {
             struct email_updateplan *plan = hash_iter_val(plan_iter);
-            if (hash_lookup(plan->mbox_id, &erroneous_plans))
+            if (!hash_lookup(plan->mbox_id, &erroneous_plans))
                 continue;
 
             int i;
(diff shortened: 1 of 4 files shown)


imap/jmap_mail.c
```diff
@@ -11510,6 +11510,12 @@ struct email_update {
     int patch_snoozed;        /* True if snoozed is a patch object */
     struct email_uidrec *snoozed_uidrec; /* Currently snoozed email */
     char *snooze_in_mboxid;   /* Snooze the email in this mailboxid */
+
+    /* JMAP Email/copy specific fields */
+    const char *creation_id;  /* Input: creation id copied message */
```
rsto, line 11515: nit: id _of_ copied message

imap/jmap_mail.c
```diff
@@ -12313,7 +12321,8 @@ static int _email_bulkupdate_plan_keywords(struct email_bulkupdate *bulk, ptrarr
     int i;
 
     /* Open seen.db, if required */
-    if (strcmp(bulk->req->accountid, bulk->req->userid)) {
+    if (bulk->from_accountid ||
```
rsto, line 12324: Maybe change `bulk->from_accountid` from `const char*` to `char *` which never is NULL? Would make it simpler to reason about the cost of malloc is small in comparison to the rest of the code anyway.

imap/jmap_mail.c
```diff
@@ -12898,6 +12914,9 @@ static int _email_bulkupdate_open(jmap_req_t *req, struct email_bulkupdate *bulk
     }
     strarray_fini(&email_ids);
 
+    /* Always use to account to open destination mailboxes */
+    bulk->req->accountid = accountid;
```
rsto, line 12918: This looks dangerous to, this part of the code really shouldn't mess with `req`. It also looks as if this function to reset `req->accountid` to its former value again?

rsto (commented):
I like the idea of reusing the bulk update code for this. The one that sticks out to me is mangling the jmap_req accountid on the fly, which looks error-prone and might even introduce an error in this particular code.

ksmurchison:
@rsto I was able to remove the req->accountid swapping.  The issue was the scoping to ownership by req->accountid that jmap_mbentry_by_uniqueid() was doing.  I created a new API for use by Email/copy which doesn't do that scoping.

imap/jmap_mail.c
```diff
@@ -12313,7 +12322,8 @@ static int _email_bulkupdate_plan_keywords(struct email_bulkupdate *bulk, ptrarr
     int i;
 
     /* Open seen.db, if required */
-    if (strcmp(bulk->req->accountid, bulk->req->userid)) {
+    if (bulk->from_account_id ||  // always need seen.db for Email/copy
```
rsto, line 12325: Did you deliberately keep this nullable? I figure it might be simpler to reason about if its always set but might differ from req->accountid. note that req->accountid also always is set.

rsto (commented):
Adding `mbentry_all` to the jmap API is a good solution. Just having one nit left about the from_accountid field.

imap/jmap_mail.c
ksmurchison: I'm not sure I follow what you're saying. For /copy, accountid and from_accountid will always differ but accountid and userid might be the same, which is why I added the check for from_accountid. 

The alternative would be to open seen.db in email_copy_bulk() if needed and check if it's already open in exec_keywords

rsto: I mean: why can `bulk->from_account_id` either be NULL or some C string value. Can't it be always set to the accountid from which the messages are copied? Just as req->accountid always is set to some C-string, it might be helpful to also know that bulk->from_account_id is set to some value.

ksmurchison: from_account_id being set is the indicator that we're doing a /copy. If it's not set then we're doing /set{update}. If you think it needs clarification, I can add an is_copy field to struct bulkupdate and use that for triggering the copy logic

rsto: yes, I'd prefer either that or rename `from_account_id` to `copy_from_account_id`

rsto (approved):

ksmurchison:
@rsto This should be good to go.  The fix to the bulkupdate error reporting was too aggressive and incorrect.  I falsely assumed that all sub-plan lists were the same.  Only the copy list is a ptrarray_t of ptrarray_t.  The others (delete, setflags, etc) are just flat ptrarray_t.

rsto (approved):

PR #5024, opened 2024-09-07
Caldav/Carddav admin html page: enable editing DAV:displayname

dilyanpalauzov:
I have some changes to the generated admin pages, which enable editing WebDAV properties.

The current, first part reorganizes list_calendars() and list_addressbooks() to enable editing the properties over AJAX.
    
- merge imap/http_carddav.js into imap/http_caldav.js and make the latter more generic, in order to serve both CalDAV and CardDAV administration.
- all calendars are numbered in the generated HTML, starting from zero and each JavaScript function receives as parameter the calendar/addressbook number, on which to enact an action.

I think I have switched away from passing displayname to JavaScript functions, because its content must be HTML-escaped to work reliably, but is not - cf. https://github.com/cyrusimap/cyrus-imapd/pull/3389.


ksmurchison:
This looks good.  I think we would want to rename http_caldav.js to http_calcarddav.js or maybe even http_cal_abook.js

dilyanpalauzov:
> I think we would want to rename http_caldav.js to http_calcarddav.js or maybe even http_cal_abook.js

Renaming http_caldav.js can be done later.

ksmurchison (approved):

PR #5031, opened 2024-09-11
Add support to IETF standardized JMAP Blob & Sieve extensions

ksmurchison:
1. Allow two versions of a JMAP method, by storing a `ptrarray_t` of `jmap_method_t`  in the `settings->methods` hash table - **Done**
2. Update jmap_sieve.c to `urn:ietf:params:jmap:sieve` - **Done**
3. Split JMAPSieve.pm into tiny-tests - **Done**
4. Split JMAPCore.pm into tiny-tests, and split out JMAPBlob.pm - **Done**
5. Split Blob/* functions from jmap_core.c into jmap_blob.c - **Done**
6. Update jmap_blob.c to `urn:ietf:params:jmap:blob` - **Done**


rsto (approved):
I'm not a fan of this whole idea, but this implementation is a good choice under the circumstances.

ksmurchison:
@rsto This is ready for review

rsto (approved):

PR #5032, opened 2024-09-12
Lmtp ratelimit

brong:
We're running into issues in Fastmail production where a single mailbox locked up for a crazy amount of time can cause starvation by having hundreds of lmtp processes waiting on locks.

We could test for locks, but it's OK to have a few waiting - we should just tell the rest to try again later.  This re-uses the
proc_checklimit in lmtpd to check the limits for each userid before trying to lock their conversations.db, and returns a useful 451 code to the sender saying that there are too many connections for this user.

We'll probably set something like a 10 process limit for our servers, which allows for all the MXes to send one and a few spares still open.

### DIFF
--- a/imap/lmtpd.c
+++ b/imap/lmtpd.c
@@ -88,6 +88,7 @@
 #include "mupdate.h"
 #include "notify.h"
 #include "prometheus.h"
+#include "proc.h"
 #include "prot.h"
 #include "proxy.h"
 #include "slowio.h"
@@ -151,6 +152,8 @@ static int dupelim = 1;         /* eliminate duplicate messages with
 static int singleinstance = 1;  /* attempt single instance store */
 static int isproxy = 0;
 static strarray_t *excluded_specialuse = NULL;
+static const char *lmtpd_clienthost = "[local]";
+static struct proc_handle *proc_handle = NULL;
 
 static struct stagemsg *stage = NULL;
 
@@ -256,6 +259,7 @@ int service_init(int argc __attribute__((unused)),
 int service_main(int argc, char **argv,
                  char **envp __attribute__((unused)))
 {
+    const char *localip, *remoteip;
     int opt;
 
     struct io_count *io_count_start = NULL;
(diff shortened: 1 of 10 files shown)


brong:
Sadly there's no test cases for maxlogins in general, and it's particularly hard to test for this case, because it only limits while the lmtp delivery itself is happening, so you'd need to hold a lock on the user's conversationsdb!

rsto (approved):
Approved under the condition that we won't merge until further testing shows this is sane. Tagged as Do-Not-Merge.

lib/proc.c
```diff
@@ -342,6 +342,10 @@ static int procusage_cb(pid_t pid __attribute__((unused)),
     /* we only count logged in sessions */
     if (!userid) return 0;
 
+    /* only check for logins to the particular protocol */
+    if (limitsp->servicename && strcmp(servicename, limitsp->servicename))
```
rsto, line 346: nit: Is `servicename` ever allowed to be NULL?

brong: it's not called with it being NULL on this codepath, but you could in theory ask for a total set of ALL the connections that a user has.  We do that in another case: when we're killing all user connections.

brong:
I rebased and merged in some patch code for the return code - I had originally had it coming from lib/proc.c but we don't have IMAP errors there, and when I removed that bit I forgot to add it to the caller.  That's now fixed :)

imap/lmtpd.c
```diff
@@ -883,11 +893,25 @@ int deliver(message_data_t *msgdata, char *authuser,
...
             struct imap4flags imap4flags = { &flags, authstate };
 
+            const char *userid = mbname_userid(mbname);
+            struct proc_limits limits;
+            limits.servicename = config_ident;
+            limits.clienthost = lmtpd_clienthost;
+            limits.userid = userid;
+            if (proc_checklimits(&limits)) {
+                if (limits.maxhost && limits.maxhost <= limits.host) r = IMAP_LIMIT_HOST;
+                else if (limts.maxuser && limits.maxuser <= limits.user) r = IMAP_LIMIT_USER;
```
elliefm, line 903: `limts` typo... why isn't this warning??

elliefm: Oh it's a compile error, good, and that's why CI is failing:

```
imap/lmtpd.c: In function 'deliver':
imap/lmtpd.c:903:26: error: 'limts' undeclared (first use in this function); did you mean 'limits'?
                 else if (limts.maxuser && limits.maxuser <= limits.user) r = IMAP_LIMIT_USER;
                          ^~~~~
                          limits
imap/lmtpd.c:903:26: note: each undeclared identifier is reported only once for each function it appears in
  CC       imap/lmtpd-sync_support.o
```

I was worried we might have some stray `limts` symbol somewhere that was being used instead

brong: yeah, just a typo -fixed and pushed!

brong:
This has been running happily in Fastmail production and legitimately delaying some mail when people are getting hammered, so it's doing the job it was designed to do!

PR #5034, opened 2024-09-12
xapian_wrap.cpp: zero-terminate string passed to CLD2 library

rsto:


ksmurchison (approved):
Assuming CI passes, which it should.

PR #5036, opened 2024-09-15
Master babysit retry

brong:
I recently moved sync_client to be managed by Cyrus rather than a separate "monitorsync" script.

Likewise, we have other things started under the DAEMON block which are supposed to run forever at Fastmail:

pusher, calalarmd, squatter, idled, bayesrpcd and one day again: promstatd.

This change means that if we hit the restart limit, we retry every 10 seconds forever for anything with the 'babysit' flag - which includes all daemons.  We only try once per 10 seconds unless we get a clean 10 seconds, in which case it resets back to allowing 5 failures with immediate restart.  This should stop it causing a totally crazy amount of restarts.

Finally: I added the string "ERROR" to the messages about hitting the limits, which will help with Fastmail's log tracking.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Master.pm
+++ b/cassandane/Cassandane/Cyrus/Master.pm
@@ -531,6 +531,10 @@ sub test_service_exit_during_start
     $self->assert_null($lemm);
     $self->assert_deep_equals({ A => { live => 0, dead => 5 } },
                               $self->lemming_census());
+
+    xlog $self, "check that the error was syslogged";
+    my @loglines = $self->{instance}->getsyslog();
+    $self->assert(grep { m/too many failures for service/ } @loglines);
 }
 
 sub test_startup
--- a/cassandane/Cassandane/Cyrus/Master.pm
+++ b/cassandane/Cassandane/Cyrus/Master.pm
@@ -532,9 +532,12 @@ sub test_service_exit_during_start
     $self->assert_deep_equals({ A => { live => 0, dead => 5 } },
                               $self->lemming_census());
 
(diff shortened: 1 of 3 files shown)


cassandane/Cassandane/Cyrus/Master.pm
```diff
@@ -531,6 +531,10 @@ sub test_service_exit_during_start
     $self->assert_null($lemm);
     $self->assert_deep_equals({ A => { live => 0, dead => 5 } },
                               $self->lemming_census());
+
+    xlog $self, "check that the error was syslogged";
+    my @loglines = $self->{instance}->getsyslog();
+    $self->assert(grep { m/too many failures for service/ } @loglines);
```
elliefm, line 537: This will give false failures if the syslog replacement injection had failed.  Have a look at `$self->assert_syslog_matches()`... I can't remember if it's exactly suitable here, but if it's not, at least the implementation will show the right way to handle this sort of thing.

elliefm (commented):
Looks okay, one nit in the test update.  I wonder if we can test the new behaviour somehow.

cassandane/Cassandane/Cyrus/Master.pm
brong: yes, can probably just use that :)  Thanks

brong: I added this!

We can test new behaviour by hand by killing something every second and watching the logs

elliefm (approved):

ksmurchison (approved):

PR #5038, opened 2024-09-16
Makefile.am: 4-space indents, except where we can't

elliefm:
This file is an Automake file, which means we can use 4-space indents like the rest of our code.  We were already doing this in some places in this file; now we do it throughout.

The exceptions are the snippets of Makefile syntax, where hard tabs are required by make.

Obsoletes #3438


ksmurchison (approved):

elliefm:
Marking this one for backport because it's likely to cause conflicts with every other thing I ever want to backport, so might as well just get it out of the way first

PR #5039, opened 2024-09-16
imapd.c: consolidate error reporting for admins-only commands

ksmurchison:


imap/imapd.c
```diff
@@ -2892,6 +2892,16 @@ static void cmdloop(void)
                     " after UIDONLY is enabled\r\n", tag.s, cmd.s);
         eatline(imapd_in, c);
         continue;
+
+    adminsonly:
+        /* administrators only please */
+        syslog(LOG_ERR, "Invalid user %s trying to use %s command",
```
rsto, line 2898: I find the use of "invalid user" here strange, isn't that rather "unauthorized"? But I know you went with what log message was used before, I just don't think it wasn't a particularly good message either.

rsto (approved):

rsto (approved):

PR #5048, opened 2024-09-19
iTIP with URL (percent) encoded addresses

ksmurchison:
This solves the original user issue using a plus address to create a scheduling event over JMAP.  I need to audit the rest of the JMAP and CalDAV code to handle percent encoded addresses.

### DIFF
--- a/cassandane/tiny-tests/JMAPCalendars/calendarevent_set_plusaddr_itip
+++ b/cassandane/tiny-tests/JMAPCalendars/calendarevent_set_plusaddr_itip
@@ -1,0 +1,79 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_calendarevent_set_plusaddr_itip
+{
+    my ($self) = @_;
+
+    my $jmap = $self->{jmap};
+    my $caldav = $self->{caldav};
+
+    xlog "Set scheduling addresses via CalDAV";
+    my $xml = <<'EOF';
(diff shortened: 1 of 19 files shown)


rsto (commented):
Looks good so far, will re-review when this is completed.

cassandane/tiny-tests/JMAPCalendars/calendarevent_set_plusaddr_itip
```diff
@@ -0,0 +1,89 @@
...
+                        }
+                    }
+                }
+            }
+        }, 'R1'],
+    ]);
+    my $id = $res->[0][1]{created}{1}{id};
+    $self->assert_not_null($id);
+
+    xlog "Assert that iTIP notification is sent";
```
rsto, line 69: Are all these asserts necessary to validate that plus-addresses trigger an iTIP notification? I think this could all be made simpler by using `assert_caldav_notified`, e.g. at https://github.com/cyrusimap/cyrus-imapd/blob/master/cassandane/tiny-tests/Caldav/invite#L65

rsto (approved):
I think this all looks great, but the JMAP test code looks unnecessary complex to me.

cassandane/tiny-tests/JMAPCalendars/calendarevent_set_plusaddr_itip
ksmurchison: Simplified

PR #5050, opened 2024-09-20
iTIP: convert \n from iCalendar to “\r\n  <br>” to the MSA

dilyanpalauzov:
When an `\n` from an iCalendar property value is converted in http_caldav_sched.c:imip_send_sendmail() to HTML, the function used to send to the MSA `\n  <br>`.  That is: a single LF without CR.  This does not work any more, the MSA replies:
```
smtpclient: unexpected response to EOT: code=421 text=4.5.0 Bare linefeed (LF) not allowed
smtpclient: Bad protocol during DATA
```

This change sends in such cases `\r\n` instead of `\n`.


ksmurchison (approved):

dilyanpalauzov:
This should be backported … to 3.0..3.10.

PR #5055, opened 2024-09-25
imapd: track characters used in IMAP tags

elliefm:
This extends the existing IMAP client behaviour tracking to record categories of characters used in tags

We don't have existing tests for the client behaviour logging, except for a couple of cursory checks in Simple and ID. I might feel more comfortable with some tests that verify the character class mappings detect as expected, but it'd be a whole new test suite, is it worth the effort?  I dunno.

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -1565,6 +1565,9 @@ static void record_client_tag_behaviour(const char *tag)
         return;
     }
 
+    /* if we haven't returned yet there's some odd stuff in there */
+    xsyslog(LOG_DEBUG, "saw an unusual tag", "tag=<%s>", tag);
+
     /* inclusive checks */
     if ((saw & TAG_SAW_DOT)) {
         client_behavior.did_tag_dot = 1;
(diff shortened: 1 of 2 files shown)


rsto (approved):

ksmurchison (approved):

elliefm:
We think we don't want to actually merge this, just include it in our builds until we've got the data we're interested in, then drop it.

rsto:
@elliefm I removed the `include-in-fastmail` tag from this PR because it conflicts with the main branch.

elliefm:
@rsto I've fixed the conflicts, so we could re-tag it now.  But actually, I think I'll take the opportunity to change the data it collects based on what we saw from the first run.  So let's not re-tag it just yet, wait till I've got it ready for the next lot of stats.

elliefm:
@ksmurchison @rsto I've updated this to log unusual tags at LOG_INFO rather than LOG_DEBUG, and improve the tracking and reporting based on shortcomings I noticed in the original stats.  The last 3 commits are the new ones.

ksmurchison (approved):

elliefm:
No longer need this, we have the data

PR #5057, opened 2024-09-25
Caldav.pm: add iTIP regression test for event timezone changes

rsto:
This adds a couple of regression tests that assert that changing the temporal properties of a VEVENT generally triggers an iTIP message. This includes changing the timezone of DTSTART or DTEND, even if the new timezone has the same UTC offset as the old one.

The one exception is changing from DTEND to DURATION does not trigger iTIP, if the new event duration has the same time span as the old one. See Caldav.invite_switch_duration_to_dtend for that.


ksmurchison (approved):
Looks good, provided CI passes

PR #5058, opened 2024-09-26
message.c: parse unbracketed dot-atom-text message-ids

rsto:
This updates `message_iter_msgid` to parse ids in the form

    dot-atom-text ['@' dot-atom-text]

with or without enclosing angle brackets.

As additional restriction each label in the dot-atom-text must start with an alphanumeric. Note that the domain part is optional, which is invalid according to RFC 5322.

We parse this form because some mail agents produce invalid message id headers such as `3a53c9f21a8e41f0bb9558c8029e16f0` or `<3a53c9f2-1a8e-41f0-bb95-58c8029e16f0>` (e.g. dumping hex-encoded uuids in the header value, with or without bracket).

As the current message id parser already accepts all kinds of invalid values, this just makes it some more lenient. It brings the advantage to support threading emails with such invalid header values.


ksmurchison (approved):
LGTM

rsto:
Fixes a bug when parsing malformed angle-bracket ids

PR #5059, opened 2024-09-26
Update CalDAV to draft-ietf-calext-vpoll-06

ksmurchison:


dilyanpalauzov:
You can replace draft-york-vpoll with draft-ietf-calext-vpoll in rfc-support.rst.

rsto (approved):

PR #5060, opened 2024-10-01
proppatch_scheddefault: store the value for the Inbox’s owner

dilyanpalauzov:
When user `a` has `w` rights on `b`’s scheduling Inbox (user/b/#calendars/Inbox), then calling
```
curl -XPROPPATCH -Hcontent-type:application/xml -ua:a --data-binary @- <<EOF http://server/dav/calendars/user/b/Inbox
<propertyupdate xmlns="DAV:">
  <set>
    <prop>
      <c:schedule-default-calendar-URL xmlns:c="urn:ietf:params:xml:ns:caldav">
        <href>/dav/calendars/user/a/c</href>
      </c:schedule-default-calendar-URL>
    </prop>
  </set>
</propertyupdate>
EOF
```
must be rejected, as b’s default scheduling calendar must be owned by b (and the database stores anyway only the last part of the URL - here `c`).

Before this change, when `a` changed the schedule-default-calendar-URL on b/Inbox, then a/Inbox was in fact modified, not b‘s. 

As the property is not masked, it must be stored either for "" (shared namespace), or the mailbox owner (b).  This does not mean that it can be stored always for `httpd_userid`.

Disallow setting as default scheduling calendar `Inbox`, `Outbox` and `Attachments`.

### DIFF
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -955,9 +955,13 @@ static int proppatch_scheddefault(xmlNodePtr prop, unsigned set,
                                   void *rock __attribute__((unused)))
 {
     /* Only allow PROPPATCH on CALDAV:schedule-inbox-URL */
-    if ((pctx->txn->req_tgt.flags != TGT_SCHED_INBOX) || !set) {
+    if ((pctx->txn->req_tgt.flags != TGT_SCHED_INBOX) || !set || pctx->txn->req_tgt.resource) {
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
Should for non-mask-able properties the userid be `""` or the mailbox-owner?

The property `schedule-default-calendar-URL` is considered mask-able in caldav_scheddefault() and propfind_calurl(), as the looup is done using `annotatemore_lookupmask()`.  jmap_calendar.c:set_scheddefault() uses annotate_state_writemask(). But the property is not supposed to be masked (be per user for one and the same mailbox owner).

PR #5065, opened 2024-10-03
http_caldav.c: do not require tzdist module for calendar-timezone-id

rsto:


ksmurchison (commented):
This works, but couldn't we just always set ALLOW_CAL_NOTZ?

rsto:
I thought that, too. We could but it would enable a couple of other features, too.

ksmurchison (commented):
Yeah, you're right.  This is probably safer for now.  However, I also think we want this change in propfind_timezone() as well.

rsto:
Agreed. Applied change and restarted CI.

ksmurchison (approved):

rsto:
Could not reproduce CI failures on local build. Most likely a VM glitch. Rerunning CI.

dilyanpalauzov:
proppatch_tzid() could reject acting on Scheduling Outbox and Attachments.

caldav_get_calendar_tz() looks for the timezone in the collection and in the calendar-homeset.  It is reasonable to allow proppatch_tzid() and proppatch_timezone() when req_tgt.userid is set, fctx->req_tgt->resource is not set, and irrespective of whether fctx->req_tgt->collection is NULL.

PR #5067, opened 2024-10-03
lmtp_sieve.c: only process data that consists of a single VCALENDAR

ksmurchison:


imap/lmtp_sieve.c
```diff
@@ -1473,6 +1473,16 @@ static int sieve_processcal(void *ac, void *ic, void *sc, void *mc,
         goto done;
     }
 
+    if (icalcomponent_isa(itip) != ICAL_VCALENDAR_COMPONENT) {
+        buf_setcstr(&cal->outcome, "error");
+        buf_setcstr(&cal->reason,
+                    "does NOT contain a single VCALENDAR component");
```
elliefm, line 1479: AIUI the problem in this case is that there are several VCALENDAR components but we expected exactly one, but the "does not contain a single" phrasing in the reason can be read like "there were no VCALENDAR components at all!".  That will be confusing to anyone who sees it, because there clearly are.  And the "END:VCALENDAR/BEGIN:VCALENDAR" pair of lines in the middle of the data is easy to miss too, making it's hard to see that there's two of them rather than one.

If there were _no_ VCALENDAR components in the data, would they also hit this error message, or is that case detected separately somewhere else?

I think I'd rephrase the reason string to be something like "must have exactly one VCALENDAR component" or something, so that we're not overlapping with the colloquial "not a single ..." usage.  I might phrase this slightly differently depending on whether the 0 case gets here or not, but maybe not.

elliefm (approved):
Looks good, but I suggest changing the "reason" string to prevent a confusing misread

rsto (approved):
Looks good, except for the point that elliefm mentions.

imap/lmtp_sieve.c
ksmurchison: If there are no VCALENDAR components in the data, ical_string_as_icalcomponent() will return NULL, which results in a reason string of "unable to find & parse text/calendar part".

PR #5069, opened 2024-10-04
add attribute noreturn where useful

elliefm:
This is a redo of #3737 by @dilyanpalauzov (thanks!)

Adds `__attribute__((noreturn))` to a number of choice places detected by GCC's `-Wsuggest-attribute=noreturn` option.  In particular, it adds it to the places where GCC's suggestion is correct.  Most of these are `usage()`, `shut_down()`, and `service_abort()`, but there's a few others.

The advantage of adding this attribute to functions that intentionally never return is that GCC can now tell us when functions that _should_ return _don't_.  There's a few of those, and this PR does not address them, because they're not critical problems, and they'll need more care than I'm investing right now to deal with properly.  At the time of writing, these are:

* mupdate: `mupdate_client_start()` thread body loops forever (eventually terminated when the process exits)
* fud, notifyd, timsieved: lazy (but safe) `service_main()` implementation whereby the whole process exits after serving one client connection, instead of resetting its state ready for the next one

### DIFF
--- a/imap/ctl_zoneinfo.c
+++ b/imap/ctl_zoneinfo.c
@@ -75,7 +75,7 @@ const int config_need_data = 0;
 static int verbose = 0;
 
 /* forward declarations */
-void usage(void);
+void usage(void) __attribute__((noreturn));
 void free_zoneinfo(void *data);
 void store_zoneinfo(const char *tzid, void *data, void *rock);
 void do_zonedir(const char *prefix, struct hash_table *tzentries,
--- a/imap/ctl_zoneinfo.c
+++ b/imap/ctl_zoneinfo.c
@@ -80,7 +80,7 @@ void free_zoneinfo(void *data);
 void store_zoneinfo(const char *tzid, void *data, void *rock);
 void do_zonedir(const char *prefix, struct hash_table *tzentries,
                 struct zoneinfo *info);
-void shut_down(int code);
+void shut_down(int code) __attribute__((noreturn));
 
 
 int main(int argc, char **argv)
@@ -503,7 +503,6 @@ void store_zoneinfo(const char *tzid, void *data, void *rock)
 /*
  * Cleanly shut down and exit
  */
-void shut_down(int code) __attribute__((noreturn));
 void shut_down(int code)
 {
     in_shutdown = 1;
--- a/imap/cyr_synclog.c
+++ b/imap/cyr_synclog.c
@@ -58,6 +58,7 @@
 #include "util.h"
 #include "xmalloc.h"
 
+__attribute__((noreturn))
 void usage(const char *name) {
     fprintf(stderr, "Usage: %s [-C altconfig] [-{type}] value\n", name);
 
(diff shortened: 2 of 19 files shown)


rsto (approved):

dilyanpalauzov:
I do not understand why jumping from https://github.com/cyrusimap/cyrus-imapd/pull/3737 to here made the attribute disappear on numerous places, like quota.c/usage() or sieve/script.c:stub_generic().

ksmurchison (commented):
Looks good, but I think Dilyan found a couple other places where we need this attribute.

elliefm:
> I do not understand why jumping from #3737 to here made the attribute disappear on numerous places, like quota.c/usage() or sieve/script.c:stub_generic().

GCC didn't complain about those, not sure why. I'll have a look.

elliefm:
Ahh:

> Warn about functions that might be candidates for attributes `pure`, `const`, `noreturn`, `malloc` or `returns_nonnull`.  The compiler **only warns for functions visible in other compilation units** or (in the case of `pure` and `const`) if it cannot prove that the function returns normally. 

Those ones are declared `static`, so GCC thinks the `noreturn` attribute is not useful for them.  I don't understand why, but since it's GCC's attribute, I'll trust it to know what it wants.

Off the top of my head, `usage()` functions could probably always be static anyway, so the other way we might have stopped it wanting `noreturn` on those would be to make them all static instead.  But I don't think it's worth chasing.  @ksmurchison What do you think?

dilyanpalauzov:
The noreturn attribute is part of [C23](https://en.cppreference.com/w/c/language/attributes/noreturn), C11 introduced [_Noreturn keyword](https://learn.microsoft.com/en-us/cpp/c-language/noreturn?view=msvc-170).  `noreturn` is not GCC specific.

As stated in https://github.com/cyrusimap/cyrus-imapd/pull/3737 the proposed changes were made after passing `-Wmissing-noreturn` when compiling Cyrus IMAP with clang.

dilyanpalauzov:
I asked at https://gcc.gnu.org/pipermail/gcc-patches/2024-October/664606.html how does `noreturn` work.  It turns out it does not modify the generated code of the `noreturn`-function itself.  It only optimizes how the function is called from other functions.

In case of functions in the same .c file (translation unit), GCC can detect that a function is `noreturn` and thus optimize how it is invoked.

In case the function is called from other translation units, GCC sees only the function declaration and then cannot know if the call to the function invocation can be optimized in the other translation units, unless it has `noreturn` attribute.

dilyanpalauzov:
Making as much functions as possible `static` is always a good idea.

elliefm:
Ahh, that makes sense, thanks for looking deeper into it.

I took a stab at making all the `usage()` functions static instead of noreturn, but there's even more mess there (e.g. a lot of them have type `int` instead of `void`, even though they never return).  If I start cleaning that up too, the changes cascade.  But if I only partially fix it, it becomes even less consistent than it is now.  So I think I'm gonna leave this one as it already is, just adding the noreturns.

PR #5071, opened 2024-10-04
proppatch_todb_nomask() - store for user ""

dilyanpalauzov:
After calling proppatch_todb_nomask() the property should be stored for user `""`, not for httpd_userid.  The latter might or might not be the mailbox owner.

### DIFF
--- a/imap/http_dav.c
+++ b/imap/http_dav.c
@@ -3379,8 +3379,7 @@ static int proppatch_todb_internal(xmlNodePtr prop, unsigned set,
     if (!r) r = mask ?
         annotate_state_writemask(astate, buf_cstring(&pctx->buf),
                 httpd_userid, &value) :
-        annotate_state_write(astate, buf_cstring(&pctx->buf),
-                httpd_userid, &value);
+        annotate_state_write(astate, buf_cstring(&pctx->buf), "", &value);
 
     if (!r) {
         xml_add_prop(HTTP_OK, pctx->ns[NS_DAV], &propstat[PROPSTAT_OK],


brong:
No, this isn't right.  nomask means "it's always per-user".  The masked version makes it visible to someone else; and this is used for things like alarms which should always be per-user, so the owner's alarms shouldn't leak to other viewers.

brong (changes requested):
If there's a use-case which needs a 'shared' version we could extend the function, but this doesn't do what was intended when Robert wrote this.

brong:
@rsto adding you as a reviewer as well, because this touches code you wrote fairly recently.

dilyanpalauzov:
I found this note on my system:
```diff
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -5843,7 +5843,7 @@ int propfind_calurl(const xmlChar *name, xmlNsPtr ns,
         if (annotname) {
             char *mailboxname = caldav_mboxname(httpd_userid, NULL);
             int r = annotatemore_lookupmask(mailboxname, annotname,
-                                            httpd_userid, &calbuf);
+                                            httpd_userid, &calbuf); //NOT MASKABLE
             free(mailboxname);
             if (!r && calbuf.len) {
                 buf_putc(&calbuf, '/');


```
Should the values returned by `propfind_calurl()` - CALDAV:calendar-home-set, CALDAV:schedule-inbox-URL, and CALDAV:schedule-outbox-URL depend only on the request URI, or they should in addition depend on the authenticated id (httpd_userid)?  I think they should depend only on the request URI, but currently depend also on httpd_userid.

dilyanpalauzov:
See also https://github.com/cyrusimap/cyrus-imapd/pull/5338#issuecomment-2737987382.

rsto:
@dilyanpalauzov the `proppatch_todb_nomask` function only is called in the current codebase for the `default-alarm-vevent-datetime` and `default-alarm-vevent-date` DAV properties in the `CALDAV:` namespace. There `httpd_userid` instead of `""` as the user_id string in the annotation makes it a per-user-only property.

Is this PR probably stale by now?

rsto (commented):
This comment is just here to make the Github review process mark this PR as reviewed. I sent my actual comment outside that review tool.

dilyanpalauzov:
These things need clarifications:
* does "nomask”/"mask" mean Cyrus-wide “same value for all users”/“per-user value” property, or the opposite.
* If for all users the same value is stored (so no per-user property) should this be stored for user=`""` or for user="the mailbox-owner" (or here `httpd_userid`, which is not necessary the mailbox owner, but should be the mailbox owner).

Once it is clear how these things should work in all places in Cyrus, the concrete case can be discussed.

rsto:
>     * does "nomask”/"mask" mean Cyrus-wide “same value for all users”/“per-user value” property, or the opposite.

proppatch_todb_nomask: stores the annotation for that user exclusively.

proppatch_todb_mask: stores the annotation as a shared value for all users if the authenticated user during the write is an admin. Otherwise falls back to a nomask_write.
 
How these values are read depends on if annotatemore_lookupmask or annotatemore_lookupmask is called.

>     * If for all users the same value is stored (so no per-user property) should this be stored for user=`""` or for user="the mailbox-owner" (or here `httpd_userid`, which is not necessary the mailbox owner, but should be the mailbox owner).

If you want to share a value across users then store it with `""` and either always read it for user `""` or make sure to not ever call annotatemore_lookupmask or annotatemore_lookup with a userid for that annotation.

dilyanpalauzov:
I am not asking what the function `proppatch_todb_nomask:()` does.  I am asking what means that a property is masked or not masked in Cyrus Imap.  Is colour per collection / resource a masked or not-masked property?  It is a per-user property.

dilyanpalauzov:
.. or rather maskable/not-mashable property.  If it is maskable, it should use only functions for read/write which work with maskable properties.  If it is not-maskable, it should use only functions which are for not-maskable properties.

dilyanpalauzov:
I will write in June documentation on how this is supposed to work, then we can agree on it.

PR #5073, opened 2024-10-08
lmtpengine: don't rely on error_message(0) for success string

elliefm:
On some platforms this produces strings like "Unknown error: 0", "Undefined error: 0", etc

Based on a patch from @ksmurchison and discussion in #3035 and #3826


rsto (approved):

cwalther:
Looks good to me based on the discussion in #3826, but I haven’t tried it.

ksmurchison (approved):

PR #5075, opened 2024-10-09
CalDAV HTML Admin page: show/edit scheduling-enabled

dilyanpalauzov:


dilyanpalauzov:
While `_scheduling_enabled()` calls

> annotatemore_lookupmask_mbox(mailbox, DAV_ANNOT_NS "<" XML_NS_CYRUS ">scheduling-enabled", "", &buf);`

`list_cal_cb()` executes

> annotatemore_lookup_mbe(mbentry, DAV_ANNOT_NS "<" XML_NS_CYRUS ">scheduling-enabled", "", &buf);`

The difference is, that the former also invokes
```c
/* and because of Bron's use of NULL rather than "" at FastMail... */
if (value->len == 0)
    r = _annotate_lookup(mboxname, mboxid, uid, entry, NULL, value);
```
I do not understand how this NULL is supposed to work at FastMail.

brong (approved):
This looks fine.  I'll answer the question inline separately

brong:
@dilyanpalauzov you'll need to rebase out the changes/next item because it was removed when the last release was done!

Regarding the NULL - it was supposed to work the same as "", but the storage format is slightly different so we needed to look up both.  I'm not sure if the data is still in that state, but I wouldn't necessarily change it here - we can use those lookup functions interchangeably.

PR #5077, opened 2024-10-09
rfc-support.rst: clarify for Lemonade that NOTIFY is not missing

dilyanpalauzov:


dilyanpalauzov:
Good for backporting to 3.10.

ksmurchison (approved):

PR #5078, opened 2024-10-10
Add compatibility layer to support libical icalrecur by ref and by value

ksmurchison:
Also handles the by_* arrays being renamed as part of the icalrecur-by-ref work

### DIFF
--- a/imap/ical_support.c
+++ b/imap/ical_support.c
@@ -407,10 +407,23 @@ static int icalrecur_compare(struct icalrecurrencetype *a,
     cmp = a->week_start - b->week_start;
     if (cmp) return cmp;
 
-    cmp = memcmp(a->by_second, b->by_second,
-                 sizeof(a->by_second) / sizeof(a->by_second[0]));
+    cmp = strcasecmpsafe(a->rscale, b->rscale);
     if (cmp) return cmp;
 
+    cmp = a->skip - b->skip;
+    if (cmp) return cmp;
+
+#ifdef HAVE_RECUR_BY_REF
+    short i;
+    for (i = 0; i < ICAL_BY_NUM_PARTS; i++) {
+        cmp = a->by[i].size - b->by[i].size;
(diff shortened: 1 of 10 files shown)


imap/ical_support.c
```diff
@@ -3260,4 +3273,131 @@ EXPORTED void icalcomponent_set_jmapid(icalcomponent *comp, const char *id)
...
+    }
+}
+
+EXPORTED short icalrecur_byrule_size(icalrecurrencetype_t *rt,
+                                     icalrecurrencetype_byrule rule)
+{
+    switch (rule) {
+    case ICAL_BY_MONTH:
+        if (rt->by_month[0] == ICAL_RECURRENCE_ARRAY_MAX) return 0;
+        else return ICAL_BY_MONTH_SIZE;
```
rsto, line 3372: Shouldn't this check for the position of the first occurrence of ICAL_RECURRENCE_ARRAY_MAX in the by_xxx array and return that, falling back to ICAL_BY_XXX_SIZE only if none could be found in the first ICAL_BY_XXX_SIZE positions?

This is what's [documented for version 3.0](https://github.com/libical/libical/blob/e8b159db8dcf25371689e4cabb7996aa86f600a9/src/libical/icalrecur.h#L168-L170): 
```
     * Each of the lists is terminated with the value
     * ICAL_RECURRENCE_ARRAY_MAX unless the list is full.
     */
```

Instead, the version 4.0 implementation does not make such guarantee.

rsto (commented):
This looks mostly correct to me, but I have doubts about icalrecur_byrule_size

imap/ical_support.c
ksmurchison: So, good point.  The existing code that uses this call already checks for ICAL_RECURRENCE_ARRAY_MAX, but I guess I should just move that here and then we won't be checking for that value for the new dynamic arrays.

rsto (approved):

Issue #5072, opened 2024-10-06
Use after free by http_cgi:meth_get()


dilyanpalauzov:
master/service.c calls in a loop httpd.c:service_main().

On the first iteration it calls global.c:get_clienthost() -> libc/(inet or nss)/getnameinfo.c:getnameinfo().  This could be resolved by systemd’s:nss-resolv.c:query_flags(), which calls getenv() for SYSTEMD_NSS_RESOLVE_VALIDATE, SYSTEMD_NSS_RESOLVE_SYNTHESIZE, SYSTEMD_NSS_RESOLVE_CACHE, SYSTEMD_NSS_RESOLVE_ZONE, SYSTEMD_NSS_RESOLVE_TRUST_ANCHOR and SYSTEMD_NSS_RESOLVE_NETWORK.  The environment variable resolution is done by looking in `char** environ`, exported by glibc as weak object.  This [must be](https://man7.org/linux/man-pages/man7/environ.7.html) an array of pointers to strings, the last pointer in this array must have the value NULL. 

Then http_cgi.c:meth_get() does:
```c
    extern char **environ;
    env = strarray_splitm(NULL, buf_release(&txn->buf), "\t", 0);
    strarray_append(env, NULL);
    environ = env->data;
    strarray_free(env); // ⇒ free(env), env->data ⇔ environ is now invalid
```
On the second iteration of service.c → service_main() again calls getenv("SYSTEMD_NSS_RESOLVE_VALIDATE"), but this results reading invalid memory, as `environ` points nowhere.

PR #5079, opened 2024-10-10
http_cgi.c:service_main() avoid use after free(environ)

dilyanpalauzov:
environ is freed below and then in the next iteration of `service_main()` → `get_clienthost()` → `getnameinfo()` can call systemd’s nss-resolve:`getenv()`.

Fixes https://github.com/cyrusimap/cyrus-imapd/issues/5072.

### DIFF
--- a/imap/http_cgi.c
+++ b/imap/http_cgi.c
@@ -248,13 +248,16 @@ static int meth_get(struct transaction_t *txn,
 
     env = strarray_splitm(NULL, buf_release(&txn->buf), "\t", 0);
     strarray_append(env, NULL);
+    char **const environ_orig = environ;
     environ = env->data;
 
     /* Run script */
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
I httpd.c:service_main()
```
   /* Find out name of client host */
    http_conn.clienthost = get_clienthost(0, &httpd_localip, &httpd_remoteip);
```
Can httpd_remoteip and httpd_localip change on subsequent iterations of service_main() in the same process?  if not, then an alternative approach would be so skip calling get_clienthost, when httpd_localip and httpd_remoteip are set.

ksmurchison (approved):

dilyanpalauzov:
As this is security related, the current change should go all the way back to where http_cgi.c was born: 3.2.

PR #5083, opened 2024-10-11
jmap_ical.c: protect against a NULL geo: uri

ksmurchison:


elliefm:
I don't know these API's well and haven't seen the backtrace, but on a code read it looks like the crash happens inside `unescape_ical_text()` when `s` is NULL, and the changes here stop us getting there.

Do we want to make `unescape_ical_text()` safe too while we're in here?  Perhaps something like:

```
diff --git a/imap/jmap_ical.c b/imap/jmap_ical.c
index beafa3745..2bae2ab86 100644
--- a/imap/jmap_ical.c
+++ b/imap/jmap_ical.c
@@ -637,7 +637,7 @@ get_icalxparam_value(icalproperty *prop, const char *name)
 
 static void unescape_ical_text(struct buf *buf, const char *s)
 {
-    for (; *s; s++) {
+    for (; s && *s; s++) {
         if (*s == '\\') {
             switch (*++s) {
                 case 'n':
```

This would make it do nothing when `s` is NULL, rather than crashing.  Not sure what the consequences of that would be though, maybe the crash is preferable so we find out if it ever happens again.  Looks like there's 9 locations where this is called, some are protected by like `if (s) unescape_ical_text(buf, s)` checks, but not all are.

rsto (approved):
LGMT. I agree with @elliefm that unescape_ical_text could benefit of NULL checks, too.

PR #5086, opened 2024-10-14
make mbexamine handle (non-existent) intermediate mailboxes

### LINKED ISSUE #5098: mbexamine stops when called with a (non-existent) intermediate mailbox
If you have the mailboxes

user.test
user.test.Foo.Bar

where "user.test.Foo" is an intermediate mailbox which does not exist. Then (with separator '/') a call of ```mbexamine -q user/test/*``` stops at ```Examining user/test/Foo...```.

Instead I would expect: ```Examining user/test/Foo/Bar...  Mailbox has CORRECT total quota usage```.

This is tested with Cyrus 3.4.
Since Cyrus 3.6 you cannot create new intermediate mailboxes anymore. But there might still be intermediate mailboxes created before an upgrade. So ```mbexamine``` should be able to handle them.

benmuellnix:
At the moment mbexamine stops when called with a (non-existent) intermediate mailbox.
This patch is especially useful if there are intermediate folders and mbexamine is called with wildcards.

fixes #5098

### DIFF
--- a/imap/mbexamine.c
+++ b/imap/mbexamine.c
@@ -217,9 +217,10 @@ static int do_examine(struct findall_data *data, void *rock)
     int j;
     int *ok_count = (int *) rock;
 
-    /* don't want partial matches */
+    /* don't want partial matches or intermediate mailboxes */
     if (!data) return 0;
     if (!data->is_exactmatch) return 0;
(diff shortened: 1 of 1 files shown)


benmuellnix:
fixes #5098

ksmurchison (approved):

PR #5088, opened 2024-10-14
httpd.c: cleanup funky version detection/reporting

ksmurchison:


imap/httpd.c
```diff
@@ -1383,6 +1383,11 @@ static int reset_saslconn(sasl_conn_t **conn)
 }
 
 
+static const char *http_versions[] = {
+    "HTTP/1.0", HTTP_VERSION, HTTP2_VERSION
+};
+
```
elliefm, line 1389: If you add something like this here...

```suggestion
static const char *http_versions[] = {
    "HTTP/1.0", HTTP_VERSION, HTTP2_VERSION
};
static const size_t n_http_versions = sizeof(http_versions) / sizeof(http_versions[0]);
```

imap/httpd.c
```diff
@@ -2595,12 +2597,7 @@ EXPORTED const char *http_statusline(unsigned ver, long code)
 {
     static struct buf statline = BUF_INITIALIZER;
 
-    if (ver == VER_2) buf_setcstr(&statline, HTTP2_VERSION);
-    else {
-        buf_setmap(&statline, HTTP_VERSION, HTTP_VERSION_LEN-1);
-        buf_putc(&statline, ver + '0');
-    }
-
+    buf_setcstr(&statline, http_versions[ver]);
```
elliefm, line 2600: ... then you could add an assertion here for safety:

```suggestion
    assert(ver < n_http_versions);
    buf_setcstr(&statline, http_versions[ver]);
```

elliefm (approved):
Looks good, one suggestion if you like

imap/httpd.c
rsto: In addition, you might want to change
`static const char *http_versions[]`
to
`static const char * const http_versions[]`

rsto (approved):

imap/httpd.c
ksmurchison: done

imap/httpd.c
ksmurchison: done

PR #5089, opened 2024-10-14
imapparse.c: don't allow empty Esearch source-opts

ksmurchison:
This is prevents a crash when in unselected state and is also not allowed per the ABNF:

`esearch-source-opts =  "IN" SP "(" source-mbox [SP
           "(" scope-options ")"] ")"`


rsto (commented):
Could you provide a test case? It looks like a single ESEARCH call would be enough to assert that this gets rejected.

rsto (approved):
LGTM - assuming that CI passes.

PR #5090, opened 2024-10-16
slowio.testc: don't expect lightning fast response under valgrind

elliefm:
One of the slowio cunit tests fails under valgrind because it expects a thing to happen instantly but valgrind slows it down.  This makes it not expect to be fast when running under valgrind.

### DIFF
--- a/cunit/slowio.testc
+++ b/cunit/slowio.testc
@@ -4,6 +4,16 @@
 
 #include <inttypes.h>
 
+#if HAVE_VALGRIND_VALGRIND_H
+#include <valgrind/valgrind.h>
+#else
+/* no valgrind header? assume we're NOT running on valgrind. might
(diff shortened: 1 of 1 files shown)


rsto (commented):
Looks good to me, but CI failed with some likely unrelated issue.

ksmurchison (approved):
LGTM.  Not sure why CI doesn' t have/find RUNNING_ON_VALGRIND

elliefm:
I guess it doesn't have valgrind.h, so it doesn't include the header that would've defined the macro.  We use RUNNING_ON_VALGRIND already elsewhere, so I wonder what I've done wrong here.  The trick might be to define RUNNING_ON_VALGRIND myself if the header wasn't found, but I'll see what we already do in the other place we use this.

ksmurchison (approved):

PR #5091, opened 2024-10-16
imap/sync_support: make some functions static

dilyanpalauzov:
Remove unused functions sync_send_restore() and sync_msgid_remove().


brong (approved):
This seems fine.  I suspect we MIGHT need to make some of these functions public again later, they're part of a whole API - but I'm happy to merge this and re-make them public when we need them.

rjbs:
This was merged as part of a busted batch of merges.  Master was immediately rewound.  I will be re-applying it differently shortly.

rjbs:
Done.

PR #5092, opened 2024-10-16
master: unlink ready file on shutdown

brong:
This means we can wrap a "is shutdown finished" check on the ready file as well


elliefm (commented):
One suggestion inline

Also, might be nice to update Cassandane to make sure the ready file is gone after shutdown.  Perhaps add a check to `Instance::stop()` along with the existing valgrind/cores/syslog checks?

brong:
Sending for another review.  I've added Cassandane tests, used xunlink, and ALSO made it die if the file can't be created.

elliefm (approved):
Great, looks good

PR #5094, opened 2024-10-16
Only allow PROPFIND calendar-timezone[-id] on calendar[-home-set]

ksmurchison:
Fixes a crasher when the PROPFIND url is /dav/calendars


rsto (approved):

dilyanpalauzov:
This should be backported, as it fixes a crash.

PR #5099, opened 2024-10-22
http_caldav: only lookup annotation on real mailboxes

rsto:
This fixes a crash for OPTIONS /dav/calendars and adds tests that should've been part of the original PR

### DIFF
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -8220,10 +8220,10 @@ static int meth_options_cal(struct transaction_t *txn, void *params)
                    txn->req_tgt.path);
         strarray_appendm(&txn->resp_body.links, buf_release(&link));
     }
-    if (txn->req_tgt.allow & ALLOW_CAL_SCHED) {
+    if ((txn->req_tgt.allow & ALLOW_CAL_SCHED) && txn->req_tgt.mbentry) {
         struct buf temp = BUF_INITIALIZER;
         if (!annotatemore_lookup_mbe(txn->req_tgt.mbentry, DAV_ANNOT_NS "<" XML_NS_CYRUS ">scheduling-enabled", "", &temp)
-            && (!strcasecmp(buf_cstring(&temp), "F") || !strcasecmp(temp.s, "no")))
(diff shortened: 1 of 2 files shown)


ksmurchison (changes requested):
I was working on something similar yesterday but yours is farther along than mine. Should we be checking for calendar-auto-schedule rather than calendar-access?  And should we add a test case where the annotation has been set to false on a calendar, just for completeness?

rsto:
> I was working on something similar yesterday but yours is farther along than mine. Should we be checking for calendar-auto-schedule rather than calendar-access? And should we add a test case where the annotation has been set to false on a calendar, just for completeness?

this all sounds good, but I won't find time to do that this week anymore. we can either merge this PR as a basis for further improvement, or would you take over the branch and add the additional tests on top?

ksmurchison:
I can take it over

imap/http_caldav.c
```diff
@@ -8220,10 +8220,10 @@ static int meth_options_cal(struct transaction_t *txn, void *params)
                    txn->req_tgt.path);
         strarray_appendm(&txn->resp_body.links, buf_release(&link));
     }
-    if (txn->req_tgt.allow & ALLOW_CAL_SCHED) {
+    if ((txn->req_tgt.allow & ALLOW_CAL_SCHED) && txn->req_tgt.mbentry) {
         struct buf temp = BUF_INITIALIZER;
         if (!annotatemore_lookup_mbe(txn->req_tgt.mbentry, DAV_ANNOT_NS "<" XML_NS_CYRUS ">scheduling-enabled", "", &temp)
-            && (!strcasecmp(buf_cstring(&temp), "F") || !strcasecmp(temp.s, "no")))
+            && (!strcasecmp(buf_cstring(&temp), "F") || !strcasecmp(buf_cstring(&temp), "no")))
```
dilyanpalauzov, line 8226: This is not needed.  The first call of buf_cstring(&temp) has to ensure that temp.s is `\0`-terminated.  The second call, added here, will allocate one more byte, if not yet allocated, for a `\0` termination, irrespective on whether the temp.s already ends with `\0`.

ksmurchison (approved):

PR #5100, opened 2024-10-23
cassandane: clean up properly after _fork_command exec failure

elliefm:
This fix prevents a huge mess from occurring if Cassandane tries to `run_command()` something that isn't executable for whatever reason.  Also adds a new test, Simple.fork_noexec, which invokes the failure condition and verifies as best it can that Cassandane hasn't chewed itself to bits.

We saw this from the new LibCyrus tests when Cassandane's rootdir is on a noexec filesystem, which means the compiled libcyrus examples can't be executed.  I'll also try to update LibCyrus to skip its tests in that situation, but I'll do that as a separate PR.  This one just fixes the infrastructure bug.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Simple.pm
+++ b/cassandane/Cassandane/Cyrus/Simple.pm
@@ -40,6 +40,7 @@
 package Cassandane::Cyrus::Simple;
 use strict;
 use warnings;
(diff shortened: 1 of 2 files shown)


rsto (approved):

wolfsage:
Awesome, thanks

PR #5101, opened 2024-10-23
LibCyrus: don't run compiled examples if not executable

elliefm:
The LibCyrus tests compile and then run some example code, but attempting to run the compiled code will fail when Cassandane's rootdir is on a filesystem that is mounted noexec.

This PR makes these tests skip trying to run the compiled code if it looks like it won't be executable.  Detecting the noexec mount option requires the program `findmnt` -- if it's not installed, the noexec mount option won't be detected, and the tests will expect to be able to execute (and then fail if they can't).

Tested by mounting a new tmpfs partition with the `noexec` option, and then configuring cassandane.ini use a directory on that partition for its `rootdir`


rsto (approved):
I ran this on a test instance and can confirm it fixes the LibCyrus failures.

PR #5102, opened 2024-10-23
http_h2.c: don't copy response body data

ksmurchison:
At one point, there WAS a reason why I was copying the data, but I BELIEVE that reason became obsolete when I added a call to http2_output() inside resp_body_chunk().

HTTP/2 still works for me locally.  I don't think we allow direct client access to Cyrus @ Fastmail, so I guess we will see if anyone in the community stumbles upon a failure mode.

### DIFF
--- a/imap/http_h2.c
+++ b/imap/http_h2.c
@@ -721,9 +721,6 @@ static int resp_body_chunk(struct transaction_t *txn,
     static unsigned char md5[MD5_DIGEST_LENGTH];
     struct http2_context *ctx = (struct http2_context *) txn->conn->sess_ctx;
     struct http2_stream *strm = (struct http2_stream *) txn->strm_ctx;
-    uint8_t flags = NGHTTP2_FLAG_END_STREAM;
-    nghttp2_data_provider prd;
-    int r;
 
     syslog(LOG_DEBUG, "http2_resp_data_chunk(datalen=%u, last=%d)",
            datalen, last_chunk);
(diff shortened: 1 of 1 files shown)


imap/http_h2.c
```diff
@@ -749,15 +746,13 @@ static int resp_body_chunk(struct transaction_t *txn,
...
     */
-    struct protstream *s = prot_readmap(xmemdup(data, datalen), datalen);
-    s->buf = s->ptr;  /* So prot_free() will free the copy of the data */
-
-    prd.source.ptr = s;
-    prd.read_callback = data_source_read_cb;
+    nghttp2_data_provider prd = {
+        .source.ptr = prot_readmap(data, datalen),
+        data_source_read_cb
+    };
```
elliefm, line 753: Style nit only, but I think I like it better when, if any struct fields are initialised by name, then the rest that are being initialised are also specified by name.

```suggestion
    nghttp2_data_provider prd = {
        .source.ptr = prot_readmap(data, datalen),
        .read_callback = data_source_read_cb,
    };
```

Though I think the way you wrote it is actually a positional initialisation of all fields, and .source.ptr is only named because source is a union rather than a struct, and you need to specify which member you're initialising.  Another way you might initialise the outer struct positionally is using a C99 compound literal that names the union member for the source field, something like this (untested...):

```suggestion
    nghttp2_data_provider prd = {
        (nghttp2_data_source){ .ptr = prot_readmap(data, datalen) },
        data_source_read_cb,
    };
```

But we don't presently make much (any?) use of C99 compound literals in Cyrus, and I've noticed in my own projects that editor syntax highlighting can find them confusing, so I present this option mostly as a point of interest.

elliefm (commented):
One minor style suggestion, but it looks fine, with the caveat that I don't know the internals and can't usefully reason about what might be different when the buffer is no longer duplicated.  I do like that we no longer need to rummage in the protstream's internals to make sure the duplicate gets freed.

If nothing uses it in anger yet it's hard to see the impact, but if nothing uses it in anger not much can break.  Do any of our tests use HTTP/2?

imap/http_h2.c
ksmurchison: Added your first suggestion.

ksmurchison:
@elliefm  All good on this one?

ksmurchison:
@elliefm  All good on this one?

elliefm (approved):

PR #5103, opened 2024-10-24
xapian_wrap.cpp: simplifications with std::string

dilyanpalauzov:


rsto (approved):
That's better, thanks.

PR #5108, opened 2024-10-29
Move buf_ declarations to a separate, installed .h file

dilyanpalauzov:
because sieve_intefrace.h depends on `struct buf` declaration and sieve_interface.h is installed.

-- install libconfig.h, imap/message_guid.h
-- charset.h (installed): do not depend on util.h and thus on config.h


ksmurchison:
This looks fine.  Is there any reason why you didn't also split the buf function implememtations into their own buf.c source file?

dilyanpalauzov:
The reason is that this is more work.

PR #5117, opened 2024-11-05
Cassandane::Tiny: Import strict/warnings to consumers

wolfsage:
This way any tiny test gets them by default, avoiding mistakes like forgetting to declare variables, etc...

### DIFF
--- a/cassandane/tiny-tests/JMAPCalendars/calendar_set_defaultalerts_shared
+++ b/cassandane/tiny-tests/JMAPCalendars/calendar_set_defaultalerts_shared
@@ -140,7 +140,7 @@ sub test_calendar_set_defaultalerts_shared
     $self->assert_shared_defaultalerts($t, 'no alarm!', 'no alarm!');
 }
 
-sub _can_match {
+sub _can_match_b {
     my $event = shift;
     my $want = shift;
 
@@ -177,7 +177,7 @@ sub assert_alarm_notifs {
         my $found = 0;
         my @newwant;
         foreach my $data (@want) {
-            if (not $found and _can_match($event, $data)) {
+            if (not $found and _can_match_b($event, $data)) {
                 $found = 1;
             }
             else {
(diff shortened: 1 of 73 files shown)


wolfsage:
(This identified a bunch of compilation problems that need to be cleaned up; I'll work on that and include them in this MR...)

wolfsage:
This should be all set

rsto (approved):
Woof - nice!

rsto:
Waiting for @elliefm to chime in, too.

rjbs:
Wow, thanks.

cassandane/tiny-tests/JMAPCalendars/calendar_set_defaultalerts_shared
```diff
@@ -134,10 +137,10 @@ sub test_calendar_set_defaultalerts_shared
         }, 'R1'],
     ]);
 
-    $self->assert_shared_defaultalerts($t);
+    $self->assert_shared_defaultalerts($t, 'no alarm!', 'no alarm!');
 }
 
-sub _can_match {
+sub _can_match_b {
```
elliefm, line 143: Are these two functions identical?  In that case I would've just moved it up to the parent.  The fact that you gave them distinct names instead makes me think maybe they were distinct functions sharing a name, but the bodies don't show in the diff and I didn't look any deeper to find out

elliefm (approved):
Nice

ksmurchison (approved):

rjbs (approved):

cassandane/tiny-tests/JMAPCalendars/calendar_set_defaultalerts_shared
rjbs: They are identical and also appear in cassandane/Cassandane/Cyrus/CaldavAlarm.pm

wolfsage:
Hey thanks.

PR #5119, opened 2024-11-06
xapian_wrap.cpp: fix language code checks for "zr-hant" and "sr-me"

### LINKED ISSUE #5118: xapian_wrap.cpp:detect_language() compares lower case string to mixed case string - this always fails
xapian_wrap.cpp:detect_language() first converts the value of the `code` variable to lower-case, then compares that value to `zh-Hant` and `sr-ME`.  This case-sensitive comparison always fails.

https://github.com/cyrusimap/cyrus-imapd/blob/58ffda28d50aea81fe53a52a766b3053d7fa6a05/imap/xapian_wrap.cpp#L398-L410

rsto:
These language code never were detected correctly, because they erroneously were compared lowercase against mixed-case strings.

Thanks to @dilyanpalauzov

Fixes #5118

### DIFF
--- a/imap/xapian_wrap.cpp
+++ b/imap/xapian_wrap.cpp
@@ -399,13 +399,13 @@ static std::string detect_language(const struct buf *part)
         std::string code(CLD2::LanguageCode(lang));
         std::transform(code.begin(), code.end(), code.begin(), ::tolower);
         // Map CLD2 special codes to ISO 639.
-        if (!code.compare("zh-Hant")) {
+        if (code == "zh-hant") {
             code = "zh";
         }
-        else if (!code.compare("sr-ME" )) {
+        else if (code == "sr-me" ) {
             code = "sr"; // not a political statement!
         }
-        else if (!code.compare("xxx")) {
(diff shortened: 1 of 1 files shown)


imap/xapian_wrap.cpp
```diff
@@ -399,10 +399,10 @@ static std::string detect_language(const struct buf *part)
         std::string code(CLD2::LanguageCode(lang));
         std::transform(code.begin(), code.end(), code.begin(), ::tolower);
         // Map CLD2 special codes to ISO 639.
-        if (!code.compare("zh-Hant")) {
+        if (!code.compare("zh-hant")) {
```
dilyanpalauzov, line 402: I suggest `if (code == "zh-hant") {`.

rsto: ack

ksmurchison (approved):

dilyanpalauzov:
This can be easily backported.

elliefm:
Don't know how far back this can go, labelling them all and I'll find out when I get there.

PR #5120, opened 2024-11-07
imapd.c: Implement IMAP UIDBATCHES

ksmurchison:
This PR will track the progress of [draft-ietf-mailmaint-imap-uidbatches](https://datatracker.ietf.org/doc/draft-ietf-mailmaint-imap-uidbatches/)

I sent the output of the Cass test to Daniel and he says it looks correct.  I'd like to get this on a FM server so he can test against it.

### DIFF
--- a/cassandane/Cassandane/Cyrus/UIDbatches.pm
+++ b/cassandane/Cassandane/Cyrus/UIDbatches.pm
@@ -1,0 +1,198 @@
+#!/usr/bin/perl
+#
+#  Copyright (c) 2011-2017 FastMail Pty Ltd. All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions
+#  are met:
+#
+#  1. Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#
+#  2. Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in
+#     the documentation and/or other materials provided with the
+#     distribution.
+#
+#  3. The name "Fastmail Pty Ltd" must not be used to
+#     endorse or promote products derived from this software without
+#     prior written permission. For permission or any legal
+#     details, please contact
+#      FastMail Pty Ltd
+#      PO Box 234
+#      Collins St West 8007
+#      Victoria
+#      Australia
+#
+#  4. Redistributions of any form whatsoever must retain the following
+#     acknowledgment:
+#     "This product includes software developed by Fastmail Pty. Ltd."
+#
+#  FASTMAIL PTY LTD DISCLAIMS ALL WARRANTIES WITH REGARD TO
+#  THIS SOFTWARE, INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY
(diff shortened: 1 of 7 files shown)


imap/imapd.c
```diff
@@ -2580,6 +2583,26 @@ static void cmdloop(void)
...
+                if (c <= EOF) goto missingargs;
+                if (c == ' ' ) {
+                    c = getuint32(imapd_in, &low);
+                    if (c != ':') goto badsequence;
+                    c = getuint32(imapd_in, &high);
+                    if (c <= EOF) goto badsequence;
+                }
+                if (!IS_EOL(c, imapd_in)) goto extraargs;
+
+                cmd_uidbatches(tag.s, size, low, high);
```
elliefm, line 2602: What if `high < low`, do we need to detect that and `goto badsequence`?  Or accept it and swap them around?  Or just produce no results?

Whatever we're doing, it'd be nice to see a test exercising it.

imap/imapd.c
```diff
@@ -15807,3 +15830,38 @@ static void cmd_getjmapaccess(char *tag)
...
+
+    prot_printf(imapd_out, "* UIDBATCHES (TAG \"%s\") UID", tag);
+    if (low <= (imapd_index->exists + size - 1) / size) {
+        int64_t msgno = imapd_index->exists - ((low - 1) * size);
+        uint32_t batch = low;
+        char sep = ' ';
+
+        prot_puts(imapd_out, " ALL");
+
+        for (; msgno > 0 && batch <= high; batch++) {
```
elliefm, line 15855: I think this loop body gets skipped if `low > high`, but we've already started an output line, does that result in us producing invalid output?

elliefm (commented):
Looks okay, but it's not clear what we expect when the sequence is provided backwards (e.g. `5:2`), and it'd be nice to see a test exercising that

rsto (approved):

imap/imapd.c
ksmurchison: Good question.  I've asked this question of the author and on the MAILMAINT list.  I will update the code when we have consensus on the correct answer

wolfsage (approved):

cassandane/Cassandane/Cyrus/UIDbatches.pm
```diff
@@ -0,0 +1,181 @@
...
+    $self->assert_null($results[0][2]);
+
+    # batch size of 10, second & third batches
+    @results = ();
+    $res = $imaptalk->_imap_cmd('UIDBATCHES', 0, \%handlers, '10', '2:3');
+    $self->assert_str_equals('ok', $imaptalk->get_last_completion_response());
+    $self->assert_str_equals('37:28,27:18', $results[0][1]);
+    $self->assert_null($results[0][2]);
+
+    # batch size of 10, fifth * sixth batches
```
elliefm, line 167: I think you meant "fifth & sixth" :)

ksmurchison: Fixed

elliefm (approved):
Yeah nice, I like that it's strict

rjbs (commented):
I'm sure everybody else gave this a good review and you don't really need me here. 😉

PR #5123, opened 2024-11-08
clean up proc entries for crashed services

elliefm:
Quoth @wolfsage:
> Say for example lmtpd repeatedly crashes for a user and then stops crashing. Now that user can't get **any** mail delivered since cyrus thinks the user now has too many active lmtpd connections!
> 
> Probably, when master records that a child has crashed, it should remove that child's pid from the cyrus proc dir…

Service processes clean up their own proc entry on clean shutdown, but they can't do that if they've crashed out.  This PR makes master detect this case and clean it up itself.


ksmurchison (approved):
LGTM

wolfsage (approved):

wolfsage:
Awesome, thanks

PR #5133, opened 2024-11-15
shut_down() the service if TLS negotiation fails

ksmurchison:
sending a cleartext response to the client on failure is nonsense


ksmurchison:
Prometheus.connection_setup_failure_imapd  fails.
@elliefm You might know how to fix this better than I do

ksmurchison:
Prometheus.connection_setup_failure_imapd  fails.
@elliefm You might know how to fix this better than I do

PR #5137, opened 2024-11-18
configure.ac: do not check for python2

dilyanpalauzov:
Python module `git` is used only from docsrc/exts/sphinxlocal/builders/gitstamp.py and sphinx-doc has removed support for Python 2 in version [2.0](https://www.sphinx-doc.org/en/master/changes/2.0.html#incompatible-changes) released April 2019.


dilyanpalauzov:
This was re-implemented at https://github.com/cyrusimap/cyrus-imapd/pull/5343 , so creating the current change was not really a good idea.

PR #5139, opened 2024-11-18
Backport proxy Uid search fix for 3.6

dhke:
- add "cmd" parameter to cmd_search() as in 3.8 and up
- get rid of goto search, call cmd_search() directly for all occurrences
- pass proper search parameter name for "Uid Search"
- pass Uid command to backend if proxying


ksmurchison:
Superseded by #5141

PR #5140, opened 2024-11-18
Backport "Uid Search" Proxy fix for 3.8

dhke:
- re-introduce "usinguid" parameter to cmd_search()
- get rid of goto search, call cmd_search() directly for all occurences
- pass proper search parameter name for "Uid Search"
- pass Uid command to backend if proxying


ksmurchison:
Superseded by #5141

PR #5138, opened 2024-11-18
Fix proxy uid search

dhke:
Fix "Uid Search" command when using imapd proxy
    - re-introduce "usinguid" parameter to cmd_search()
    - get rid of goto search, call cmd_search() directly for all occurences
    - pass proper search parameter name for "Uid Search"
    - pass Uid command to backend if proxying

Closes #5047

Issue #5047, opened 2024-09-19
UID SEARCH broken on >= 3.6


dhke:
**Affects:** cyrus-imapd > 3.4 when running in proxy-backend configuration.

The issue appeared when I tried to upgrade our test systems to cyrus-imapd 3.6 from Ubuntu 24.04. The installation almost immediately breaks access e.g. via Thunderbird, which does an `UID SEARCH RECENT` when accessing some mailbox folders. The client gets a 

`Unrecognized UID subcommand`

response from the IMAP server.

I first assumed a packaging or distribution patch problem, but that does not seem to be the case.

If using cyrus-imapd >= 3.6 in proxy mode, the proxy receives

`<76 uid SEARCH UNDELETED SUBJECT "foo"`

but sends to the backend:

`<76 Uid UNDELETED SUBJECT "foo"`

Note the absence of the actual `SEARCH` command on the backend request.

It looks like #4321 (with commit 2e6720278c1e9e0ffdb1504a1d043a1467bf7a2d) broke "UID SEARCH" when using a frontend-backend configuration with imapd acting as proxy. The commit drops `usinguid` from the parameters of `cmd_search()` and does not seem to handle that case before passing the command to the backend:

```
    if (backend_current) {
        /* remote mailbox */
        prot_printf(backend_current->out, "%s %s ", tag, cmd);
        if (!pipe_command(backend_current, 65536)) {
            pipe_including_tag(backend_current, tag, 0);
        }
        return;
    }
```

Since `cmdloop()` consumes the next word after `Uid` using `c = getword(imapd_in, &arg1);`, the actual `SEARCH` command never reaches the backend, which responds with 

`BAD Unrecognized UID subcommand`

The fix might be easy, but it needs to be backported to cyrus 3.6 and 3.8.

Update: Remove suggested patch that breaks esearch.

```
+++ b/imap/imapd.c
@@ -532,7 +532,7 @@ static int parse_fetch_args(const char *tag, const char *cmd,
 static void fetchargs_fini (struct fetchargs *fa);
 static void cmd_fetch(char *tag, char *sequence, int usinguid);
 static void cmd_store(char *tag, char *sequence, int usinguid);
-static void cmd_search(char *tag, char *cmd);
+static void cmd_search(char *tag, char *cmd, int usinguid);
 static void cmd_sort(char *tag, int usinguid);
 static void cmd_thread(char *tag, int usinguid);
[...]

PR #5141, opened 2024-11-19
imapd.c: fix proxying UID Search (issue #5047)

ksmurchison:
I think this is a much simpler fix than #5138.  If approved, we should close #5138 #5139 #5140

### DIFF
--- a/cassandane/Cassandane/Cyrus/MurderIMAP.pm
+++ b/cassandane/Cassandane/Cyrus/MurderIMAP.pm
@@ -1221,4 +1221,70 @@ sub test_replace_across_backends
     $self->check_messages(\%exp, store => $self->{frontend_store});
 }
 
+sub test_proxy_search
+{
+    my ($self) = @_;
+    my $result;
+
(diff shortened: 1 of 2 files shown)


elliefm:
The fix does look simple.  I guess this was able to be broken because we didn't have any tests for it, so we should add a test now so we don't break it again.

If I understand the problem correctly, it might be enough to add a UID search to the existing MurderIMAP.frontend_commands test.  Or perhaps add a regular search to MurderIMAP.frontend_commands, and then make a duplicate frontend_commands_uid test that does all the same things but with UID.  I think you can do this by passing `UsingUID=>1` or something like that to the `get_client()` call at the start of the test.

ksmurchison:
@elliefm Yeah, good call.  I will add tests tomorrow

elliefm (approved):
Nice test, thanks!

elliefm:
Turns out this didn't need to be backported to 3.6 after all, the bug it fixes was introduced with multisearch, which 3.6 doesn't have.  I backported the test though, which passes.

PR #5142, opened 2024-11-21
lmtp_sieve.c: better snooze failure logging

ksmurchison:


rjbs (approved):
If I read correctly:  yes.

Do we have the capability to write tests for this easily?  I think we have "assert logs contained", which should make the rest pretty easy?

ksmurchison:
> If I read correctly: yes.
> 
> Do we have the capability to write tests for this easily? I think we have "assert logs contained", which should make the rest pretty easy?

@rjbs Tests added

PR #5147, opened 2024-11-26
httpd.c: Add support for HTTP/0.9

ksmurchison:
Mostly just for completeness, and makes doing GET requests via telnet simpler when debugging.

Also, I can leverage this when starting to implement HTTP over QUIC - the ngtcp2 QUIC library has a sample/test HTTP/0.9 over QUIC client


elliefm (approved):
Looks good

PR #5148, opened 2024-11-27
add --enable-release-checks configure option

elliefm:
Adds a new option for configure: `--enable-release-checks`

With this option, configure/make will error out for these conditions:

* imapoptions entries that are marked UNRELEASED
* missing dependencies required to build the documentation
* files other than the template in changes/next

Without the option, these conditions are just warnings or ignored as usual.

I expect to use this option only in the final stages of building new public releases, to double check that everything is taken care of.  Between releases, this stuff doesn't matter and/or is the expected state of affairs, so the option isn't useful most of the time.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1587,8 +1587,18 @@ lib/chartable.c: lib/mkchartable.pl lib/charset/unifix.txt \
 
 lib/imapopts.h: lib/imapopts.c
 
+if ENABLE_RELEASE_CHECKS
+    CFG2HDR_FORBID_UNRELEASED="--forbid-unreleased"
+else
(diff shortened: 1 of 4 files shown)


wolfsage (approved):
(I did not try this build out myself)

This seems fine, though I think you'd be happier in the perl script if you used Getopt::Long or similar - could delete a lot of code

elliefm:
Ahh yep.  For whatever reason (inertia?) we don't usually use Getopt::Long in Cyrus perl, but I just checked and we also don't _not_ use it.  I'll keep it in mind for next time, but I did that command line parser rewrite like 18 months ago, and I don't feel like revisiting it now.

PR #5155, opened 2024-12-02
fix shared mailbox subscription bug when one user is prefix of other

### LINKED ISSUE #5146: Cannot subscribe shared folder if destination mailbox name is a prefix of source mailbox name
Hello,

we’ve found a bug in Cyrus 3.6 where it is impossible to share a mailbox folder from one mailbox to another when the destination mailbox name is a prefix of the source mailbox name. It’s reproducible by these steps:

1. Create two mailboxes, `martin@somedomain.com` and `martin123@somedomain.com`. It’s crucial that `martin` is a prefix of `martin123`.
2. Create a subfolder `martin123@somedomain.com/SharedFolder`.
3. Set ACL so that `martin@somedomain.com` has access to `martin123@somedomain.com/SharedFolder`. I.e. `cyradm listacl` looks like: 
```
user/martin123/SharedFolder@somedomain.com:
  martin@somedomain.com lrswipkxtecd
  martin123@somedomain.com lrswipkxtecdan
```
4. Try to subscribe `SharedFolder` from `martin@somedomain.com`. Then, although ACL is properly set, the subscription doesn’t happen and `martin@somedomain.com` doesn’t see the shared folder.

While investigating this issue we noticed that there’s a strange record in martin.sub subscriptions file: instead of the expected `Nsomedomain.com.user.martin123.SharedFolder`, there’s `NINBOX123.SharedFolder`. That is, the common prefix of the two mailboxes was replaced by string `INBOX`. If we copy .sub file from another mailbox that contains `Nsomedomain.com.user.martin123.SharedFolder` , subscription of the shared folder works fine.

After quick inspection of the source code I think the issue is caused by function `mboxlist_dbname_to_key()` in mboxlist.c. This function rewrites mbname to `"INBOX"` based on a trivial `strncmp()` comparison and ignores the possibility that mbname can be a partial prefix of a longer mailbox name. Then, when called from `mboxlist_changesub()`, `dbname` points to a different mailbox than `userid` and `mboxlist_dbname_to_key()` rewrites `dbname` to a nonexisting bogus name. 

I think that `mboxlist_dbname_to_key()` should rewrite mbname to `"INBOX"` **only if it’s followed by a separator or end of string**. However, `mboxlist_dbname_to_key()` is called from a number of other places, so I don’t want to propose a patch without deeper knowledge of the code. Finally, note that there’s similar `strncmp()`-based string replacement of `"INBOX"` to mbname in `mboxlist_dbname_from_key()`. Perhaps this function should perform prefix sanity checks as well?

We’ve found this bug in Cyrus IMAP 3.6 (3.6.1-4+deb12u3) in Debian 12. Cyrus IMAP 3.4.x and older seem to be unaffected. Master branch on github contains the same `mboxlist_dbname_to_key()` and `mboxlist_changesub()` code so I assume that all versions after 3.6 have this bug too. Note that our setup uses options `altnamespace: yes` and `mailbox_legacy_dirs: yes` (not sure if it’s relevant).

Best regards,
Martin

elliefm:
**Reviewers**: I'd suggest stepping through this one commit by commit

* Fixes #5146 by more strictly checking boundaries when converting between mboxlist dbnames and keys.
* Adds a new Cassandane suite "Shared" with tests for subscribing to shared mailboxes.  I expect we'll accumulate other shared mailbox tests here over time.
* Bumps the subscriptions database version number from 2 to 3, and makes the upgrade-on-open path fix any broken v2 entries it finds.  It fixes the bad subscriptions by string manipulation -- it does _not_ check whether those mailboxes actually exist, nor whether the subscribed user has access to them.  I don't think this is a problem, as I believe you can ordinarily be subscribed to mailboxes that don't exist, or which you cannot see, and you just won't see them, or they'll have the `\Nonexistent` flag if you `LSUB` or `LIST (SUBSCRIBED) ...` them.
* Renames the existing Cassandane "Lsub" test suite to "Subscriptions", and adds tests that create a v2 subscriptions db with good and bad records, and verify that it's correctly upgraded and fixed on first open.
* Extends the old and barely-used Cassandane::Mboxname class to better support non-admin namespace mailboxes, because I needed that functionality for these tests.  These improvements should make this class useful in many other tests too.
* Also improves Cassandane::Util::Words a little, cause I needed lists of random words for the tests.

### DIFF
--- a/cassandane/Cassandane/Mboxname.pm
+++ b/cassandane/Cassandane/Mboxname.pm
@@ -143,7 +143,7 @@ sub _external_separator_regexp
     return $self->{config}->get_bool('unixhierarchysep', 'on') ? qr/\// : qr/\./;
 }
 
-
+# XXX assumes input is always in the admin namespace
 sub from_external
 {
     my ($self, $s) = @_;
@@ -161,20 +161,68 @@ sub from_external
     die "Bad external name \"$s\""
         if !defined $userid || $prefix ne 'user';
 
+    s/\./\^/g for @comps;
     $self->_set($domain, $userid, join('.', @comps));
 }
 
 sub to_external
 {
-    my ($self) = @_;
+    my ($self, $namespace) = @_;
 
-    my @comps;
-    push(@comps, 'user', $self->{userid}) if defined $self->{userid};
-    push(@comps, split(/\./, $self->{box})) if defined $self->{box};
-    my $s = join($self->_external_separator, @comps);
-    $s .= '@' . $self->{domain} if defined $self->{domain};
+    $namespace ||= 'admin';
 
-    return ($s eq '' ? undef : $s);
+    my $altnamespace = $self->{config}->get_bool('altnamespace', 'on');
+
+    my @boxes;
+    @boxes = split/\./, $self->{box} if $self->{box};
+    if ($self->{config}->get_bool('unixhierarchysep', 'on')) {
+        map { s/\^/\./g } @boxes;
+    }
+
+    if ($namespace eq 'admin' || !$altnamespace) {
+        my @comps;
+
+        push @comps, 'user', $self->{userid} if defined $self->{userid};
+        push @comps, @boxes;
+
+        my $s = join($self->_external_separator, @comps);
+        $s .= '@' . $self->{domain} if defined $self->{domain};
+
+        return ($s eq '' ? undef : $s);
+    }
+    else {
+        # altnamespace is active, behaviour depends on whether we're
+        # the mailbox's owner or someone else
+        my @comps;
+
+        if ($namespace ne 'owner') {
+            die "Can't make external name for other user without userid"
+                if not $self->{userid};
+
+            my $userprefix = $self->{config}->get('userprefix');
+            $userprefix ||= 'Other Users';
+
+            my $userid = $self->{userid};
+            my $domain = $self->{domain};
+
+            if ($domain) {
+                # XXX assumes crossdomains=yes and crossdomains_onlyother=no!
+                if (!$self->{config}->get_bool('unixhierarchysep', 'on')) {
+                    $domain =~ s/\./^/g;
+                }
+                $userid .= '@' . $domain;
+            }
+
+            push @comps, $userprefix, $userid, @boxes;
+        }
+        else {
+            push @comps, (@boxes ? @boxes : 'INBOX');
+        }
+
+        my $s = join($self->_external_separator, @comps);
+
+        return ($s eq '' ? undef : $s);
+    }
 }
 
 sub from_internal
--- a/cassandane/Cassandane/Mboxname.pm
+++ b/cassandane/Cassandane/Mboxname.pm
@@ -132,7 +132,7 @@ sub _external_separator
     my ($self) = @_;
     die "No Config specified"
         unless defined $self->{config};
-    return $self->{config}->get_bool('unixhierarchysep', 'off') ? '/' : '.';
+    return $self->{config}->get_bool('unixhierarchysep', 'on') ? '/' : '.';
 }
 
 sub _external_separator_regexp
@@ -140,7 +140,7 @@ sub _external_separator_regexp
     my ($self) = @_;
     die "No Config specified"
         unless defined $self->{config};
-    return $self->{config}->get_bool('unixhierarchysep', 'off') ? qr/\// : qr/\./;
+    return $self->{config}->get_bool('unixhierarchysep', 'on') ? qr/\// : qr/\./;
 }
 
 
--- a/cassandane/Cassandane/Util/Words.pm
+++ b/cassandane/Cassandane/Util/Words.pm
@@ -45,7 +45,8 @@ use Exporter ();
 our @ISA = qw(Exporter);
 our @EXPORT = qw(
     &random_word
-    );
+    &random_words
+);
 
 my @words;
 my @remaining;
@@ -82,6 +83,18 @@ sub random_word
         if (!scalar @words);
     @remaining = @words unless scalar @remaining;
     return $remaining[int(rand(scalar @remaining))];
+}
+
+sub random_words
+{
+    my ($count) = @_;
+    my @random_words;
+
+    while ($count-- > 0) {
+        push @random_words, random_word();
+    }
+
+    return wantarray ? @random_words : "@random_words";
 }
 
 1;
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -378,7 +378,12 @@ static void mboxlist_dbname_to_key(const char *dbname, size_t len,
         char *inbox = mbname_dbname(mbname);
         size_t inboxlen = strlen(inbox);
 
-        if (len >= inboxlen && !strncmp(dbname, inbox, inboxlen)) {
+        if (len >= inboxlen
+            && (len == inboxlen
+                || dbname[inboxlen] == '\0'
+                || dbname[inboxlen] == DB_HIERSEP_CHAR)
+            && !strncmp(dbname, inbox, inboxlen))
+        {
             buf_appendcstr(key, "INBOX");
             dbname += inboxlen;
             len -= inboxlen;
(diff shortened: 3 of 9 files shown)


ksmurchison (approved):

cassandane/Cassandane/Mboxname.pm
```diff
@@ -132,15 +132,15 @@ sub _external_separator
     my ($self) = @_;
     die "No Config specified"
         unless defined $self->{config};
-    return $self->{config}->get_bool('unixhierarchysep', 'off') ? '/' : '.';
+    return $self->{config}->get_bool('unixhierarchysep', 'on') ? '/' : '.';
```
wolfsage, line 135: Will this matter for any tests that maybe expected this to be off I wonder?

cassandane/Cassandane/Mboxname.pm
```diff
@@ -161,20 +161,68 @@ sub from_external
     die "Bad external name \"$s\""
         if !defined $userid || $prefix ne 'user';
 
+    map { s/\./\^/g } @comps;
```
wolfsage, line 164: nit: map in void is usually not needed, instead `s/// for @comps;`

Bigger question: Why assume here instead of checking config?

imap/mboxlist.c
```diff
@@ -5699,26 +5714,139 @@ EXPORTED int mboxlist_upgrade(int *upgraded)
...
+
+    r = cyrusdb_fetch(subs, key, keylen, &data, &datalen, NULL);
+    switch (r) {
+    case CYRUSDB_NOTFOUND:
+        version = 0;
+        break;
+    case CYRUSDB_OK:
+        /* need a copy that's definitely null-terminated for strtol */
+        tmp = xstrndup(data, datalen);
+        version = strtol(tmp, &endptr, 10);
```
wolfsage, line 5737: nit: Should probably check errno for overflow/underflow?

cassandane/Cassandane/Mboxname.pm
elliefm: Those tests will have configured it to be off, so it will be set to off in the config, so they will get back the false value they have set.  This line is fixing a bug where it assumed missing meant 'off', but actually missing means 'on'.

It matters, in that tests that assumed the wrong thing and were broken will now assume the right thing and work!  (Except that no such tests existed until I started writing these, and found the bug.)

cassandane/Cassandane/Mboxname.pm
elliefm: (I do not particularly like this API fwiw, but I don't have a better one in mind yet.  I've used it enough to be annoyed by it, but not enough to know what I'd prefer.)

The next line passes these comps to `_set()`, which expects them to be in the internal format, where dots within mailbox name components are always represented by `^`, regardless of config.

I used this API both ways (i.e. with/without unixhierarchysep) while developing these improvements, and found this part wanted to be unconditional to be ergonomic.  Otherwise, when constructing an Mboxname object, you'd need to be mindful of the config, defeating the purpose of the module managing that detail for you.

Perhaps a better fix would be for `from_external` to take a namespace argument like `to_external` now does, and then die if the input format doesn't match match the format the declared namespace dictates.    But I didn't need that complexity, so I didn't add it.

This is mostly just a crude perl rewrite of the actual mangling that happens in mboxname.c.  So perhaps the even better rewrite would be exposing that so it can be used from perl.

cassandane/Cassandane/Util/Words.pm
```diff
@@ -84,4 +85,16 @@ sub random_word
...
+sub random_words
+{
+    my ($count) = @_;
+    my @random_words;
+
+    while ($count-- > 0) {
+        push @random_words, random_word();
+    }
+
+    return wantarray ? @random_words : join($", @random_words);
```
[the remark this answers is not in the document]
elliefm: Hmm.  In my mind interpolation is relatively slow, so if there's a specific way to do something, it's better to do that rather than interpolate.  But I don't know where I got that from, maybe it's rubbish anyway!

The irony is, if I'd been reading code with the array interpolated, I wouldn't have known off the top of my head how the words would be separated, because I didn't know about `$"` until I was reading the docs to write this line.  So in a way, this line is a note to my future self to remember how arrays interpolate. :)

imap/mboxlist.c
elliefm: Hmm.  I think if that happens then whatever remaining digits it couldn't parse will wind up in *endptr, and we'll hit the error check below.  But I'm rereading the man page now, and it's not clear what happens with endptr on overflow.  So yeah I'd better tighten that up

cassandane/Cassandane/Util/Words.pm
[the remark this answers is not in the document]
wolfsage: Under the hood they are exactly the same:
```
alh@lockedtomb:~$ perl -MO=Concise -e 'my $x = "@list"'
a  <@> leave[1 ref] vKP/REFC ->(end)
1     <0> enter v ->2
2     <;> nextstate(main 1 -e:1) v:{ ->3
9     <2> sassign vKS/2 ->a
7        <@> join[t5] sK/2 ->8
3           <0> pushmark s ->4
-           <1> ex-rv2sv sK/1 ->5
4              <#> gvsv[*"] s ->5
6           <1> rv2av[t4] lK/1 ->7
5              <#> gv[*list] s ->6
8        <0> padsv[$x:1,2] sRM*/LVINTRO ->9
-e syntax OK
```
```
alh@lockedtomb:~$ perl -MO=Concise -e 'my $x = join($", @list)'
a  <@> leave[1 ref] vKP/REFC ->(end)
1     <0> enter v ->2
2     <;> nextstate(main 1 -e:1) v:{ ->3
9     <2> sassign vKS/2 ->a
7        <@> join[t5] sK/2 ->8
3           <0> pushmark s ->4
-           <1> ex-rv2sv sK/1 ->5
4              <#> gvsv[*"] s ->5
6           <1> rv2av[t4] lK/1 ->7
5              <#> gv[*list] s ->6
8        <0> padsv[$x:1,2] sRM*/LVINTRO ->9
-e syntax OK
```

I had to look up `$"`, I was fairly sure what it was going to be but had to be sure heh. It's fine.

elliefm:
Even though it's just a niche bug fix, I think the sub.db version bump and autoupgrade makes this consequential to need a changes file

wolfsage (approved):

imap/mboxlist.c
```diff
@@ -394,7 +403,13 @@ static void mboxlist_dbname_to_key(const char *dbname, size_t len,
 static void mboxlist_dbname_from_key(const char *key, size_t len,
                                      const char *userid, struct buf *dbname)
 {
-    if (userid && len >= 6 && !strncmp(key+1, "INBOX", 5)) {
+    assert(key[0] == KEY_TYPE_NAME);
+
+    if (userid
+        && len >= 6
+        && (!key[6] || key[6] == DB_HIERSEP_CHAR)
```
wolfsage, line 410: Valgrind is unhappy about this. I think it's looking at key[6].

I'm guessing that cyrusdb_foreach is giving us a pointer to exactly 6 bytes, not 6 bytes plus 1 byte for null termination (which is why we have a key + keylen interface I imagine?)
```
==87065== Conditional jump or move depends on uninitialised value(s)
==87065==    at 0x4915E12: mboxlist_dbname_from_key (mboxlist.c:410)
==87065==    by 0x4917682: find_p (mboxlist.c:3584)
==87065==    by 0x4ACD55A: cyrusdb_forone (cyrusdb.c:295)
==87065==    by 0x491E600: mboxlist_do_find (mboxlist.c:4239)
==87065==    by 0x491FD58: mboxlist_findsubmulti_withp (mboxlist.c:5075)
==87065==    by 0x491FDD9: mboxlist_findsubmulti (mboxlist.c:5033)
==87065==    by 0x123C14: list_data (imapd.c:13865)
==87065==    by 0x123DBE: cmd_list (imapd.c:8545)
==87065==    by 0x13C769: cmdloop (imapd.c:2149)
==87065==    by 0x13F3E5: service_main (imapd.c:1264)
==87065==    by 0x153AA1: main (service.c:647)
```

wolfsage (changes requested):

imap/mboxlist.c
elliefm: Oh yeah right, I see the issue.  I'm expecting `[6]` to either be a `'\0'` or a separator, and that's fine when `len > 6`, but my condition on the line above is `len >= 6` -- so there might not be a `[6]`.  Thanks!

elliefm: There's a couple of spots with this shape of thing, I'll have a close look at the other one too while I'm in there...

elliefm: And yes, you're right about the cyrusdb key/keylen, data/datalen interfaces using arbitrary memory buffers, not C strings.

elliefm: I've fixed both `mbname_dbname_to_key()` and `mbname_dbname_from_key()` to not read `[n]` when `len==n`, and the Shared and Subscriptions tests now pass cleanly for me with `--valgrind`

wolfsage (approved):

PR #5156, opened 2024-12-02
tls: Use auto DH selection with OpenSSL 1.1.0 and later

notroj:
The current code only uses auto DH parameter selection with OpenSSL 3.0+

`SSL_CTX_set_dh_auto` was introduced in OpenSSL 1.1.0 in this commit https://github.com/openssl/openssl/commit/09599b52d4e295c380512ba39958a11994d63401

### DIFF
--- a/imap/tls.c
+++ b/imap/tls.c
@@ -150,7 +150,8 @@ static int tls_serverengine = 0; /* server engine initialized? */
 static int tls_clientengine = 0; /* client engine initialized? */
 static int do_dump = 0;         /* actively dumping protocol? */
 
-#if (OPENSSL_VERSION_NUMBER >= 0x0090800fL) && (OPENSSL_VERSION_NUMBER < 0x30000000L)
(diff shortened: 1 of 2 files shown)


dilyanpalauzov:
This needs adjustment:

https://github.com/cyrusimap/cyrus-imapd/blob/fdf228cae14803e9fc38e95dad905c9c23ace920/lib/imapoptions#L3142

notroj:
Thanks @dilyanpalauzov - updated

notroj:
One other note: the approach taken in the existing code is to ignore any manually configured DH parameters for OpenSSL versions which support `SSL_CTX_set_dh_auto`.

It would be possible to change the logic to:
a) use manually configured params if present, or else
b) fall back to the automatic logic otherwise.
but it'd be more complex change.

notroj:
Rebased to hopefully address the compiler warnings which were triggering CI failures.

notroj:
Apparently too difficult for me to track all the places using that `#define` combination. I've adjusted it but if there is a preferred coding style here I can change that. Also would welcome any feedback on comment above before iterating further. https://github.com/cyrusimap/cyrus-imapd/pull/5156#issuecomment-2511732873

PR #5157, opened 2024-12-05
Twom database format and first implementation

brong:
Throwing this open for wide reviews now!  It's been rewritten a BUNCH of times, and optimised a fair bit - it's got a simpler internal structure and basically everything aligned with the right datastructures I think.  It does MVCC reads nicely, can keep its transaction in foreach unless the callback writes, in which case it STILL does the right thing and re-locks and re-finds itself.

On disk format of records is:
image


lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2634 @@
...
+    /* no need to run recovery if we're consistent */
+    if (db_is_clean(db))
+        return 0;
+
+    r = recovery1(db, &count);
+    if (r) {
+        xsyslog(LOG_ERR, "DBERROR: recovery1 failed",
+                         "filename=<%s>",
+                         FNAME(db));
+        if (r) return r;
```
dilyanpalauzov, line 2542: Here `r` is never 0.

rjbs:
I've removed include-in-fm while tests are failing.

lib/cyrusdb_twom.c
brong: Thanks I fixed that in the newer code

brong:
Bah, CI failed because `errno.h` is required.  I took it out because it's not needed in my machine and I was trying to trim back all the required headers

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2597 @@
...
+ * end can be used to see if anything in
+ * the file may have changed and needs relocation.
+ */
+struct tm_loc {
+    struct tm_file *file;
+    size_t end;  // pointers are only valid when end matches txn end
+    // location
+    size_t offset;     // current position
+    size_t backloc[MAXLEVEL+1];
+    int is_exactmatch; // key was passed in; did we match it?
```
elliefm, line 150: Might be advantageous to put `is_exactmatch` next to `offset` so they're more likely to be in the same cache line?   I get the impression from the comment that they'll typically be accessed together, but with `backloc[]` in between, they'll be 8 * 32 bytes away from each other in memory.

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2597 @@
...
+    cr.db = NULL;
+    cr.tid = NULL;
+    r = opendb(newfname, CYRUSDB_CREATE, &cr.db, &cr.tid);
+    if (r) return r;
+
+    // has it already repacked?
+    if (cr.db->openfile->header.num_records ||
+        cr.db->openfile->header.generation > 1 ||
+        !memcmp(cr.db->openfile->header.uuid, db->openfile->header.uuid, 16)) {
+        r = myabort(cr.db, cr.tid);
```
elliefm, line 2008: IMO with long conditions it's easier to see their shape and where they end at a glance if they're formatted like this:

```suggestion
    if (cr.db->openfile->header.num_records
        || cr.db->openfile->header.generation > 1
        || !memcmp(cr.db->openfile->header.uuid, db->openfile->header.uuid, 16))
    {
        r = myabort(cr.db, cr.tid);
```

elliefm (commented):
Haven't finished looking at this yet, but need to stop for lunch, so here's what I have so far.

imap/hammer_cyrusdb.c
```diff
@@ -0,0 +1,202 @@
...
+        struct txn *tid = NULL;
+        struct txn **tp;
+        char key[100];
+        char value[100];
+        int klen, vlen, i, r = 0;
+        int rop;
+
+        /* protect against silly things */
+
+        tp = (rand() % 2) ? &tid : NULL;
```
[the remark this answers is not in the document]
brong: hmm, you're right - srand is only called in services; which also means that all our commandline tools don't do it!

We should just put it in `cyrus_init()`.

lib/cyrusdb_twom.c
brong: Good point, moved them

lib/imapoptions
```diff
@@ -217,7 +217,7 @@ are listed with ``<none>''.
 /* Alternative INBOX spellings that can't be accessed in altnamespace
    otherwise go under here. */
 
-{ "annotation_db", "twoskip", STRINGLIST("skiplist", "twoskip", "zeroskip"), "3.1.6" }
+{ "annotation_db", "twoskip", STRINGLIST("skiplist", "twom", "twoskip", "zeroskip"), "3.1.6" }
```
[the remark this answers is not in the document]
brong: Fixed

lib/imapoptions
```diff
@@ -549,7 +549,7 @@ Blank lines and lines beginning with ``#'' are ignored.
 { "backup_retention", "7d", DURATION, "UNRELEASED", "UNRELEASED" }
 /* Deprecated. No longer used. */
 
-{ "backup_db", "twoskip", STRINGLIST("skiplist", "sql", "twoskip", "zeroskip"), "UNRELEASED", "UNRELEASED" }
+{ "backup_db", "twoskip", STRINGLIST("skiplist", "sql", "twom", "twoskip", "zeroskip"), "UNRELEASED", "UNRELEASED" }
```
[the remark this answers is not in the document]
brong: Reverted that one :)

lib/cyrusdb_twom.c
brong: Sure, I've reordered the tests too, looks a bit neater and speed will be similar

imap/hammer_cyrusdb.c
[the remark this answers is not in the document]
brong: Put it in cyrus_init()

imap/hammer_cyrusdb.c
```diff
@@ -0,0 +1,202 @@
...
+ * AND FITNESS, IN NO EVENT SHALL CARNEGIE MELLON UNIVERSITY BE LIABLE
+ * FOR ANY SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
+ * AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING
+ * OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ *
+ */
+/*
+ * $Id: hammer_cyrusdb.c,v 1.4 2007/09/28 02:27:46 murch Exp $
+ */
```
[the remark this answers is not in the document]
brong: Removed

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2609 @@
...
+    struct txn *read_txn;
+
+    unsigned int readonly:1;
+    unsigned int nocsum:1;
+    unsigned int nosync:1;
+    unsigned int nocompact:1;
+    uint64_t foreach_lock_release;
+    int refcount;
+    struct dbengine *next;
+};
```
dilyanpalauzov, line 179: sizeof(struct dbengine) is 352 with paddings before foreach_lock_release and before *next.  If the lines refcount and foreach_lock_release are swapped, on my system there is one padding less and sizeof(struct dbengine) is 344.

Likewise, sizeof(struct tm_file) is 144 with padding before header and next.  If refcount is moved before header, there is one padding less and sizeof(struct tm_file) is 136.

```c
struct tm_file {
    // file descriptor
    int fd;
    int has_lock;
    // mmap
    char *base;
    size_t size;
    int dirty;
    // tracking
    int refcount;
    // header info
    struct tm_header header;
    uint32_t (*csum)(const char *base, size_t len);
    int (*compar)(const char *a, size_t alen, const char *b, size_t blen);
    struct tm_file *next;
};

struct dbengine {
    /* file data */
    char *fname;

    /* tracking info */
    struct tm_loc loc;
    struct tm_file *openfile;
    struct txn *write_txn;
    struct txn *read_txn;

    unsigned int readonly:1;
    unsigned int nocsum:1;
    unsigned int nosync:1;
    unsigned int nocompact:1;
    int refcount;
    uint64_t foreach_lock_release;
    struct dbengine *next;
};
```

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2609 @@
...
+    file->size = offset;
+
+    return 0;
+}
+
+/* given an open, mapped, locked db, write the header information */
+static inline int write_header(struct dbengine *db, struct tm_header *header)
+{
+    int r = tm_ensure(db, HEADER_SIZE);
+    if (r) return r;
```
dilyanpalauzov, line 517: The current and the next function `commit_header` can easily be rewritten not to use variable `r`.

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2609 @@
...
+    }
+    else if (!strcmp(checksum_engine, "xxh32")) {
+        twom_default_checksum_engine = TWOM_CHECKSUM_XXH32;
+    }
+    else {
+        // default
+        twom_default_checksum_engine = TWOM_CHECKSUM_XXH64;
+    }
+    return 0;
+    return 0;
```
dilyanpalauzov, line 2577: Twice `return 0` in a row.

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2609 @@
...
+        libcyrus_delayed_action(drock->fname, _delayed_checkpoint,
+                                _delayed_checkpoint_free, drock);
+    }
+
+ done:
+    if (r) {
+        int r2;
+
+        /* error during commit; we must abort */
+        r2 = abort_locked(db, txn);
```
dilyanpalauzov, line 1550:
```c
int r2;
r2 = abort_locked(db, txn);
```
can be put on one line.  But since the value of `r2` is used only once, that variable can be completely removed:
```c
/* error during commit; we must abort */
if (abort_locked(db, txn)) {
           xsyslog(LOG_ERR, "DBERROR: commit AND abort failed", "filename=<%s>", db->fname);
}
return r;

```

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2596 @@
...
+    struct txn *next = this->next;
+    this->file->refcount--;
+    free(this);
+    *ptr = next;
+}
+
+static void _remove_file(struct tm_file **ptr)
+{
+    struct tm_file *this = *ptr;
+    struct tm_file *next = (*ptr)->next;
```
wolfsage, line 426: Why isn't this `this->next` like above in _remove_txn?

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2596 @@
...
+        }
+
+        // new file, create a new mapping
+        file = xzmalloc(sizeof(struct tm_file));
+        file->fd = newfd;
+        file->next = db->openfile;
+        db->openfile = file;
+    }
+
+    db->openfile->has_lock = 2;
```
wolfsage, line 1102: Constants would make this easier to understand here and elsewhere

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2596 @@
...
+
+/* end cleanup */
+
+static inline int tm_commit(struct dbengine *db, size_t len)
+{
+    assert(db->openfile);
+    if (!db->openfile->dirty) return 0;
+    assert(db->openfile->has_lock == 2);
+    db->openfile->dirty = 0;
+    return msync(db->openfile->base, len, MS_SYNC|MS_INVALIDATE);
```
wolfsage, line 474: What is tm_commit's purpose? My understanding of msync is that it exists to make sure data is flushed to disk (or at least... to the filesystem layer). I think the only think it protects us from is power loss / server crash and reboot. Is that what we're aiming for here?

What is the MS_INVALIDATE for? With MAP_SHARED, all processes using this mmap'd region should always only see the updated data, I think MS_INVALIDATE only affects things reading the file from the OS or not using MAP_SHARED?

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2596 @@
...
+}
+
+static inline int tm_ensure(struct dbengine *db, size_t offset)
+{
+    struct tm_file *file = db->openfile;
+    if (offset <= file->size) return 0;
+
+    assert(file);
```
wolfsage, line 482: On line 478 we've done `file->size` so I think this assert is too late?

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2596 @@
...
+    assert(file->has_lock == 2);
+
+    offset = tm_roundup(offset);
+
+    // make sure anything we have mapped in has committed to disk
+    if (file->dirty) {
+        if (msync(file->base, file->size, MS_ASYNC)) {
```
wolfsage: MS_ASYNC is a no-op on linux, what are we trying to do here (or is this more about portability?)

lib/cyrusdb_twom.c
wolfsage: Any time we msync with MS_SYNC we risk blocking on file IO I think. Is that okay?

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2596 @@
...
+#endif
+static inline int check_tailcsum(struct dbengine *db, struct tm_file *file, const char *ptr, size_t offset)
+{
+    if (db->nocsum) return 0;
+    size_t taillen = TAILLEN(ptr);
+    if (!taillen) return 0;
+    uint32_t csum = file->csum(KEYPTR(ptr), taillen);
+    if (csum != TAILCSUM(ptr)) {
+        xsyslog(LOG_ERR, "DBERROR: invalid tail checksum",
+                         "filename=<%s> offset=<%08llX>",
```
wolfsage, line 563: Offset is unused except for error reporting? Okay!

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2596 @@
...
+        if (next != offset) {
+            offset = next;
+            if (offset) {
+                ptr = safeptr(loc, offset);
+                if (!ptr) return CYRUSDB_IOERROR;
+                assert(LEVEL(ptr) >= level);
+
+                cmp = file->compar(KEYPTR(ptr), KEYLEN(ptr),
+                                   key, keylen);
+
```
wolfsage, line 672: I'm unclear what this function is doing so excuse my ignorance: If cmp == 0, doesn't that mean we've found an exact match and can bail early? It seems like instead we keep going...

wolfsage (commented):
Half way through, more later/tomorrow...

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2596 @@
...
+static int mycheckpoint(struct dbengine *db)
+{
+    char newfname[1024];
+    clock_t start = sclock();
+    struct copy_rock cr;
+    int r = 0;
+    struct txn *txn = NULL;
+
+    r = refreshtxn(db, 1/*shared*/, &txn);
```
wolfsage, line 1980: This is taking a shared lock? Does that mean 2 processes can checkpoint at the same time? How do they not interfere?

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2596 @@
...
+                 const char *key, size_t keylen,
+                 struct txn **tid, int force)
+{
+    return mystore(db, key, keylen, NULL, 0, tid, force);
+}
+
+static int myinit(const char *dbdir __attribute__((unused)), int flags __attribute__((unused)))
+{
+    const char *checksum_engine = libcyrus_config_getstring(CYRUSOPT_TWOM_CHECKSUM_ENGINE);
+    if (!checksum_engine) return TWOM_CHECKSUM_XXH64;
```
wolfsage, line 2562: This looks wrong. If we *don't* have a checksum_engine, we return TWOM_CHECKSUM_XXH64, which will be treated as an error code by the caller?

wolfsage:
(Will play with it a bit tomorrow, today was mostly reading)

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2596 @@
...
+        file->fd = newfd;
+        file->next = db->openfile;
+        db->openfile = file;
+    }
+
+    file->has_lock = 1;
+
+    // if the existing map it too small, replace it
+    if (file->size < (size_t)sbuf.st_size) {
+        /* map the new space (note: we map READ|WRITE even for readonly locks,
```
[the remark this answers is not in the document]
brong: This is the read_lock function.  We map READ|WRITE if we're taking a read lock but the database itself is read_write (aka: could be write locked later)

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2596 @@
...
+    if (r) return r;
+
+    /* we can't read an unclean database */
+    if (!db_is_clean(db)) {
+    badfile:
+        /* we have to be able to re-lock safely */
+        if (db->readonly) return -EIO;
+        if (forcefile) return -EIO; // bogus; you can't have an unclean old file!
+        /* if we take a write lock, that will repair it */
+        unlock(db, file);
```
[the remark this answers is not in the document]
brong: Yes, it's intentional.  It's OK because either the something else damages the file again, in which case we're gonna be repeating this whole sorry exercise, or it's going to still be fine to get a new read_lock after.

Having said that, maybe this should have `goto retry` rather than recursing.

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2596 @@
...
+
+    header->version
+        = ntohs(*((uint16_t *)(base + OFFSET_VERSION)));
+
+    header->checksum_engine
+        = ntohs(*((uint16_t *)(base + OFFSET_CHECKSUM_ENGINE)));
+
+    set_csum_engine(file, header->checksum_engine);
+
+    if (header->version > VERSION) {
```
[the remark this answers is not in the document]
brong: good point - we should check the version first

lib/cyrusdb_twom.c
brong: Yes, that's the point of this - tm_commit's purpose is "make sure this stuff is on disk".  It's part of the 3 commit safety:
1) mark header dirty; then sync
2) make changes which are all revertable by recovery1 in case of an abort or crash; then sync
3) mark header clean; then sync

So the atomic bit is just (3), and it's only a write of 96 bytes at the top of the file, so should always be within a page.


I should read more about MS_INVALIDATE and see if it's necessary.

lib/cyrusdb_twom.c
brong: Yeah, that's portability.  Just making sure that anything in the mmap is written down to disk before we unmap it.

lib/cyrusdb_twom.c
brong: hmm, yeah.  That's totally bogus.  Luckily it got removed in the rebase (though I need to remove the associated config option now too)

lib/cyrusdb_twom.c
brong: I have re-organised these somewhat in the new twom_* layout.  I'll put dirty and has_lock together as well since they only need to be a couple of bits each.

lib/cyrusdb_twom.c
brong: Yes thanks, I've fixed it

lib/cyrusdb_twom.c
brong: Yes, but that makes it hard to instrument tests (or step though with a debugger).  Variables cost nothing when the compiler optimises them out.

lib/cyrusdb_twom.c
brong: https://stackoverflow.com/questions/60547532/whats-the-exact-meaning-of-the-flag-ms-invalidate-in-msync - says it's basically a NOOP (as is MS_ASYNC), but making things posixly correct seems sensible.

Makefile.am
```diff
@@ -802,6 +803,7 @@ include_HEADERS = \
     lib/stristr.h \
     lib/times.h \
     lib/tok.h \
+	lib/twom.h \
```
dilyanpalauzov, line 806: Bad indentation.

lib/twom.c
```diff
@@ -0,0 +1,2510 @@
...
+    struct tm_file *file = db->openfile;
+    if (!file) goto newfile;
+
+    for (;;) {
+        fl.l_type = F_WRLCK;
+        fl.l_whence = SEEK_SET;
+        fl.l_start = 0;
+        fl.l_len = 0;
+        if (fcntl(db->openfile->fd, F_SETLKW, &fl) < 0) {
+            if (errno = EINTR) continue;
```
dilyanpalauzov, line 1068: This is wrong.

lib/twom.c
```diff
@@ -0,0 +1,2510 @@
...
+    struct flock fl;
+    struct tm_file *file = forcefile ? forcefile : db->openfile;
+
+    for (;;) {
+        fl.l_type = F_RDLCK;
+        fl.l_whence = SEEK_SET;
+        fl.l_start = 0;
+        fl.l_len = 0;
+        if (fcntl(file->fd, F_SETLKW, &fl) < 0) {
+            if (errno = EINTR) continue;
```
dilyanpalauzov, line 1196: This is wrong.

lib/twom.c
```diff
@@ -0,0 +1,2510 @@
...
+    if (header->dirty_size > MINREWRITE
+        && header->current_size < 4 * header->dirty_size) return 1;
+    return 0;
+}
+
+int twom_repack(struct twom_db *db, struct twom_txn **txnp)
+{
+    char newfname[1024];
+    struct copy_rock cr;
+    int r = twom_begin(db, TWOM_SHARED, txnp);
```
dilyanpalauzov, line 1976: Value stored in `r` is never read.

lib/twom.c
```diff
@@ -0,0 +1,2510 @@
...
+    r = opendb(newfname, &init, &cr.db, &cr.tid);
+    if (r) return r;
+
+    // has it already repacked?
+    if (cr.db->openfile->header.num_records ||
+        cr.db->openfile->header.generation > 1 ||
+        !memcmp(cr.db->openfile->header.uuid, db->openfile->header.uuid, 16)) {
+        r = twom_abort(cr.db, &cr.tid);
+        if (r) goto err;
+        r = twom_close(&cr.db);
```
dilyanpalauzov, line 2005: Value stored in `r` is never read.

lib/twom.c
```diff
@@ -0,0 +1,2510 @@
...
+    // we're just doing small copies, release less frequently
+    db->foreach_lock_release *= 64;
+
+    // mvcc process all the existing records
+    r = twom_foreach(db, NULL, 0, NULL, copy_cb, &cr, txnp, /*flags*/0);
+    if (r) goto err;
+
+    // replay all the remaining changes to the end of the file
+    r = myreplay(db, *txnp, replay_cb, &cr);
+
```
dilyanpalauzov, line 2031: Value stored in `r` is never read.

dilyanpalauzov (commented):
Reading the outcome of Clang Static Analyzer might reveal further potential problems.

lib/cyrusdb_twom.c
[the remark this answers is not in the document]
brong: Done

lib/cyrusdb_twom.c
brong: I've patched this now, there's a nonblocking option and the repacker uses it.  Still has to check for the race just in case of course

Makefile.am
brong: Fixed

lib/twom.c
brong: This all got changed

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,2596 @@
...
+            xsyslog(LOG_ERR, "IOERROR: stat failed",
+                             "filename=<%s>", db->fname);
+            unlock(db, NULL);
+            return -EIO;
+        }
+
+        if (sbuf.st_ino == sbuffile.st_ino) break;
+
+    newfile:
+        int newfd = open(db->fname, O_RDWR, 0644);
```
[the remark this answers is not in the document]
brong: Yeah that all ugly.  I'll just define it above.

lib/cyrusdb_twom.c
brong: I don't care about that.  Easier to debug with variables and the compiler will clean them up.  Readability trumps micro optimisation in this case, it's only called twice per commit.  I'm all for micro-optimising advance_loc and other inner loop stuff, but not this.

brong:
There's a whole new API in twom.c now; much more the API I want, though I'm going to just make non-locked versions of the APIs too

lib/twom.c
dilyanpalauzov: This all got changed.  The value stored from `myreplay()` in `r` is never read.

imap/hammer_cyrusdb.c
```diff
@@ -0,0 +1,199 @@
...
+
+#include <sys/ipc.h>
+#include <sys/msg.h>
+
+#include "acl.h"
+#include "assert.h"
+#include "auth.h"
+#include "cyrusdb.h"
+#include "glob.h"
+#include "imap_err.h"
```
dilyanpalauzov, line 67: This must be `#include imap/imap_err.h` otherwise VPATH builds fail:
```
$ make imap/hammer_cyrusdb
  CC       imap/hammer_cyrusdb.o
../imap/hammer_cyrusdb.c:67:10: fatal error: imap_err.h: No such file or directory
   67 | #include "imap_err.h"
      |          ^~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:5496: imap/hammer_cyrusdb.o] Error 1
```

imap/hammer_cyrusdb.c
```diff
@@ -0,0 +1,199 @@
...
+#include "global.h"
+#include "mailbox.h"
+#include "util.h"
+#include "xmalloc.h"
+
+const int config_need_data = 0;
+struct cyrusdb_backend *OLDDB = NULL;
```
dilyanpalauzov, line 74: Where is variable `OLDDB` used?

brong:
> @brong I removed the `include-in-fastmail` label for this PR. The Rename test suite tests are failing.

Thanks!  yeah, there were more issues but I believe it's all clean now!

imap/hammer_cyrusdb.c
```diff
@@ -0,0 +1,199 @@
...
+	    alt_config = optarg;
+	    break;
+	case 'n': /* create new */
+	    db_flags |= CYRUSDB_CREATE;
+	    break;
+	}
+    }
+	
+    if((argc - optind) < 1) {
+	fprintf(stderr, "Usage: %s [-C altconfig] <dbfile> [<dbtype>]\n", argv[0]);
```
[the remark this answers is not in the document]
brong: It's kind of rubbish - I'm going to drop it before merging.  It was just some old code for testing.

imap/hammer_cyrusdb.c
brong: removed it

brong:
OK!  Major messing around.
* I've added more comments
* Standalone twom library is API made more consistent
* It's actually passing tests (there were plenty of fun bugs)
* Repack now happens in-object a bit more rather than creating a new DB.

lib/imapoptions
```diff
@@ -3178,6 +3178,11 @@ product version in the capabilities
    of UID, "cyrus" is UIDVALIDITY.UID.  Dovecot is 8 digits of leading
    hex (lower case) each UID UIDVALIDITY. Courier is UIDVALIDITY-UID. */
 
+{ "twom_checksum_engine", "xxh64", STRING, "UNRELEASED" }
+/* The default checksum engine for new twom databases, string values
+   are currently "null", "crc32", "xxh32", "xxh64".  The default if
+   not set is xxh64. */
```
dilyanpalauzov, line 3184: “if not set” is redundant.  As the default value is anyway shown in the rendered documentation, there is no need to spell it explicitly (cf. https://github.com/cyrusimap/cyrus-imapd/pull/4818).

lib/imapoptions
```diff
@@ -2226,7 +2226,7 @@ If all partitions are over that limit, this feature is not used anymore.
 /* Unix domain socket that ptloader listens on.
    (defaults to configdirectory/ptclient/ptsock). */
 
-{ "ptscache_db", "twoskip", STRINGLIST("skiplist", "twoskip", "zeroskip"), "3.1.6" }
+{ "ptscache_db", "twoskip", STRINGLIST("skiplist", "twom", "twoskip", "zeroskip"), "UNRELEASED" }
```
dilyanpalauzov, line 2229: docsrc/imap/concepts/deployment/databases.rst says for ptscache.db
> File type can be: `twoskip`_ (default) or `skiplist`_.

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,407 @@
...
+#include "bsearch.h"
+#include "cyrusdb.h"
+#include "libcyr_cfg.h"
+#include "twom.h"
+#include "util.h"
+#include "xmalloc.h"
+
+/* type aliases */
+#define LLU long long unsigned int
+#define LU long unsigned int
```
dilyanpalauzov, line 59: `LU` is nowhere used.

lib/imapoptions
brong: Actually this whole thing isn't being used, I'll just strip it.

lib/cyrusdb_twom.c
brong: I'm pretty sure it is:
```
    printf("HEADER: v=%lu g=%llu fl=%08lX num=(%llu/%llu) sz=(%08llX/%08llX/%08llX) ml=%lu\n",
          (LU)header->version,
          (LLU)header->generation,
          (LU)header->flags,
          (LLU)header->num_records,
          (LLU)header->num_commits,
          (LLU)header->dirty_size,
          (LLU)header->current_size,
          (LLU)header->repack_size,
          (LU)header->maxlevel);
```

dilyanpalauzov: I meant `LU` is defined in libcyrus_twom.c, but is not used in libcyrus_twom.c.

brong:
Ok I've rewritten to have DELETE just be an ancestor pointer in the level0 chain (no key or value) and the most recent ADD or REPLACE always be in the full multi-level links except for the DELETE being immediately before it if present.

This added a 'deleted_offset' to struct tm_loc, and a bunch of extra logic (slightly more code overall) - but partly because I just open-coded it everywhere, it could do with some code refactoring but it's working and the structure is what I want, so I figured it's time to ask for reviews again!

I also wrote up a bunch of extra doc about it which I'll pop in the repo.

docsrc/imap/concepts/deployment/databases.rst
```diff
@@ -223,7 +223,7 @@ have that conversations ID in them, and counts of messages and flags.
 Finally there are records for each folder with the counts of conversations in
 that folder.
 
-File type can be: `skiplist`_ (default), `sql`_, or `twoskip`_.
+File type can be: `skiplist`_ (default), `sql`_, `twoskip`_, or `twom`_.
```
dilyanpalauzov, line 226: imapoptions says the default is `twom`.

docsrc/imap/concepts/deployment/databases.rst
```diff
@@ -250,7 +250,11 @@ This is either cyrus.squat in each folder, or if you're using Xapian a single
 <userid>.xapianactive file listing active databases with tier name and number.
 
 cyrus.indexed.db is used by the Xapian search engine.  Its file type
+<<<<<<< HEAD
```
dilyanpalauzov, line 253: ?

lib/twom.c
```diff
@@ -0,0 +1,3220 @@
...
+        }
+
+        offset += RECLEN(ptr);
+    }
+done:
+
+    return r;
+}
+
+int twom_db_fetch(struct twom_db *db,
```
dilyanpalauzov, line 2961: Can the public functions in this file be marked as `HIDDEN` ⇔ `__attribute__((__visibility__("hidden")))`, as they are not called from outside libcyrus.so?

docsrc/imap/concepts/deployment/databases.rst
brong: Hmm, oops - that's noise from the zeroskip rebase.  Will fix.

docsrc/imap/concepts/deployment/databases.rst
brong: hmm, I should undo any default changes I think; we should only change defaults in major versions.

brong: Yeah, fixed the default back - as bad as skiplist is as a choice for conversationsdb, it's the default for the current versions

docsrc/imap/concepts/deployment/databases.rst
```diff
@@ -61,7 +61,7 @@ The format of each record is as follows::
 
     Data: <Type Number>SP<Partition>SP<ACL (space-separated userid/rights pairs)>
 
-File type can be: `twoskip`_ (default), `flat`_, `skiplist`_, or `sql`_.
+File type can be: `twoskip`_ (default), `twom`_, `flat`_, `skiplist`_, or `sql`_.
```
dilyanpalauzov, line 64: Before the current change in this file after the default type, the remaining formats were sorted in alphabetical order.

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,412 @@
...
```
dilyanpalauzov, line 59: `LU` is `#define`’d only in cyrusdb_twom.c, but is not used by cyrusdb_twom.c.

lib/twom.c
```diff
@@ -0,0 +1,3293 @@
...
+    r = twom_txn_foreach(txn, NULL, 0, NULL, copy_cb, newtxn, /*flags*/0);
+    if (r) goto fail;
+
+    /* remember the repack size at this point, because that's everything
+     * which is in order!  The remaing replayed items might be deletes or
+     * replaces, so they still dirty up the ordering as much as new changes. */
+    newtxn->file->header.repack_size = newtxn->end;
+
+    // replay all the remaining changes to the end of the file
+    r = myreplay(txn, replay_cb, newtxn);
```
dilyanpalauzov, line 3229: The value stored here in `r` is never read.

docsrc/imap/reference/manpages/systemcommands/cyr_dbtool.rst
```diff
@@ -16,7 +16,7 @@ Synopsis
 
 .. parsed-literal::
 
-    **cyr_dbtool** [ **-C** *config-file* ] [ **-n** ] [ **-o** ] [ **-T** ]
+    **cyr_dbtool** [ **-C** *config-file* ] [ **-n** ] [ **-o** ] [ **-S** ] [ **-N** ] [ **-T** ]
```
dilyanpalauzov, line 19: Option `-R` is not documented.  Can `-o` be removed from here?

docsrc/imap/concepts/deployment/databases.rst
brong: Yes, good catch - I have fixed this for all types

lib/cyrusdb_twom.c
brong: Ahh yep, removed

lib/twom.c
brong: That was a real bug.  It's fixed with an `if (r) goto fail;`

docsrc/imap/reference/manpages/systemcommands/cyr_dbtool.rst
brong: Yes, have removed -o, documented it, and also fixed EVERYTHING to be documented and sorted case-insignificant alphabetically.

lib/twom.c
```diff
@@ -0,0 +1,3294 @@
...
+    }
+
+done:
+    unlock(db, file);
+    return r;
+}
+
+/***********************  OPEN AND CLOSE *************************************/
+
+static void errors_to_stderr(const char *msg, const char *fmt, ...)
```
dilyanpalauzov, line 1899: Adding here `__attribute__((format(printf, 2, 3)))` avoids warnings on the line  1906 `vfprintf(stderr, fmt, args);`:

> format string is not a string literal
> clang-diagnostic-format-nonliteral

lib/cyrusdb_twom.c
```diff
@@ -0,0 +1,411 @@
...
+#include "cyrusdb.h"
+#include "libcyr_cfg.h"
+#include "twom.h"
+#include "util.h"
+#include "xmalloc.h"
+
+/* type aliases */
+#define LLU long long unsigned int
+
+static void _twom_error_callback(const char *msg, const char *fmt, ...)
```
dilyanpalauzov, line 60: Adding here `__attribute__((format(printf, 2, 3)))` prevents warning on line 71 `buf_vprintf(&buf, fmt, args);`:
> format string is not a string literal
> clang-diagnostic-format-nonliteral

lib/twom.c
```diff
@@ -0,0 +1,3294 @@
...
+
+    return r;
+}
+
+// useful for when you started with an external transaction that you're
+// not finished with, just clean up the cursor
+int twom_cursor_fini(struct twom_cursor **curp)
+{
+    struct twom_cursor *cur = *curp;
+    if (cur->loc.file) {
```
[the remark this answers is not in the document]
brong: fixed with a bit more tidying up of error handling in begin as well

lib/twom.c
brong: added

lib/twom.c
```diff
@@ -0,0 +1,3294 @@
...
+    addr += 8;
+
+    // store all the backwards and forwards locations
+    size_t oldlevel = LEVEL(locptr);
+    uint8_t i;
+
+    // if it wasn't an exact match, we'll be adding the new record afterwards,
+    // so the backloc now needs to point at that record so it gets updated below
+    if (!loc->is_exactmatch)
+        for (i = 0; i < oldlevel; i++)
```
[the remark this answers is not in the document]
brong: interesting.  They're both limited to 31; but I've made all levels uint8_t throughout.

imap/cyr_dbtool.c
```diff
@@ -325,16 +337,21 @@ int main(int argc, char *argv[])
         fprintf(stderr, "\n");
         fprintf(stderr, "Options:\n");
         fprintf(stderr, "  -c     convert database to named backend if not already\n");
-        fprintf(stderr, "  -M     use \"improved_mboxlist_sort\" order\n");
+        fprintf(stderr, "  -N     minimise use of checksums (don't check on read, create nochecksum if supported)\n");
+        fprintf(stderr, "  -n     create the database if it doesn't exist\n");
+        fprintf(stderr, "  -R     open the database readonly (won't create a new DB)\n");
```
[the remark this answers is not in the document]
brong: Thanks, fixed

doc/README.twom.md
```diff
@@ -0,0 +1,189 @@
...
+
+- Three fsyncs per transaction is kinda expensive!
+- Every write causes the file to be extended, as twoskip format uses exact file size and doesn’t allocate slop.
+- A single write does amplify into 1+N (the level of the record) writes; the append of the new record, and the updates of the immediately previous records in the skiplist at each level.  Since half the records are level 1, half that level 2, etc - the average write updates three separate previous records.  In a recently repacked file, the previous records are likely to be adjacent or nearly so for the lower levels, but this becomes less so over time, so repacks are valuable for both data locality and size reduction.
+- The repack operation requires the database to be remain exclusively locked for the duration, which is slow on large files (particularly along with the above lack of slop).
+
+## The Twom format
+
+Why the name?  It’s twoskip, but with mmap and MVCC.  Two Ms.  And the “tombstone” records, rather than just causing the skiplist to be rewritten without the deleted record in it, remain in the linked lists right until a repack, with an ANCESTOR pointer back to the previous entity with that key.
```
rsto, line 34: Similar to previous comments, this mentions tombstone records without more explanation.

doc/README.twom.md
```diff
@@ -0,0 +1,189 @@
...
+
+This tombstone allows the MVCC transaction, because you can walk the current lowest level linked list, and any time you find a record with an offset in the future of the file size at the time you started the transaction, walk back up the ancestor chain (which is a single linked list backward through the file, to the original ADD record which was the first time the key was seen in this file).
```
rsto, line 36: I would only walk the ancestor chain until the file offset of the record is less than then file size of the time I started the transaction.

doc/README.twom.md
```diff
@@ -0,0 +1,189 @@
...
+
+With MVCC transactions, we can then re-implement the repack operation; instead of having to lock the file for the entire repack, we can take a read-only lock and yield it many times while walking the original records and adding them to the new file in sorted order, and then (still releasing locks as many times as we like) read the remainder of the file and replay the operations (adds, replaces and deletes) to the new file.  This doesn’t require any exclusive lock at any point, so readers can continue throughout the repack, even until the rename - and either get the old file or the new file.
```
rsto, line 38: If I read this correctly, then the repack won't result in an optimally packed file, e.g. it might include tombstones and ancestors. The determining factors seem to be the number of read-only lock yields by the repacker and the number of writes between those yields. Is there some heuristic to choose how often to yield the read-only locks when the average number of writes is known?

doc/README.twom.md
```diff
@@ -0,0 +1,189 @@
+This document describes the twom database format for key-value store within Cyrus.
+
+## Skiplists
+
+A [skiplist](https://en.wikipedia.org/wiki/Skip_list) is a data structure gives O(log N) access to data; by creating multiple linked lists.  Every node contains a pointer to the immediately next node in a sorted order, however some nodes also contain links that “skip over” some number of nodes, hence “skip”.  I won’t describe in detail how skiplists work here, as that’s described well in other places.
```
rsto, line 5: In addition to the Wikipedia, this paper gives a good overview https://arxiv.org/abs/2403.04582

doc/README.twom.md
```diff
@@ -0,0 +1,189 @@
...
+    - Has a reference counted on the file into which it points, which is required to allow relocation without having to copy the key content.
+- `struct tm_file` an mmaped file.  Has the file handle and tracks the lock on it.  Also has a parsed version of the header of the file, which is updated every time it’s locked.  (strictly this isn’t necessary, we could do it all with MMAP reads directly from the file).
+    - Also contains pointers to the checksum and comparison functions for this file.  These CANNOT BE CHANGED after file creation, but can either use the ones built into twom, or you can pass external functions at file creation time.
+    - Refcounted by txn and loc above.
+- `struct twom_cursor` a wrapper around a loc and a txn, used by public API for cursor operations.
+
+## Locking
+
+Twom uses a two-phase locking system, as described in [a stackoverflow post](https://stackoverflow.com/questions/27625597/how-to-implement-a-writer-preferring-read-write-lock-for-nix-processes).
```
rsto, line 63: Maybe more clearer e.g.: "Twom uses fcntl(2) for locking the database file, specifically a two-phase locking system..."
Maybe of interest is https://arxiv.org/html/2403.04582v3#S4 section 4.7 describing other implementations.

doc/README.twom.md
```diff
@@ -0,0 +1,189 @@
...
+
+Remaining issues:
+
+- We still use 3 fsyncs - there’s no way around this without risking a crash leaving a file in a broken state.  We could theoretically still get a clean read after a single failure by checking for end pointers during the “locate” phase and falling back to the lowest level pointers - but if we then started another write; there would be invalid pointers somewhere in the file and it would corrupt over time.  Nope.
+- The write amplification issue is still the same, twom is basically the same format as twoskip, just with the addition of ancestor pointers and richer tombstones.
+
+### Navigating the code - data structures:
```
rsto, line 45: Both "Navigating the code" sections might better go in the C file?

rsto (commented):
This is the first of multiple review rounds. In this round, I only reviewed the README with the idea that someone new to twom in Cyrus would first look at that file.

The very first sentence in the README states that it describes the database format. It does, kind of, but I think it could be more to the point.

Motivating twom by first introducing skiplist and twoskip maybe isn't all that useful in the long run. Why would anyone use skiplist or twoskip when there's twom, and how long will the two formats still be available in Cyrus?

Instead, I would prefer if the README only describes twom, but more rigorously: explain all the record types, the main fields in the header and locking, at best also including an example for the insert, delete, search and repack operations. Preferably, the README contains everything that is required to understand if the design is sound and allows to check if the implementation lives up to that design.

doc/README.twom.md
brong: Strictly, repack will be optimally repacked for everything in the file that was present at the time the repack started, so for sure that much will be optimal (and it was already so poor at that point that it calculates it will save a bunch of space) - so it'll definitely be better than not repacking, but could still be non-ideal if there's a lot of writes while the repack is taking place.

The heuristic is to release 64 times less often when repacking than in a usual 'foreach' to make repack a bit more greedy.  So that's every 64k records (usually it releases every 1024 records)

doc/README.twom.md
brong: yes, that's what the code does - but you can create a transaction at any previous commit and read the state that it was at that commit.

doc/README.twom.md
```diff
@@ -0,0 +1,189 @@
...
+2. once all new records are written, fsync the entire file (still marked dirtry)
+3. mark the header NOT-DIRTY along with the new file length, and fsync
+
+This means that so long as the final header write is atomic, transactions are always completely safe.
+
+## Twoskip downsides:
+
+- Three fsyncs per transaction is kinda expensive!
+- Every write causes the file to be extended, as twoskip format uses exact file size and doesn’t allocate slop.
+- A single write does amplify into 1+N (the level of the record) writes; the append of the new record, and the updates of the immediately previous records in the skiplist at each level.  Since half the records are level 1, half that level 2, etc - the average write updates three separate previous records.  In a recently repacked file, the previous records are likely to be adjacent or nearly so for the lower levels, but this becomes less so over time, so repacks are valuable for both data locality and size reduction.
```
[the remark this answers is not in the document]
brong: probabilistic skiplist.  The level picking function finds the first set bit in a random number:
```
// find random level 1-maxlevel (up to 31 on Linux)
static inline uint8_t randlvl(uint8_t lvl, uint8_t maxlvl)
{
    uint32_t v = random();
    uint8_t i;
    for(i = lvl-1; i < maxlvl-1; i++)
        if (v & (1<<i)) break;
    return i+1;
}
```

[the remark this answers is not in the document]
brong: which is always called with (1, 31) so it's guaranteed to be a level in the range `1:30`

doc/README.twom.md
```diff
@@ -0,0 +1,189 @@
...
+
+## The twoskip file format
+
+Twoskip added an extra innovation to the skiplist format - a second linked list at the bottom level, which includes every single record in order - but with a twist.  The two pointers leapfrog each other, so it’s always the lowest pointer which is updated each time.  This means that if there’s a crash, the pointer which doesn’t point past the old size of the file will always point to the record which used to be the next in the valid records - so a repair is just a matter of walking all the valid level 0 pointers, and zeroing out any future pointers plus fixing up the higher level  pointers.  This is so fast that transaction `abort()` is implemented as just running recovery on the file.
+
+Along with a three phase commit, this makes the twoskip file format incredibly robust.  The three phases are:
+
+1. before writing any new record to the end of file, mark the header as DIRTY, and fsync
+2. once all new records are written, fsync the entire file (still marked dirtry)
+3. mark the header NOT-DIRTY along with the new file length, and fsync
```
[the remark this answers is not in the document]
brong: That makes sense, I'll re-write with the header and record information higher up

doc/README.twom.md
```diff
@@ -0,0 +1,189 @@
...
+
+Since there is no way to “free” a pointer in this format, garbage collection is done by creating a new file and copying all the records into the new file.  This was done with a special INORDER record type, which was otherwise identical to an ADD but only occurred after a repack.
+
+## The twoskip file format
+
+Twoskip added an extra innovation to the skiplist format - a second linked list at the bottom level, which includes every single record in order - but with a twist.  The two pointers leapfrog each other, so it’s always the lowest pointer which is updated each time.  This means that if there’s a crash, the pointer which doesn’t point past the old size of the file will always point to the record which used to be the next in the valid records - so a repair is just a matter of walking all the valid level 0 pointers, and zeroing out any future pointers plus fixing up the higher level  pointers.  This is so fast that transaction `abort()` is implemented as just running recovery on the file.
```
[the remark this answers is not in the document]
brong: In order of the compar function - which will be just a memcmp on the keys unless you pass in an external comparator.

The way twoskip (and then twom) work is to have two values for that linked list.  Do you have a proposed better word than "leapfrog" for "the records are incremented in alternating order, with the lower one always being updated to become the higher one"?  The children's game works precisely like that - the child at the back jumps over the other child and becomes the front.

OK, I'll replace "future pointer" with "higher level pointer that points past the end of the file".  The `recovery1` function remembers each of these, and when it next sees a record with that level, writes its offset into place, repairing the linked list at that level.

doc/README.twom.md
```diff
@@ -0,0 +1,189 @@
+This document describes the twom database format for key-value store within Cyrus.
```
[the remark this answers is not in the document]
brong: Will do

doc/README.twom.md
brong: Nice :)  I think the wikipedia link is enough to have in this doc, but maybe I should read it!

doc/README.twom.md
brong: There's a bunch of comments in the C file too!

docsrc/imap/reference/manpages/systemcommands/cyr_dbtool.rst
```diff
@@ -32,22 +33,32 @@ Description
...
+The *consistent* action runs a consistency check on the DB by calling
 'myconsistent' on it.
 
 The *repack* action will compress the database by removing stale data
 on backends which support it.  It's a NOOP otherwise.
 
+The *damage* action makes the file dirty and then crashes, so it will need
+to be repaired.  It's useful for testing crash recovery speed.
+
+The *batch* action takes commands as bastrings on stdin, and writes results
```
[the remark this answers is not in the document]
brong:
```imap/imapparse.h
81:#define getbastring(pin, pout, buf) getxstring((pin), (pout), (buf), IMAP_BIN_ASTRING)```

A binary astring, and an "astring" is either an atom, a quoted string, or a literal.

File: lib/imapoptions [hunk 60]
```diff
@@ -762,6 +762,10 @@ Blank lines and lines beginning with ``#'' are ignored.
 /* only show the domain for users in other domains than your own (for
    backwards compatibility if you're already sharing. */
 
+{ "cyrusdb_autoconvert", 0, SWITCH, "UNRELEASED" }
+/* automatically convert any existing db to the configured engine for that
```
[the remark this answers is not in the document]
brong: fixed

lib/twom.c
```diff
@@ -0,0 +1,3297 @@
...
+// find_loc is an optimisation on locate.  In the cases where the file hasn't changed and
+// either the current key matches exactly, or is in the gap, or is the immediately next
+// record, we can avoid the full cost of filling out the location.
+//
+// This function can also be used to initialise a blank location, since it will detect
+// that as a "file has changed" and fill out the right values
+static int find_loc(struct twom_txn *txn, struct tm_loc *loc, const char *key, size_t keylen)
+{
+    // the old location is for an old file or this file has been extended
+    if (loc->file != txn->file || loc->end != loc->file->written_size) {
```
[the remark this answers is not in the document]
brong: txn->file can't be NULL, so this can't be a NULL dereference.

doc/README.twom.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Skiplists
+
+A [skiplist](https://en.wikipedia.org/wiki/Skip_list) is a data structure gives O(log N) access to data; by creating multiple linked lists.  Every node contains a pointer to the immediately next node in a sorted order, however some nodes also contain links that “skip over” some number of nodes, hence “skip”.  I won’t describe in detail how skiplists work here, as that’s described well in other places.
+
+## The original cyrus skiplist format
+
+This format is purely just an in-memory skiplist, where new “allocations” are just appending a set of bytes to a file, and the pointers for are all relative pointers, offsets within the file.
+
+Since there is no way to “free” a pointer in this format, garbage collection is done by creating a new file and copying all the records into the new file.  This was done with a special INORDER record type, which was otherwise identical to an ADD but only occurred after a repack.
```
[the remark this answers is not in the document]
brong: I just removed this bit

doc/README.twom.md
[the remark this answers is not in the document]
brong: I rewrote this

brong:
OK I've updated the documentation somewhat, and also updated the image in the description of this PR to be accurate to the current format.

docsrc/imap/reference/manpages/systemcommands/cyr_dbtool.rst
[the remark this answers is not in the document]
dilyanpalauzov: Some examples for cyr_dbtools using strings will be very useful.  E.g. how to get DAV:displayname for a mailbox using `get` and `show`.  The current example `**cyr_dbtool** /var/lib/imap/user_deny.db flat baduser "2<tab>pop3,imap<tab>Denied"` does not contain the mandatory **action** parameter.

lib/twom.h
```diff
@@ -0,0 +1,162 @@
...
+#define INCLUDED_TWOM_H
+
+#include <stdlib.h>
+#include <stdint.h>
+
+struct twom_db;
+struct twom_txn;
+struct twom_cursor;
+
+enum twom_ret {
```
rsto, line 26: Do these enums have to match the ones from CYRUSDB? `_errormap` in cyrusdb_twom.c converts them to CYRUSDB errors anyway. If these enums still better match exactly the CYRUSDB error values, state it in a comment here. The missing -7 and -8 look particularly odd otherwise.

lib/twom.h
```diff
@@ -0,0 +1,162 @@
...
+// header info
+extern size_t twom_db_generation(struct twom_db *db);
+extern size_t twom_db_num_records(struct twom_db *db);
+extern size_t twom_db_size(struct twom_db *db);
+extern const char *twom_db_fname(struct twom_db *db);
+extern const char *twom_db_uuid(struct twom_db *db);
+
+extern int twom_db_sync(struct twom_db *db);
+
+extern const char *twom_strerror(int r);
```
rsto, line 160: This function is not defined.

lib/twom.h
```diff
@@ -0,0 +1,162 @@
...
+    TWOM_IOERROR = -1,
+    TWOM_AGAIN = -2,
+    TWOM_EXISTS = -3,
+    TWOM_INTERNAL = -4,
+    TWOM_NOTFOUND = -5,
+    TWOM_LOCKED = -6,
+    TWOM_READONLY = -9,
+};
+
+// we don't reuse flags, there's 32 bits of space, and we spread through them
```
rsto, line 38: I stumbled over "reuse" here. Do you mean, each flag has one and only one meaning, irrespective if other flags are set?

lib/twom.h
```diff
@@ -0,0 +1,162 @@
...
+enum twom_flagspec {
+    TWOM_CREATE          = 1<<0,    /* Create the database if not existant */
+    TWOM_SHARED          = 1<<1,    /* Open in shared lock mode */
+    TWOM_NOCSUM          = 1<<2,    /* Don't check checksums on read */
+    TWOM_NOSYNC          = 1<<3,    /* Don't msync/fsync on write */
+
+    TWOM_ALWAYSYIELD     = 1<<9,
```
rsto, line 45: I can guess the purpose of most of these flags, but a one-liner docstring like for the first four flags would help.

lib/twom.h
```diff
@@ -0,0 +1,162 @@
...
+typedef uint32_t twom_csum(const char *s, size_t l);
+
+struct twom_open_data {
+    uint32_t flags;
+    int (*compar)(const char *s1, size_t l1, const char *s2, size_t l2);
+    uint32_t (*csum)(const char *s, size_t l);
+    void (*error)(const char *msg, const char *fmt, ...);
+};
+
+typedef struct twom_open_data twom_init;
```
rsto, line 75: The `twom_init` type is defined but not used anywhere in the public API. You could get rid of that typedef and could replace its two occurrences in `cyrusdb_twom.c` with the twom_open_data struct. For better or worse, it would also be more common in Cyrus code to rename `TWOM_INIT` to `TWOM_OPEN_DATA_INITIALIZER`.

lib/twom.h
```diff
@@ -0,0 +1,162 @@
...
+#define TWOM_INIT { 0, NULL, NULL, NULL }
+
+// database operations
+extern int twom_fname_open_db(const char *fname, struct twom_open_data *setup,
```
rsto, line 79: bikeshed: `twom_open_db_fname` would emphasize the operation before the specific mode of operation.

lib/twom.h
```diff
@@ -0,0 +1,162 @@
...
+                           const char *prefix, size_t prefixlen,
+                           twom_cb *p, twom_cb *cb, void *rock,
+                           int flags);
+extern int twom_db_store(struct twom_db *db,
+                         const char *key, size_t keylen,
+                         const char *val, size_t vallen,
+                         int flags);
+
+// utility
+extern int twom_db_consistent(struct twom_db *db);
```
rsto, line 100: Is that a boolean? Since C99 there's the `bool` type and we lean towards requiring C11 anyway? Regardless of the return type, `twom_db_is_consistent` would be clearer.

rsto (commented):
Second round of review, this time for `twom.h` only. I just got a couple of minor points this time. Next round will cover `twom.c`.

doc/README.twom.md
rsto: OK. You still might want to move both of the "Navigating the code" section at the of the README, so that that the Locking section isn't interleaved.

rsto:
I re-read the README and it's so much clearer to me now!

lib/twom.h
brong: Ooh, you're right - I need to add that.

lib/twom.h
brong: They're passed to different functions, so you could have "enum twom_open_flags", "enum twom_store_flags", "enum twom_foreach_flags" etc, but the values are all disjoint so you can't have something mis-interpreted by passing it to the wrong function.

lib/twom.h
brong: I've named everything else with:

twom_input_operation_output (e.g. twom_db_begin_txn, twom_db_begin_cursor, twom_txn_begin_cursor) - but this one is a bit weird.  Maybe it should just be `twom_db_open()`

lib/twom.h
brong: yes, _is_ is better, will do.

brong: Hmm... - no, it's not a boolean.  It returns either IOERROR or CHECKSUM depending what's wrong

lib/twom.h
brong: done

lib/twom.h
brong: done, alphabetised as well

lib/twom.h
rsto: Thanks, that explains this. If they are disjoint, you might want to organize them in the declaration such that each set of flags is visually separated from the other, e.g. 

    ...

     // repack flags
    TWOM_ALWAYSYIELD     = 1<<9,
    TWOM_NOYIELD     = 1<<10,

    // fetch flags
    ...

lib/twom.h
brong: I did rename it, but to "check_consistency"

lib/twom.h
rsto: I'm not against the typedef but I don't understand why you introduce a type name in the header that's nowhere used in any function or struct inside that same header. The `twom_db_open` function takes a `struct twom_open_data` as argument, not a `twom_init`. If you think the typedef is worth it, then I would suggest to typedef the struct to the name `twom_open_data`, though.

Just for completeness, I find the arguments against typedef in the [FreeBSD style guide](https://man.freebsd.org/cgi/man.cgi?query=style&sektion=9) convincing, but the Cyrus code base, including mine, is inconsistent in this regard anyway:
> Avoid using typedefs for structure types. Typedefs are problematic because they do not properly hide their underlying type; for example you need to know if the typedef is the structure itself or a pointer to the structure. In addition they must be declared exactly once, whereas an incomplete structure type can be mentioned as many times as necessary. Typedefs are difficult to use in stand-alone header files: the header that defines the typedef must be included before the header that uses it, or by the header that uses it (which causes namespace pollution), or there must be a back-door mechanism for obtaining the typedef.

lib/twom.c
```diff
@@ -0,0 +1,3314 @@
...
+#define DELETE 6
+#define COMMIT 7
+static const char *typestr[] = { NULL, "DUMMY", "ADD", "FATADD",
+                          "REPLACE", "FATREPLACE", "DELETE", "COMMIT" };
+static uint8_t ptroffset[8]      = { 0,  8,  8, 24, 16, 32,  8,  8 };
+static uint8_t ancestoroffset[8] = { 0,  0,  0,  0,  8, 24,  8,  0 };
+static uint8_t fatrecord[8]      = { 0,  0,  0,  1,  0,  1,  0,  0 };
+static uint8_t hastail[8]        = { 0,  0,  1,  1,  1,  1,  0,  0 };
+
+#define SETUP_FLAGS (TWOM_CSUM_NULL|TWOM_CSUM_XXH64|TWOM_CSUM_EXTERNAL)
```
dilyanpalauzov, line 69: This macro is not used.

lib/twom.h
brong: I'm convinced, have made both these changes

lib/twom.c
brong: Removed

lib/twom.c
```diff
@@ -0,0 +1,3312 @@
...
+                 struct twom_db **ret, struct twom_txn **txnp)
+{
+    struct twom_db *mydb;
+    int r = 0;
+
+    /* do we already have this DB open? */
+    for (mydb = open_twom; mydb; mydb = mydb->next) {
+        if (strcmp(mydb->fname, fname)) continue;
+        if (txnp) {
+            r = twom_db_begin_txn(mydb, setup->flags & TWOM_SHARED, txnp);
```
rsto, line 2406: I want to clarify the purpose of thatsetup->flags & TWOM_SHARED: This deliberately removes all flags, only keeping TWOM_SHARED if that's what the caller passed as flags. Flags such as NOSYNC get wiped, presumably because a second open on an already open database should not reset that flag.
Isn't that then kind of lying to the caller, e.g. if I write the code to expect NOSYNC being enabled, but the db already got opened without NOSYNC in the first open?

lib/twom.c
```diff
@@ -0,0 +1,3312 @@
...
+    return db->fname;
+}
+
+int twom_db_sync(struct twom_db *db)
+{
+    if (!db->openfile) return 0;
+    return tm_commit(db, db->openfile->size);
+}
+
+size_t twom_db_generation(struct twom_db *db)
```
dilyanpalauzov, line 2879: This function is not called from anywhere.

rsto (approved):
This looks good to me.

Since this is a big change, I focused mostly on cross-checking the code with the design outlined in the README. I only found one potentially too-early unmarked dirty flag, but haven't dug into its implications.

Other than that, error reporting might be notched up a bit in the standard error handler, but I guess we'll use xsyslog in Cyrus for error reporting and that reports `errno` for us anyway.

rsto:
One more thing: I built and cunit-tested this on my aarch64 setup without issues, including valgrind.

lib/twom.h
```diff
@@ -0,0 +1,160 @@
...
+    TWOM_CSUM_XXH64      = 1<<28,   /* use the XXH64 checksum algorithm when creating or repacking */
+    TWOM_CSUM_EXTERNAL   = 1<<29,   /* use an external checksum algorithm (must be passed in init) */
+    TWOM_COMPAR_EXTERNAL = 1<<30    /* use an external comparison function (must be passed in init) */
+};
+
+typedef int twom_cb(void *rock,
+                    const char *key, size_t keylen,
+                    const char *data, size_t datalen);
+typedef int twom_compar(const char *s1, size_t l1, const char *s2, size_t l2);
+typedef uint32_t twom_csum(const char *s, size_t l);
```
dilyanpalauzov, line 65: The terms `twom_compar` and `twom_csum` appear only here and thus only once.

lib/twom.h
```diff
@@ -0,0 +1,160 @@
...
+    uint32_t flags;
+    int (*compar)(const char *s1, size_t l1, const char *s2, size_t l2);
+    uint32_t (*csum)(const char *s, size_t l);
+    void (*error)(const char *msg, const char *fmt, ...);
+};
+
+#define TWOM_OPEN_DATA_INITIALIZER { 0, NULL, NULL, NULL }
+
+// database operations
+extern int twom_db_open(const char *fname, struct twom_open_data *setup,
```
dilyanpalauzov, line 77: What is the difference between with this `extern` and without it?

lib/twom.h
```diff
@@ -0,0 +1,160 @@
...
+// there's 32 bits of space available - though not all flags have meaning in all contexts.
+enum twom_flagspec {
+    TWOM_CREATE          = 1<<0,    /* Create the database if not existant */
+    TWOM_SHARED          = 1<<1,    /* Open in shared lock mode */
+    TWOM_NOCSUM          = 1<<2,    /* Don't check checksums on read */
+    TWOM_NOSYNC          = 1<<3,    /* Don't msync/fsync on write */
+    TWOM_NONBLOCKING     = 1<<4,    /* When taking a lock, return immediately if the file is already locked */
+
+    TWOM_ALWAYSYIELD     = 1<<9,    /* Yield foreach before every callback */
+    TWOM_NOYIELD         = 1<<10,   /* Never yield a read transaction lock */
```
dilyanpalauzov, line 48: ` TWOM_NOYIELD` isn‘t used.

lib/twom.c
```diff
@@ -0,0 +1,3312 @@
...
+
+    if (foundkey) *foundkey = KEYPTR(ptr);
+    if (foundkeylen) *foundkeylen = KEYLEN(ptr);
+    if (data) *data = VALPTR(ptr);
+    if (datalen) *datalen = VALLEN(ptr);
+
+    return 0;
+}
+
+int twom_db_begin_cursor(struct twom_db *db,
```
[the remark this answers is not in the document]
brong: That's deliberate - the idea here is to have a fairly complete API covering the things that a caller might want, even if cyrusdb_twom doesn't use them all itself.

lib/twom.c
brong: Likewise

lib/twom.c
```diff
@@ -0,0 +1,3312 @@
...
+        }
+
+        if (sbuf.st_size < HEADER_SIZE + HLCALC(DUMMY, MAXLEVEL) + 8)
+            goto badfile;
+
+        // we're not interested in getting the latest file
+        if (forcefile) break;
+
+        if (stat(db->fname, &sbuffile) == -1) {
+            db->error("read_lock stat failed",
```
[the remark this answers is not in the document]
brong: As you noted, db->error can read and report errno itself, since it's a global!

doc/README.twom.md
```diff
@@ -0,0 +1,235 @@
...
+the start of the range of records that it commits).
+
+### Navigating the code - data structures:
+
+Structures named `twom_` are public, `tm_` are entirely internal.
+
+- `struct twom_db` - this is refcounted and contains pointers to all the refcounted resources that it has open.  When you open a DB again, you get another pointer to the same file, and when you open a transaction or a location on the database, you hold a reference that the file won’t be closed under you.
+- `struct twom_txn` - a transaction on the database.
+    - If this is an exclusive transaction (not TWOM_SHARED) then you can write to the file and either abort or commit.  A transaction dirties the file on first write.  An exclusive transaction forces the file to remain locked exclusively for the duration of the transaction.
+    - If this is read-only transaction, it can be either MVCC or not.  If it is MVCC, then it never changes file, it always reads from the same file, though it may still need to refresh the length if the lock has been released.  Read-only transactions can release the lock, either explicitly using `twom_yield` or implicitly every N transactions to avoid starving other database users.
```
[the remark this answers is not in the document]
brong: I think I removed that somewhere along the way.  I'll re-add the API even though we're not using it at the moment.  I'd like to eventually plumb it through.

lib/twom.h
brong: https://stackoverflow.com/questions/496448/how-to-correctly-use-the-extern-keyword-in-c says "it's the default, you don't need to be explicit" - fair enough.  So "nothing".

lib/twom.c
```diff
@@ -0,0 +1,3312 @@
...
+    size_t page_size = 1<<14; // 16k
+    return ((offset + offset / 4) + page_size - 1) & ~(page_size - 1);
+}
+
+static inline int tm_commit(struct twom_db *db, size_t len)
+{
+    assert(db->openfile);
+    if (!db->openfile->dirty) return 0;
+    assert(db->openfile->has_datalock == 2);
+    db->openfile->dirty = 0;
```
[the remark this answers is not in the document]
brong: Hmm, I guess so!

I looked at doing this, but I'm not sure what we can actually DO.  A failed msync basically means that the drive has gone away, so leaving this not-dirty will mean we don't fail all the asserts as we close out.  I will add a comment.

lib/twom.c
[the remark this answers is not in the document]
brong: added to buildin error handler.

lib/twom.c
brong: Yeah, that was a mistake - this USED to take a boolean.  I've switched it to pass the whole flags through.

It might be useful to check flags and return a "you're holding it wrong" error.  I'll add a comment there, but otherwise leave it for now.

doc/README.twom.md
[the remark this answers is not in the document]
brong: Ahh yeah it was renamed twom_db_yield in the great API realignment.  Fixed.

lib/twom.h
brong: I removed them all

lib/twom.h
brong: Now it is :)

lib/twom.c
```diff
@@ -0,0 +1,3312 @@
...
+        }
+        if (TYPE(nextptr) == DELETE) {
+            deleted_offset = next[0];
+            dirty_size += 24;
+            next[0] = ANCESTOR(nextptr);
+            nextptr = safeptr(loc, next[0]);
+            if (!nextptr) return TWOM_IOERROR;
+        }
+
+        cmp = file->compar(KEYPTR(nextptr), KEYLEN(nextptr),
```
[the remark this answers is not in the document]
brong: They never should, but... yeah OK it's NULL but zero length.  Hmm.  Maybe it should return "" for KEYPTR for the DUMMY.   Fixed.

lib/twom.h
brong: I've now used them in the struct below :)

brong:
Thanks @rsto I've added some more doc, made a few tiny fixes, and fixed up cyr_dbtool to do better error handling and patch readonly through.

lib/twom.c
```diff
@@ -0,0 +1,3328 @@
...
+
+    struct twom_db *db = txn->db;
+    struct tm_file *file = txn->file;
+
+    while (txn->end + 24 <= file->committed_size) {
+        const char *ptr = file->base + txn->end;
+        size_t reclen = RECLEN(ptr);
+        // ensure the entire record fits!
+        if (txn->end + reclen > file->committed_size)
+             return TWOM_IOERROR;
```
dilyanpalauzov, line 2108: Indentation.

PR #5159, opened 2024-12-05
mbname_from_path: Don't crash with '..' at top of tree

wolfsage:
If we're at the top of the tree and we try to `mbpath -p ..`, we have no mbname pointer anymore, so don't try to call mbname_userid() on it.


ksmurchison (approved):
Approved, provided that CI succeeds, which I expect it will

PR #5162, opened 2024-12-09
Sieve: allow private Final-Recipient in rejects, allow non-Original-Recipient header for original recipient.

rjbs:
This replaces a longstanding private commit in Fastmail's Cyrus, which replaces Final-Recipient with an encrypted version of the original recipient.  The goal is to provide a string that can be traced back to a recipient by the Cyrus operator, but doesn't disclose the final delivery username to the bounce recipient.

Using session_id should accomplish the same result, at least if logs are maintained.  It's also a lot less complex than shipping our own little RC4 implementation!

This then replaces *another* private commit, allowing the Sieve rejection code to pluck the original recipient from a header other than Original-Recipient.


wolfsage:
lgtm, but I think we also need a changes/next/* entry? Like b48001d0e96

wolfsage (approved):

rjbs:
> lgtm, but I think we also need a changes/next/* entry? Like [b48001d](https://github.com/cyrusimap/cyrus-imapd/commit/b48001d0e96ad798a5ecb8a9f311f2ac75c800bf)

Great catch, thank you.

changes/next/private_sieve_mdn
```diff
@@ -0,0 +1,19 @@
+Description:
+
+The new `private_sieve_mdn` option helps prevent leaking usernames.
+
+Config changes:
```
wolfsage, line 5: Nit: For consistency should have another new line before "Config changes"

imap/lmtp_sieve.c
```diff
@@ -492,7 +492,12 @@ static int send_rejection(const char *userid,
             config_servername, CYRUS_VERSION, SIEVE_VERSION);
     if (origreceip)
         buf_printf(&msgbuf, "Original-Recipient: rfc822; %s\r\n", origreceip);
-    buf_printf(&msgbuf, "Final-Recipient: rfc822; %s\r\n", mailreceip);
+
+    if (config_getswitch(IMAPOPT_PRIVATE_SIEVE_MDN))
+      buf_printf(&msgbuf, "Final-Recipient: rfc822; %s\r\n", session_id());
+    else
+      buf_printf(&msgbuf, "Final-Recipient: rfc822; %s\r\n", mailreceip);
+
```
elliefm, line 500: Cryus uses four-space indents, not two :)

changes/next/private_sieve_mdn
elliefm: In my mind, we have two blank lines before headings in these files, so that single blank lines can be paragraph separators within a section, though in practice the section bodies are often single paragraphs anyway, or even single lines.  It's going to be transcribed to release notes and upgrade docs by a human anyway, so as long as the required information is there (and in this case it is, thanks!), the syntax doesn't matter too much.

I imagine at some point we'll automate collating these changes files into release notes entries, and to support that we'll probably need to be more specific and strict about the contents and syntax of these files, but we're not there yet.

elliefm (approved):
One nit about four-space indents, but I'm approving it now because it does not need my re-review once that's fixed.

elliefm:
> Shouldn't merge until v3.12 forked. Will require tweak to Fm internal builds.

I'll add the Do Not Merge label so it (hopefully) doesn't get merged accidentally

changes/next/private_sieve_mdn
rjbs: I have added the second blank line, but may push back on this in the future, depending on how the future goes!  For now, just lookin' to get it done. :)

changes/next/sieve_mdn_original_recipient_header
```diff
@@ -0,0 +1,21 @@
+Description:
+
+The new `sieve_mdn_original_recipient_header` allows an install to get the
+original recipient from a header other than Original-Recipient.
+
+
+Config changes:
+
+This adds `sieve_mdn_original_recipient_header`, a yes/no option which, if set,
```
wolfsage, line 9: This "yes/no option" is incorrect, since you're providing a header name

wolfsage (changes requested):

changes/next/sieve_mdn_original_recipient_header
rjbs: Fixed in amendment.

wolfsage (approved):

PR #5163, opened 2024-12-09
GitHub action: build on Cyrus Bookworm image

rjbs:


wolfsage (approved):

elliefm:
I don't know how this stuff works, do we expect the old buster image to still be available indefinitely?  If not, I'll need to backport this to the other branches so they start building on bookworm too.  Though actually, that's a good idea anyway so I'll just tag it up now (but sing out if you object, or have more context or whatever)

rjbs:
@elliefm They should be available basically indefinitely, but updating is reasonable.

PR #5164, opened 2024-12-10
imapd: don't auto-expunge if replicaonly

brong:
This came up where a test script was logging into a replica and SELECTing the INBOX.  If there was a DELETED message that wasn't expunged, then it would try to write quote while expunging on the replica, and fail an assert.

This simplifies the code by putting it in a function.

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -928,6 +928,14 @@ static void imapd_log_client_behavior(void)
                         client_behavior.did_xlist       ? " xlist=<1>"        : "");
 }
 
+static void maybe_autoexpunge(void)
+{
+    if (config_getswitch(IMAPOPT_REPLICAONLY)) return;
+    if (!config_getswitch(IMAPOPT_AUTOEXPUNGE)) return;
+    if (!index_hasrights(imapd_index, ACL_EXPUNGE)) return;
+    index_expunge(imapd_index, NULL, 1);
+}
+
 static void imapd_reset(void)
 {
     int i;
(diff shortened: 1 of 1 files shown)


imap/imapd.c
```diff
@@ -5004,10 +5009,13 @@ static void cmd_close(char *tag, char *cmd)
     }
 
     /* local mailbox */
-    if ((cmd[0] == 'C' || config_getswitch(IMAPOPT_AUTOEXPUNGE)) && index_hasrights(imapd_index, ACL_EXPUNGE)) {
+    if (cmd[0] == 'C') {
```
wolfsage, line 5012: This looks like a mistake - now we'll call `index_expunge(imapd_index, NULL, 1)` on `close` without checking if they have EXPUNGE rights, when before we'd check 

If I open two ctel sessions, and do  `x SELECT INBOX` and then `x STORE 1 +Flags \Deleted` in one, and then in the other I do `x EXAMINE INBOX` and `x CLOSE`, imapd crashes with:
```
2024-12-10T10:02:19.351166-05:00 box-fastmail-01 (err) slotf1n01/imapadmin[3149412]:  Internal error: assertion failed (aborting): imap/mailbox.c: 4516: mailbox_index_islocked(mailbox, 1)

```

wolfsage (changes requested):

imap/imapd.c
brong: Oh yeah!  You're right, the `||` wasn't unconditional.  Fixed

rsto (approved):
LGTM

PR #5166, opened 2024-12-11
Don't send the "action strings" when notifying

rjbs:
These strings are largely useful for debugging, not end user consumption.

(This commit message was written by Ricardo Signes, replacing the previous internal Fastmail commit message.  The code change is by Bron Gondwana.)

### DIFF
--- a/sieve/script.c
+++ b/sieve/script.c
@@ -430,7 +430,6 @@ static int build_notify_message(sieve_interp_t *i,
 static int send_notify_callback(sieve_interp_t *interp,
                                 void *message_context,
                                 void *script_context, notify_list_t *notify,
-                                char *actions_string,
                                 const char **errmsg)
 {
     sieve_notify_context_t nc;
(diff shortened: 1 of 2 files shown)


rjbs:
Marked do not merge until freeze is over.

sieve/script.c
```diff
@@ -430,7 +430,7 @@ static int build_notify_message(sieve_interp_t *i,
 static int send_notify_callback(sieve_interp_t *interp,
                                 void *message_context,
                                 void *script_context, notify_list_t *notify,
-                                char *actions_string,
+                                char *actions_string __attribute__((unused)),
```
ksmurchison, line 433: Do we want to just remove the `actions_string` parameter all together rather than just commenting out the place where it is used?

rjbs: Yes, probably. :)

rjbs: Done.

ksmurchison (approved):

PR #5167, opened 2024-12-11
don't add timezone data in http_caldav_sched.c

rjbs:
We removed this line quite some time ago in Fastmail builds, because we sometimes saw downstream problems with Exchange.

Ken Murchison's advice was that in general clients should not (or no longer) be looking at TZ data included in events, anyway.  Adding it could cause problems, but probably helps no one.  Instead, we'll just leave it be.

(This commit message was written by Ricardo Signes, replacing the previous internal Fastmail commit message.  The code change is by Bron Gondwana.)


rjbs:
Labeled `do-not-merge` until freeze is over.

ksmurchison (approved):

PR #5169, opened 2024-12-13
jmap_contact.c: base64url encode vCard UIDs as JMAP ids

ksmurchison:
This prevents us from return "urn:uuid:xxx" as an object id, but also allows us to derive the vCard UID from the object id to avoid any changes to dav.db


rsto:
@ksmurchison I removed myself as reviewer while this is a draft. Please re-add me for review when it's ready.

rjbs:
Ken says this should be closed and not merged.  Agreed.

PR #5171, opened 2024-12-16
message.c: don't try to access past end of string

ksmurchison:


imap/message.c
```diff
@@ -1964,8 +1964,7 @@ static void message_parse_content(struct msg *msg, struct body *body,
         len = endline - line;
         msg->offset += len;
 
-        if (line[0] == '-' && line[1] == '-' &&
-            message_pendingboundary(line, len, boundaries)) {
+        if (message_pendingboundary(line, len, boundaries)) {
```
wolfsage, line 1967: This does mean we'll call this method for every single line in an email when before we wouldn't. Performance wise this maybe doesn't matter since messages don't have too many lines. What do you think?

ksmurchison: Switched back to the original test but checking length first

wolfsage (approved):

ksmurchison:
@elliefm this is a fix for a crasher.  Should we merge it before the next release or wait?

elliefm:
@ksmurchison Bugfix, so it's fine to merge :)

PR #5174, opened 2024-12-17
in GitHub Actions, use cyrus-docker container from GHCR

rjbs:


elliefm (approved):

rjbs:
@elliefm I agree with your assessment.

elliefm:
I don't remember if I've been consistent about adding backport tags to your other recent github actions changes...  I think we always want to backport github actions changes, so we're not surprised by CI runs for the released versions doing different things.

Generally, please do just add backport tags to your own PRs if you know they'll need backporting, or ask if you're not sure.  I'll usually add them myself if I happen to look at the PR on the way through and think it's needed, but I don't always look at every PR.

PR #5176, opened 2024-12-18
SieveScript/test: Allow prepopulating variables so your sieve scripts can behave differently during test

wolfsage:


ksmurchison (approved):
This all looks sane to me

imap/jmap_sieve.c
```diff
@@ -2019,6 +2024,26 @@ static int jmap_sieve_test(struct jmap_req *req)
         envelope = NULL;
     }
 
+    /* variables */
+    if (JNOTNULL(variables) && json_is_object(variables)) {
+        jmap_parser_push(&parser, "variables");
+
+        construct_hash_table(&vars, 64, 0);
```
wolfsage, line 2031: Probably I should tune this down to 16 or something?

ksmurchison: I wouldn't argue with 16 or 32

wolfsage: Yeah 16 buckets seems like more than enough

imap/jmap_sieve.c
```diff
@@ -2257,7 +2259,6 @@ static int jmap_sieve_test(struct jmap_req *req)
         }
     }
 
-    if (interp_ctx.cstate) conversations_commit(&interp_ctx.cstate);
```
wolfsage, line 2190: The convdb is opened for us by jmap_api. If we commit it here, we'll crash later because jmap_api.c will try to commit it too

ksmurchison (approved):

wolfsage:
@rsto I've added include-in-fastmail to this one o/

PR #5177, opened 2024-12-19
WIP: 64bit nanosecond jmapids

ksmurchison:
TODO:

- Add code to upgrade mailboxes.db to include createdmodseqs for all mailboxes, and resolve any createdmodseq conflicts


ksmurchison:
This isn't complete yet but could probably use a sanity check

rsto:
@ksmurchison since this is a WIP review, could you please let me know the main file or functional areas you would like me to sanity check?

ksmurchison:
> @ksmurchison since this is a WIP review, could you please let me know the main file or functional areas you would like me to sanity check?

You should probably look at conversations.c and jmap_mail.c

rsto:
> You should probably look at conversations.c and jmap_mail.c

Thanks, will do.

imap/conversations.h
```diff
@@ -289,6 +289,12 @@ extern int conversations_guid_cid_lookup(struct conversations_state *state,
    strcmpsafe(conv_guidrec_uniqueid(rec), mailbox_uniqueid(mbox)) : \
    strcmpsafe(conv_guidrec_mboxname(rec), mailbox_name(mbox)) \
  )
+
+/* J record items */
+#define CONV_JMAPID_SIZE 11  // 64-bits base64-encoded w/o padding
+extern int conversations_lookup_jmapid(struct conversations_state *state,
```
rsto, line 295: The other lookup function in conversations.h is named `conversations_guid_cid_lookup`, so should this be named consistently, e.g. as `conversations_jmapid_guidrep_lookup`? Not that I think that would be a great name, but I would prefer the API to use consistent naming either way.

imap/conversations.c
```diff
@@ -2357,15 +2361,57 @@ static int conversations_set_guid(struct conversations_state *state,
...
+        NANOSEC_TO_JMAPID(&key, internaldate);
+
+        /* Fetch existing record (for ref_count) */
+        r = cyrusdb_fetch(state->db, buf_base(&key), buf_len(&key),
+                          &data, &datalen, &state->txn);
+        if (!r || (add && r == CYRUSDB_NOTFOUND)) {
+            uint32_t ref_count = 0;
+            char *p;
+
+            if (datalen) ref_count = strtoul(data, &p, 16);
```
rsto, line 2388: Why are the jmapids refcounted by GUID?

imap/conversations.c
```diff
@@ -3305,4 +3352,33 @@ EXPORTED int conversations_zero_modseq(struct conversations_state *state)
...
 
+EXPORTED int conversations_lookup_jmapid(struct conversations_state *state,
+                                         const char *jidrep, char guidrep[])
+{
+    char key[CONV_JMAPID_SIZE+2] = "J";
+    size_t datalen = 0;
+    const char *data, *p;
+    int r;
+
+    if (strlen(jidrep) != CONV_JMAPID_SIZE)
```
rsto, line 3363: Should this better return an internal error? Or is there any scenario where the caller might legitimately pass a jidrep that doesn't match the expected size?

imap/jmap_mail.c
```diff
@@ -1077,9 +1077,14 @@ static char *_emailbodies_to_html(struct emailbodies *bodies, const struct buf *
...
 }
 
-static const char *_guid_from_id(const char *msgid)
+static const char *_guid_from_id(struct conversations_state *cstate,
+                                 const char *emailid)
 {
-    return msgid + 1;
+    static char guidrep[2*MESSAGE_GUID_SIZE+1];
+
+    int r = conversations_lookup_jmapid(cstate, emailid + 1, guidrep);
```
rsto, line 1085: non-actionable comment: I get why this is done here, but it sucks to have a db call introduced into a function that formerly only had the cost of simple pointer arithmetic.

imap/jmap_mail.c
```diff
@@ -9111,11 +9181,14 @@ static void _email_append(jmap_req_t *req,
     fclose(f);
     f = NULL;
 
+    jmap_set_emailid_ts(internaldate, detail->email_id);
```
rsto, line 9184: I would prefer the jmap_set_emailid functions to be more explicitly named: e.g. jmap_set_email_id_from_timespec and jmap_set_email_id_from_nanoepoch. I had to grep through to the code to figure when to pass what.

rsto (commented):
I gave conversations.c and jmap_mail.c a read. It looks to implement the requirements, I just have a question about refcounts as well as some thoughts about function names.

imap/conversations.h
ksmurchison: Sure.  I can rename it.

imap/conversations.c
ksmurchison: Because there could be multiple G records with the same value.  If I have the same message in more than one mailbox, I don't want to remove the J key when I remove one instance of the message.

This popped up with one of the Cass tests.

imap/conversations.c
ksmurchison: I guess the JMAP code should be checking the id early, and this function can assume that its well-formed.

Should I remove this check?

imap/jmap_mail.c
ksmurchison: sure.

imap/jmap_mail.c
rsto: Already the emailid+1 was really a code smell and I figured doing that in a single function at least contains that. It's not a big deal to call into conversations now, I guess, at least I don't see this being called in a hot loop (or is it?)

ksmurchison: It shouldn't be

imap/conversations.c
rsto: I would rather keep that check but either return CYRUSDB_INTERNAL (and assume that this gets logged properly by the caller), or log the incorrect size here and return NOTFOUND. My point is: if calling that with an invalid size is actually an error by the caller, then this function should not hide that just behind a NOTFOUND.

imap/conversations.c
rsto: That refcount only seems to be relevant when updating a entry in conversations.db, rather than when reading it, right? In this case, it is really necessary to denormalize the refcount here, or would be OK to lookup if any G record on a top-level mailbox for that guid still exists?

rsto: Yes, I'm uncertain if adding the refcount here is useful enough to warrant a potential data inconsistency. I'm only challenging that implementation, I'm not insisting on removing it.

Typically, one would denormalize data if it brings enough performance improvement. But it looks to me as if in this particular case, the refcount only is evaluated during a write (already a slower operation in comparison to reads) and it only saves reading one row in the db. In contrast, in jmap_mail.c we now do a read for every `_guid_from_id` call because it's cheap to do so. Why isn't it cheap enough here for the same reasons?

Again, this is more an open discussion, rather than an explicit request to remove the refcount.

ksmurchison: The lookup for an existing GUID would have to be a foreach loop because we need to ignore any G records that match the one we're writing and any that have mime parts.  Or am I thinking about this wrong?

rsto: yes, it would need to be a loop and only check for G record without imap part (the same as for guid_from_id, right??). Might be that this is too costly, maybe not, hard to tell without actually seeing this in practice. It's OK to keep refcounting.

imap/conversations.c
ksmurchison: do you have a preference for the log level?

ksmurchison: I guess I will have to audit every place where _guid_from_id() is called to see if it tries to validate the emailid.

I added the check just to short-circuit doing a DB lookup, because we know its going to fail.

imap/conversations.c
ksmurchison: Turns out that ref-counting was broken because it was incrementing everytime a G record was written, including for flag updates.  I removed the ref-counting and just looked for an existing G record to determine whether to add/delete the J record.

imap/conversations.c
elliefm: [This is just a drive by comment cause something caught my eye in the email notification. I haven't read any of this closely.]

> if calling that with an invalid size is actually an error by the caller [...]

perhaps an assert rather than log-and-return, if it indicates a programmer error or unexpected runtime state

ksmurchison: Its not really meant to be a check on a programmer error as in some cases jmap_mail.c just passes whatever value it received by the client.  The current comment and code is this:

`    /* sanity check the ID:
       - to prevent strcat() from overrunning a fixed-length buffer
       - to short-circuit an ill-fated DB fetch
    */
    idlen = strspn(jidrep, jmapid_alphabet);
    if (idlen != CONV_JMAPID_SIZE || jidrep[idlen] != '\0') {
        /* we know this J key can't exist */
        return CYRUSDB_NOTFOUND;
    }
`

imap/mailbox.c
```diff
@@ -2105,7 +2105,7 @@ static int mailbox_buf_to_index_record(const char *buf, int version,
     }
 
     cache_offset_field = ntohl(*((bit32 *)(buf+OFFSET_CACHE_OFFSET)));
-    record->internaldate = ntohll(*((bit64 *)(buf+OFFSET_INTERNALDATE)));
+    record->internaldate.tv_sec  = ntohll(*((bit64 *)(buf+OFFSET_INTERNALDATE)));
```
wolfsage: Is this portable on a 32 bit system where long long will be 64 bits but time_t could be 32? Do we care?

imap/mailbox.c
```diff
@@ -4799,8 +4815,12 @@ EXPORTED int mailbox_append_index_record(struct mailbox *mailbox,
         }
     }
 
-    if (!record->internaldate.tv_sec)
-        record->internaldate.tv_sec = time(NULL);
+    clock_gettime(CLOCK_REALTIME, &now);
```
wolfsage, line 4838: This clock_gettime() could be moved into the if block I think? Doesn't seem like now is used anywhere else

wolfsage (commented):
Some small notes for now. More tomorrow.

imap/lmtp_sieve.c
```diff
@@ -191,7 +191,9 @@ static int getmailboxidexists(void *sc, const char *extname)
 {
     script_data_t *sd = (script_data_t *)sc;
     const char *userid = mbname_userid(sd->mbname);
-    char *intname = mboxlist_find_uniqueid(extname, userid, sd->authstate);
```
wolfsage, line 194: Does jmap_sieve.c need this same change?

imap/mailbox.c
ksmurchison: We would have the same problem before this change with record->internaldate which was also time_t.

Do you have any suggestions?

imap/mailbox.c
ksmurchison: Yep.  Will do.  I think I did this to keep the diff smaller, but that was dumb.

cassandane/tiny-tests/Append/append_same_internaldate
```diff
@@ -0,0 +1,36 @@
...
+
+        $imaptalk->append('INBOX',
+                          '26-Jan-2025 15:13:54 -0500', $mimeMessage) || die $@;
+    }
+
+    xlog $self, "Verify that all messages have the same internaldates";
+    $imaptalk->examine('INBOX');
+    my $res = $imaptalk->fetch('1:*', 'INTERNALDATE');
+
+    # Did we log an conflicts?
```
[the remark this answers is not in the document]
ksmurchison: Thx.  I will fix

imap/lmtp_sieve.c
wolfsage: Also imap/http_applepush.c

lib/util.h
```diff
@@ -168,6 +168,11 @@ extern const unsigned char convert_to_uppercase[256];
         (ts)->tv_nsec = (nanosec) % 1000000000; \
 }
 
+#define NANOSEC_TO_JMAPID(buf, nanosec) {                                   \
+        uint64_t u64 = htonll(UINT64_MAX - (nanosec));                      \
```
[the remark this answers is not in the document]
ksmurchison: Yes, as requested by Neil

imap/mailbox.c
wolfsage: Hmm fair point. I suppose on such systems internaldate will be 32bits anyway so maybe this is fine? Alright ignore this

imap/index.c
```diff
@@ -6858,6 +6858,13 @@ static int index_sort_compare(MsgData *md1, MsgData *md2,
         case SORT_GUID:
             ret = message_guid_cmp(&md1->guid, &md2->guid);
             break;
+        case SORT_EMAILID:
+            // EMAILIDs are an ASCII-order encoding of
+            // (INT_MAX - nano_internaldate) =>
+            // reverse the order of nano_internaldate comparison
+            ret = numcmp(TIMESPEC_TO_NANOSEC(&md2->internaldate),
+                         TIMESPEC_TO_NANOSEC(&md1->internaldate));
```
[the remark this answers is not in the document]
ksmurchison: So, numcmp(a, b) evaluates (a - b)

In the case of new email ids:
a = MAX - nano1
b = MAX - nano2

Doing some algebra:
(MAX - nano1) - (MAX - nano2) = MAX - nano1 - MAX + nano2 = nano2 - nano1 = numcmp(nano2, nano1)

Should I spell this out in the comment

imap/conversations.c
```diff
@@ -3368,4 +3368,42 @@ EXPORTED int conversations_zero_modseq(struct conversations_state *state)
...
+                      &data, &datalen, &state->txn);
+    if (r) return r;
+
+    if (datalen < 2*MESSAGE_GUID_SIZE) {
+        xsyslog(LOG_NOTICE, "IOERROR: malformed data value in J record",
+                "key=<%s>", key);
+        return CYRUSDB_NOTFOUND;
+    }
+
+    strncpy(guidrep, data, datalen);
```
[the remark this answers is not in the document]
ksmurchison: Fixed

imap/lmtp_sieve.c
ksmurchison: jmap_sieve.c for sure.  Uncertain on http_applepush.c.  Let's discuss with @brong on Monday

imap/jmap_mailbox.c
```diff
@@ -750,7 +770,7 @@ static int jmap_mailbox_get_cb(const mbentry_t *mbentry, void *_rock)
 
     /* Move this mailbox of the lookup list */
     if (rock->want) {
-        hash_del(mbentry->uniqueid, rock->want);
+        hash_del(id, rock->want);
```
wolfsage, line 773: This fails to compile with `-Og`:
```
imap/jmap_mailbox.c: In function 'jmap_mailbox_get_cb':
imap/jmap_mailbox.c:773:9: error: 'id' may be used uninitialized [-Werror=maybe-uninitialized]
  773 |         hash_del(id, rock->want);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~
imap/jmap_mailbox.c:740:17: note: 'id' was declared here
  740 |     const char *id;
```

Probably just `const char *id = NULL;` ...

ksmurchison: Fixed

imap/mailbox.c
ksmurchison: Actually, this conversion isn't even in the current code anymore.  We do assign tv_sec the result of  uint64 / 1000000000, but we won't have to worry about that until 2038.

elliefm: FWIW: We already don't support weird platforms where time_t is not the system's native word size (e.g. some mobile device architectures have 32-bit cpus but a 64-bit time_t), and we barely support platforms where both are 32-bit.  See extensive discussion in #4876 and related PRs etc.

Though as far as I understand, any OS we support, on any hardware that can provide a 64-bit integer, will use that for time_t.  So we might need to worry about 'long' or 'int' being different from time_t (e.g. those 32-bit mobile devices), but not 'long long'.  At least until 128-bit integer types become common...

imap/lmtp_sieve.c
wolfsage: Oh did we figure out the applepush questioN?

imap/mboxlist.c
```diff
@@ -5651,7 +5684,20 @@ static int _check_rec_cb(void *rock,
...
-            *do_upgrade = (mbentry->name == NULL);
+            crock->do_upgrade = (mbentry->name == NULL);
+            mboxlist_entry_free(&mbentry);
+            // continue checking for J records
+        }
+        break;
+    }
+
+    case KEY_TYPE_JID: {
+        /* Verify that we have a valid I record */
```
wolfsage, line 5695: This should be "valid J record" shouldn't it?

ksmurchison:
Superseded by #5392

PR #5181, opened 2024-12-20
Fix conversations issues

brong:
There's two issues here:

1) basecid inconsistency causing TEMP/REAL mismatch

2) poor inbox detection on user rename causing bigger issues.  The second one is ALSO a known Fastmail issue on user rename, such that we do a full conversations rebuild after renaming.  This will fix that.

The reason these weren't detected earlier is that `ctl_conversationsdb -A` ONLY works on databases which support fetchnext, which skiplist doesn't, so we didn't notice these until switching to make twoskip the default.

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -7605,6 +7605,8 @@ static void cmd_rename(char *tag, char *oldname, char *newname, char *location, 
     int rename_user = 0;
     int mbtype = 0;
     mbentry_t *mbentry = NULL;
+    mbentry_t *olddestmbentry = NULL;
+    mbentry_t *newdestmbentry = NULL;
     struct renrock rock = {0};
     const char *orig_oldname = oldname;
     const char *orig_newname = newname;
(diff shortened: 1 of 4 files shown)


ksmurchison (commented):
The changes make sense, but Replication.full_rename failed.
Don't know if this is an issue with the test or the actual code

imap/imapd.c
```diff
@@ -7875,6 +7875,25 @@ static void cmd_rename(char *tag, char *oldname, char *newname, char *location,
              imapd_userisadmin) {
 
         rename_user = 1;
+
+        // we can't rename users if the new inbox already exists!
+        mbentry_t *destmbentry = NULL;
+        r = mboxlist_lookup_allow_all(newmailboxname, &destmbentry, NULL);
+        if (r != IMAP_MAILBOX_NONEXISTENT) {
+            mboxlist_entry_free(&destmbentry);
+            r = IMAP_MAILBOX_EXISTS;
```
wolfsage, line 7884: This could mask other weird errors couldn't it? Maybe they'll all get log entries, but it's not clear.
Should we log something if r != IMAP_MAILBOX_EXISTS already?

imap/sync_support.c
```diff
@@ -3866,17 +3867,47 @@ int sync_apply_rename(struct dlist *kin, struct sync_state *sstate)
...
+    if (r) goto done;
+
+    if (mboxname_isusermailbox(oldmboxname, 1) &&
+        mboxname_isusermailbox(newmboxname, 1)) {
+        // we can't rename users if the new inbox already exists!
+        mbentry_t *destmbentry = NULL;
+        r = mboxlist_lookup_allow_all(newmboxname, &destmbentry, NULL);
+        if (r != IMAP_MAILBOX_NONEXISTENT) {
+            mboxlist_entry_free(&destmbentry);
+            r = IMAP_MAILBOX_EXISTS;
```
wolfsage, line 3879: (Same concern here about hiding other errors)

brong:
@wolfsage both your comments are fixed in https://github.com/cyrusimap/cyrus-imapd/pull/5181/commits/75509c3a31fe3753092fb046a4f4354255f8436a

@ksmurchison - the fixes above also fixed the Replication test!  Yep, looks like it was a bug before.

imap/imapd.c
```diff
@@ -8056,8 +8065,18 @@ static void cmd_rename(char *tag, char *oldname, char *newname, char *location,
...
     if (r) {
         prot_printf(imapd_out, "%s NO %s\r\n", tag, error_message(r));
-        // ensure the mboxlist entry gets fixed up
-        if (rename_user) mboxlist_update_full(mbentry, /*localonly*/1, /*silent*/1);
+        // ensure the mboxlist entry gets fixed up or removed
+        if (olddestmbentry) {
+            int r2 = mboxlist_update_full(olddestmbentry, /*localonly*/1, /*silent*/1);
+            if (r2)
+                xsyslog(LOG_ERR, "IOERROR: replacing old destination tombstone after rename error",
+                        "mboxname=<%s>, error=<%s>", olddestmbentry->name, error_message(r));
```
wolfsage, line 8073: In all four cases here should we also be logging error_message(r2) so we know why the cleanup failed?

`"mboxname=<%s>, error=<%s>, cleanuperror=<%s>", oldestmbentry->name, error_message(r), error_message(r2));`

brong: I was going to say yes, but... actually.  *sigh*.  error_message isn't re-entrant, you can't call it twice in a single syslog because the old one will get whacked, so I'm tempted to just do two syslogs, one for the `r` and one for the `r2`.

wolfsage (approved):
Thanks, this looks good to me

PR #5182, opened 2024-12-20
cyr_expire: batch deletes to 4096 at a time

brong:
As promised, here's a batching version of cyr_expire.  It can wind up calling both codepaths multiple times but I think that's OK, it'll be faster without changes each time.  Without that, we'd have to keep a cursor to say where to restart each time, and that's a more complex intervention.


ksmurchison (approved):
LGTM

PR #5184, opened 2024-12-23
xunlink: clear errno and return 0 on ENOENT

elliefm:
We almost always ignore this case anyway. Swallowing it here prevents some `syserror=<No such file or directory>` pollution in unrelated log lines.
    
With this change, if you DO care to differentiate between "file has been unlinked" and "file didn't exist anyway", you need to use unlink directly (and check/log errors yourself!)

Fixes some call sites to not do redundant logging or ENOENT checking

### DIFF
--- a/imap/append.c
+++ b/imap/append.c
@@ -1224,10 +1224,7 @@ EXPORTED int append_removestage(struct stagemsg *stage)
 
     while ((p = strarray_pop(&stage->parts))) {
         /* unlink the staging file */
-        if (xunlink(p) != 0) {
-            xsyslog(LOG_ERR, "IOERROR: error unlinking file",
-                             "filename=<%s>", p);
-        }
+        xunlink(p);
         free(p);
     }
 
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -5536,9 +5536,7 @@ EXPORTED void mailbox_archive(struct mailbox *mailbox,
                     // didn't manage to store it, so remove the ARCHIVED flag
                     continue;
                 }
-                r = xunlink (srcname);
-                if (r < 0)
-                    syslog(LOG_ERR, "unlink(%s) failed: %m", srcname);
+                xunlink(srcname);
             }
 #endif
             copyrecord.internal_flags |= FLAG_INTERNAL_ARCHIVED | FLAG_INTERNAL_NEEDS_CLEANUP;
--- a/lib/cyrusdb_quotalegacy.c
+++ b/lib/cyrusdb_quotalegacy.c
@@ -259,13 +259,8 @@ static int commit_subtxn(const char *fname, struct subtxn *tid)
         free(tid->fnamenew);
     } else if (tid->delete) {
         /* delete file */
-        r = xunlink(fname);
-        if (r == -1) {
-            xsyslog(LOG_ERR, "IOERROR: unlink failed",
-                             "fname=<%s>",
-                             fname);
+        if (xunlink(fname))
             r = CYRUSDB_IOERROR;
-        }
     } else {
         /* read-only txn */
     }
--- a/lib/xunlink.h
+++ b/lib/xunlink.h
@@ -49,9 +49,13 @@
  * N.B. These are NOT "error-handling" wrappers; you still must check the
  * return value, errno, etc, and handle errors accordingly!
  *
- * These wrappers present the same interface as the original functions,
- * but will also log an error if the original function did anything
- * other than succeed or fail-due-to-ENOENT
+ * These wrappers present almost the same interface as the original functions,
+ * except that:
+ *  1) errors are logged for you; and
+ *  2) ENOENT is treated as not an error: nothing is logged, errno is cleared,
+ *     and 0 is returned
+ *
+ * You cannot use these wrappers if you need to detect ENOENT failures.
  **/
 
 #define xunlink(pathname)                                                     \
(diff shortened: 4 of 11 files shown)


ksmurchison (approved):
I like it

rjbs (approved):
I am not reviewing the specific code, but we discussed this conceptual change and I agreed it was a good one.

imap/append.c
```diff
@@ -1224,10 +1224,7 @@ EXPORTED int append_removestage(struct stagemsg *stage)
 
     while ((p = strarray_pop(&stage->parts))) {
         /* unlink the staging file */
-        if (xunlink(p) != 0) {
```
wolfsage, line 1227: Here ENOENT would cause a log failure, now it doesn't. Are we sure we don't want to know if the file didn't exist in this case?

imap/mailbox.c
```diff
@@ -5536,9 +5536,7 @@ EXPORTED void mailbox_archive(struct mailbox *mailbox,
                     // didn't manage to store it, so remove the ARCHIVED flag
                     continue;
                 }
-                r = xunlink (srcname);
-                if (r < 0)
-                    syslog(LOG_ERR, "unlink(%s) failed: %m", srcname);
+                xunlink(srcname);
```
wolfsage, line 5539: Same here

lib/cyrusdb_quotalegacy.c
```diff
@@ -259,13 +259,8 @@ static int commit_subtxn(const char *fname, struct subtxn *tid)
         free(tid->fnamenew);
     } else if (tid->delete) {
         /* delete file */
-        r = xunlink(fname);
-        if (r == -1) {
-            xsyslog(LOG_ERR, "IOERROR: unlink failed",
-                             "fname=<%s>",
-                             fname);
```
wolfsage, line 266: Same here

lib/xunlink.h
```diff
@@ -52,6 +52,10 @@
  * These wrappers present the same interface as the original functions,
  * but will also log an error if the original function did anything
  * other than succeed or fail-due-to-ENOENT
+ *
+ * If the original function fails with ENOENT, errno will be reset to zero.
+ * This prevents errno leaks in the common case, but it means you can't use
+ * these wrappers if you need to detect ENOENT failures.
  **/
```
wolfsage, line 59: This should also be adjusted to say that the function will return 0 for ENOENT instead of -1

wolfsage (changes requested):
I'm confused by a few of the changes, it seems we'll no longer log on ENOENT when we would before. Is that really intentional?

imap/append.c
elliefm: You're correct about the changed behaviour, but I think the changed behaviour is what we really want here.

Like, I suspect the lack of an exception for ENOENT in this case originally was an oversight.  This suspicion is based on the fairly cavalier way we've used unlink elsewhere in the code.  I think the places with ENOENT exceptions were added reactively and individually after some excess logging was annoying, rather than systematically and deliberately.  In situations like this where the stage file probably always exists when we try to remove it, we wouldn't have had any log noise about it not existing to silence.

I guess the way to find out would be to see if we're currently ever logging this ENOENT in practice.  If we are, it hasn't annoyed us and maybe we do want it; if we aren't, that would explain why we never got annoyed enough to add an exception for it.

Though it's also notable that we don't record any kind of error when _any_ unlink failure happens here, we just log and move on.  This also makes me think we don't really care much in the moment whether the unlink succeeds or not, we just log in case we need to know what happened later if we find some file still lying around.  In the case of an ENOENT, there won't be any files still lying around to need explaining.

lib/xunlink.h
elliefm: Ahh yeah good catch!

imap/mailbox.c
elliefm: Same thinking as the earlier case -- we didn't care about any unlink failure except to log it, and we don't think ENOENT is worth logging because the end state is the state we were trying to reach by unlinking

lib/cyrusdb_quotalegacy.c
elliefm: This one is actually a meaningful behaviour change and not just a logging change, since we used to set `r = CYRUS_DBERROR` for ENOENT and with this change we'll ignore it.  I believe it's what we want though.

I guess this is another one where we could look for whether we ever log ENOENT here.  If it never happens in practice, then we never returned an error on this case, and perhaps the lack of an ENOENT exception was another oversight.  If we have been logging it, and haven't been annoyed by it, then maybe we do want it after all.  Perhaps we don't even use the quotalegacy db format, idk.

lib/xunlink.h
elliefm: I've rewritten this comment quite a bit, what do you think?

elliefm: It would actually be more in keeping with errno's usual behaviour to reset it to its previous value rather than 0 when ENOENT is detected (since a syscall that succeeds does not change errno at all).  I don't think it matters much in practice, but I'm overthinking the details now.

imap/append.c
wolfsage: Yeah I'm not sure. Maybe someone who knows these bits of code better can comment. It feels like "We're cleaning up staging files. The file *should* exist." and in later ones we're cleaning up other files that *should* exist also.

I won't block the review MR, I'm just... uncomfortable unless we're really sure this is right

wolfsage (approved):

PR #5186, opened 2024-12-23
Add xsyslog_ev function and related

mjdominus-cyrus:
Per https://linear.app/fastmail/issue/CYR-1407/produce-new-xsyslog-ev-function

Also:
* Adds a new test suite, `util.testc`
* Tests pass and `valgrind` is clean
* Includes some unrelated minor documentation fixes and other corrections

### DIFF
--- a/cunit/util.testc
+++ b/cunit/util.testc
@@ -2,7 +2,7 @@
 #include "util.h"
 #include <malloc.h>
 
-extern char *xsyslog_ev_escape_value(const char *);
+extern const char *__test_xsyslog_ev_escape_value(struct buf *);
 extern char __test_xsyslog_ev_guess_printf_escape(const char *);
 
 static void test_xsyslog_ev_fmt(void)
@@ -97,13 +97,14 @@ static void test_xsyslog_ev_escape_value(void)
      };
 
      unsigned i;
+     struct buf in = BUF_INITIALIZER;
      for (i=0; test[i]; i += 2) {
-         const char *in = test[i];
+         buf_setcstr(&in, test[i]);
          const char *expected = test[i+1];
-         char *actual = __test_xsyslog_ev_escape_value(in);
+         const char *actual = __test_xsyslog_ev_escape_value(&in);
          CU_ASSERT_STRING_EQUAL(actual, expected);
-         free(actual);
      };
+     buf_free(&in);
 }
 
 static void test_xsyslog_ev_guess_printf_escape(void)
@@ -163,9 +164,10 @@ static void test_xsyslog_ev_guess_printf_escape(void)
         const char *in = test[i];
         const char *expected = test[i+1];
         char actual = __test_xsyslog_ev_guess_printf_escape(in);
-        if (actual != *expected) fprintf(stderr, "%s -> %c ? s/b %c \n", in, actual, *expected);
+        if (actual != *expected) fprintf(stderr, "%s -> %c ? s/b %c (%hd) \n", in, actual, *expected, *expected);
         CU_ASSERT_EQUAL(actual, expected[0]);
     }
 }
+
 
 
--- a/cunit/util.testc
+++ b/cunit/util.testc
@@ -1,7 +1,6 @@
 #include "cunit/cyrunit.h"
 #include "util.h"
-
-extern char *__test_xsyslog_ev_fmt(const char *, ...);
+#include <malloc.h>
 
 static void test_xsyslog_ev_fmt(void)
 {
@@ -9,18 +8,18 @@ static void test_xsyslog_ev_fmt(void)
     {
 	char *result;
 
-	result = __test_xsyslog_ev_fmt("event-name", NULL);
+	result = xsyslog_ev_fmt("event-name", NULL);
 	CU_ASSERT_STRING_EQUAL(result, "event=event-name");
 	free(result);
 
-	result = __test_xsyslog_ev_fmt(
+	result = xsyslog_ev_fmt(
 	       "event-name",
 	       "firstname", "%s", "Fred",
 	       NULL);
 	CU_ASSERT_STRING_EQUAL(result, "event=event-name firstname=Fred");
 	free(result);
 
-	result = __test_xsyslog_ev_fmt(
+	result = xsyslog_ev_fmt(
 	       "event-name",
 	       "firstname", "%s", "Fred",
 	       "lastname", "%s", "Flintstone",
@@ -29,7 +28,7 @@ static void test_xsyslog_ev_fmt(void)
 	free(result);
 
 	// now with escaping
-	result = __test_xsyslog_ev_fmt(
+	result = xsyslog_ev_fmt(
 	       "announce name",
 	       "name", "%s", "Fred Flintstone",
 	       NULL);
@@ -37,7 +36,7 @@ static void test_xsyslog_ev_fmt(void)
 	free(result);
 
 	// now with various formats
-	result = __test_xsyslog_ev_fmt(
+	result = xsyslog_ev_fmt(
 				"formats",
 				"strlt", "%-10s", "foo",
 				"strrt", "%10s", "bar",
@@ -57,7 +56,7 @@ static void test_xsyslog_ev_fmt(void)
 	free(result);
 
 	// common case: multiline text output
-        result = __test_xsyslog_ev_fmt(
+        result = xsyslog_ev_fmt(
 				"multiline",
 				"text", "%s", "I\r\nlike\r\npie!\r\n",
 				NULL
@@ -69,7 +68,7 @@ static void test_xsyslog_ev_fmt(void)
     // tests for weird __sep__ feature
     {
       char *result;
-      result = __test_xsyslog_ev_fmt(
+      result = xsyslog_ev_fmt(
 			      "sep-test",
 			      "__sep__", "!!!",
 			      "name", "%s", "Fred Flintstone",
@@ -80,6 +79,7 @@ static void test_xsyslog_ev_fmt(void)
     }
 }
 
+extern char *xsyslog_ev_escape_value(const char *);
 extern char *__test_xsyslog_ev_escape_value(const char *);
 
 static void test_xsyslog_ev_escape_value(void)
@@ -104,3 +104,4 @@ static void test_xsyslog_ev_escape_value(void)
 	free(actual);
     }
 }
+
--- a/cunit/util.testc
+++ b/cunit/util.testc
@@ -93,6 +93,7 @@ static void test_xsyslog_ev_escape_value(void)
          "\r",               "\"\\r\"",
          "\r\n",             "\"\\r\\n\"",
          "-12.345e67",       "-12.345e67",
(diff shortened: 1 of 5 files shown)


lib/util.c
```diff
@@ -2187,9 +2194,324 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
...
+ *   p   if the expected argument is a (void *)
+ *   -1  if it's something else
+ */
+static char xsyslog_ev_guess_printf_escape(const char *s) {
+    unsigned is_long = 0;
+ TOP:
+    while (*s && *s != '%') s++; // seek to first % sign or to end of string
+    if (*s == 0) return 0;       // String has no percent signs at all
+    s++;
+    if (*s == '%') { s++; goto TOP; } // it's "%%", skip it and continue
```
rsto, line 2387: nit: I know you asked before writing this code about when to and when to not use `goto`, so I won't request to change this now. But I find this goto here not only unnecessary but confusing. A `while` loop with a break or continue statement would make that code likely more simple to read.

As a rule of thumb: `goto` in the Cyrus codebase is perfectly fine for early function exits that require cleanup (e.g. the `done` label). In all other cases, they are the least preferred option.

rsto (changes requested):
I like improving the log format! It's also great we got proper unit tests for this.

Apart from the few items I found in the code, I wonder:
Is there a canonical specification for logfmt somewhere? I see our tests don't assert how empty strings in the key or value are handled, would that be the empty string or `""`?

I'm a bit wary of the magic printf formatter guessing, but it looks well thought out so let's see how it turns out in practice.

One other thing that stands out to me are the commit message: in Cyrus, we typically prefix the commit message with the filename that mostly got changed in this commit (e.g. `util.c`). Also, our commit messages almost always start with a verb.

rjbs:
@rsto `logfmt` is, in my experience, insufficiently documented in the wild.  The most cited source is https://brandur.org/logfmt which I find useful but … well, it wouldn't make a good I-D.

I thought we'd improved the clarity of this in the Fastmail internal spec, but I'm wrong.  We just use Perl's Log::Fmt, which I wrote.  In that, the value must have length, so `x=""` and not `x=`.  (It serializes Perl's `undef` value as `~missing~`.)

lib/util.c
```diff
@@ -2187,9 +2194,324 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
...
+    case 'f': (void) va_arg(ap, double);   break;
+    case 's': (void) va_arg(ap, char *); break;
+    case 'p': (void) va_arg(ap, void *); break;
+    case 0:
+        // we could handle this by skipping the va_arg call entirely, which would allow the user to write
+        //     xsyslog_ev(..., key, "foo", ...)
+        // in place of
+        //     xsyslog_ev(..., key, "foo", DUMMY, ...)
+        // but that seems morelikely to be confusing than otherwise
+    case -1:
```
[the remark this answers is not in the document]
mjdominus-cyrus: I can do this.  Would it also work to declare the function as returning `signed char`?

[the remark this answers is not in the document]
mjdominus-cyrus: Also, it's not clear to me that the -1 return is ever useful, and I considered removing it, since the current code doesn't distinguish it from 0. But my heart told me that it was reasonable to distinguish "unparseable" from "found no escape sequences" even if no current code cared about it.

mjdominus-cyrus:
The new version has exactly the same commits as before, except:
* I redid all the commit messages
* There are three new commits on the end to address the `case -1`, the `goto`, and the escaped version of the empty string
*

lib/util.c
```diff
@@ -2187,9 +2194,332 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
...
+ *   1. Format a key-value list for syslogging
+ *   2. Allocate a string to hold the result
+ *   3. The CALLER is responsible for freeing this string
+ *   4. Returns NULL if something goes wrong
+ *
+ * EXAMPLE:
+ *
+ *     xsyslog_ev_fmt("earthquake",
+ *                    "location", "%s", "Dubuque",
+ *                    "richter", "%5.1f", 7.9)
```
wolfsage, line 2236: This doesn't document the final necessary NULL, we should in all of the examples, etc...

lib/util.c
```diff
@@ -2187,9 +2194,332 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
...
+ *
+ * will return the string
+ *
+ *     event=earthquake location=Dubuque richter="  7.9"
+ *
+ * char *s = xsyslog_ev_fmt(event_name,
+ *                          key1, fmt1, val1,
+ *                          key2, fmt2, val2,
+ *                          ...);
```
wolfsage, line 2245: Maybe `..., NULL);`

lib/util.c
```diff
@@ -2187,9 +2194,332 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
...
+ *
+ * A sequence of chunks is built, each with the form "key_n=escaped_formatted_val_n"
+ *
+ * The chunks are concatenated with single spaces in between.
+ *
+ * The resulting string is returned.  The CALLER is responsible for freeing it.
+ *
+ * WARNING:
+ *   the function only supports a subset of sprintf-style escapes.
+ *   in particular, each format string MUST contain EXACTLY ONE %-escape.
```
wolfsage, line 2262: This says MUST but that restriction isn't enforced. It might be a good idea to fatal() if the formats have more than one %... specification

lib/util.c
```diff
@@ -2187,9 +2194,332 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
...
+     * and then the result will be escaped (if necessary)
+     *
+     * finally, the resulting string "key=escaped-formatted-value"
+     * will be appended to the log buffer
+     */
+    while ((key = va_arg(ap, const char *))) {
+        // special case: __sep__ means reconfigure the separator
+        if (!strcmp(key, "__sep__")) {
+            sep = va_arg(ap, const char *);
+        }
```
elliefm, line 2317: Is this actually useful?  Surely if we're going to all this effort it's because we want our logs to be in a consistent format

lib/util.c
```diff
@@ -2187,9 +2194,332 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
...
+    case 0:
+        // we could handle this by skipping the va_arg call entirely, which would allow the user to write
+        //     xsyslog_ev(..., key, "foo", ...)
+        // in place of
+        //     xsyslog_ev(..., key, "foo", DUMMY, ...)
+        // but that seems more likely to be confusing than otherwise
+    default:
+        // This seems extreme at first but the alternative is to leave ap pointing to the wrong argument,
+        // and that is already undefined behavior, likely a segfault, so this is really the best we can do
+        abort();
```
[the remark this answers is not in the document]
elliefm: Yep, `fatal()` is the thing to do here.

lib/util.c
```diff
@@ -2187,9 +2194,332 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
...
+    }
+}
+
+/* Quick-and-dirty parser for SINGLE, COMMON printf-type escape strings
```
elliefm, line 2368: If we have to have our own parser for printf escape sequences, have we not lost all value we were going to gain by using them in the first place?

lib/util.c
```diff
@@ -2187,9 +2194,332 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
...
+ * return 0 if string contains no %-escapes
+ * otherwise return:
+ *   d   if the expected argument is an int
+ *   l   if the expected argument is a long
+ *   f   if the expected argument is a double
+ *   s   if the expected argument is a (char *)
+ *   p   if the expected argument is a (void *)
+ *   -1  if it's something else
+ */
+static char xsyslog_ev_guess_printf_escape(const char *s) {
```
elliefm, line 2380: This should probably return `int` rather than `char`.  It's fine for it to be logically a single character still, but it's usual in C to use the `int` type for single characters, for reasons I don't quite remember, but which might be related to the implicit type promotion rules and/or the fact that the signedness of `char` is platform-dependent.

See for example the `int c` argument to `char *strchr(const char *s, int c);`

elliefm (commented):
My review is mostly a handful of style nits and advicey comments.

Bigger picture, I don't like the design; but given the design, this seems like a solid implementation of it.  If we're considering this a proof of concept, I'd say it proves I don't like the design. :)

cunit/util.testc
```diff
@@ -0,0 +1,174 @@
...
+        char *result;
+
+        result = xsyslog_ev_fmt("event-name", NULL);
+        CU_ASSERT_STRING_EQUAL(result, "event=event-name");
+        free(result);
+
+        result = xsyslog_ev_fmt(
+               "event-name",
+               "firstname", "%s", "Fred",
+               NULL);
```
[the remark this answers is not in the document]
elliefm: Further into the review, I saw you formatted this sort of thing elsewhere in the way I suggest here, but I forgot to come back and update this comment.

rsto:
> We just use Perl's Log::Fmt, which I wrote. In that, the value must have length, so `x=""` and not `x=`. (It serializes Perl's `undef` value as `~missing~`.)

Requiring a non-zero length value sounds sane to me, we should do the same in the Cyrus implementation.

lib/util.c
ksmurchison: Yeah.  I thought @rjbs had suggested that rather than using a printf-style format string, we just use an enum like [CHAR, STRING, INT, FLOAT] and then convert those into format strings

cunit/util.testc
```diff
@@ -0,0 +1,174 @@
...
+         "foo bar",          "\"foo bar\"",
+         "call me \"fred\"", "\"call me \\\"fred\\\"\"",
+         "\n",               "\"\\n\"",
+         "\r",               "\"\\r\"",
+         "\r\n",             "\"\\r\\n\"",
+         "-12.345e67",       "-12.345e67",
+         "",                 "\"\"",
+         NULL,
+     };
+
```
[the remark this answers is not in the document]
mjdominus-cyrus: I see little benefit either way.

lib/util.c
mjdominus-cyrus: Probably not. I only threw it in as a possibly interesting suggestion.

YAGNI says to eliminate it and its associated tests.

lib/util.c
mjdominus-cyrus: Rik and I discussed that.  i think I persuaded him that this is better, because:

1. You can write `"%5.2f"` and `"%-20s"` and other possibly useful formats
2. Why use and maintain a set of enum constants that mean `"%s"` and `"%d"` when you can just use self-documenting `"%s"` and `"%d"`

mjdominus-cyrus:
Today is my last day in this job.  Rather than make further revisions to this code, I will leave it to you folks to decide what to do with it, if anything.

rsto:
Closing this in favor of https://github.com/cyrusimap/cyrus-imapd/pull/5241

PR #5187, opened 2024-12-24
jmap_ical: use verbatim RELTYPE parameter value

rsto:
Recent libical icalparameter_get_xvalue returns NULL if the parameter value actually isn't an x-enum value. This patch changes reading RELTYPE as either an IANA-enum or X-enum value, rather than special-casing a few known relation types when setting the relatedTo property in JSCalendar.


rsto:
Testing is done by passing the following existing tests again:

- JMAPCalendars.calendarevent_get_relatedto
- JMAPCalendars.calendarevent_set_relatedto

ksmurchison (approved):

PR #5188, opened 2024-12-27
index: parse '$' in vanished as well

brong:
As found, the QRESYNC vanished version of fetch '$' failed

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -426,7 +426,7 @@ EXPORTED int index_expunge(struct index_state *state, const char *sequence,
             continue; /* no \Deleted flag */
 
         /* if there is a sequence list, check it */
-        if (sequence && !seqset_ismember(seq, im->uid))
+        if (seq && !seqset_ismember(seq, im->uid))
             continue; /* not in the list */
 
         /* load first once we know we have to process this one */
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):
The code LGTM

ksmurchison (changes requested):
We should probably add a test using the example that caused the crasher

imap/index.c
```diff
@@ -1211,11 +1211,11 @@ EXPORTED int index_fetch(struct index_state *state,
     r = index_lock(state, readonly);
     if (r) return r;
 
-    if (!strcmp("$", sequence)) {
-        seq = state->searchres;
+    seq = _parse_sequence(state, sequence, usinguid);
+    if (!strcmpsafe("$", sequence)) {
         usinguid = 1;
```
wolfsage, line 1216: After we've called _parse_sequence, we're setting usinguid to 1. But _parse_sequence has different behaviour depending on usinguid. Does that matter?

ksmurchison: It shouldn't, as `_parse_sequence()` will return immediately if `sequence == "$"`.  I don't believe this changes the existing behavior and the Cass tests still pass, assuming that we have proper coverage.

wolfsage (approved):

brong:
> We should probably add a test using the example that caused the crasher

Done!

ksmurchison (approved):

PR #5189, opened 2024-12-30
configure.ac: check for cyruslibs dependencies also without pkg-config

rsto:
This updates configure.ac to fall back checking for the cld2 and wslay dependencies even if pkg-config could not find them. These dependencies have now been removed from the https://github.com/cyrusimap/cyruslibs repository and their upstream version does not support pkg-config. This patch allows detecting them in the standard include and library paths.


ksmurchison (approved):
WFM

PR #5191, opened 2025-01-03
docs: strip out custom theming and use stock RTD

rjbs:
This means that the box building the docs will require python3-sphinx-rtd-theme (or however else you get that theme).

**Update**:  Okay, I got a *little* carried away here, but I've got the docs building clean, with no errors or warnings.  This makes life better because any new warning or error is now worth looking at.  Also, this might get us ready to use Sphinx v8, which will be in Debian Trixie, likely out in mid-2025.


rjbs:
When approved, I will immediately backport this to maint branches and update the system doing the automated builds to Debian bookworm's default sphinx.

rjbs:
Woo, with these changes, the site also builds nicely in Sphinx v8, which we'll get on Trixie next year.  (We could use a special-purpose site builder image for that now, but I don't think we need anything new.)

docsrc/conf.py
```diff
@@ -163,7 +163,7 @@
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
-html_theme_options = { 'github_version': 'master'}
+html_theme_options = { }
```
elliefm, line 166: Ooooooh interesting, what consequence does this discarding this detail have?  The stable branches all have `'github_version': 'cyrus-imapd-3.10'` or whatever here.

docsrc/imap/download/release-notes/3.0/index.rst
```diff
@@ -8,7 +8,6 @@ Cyrus IMAP 3.0 Releases
     :maxdepth: 1
     :glob:
 
-    x/?.?.*-beta*
-    x/?.?.*-rc*
-    x/?.?.?
-    x/?.?.??
+    x/*-beta*
+    x/*-rc*
+    x/*
```
elliefm, line 13: Does newer sphinx order these sensibly?  sphinx 3.x sorts globs in ascii order, which leads to releases being listed like:
* 3.x.1
* 3.x.10
* 3.x.2

The fiddlier listings here are meant to force the releases to be listed in release order, and we (I) ignore the warnings for the empty two-digit expansions on branches that haven't had a two-digit minor release yet.

docsrc/imap/download/release-notes/index.rst
```diff
@@ -107,6 +107,8 @@ Development snapshots
 .. toctree::
     :maxdepth: 1
 
+    3.11/index
+    3.9/index
```
elliefm, line 111: Oh, this'll need `3.13/index` now too, guess I missed it when doing the 3.12 fork the other week.

elliefm (commented):
Couple of points for consideration where you've removed or changed things that are (or used to be) load-bearing.  Though they might just not be load-bearing anymore, idk!

elliefm:
Forgot to comment on this in my earlier review because it didn't come up in the diff:

> building the docs will require python3-sphinx-rtd-theme (or however else you get that theme)

We'll need a configure check for this then, like we have for pod::pom:view::restructured and the other doc build dependencies.

elliefm:
Can you rebase this on current upstream master? Since there's a bunch of doc changes on master following the 3.12 fork, and there might be stuff in my changes that'll warn/etc with newer sphinx.

docsrc/imap/download/release-notes/3.0/index.rst
rjbs: Discussion on Slack suggested that these are still sorting as we'd like.

Also, I said that I'd like to do some future work to replace the globbing with exact links.  Then, since we'll normally be warning free, we can notice warnings that arise from "document not included".

docsrc/imap/download/release-notes/index.rst
rjbs: Done, post rebase.

rjbs:
> Forgot to comment on this in my earlier review because it didn't come up in the diff:
> 
> > building the docs will require python3-sphinx-rtd-theme (or however else you get that theme)
> 
> We'll need a configure check for this then, like we have for pod::pom:view::restructured and the other doc build dependencies.

ellie agreed to do this, for which I've made a separate issue.

docsrc/conf.py
rjbs: I believe this is now redundant to the `html_context` configuration, which is already present in conf.py.  I've made internal issue CYR-1481 to investigate this and fix if needed

rjbs:
The level of approval here was unclear (to me), but I've addressed a few things and asked @wolfsage to give another round of review to keep this moving.  I'll catch up with you later so we can make sure we're both happy.

wolfsage:
The commit "[docs: strip out custom theming and use stock RTD](https://github.com/cyrusimap/cyrus-imapd/pull/5191/commits/97f66c4782f1c95bf7836eb05982d1071534ca3c)" doesn't say why we're doing this - what's the reason to drop the custom theme?

wolfsage:
Seems good! It makes search work again (who knew it was broken?!)

I don't love the yellow background of code blocks or the pages only taking up half of my wide screen vs the old style, but I can live with it

wolfsage (approved):

PR #5193, opened 2025-01-03
twoskip: free transaction after repack

brong:
This mirrors the code in commit/abort after unlocking.  We're about to replace *db with data from the new DB, so free up the memory.


ksmurchison (commented):
Cassandane is unhappy with this change

brong:
Yeah wow, clearly it's not safe!  Interesting.  Something must refcount it in interesting ways.  Just gonna dump this, it's not a big deal.

brong:
Oh yeah - it can be called with or without a transaction :(  I had to add local transaction tracking!

ksmurchison (approved):

PR #5194, opened 2025-01-03
http_caldav_sched.c: fix for organizer getting stomped in sched_request()

ksmurchison:
When the organizer gets stomped it can result in binary data being used when building the JSON structure which results in an 'Invalid UTF-8 string' error.

The stomping of the organizer value seems consistent if the iTIP request contains a large number of ATTENDEEs, but the JSON failure is random depending on what data the stomped organizer contains.

So, this is a hard problem/fix to write a test for.

### DIFF
--- a/cassandane/tiny-tests/Caldav/invite_lots_of_attendees
+++ b/cassandane/tiny-tests/Caldav/invite_lots_of_attendees
@@ -1,0 +1,89 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_invite_lots_of_attendees
+    :VirtDomains :min_version_3_0
+{
+    my ($self) = @_;
+
+    my $service = $self->{instance}->get_service("http");
+    my $CalDAV = Net::CalDAVTalk->new(
+        user => "cassandane%example.com",
+        password => 'pass',
+        host => $service->host(),
+        port => $service->port(),
+        scheme => 'http',
+        url => '/',
+        expandurl => 1,
+    );
+
+    my $CalendarId = $CalDAV->NewCalendar({name => 'hello'});
+    $self->assert_not_null($CalendarId);
+
+    my $uuid = "6de280c9-edff-4019-8ebd-cfebc73f8201";
+    my $href = "$CalendarId/$uuid.ics";
+    my $card = <<EOF;
+BEGIN:VCALENDAR
+VERSION:2.0
+PRODID:-//Apple Inc.//Mac OS X 10.10.4//EN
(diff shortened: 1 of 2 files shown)


imap/http_caldav_sched.c
```diff
@@ -725,22 +725,31 @@ static int imip_send(const char *cal_ownerid, const char *sched_userid,
...
+                               "method", imip_method,
+                               "id", id,
+                               "ical", ical_str,
+                               "jsevent", jsevent,
+                               "patch", patch,
+                               "is_update", SCHED_IS_UPDATE(sched_data),
+                               "schedulingMechanism",
+                               sched_mechanisms[sched_data->mech]);
+    if (!val) {
+        xsyslog(LOG_ERR, "failed to create iMIP notification message",
```
wolfsage, line 743: Should we set r to something in this case so the error bubbles up, or no? We don't above if  icalcomponent_as_ical_string fails, so maybe that's okay?

wolfsage (approved):

imap/http_caldav_sched.c
ksmurchison: Yeah, I'll do that

wolfsage: (We should probably add a log line for when icalcomponent_as_ical_string fails too)

imap/http_caldav_sched.c
```diff
@@ -720,27 +720,40 @@ static int imip_send(const char *cal_ownerid, const char *sched_userid,
 
     /* Don't send a bogus message - check late to not allocate our own copy */
     const char *ical_str = icalcomponent_as_ical_string(sched_data->itip);
-    if (!ical_str) goto done;
+    if (!ical_str) {
+        r = IMAP_INTERNAL;
```
wolfsage, line 724: Should we log something here or do you think the r of IMAP_INTERNAL will be enough?

wolfsage (approved):

cassandane/tiny-tests/Caldav/invite_lots_of_attendees
```diff
@@ -0,0 +1,89 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_invite_lots_of_attendees
```
rsto, line 4: Can this test deterministically reproduce the issue? If not, maybe we should get rid of it, as it will only make the tests run longer?

rsto (approved):
LGTM - but the underlying memory management issue in libical is a real pain.

PR #5195, opened 2025-01-03
cassandane: Do syntax checks only for changed files

wolfsage:
Instead of `make` inside of cassandane always re-checking syntax, only check syntax if one of the files has actually changed.

This is more in the spirit of make with only rebuilding things on changes, and also makes `cyd test` much faster...

### DIFF
--- a/cassandane/Makefile
+++ b/cassandane/Makefile
@@ -63,18 +63,11 @@ ifndef NOCYRUS
     CYRUS_PERL_PATHS := $(shell $(PERL) utils/cyrus-perl-paths.pl)
 endif
 
-SYNTAX_rules =
+syntax: .cassandane-syntax-check
 
-define SYNTAX_template
- $(1)_syntax: $(1)
-	@$(PERL) $(CYRUS_PERL_PATHS) -c $(1)
- SYNTAX_rules += $(1)_syntax
-endef
+.cassandane-syntax-check: $(SCRIPTS) $(MODULES)
+	rm -f .cassandane-syntax-check
+	$(foreach path,$?,$(PERL) $(CYRUS_PERL_PATHS) -c $(path) || exit;)
+	touch .cassandane-syntax-check
 
-$(foreach s,$(SCRIPTS),$(eval $(call SYNTAX_template,$(s))))
-
-$(foreach m,$(MODULES),$(eval $(call SYNTAX_template,$(m))))
-
-syntax: $(SYNTAX_rules)
-
-.PHONY: all syntax $(SYNTAX_rules)
+.PHONY: syntax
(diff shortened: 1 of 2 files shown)


wolfsage:
Mmm, this won't pick up on changes in tiny-tests. Tabling for now.

elliefm:
IIRC it was a deliberate choice on my part to have it syntax check everything, because if you only check the changed modules, you won't find errors in dependent ones that haven't been updated to match.  For this to really work in a make-like way, make would need to know the perl module dependency graph, and it doesn't just have that already because it's not usually needed.

If there's syntax errors, there'll be test failures for them anyway.  But a test failure adds a lot of noise which makes it harder to diagnose and repair the syntax error when you discover it that way.  So we front load a syntax check into make, and that way if there's any syntax errors we can see them clearly, without the delay and layers of distraction.  But the syntax check isn't necessary, just a development cycle convenience.

So really, the `syntax` target doesn't need to be included in `all`.  It's like that cause that's how it has been, but it wouldn't be a problem for it to stand alone and require an explicit `make syntax`.  A developer knows when they've made changes, and can ask for the syntax check if they want it.

elliefm:
I guess a middle point would be "if any perl file has changed, syntax check all the perl files" (but if none have changed, skip the syntax check).  `make` would need to be able to find "all the perl files" (because we don't want to have to maintain a list of them), but that's easy enough given that we're already wedded to GNU make syntax anyway.

cassandane/Makefile
```diff
@@ -63,18 +63,11 @@ ifndef NOCYRUS
...
+syntax: .cassandane-syntax-check
 
-define SYNTAX_template
- $(1)_syntax: $(1)
-	@$(PERL) $(CYRUS_PERL_PATHS) -c $(1)
- SYNTAX_rules += $(1)_syntax
-endef
+.cassandane-syntax-check: $(SCRIPTS) $(MODULES)
+	rm -f .cassandane-syntax-check
+	$(foreach path,$?,$(PERL) $(CYRUS_PERL_PATHS) -c $(path) || exit;)
```
elliefm, line 70: This implementation, where the foreach is inside a rule, serialises the checks.  We can no longer benefit from parallel make (`make -j4` or whatever).

The purpose of the odd-looking `SYNTAX_template` design, where the foreach calls define rules rather than performing actions, is so that we end up with one rule per file.  This means a parallel make can run a bunch of the checks at a time.  This is a significant speedup vs running them serially:

```
ellie@debian:master:~/fastmail/cyrus-imapd/cassandane$time make
...
real	0m54.897s
user	0m47.373s
sys	0m6.841s
ellie@debian:master:~/fastmail/cyrus-imapd/cassandane$time make -j4
...
real	0m32.507s
user	0m51.438s
sys	0m8.984s
```

wolfsage: Yeah I discovered that a bit later :)

wolfsage:
Thanks for all of the comments and background, good to know!

PR #5196, opened 2025-01-07
GitHub Action: import updated workflow from master

rjbs:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,7 @@
 name: Cyrus IMAP CI
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - master
@@ -15,7 +16,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     container:
-        image: cyrusimapdocker/cyrus-buster:latest
+        image: ghcr.io/cyrusimap/cyrus-docker:bookworm-libs56
         options: --sysctl net.ipv6.conf.all.disable_ipv6=0 --init
     steps:
     - uses: actions/checkout@v4
@@ -24,9 +25,6 @@ jobs:
     - name: install missing or frequently-updated deps
       shell: bash
       run: |
-        cpanm IO::File::fcntl
-        cpanm Digest::CRC
-        cpanm XML::Simple # XXX could be grabbed from apt by docker image
         cpanm Mail::IMAPTalk # in image, but fetch latest!
     - name: setup git safe directory
       shell: bash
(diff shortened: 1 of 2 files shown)


rjbs:
So, we talked about backporting the changes to CI, but it isn't straightforward.  We've "always" used a single Docker container as the basis for testing, with a single version of cyruslibs.  I think here, we reach a point where master needs cyruslibs-v60 (or something) and older branches won't build with it.  We can build versions with older cyruslibs and leave them be, without updating them every day.  Is that the preferred plan here?

elliefm:
Hmmm.

So, I think the reason we have cyruslibs at all is because the bleeding edge Cyrus usually wants libraries or library features that aren't necessarily available in distro packages.

But if the docker image is based on a reasonably current distro, and we're trying to run CI for an older release branch (say, 3.6 and earlier), perhaps it's sufficient to use distro packages for the dependencies rather than cyruslibs.  And then it doesn't matter that the older code can't build with current cyruslibs.

But there's an awkward spot in between, because we don't coordinate Cyrus releases with Debian releases.  3.12 certainly, 3.10 probably, and perhaps even 3.8 will depend on stuff that Bookworm doesn't have, and will thus need cyruslibs.  But if cyruslibs moves out from under them and they no longer build with the latest cyruslibs, uh oh.

Do we need to start versioning the public cyruslibs repo?  We already version our FM builds of it when doing internal deploys, and for much the same reason, so maybe the answer is something like:

1. we start routinely pushing cyruslibs version tags to the cyrusimap/cyruslibs repo (and perhaps some historical ones, especially the last one before things started breaking older builds)
2. we keep a file around on each cyrus-imapd-version branch indicating which cyruslibs version tag it wants to use for CI
3. we make CI build the appropriate cyruslibs for whatever is being tested (always latest for master)
4. presumably the finite set of required-older-cyruslibs-versions doesn't change much, so maybe it's worthwhile prebuilding them in the image, rather than building them hot each time CI runs? but then, we also don't run CI on those older branches very often, so there's an image size vs run time tradeoff there which I don't have an opinion about

elliefm:
5. a new step in the annual major release process will be to add a ci-requires-cyruslibs-version file (or whatever we call it) to the new release branch with whatever the current cyruslibs version is at the time we do this step

rjbs:
We already publish (to the world) tags for versioned cyruslibs.  I reckon what I'd do next is create a one-off version of the cyrus-docker image for the last few versions (known to pass, anyway) of cyrus, and make their CI branches use those.

rjbs:
Replaced by #5200 because of bad branch naming on my part.

PR #5197, opened 2025-01-07
ical_support.c: bump refcount of rrules in multirrule iterator

rsto:
Fixes a memory leak that got introduced by libical's recent change to rrule-by-ref.

Run cunit tests with valgrind to reproduce the issue:

```
make VG=1 check-ical_support
```

### DIFF
--- a/imap/ical_support.c
+++ b/imap/ical_support.c
@@ -516,7 +516,11 @@ static void multirrule_iterator_add(struct multirrule_iterator *iter,
     entry->icaliter = icaliter;
     entry->next = icalrecur_iterator_next(entry->icaliter);
     entry->recur = recur;
+#ifdef HAVE_RECUR_BY_REF
+    icalrecurrencetype_ref(recur);
+#else /* !HAVE_RECUR_BY_REF */
     entry->recur->rscale = xstrdupnull(recur->rscale);
+#endif /* HAVE_RECUR_BY_REF */
 }
 
 static icaltimetype multirrule_iterator_next(struct multirrule_iterator *iter)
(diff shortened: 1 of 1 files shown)


ksmurchison (commented):
Do we actually want these #ifdef protected?  Won't we leak recur->rscale in the non-by-ref case?

rsto:
I don't have an old-enough libical around to check this. Is the unref function also available without recur-by-ref

ksmurchison:
> I don't have an old-enough libical around to check this. Is the unref function also available without recur-by-ref

Yes, I defined one in ical_support.c:3299

imap/ical_support.c
```diff
@@ -516,7 +516,7 @@ static void multirrule_iterator_add(struct multirrule_iterator *iter,
     entry->icaliter = icaliter;
     entry->next = icalrecur_iterator_next(entry->icaliter);
     entry->recur = recur;
-    entry->recur->rscale = xstrdupnull(recur->rscale);
+    icalrecurrencetype_ref(recur);
```
ksmurchison, line 519: Hmm.  I misread this initially as icalrecurrencetype_unref().  We don't have a non-recur-by-ref version of this function.  We could make a macro which does nothing or #ifdef protect this.

rsto: alright. I'll re-add the unref then.

rsto: I mean, I'll re-add guarding the unref then.

elliefm:
I might have an old enough cyuslibs still, I'll pull this branch and see what happens

elliefm:
Is there something I can tell you about my cyruslibs install that will confirm whether it's old enough?  Oh actually (happening as I type), here's a thing -- ical_support no longer building for me:

```
imap/ical_support.c: In function 'multirrule_iterator_add':
imap/ical_support.c:519:5: error: implicit declaration of function 'icalrecurrencetype_ref'; did you mean 'icalrecurrencetype_unref'? [-Werror=implicit-function-declaration]
  519 |     icalrecurrencetype_ref(recur);
      |     ^~~~~~~~~~~~~~~~~~~~~~
      |     icalrecurrencetype_unref
  CC       imap/libcyrus_imap_la-webdav_db.lo
cc1: all warnings being treated as errors
```

ksmurchison:
@elliefm Eithwr we need to #ifdef protect that call of create an empty macro for that function

rsto:
Now compiles and tests/valgrinds cleanly with both recur-by-ref and recur-by-value libical.

ksmurchison (approved):
If it works and is memory clean, WFM

PR #5198, opened 2025-01-07
configure.ac: want sphinx_rtd_theme for doc builds

elliefm:
Now that we're using the "sphinx_rtd_theme" them rather than a custom one (#5191), we need configure to notice and report when the dependency is missing, like the other documentation dependencies.


rjbs (approved):
I am approving this based on (a) trusting you and (b) sort of squinting at other bits of configure.ac.  This stuff looks *weird* and alien to me.  I will try to learn more of it.

elliefm:
For backporting, this is easier if #5148 is backported first

PR #5199, opened 2025-01-08
Remove experimental zeroskip database backend

rsto:
It never was ready for production use, is not distributed anymore in cyruslibs and is not developed further.


cunit/aaa-db.testc
```diff
@@ -190,15 +190,6 @@ static char *make_basedir(const char * const *reldirs)
...
 {
-    /* cunit.pl doesn't play nice with #ifdef'ed CUNIT_PARAMS */
-
-    if (!strcmp(backend, "zeroskip")) {
-#ifdef HAVE_ZEROSKIP
-        return 0;
-#else
-        return 1;
-#endif
-    }
```
elliefm, line 201: I think you could remove the whole `skiptest()` function and all calls to it, it only existed because of zeroskip

cunit/libconfig.testc
```diff
@@ -994,20 +994,6 @@ static void test_deprecated_stringlist(void)
...
-        "configdirectory: "DBDIR"/conf\n"
-        "tlscache_db: sql\n"
-        "tls_sessions_db: zeroskip\n"
-    );
-    CU_ASSERT_SYSLOG(/*all*/0, 1);
-
-    /* should read new value at the new name */
-    val = config_getstring(IMAPOPT_TLS_SESSIONS_DB);
-    CU_ASSERT_PTR_NOT_NULL(val);
-    CU_ASSERT_STRING_EQUAL(val, "zeroskip");
```
elliefm, line 1010: We need to keep this part of the test.  The point is to make sure the config parser behaves correctly with deprecated a STRINGLIST.  It just so happens that tlscache_db/tls_sessions_db is an example of such a thing, which we're using to exercise the behaviours we're testing.

Instead of removing this chunk, it should be fine to use the value "twoskip" instead of "zeroskip".  "twoskip" is a valid value for these options, and is different from "sql", so the interesting behaviour will still be exercised.

lib/imapoptions
```diff
@@ -217,7 +217,7 @@ are listed with ``<none>''.
 /* Alternative INBOX spellings that can't be accessed in altnamespace
    otherwise go under here. */
 
-{ "annotation_db", "twoskip", STRINGLIST("skiplist", "twoskip", "zeroskip"), "3.1.6" }
+{ "annotation_db", "twoskip", STRINGLIST("skiplist", "twoskip"), "3.1.6" }
```
elliefm, line 220: Need to replace the version number with "UNRELEASED".  This applies to all the other foo_db option changes too, except for backup_db which I'll address separately

lib/imapoptions
```diff
@@ -549,7 +549,7 @@ Blank lines and lines beginning with ``#'' are ignored.
 { "backup_retention", "7d", DURATION, "3.12.0", "3.12.0" }
 /* Deprecated. No longer used. */
 
-{ "backup_db", "twoskip", STRINGLIST("skiplist", "sql", "twoskip", "zeroskip"), "3.12.0", "3.12.0" }
+{ "backup_db", "twoskip", STRINGLIST("skiplist", "sql", "twoskip"), "3.12.0", "3.12.0" }
```
elliefm, line 552: This case is interesting, because the option is already deprecated (since 3.12.0), but we're changing it again (by necessity I think).

I think perhaps you should change the first version number (which is the `last_modified` field) to "UNRELEASED", to reflect that it has been modified; but leave the second version number (the `deprecated_since` field) as "3.12.0", since that was when it was deprecated and that fact hasn't changed.

If that doesn't work, then I guess leave both version numbers as "3.12.0".

elliefm (changes requested):

lib/imapoptions
rsto: done. please recheck.

lib/imapoptions
rsto: done. please recheck.

ksmurchison (approved):
LGTM

elliefm (approved):

PR #5200, opened 2025-01-08
import new testing from master branch to 3.10

rjbs:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,6 +1,7 @@
 name: Cyrus IMAP CI
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - master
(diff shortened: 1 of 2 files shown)


rjbs:
@elliefm So, with a newly built bookworm+cyruslibs-v56, this gets much closer to done.  I don't know what the problem here might be.  Any thoughts before I blindly try an older cyruslibs?

elliefm:
None of those errors mean anything to me, my first guess would be a cyruslibs incompatibility too.  For whatever it's worth, my local cyruslibs seems to be commit 8670334070cf0e8f231093dda408ad0ab7bf78e9 from December 2022, and 3.10 builds/tests fine with that.

rjbs:
I had build images with v50 - v60 and none tested to completion.

The previous automation uses the image cyrusimapdocker/cyrus-buster which has cyruslibs at a853b0dc1d72c04cd177cdf9846691f97d6f216e.  That'll be my next try.

rjbs:
@elliefm Heyyyy, this seems to build with libs-v42.  I'm going to set up branches for older versions, but we may have a winner here.

rjbs:
I think this can be applied at will.  Future CI backports will likely be simpler after this one.

elliefm (approved):

PR #5204, opened 2025-01-09
update GitHub workflow to use cyd

rjbs:


.github/workflows/main.yml
```diff
@@ -40,28 +40,18 @@ jobs:
...
-      run: |
-        cp -af cassandane.ini.dockertests cassandane.ini
-        chown cyrus:mail cassandane.ini
-        make -j8
     - name: run cassandane quietly
       id: cass1
       continue-on-error: true
-      working-directory: cassandane
-      run: setpriv --reuid=cyrus --regid=mail --clear-groups --inh-caps='-chown,-dac_override,-dac_read_search,-fowner,-fsetid,-kill,-setgid,-setuid,-setpcap,-linux_immutable,-net_bind_service,-net_broadcast,-net_admin,-net_raw,-ipc_lock,-ipc_owner,-sys_module,-sys_rawio,-sys_chroot,-sys_ptrace,-sys_pacct,-sys_admin,-sys_boot,-sys_nice,-sys_resource,-sys_time,-sys_tty_config,-mknod,-lease,-audit_write,-audit_control,-setfcap,-mac_override,-mac_admin,-syslog,-wake_alarm,-block_suspend,-audit_read,-cap_38,-cap_39,-cap_40' ./testrunner.pl --slow -f prettier -j 8 !Test::Core
+      run: cyd test
```
elliefm, line 54: CI for this one is failing on the Test::Core tests, which we used to skip in CI (`!Test::Core` at the end of the run command), but with this change I guess they're no longer skipped.

These tests prove that Cassandane can detect when Cyrus crashed -- which means, if these tests succeed, and everything else also succeeded, then nothing crashed.  When these tests fail (or are skipped), that means Cassandane can't detect Cyrus crashes, and tests that result in Cyrus crashes might accidentally report success if the test itself didn't notice the crash. 

So it would be good to not skip them, but I never figured out how to get them working in CI.  On a normal system, you need to configure a few things to get crashed processes to dump core.  On CI, between all the layers of docker and stuff, I don't know what's configured or where to change it.

rjbs: Interesting.  They must(?) be passing in Docker for cyd/dar to work, so it'll be something more constrained on GitHub Actions.

rjbs: We're skipping the Test::Core tests now, but we'll see if we can't bring them back in the future…

rjbs:
I think this is good to go.  It changes the name of some checks, so when this is merged, we'll need to update the required check list.

rjbs:
It would be good to merge https://github.com/cyrusimap/cyrus-docker/pull/35 shortly after (or before) this.

.github/workflows/main.yml
```diff
@@ -32,15 +35,11 @@ jobs:
...
-    - name: install missing or frequently-updated deps
-      shell: bash
-      run: |
-        cpanm Mail::IMAPTalk # in image, but fetch latest!
     - name: setup git safe directory
       shell: bash
       run: git config --global --add safe.directory /__w/cyrus-imapd/cyrus-imapd
     - name: fetch upstream release tags
-      if: ${{ github.repository != 'cyrusimap/cyrus-imapd' }}
+      # if: ${{ github.repository != 'cyrusimap/cyrus-imapd' }}
```
wolfsage, line 42: Is this intentional? We only need to fetch upstream tags if we are on someone elses branch IIRC?

wolfsage (changes requested):

.github/workflows/main.yml
rjbs: I can't recall. 😕  I've removed the commenting-out and pushed.

wolfsage:
Seems good. Does this change matter: Something somewhere is requiring `Cassandane (gcc)` but now it's `Cassandane (--gcc)` .... (same with clang)

Where is that configured?

wolfsage (approved):

rjbs:
@wolfsage Yeah, you'll see I wrote:
> It changes the name of some checks, so when this is merged, we'll need to update the required check list.

It's in the protected branches for the repository.  I'll update it when this is shipped.

PR #5205, opened 2025-01-09
message.c: reject to parse MIME headers with NUL byte

rsto:
Before, the parse continued but set incorrect header and body offsets for the body part. This resulted in log spam noting the discrepancy between the file size and header+body offsets, as well as a bogus in-memory MIME body structure.

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_import_zerobyte
+++ b/cassandane/tiny-tests/JMAPEmail/email_import_zerobyte
@@ -2,13 +2,14 @@
 use Cassandane::Tiny;
 
 sub test_email_import_zerobyte
-    :min_version_3_1 :needs_component_sieve
+  :needs_component_sieve
 {
     my ($self) = @_;
     my $jmap = $self->{jmap};
 
-    # A bogus email with an unencoded zero byte
-    my $email = <<"EOF";
+    my @testCases = ({
+        desc => "unencoded zero byte in MIME body",
+        mime => <<"EOF",
 From: \"Some Example Sender\" <example\@local>\r\n
 To: baseball\@local\r\n
 Subject: test email\r\n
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):

imap/message.c
```diff
@@ -4956,7 +4969,8 @@ static int body_foreach_section(struct body *body, struct message *message,
             msg.len = body->header_size;
             msg.offset = 0;
             msg.encode = 0;
-            message_parse_headers(&msg, tmpbody, "text/plain", &boundaries, NULL);
+            message_parse_headers(&msg, tmpbody, "text/plain", &boundaries,
```
wolfsage, line 4972: Kinda surprised we don't care about errors in parsing here

imap/message.c
```diff
@@ -522,36 +523,38 @@ EXPORTED int message_parse_mapped(const char *msg_base, unsigned long msg_len,
...
-            /* XXX IOERROR but only LOG_NOTICE?? */
-            xsyslog(LOG_NOTICE, "IOERROR: size mismatch on parse",
-                                "guid=<%s> filename=<%s> "
-                                "filesize=<%" PRIu32 "> bodysize=<%" PRIu32 ">",
-                                message_guid_encode(&body->guid), efname,
-                                body->filesize,
-                                body->header_size + body->content_size);
+            xsyslog(LOG_ERR, "IOERROR: size mismatch on parse",
+                             "guid=<%s> filename=<%s> "
+                             "filesize=<%" PRIu32 "> bodysize=<%" PRIu32 ">",
```
elliefm, line 544: I think I probably formatted this xsyslog call originally, but my thinking has changed since.  I'd like to see this:

```suggestion
            xsyslog(LOG_ERR, "IOERROR: size mismatch on parse",
                             "guid=<%s> filename=<%s>"
                             " filesize=<%" PRIu32 "> bodysize=<%" PRIu32 ">",
```

That is, with the separating space at the start of the next line, instead of the end of the previous line.  It's visually a little jarring, which I think draws attention to the fact that the format argument is a single string literal being concatenated together from adjacent pieces (i.e. no comma in between) which are wrapped over multiple lines.

cassandane/tiny-tests/JMAPEmail/email_import_zerobyte
```diff
@@ -2,13 +2,13 @@
 use Cassandane::Tiny;
 
 sub test_email_import_zerobyte
-    :min_version_3_1 :needs_component_sieve
-{
+  : needs_component_sieve {
```
elliefm, line 5: style nit

```suggestion
  :needs_component_sieve
{
```

elliefm (commented):
(couple of style nits)

imap/message.c
rsto: OK. I adapted the line for both xsyslog calls.

cassandane/tiny-tests/JMAPEmail/email_import_zerobyte
rsto: done

imap/message.c
rsto: Here, we parse the headers for on an already parsed body-part here, e.g. these headers already were validated otherwise that body part wouldn't exist. We didn't check for errors previously here, because message_parse_headers didn't check for any.

It might be good to check for message_parse_headers errors here, too, but in case of errors it's a very different situation we are in: it means that an already accepted email on disk could now not be read anymore. We would need to treat this differently and I'd prefer to keep the damage radius of this change small.

PR #5206, opened 2025-01-09
xapian_wrap.cpp: support phrase search for email addresses

rsto:
Quotes signify phrase search in free text queries but were not supported in email queries since we supported querying for sub-parts of an email address. This reintroduces phrase search for email queries.


ksmurchison (approved):
LGTM

PR #5207, opened 2025-01-09
Fix compile error on macOS: unknown type name 'ulong'

cwalther:
Trying to build cyrus-imapd on macOS (12.5–15.2) fails with a compiler error `unknown type name 'ulong'` in imap/http_client.h and imap/httpd.h (even with no HTTP features enabled). That type does not appear to be defined anywhere in the source code.

Checking on Linux, it appears to be a Linux-specific thing, in my case coming from the following part of /usr/include/x86_64-linux-gnu/sys/types.h:

```c
#ifdef __USE_MISC
/* Old compatibility names for C types.  */
typedef unsigned long int ulong;
```

Using that typedef does not seem to add any value, replace it by the standard type.

### DIFF
--- a/imap/http_client.h
+++ b/imap/http_client.h
@@ -57,7 +57,7 @@ struct body_t {
     unsigned char framing;              /* Message framing   */
     unsigned char te;                   /* Transfer-Encoding */
     unsigned max;                       /* Max allowed len   */
-    ulong len;                          /* Content-Length    */
+    unsigned long len;                  /* Content-Length    */
     struct buf payload;                 /* Payload           */
 };
 
--- a/imap/httpd.h
+++ b/imap/httpd.h
@@ -280,7 +280,7 @@ typedef int (*encode_proc_t)(struct transaction_t *txn,
 
 /* Meta-data for response body (payload & representation headers) */
 struct resp_body_t {
-    ulong len;                          /* Content-Length   */
+    unsigned long len;                  /* Content-Length   */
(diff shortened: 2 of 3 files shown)


rjbs (approved):

rjbs:
Seems like this should also patch `imap/jmap_mail_submission.c`.  Will get this sorted.

cwalther:
Ah, good point. It didn’t occur to me to search for all occurrences, I just fixed what got in my way, and I didn’t have JMAP enabled.

rjbs:
@cwalther Could you click "Allow edits and access to secrets by maintainers" on this PR?  Failing that, I'll make a new one based on it, tomorrow.

cwalther:
I don’t see that checkbox anymore. It used to be in the right sidebar, right?
But I can force-push an amended commit with the additional change to jmap_mail_submission.c.

cwalther:
GitHub Docs say
> On user-owned forks, if you want to allow anyone with push access to the upstream repository to make changes to your pull request, select Allow edits from maintainers.

so maybe the checkbox isn’t there because this is an organization-owned fork (no idea why that should make a difference).

Anyway, if the new commit isn’t what you wanted, let me know or proceed with the previous one as you wish.

rjbs:
Thanks!  I expect I'll merge this tomorrow when I'm back on the clock! 👍🏽

cwalther:
Thank you!

PR #5208, opened 2025-01-09
Fix build without libical

cwalther:
Building cyrus-imapd on a system that does not have libical installed (with iCal features accordingly disabled by `configure`) fails with a compiler error `'libical/ical.h' file not found` in various places. Checked on macOS (15.2) and Linux (Ubuntu 20.04).

This appears to have been broken when #3467, #3815, #3845 started including code that was previously exclusive to CalDAV features also when CalDAV is disabled. It was presumably not noticed when testing only on systems that do have libical installed, even with CalDAV explicitly disabled.

I can fix this with the `#ifdefs` added in the attached commit, but am not sure whether these are exactly the right macros to test.

### DIFF
--- a/imap/dav_db.c
+++ b/imap/dav_db.c
@@ -55,7 +55,9 @@
 #include <sys/wait.h>
 
 #include "assert.h"
+#ifdef WITH_DAV
 #include "caldav_alarm.h"
+#endif
 #include "cyrusdb.h"
 #include "dav_db.h"
 #include "global.h"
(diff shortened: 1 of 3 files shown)


ksmurchison:
Is/was your goal here to build without httpd (DAV, JMAP, etc) entirely, or just build without support for calendars?

cwalther:
Yes, without httpd. All I need is POP and IMAP.

cwalther:
Rebased on master, with no modifications, because I still need it to fix my build.

PR #5211, opened 2025-01-10
WIP: testing rejiggering of docs (and Markdown)

rjbs:


elliefm (changes requested):
Looks good!

A few things still needed to complete the job:

* around line 2425 of configure.ac, add myst_parser to the rest of the "Documentation dependencies" outputs that are reported to the terminal
* also add it to the dependencies list in docsr/developer/compiling.rst.  The deb package is called 'python3-myst-parser'
* changes file

changes/next/markdown-docs
```diff
@@ -0,0 +1,26 @@
+Description:
+
+Content for the documentation site (cyrusimap.org) can now be written in
+Markdown.
```
elliefm: So, from the point of view of someone installing, upgrading, or packaging Cyrus (i.e. the audience of the release notes), the interesting change is that they will now need myst-parser if they want to (re)build the documentation.  Last time they built Cyrus, Cyrus didn't need it, so they probably don't just have it already.  This is the kind of thing I would stick under the "Build changes" heading.

Being able to write documentation in markdown is interesting to people writing (or avoiding writing) documentation, but not to anyone else, I don't think.

rjbs: I have adjusted this bit!

PR #5212, opened 2025-01-11
import new testing from master branch to 3.8

rjbs:


rjbs:
I think this can be applied at will.  Future CI backports will likely be simpler after this one.

elliefm (approved):

PR #5215, opened 2025-01-14
iCalendar: don't crash if RRULE exists with no value

ksmurchison:


rsto (approved):
LGTM - would prefer for libical to reject/ignore such properties, though.

ksmurchison:
> LGTM - would prefer for libical to reject/ignore such properties, though.

I'd have to check, but I think it rejects all empty properties by default, and -DICAL_ALLOW_EMPTY_PROPERTIES=true is what allows these.  We should probably limit that to only allowing empty text-valued properties.

rsto:
Oh, didn't think of `ICAL_ALLOW_EMPTY_PROPERTIES=true`. Yeah, we'd only want this enabled for TEXT values.

> Do you want me to stash this PR and work on libical instead?

I'd say no, handling this in Cyrus at least works also with other libical installations that might have the same build options.
The long-sought iCalendar repair layer might make this all more straight-forward to handle.

PR #5217, opened 2025-01-15
add a cast needed to silence an error in clang

rjbs:
buf_putc wants a char (no specified sign), and we're passing an int with the high bit very intentionally set.  We need to cast the resulting expression to (char).


rsto:
I cherry-picked that. Thanks.

PR #5221, opened 2025-01-17
Minifixes

brong:
This is three little fixes from the twom branch that make sense to backport to earlier releases.


elliefm:
from slack:
> when conversationsdb is built, the basecid gets saved in the G values
> but when you rebuild (or audit), it doesn't get read in, so the value changes
> it makes audit fail on tests where you actually have a cyrusdb_next (e.g. twoskip and twom)
> but it also makes audit ugly in prod

(and the fix makes it read the existing basecid)

elliefm (approved):

PR #5225, opened 2025-01-20
message: make some strcmps "safe"

brong:
This fixes crashers with parsing malformed MESSAGE/RFC822


lerlacher-fm (approved):
These changes might hide some parsing logic errors and should be looked at to see when we might run into the null cases for valid messages. Apart from that it's straightforward and safe.

brong:
I've added @ksmurchison for review as well before merging it upstream.

ksmurchison (approved):
LGTM.  Can we collapse this into one commit?

brong:
> LGTM. Can we collapse this into one commit?

sure!

PR #5229, opened 2025-01-22
cass: support newer HTTP::Tiny, Net::DAVTalk, Mail::JMAPTalk

elliefm:
Newer HTTP::Tiny turns on SSL certificate verification by default, which meant tests using HTTPS were failing.  This was previously worked around in #5227.

There have since been updates to Mail::JMAPTalk and Net::DAVTalk to pass through our SSL_options, so if the new versions of those are installed we can skip the workaround and do it properly instead, with certificate verification enabled.

Also added a simple test that exercises JMAP-over-HTTPS, because this was broken just like DAV-over-HTTPS was, just not visibly so because nothing used it.

### DIFF
--- a/cassandane/Cassandane/Cyrus/TestCase.pm
+++ b/cassandane/Cassandane/Cyrus/TestCase.pm
@@ -41,12 +41,14 @@ package Cassandane::Cyrus::TestCase;
 use strict;
 use warnings;
 use attributes;
+use version 0.77;
+use Cwd qw(abs_path);
 use Data::Dumper;
-use Scalar::Util qw(refaddr);
-use List::Util qw(uniq);
 use Digest::file qw(digest_file_hex);
-use File::Temp qw(tempfile);
 use File::Path qw(rmtree);
+use File::Temp qw(tempfile);
+use List::Util qw(uniq);
+use Scalar::Util qw(refaddr);
 
 use lib '.';
 use base qw(Cassandane::Unit::TestCase);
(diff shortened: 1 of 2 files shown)


cassandane/tiny-tests/JMAPCore/echo-tls
```diff
@@ -0,0 +1,28 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_echo_tls
```
rsto, line 4: The file name contains a dash but the test name an underscore. Please change the file name to match the test name.

rsto (changes requested):
Works for me:

$ perl -MHTTP::Tiny -le 'print $HTTP::Tiny::VERSION'
0.080

$ perl -MNet::DAVTalk -le 'print $Net::DAVTalk::VERSION'
0.2

$ perl -MMail::JMAPTalk -le 'print $Mail::JMAPTalk::VERSION'
0.15

./testrunner.pl --config ./cassandane.ini -j 4 -f pretty JMAPCore.echo_tls
[  OK  ] Cyrus::JMAPCore.echo_tls

cassandane/tiny-tests/JMAPCore/echo-tls
wolfsage: Rik usually prefers file names not have underscores because they can be hard to see in terminal output or other situations, so we use dashes instead.

Perl subroutines can't contain dashes, so we use underscores instead.

I think it would be fair to say files use dashes and subs use underscores. We have a mix of both right now but settling on one would be better.

What do you think @rsto?

wolfsage (approved):
If this works, seems fine to me.

An alternate option to consider would be to require the latest of everything. Folks will have to update some modules once, and then we don't have this confusing situation anymore?

rsto:
@wolfsage In the case of Cassandane/tiny-tests I prefer we keep using underscore also in the file name. I'm surprised we do have file names with dashes because Rik even renamed the test files from using dash to underscore after we agreed to use underscores: https://github.com/cyrusimap/cyrus-imapd/commit/115323ee1203da8b3d41031ac4cf76cbb3973e0c

The reason for that is that it's more ergonomic being able to copy the name of a failing test and paste that name into the command line, without having to replace all underscores.

wolfsage:
> The reason for that is that it's more ergonomic being able to copy the name of a failing test and paste that name into the command line, without having to replace all underscores.

Okay, works for me!

elliefm:
> If this works, seems fine to me.
> 
> An alternate option to consider would be to require the latest of everything. Folks will have to update some modules once, and then we don't have this confusing situation anymore?

Yep, I agree, except that we don't `use Foo;` these modules in this file, so we can't `use Foo <version>;` them.  We'd need to do `require Foo; Foo->VERSION('<version>');` or something, which I tried briefly and hated.

The reason we don't just `use Foo;` in this file is so that when Cyrus was built without http support, someone running Cassandane isn't forced to install a bunch of perl modules that won't actually be used (because all the tests that would care will be skipped anyway).

We intend at some point to make http support non-optional; at that point, this mess can go away and we can just `use Foo <version>;` these modules at the top like usual.

elliefm:
@rsto I've renamed the echo-tls file to echo_tls.

rsto (approved):

PR #5232, opened 2025-01-22
Makefile.am: enable compile warnings in CFLAGS

rsto:
This enables the "all" and "extra" warning diagnostics of the compiler and instructs it to treat any warning as compile error.

Most of the current Cyrus IMAP maintainers have these flags enabled in their custom build scripts, but we now want to establish them as baseline for any builds.

Tested with gcc version 12.2.0 and clang version 14.0.6.


ksmurchison (approved):

dilyanpalauzov:
I am against.  The mechanism to plug arbitrary compiler flags is by using CFLAGS/CXXFLAGS when calling `./configure` or `make`.  `clang` does not use `-Wall -Wextra` but `-Weverything`.  Moreover if I remember correctly, `-O3 -flto -Wall -Wextra` produces even more warnings than `-Wall -Wextra`.  That said, this is out of scope for the build system.

To ensure that these flags are used whenever `./configure` is invoked, one can put in /usr/local/etc/config.site:
```
export CFLAGS='-Wall -Wextra -O3'
export CXXFLAGS='-Wall -Wextra -O3'
```
(I do have this)

rsto:

elliefm:
AM_XXX is the right place for these warning options.  We expect every build to have them enabled, we expect PR submissions to be free from these warning classes, etc.

Dilyan is correct that mechanisms exist for overriding them, and we also expect this.  If you want to add extra flags, or need to disable some on your build system for some reason, then you can use those mechanisms to do so.  But if you have not configured your system in any particular way, you will get what AM_XXX sets, so AM_XXX should contain the correct defaults for a normal build of Cyrus.

Makefile.am
```diff
@@ -41,11 +41,15 @@
 # OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 #
 
+WARN_CFLAGS = -Wall -Wextra -Werror
+WARN_CPPFLAGS = -Wall -Wextra -Werror
```
elliefm, line 45: This distinction isn't really meaningful, unless you mistook "CPP" for C++ (in this context, it means "C pre-processor").  A single `WARN_FLAGS = ...` would be fine.

Makefile.am
```diff
@@ -41,11 +41,15 @@
...
+
 ACLOCAL_AMFLAGS = -I cmulocal
-AM_CFLAGS = @PERL_CCCDLFLAGS@ $(GCOV_CFLAGS)
+AM_CFLAGS = $(WARN_CFLAGS) @PERL_CCCDLFLAGS@ $(GCOV_CFLAGS)
 AM_CXXFLAGS = $(GCOV_CXXFLAGS)
```
elliefm, line 49: Any reason you didn't update `AM_CXXFLAGS`?  We do have C++ code in Cyrus (xapian_wrap.cpp, icu_wrap.cpp), and we want warnings-as-errors for those files too.

My own wrapper script already sets these options in `CXXFLAGS`, and the C++ code builds clean, for what it's worth.

elliefm (changes requested):

Makefile.am
elliefm: If we _did_ start wanting different warning options for C vs C++, we'd run into a big annoyance anyway in that they'll both use CPPFLAGS, so we'd need some way to set that differently for different targets.  But we expect it to all just be GCC (or Clang) anyway, it's not like people are gonna be compiling the C code with Intel and the C++ with MSVC or something, so there's no reason to differentiate.

dilyanpalauzov:
https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html describes the difference between `AM_CFLAGS` and `CFLAGS` variables, and any other X versus AM_X.  CFLAGS can be set by the users, AM_CFLAGS is set in Makefile.am.

>  If you … need to disable some flags on your build system for some reason, then you can use those mechanisms to do so. 

So the way to disable `-Werror` because for some reason with a new compiler a new `-Wall` warning appears, users are supposed to set `CFLASG=-Wno-error`.  This is possible, but I guess this is hard to figure out for an average person who tries to compile the software by herself.

The automake build system should just not utilize optional compiler flags, not even `__attribute__((optimize("-O3")))` and leave the choice of optional flags to the person compiling the software.

> we expect PR submissions to be free from these warning classes

Rejecting changes, which trigger warnings, can be enforced by github checks, which utilize `-wall -wextra -werror`, the current changes is not essential for this.

Makefile.am
rsto: My bad, I mistook CPPFLAGS with CXXFLAGS.

Makefile.am
rsto: Agreed. I now set CXXFLAGS to CFLAGS, but locally I ran CXXFLAGS additionally with `-Wpedantic` for quite some time.

elliefm (approved):

dilyanpalauzov:
```
$ make imtest/imtest-imtest.o V=1
clang -DHAVE_CONFIG_H -I.   -I. -I./lib -I. -I./lib -DLIBEXEC_DIR=\"/usr/local/libexec\" -DSBIN_DIR=\"/usr/local/sbin\" -DSYSCONF_DIR=\"/etc\" -DHAVE_CONFIG_H      -I/usr/include/libxml2 -DWITH_GZFILEOP  -I/usr/local/include    -I/usr/local/include    -I/usr/local/include/chardet -Wall -Wextra -g -fvisibility=hidden    -Wall -Wextra -Werror   -fvisibility=hidden -Wall -Wextra -O3 -DPCRE2POSIX_SHARED -MT imtest/imtest-imtest.o -MD -MP -MF imtest/.deps/imtest-imtest.Tpo -c -o imtest/imtest-imtest.o `test -f 'imtest/imtest.c' || echo './'`imtest/imtest.c
imtest/imtest.c:151:25: error: cast from 'int (*)(void *, const char *, const char *, const char **, unsigned int *)' to 'int (*)(void)' converts to incompatible function type [-Werror,-Wcast-function-type-mismatch]
  151 |         SASL_CB_GETOPT, (int (*)(void))&mysasl_config, NULL
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make: *** [Makefile:7164: imtest/imtest-imtest.o] Error 1

```

rsto:
Which clang version are you using?

dilyanpalauzov:
> Which clang version are you using?

clang version 19.1.7

PR #5236, opened 2025-01-22
append.c: only call out to annotator on email

rjbs:
The annotator exists so that installations can update email messages to have extra flags or annotations.  Because "everything" is an email, it was also running on sieve scripts and other non-emails.  This was probably harmless on many installations, but not intended.

This branch updates the annotator to run only on messages being appended to mailboxes of type `e`.

It also fixes a bug in `sieve_db`, wherein we only set the authstate for the `onSuccessActivateScript` side effect, not the rest of script creation.

### DIFF
--- a/imap/append.c
+++ b/imap/append.c
@@ -1107,10 +1107,8 @@ havefile:
         goto out;
     }
 
-    if (
-        config_getstring(IMAPOPT_ANNOTATION_CALLOUT)
-        && (mbtype_isa(mailbox_mbtype(mailbox)) == MBTYPE_EMAIL)
-    ) {
+    if (config_getstring(IMAPOPT_ANNOTATION_CALLOUT) &&
+        (mbtype_isa(mailbox_mbtype(mailbox)) == MBTYPE_EMAIL)) {
         if (flags)
             newflags = strarray_dup(flags);
         else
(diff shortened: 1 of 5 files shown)


imap/append.c
```diff
@@ -1107,7 +1107,10 @@ EXPORTED int append_fromstage_full(struct appendstate *as, struct body **body,
         goto out;
     }
 
-    if (config_getstring(IMAPOPT_ANNOTATION_CALLOUT)) {
+    if (
+        config_getstring(IMAPOPT_ANNOTATION_CALLOUT)
```
ksmurchison, line 1111: Just a nit.  We (the core devs) typically don't put like breaks after the opening '(' and before the closing ')'

ksmurchison (approved):
Othe rthan my comment, LGTM

imap/append.c
rjbs: Happy to address.  (Also, eager to see what autoformatting we could agree on…)

rjbs: @ksmurchison I believe my latest commit matches the style I've now found elsewhere in append.c.  Look right to you?

ksmurchison: Yes sir.  All good!

PR #5240, opened 2025-01-28
Add coding style for emacs

dilyanpalauzov:
This convinces emacs to do the right indentation by default, when typing:
* no tabs policy
* indent jumps by four and not by two spaces


brong (approved):
I don't use emacs but we may as well make it nice for people who do

rjbs:

PR #5241, opened 2025-01-28
Add xsyslog_ev - a logfmt style logger

wolfsage:
This implements roughly the same logging format as https://metacpan.org/dist/Log-Dispatchouli/source/lib/Log/Fmt.pm.

It's structed in a way to maintain compiler time safety checks around variable assignments, type conversions, and printf format strings.

A simple usage looks like:

    int errors = 0;
    ...
    xsyslog_ev(LOG_ERR, "name of event", lf_d("error-count", errors));

which will produce a log entry looking like:

    event="name of event" error-count=3

This provides a number of common printf-style formats used in our existing logs as macros to get the same behaviour. For example, for '%d', you'd use lf_d("key", value), for '%s', lf_s(...), etc...

If you want more control over formatting, there's also lf_raw:

    xsyslog_ev(LOG_ERR, "event", lf_raw("float", "%.02f", 3.14159));

which produces:

    event="event" float=3.14

(lf_raw should probably be used sparingly as it adds more inline code to each place that uses it, increasing binary size...)

You can use as many macros together as necessary. Taking a real example:

    xsyslog(LOG_ERR, "DBERROR: database needs upgrade",
                     "fname=<%s> have=<%d> want=<%d> error=<%s>",
                     open->fname, open->version, version,
                     sqlite3_errmsg(open->db));

becomes:

    xsyslog_ev(LOG_ERR, "DBERROR: database needs upgrade",
        lf_s("fname", open->fname),
        lf_d("have", open->version),
        lf_d("want", version),
        lf_s("error", sqlite3_errmsg(open->db))
    );

_xsyslog_ev_escape() comes from Ken Murchison with modifications by Mark Jason Dominus and myself.

This work is based off of some back and forth conversations with ellie.

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -2193,3 +2193,137 @@ EXPORTED char *modseqtoa(modseq_t modseq)
     buf_printf(&buf, MODSEQ_FMT, modseq);
     return buf_release(&buf);
 }
+
+static char *_xsyslog_ev_escape(const char *val)
+{
+    struct buf buf = BUF_INITIALIZER;
+    int needs_escaping = 0, orig_len, escaped_len;
+    const char *p;
+
+    buf_setcstr(&buf, val);
+
+    escaped_len = orig_len = buf_len(&buf);
+
+    // Make sure the empty string is visible
+    if (orig_len == 0) {
+        buf_setcstr(&buf, "\"\"");
+        return buf_release(&buf);
+    }
+
+    for (p = buf_cstring(&buf); *p; p++) {
+        switch (*p) {
+        case '\\':
+        case '\"':
+        case '\n':
+        case '\r':
+            ++escaped_len;  // add 1 for the backslash
+
+        // FALL THROUGH
+        case ' ':
+            needs_escaping = 1;
+            break;
+        }
+    }
+
+    if (needs_escaping) {
+        char *q;
+
+        escaped_len += 2;  // add 2 for surrounding DQUOTEs
+
+        buf_truncate(&buf, escaped_len);  // grow the buffer to escaped length
+
+        // we can now build the escaped value in place, tail to head
+        q = (char *) buf_base(&buf) + escaped_len - 1;
+        *q-- = '\"';  // closing DQUOTE
+
+        for (p = buf_base(&buf) + orig_len - 1; p >= buf_base(&buf); p--) {
+            char c = *p;
+
+            switch (c) {
+            case '\\':
+            case '\"':
+                needs_escaping = 1;
+                break;
+
+            case '\n':
+                needs_escaping = 1;
+                c = 'n';
+                break;
+
+            case '\r':
+                needs_escaping = 1;
+                c = 'r';
+                break;
+
+            default:
+                needs_escaping = 0;
+                break;
+            }
+
+            *q-- = c;
+
+            if (needs_escaping) *q-- = '\\';
+        }
+
+        assert(q == buf_base(&buf));
+        *q = '\"';  // opening DQUOTE
+    }
+
+    return buf_release(&buf);
+}
+
+EXPORTED void _xsyslog_ev(int priority, const char *event,
+                          logfmt_arg_list *arg)
+{
+    static struct buf buf = BUF_INITIALIZER;
+    int saved_errno = errno;
+
+    char *escaped = _xsyslog_ev_escape(event);
+    buf_printf(&buf, "event=%s", escaped);
+    free(escaped);
+
+    for (size_t i = 0; i < arg->nmemb; i++) {
+        buf_printf(&buf, " %s=", arg->data[i].name);
+
+        switch(arg->data[i].type) {
+        case LF_C:   buf_printf(&buf, "%c",   arg->data[i].c);   break;
+        case LF_D:   buf_printf(&buf, "%d",   arg->data[i].d);   break;
+        case LF_LD:  buf_printf(&buf, "%ld",  arg->data[i].ld);  break;
+        case LF_LLD: buf_printf(&buf, "%lld", arg->data[i].lld); break;
+        case LF_U:   buf_printf(&buf, "%u",   arg->data[i].u);   break;
+        case LF_LU:  buf_printf(&buf, "%lu",  arg->data[i].lu);  break;
+        case LF_LLU: buf_printf(&buf, "%llu", arg->data[i].llu); break;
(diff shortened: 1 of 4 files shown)


ksmurchison (approved):
LGTM

lib/util.c
```diff
@@ -2193,3 +2193,137 @@ EXPORTED char *modseqtoa(modseq_t modseq)
...
+        case LF_LD:  buf_printf(&buf, "%ld",  arg->data[i].ld);  break;
+        case LF_LLD: buf_printf(&buf, "%lld", arg->data[i].lld); break;
+        case LF_U:   buf_printf(&buf, "%u",   arg->data[i].u);   break;
+        case LF_LU:  buf_printf(&buf, "%lu",  arg->data[i].lu);  break;
+        case LF_LLU: buf_printf(&buf, "%llu", arg->data[i].llu); break;
+        case LF_ZD:  buf_printf(&buf, "%zd",  arg->data[i].zd);  break;
+        case LF_ZU:  buf_printf(&buf, "%zu",  arg->data[i].zu);  break;
+        case LF_LLX: buf_printf(&buf, "%llx", arg->data[i].llu); break;
+        case LF_F:   buf_printf(&buf, "%f",   arg->data[i].f);   break;
+        case LF_M:
```
wolfsage, line 2300: I wonder if we even want to support/use `%m` or errno in this way. It's *possible* something could change errno by the time we get to it here due to one of the buf_printfs doing something weird, or if a xsyslog_ev(...) uses a function as an argument that changes errno...

I can work around it, but it makes me think our existing uses of '%m' could be wrong and that xsyslog even has this wrong...

lib/util.c
```diff
@@ -2193,3 +2193,137 @@ EXPORTED char *modseqtoa(modseq_t modseq)
...
+
+static char *_xsyslog_ev_escape(const char *val)
+{
+    struct buf buf = BUF_INITIALIZER;
+    int needs_escaping = 0, orig_len, escaped_len;
+    const char *p;
+
+    buf_setcstr(&buf, val);
+
+    escaped_len = orig_len = buf_len(&buf);
```
rsto, line 2205: buf_len returns a size_t so why cast implicitly to int here?
nit: from C99 onwards there's no need to declare variables at the start of the function.

lib/util.c
```diff
@@ -2193,3 +2193,137 @@ EXPORTED char *modseqtoa(modseq_t modseq)
...
+            break;
+        }
+    }
+
+    if (needs_escaping) {
+        char *q;
+
+        escaped_len += 2;  // add 2 for surrounding DQUOTEs
+
+        buf_truncate(&buf, escaped_len);  // grow the buffer to escaped length
```
rsto, line 2233: while buf_truncate does what it's supposed to do here, I would say that calling `buf_ensure` would make intention clearer (probably then even without the comment).

lib/util.c
```diff
@@ -2193,3 +2193,137 @@ EXPORTED char *modseqtoa(modseq_t modseq)
...
+}
+
+EXPORTED void _xsyslog_ev(int priority, const char *event,
+                          logfmt_arg_list *arg)
+{
+    static struct buf buf = BUF_INITIALIZER;
+    int saved_errno = errno;
+
+    char *escaped = _xsyslog_ev_escape(event);
+    buf_printf(&buf, "event=%s", escaped);
```
rsto, line 2282: It's a micro-optimization in light of the heap allocations later in this function and maybe it's even better to keep `buf_print` for consistency with the LF_ formatting below, but: in general we tend to avoid `buf_printf` for simple formatting. It's surprisingly costly when compared to

```
buf_setcstr(&buf, "event=");
buf_appendcstr(&buf, escaped);
```

lib/util.c
```diff
@@ -2193,3 +2193,137 @@ EXPORTED char *modseqtoa(modseq_t modseq)
...
+    for (p = buf_cstring(&buf); *p; p++) {
+        switch (*p) {
+        case '\\':
+        case '\"':
+        case '\n':
+        case '\r':
+            ++escaped_len;  // add 1 for the backslash
+
+        // FALL THROUGH
+        case ' ':
```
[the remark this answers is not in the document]
wolfsage: Fixed, see what you think?

Relatedly... you say `\x3E-x7E` but that includes 0x5C which is a '\' which you then say should be quoted and escaped.

It looks like the perl implementation doesn't detect '\' as needing escaping or quoting (unless something else is detected as needed escaping). I think that's a bug?

lib/util.c
wolfsage: I've gone and applied this patch. I think it's right.

lib/util.c
wolfsage: Fixed the size_t issue. I think there's still a habit of initialize at the top we haven't fully broken yet :) I've left that bit...

lib/util.c
wolfsage: I think truncate/ensure have unfortunate names. But this looks clearer (to me) than:
```
buf_ensure(&buf, escaped_len - orig_len);
```

Rather than do the math, using the precomputed value seems most sensible?

lib/util.c
wolfsage: I've made some changes here, see what you think?

lib/util.c
elliefm: Collecting variable declarations at the top of their scope is a good habit in languages with manual memory management.  We don't have to copy C++'s bad habits just because we now can.

rsto: > Collecting variable declarations at the top of their scope is a good habit in languages with manual memory management. We don't have to copy C++'s bad habits just because we now can.

I disagree. But there is no point in back and forth without an agreed code style for this project, so I'll resolve this comment.

lib/util.c
rsto: Yes, let's keep it as-is, I forgot about buf_ensure having this weird API.

rsto (approved):

lib/util.c
wolfsage: Ah, gcc does when -O is involved. Wild.

lib/util.c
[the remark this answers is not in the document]
rjbs: 0x5C is not `"`, it's `\`

`"` is 0x22.  So, we do NOT quote things with `\`.  I think this is fine:

```
  DB<4> x Log::Fmt->format_event_string([ a => qq{pan\\ama} ])
0  'a=pan\\ama'
```

It is not ambiguous when parsing.

PR #5246, opened 2025-01-31
http_jmap.c: allow EventSource without WebSocket

ksmurchison:
Also, jmap_push_poll:0 should disable EventSource


elliefm (approved):
Good catch, thanks

PR #5248, opened 2025-02-03
cyr_expire: add batchsize

brong:
This makes it easier to do larger batches when cleaning up a mess on a replica!

### DIFF
--- a/docsrc/imap/reference/manpages/systemcommands/cyr_expire.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/cyr_expire.rst
@@ -19,7 +19,7 @@ Synopsis
     **cyr_expire** [ **-C** *config-file* ] [ **-A** *archive-duration* ]
     [ **-D** *delete-duration* ] [ **-E** *expire-duration* ] [ **-X** *expunge-duration* ]
     [ **-p** *mailbox-pre‐fix* ] [ **-u** *username* ] [ **-t** ] [ **-v** ]
-    [ **-a** ] [ **-c** ] [ **-x** ]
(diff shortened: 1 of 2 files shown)


docsrc/imap/reference/manpages/systemcommands/cyr_expire.rst
```diff
@@ -160,6 +160,11 @@ Options
     Remove any user flags which are not used by remaining (not expunged)
     messages.
 
+.. option:: -b num, --batchsize=num
```
dilyanpalauzov, line 163: In the synopsys the parameter is called batchsize.

brong: I have changes this to match, in the same style as the others.

elliefm (approved):

dilyanpalauzov:
For cyr_expire.rst in the synopsis the parameter names (config-file, archive-duration, delete-duration, etc) match the names, that are used in the long parameter description afterwards.

For **-b** the parameter name is *batchsize* in the synopsis, but the long description for **-b** names its parameter *num*.

chittaranjan870:
**Expired 0 and expunged 0 out of 90 messages from 270 mailboxes**

what did the meaning of above log, I don't know why  the archive command is not working.
Is there any wrong I am doing.

/usr/lib/cyrus/bin/cyr_expire  -A 7d

imap.conf
# Which partition to use for default mailboxes
defaultpartition: default
partition-default: /var/spool/cyrus/mail
partition-archive: /var/spool/cyrus/archive

archive_enabled: 1
archive_after: 7d
archive_maxsize: 1024
archive_keepflagged: 1

PR #5249, opened 2025-02-03
ContactCard uses addressBookIds, not addressBookId

wolfsage:


imap/jmap_contact.c
```diff
@@ -7727,8 +7733,8 @@ static int getcards_cb(void *rock, struct carddav_data *cdata)
     }
 
     json_object_set_new(obj, "id", json_string(cdata->vcard_uid));
-    json_object_set_new(obj, "addressBookId",
-                        json_string(strrchr(mbentry->name, '.')+1));
+    json_object_set_new(obj, "addressBookIds",
+                             json_pack("{s:b}", strrchr(mbentry->name, '.')+1));
```
ksmurchison, line 7738: Aren't we missing the boolean value here?  We mustbe getting lucky that its defaulting to 1 (true)

wolfsage: Dang, good call. And this is why I prefer to let the compiler handle printf formatting. From jansson:
```        case 'b': /* boolean */
            return va_arg(*ap, int) ? json_true() : json_false();```

So that's just reading past the argument list into random memory probably!

wolfsage: Fixed

ksmurchison (approved):
Approved, assuming CI passes again

PR #5253, opened 2025-02-04
message.c: do not use strlen to calculate header section size

rsto:
This reverts commit 0f137e7ab8dfbfaa14b0113cc8def0cc9eb2d94d and introduces a different way to handle NUL bytes in MIME messages.

It changes the MIME parser to include any illegal NUL bytes in the header size of a body part. Before, it read those NUL bytes as part of the header section but set the header size only to the count of bytes leading up to the NUL byte.

This mismatch in the stored header size led to an error message indicating that the file size of the message does not match the sum of its header and content size. This was a red herring and now is prevented by setting the correct size and offsets.

Note that the MIME parser did not reject MIME message with NUL bytes, and this patch keeps it at that. This is to keep parsing such messages when they already are stored on disk. New messages with NUL bytes already are rejected at delivery.

### DIFF
--- a/imap/message.c
+++ b/imap/message.c
@@ -111,12 +111,12 @@ static int message_parse_body(struct msg *msg,
                               const char *defaultContentType,
                               strarray_t *boundaries,
                               const char *efname);
-static int message_parse_headers(struct msg *msg,
-                                 struct body *body,
-                                 const char *defaultContentType,
-                                 strarray_t *boundaries,
-                                 const char *efname,
-                                 int *sawboundaryp);
+static void message_parse_headers(struct msg *msg,
+                                  struct body *body,
+                                  const char *defaultContentType,
+                                  strarray_t *boundaries,
+                                  const char *efname,
+                                  int *sawboundaryp);
 
 static void message_parse_address(const char *hdr, struct address **addrp);
 static void message_parse_encoding(const char *hdr, char **hdrp);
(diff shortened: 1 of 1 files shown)


imap/message.c
```diff
@@ -908,7 +897,8 @@ static int message_parse_headers(struct msg *msg, struct body *body,
     }
 
     body->content_offset = msg->offset;
-    body->header_size = strlen(headers.s+1);
+    if (buf_len(&headers) >= body->boundary_size + 1)
+        body->header_size = buf_len(&headers) - body->boundary_size - 1;
```
ksmurchison, line 902: So, I assume body->header_size == 0 otherwise, and that is correct?

Also, unrelated to this PR, can we change the other instances of headers.s to buf_base(&headers)?

rsto: yes, header_size is initialized to zero.

I had wanted to change the use of header.s to buf_base and I still can. But it will introduce more line noise, because we then should also switch to using a const char* instead of a `char*` that points into the buffer. I can do that, but it will change that code more and it's an egg dance already.

ksmurchison: Ok.  Keep it simple for now

ksmurchison (approved):

rsto:
I'm going to pull this back to Draft state and meanwhile just revert the recent MIME parser changes: https://github.com/cyrusimap/cyrus-imapd/pull/5254
Sorry for the noise

rsto:
Closing this one, we'll keep the current implementation for now.

PR #5257, opened 2025-02-04
jmap_mail_submission.c: use SORT_SEQUENCE vs SORT_UID as last criterion

ksmurchison:


wolfsage (changes requested):
Could you please update the commit message to explain why this change is happening?

rsto (approved):
Please add an explanation to the commit message why we are doing this. Otherwise LGTM.

wolfsage (approved):
Thanks

PR #5258, opened 2025-02-05
Ctl conversations basecid

brong:
As discussed this morning, this should do two things:

1) avoid us finding a diff when only the basecid is different in G keys
2) replace binary gunk with human readable output

I did NOT make cyr_dbtool ungunk the binary output because that's not needed for this immediate issue.

### DIFF
--- a/imap/ctl_conversationsdb.c
+++ b/imap/ctl_conversationsdb.c
@@ -537,7 +537,7 @@ static unsigned int diff_records(struct conversations_state *a,
             continue;
         }
 
-        /* both exist an are the same key */
+        /* both exist and are the same key */
         delta = blob_compare(ca.data, ca.datalen, cb.data, cb.datalen);
         if (delta) {
             ndiffs++;
@@ -560,6 +560,7 @@ static int fix_modseqs(struct conversations_state *a,
 {
     int ra, rb;
     struct cursor ca, cb;
+    char buf[80];
     int keydelta;
     int r;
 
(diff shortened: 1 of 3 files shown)


imap/ctl_conversationsdb.c
```diff
@@ -501,6 +501,29 @@ static int next_diffable_record(struct cursor *c)
...
+static void printer(const char *type, struct cursor *c)
+{
+    if (!verbose) return;
+    static struct buf keybuf = BUF_INITIALIZER;
+    static struct buf databuf = BUF_INITIALIZER;
+    buf_reset(&keybuf);
+    buf_reset(&databuf);
+    size_t i;
+    for (i = 0; i < c->keylen; i++) {
+        if (c->key[i] > 31 && c->data[i] < 127)
```
elliefm, line 513: There's a bug here, cf `c->key[i]` in the first term and `c->data[i]` in the second

Also, I think this could just be `if (isprint(c->key[i]))` using isprint() from ctype.h anyway

imap/ctl_conversationsdb.c
```diff
@@ -501,6 +501,29 @@ static int next_diffable_record(struct cursor *c)
...
+            buf_putc(&keybuf, c->key[i]);
+        else
+            buf_printf(&keybuf, "<%d>", (int)c->key[i]);
+    }
+    for (i = 0; i < c->datalen; i++) {
+        if (c->data[i] > 31 && c->data[i] < 127)
```
elliefm, line 519: This one looks bug-free, but could also probably just use isprint()

imap/ctl_conversationsdb.c
```diff
@@ -702,6 +719,25 @@ static int fix_modseqs(struct conversations_state *a,
                 return r;
             }
         }
+        if (ca.key[0] == 'G') {
+            // basecid might be different on the old record, so if they're both v3, then copy the data over
+            if (ca.datalen >= 32 && cb.datalen >= 32 && ca.data[0] == (char)0x83 && cb.data[0] == (char)0x83) {
```
elliefm, line 724: What's this 0x83 magic number? Looks like `'S'`?

imap/ctl_conversationsdb.c
brong: ooh nice.  It should also be casting the value to (uint8_t) and using %u really.   Bah.  But it's quite readable.

imap/ctl_conversationsdb.c
brong: Ouch, yeah that's a copy-paster

imap/ctl_conversationsdb.c
brong: yeah, it's a bit magic.  The version number `if (c & 0x80) version = c & 0x7f`.  I just combined the two.  Sadly the code in conversations.db isn't neatly split into a parser, so it was easier to just hardcode.

brong: I've added a comment

elliefm (approved):

PR #5259, opened 2025-02-05
util.c: use block scope to declare variables in switch case

rsto:
gcc-12 leniently allows this without block scope but clang-14 fails.


wolfsage:
Mm when I wrote this I was surprised I needed different variable names because they clashed. Of course this is why. Thanks

wolfsage (approved):

PR #5260, opened 2025-02-05
mailbox: assert fail if conversations open error

brong:
Right now it can silently just not record things if it fails to open the DB for whatever reason


rsto (approved):

brong:
This is valuable so I'm going to merge it.  I also think we could create the conversationsdb early and open without recreate; so that you can't accidentally have the data path not be present and cause out of sync conversations and mailboxes.

PR #4566, opened 2023-08-11
make 'check' etc fail on cunit valgrind errors

elliefm:
This makes it so that cunit-under-valgrind will fail if valgrind detects memory errors, instead of reporting the valgrind errors to the terminal but still succeeding.  This will be more useful for automated testing.
    
Specifically, this affects the following make targets:
* check
* check-local
* check-%
    
The check-discrete make target already behaved like this, and is no longer a special case.

Also fixes a few memory leaks that had snuck into our unit tests since last time I looked at them.

**Reviewers**: to test this, run `make VG=1 check` after introducing a memory leak somewhere (such as by temporarily reverting one of the testc fixes on this branch).  Expect `echo $?` to be non-zero afterwards.  For comparison, the same on the master branch would report the leak, but `echo $?` would be zero.

PR #5262, opened 2025-02-05
Makefile.am: correct path to libtool

wolfsage:
This fix comes from https://github.com/cyrusimap/cyrus-imapd/pull/5186/

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -698,7 +698,7 @@ cunit/registers.h: $(CUNIT_PROJECT)
 	$(AM_V_GEN)$(CUNIT_PL) --generate-register-function $@
 
 # To run under Valgrind, do: make VG=1 check
-VALGRIND = libtool --mode=execute valgrind \
+VALGRIND = ${abs_top_srcdir}/libtool --mode=execute valgrind \
             --tool=memcheck \
             --leak-check=full \
             --suppressions=${abs_top_srcdir}/cunit/vg.supp \
--- a/doc/internal/unit-tests.html
+++ b/doc/internal/unit-tests.html
@@ -304,7 +304,7 @@ an executable, it's a shell script which sets up some environment variables
 before running the real executable which is hidden away.  This makes debugging
 a failing test somewhat challenging.  The solution is:</p>
 
-<blockquote class="cmds">me@mybox&gt; <b>( cd cunit ; libtool --mode=execute gdb --args unit -t crc32 )</b>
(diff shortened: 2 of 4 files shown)


Makefile.am
```diff
@@ -698,7 +698,8 @@ cunit/registers.h: $(CUNIT_PROJECT)
 	$(AM_V_GEN)$(CUNIT_PL) --generate-register-function $@
 
 # To run under Valgrind, do: make VG=1 check
-VALGRIND = libtool --mode=execute valgrind \
+# The path to libtool is relative to the cunit/ directory
```
elliefm, line 701: This comment is weird?  It might be explaining the `${abs_top_srcdir}` variable, but I don't think that's the detail that needs explaining here, it's just an ordinary part of automake.

IMO the interesting (and surprising to me) detail is that we're now using a libtool from the source directory, rather than the system libtool from `$PATH` (in my case, `which libtool` says `/usr/bin/libtool`).  And we do indeed have a libtool in the source directory!  Looks like it gets generated during configure.  Knowing that it's there, using it smells correct, but I'd never thought about this before.  On the command line, I've always used the system one (e.g. for invoking gdb/ldd/etc on some executable, like `libtool --mode=execute gdb imap/imapd` or whatever), so I should retrain my fingers.

I think we mostly don't invoke libtool directly during build, but rather the stuff generated by autotools does it correctly for us when it has to, and this case is exceptional because these rules were hand-written.  So I do wonder if there's other hand-written invocations of it like this that have been missed....  Scanning the output of `ack libtool`, I don't see any other misuses of it, but there is one spot in our documentation that needs updating too:

```
docsrc/imap/reference/faqs/o-gdb.rst
24:source/build tree), you need to run gdb from ``libtool`` in order for everything
31:    $ libtool --mode=execute gdb imap/cyr_virusscan
```

I think that invocation on line 31 should be `./libtool`.  Can you fix that while you're in here?  Thanks

And given that I now think it's correct to use the source directory libtool and not the system one, I guess the comment that originally bothered me doesn't need to exist at all.  We were using the wrong libtool, now we're using the right libtool, and the commit message adequately explains that.

elliefm (changes requested):
Fascinating! Detailed commentary inline

Makefile.am
wolfsage: I've removed that comment and found a few more references. See what you think? Thanks

elliefm (approved):

elliefm:
**Backporting:** probably grab #4566 too

PR #5265, opened 2025-02-05
Fix Cyrus builds for clang version 19

rsto:
Fixes a couple of issues that showed up compiling Cyrus with clang version 19.


ksmurchison (approved):
WFM

dilyanpalauzov:
These should be backported.

PR #5266, opened 2025-02-05
jmap_ical.c: fix bug when guessing non-standard timezones and improve caching

rsto:
This fixes a bug when guessting non-standard timezones which led to some VTIMEZONEs having non-zero minutes in their UTC offset being mapped to floating time. Related to this, this pull request also adds caching of timezones to CalendarEvent/parse, improving latency for that JMAP method.


ksmurchison (approved):
LGTM

PR #5267, opened 2025-02-06
Userpath

brong:
This lets us specify a path for user/domain data WITHOUT having to use symlinks


brong:
> Why is `char *mboxlist = mboxlist_fname();` added in `list_calendars()` below `/* stat() mailboxes.db for Last-Modified and ETag */`, but the same change is not applied in `list_addressbooks()` and `list_feeds()`?

I guess I missed them, thanks.  It needs to be the same because otherwise it might look at the wrong file.

PR #5268, opened 2025-02-06
cassandane: rename tiny tests to use low line, not hyphen minus

rjbs:


ksmurchison (approved):

cassandane/utils/tiny-test-splitter.pl
```diff
@@ -56,7 +56,6 @@
 
 LINE: while ($_ = shift @lines) {
   if (/^sub test_(\S+)/) {
-    my $file = $1 =~ s/_/-/gr;
```
elliefm, line 59: Does this change work?  Looks to me like `$file` is no longer declared

elliefm (changes requested):

elliefm:
I assume you meant to "request review" rather than "assign", so I've reviewed and dropped my assignment.  If I'm wrong lmk :)

cassandane/utils/tiny-test-splitter.pl
rjbs: Woof, thanks.  Of course this file isn't covered by tests.

rjbs (commented):
I believe I have fixed the silly mistake.

elliefm (approved):

elliefm:
I'm too fast, approved it already but the new CI run hasn't finished yet, heh.  I assume it'll pass because the only thing that was previously bad was untested anyway.

PR #5269, opened 2025-02-06
mboxlist: auditlog subscriptions

brong:
This might be closing the barn door after a horse has bolted, but it's useful to know when things were subscribed or unsubscribed!

### DIFF
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -5331,6 +5331,17 @@ EXPORTED int mboxlist_changesub(const char *name, const char *userid,
         mboxevent_free(&mboxevent);
     }
 
+    if (add) {
+        xsyslog(LOG_NOTICE, "auditlog: subscribe",
+                            "sessionid=<%s> userid=<%s> mailbox=<%s>",
+                session_id(), userid, name);
+    }
+    else {
+        xsyslog(LOG_NOTICE, "auditlog: unsubscribe",
+                            "sessionid=<%s> userid=<%s> mailbox=<%s>",
+                session_id(), userid, name);
+    }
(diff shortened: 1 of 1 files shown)


imap/mboxlist.c
```diff
@@ -5331,6 +5331,11 @@ EXPORTED int mboxlist_changesub(const char *name, const char *userid,
         mboxevent_free(&mboxevent);
     }
 
+    xsyslog(LOG_NOTICE, "auditlog: %s",
+                        "sessionid=<%s> userid=<%s> mailbox=<%s>",
+            (add ? "subscribe" : "unsubscribe"),
+            session_id(), userid, name);
+
```
elliefm, line 5338: This won't work, the first argument to xsyslog is a simple event description, not a printf format string.  This is going to end up logging something like:

```
auditlog: %s: sessionid=<subscribe> userid=<[session id]> mailbox=<[userid]> 
```

It might be nice to add an auditlog() function specifically for audit logging, but for now we don't have one, just use regular syslog.

elliefm (changes requested):
Good thinking.  The thing about closing the barn door is that your other horses don't bolt too!

Won't do what you want as written though, xsyslog isn't suitable for auditlogs

imap/mboxlist.c
brong: I just did the simplest possible dumb thing (split it into two separate xsyslogs, one for each case)

elliefm:
Did you push the wrong thing?  This still has the bad xsyslog call, but _nothing else_

elliefm (approved):

PR #5270, opened 2025-02-06
fix remaining mboxlist usage

brong:
This fixes items missed in https://github.com/cyrusimap/cyrus-imapd/pull/5267 which fixed the mboxlist pathname detection for caldav but not carddav or rss.


brong:
@dilyanpalauzov here's the others :)

elliefm (approved):

PR #5271, opened 2025-02-07
document mboxgroups

### LINKED ISSUE #5252: Document 'mboxgroups' auth backend
Cyrus IMAP 3.12 will have the  'mboxgroups' authentication backend.  As there is no documentation for it, this authentication mechanism cannot be used.

elliefm:
**Reviewers: suggest reading this commit by commit**

* Adds basic documentation for the new `auth_mech: mboxgroups` authorization mechanism.

  This was the only place any authorization mechanisms were documented, so I've added it here.  I'm not convinced this is the _best_ place to document authorization mechanisms generally, but I also don't want to undergo a major restructure of the documentation, so this'll do.  I tidied up the existing stuff a little while I was in there.
* Makes IMAP GETUSERGROUP command admin-only.  With this, a normal user could see any other user's groups, or any group's membership.  That felt wrong so I fixed it.
* Adds tests for the group management commands, cause I needed to figure out how they worked anyway

Closes #5252

### DIFF
--- a/cassandane/Cassandane/Cyrus/Mboxgroups.pm
+++ b/cassandane/Cassandane/Cyrus/Mboxgroups.pm
@@ -97,6 +97,28 @@ sub tear_down
     $self->SUPER::tear_down();
 }
 
+sub imap_getusergroup
+{
+    my ($self, $talk, $item) = @_;
+
+    my $usergroups = {};
+    my $handlers = {
+        'usergroup' => sub {
+            my (undef, $response) = @_;
+
+            my %ug = @{$response};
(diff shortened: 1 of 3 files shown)


rjbs:
This all matches my understanding, but the more valuable reviewer here will be @brong, who wrote the thing being documented!

rjbs (approved):

brong (approved):
This looks great - thanks for documenting it!

I do wonder if we need a MYUSERGROUPS command that every user can use which returns the list of groups that you're a member of.

PR #5273, opened 2025-02-07
Sanitizer ci

wolfsage:
Add an asan make check phase to CI


rsto (approved):

dilyanpalauzov:
While the current change adds ASAN builds, https://github.com/cyrusimap/cyrus-imapd/pull/5288 does include Valgrind checks.  Don’t both things overlap to 100%?  What can the one detect, which cannot be detected by the other?

wolfsage:
I think the answer is "We don't know and won't know for some time", but keeping the valgrind build around is low cost, so worth doing IMO.

PR #5274, opened 2025-02-07
cyr_info: Don't misinterpret bitfields with flags >= 1<<31 or multipl…

wolfsage:
…e names

1<<31 or higher cannot be represented by an integer, so may wrap around or trigger other undefined behaviour. The end result of this was that:

    sieve_extensions: fileinto

reported as:

    sieve_extensions: fileinto vnd.cyrus.jmapquery

Also, some bitfields can have multiple names, and these take up actual slots in the enum_options array. These multi-named options would increment 'j' an extra time, meaning that this:

    sieve_extensions: snooze

would be reported as:

    sieve_extensions: x-cyrus-log

Rather than bit shifting from j, compare directly with the value of the particular flags.

### DIFF
--- a/imap/cyr_info.c
+++ b/imap/cyr_info.c
@@ -143,7 +143,10 @@ static void do_conf(int only_changed, int want_since, uint32_t since)
                     highlight(imapopts[i].last_modified);
                 printf("%s:", imapopts[i].optname);
                 for (j = 0; imapopts[i].enum_options[j].name; j++) {
-                    if (imapopts[i].val.x & (1<<j)) {
+                    /* multiple names? Use only the non-legacy (first) one */
+                    if (j && imapopts[i].enum_options[j].val == imapopts[i].enum_options[j-1].val)
+                        continue;
+                    if (imapopts[i].val.x & imapopts[i].enum_options[j].val) {
                         printf(" %s", imapopts[i].enum_options[j].name);
                     }
                 }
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):

imap/cyr_info.c
```diff
@@ -143,7 +143,10 @@ static void do_conf(int only_changed, int want_since, uint32_t since)
                     highlight(imapopts[i].last_modified);
                 printf("%s:", imapopts[i].optname);
                 for (j = 0; imapopts[i].enum_options[j].name; j++) {
-                    if (imapopts[i].val.x & (1<<j)) {
+                    /* multiple names? Use only the non-legacy (first) one */
+                    if (j && imapopts[i].enum_options[j].val == imapopts[i].enum_options[j-1].val)
+                        continue;
+                    if (imapopts[i].val.x & imapopts[i].enum_options[j].val) {
```
elliefm, line 149:
```suggestion
                    if ((imapopts[i].val.x & imapopts[i].enum_options[j].val)) {
```

When using the truthiness of a bitwise operation in a condition like this, I find it helps readability to put the operation in an extra set of parens.  Makes it look weird, so you read it more closely, and don't misread the `&` as `&&` (or `|`/`||`), plus the extra level of typing effort conveys intentionality, clarifying that the `&` isn't a typo either.

Once upon a time I used to get compiler warnings suggesting this practice, and I liked the suggestion and adopted it as habit.  I haven't seen those warnings in a while, so I wonder if there's some warning option I used to use and forgot to copy over to a new machine at some point...

This isn't a strict Cyrus style thing, it's just a me thing, take it or leave it :)

wolfsage: I think I'll leave it. The extra parens would surprise me.

-Wall includes -Wparentheses which will warn if you try this:

    if (x & 1 == 0)

because that's really

    if (x & (1 == 0))

I don't know if there's a more noisy one

PR #5282, opened 2025-02-12
jmap_contact.c: populate JSContact members with the proper UIDs

ksmurchison:


cassandane/tiny-tests/JMAPContacts/contactgroup_set_cardgroup_get
```diff
@@ -0,0 +1,43 @@
...
+    xlog $self, "create contact group";
+    $res = $jmap->CallMethods([['ContactGroup/set', {
+        create => {
+            "1" => {
+                name => "group1",
+                contactIds => [$contact1, $contact2]
+            }
+        }
+    }, "R2"]]);
+    $self->assert_not_null($res);
```
rsto, line 28: nit: This line and the next 2 lines are unnecessary noise. We already have tests that assert that Cyrus adheres to the core spec of JMAP, there is no need to reassert this in every test.

cassandane/tiny-tests/JMAPContacts/contactgroup_set_cardgroup_get
```diff
@@ -0,0 +1,43 @@
...
+    $self->assert_str_equals('ContactGroup/set', $res->[0][0]);
+    $self->assert_str_equals('R2', $res->[0][2]);
+    my $id = $res->[0][1]{created}{"1"}{id};
+
+    xlog $self, "get contact group $id";
+    $res = $jmap->CallMethods([['ContactCard/get', { ids => [$id] }, "R3"]]);
+    $self->assert_not_null($res);
```
rsto, line 35: nit: same, unnecessary noise

rsto (approved):
This LGTM, apart for a few nits in the test. For ContactCard, we do not expect the JMAP id of the Card to be equal to the uid, right?

ksmurchison:
> This LGTM, apart for a few nits in the test. For ContactCard, we do not expect the JMAP id of the Card to be equal to the uid, right?

We don't, but until we switch to nanosecond-based IDs for contacts, they will continue to be the same.

cassandane/tiny-tests/JMAPContacts/contactgroup_set_cardgroup_get
ksmurchison: Right.  Fixed.

PR #5283, opened 2025-02-12
twom: if file is dirty, run recovery even if opened "SHARED"

brong:


rsto (approved):

brong:
This was added in another PR along with the fixes to it.

PR #5288, opened 2025-02-14
.github/workflows/main: Add a valgrind make check ci build

wolfsage:


wolfsage:
Needs https://github.com/cyrusimap/cyrus-docker/pull/31/files

wolfsage:
(The check will fail right now because valgrind isn't in the docker build (yet))

.github/workflows/main.yml
```diff
@@ -107,3 +107,39 @@ jobs:
...
+        git fetch --no-tags upstream 'refs/tags/cyrus-imapd-*:refs/tags/cyrus-imapd-*'
+    - name: configure and build
+      shell: bash
+      run: |
+        echo "building cyrus version" $(./tools/git-version.sh)
+        ./tools/build-with-cyruslibs.sh
+    - name: make check with valgrind
+      shell: bash
+      run: |
+        make VG=1 check
```
wolfsage, line 139: check discreet takes 18m. The regular check is much faster and the output is readable enough to find the real errors I think

rjbs (approved):
Reviewed.  Did a manual run.  Came very close to just merging, but decided maybe you want ellie to take a proper look.  But I think it looks good to go and the docker image is ready for it. 👍🏽

.github/workflows/main.yml
elliefm: Yeah that seems fine.  In my mind, the main utility of running valgrind in CI is regularly knowing _if_ something is leaky.  Once you know something's leaky, you're probably going to do targeted local valgrind runs in order to track down and fix the leak, so the CI output doesn't really need to be diagnostically useful in its own right.

.github/workflows/main.yml
```diff
@@ -107,3 +107,39 @@ jobs:
...
+    - name: report version information
```
elliefm, line 140: I don't think it matters, but it surprised me to have the version information at the end instead of before the tests. I guess this is probably the same order as the regular CI ("after build and check, before cassandane"), and the difference is that this one doesn't run cassandane

elliefm (approved):

.github/workflows/main.yml
wolfsage: Yeah... I've been mostly copying what we have and pinning it down. I tried to find a way to make this all reusable but it doesn't seem easy or obvious to modularize this stuff. /shrug

PR #5293, opened 2025-02-18
Twom fixes

brong:
This fixes some issues found with twom in production:
* full disk can cause the mmap to create 16k of blanks then fail to write the headers!  Use write instead for initial header and dummy
* readonly database can read a dirty file after a crash, meaning that it returns errors, always open RDWR and mmap readwrite.

Plus just makes the dump format a bit more regular.

### DIFF
--- a/lib/twom.c
+++ b/lib/twom.c
@@ -1811,7 +1811,7 @@ static int read_lock(struct twom_db *db, struct twom_txn **txnp,
         r = unlock(db, file);
         if (r) goto done;
 
-        int newfd = open(db->fname, db->readonly ? O_RDONLY : O_RDWR, 0644);
+        int newfd = open(db->fname, O_RDWR, 0644);
         if (newfd == -1) {
             db->error("read_lock open failed",
                       "filename=<%s>", db->fname);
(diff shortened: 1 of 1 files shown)


lib/twom.c
```diff
@@ -2962,7 +2962,16 @@ static int twom_txn_dump(struct twom_txn *txn, int detail)
             printf("COMMIT start=%08llX\n", (LLU)NEXT0(ptr, 0));
         }
         else if (type == DELETE) {
-            printf("DELETE ancestor=%08llX\n", (LLU)NEXT0(ptr, 0));
+            size_t parent_offset = NEXT0(ptr, 0);
+            const char *key = KEYPTR(loc->file->base + parent_offset);
+            int len = KEYLEN(loc->file->base + parent_offset);
```
rsto, line 2967: Signed integer overflow is undefined behavior. In practice, we could expect len to become negative if KEYLEN returns a very lange length, and the conditional in the next line won't catch that. Better to avoid either.

rsto (approved):

lib/twom.c
brong: yeah ,OK - it should be a size_t.  I just copied this from the other one nearby, so I'll fix them all.

PR #5294, opened 2025-02-18
index.c: use "utf-8" as default charset for vCard data

rsto:
A MIME body part having Content-Type: text/vcard without the charset parameter would otherwise be decoded as US-ASCII, leading to any non-ASCII UTF-8 octets being replaced with the UTF-8 replacement character.


brong (approved):
This looks reasonable.  I guess it's OK to treat us-ascii as UTF-8 since they'll be the same for all "legal" chars

PR #5296, opened 2025-02-18
Twom fix compar

brong:


wolfsage (approved):
Approved with minor nit: commit messages should explain their rationale

PR #5297, opened 2025-02-18
jmap_contact.c: properly handle VALUE=X param on text-valued properties

ksmurchison:
vcardproperty_get_prodid() will return NULL if the property contains VALUE=X because the value type isn't TEXT and is stored in a different field of struct vcardvalue.

This patch treats VALUE=X as a text string


rsto (approved):
LGTM, but where is VALUE=X even coming from?

PR #5298, opened 2025-02-18
Add test_restore_calendars_scheduling to check iMIP processing after restore

ksmurchison:


rjbs (approved):
This seems right.  One thought:  I might make this use two calendars.  I don't know enough about the internals, but I can imagine a world where we have some obnoxious bug about an event being deleted/restored more than once not doing iMIP and that's why it's passing.  That said:  I doubt this is really a problem.  Then again, I think making that happen is probably about 10 more LOC.

ksmurchison:
> This seems right. One thought: I might make this use two calendars. I don't know enough about the internals, but I can imagine a world where we have some obnoxious bug about an event being deleted/restored more than once not doing iMIP and that's why it's passing. That said: I doubt this is really a problem. Then again, I think making that happen is probably about 10 more LOC.

Sure.  I can use a second calendar.

PR #5299, opened 2025-02-18
master.c: Actually wait (at least) 10 seconds before attempting to restart crashing children

wolfsage:


cassandane/Cassandane/Cyrus/Master.pm
```diff
@@ -1399,4 +1406,54 @@ sub test_daemon_exits
                                  qr{too many failures for});
 }
 
+sub test_babysit
+{
+    my ($self) = @_;
+
+    $self->{instance}->{no_check_syslog} = 1;
+
```
elliefm, line 1414: There's a magic attribute for this already, something like this should do:

```suggestion
sub test_babysit
    :NoCheckSyslog
{
    my ($self) = @_;

```

Though maybe you tried that already and it _didn't_ work, which would be very interesting to hear about

cassandane/Cassandane/Instance.pm
```diff
@@ -1745,7 +1745,7 @@ sub stop
     push @errors, $self->_check_sanitizer_logs("asan");
     push @errors, $self->_check_sanitizer_logs("ubsan");
     push @errors, $self->_check_cores();
-    push @errors, $self->_check_syslog() unless $params{no_check_syslog};
+    push @errors, $self->_check_syslog() unless $params{no_check_syslog} || $self->{no_check_syslog};
```
elliefm, line 1748: per above, if the magic attribute works properly, then this isn't necessary

master/master.c
```diff
@@ -2839,22 +2839,7 @@ static void reread_conf(struct timeval now)
...
-        }
-
-        while (j-- > 0) {
-            spawn_service(s, si, wdi);
-        }
-    } else if (s->exec
+    if (s->exec
                 && s->babysit
                 && s->nactive == 0) {
         if (s->nreadyfails >= MAX_READY_FAILS) {
```
elliefm, line 2845: Style nit:

```suggestion
    if (s->exec
        && s->babysit
        && s->nactive == 0)
    {
        if (s->nreadyfails >= MAX_READY_FAILS) {
```

* align the terms in the condition
* since this puts the condition terms at the same indent level as the block body (due to `if (` being four chars), shunt the `{` to its own line to make them visually distinct

elliefm (approved):
Do I understand correctly that the main part of this change is switching the order, so that check_undermanned checks the babysat children before it checks general ones?  I guess previously the more general check was accidentally restarting the babysat ones too, and so the babysit-specific handling was never firing.  And then there was a small bug in the babysit handling that was never seen cause it never ran.  Good catch!

I've left suggestions for a better way to disable syslog checks per-test, and a style nit in the C.  Marking this "Approve" rather than "Request changes" though, because I don't think this needs to round-trip back through me again before it lands.

elliefm:
(I just wrote up the 3.12 release notes entry for this feature a couple of days ago, hah.  Didn't realise it didn't actually work, glad it now does!)

cassandane/Cassandane/Cyrus/Master.pm
wolfsage: Ah! I had found another test use no_check_syslog => 1 in stop() so figured that was the state of the art. Thank you, fixed

wolfsage:
> Do I understand correctly that the main part of this change is switching the order

Yup! (This is explained in the commit message)

cassandane/Cassandane/Instance.pm
wolfsage: Yup, fixed

cassandane/Cassandane/Cyrus/Master.pm
elliefm: Ahh yep.  Useful rule of thumb: any time you need to do something unusual at the start of a test, have a look at Cassandane::Cyrus::TestCase and see if there's already a magic for it :)

PR #5303, opened 2025-02-19
Ubsan fixes

brong:
I got sick of compiling with ubsan and then not being able to run Cassandane tests, so this fixes a lot of the noise!


wolfsage (approved):
lgtm. Neat trick with the union for aligned reads. Probably we could use alignas at this point instead but it seems fine.

Thanks for this!

brong:
> lgtm. Neat trick with the union for aligned reads. Probably we could use alignas at this point instead but it seems fine.
> 
> Thanks for this!

Yeah, this trick was used in places in the code before (may still be) - have fixed a couple more now.

PR #5304, opened 2025-02-20
Cassandane: use config for all DB types

brong:
Remove all the 'twoskip' hard-coded values (this should basically be a NOOP since it sets the same defaults; but you can more easily change them in cassandane.ini without having the affected tests behave badly.


rjbs (approved):
👍🏽

Not sure I love the second commit but (a) I don't have a better suggestion atm and (b) it gets the rest of the work in place, which is really good.

PR #5307, opened 2025-02-21
util.c: bin_to_lchex needs an aligned hex buffer

wolfsage:
bin_to_lchex casts a char* to a uint16_t*, then uses that pointer to write 2 bytes at a time. This requires an aligned pointer, so we'll use malloc to give us a scratch space if the incoming hex pointer is unaligned.

malloc is gauranteed to return aligned pointers for any built-in types.

With this, gcc 12.2.0 (Debian 12.2.0-14) with -fsanitize=undefined (UBSan) runs clean.


rjbs (approved):

brong (changes requested):
I'd prefer to handle either both; or at least the misaligned case; with per-character logic:
```
    const char *xd = "0123456789abcdef";
    uint8_t *v = (uint8_t *)bin;
    char *p = hex;
    size_t i;
    for (i = 0; i < binlen; i++) {
        *p++ = xd[(v[i] >> 4) & 0xf];
        *p++ = xd[v[i] & 0xf];
    }
    *p = '\0';
```

elliefm:
You could do one single-byte read before the main loop if necessary (i.e. if the original pointer address has the least significant bit set), and one more after if necessary, and still use aligned two-byte reads for the bulk of the work, without needing to copy to scratch space.  But IRL we're on 64 bit platforms, not 16, so if we're optimising like this, it's probably better to do four- or eight-byte reads, with up-to-3 or up-to-7 leading/trailing bytes...

But you could also just do the whole thing with single-byte reads like Bron suggested, and suppose that this pattern is probably common enough for the compiler to choose an appropriate optimisation for the target architecture.

wolfsage:
Alright, I've updated this MR to remove bin_to_lchex and use bin_to_hex instead

ksmurchison (approved):

rsto (approved):
LGTM!

PR #5308, opened 2025-02-24
always build/install/distribute manpages

### LINKED ISSUE #4797: always build/install manpages
We have a bunch of conditionalisation in Makefile.am to try to make it so that we don't build/install man pages for features that were disabled during configure.  But it's brittle, and causes headaches (e.g. https://github.com/cyrusimap/cyrus-imapd/pull/3867#issuecomment-1902759455)

We should remove the conditionalisation, and just have all our manpages always built and installed.  That'll be easier to maintain, and more robust in use.

elliefm:
Removes a bunch of brittleness from Makefile.am by not applying feature-conditional logic to man pages.  Closes #4797.

Adds a couple of newer man pages to what's installed/distributed which had been overlooked previously.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1607,6 +1607,7 @@ imap_cvt_xlist_specialuse_LDADD = $(LD_UTILITY_ADD)
 @SET_MAKE@
 
 dist_man1_MANS = \
+    man/dav_reconstruct.1 \
     man/imtest.1 \
     man/installsieve.1 \
     man/httptest.1 \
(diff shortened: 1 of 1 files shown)


elliefm:
**backporting:** Must backport at least the first commit to 3.12, but probably just do both.

wolfsage:
It feels weird to install man pages for binaries that aren't there. Do packages (debian, etc) care?

I've approved, but I wonder if it would be better to use awk to remove deuplicates in the manlist instead...

wolfsage (approved):

elliefm:
> It feels weird to install man pages for binaries that aren't there. Do packages (debian, etc) care?

It does feel weird, but what we had has been a maintenance headache.  Long term, most of these build options will become non-optional, at which point the issue will be moot anyway.  I don't know if packagers care.
 
> I've approved, but I wonder if it would be better to use awk to remove deuplicates in the manlist instead...

That sounds like it would be just as brittle as what we had 😅 I suspect packagers can probably include/exclude specific files much more easily than we can, so if they know they've built it without (say) nttpd, then could exclude that man page from their package.

brong (approved):

Issue #5310, opened 2025-02-24
index_expunge optimization, do not run if there are no deleted messages


zisisvakras:
This issue references this todo (in imap/index.c:414 index_expunge): 
```c
/* XXX - check if not mailbox->i.deleted count and need_deleted */
```

Check the index header for deleted flagged messages, if there are none then avoid executing rest of expunge...

PR #5311, opened 2025-02-24
index_expunge optimization

zisisvakras:
do not run if there are no deleted messages

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -411,8 +411,8 @@ EXPORTED int index_expunge(struct index_state *state, const char *sequence,
     r = index_lock(state, /*readonly*/0);
     if (r) return r;
 
-    /* If mailbox has no deleted flagged messages return */
-    if (!(state->mailbox->i.deleted > 0 && need_deleted)) {
+    /* Make sure there is something to expunge */
+    if (!sequence && !(state->mailbox->i.deleted > 0 && need_deleted)) {
         index_unlock(state);
         return r;
     }
--- a/imap/index.c
+++ b/imap/index.c
@@ -411,7 +411,12 @@ EXPORTED int index_expunge(struct index_state *state, const char *sequence,
     r = index_lock(state, /*readonly*/0);
     if (r) return r;
 
-    /* XXX - check if not mailbox->i.deleted count and need_deleted */
+    /* If mailbox has no deleted flagged messages return */
+    if (!(state->mailbox->i.deleted > 0 && need_deleted)) {
+        index_unlock(state);
+        return r;
+    }
+    
     seq = _parse_sequence(state, sequence, 1);
 
     mboxevent = mboxevent_new(EVENT_MESSAGE_EXPUNGE);
--- a/imap/index.c
+++ b/imap/index.c
@@ -414,7 +414,7 @@ EXPORTED int index_expunge(struct index_state *state, const char *sequence,
     /* Make sure there is something to expunge */
     if (!sequence && !(state->mailbox->i.deleted > 0 && need_deleted)) {
         index_unlock(state);
-        return r;
+        return 0;
     }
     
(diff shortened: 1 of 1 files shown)


zisisvakras:
this PR references issue #5310

elliefm (changes requested):
Looks like this is causing a bunch of Cassandane test failures (list below).  I guess the optimisation here is not as simple as the old comment suggested.

I'm marking this "request changes" because it's not suitable to merge as is, but I can't tell what specifically is wrong with it, so I can't suggest what you could change to make it work, sorry!

```
[FAILED] Cyrus::ImapTest.move
[FAILED] Cyrus::JMAPMailbox.mailbox_counts_add_remove
[FAILED] Cyrus::JMAPEmail.email_get_multimailboxes_expunged
[FAILED] Cyrus::JMAPEmail.email_query_shared_move
[FAILED] Cyrus::JMAPEmail.thread_get_shared
[FAILED] Cyrus::JMAPMailbox.mailbox_trash_counts_ondelete
[FAILED] Cyrus::JMAPEmail.email_querychanges_deletedcopy
[FAILED] Cyrus::JMAPMailbox.mailbox_counts
[FAILED] Cyrus::Rename.rename_inbox
[FAILED] Cyrus::Rename.rename_inbox_unselected
[FAILED] Cyrus::Replace.replace_different_mailbox
[FAILED] Cyrus::Replace.replace_same_mailbox
[FAILED] Cyrus::MurderIMAP.replace_same_backend
[ERROR ] Cyrus::UIDonly.copy_move
[FAILED] Cyrus::Notify.change_selected
[FAILED] Cyrus::Notify.idle
```

imap/index.c
```diff
@@ -411,7 +411,12 @@ EXPORTED int index_expunge(struct index_state *state, const char *sequence,
     r = index_lock(state, /*readonly*/0);
     if (r) return r;
 
-    /* XXX - check if not mailbox->i.deleted count and need_deleted */
+    /* If mailbox has no deleted flagged messages return */
+    if (!(state->mailbox->i.deleted > 0 && need_deleted)) {
+        index_unlock(state);
```
ksmurchison, line 416: I think this should be:

`if (!seq && !(state-> ...)`

zisisvakras: I run Cassandane on the failed tests again with this fix and it's working :+1:

imap/index.c
```diff
@@ -411,7 +411,12 @@ EXPORTED int index_expunge(struct index_state *state, const char *sequence,
     r = index_lock(state, /*readonly*/0);
     if (r) return r;
 
-    /* XXX - check if not mailbox->i.deleted count and need_deleted */
+    /* Make sure there is something to expunge */
+    if (!sequence && !(state->mailbox->i.deleted > 0 && need_deleted)) {
+        index_unlock(state);
+        return r;
+    }
```
ksmurchison, line 418: Not a huge deal, but this could probably just be `return 0;`

zisisvakras: Yeah if execution flow gets there r is 0, might as well do it then

ksmurchison (approved):

ksmurchison:
@elliefm This looks good to me.  Do you concur?

brong (approved):
Yep, that's the thing!  This is a nice optimisation for the general "close" case.

ksmurchison:
> Actually, as discussed in the Cyrus call today - we could put this test BEFORE the lock, because if you haven't noticed that the DELETED count has gone up, then you aren't the one who needs to do the expunge anyway, so we could avoid the entire lock cost.

Actually, we CAN'T do this because state->mailbox is NULL as we enter index_expunge().  index_lock() opens the mailbox for us.

This PR is good to go as-as.

PR #5315, opened 2025-02-26
master deadlock on shutdown

elliefm:


dilyanpalauzov:
Isn't the master deadlock on shutdown already solved by https://github.com/cyrusimap/cyrus-imapd/pull/5319?

PR #5316, opened 2025-02-26
twom.c: don't truncate file->size during recovery

rsto:
This fixes a bug where aborting a twom transaction followed by closing the database leaves a dangling mmapped region as long as the process still is running.

The bug is the result of truncating the file->size field in the recovery1 function. If the file has previously been mmapped to a larger region than that new size, the munmap following a db_close leaves the remaining region mapped for the process.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Conversations.pm
+++ b/cassandane/Cassandane/Cyrus/Conversations.pm
@@ -1303,4 +1303,45 @@ sub test_rename_user_nosplitconv
     $self->check_replication('newuser');
 }
 
+sub test_unmap_failed_appends
+    :NoCheckSyslog
+{
+    my ($self) = @_;
+    my $imap = $self->{store}->get_client();
+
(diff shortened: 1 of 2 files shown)


rsto:
@brong @wolfsage This fixes the issue at hand with munmap leaving dangling memory map regions, but I'm not entirely convinced it doesn't break other code flows that build on the recovery1 function.

brong:
Yes, this is fine - the comment was wrong - it doesn't cause that, it causes it to fail to unmap the old mmap entirely :( dammit.  Better to just not truncate the map, but maybe also to not even truncate!

brong (approved):
Approved along with another commit which also removes the truncate - we don't need it really, and it introduced the mmapping bug along with it.  This way there will be junk
left after the file when a commit fails, until it gets overwritten with new transactions, but that's actually OK.  This was a last-minute fiddle.

PR #5317, opened 2025-03-02
cron-style scheduling of cyrus.conf EVENTS

elliefm:
**Reviewers:** The diffs will make more sense when read commit-by-commit.

This PR adds cron-style scheduling of cyrus.conf EVENTS.  It takes a bit of a walk on the way there, the highlights are:

* rips all the existing event handling out of master.c and into a testable module (with minimal changes), and adds tests for it
* updates the new module to have a saner interface, and fixes some old bugs
* adds a parser for cron-style date/time specifications
* adds a new cronevent module based around the new parser
* updates master to support cronevents alongside the existing events
* changes the at=hhmm event handling to use the cronevent stuff under the hood (the at=hhmm cyrus.conf syntax is still supported, so no action is needed during upgrade)

The cron parser is based on the syntax described by `crontab(5)` on my laptop.  I believe this is pretty standard.

Cron events consist of a set of bitfields describing the times the event wants to run (`struct cron_spec`), plus the name of the event and the command to execute (`struct cronevent_details`).  These two structs are maintained in separate, parallel arrays.  Once per minute, the cron_specs for all events are examined in a hot loop to see which (if any) match the current time.  Most minutes, no events will match, so we won't even look at the cronevent_details.  When an event does match, its corresponding cronevent_details are fetched and the command executed.

If Fastmail converted our existing cron jobs into cron-style cyrus.conf EVENTS, there would be about 16 such events (based on role_imaprepl.tt2).  Given the small number, I chose the hot loop check-everything approach, relying on CPU cache prefetching of sequentially-accessed memory to keep it fast, rather than devising some more complicated lookup scheme.  The cron_spec data being isolated in its own array, separate from the cronevent_details, means prefetches in the hot loop aren't wasted on data that probably won't be used.

When initialised, the cronevent subsystem registers a periodic wakeup call timed to the start of the minute, using the existing event system.  This means cronevents will typically run as close to the start of their chosen minute as possible.  The wakeup call setup is similar to what's already used by the janitor and prom_report subsystems, which want to run every ten seconds, though the alignment to the top of the minute is novel.

### DIFF
--- a/docsrc/imap/reference/manpages/configs/cyrus.conf.rst
+++ b/docsrc/imap/reference/manpages/configs/cyrus.conf.rst
@@ -225,19 +225,35 @@ scheduled cleanup/maintenance.
 
 ..
 
-        The interval (in minutes) at which to run the command.  This
-        integer value is optional, but SHOULD be a positive integer >
-        10.
+        The integer interval (in minutes) at which to run the command.
+
+        Intervals are counted starting from when :cyrusman:`master(8)` was
+        started, they are not aligned to any particular clock time.
+
+        This is optional, but one of **period** or **cron** must be specified.
 
 .. parsed-literal::
 
-    **at=**\ <hhmm>
+    **cron=**\ "min hr dom mon dow"
+
+..
+
+        A string containing five space-delimited fields describing the minutes,
+        hours, days-of-month, months, and days-of-week that this command should
+        run, just like :manpage:`crontab(5)`.  The usual cron features are
+        supported: asterisks, ranges, lists, step values, named months
+        (jan-dec), and named days of the week (sun-sat).
+
+        This is optional, but one of **period** or **cron** must be specified.
+
+.. parsed-literal::
+
+    **at=**\ hhmm
 
 ..
 
         The time (24-hour format) at which to run the command each day.
-        If set to a valid time (0000-2359), period is automatically
-        set to 1440. This string argument is optional.
+        This is an alias for ``cron="mm hh * * *"``.
 
 DAEMON
 ------
@@ -320,8 +336,8 @@ Examples
 
     EVENTS {
         checkpoint    cmd="ctl_cyrusdb -c" period=30
-        delprune      cmd="cyr_expire -E 3" at=0400
-        tlsprune      cmd="tls_prune" at=0400
+        delprune      cmd="cyr_expire -E 3" cron="0 4 * * *"
+        tlsprune      cmd="tls_prune" cron="0 4 * * *"
     }
 
     DAEMON {
@@ -353,4 +369,5 @@ See Also
 :cyrusman:`ctl_cyrusdb(8)`,
 :cyrusman:`ctl_deliver(8)`,
 :cyrusman:`tls_prune(8)`,
+:manpage:`crontab(5)`,
 :manpage:`hosts_access(5)`
--- a/docsrc/imap/reference/manpages/configs/cyrus.conf.rst
+++ b/docsrc/imap/reference/manpages/configs/cyrus.conf.rst
@@ -219,6 +219,11 @@ scheduled cleanup/maintenance.
         The command (with options) to spawn as a child process.  This
         string argument is required.
 
+        Note that this argument is parsed quite naively: arguments are
+        split on whitespace, quoting is not supported.  If you need to run
+        a complex command, make a shell script that calls the real command,
+        and set the event to run the script.
+
 .. parsed-literal::
 
     **period=**\ 0
--- a/master/cronevent.c
+++ b/master/cronevent.c
@@ -63,7 +63,10 @@ static dynarray_t cronevent_details
 
 static time_t cronevent_last_run_time = 0;
 
-EXPORTED void cronevent_add(const char *name, const char *spec, const char *cmd)
+EXPORTED void cronevent_add(const char *name,
+                            const char *spec,
+                            const char *cmd,
+                            bool ignore_err)
 {
     struct cron_spec cron_spec = {0};
     struct cronevent_details *details = NULL;
@@ -74,12 +77,14 @@ EXPORTED void cronevent_add(const char *name, const char *spec, const char *cmd)
         xsyslog(LOG_ERR, "event missing name",
                          "spec=<%s> cmd=<%s>",
                          spec, cmd);
+        if (ignore_err) return;
         fatal("event missing name", EX_CONFIG);
     }
 
     if (!cmd || !*cmd) {
         xsyslog(LOG_ERR, "event missing cmd",
                          "name=<%s>", name);
+        if (ignore_err) return;
         fatal("event missing cmd", EX_CONFIG);
     }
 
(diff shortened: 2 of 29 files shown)


elliefm:
This won't do cron-style _execution_ of EVENTS, though.  The parsing and handling of `cmd="..."` is unchanged from the original, which just does a naive split on whitespace into an array of tokens, which are passed to `execv(3)`.  I believe this means that complicated commands, such as ones with parameters containing quoted whitespace, won't work properly.  I might add a test to prove that it's broken.

Improving the parsing of `cmd="..."` could be a future extension, but in the meantime the shortcoming can be worked around by writing a small wrapper script containing the real command, and having the event invoke the script.

master/event.c
```diff
@@ -199,25 +199,23 @@ EXPORTED struct event *schedule_peek(void)
...
-        next->next = due;
-        due = next;
+    due = schedule;
+    while (due && timesub(&now, &due->mark) <= 0.0) {
+        last_due = due;
+        due = due->next;
     }
 
-    return due;
+    if (last_due) {
```
wolfsage, line 168: Ah this is all way better

master/cronevent.c
```diff
@@ -0,0 +1,168 @@
...
+    if (!name || !*name) {
+        xsyslog(LOG_ERR, "event missing name",
+                         "spec=<%s> cmd=<%s>",
+                         spec, cmd);
+        fatal("event missing name", EX_CONFIG);
+    }
+
+    if (!cmd || !*cmd) {
+        xsyslog(LOG_ERR, "event missing cmd",
+                         "name=<%s>", name);
```
wolfsage, line 86: Kinda weird we don't log spec=... here like we do if name is missing

docsrc/imap/reference/manpages/configs/cyrus.conf.rst
```diff
@@ -219,6 +219,11 @@ scheduled cleanup/maintenance.
         The command (with options) to spawn as a child process.  This
         string argument is required.
 
+        Note that this argument is parsed quite naively, only trivial commands
+        are likely to work.  If you need to run anything complex here, make a
+        shell script that calls the real command, and set the event to call
+        the script.
```
wolfsage: It might be nice if this was more explicit about what won't work, ie "Don't use spaces, or quote characters, or..."

wolfsage (commented):
Initial review, more to come.

Looks pretty good!

master/event.c
```diff
@@ -0,0 +1,177 @@
...
+EXPORTED void reschedule_event(struct event *evt, struct timeval now)
+{
+    time_t now_s = now.tv_sec;
+    time_t period = evt->period;
+    time_t mark = MIN(now_s, evt->mark.tv_sec);
+
+    assert(period > 0);
+
+    /* don't fall behind schedule if we're running slow for some reason */
+    while (mark <= now_s)
```
wolfsage: I wonder if this is worth a syslog line "Hey for some reason we processed this event behind by %s seconds..."

Should not happen right?

master/cronevent.c
elliefm: If I saw this error, the thing I'd want to know is, which event did I forget to add a command for?  And the name is enough to me to find it; the spec isn't interesting.

On the other hand, when the name is missing (apart from that possibly being impossible due to the way masterconf parses the file, idk), then I can't find the culprit by looking up the name, so every other detail is logged in the hope that it's helpful.

docsrc/imap/reference/manpages/configs/cyrus.conf.rst
elliefm: I feel like leaving it ambiguous and "here be dragons"-ey means I'm not accidentally implying that anything I didn't list works.  I haven't actually thought much about the set of what-won't-work beyond observing that it's not the empty set, and the issue isn't new (but does become more likely to be tripped over as people stick more arbitrary non-Cyrus stuff into cyrus.conf EVENTS).

rsto: Is this additional remark even necessary then? Anyone using this feature will have to try anyway.

wolfsage (approved):

master/event.c
elliefm: Before this change, I believe it was plausible for the wallclock times of a periodic event ("every x seconds") to drift over time from cumulative small timing errors.  But I don't believe it was possible for any individual event to be substantially delayed on its own (unless some other change accidentally introduces a stall somewhere).

This change stops the errors from accumulating in the first place, but since we're now considering only a single error at a time, I don't expect it to ever be much more than a single second.  It _might_ be worth logging in a "see if it ever happens" way, but I don't have any feeling for what threshold to log at.  Too high and we'll never log anyway, too low and it'll be noise.

elliefm: For the purposes of the new cron stuff, the periodic event driving it is only a wakeup call anyway.  It will make sure we wake up from pselect near the top of the minute if we weren't already awake.  But the cron stuff will kick off at the first wakeup after the minute changes, regardless of whose wakeup it was, and janitor and prom_update have every-10-seconds wakeups, so pselect never blocks for more than 10 seconds anyway.  And the non-cron stuff never cared about potential drift.

elliefm: That said, I'm uncomfortable with "this shouldn't happen" without being able to prove it to myself, so maybe I will log, and then I'll have proof one way or another...

elliefm:
@wolfsage Thanks to your comment about logging, I convinced myself to log when rescheduled periodic events have skipped entire periods.  I've also rebased on current master, but the only real change is in the newest commit.  Do you want to give that a look?  Thanks

docsrc/imap/reference/manpages/configs/cyrus.conf.rst
wolfsage: My problem with this warning is it tells a user "This may surprise you!" but does not at all provide any sort of guidance or expectations. I think "Well, I'll just use a shell script. But wait? Is it so naive I can't put a _ or a - in the name of my script? I guess I have to experiment to find out... how frustrating".

elliefm: Ahh yeah.

So, I've just (or am about to) referred the reader to crontab(5) for the definition of the date/time fields...

Real cron runs its commands by passing the remainder of the line to the `system()` call, which means the command string is parsed by `/bin/sh`; and crontab(5) is full of examples of doing all kinds of tricky things using the full power of the real shell parser.

Cyrus master can't just call `system()`, because it mustn't block waiting for the command to finish.  We parse the command ourselves (simply splitting on whitespace) and do our own fork/exec/babysit/waitpid.  This is no different from before.

My concern is that someone who's just been referred to crontab(5) for the date/time specification might mistake that referral for meaning Cyrus events' command parsing is similarly sophisticated.  I'm trying to communicate that, nope, it's the same old thing you know and love/loathe, just the date/time handling is better.  I'm not trying to document the command parsing, just pointing out that it works the same as it always did.

The right place for the command parsing to be described would be much earlier in the document (since the same parsing is used for every other section too, not just EVENTS).  It doesn't exist, so there's nothing to update, but since I'm not changing how it works, it wouldn't need an update anyway.

elliefm: I've rephrased this paragraph to hopefully clarify exactly what's naive about it.

wolfsage: Thanks!

PR #5020, opened 2024-09-05
sieve_eval_bc:B_REDIRECT make the function reentrant

dilyanpalauzov:
make by_value non-static, so that libcyrus_sieve can be used to interpret bytecode in multi-threaded environment.

I use `sieve_eval_bc()` in a multi-threaded application.

PR #5318, opened 2025-03-03
xsyslog_fn(): make `struct buf buf` non-static

dilyanpalauzov:
so that `message_parse_mapped()` can be called concurrently and each thread can log via `charset_decode_sha1()` “ignoring invalid base64 characters”.

Besides, there is no point to have the variable `buf` static here.

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -2154,12 +2154,11 @@ EXPORTED void tcp_disable_nagle(int fd)
 EXPORTED void xsyslog_fn(int priority, const char *description,
                          const char *func, const char *extra_fmt, ...)
 {
-    static struct buf buf = BUF_INITIALIZER;
+    struct buf buf = BUF_INITIALIZER;
     int saved_errno = errno;
     int want_diag = (LOG_PRI(priority) != LOG_NOTICE
                      && LOG_PRI(priority) != LOG_INFO);
 
-    buf_reset(&buf);
     buf_appendcstr(&buf, description);
     buf_appendmap(&buf, ": ", 2);
     if (extra_fmt && *extra_fmt) {


elliefm:
I think the point of that buf being static was so that buf.s doesn't need to be malloced and freed for every single call, just the first few until it's grown long enough for any further use (notice that `buf_reset()` clears `buf->len`, but leaves `buf->alloc` and `buf->s` intact).  But then, that intent is defeated by the `buf_free()` call near the end, doh!  I think the `buf_free()` being there is a bug.

Two possible paths here:

* Keep it static, remove the `buf_free()`.  Fixes the bug, reducing the malloc/free overhead as originally intended.  Might need to add a suppression if valgrind complains about it not being freed at exit.  Not re-entrant, but since Cyrus is mostly single-threaded, that doesn't matter much.  `mupdate` is the only part of Cyrus that's multi-threaded, so we should probably make sure it doesn't use xsyslog.
* As you've done.  Makes it re-entrant, but most of Cyrus won't benefit from this. Has malloc/free overhead for every call, though it did already anyway due to the bug.

What are you wanting to use it for that it needs to be thread-safe?  I don't think we provide any guarantees that any part of libcyrus is thread-safe, so this would be unusual.  Even if xsyslog/message_parse_mapped were thread safe, nothing else would be.  **I would generally recommend that for calling into libcyrus from a multi-threaded application, you should nominate a single thread for doing so, and only do so from that thread.**

If we want to achieve both goals (low malloc/free overhead, re-entrant), we could look at using pthread/C11/C23 thread-local storage within `xsyslog_fn()` for its `struct buf buf`.  If we did that naively, and the application uses many short-lived threads, buf.s would be leaked for each thread that finishes.  We could potentially do it more cleverly to avoid that; but it wouldn't be a problem for threads whose lifetime is the lifetime of the process, even if we didn't.  I don't think this is worthwhile though: no other part of libcyrus guarantees to be re-entrant, so you still wouldn't be able to do anything useful even if xsyslog was okay.

dilyanpalauzov:
I am using libcyrus_imap/libcyrus_sieve to apply Sieve scripts during SMTP delivery, and possibly reject the emails during the SMTP dialogue - https://mail.aegee.org/cgit/aegee-milter/tree/src/modules/mod_sieve/libcyrus_sieve.cpp#n238.  I do this in a multi-threaded way by calling `sieve_execute_bytecode()`, `sieve_script_load()`, `message_fetch_part()`, `message_parse_mapped()`,some `buf_…()` functions.  For over a decade I have hardly seen problems with concurrent execution, except the current case and https://github.com/cyrusimap/cyrus-imapd/pull/5020 .

If now `buf_free()` is removed from `xsyslog_fn()` once I decide to complete the documentation for the software I use, and announce it publicly, the instructions to adjusting the Cyrus IMAP source code will get even more complicated.  The current changes needed are shown at https://github.com/dilyanpalauzov/cyrus-imapd/commits/aegee-milter-3.8/ .  One of them is https://github.com/cyrusimap/cyrus-imapd/pull/5108, the others are to remove static on the return type in `user_sieve_path()` and `sievedir_get_active()`.

brong (approved):
I'm happy with this - it doesn't change the behaviour from the existing, and we're not seeing this as a hotspot anywhere.  We're calling syslog after all, and that's using tons more resources than a couple of mallocs!   Fewer static buffers is generally a good thing.

rjbs:

PR #3567, opened 2021-07-21
Add clang-format configuration to unify code style

rsto:
The Cyrus codebase currently uses a variety of code styles. This is a proposal for a single code style across all of the C and C++ source files.

It it currently a draft, and I suggest we all experiment with it.

The list of style options is documented at https://clang.llvm.org/docs/ClangFormatStyleOptions.html

Unfortunately my Debian development box only includes an ancient clang-format version (v7), which does not support a couple of options that I deem relevant. If we decide on a more recent version, I will be happy to update my local clang tools.

In the top-level cyrus-imapd directory, run ` clang-format -style=file <your source file here>` to check it out

PR #5320, opened 2025-03-05
remove hard tabs (again), this time with automation to prevent new ones

rjbs:
We remove all the tabs from our source 10yr ago, but didn't do anything to prevent them from growing back, so they did.

This branch removes them again, and also adds a GitHub action that will cause a failed check when tabs appear in "interesting" files.  I expect we'll do more about enforcing things like this in the future, but for now, this eliminates one thing we don't like.


dilyanpalauzov:
The file `.git-blame-ignore-revs` misses the commit “remove hard tabs (again), this time with automation to prevent new ones”.

While enforcing a no-tabs policy is good, it removes just one type of inconsistency.  Another type of inconsistency is the amount of leading spaces.  Enforcing the correct amount of leading spaces, as approached at https://github.com/cyrusimap/cyrus-imapd/pull/3567 and https://github.com/cyrusimap/cyrus-imapd/pull/5240, is a step in the right direction.

rjbs:
> The file `.git-blame-ignore-revs` misses the commit “remove hard tabs (again), this time with automation to prevent new ones”.

That's because I expect this branch to be rebased before merge, so I'll add the new SHA to the dotfile in the master branch subsequent to that.

> While enforcing a no-tabs policy is good, it removes just one type of inconsistency. Another type of inconsistency is the amount of leading spaces. Enforcing the correct amount of leading spaces, as approached at #3567 and #5240, is a step in the right direction.

I agree, and I'm hoping we'll enforce style issues more generally in the next few months, but this was a step I could bang out between meetings. 😉

docsrc/imap/reference/images/architecture.jpg
elliefm, line 1: what happened here?

elliefm (commented):
A bunch of jpgs under the docsrc tree got corrupted by this??

Otherwise, I guess the cleanup aspect is replacing hard tabs with align-to-next-8, which mostly does the right thing.  But curiously, there's places where it hoses the indentation.  Looks like someone was using hard tabs with tab width set to 4 at some point?  Which is very weird, because back when our code style did use hard tabs, we only used hard tabs for 8 spaces, and soft tabs otherwise.   And now that those 4-space hard tabs have been replaced with 8 real spaces, there's mess.

I dunno if you want to manually fix that up as part of this, but it's pretty easy to see where it went wrong in the diffs, because in those places the indentation visibly changes, instead of just replacing one kind of whitespace with another

rjbs:
I should make this a draft, only because I don't actually think it's ready yet!  I just pushed it up out of a constant fear of losing my local work.

The way the tab-fixer worked when I got here was very aggressive, touching all files.  I made it less aggressive later, then put that commit first… but things were already broken in chronological order, if not commit-topological order.

I'll redo the re-tabbing work.  As for the "indent got screwed up", I'll give that a real look before requesting review!

docsrc/imap/reference/images/architecture.jpg
rjbs: "All better."

The short version is that an early version of the tool operated on every file.  I updated the tool to be better, but had already caused the damage.  The branch now has the tool fixes first, then their application.

wolfsage (approved):

PR #5323, opened 2025-03-07
Asan full runs

wolfsage:
A WIP


wolfsage:
Will need to add some skips to specific tests for now until I can figure out compilation / linking (example cyruslibs, timsieved.so... etc)

wolfsage:
Closing in favor of https://github.com/cyrusimap/cyrus-imapd/pull/5370 which includes a bit of this work

PR #5324, opened 2025-03-09
Cassandane: try to speed up periodic_event_slow test

rjbs:
Before this patch, the test does:

1.  start background work
2.  wait 5m5s
3.  check that it's all good

This patch changes it to:

1.  start background work
2.  every 0.25s, check whether the work is done
3.  give up if not done by 5.5s
4.  check that it's all good

The performance seems moderately better, but not much.  I'm not sure it's worth merging yet, but it's not much more complex.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Master.pm
+++ b/cassandane/Cassandane/Cyrus/Master.pm
@@ -881,12 +881,23 @@ sub test_periodic_event_slow
 
     xlog $self, "periodic events run immediately";
 
-    xlog $self, "waiting 5 mins for events to fire, plus some slop";
-    sleep(5*60 + 5);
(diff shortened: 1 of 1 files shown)


elliefm:
The test is already marked `..._slow`, so it won't run unless you ask for it by name or invoke Cassandane with `--slow`.  When it runs as part of a big run of other tests (e.g. CI), it'll be mostly sleeping for those 5 minutes, so it won't be in contention with the other tests running in parallel.  I believe it's only a nuisance when you're running it alone, or in a small batch that would otherwise be finished in less than five minutes.  Have you found otherwise?

rjbs:
I can remove `--slow` from the default tests run by `cyd test`, but I had looked at this change because this test was often the last thing to finish by a long ways.  So, if it could be faster, it'd mean that a slow run was faster.

PR #5326, opened 2025-03-16
configure/make: Add --enable-asan and --enable-ubsan

wolfsage:
If enabled, by default these will link the sanitizer libraries statically. You can use --disable-static-asan and --disable-static-ubsan if you want them linked dynamically.

Generally, you want static linking, and the other tooling (that will make use of these configure arguments) will expect the libraries to be linked statically.

`cyr_info version` will also now use this info and display whether cyrus was built with one, both, or none of these sanitizers.

`make check` will automatically apply LSAN suppressions if built with asan. You can avoid this by doing:

    make SUPPRESS=0 check

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -79,12 +79,14 @@ if STATIC_ASAN
         AM_CXXFLAGS += -DWITH_STATIC_ASAN
 
 if IS_GCC_COMPILER
-            AM_CFLAGS += -static-libasan
-            AM_CXXFLAGS += -static-libasan
+            AM_CFLAGS += -static-libasan -DWITH_GCC
+            AM_CXXFLAGS += -static-libasan -DWITH_GCC
             AM_LDFLAGS += -static-libasan
 endif # IS_GCC_COMPILER
 else # ! STATIC_ASAN
 if IS_GCC_COMPILER
+            AM_CFLAGS += -DWITH_GCC
+            AM_CXXFLAGS += -DWITH_GCC
 else # ! IS_GCC_COMPILER
             AM_CFLAGS += -shared-libsan
             AM_CXXFLAGS += -shared-libsan
@@ -103,12 +105,14 @@ if STATIC_UBSAN
         AM_CXXFLAGS += -DWITH_STATIC_UBSAN
 
 if IS_GCC_COMPILER
-            AM_CFLAGS += -static-libubsan
-            AM_CXXFLAGS += -static-libubsan
+            AM_CFLAGS += -static-libubsan -DWITH_GCC
+            AM_CXXFLAGS += -static-libubsan -DWITH_GCC
             AM_LDFLAGS += -static-libubsan
 endif # IS_GCC_COMPILER
 else # ! STATIC_UBSAN
 if IS_GCC_COMPILER
+            AM_CFLAGS += -DWITH_GCC
+            AM_CXXFLAGS += -DWITH_GCC
 else # ! IS_GCC_COMPILER
             AM_CFLAGS += -shared-libsan
             AM_CXXFLAGS += -shared-libsan
--- a/Makefile.am
+++ b/Makefile.am
@@ -63,10 +63,59 @@ AM_CPPFLAGS = \
     $(ICU_CFLAGS)
 
 if HAVE_LDAP
-AM_CPPFLAGS += $(LDAP_CPPFLAGS)
+    AM_CPPFLAGS += $(LDAP_CPPFLAGS)
 endif # HAVE_LDAP
 
 AM_LDFLAGS = $(COV_LDFLAGS) $(ICU_LIBS)
+
+if ASAN
+    LSAN_SUPPRESS="env LSAN_OPTIONS=suppressions=leaksanitizer.suppress"
+    AM_CFLAGS += -fsanitize=address -DWITH_ASAN
+    AM_CXXFLAGS += -fsanitize=address -DWITH_ASAN
+    AM_LDFLAGS += -fsanitize=address
+
+if STATIC_ASAN
+        AM_CFLAGS += -DWITH_STATIC_ASAN
+        AM_CXXFLAGS += -DWITH_STATIC_ASAN
+
+if IS_GCC_COMPILER
+            AM_CFLAGS += -static-libasan
+            AM_CXXFLAGS += -static-libasan
+            AM_LDFLAGS += -static-libasan
+endif # IS_GCC_COMPILER
+else # ! STATIC_ASAN
+if IS_GCC_COMPILER
+else # ! IS_GCC_COMPILER
+            AM_CFLAGS += -shared-libsan
+            AM_CXXFLAGS += -shared-libsan
+            AM_LDFLAGS += -shared-libsan
+endif # ! IS_GCC_COMPILER
+endif # STATIC_ASAN
+endif # ASAN
+
+if UBSAN
+    AM_CFLAGS += -fsanitize=undefined -DWITH_UBSAN
+    AM_CXXFLAGS += -fsanitize=undefined -DWITH_UBSAN
+    AM_LDFLAGS += -fsanitize=undefined
+
+if STATIC_UBSAN
+        AM_CFLAGS += -DWITH_STATIC_UBSAN
+        AM_CXXFLAGS += -DWITH_STATIC_UBSAN
+
+if IS_GCC_COMPILER
+            AM_CFLAGS += -static-libubsan
+            AM_CXXFLAGS += -static-libubsan
+            AM_LDFLAGS += -static-libubsan
+endif # IS_GCC_COMPILER
+else # ! STATIC_UBSAN
+if IS_GCC_COMPILER
+else # ! IS_GCC_COMPILER
+            AM_CFLAGS += -shared-libsan
+            AM_CXXFLAGS += -shared-libsan
+            AM_LDFLAGS += -shared-libsan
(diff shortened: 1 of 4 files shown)


wolfsage:
I've added another commit which will also identify if gcc was used when compiling (but only when --enable-asan or --enable-ubsan is used).

Cass needs this for specific tests (MR to come...) because there are various tests that try to load shared libraries, and to use asan against shared libraries compiled with it, you must LD_PRELOAD the correct sanitizer lib...

Makefile.am
```diff
@@ -63,11 +63,64 @@ AM_CPPFLAGS = \
     $(ICU_CFLAGS)
 
 if HAVE_LDAP
-AM_CPPFLAGS += $(LDAP_CPPFLAGS)
+    AM_CPPFLAGS += $(LDAP_CPPFLAGS)
```
dilyanpalauzov, line 66: There is no consensus on indenting if-blocks and many other if-blocks start at the first column.  So I am against introducing indentation for a line, without changing it anyhow.

dilyanpalauzov:
Does cunit work with out-of-tree/VPATH builds?  For `make valgrind` I had to apply this https://github.com/cyrusimap/cyrus-imapd/commit/7eebf2d0fefd6059a6c77022ccf58254988ff148 to get with VPATH builds.

The master.c file has some memory leaks on exiting - missing `free(pidfile_lock);`  before `exit()` and `fatal()`.  Rather than adding these to suppressions lists in valgrind and asan, it would be simpler to fix the leaks.

I am against this change: there are countless ways to invoke the compilers and linkers and utilizing CFLAGS/CPPFLAGS before invoking `./configure` or during `make` is the way to adjust the build.  I stated this also at https://github.com/cyrusimap/cyrus-imapd/pull/5232#issuecomment-2608176041.  The person who invokes `./configure` is supposed to know exactly how to call the compiler and these options should not be in configure.ac/Makefile.am .

I personally use

> export   CFLAGS="-g -O1 -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=undefined -fno-omit-frame-pointer -fno-common -fsanitize-recover=address"
> export CXXFLAGS="-g -O1 -fsanitize=address -fsanitize=pointer-compare -fsanitize=pointer-subtract -fsanitize=undefined -fno-omit-frame-pointer -fno-common -fsanitize-recover=address"

because these proved to be useful in the past (I do not remember why), when I want to run Cyrus IMAP (or only its libraries) under the Address/Leak/Undefined sanitizer.

configure.ac
```diff
@@ -2589,6 +2630,26 @@ then
         LIBS="$LIBS -lstdc++";
 fi
 
+if test "X$asan" = "Xtrue";
```
elliefm, line 2633: Ditto these guys re `AS_IF()`

Makefile.am
```diff
@@ -63,11 +63,64 @@ AM_CPPFLAGS = \
...
+            AM_CFLAGS += -shared-libsan
+            AM_CXXFLAGS += -shared-libsan
+            AM_LDFLAGS += -shared-libsan
+endif # ! IS_GCC_COMPILER
+endif # STATIC_ASAN
+endif # ASAN
+
+if UBSAN
+    AM_CFLAGS += -fsanitize=undefined -DWITH_UBSAN
+    AM_CXXFLAGS += -fsanitize=undefined -DWITH_UBSAN
```
elliefm, line 100: I feel like we could maybe have done all this decision-making in configure.ac, setting up SAN_CFLAGS, SAN_CXXFLAGS, SAN_LDFLAGS with the appropriate combinations of flags (or empty if both sanitisers are disabled), and then in Makefile.am we would just need to interpolate these new variables into their respective AM_foo once, unconditionally.  We do something like this (albeit simpler) for GCOV_CFLAGS et al, so have a look at that for an example.

Off the top of my head, I'm not sure if it would be more readable/maintainable, but I notice we do already have some (all?) of these conditions in configure.ac already, e.g. for setting asan_desc etc.  If we're gonna do it all in there anyway, might as well not redo it later.

Makefile.am
```diff
@@ -708,10 +761,11 @@ check-local:
...
 		test -z "$$VG" || vg="$(VALGRIND)" ; \
+		test -z "$$SUPPRESS" && suppress=$(LSAN_SUPPRESS) ; \
 		f="-v" ; \
 		test "x$$CUFORMAT" = xjunit && f="-x" ; \
 		cd cunit ; \
-		$$vg ./unit $$f ; \
+		$$vg $$suppress ./unit $$f ; \
 		retval=$$? ; \
 		if [ "x$$CUFORMAT" = xjunit ] ; then \
 			$(RM) -rf reports ; mkdir reports ; ./cunit-to-junit.pl ; \
```
elliefm, line 771: As an aside, I really hate this rule, and keep vaguely meaning to replace the multi-line embedded shell script with a separate script that we just call with arguments.

elliefm (changes requested):
Looks good. I've got a bunch of suggestions and such, but only one comment that I think definitely needs fixing.

wolfsage:
> Does cunit work with out-of-tree/VPATH builds? For `make valgrind` I had to apply this [7eebf2d](https://github.com/cyrusimap/cyrus-imapd/commit/7eebf2d0fefd6059a6c77022ccf58254988ff148) to get with VPATH builds.

Oh interesting. I'm happy to fix this if it needs it.

> The master.c file has some memory leaks on exiting - missing `free(pidfile_lock);` before `exit()` and `fatal()`. Rather than adding these to suppressions lists in valgrind and asan, it would be simpler to fix the leaks.

I'm not sure how that's relevant here, could you explain?

> I am against this change: there are countless ways to invoke the compilers and linkers and utilizing CFLAGS/CPPFLAGS before invoking `./configure` or during `make` is the way to adjust the build. I stated this also at [#5232 (comment)](https://github.com/cyrusimap/cyrus-imapd/pull/5232#issuecomment-2608176041). The person who invokes `./configure` is supposed to know exactly how to call the compiler and these options should not be in configure.ac/Makefile.am .

I appreciate your point but you do not have to use this feature, and it won't affect builds that do not use it. In order for cassandane to work with specific tests (specifically: those that test loading of shared libraries compiled with asan) it must know how to LD_PRELOAD the correct libraries. gcc and clang use different ones, so cassandane must know which compiler was used, and that asan was used. The other option is to skip those tests entirely, and I'd rather not do that.

dilyanpalauzov:
> > The master.c file has some memory leaks on exiting - missing `free(pidfile_lock);` before `exit()` and `fatal()`. Rather than adding these to suppressions lists in valgrind and asan, it would be simpler to fix the leaks.
> 
> I'm not sure how that's relevant here, could you explain?

I mean if everything goes good with the tests (so 100% of possible cases are covered one day), these memory leak will be added soon or late in a suppression file for valgrind and asan.

> I appreciate your point but you do not have to use this feature, and it won't affect builds that do not use it. In order for cassandane to work with specific tests (specifically: those that test loading of shared libraries compiled with asan) it must know how to LD_PRELOAD the correct libraries. gcc and clang use different ones, so cassandane must know which compiler was used, and that asan was used. The other option is to skip those tests entirely, and I'd rather not do that.

Cassandane should be able to set CFLAGS and CXXFLAGS as environment variables before calling ./configure or make, and this would be equivalent to pass asan-parameters to the compiler/linker, as if a special ./configure option is used.  So what Cassandane needs, can be achieved without modifying configure.ac and Makefile.am. N.B.: I do not know exactly how Cassandane works, but it should be possible to set in build system environment variables before calling ./configure and before calling Cassandane.

Besides, I agree with the suggestion to set the flags in configure and not in Makefile.am , which also implies removing WARN_CFLAGS from Makefile.am and moving it in configure.ac . (But I still prefer if the extra flags like Werror and fsanitize=address are mentioned neither in configure.ac nor in Makefile).

wolfsage:
> Cassandane should be able to set CFLAGS and CXXFLAGS as environment variables before calling ./configure or make, and this would be equivalent to pass asan-parameters to the compiler/linker, as if a special ./configure option is used. So what Cassandane needs, can be achieved without modifying configure.ac and Makefile.am. N.B.: I do not know exactly how Cassandane works, but it should be possible to set in build system environment variables before calling ./configure and before calling Cassandane.

Setting flags isn't enough. With shared objects, LD_PRELOAD must be used, so cassandane needs to know that it must use it.

Again, this feature is not something you have to use, so it should not affect you.

dilyanpalauzov:
> With shared objects, LD_PRELOAD must be used

What must be put in LD_PRELOAD with shared objects?  Is it needed only for gcc or only for clang?  Also why the value of LD_PRELOAD must be there, in case the answer is not evident from the value of LD_PRELOAD? 

> … must know how to LD_PRELOAD the correct libraries

Which libraries exist compiled with and without sanitizer?

I suggest to consolidate the options `--enable-asan` and `--disable-static-asan` to a single option:
```
--disable-asan ⇔ --enable-asan=no
--enable-asan=shared
--enable-asan=static ⇔ --enable-asan=yes ⇔ --enable-asan
```

(It could be also `--with-asan=` instead of `--enable-asan=`).

wolfsage:
Closing in favor of https://github.com/cyrusimap/cyrus-imapd/pull/5370

Rather than fight with automake etc I'm doing this with cyd / cass itself, and skipping (for now) tests that require linking shared objects (there's only 3 of those, I don't think this is a problem in general - the surface area we cover instead is far greater)

PR #5329, opened 2025-03-18
docsrc/conf.py: apply black formatter

rjbs:
This just reformats conf.py with black, the (a) Python code formatter.


wolfsage (approved):
Sure. Still looks snake like

PR #5330, opened 2025-03-18
updates to release and install docs

rjbs:
This consolidates (or deletes) a bunch of documentation about installation and release notes.

I think there's lots more left to do, but this is a start.


docsrc/index.rst
```diff
@@ -4,44 +4,29 @@
 What is Cyrus IMAP?
 ===================
 
-Cyrus IMAP is an email, contacts and calendar server. Cyrus is free and open source.
+Cyrus IMAP is an efficient, reliable, and scalable server for email, calendar,
+and and contacts server. Cyrus is free and open source.
```
elliefm, line 8: > server for ... and **and** contacts **server**

couple of duplicated words!

docsrc/index.rst
```diff
@@ -4,44 +4,29 @@
...
-* Long list of supported standards
-
-Read more in our full :ref:`list of features <imap-features>`.
-
-Cyrus has been under active development since the year 1993 when the project was launched at Carnegie Mellon University. It's used in production
-systems around the world, at universities and in private enterprise.
+Cyrus IMAP supports many authentication options.  It uses Cyrus SASL, an
+implementation of the `Simple Authentication and Security Layer (SASL)
+<https://en.wikipedia.org/wiki/Simple_Authentication_and_Security_Layer>`_
+system.  You can read more at the `Cyrus SASL web page <http://www.cyrusimap.org/sasl>`_.
```
elliefm, line 23: You might be able to link to this using an RST thingy rather than a direct hyperlink, since the two repositories get built into a single website, and both of them publish a TOC thingy so they should know what each other's main headings are.  Especially since this file is "docsrc/index.rst" (and not "docsrc/imap/index.rst"), and is therefore supposed to be the front page for the entire common website.  There might be a magic word defined in docsrc/conf.py that you can invoke.

In practice, I would look for other references to the SASL site and see how they're linked, and do whatever they do.   Maybe you already did that and this is how it's done.

docsrc/index.rst
```diff
@@ -50,19 +35,3 @@ Need help? We have :ref:`active mailing lists <feedback-mailing-lists>`.
     operations
     developers
     support
-
---------
-
-===================
-What is Cyrus SASL?
```
elliefm, line 57: I think we should keep this section.  It's not so obvious from this repository, but this is a shared front page for the whole common website, not a Cyrus-IMAP specific page.

docsrc/imap/download/release-notes/index.rst
```diff
@@ -128,23 +118,24 @@ Development snapshots
...
     :maxdepth: 1
 
-    1/1.x.x
+    2.5/index
+    2.4-dav/index
+    2.4/index
+    2.3/index
+    2.2/2.2.x
+    2.1/2.1.x
+    2.0/2.0.x
```
elliefm: Do any of these pre-2.5 links go anywhere useful?  2.5 was the first version that we used RST for the docs, 2.4 and earlier just had a directory of html files that used to get published statically on a server that doesn't exist anymore, and I don't know/remember what happens for them anymore.

docsrc/imap/download/release-notes/index.rst
```diff
@@ -28,112 +22,86 @@ Latest development snapshot is |imap_development_release_notes|. Documentation a
...
+
+Supported Verions
+-----------------
+
+The latest stable version is |imap_stable_release_notes|.  Its documentation
+can be found at :cyrus-stable:`/`.
 
 Series 3.12
------------
+~~~~~~~~~~~
```
elliefm, line 44: This section of changes is sort of awkward.  The master branch is currently in a state of lying that 3.12 is the current stable version, because its own version is already 3.13.  Whereas the main website that's built from the 3.10 branch doesn't even know 3.12 exists yet.  Once 3.12 is released for reals, the lies will become true.  But in the meantime, landing and backporting this set of changes is a little awkward.

If we don't intend to backport these changes (or only as far as 3.12, so that the main website only sees the updates once 3.12 is released), then it's less awkward.

elliefm (commented):
Looking good.  One big pair of typos, a few notes and thoughts.

docsrc/index.rst
rjbs: Hrm.  I will do this, but I don't think it's actually a good idea for us to do this.  I will try to suck it up, it's probably fine.  This works:

```
:ref:`Cyrus SASL <cyrussasl:sasl-index>`
```

…not because they are built into a single website, but because `cyrussasl` is configured as an "intersphinx link", which means that when we build the Cyrus IMAP website, it will fetch, *via HTTP*, the TOC of the published SASL site.  So, there's sort of an ordering / bootstrapping issue.

Whatever.  We can fix it someday or not.  For now, we use it, so I've done it to be consistent.

docsrc/imap/download/release-notes/index.rst
rjbs: Yeah, I'd say that it was broken/awkward/wrong when I got here, and I've just reorganized things.  I think we should fix this, but I don't think this branch is making things worse right now.  Also, **I don't think we need to backport these changes**.

rjbs:
@elliefm LMK if you want to talk more about the change to the front page.  Mostly, I think this change is an improvement and puts SASL further forward, even while reducing the amount of text it gets.

elliefm (approved):

PR #5335, opened 2025-03-19
some small fixes

brong:
This is some of the small fixes that I just did on-the-fly as we were reviewing tickets today


ksmurchison (approved):
LGTM

PR #5074, opened 2024-10-09
http_caldav.c: make property scheduling-enabled live

dilyanpalauzov:
and enforce that the submitted value is valid.

This allows setting for the CY:scheduling-enabled property only the values `F`, `f`, `NO`, `No`, `no`, and `nO`, as all other values are irrelevant in _scheduling_enabled().  By applying also https://github.com/cyrusimap/cyrus-imapd/pull/5060 it is ensured that whoever sets the property, it will be stored for the user "", as only this user is checked by _scheduling_enabled().

As can be seen in the code, this property is stored by `proppatch_todb_nomask()`.

PR #5338, opened 2025-03-19
http_caldav.c: make property scheduling-enabled live

dilyanpalauzov:
and enforce that the submitted value is valid.

This allows setting for the CY:scheduling-enabled property only the values `F`, `f`, `NO`, `No`, `no`, and `nO`, as all other values are irrelevant in _scheduling_enabled(). By applying also https://github.com/cyrusimap/cyrus-imapd/pull/5071 it is ensured that whoever sets the property, it will be stored for the user "", as only this user is checked by _scheduling_enabled().

As can be seen in the code, this property is stored by `proppatch_todb_nomask()`.

Same as https://github.com/cyrusimap/cyrus-imapd/pull/5074 , but uses `(char*)"F"` instead of `"F"`, because of https://github.com/cyrusimap/cyrus-imapd/pull/5096.

### DIFF
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -195,6 +195,9 @@ static int propfind_scheduser(const xmlChar *name, xmlNsPtr ns,
                             struct propfind_ctx *fctx,
                             xmlNodePtr prop, xmlNodePtr resp,
                             struct propstat propstat[], void *rock);
+static int proppatch_scheduling_enabled(xmlNodePtr, unsigned,
+                                        struct proppatch_ctx*,
+                                        struct propstat[], void*);
 static int propfind_caldata(const xmlChar *name, xmlNsPtr ns,
                             struct propfind_ctx *fctx,
                             xmlNodePtr prop, xmlNodePtr resp,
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
I think the problem here is that `F` is stored by calling `proppatch_todb_nomask(prop, set, pctx, propstat, (char*) "F");` ⇔ `annotate_state_write(astate, buf_cstring(&pctx->buf), httpd_userid, &value);`.  At the same time `_scheduling_enabled()` uses `annotatemore_lookupmask_mbox(mailbox, DAV_ANNOT_NS "<" XML_NS_CYRUS ">scheduling-enabled", "", &buf);`.  So in the first case the user is httpd_userid, in the second the user is `""`.  This means that the property is not read for the user `httpd_userid`.

The details are discussed for `proppatch_todb_nomask()` at https://github.com/cyrusimap/cyrus-imapd/pull/5071.

rjbs:
As you can see, this branch seems to cause the test suite to fail.  I haven't investigated, but won't be merging it while it's failing.

dilyanpalauzov:
>  this branch seems to cause the test suite to fail

I have spelled the reason above.

Issue #4665, opened 2023-09-24
IMAP PREVIEW returns incorrectly `NIL`


dilyanpalauzov:
Per [RFC 8970](https://www.rfc-editor.org/rfc/rfc8970.html#section-3.2-6)

> A server MUST NOT return NIL to a FETCH PREVIEW request made without the LAZY modifier.

In Cyrus IMAP 3.6, which is supposed to implement RFC 8970 IMAP PREVIEW, on FETCH can be returned `NIL`.  Looking at 

https://github.com/cyrusimap/cyrus-imapd/blob/97435592f0390daf70159162d41e7134c4229924/imap/index.c#L4472-L4489

if `IMAPOPT_JMAP_PREVIEW_ANNOT` is not set, and by default it is not set, then no-lazy FETCH returns `NIL`.

ksmurchison:
Fixed by #5339

PR #5339, opened 2025-03-19
imapd.c: only advertise/accept PREVIEW when properly configured

ksmurchison:
Fixes issue #4665

### DIFF
--- a/cassandane/Cassandane/Cyrus/Fetch.pm
+++ b/cassandane/Cassandane/Cyrus/Fetch.pm
@@ -55,7 +55,13 @@ $Data::Dumper::Sortkeys = 1;
 sub new
 {
     my $class = shift;
-    return $class->SUPER::new({ adminstore => 1 }, @_);
+
+    my $config = Cassandane::Config->default()->clone();
+    $config->set(
+        jmap_preview_annot => '/shared/vendor/messagingengine.com/preview',
+    );
+
+    return $class->SUPER::new({ config => $config, adminstore => 1 }, @_);
 }
 
 sub set_up
--- a/changes/next/imap-preview-issue-4665
+++ b/changes/next/imap-preview-issue-4665
@@ -1,0 +1,23 @@
+
+Description:
+
+Use the 'jmap_preview_annot' option to enable/disable
+the IMAP PREVIEW capability and the FETCH PREVIEW attribute.
+
+Config changes:
+
+The IMAP PREVIEW extension is now enabled ONLY if the 'jmap_preview_annot'
+option is non-NULL and begins with the '/shared/' prefix.
+
+
+Upgrade instructions:
+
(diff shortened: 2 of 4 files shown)


wolfsage:
lgtm. Do any docs need updating anywhere to specify that "this config option must be true if you want PREVIEW support..." ?

cassandane/Cassandane/Cyrus/Fetch.pm
```diff
@@ -55,7 +55,12 @@ $Data::Dumper::Sortkeys = 1;
 sub new
 {
     my $class = shift;
-    return $class->SUPER::new({ adminstore => 1 }, @_);
+
+    my $config = Cassandane::Config->default()->clone();
+    $config->set(jmap_preview_annot => '/shared/vendor/messagingengine.com/preview',
+    );
```
wolfsage, line 61: nit: Weird closing parent placement. Would put this all on one line or do this if it needs to fit in 80:
```
    $config->set(
        jmap_preview_annot => '/shared/vendor/messagingengine.com/preview',
    );

```

ksmurchison:
> lgtm. Do any docs need updating anywhere to specify that "this config option must be true if you want PREVIEW support..." ?

Yeah.  Good point.  I will add

cassandane/Cassandane/Cyrus/Fetch.pm
ksmurchison: Hmm.  Quick cut-n-paste from another test and I didn't look at the formatting.  I will fix

rsto (approved):

dilyanpalauzov:
rfc-support.rst under RFC 8970 could include a Note: (as RFC 2086 does), stating when PREVIEW is enabled.

I do not think it is sufficient to have any value in IMAPOPT_JMAP_PREVIEW_ANNOT, the value must start with "/shared/" (`if (annot && !strncmp(annot, "/shared/", 8)) {`).  This is not documented anywhere.  As the PREVIEW annotation is inserted by jmap_mail.c I also suspect that `httpmodules: jmap` is necessary and overall for me it is not clear how to get PREVIEW running.

dilyanpalauzov:
If the `FETCH_PREVIEW` bit can be set only when `IMAPOPT_JMAP_PREVIEW_ANNOT` is set, then in `index_fetchreply()`:
```c
        const char *annot = config_getstring(IMAPOPT_JMAP_PREVIEW_ANNOT);
        if (annot && !strncmp(annot, "/shared/", 8)) {
```
the first part of the condition — `annot &&` — is always true.

cassandane/Cassandane/Cyrus/Fetch.pm
ksmurchison: Fixed

wolfsage (approved):
lgtm

Issue #2944, opened 2019-12-09
lib/imapoptions: convert doc comments to rst


elliefm:
The documentation C-comments in lib/imapoptions are in man format, because this file was historically converted directly into the imapd.conf.5 man page.

Nowadays, this file is converted into imapd.conf.rst by `tools/config2rst`, and then sphinx-build does the work of converting that into both the man page and the html version of the man page.  But, this means `tools/config2rst` currently has to parse all the old man formatting and convert it to rst, and the person writing documentation for an imapd.conf option has to write what they really want in man format and hope that it survives all the translations roughly intact.

We should simplify all this by converting the descriptive parts of lib/imapoptions to rst, so config2rst can just pass them through, and let sphinx be the only source of man formatting.

rjbs:
Wow.  Yes.

elliefm:
I actually have a new version of this idea written up in the FM tracker.  I looked into this particular approach years ago, and for various reasons rst doesn't embed into other things well, but I guess I forgot to come back and comment here.  And then next time it annoyed me enough to look into it again, I forgot this task existed and made a new one...

PR #5342, opened 2025-03-20
config2rst: do not double-indent topmost section text

### LINKED ISSUE #2889: generated man pages have bad indentation
**imapd.conf.5**

Notice how the NAME section is indented normally (8 spaces), but then the DESCRIPTION section has an extra half-level of indentation (12 spaces).  The 12-spaces indentation carries on all the way through to the SEE ALSO section, after which the AUTHOR and COPYRIGHT sections are back to the normal 8 spaces.

```
IMAPD.CONF(5)                     Cyrus IMAP                     IMAPD.CONF(5)

NAME
       imapd.conf - Cyrus IMAP documentation

       IMAP configuration file

DESCRIPTION
          /etc/imapd.conf is the configuration file for the Cyrus IMAP server.
          It defines local parameters for IMAP.

          Each line of the /etc/imapd.conf file has the form
                 option: value

          where option is the name of the configuration option being  set  and
          value is the value that the configuration option is being set to.

[...]
```

This file starts life as `lib/imapoptions`, and is converted to imapd.conf.rst by `tools/config2rst`, and then finally sphinx generates the 'man' and 'html' versions of it.

Back in the olden days (before our documentation was consistently rst) this file was converted directly to man format by `tools/config2man`, though I'm not sure if this problem existed back then too.

It may be a quirk of the current double-conversion.  It could also be a side-effect of the pretab/posttab change a few years ago (maybe the converter tools still assume mixed tabs/spaces and mishandle pure-spaces?).  I dunno, I haven't looked!  But these are the hunches I'd start with, if I sunk some time into chasing this down...

**cyradm.8**

This one has a bunch of 4-space indents in the COMMANDS section (other sections are fine).  It's converted from `perl/imap/IMAP/SHELL.pm` to rst by `tools/perl2rst`, and then sphinx builds the man and html documents.

**sieveshell.1**

This one is fine (for comparison), though it's also converted to rst by `tools/perl2rst` (from `perl/sieve/scripts/sieveshell.pl`).

Any fix to the man page generation must still also result in valid html generation!

rsto:
Fixes #2889

### DIFF
--- a/tools/config2rst
+++ b/tools/config2rst
@@ -153,9 +153,9 @@ sub print_man2rst {
     my $result = '';
 
     # If we're already in a section, and tabs is 0, set it to 1
-    if ($section == 1 && $tabs == 0) {
-        $tabs = 1;
-    }
+    #if ($section == 1 && $tabs == 0) {
+    #    $tabs = 1;
+    #}
 
     # We have a hanging indent from the last pass
     if ($inl > 0) {


rsto:
@elliefm this fixes the formatting for `imapd.conf.5`. I looked at `cyradm.8` but couldn't find anything that stuck out to be wrongly formatted. Do you see any issues remaining with the man page of `cyradm.8`?

elliefm (approved):
That seems to do what we want I think

elliefm:
I guess I had a lot of issues open about this stuff, heh (see also #2944)

PR #5343, opened 2025-03-21
require Python 3 for building the docs

rjbs:
This is largely to simplify our prereq detection.  Also, Python v2 is dead as heck.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2352,33 +2352,15 @@ AX_PROG_PERL_MODULES([Pod::POM::View::Restructured],
                        have_ppvr=no
                      ])
 
-AX_PYTHON_MODULE([git], [], [python2])
+AX_PYTHON_MODULE([git], [], [python3])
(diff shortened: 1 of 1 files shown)


elliefm (approved):

dilyanpalauzov:
This is the very same thing as https://github.com/cyrusimap/cyrus-imapd/pull/5137 (at the time it was created), but instead of `AX_PYTHON_MODULE([git], [], [python3])` the other change uses the more concise `AX_PYTHON_MODULE([git])`.  From my perspective, in the current situation, the efforts to create https://github.com/cyrusimap/cyrus-imapd/pull/5137 were in fact lost time.

rjbs:
@dilyanpalauzov Sorry, of course that sounds frustrating.  As you've probably seen, we're trying close out many old issues and PRs, because it's still hard to keep track of everything at the current counts.  As the numbers get closer to zero, duplicated effort should be less likely.

PR #5345, opened 2025-03-21
build-with-cyruslibs.sh: remove this, no longer used

rjbs:
This program was used by the GitHub Actions builds, but that's now covered by "cyd build" which is now self-contained.  I don't believe this has any other use, now, so it's just one more thing that could bitrot.

We can certainly keep this if it's still in use, but so far I don't think it is.  I'll be requesting review from a bunch of the usual suspects to make sure.


ksmurchison (approved):

rsto (approved):

elliefm (approved):

brong (commented):
Sure!  Kill it.  If we find we want it again it it's a simple "git revert" away anyway.

brong (approved):

PR #5346, opened 2025-03-21
imapoptions: deprecate unused "postspec" option

### LINKED ISSUE #2948: 'postspec' config item is undocumented, but also appears to be unused

rsto:
This option got introduced for version "2.3.17" but it seemingly never got implemented. The commit ref that introduced it is ced20b69d8064480d05e6f40da4093584b4c4ae9

Fixes #2948

### DIFF
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -2144,7 +2144,8 @@ If all partitions are over that limit, this feature is not used anymore.
 /* Username that is used as the 'From' address in rejection MDNs produced
    by sieve. */
 
-{ "postspec", NULL, STRING, "2.3.17" }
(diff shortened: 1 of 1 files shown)


elliefm (approved):
Can you make the two version numbers "UNRELEASED" please?  This doesn't need a re-review after that's done, so I'm clicking approve now.

Given that the option was unused, I think the only difference will be that if someone had `postspec: ...` configured uselessly in their imapd.conf, `cyr_info conf-lint` will start complaining about it after they upgrade.  With the version numbers bumped, `cyr_info -s <version> conf` will be able to tell them about it, so it shouldn't be a surprise to them.

Issue #4787, opened 2024-01-10
VANISHED not working with cyrus proxy


alecpl:
We're using Cyrus Murder 3.6.3 setup. Here's the error:
```
[10-Jan-2024 12:29:19.352378]: C: A0214 SELECT Test
[10-Jan-2024 12:29:19.399869]: S: * OK [CLOSED] Ok
[10-Jan-2024 12:29:19.400106]: S: * 1 EXISTS
[10-Jan-2024 12:29:19.400235]: S: * 1 RECENT
[10-Jan-2024 12:29:19.400332]: S: * FLAGS (\Answered \Flagged \Draft \Deleted \Seen)
[10-Jan-2024 12:29:19.400435]: S: * OK [PERMANENTFLAGS (\Answered \Flagged \Draft \Deleted \Seen \*)] Ok
[10-Jan-2024 12:29:19.400531]: S: * OK [UNSEEN 1] Ok
[10-Jan-2024 12:29:19.400628]: S: * OK [UIDVALIDITY 1577540428] Ok
[10-Jan-2024 12:29:19.400721]: S: * OK [UIDNEXT 2] Ok
[10-Jan-2024 12:29:19.400813]: S: * OK [HIGHESTMODSEQ 53] Ok
[10-Jan-2024 12:29:19.400906]: S: * OK [MAILBOXID (e7534bd5-198d-42a9-822d-e5adb478b933)] Ok
[10-Jan-2024 12:29:19.400998]: S: * OK [URLMECH INTERNAL] Ok
[10-Jan-2024 12:29:19.401090]: S: * OK [ANNOTATIONS 65536] Ok
[10-Jan-2024 12:29:19.401183]: S: A0214 OK [READ-WRITE] Completed
[10-Jan-2024 12:29:19.417594]: C: A0215 ENABLE QRESYNC
[10-Jan-2024 12:29:19.459301]: S: * ENABLED CONDSTORE QRESYNC
[10-Jan-2024 12:29:19.459683]: S: A0215 OK Completed
[10-Jan-2024 12:29:19.459931]: C: A0216 UID FETCH 1:* (FLAGS) (CHANGEDSINCE 52 VANISHED)
[10-Jan-2024 12:29:19.505736]: S: A0216 BAD Invalid UID Fetch modifier VANISHED
```
It does work when connecting to the backend directly.

ksmurchison:
Fixed by #5347

PR #5347, opened 2025-03-21
imapd.c: proxy ENABLE commands to backends

ksmurchison:
Fix for issue #4787


rsto (approved):

elliefm:
The shape of this patch expects IMAP4rev2 (RFC 9051) support, so we can't backport this to 3.8 as is.  The interesting change is the addition of the "ENABLE on all remote mailboxes", which would need to be rewritten for 3.8 without assuming IMAP4rev2 support.

Issue #2425, opened 2018-07-06
Enable PROPFIND and OPTIONS for `anyone` when it has lr rights


dilyanpalauzov:
https://www.cyrusimap.org/imap/download/installation/http/caldav.html does not say what privileges are needed to call HTTP OPTIONS.

For a calendar I set `lam user/x/#calendars/y anyone lr9`.

The GET https://\<server\>/dav/calendars/user/x/y works, but calling OPTIONS or PROPFIND returns 401 Unauthorized, so unauthenticated WebDAV Collection Syncrhonization is not possible.

dilyanpalauzov:
What about substituting all calls of `http_allow_noauth_get()` with `http_allow_noauth()` or just stripping `int (*need_auth)(struct transaction_t *)` from `imap/httpd.h:struct namespace_t` together with
```diff
diff --git a/imap/httpd.c b/imap/httpd.c
index a2e448713..f5b7c18d5 100644
--- a/imap/httpd.c
+++ b/imap/httpd.c
@@ -1760,11 +1760,6 @@ static int examine_request(struct transaction_t *txn)
                   txn->req_line.uri, txn->req_line.meth);
     buf_reset(&txn->buf);
 
-    /* Request authentication, if necessary */
-    if (!httpd_userid && namespace->need_auth(txn)) {
-        ret = HTTP_UNAUTHORIZED;
-    }
-
     if (ret) return client_need_auth(txn, r);
 
     /* Check if this is a Cross-Origin Resource Sharing request */
```
?

dilyanpalauzov:
It turns out, that [DAVdroid](https://www.davdroid.com/) insists on getting a username and password before trying.

Gnome [Evolution](https://wiki.gnome.org/Apps/Evolution) asks for password only after getting 401 Unauthorized, which does not arrive with the change above after requesting `PROPFIND <server>/dav/calendars` during the setup.

Nevertheless I think the change it correct.

ksmurchison:
If this is still an issue, please create a pull request

dilyanpalauzov:
This is a big change:  it requires removing from httpd.h:`struct namespace_t`:  `int (*need_auth)(txn_t *);  /* Function run prior to unauthorized requests */`, and the `http_allow_noauth()` and `http_allow_noauth_get()` implementations. In addition also from `examine_request()` removing:
```c
    /* Request authentication, if necessary */
    if (!httpd_userid && namespace->need_auth(txn)) {
[...]

PR #5349, opened 2025-03-23
imapd/httpd:http_allow_noauth_get: don't request authentication depending on the HTTP verb

dilyanpalauzov:
Instead always check the ACLs, so that anonymous users can subscribe a CalDAV resource.

This is https://github.com/cyrusimap/cyrus-imapd/issues/2425, 

Let’s what will happen.

### DIFF
--- a/imap/http_prometheus.c
+++ b/imap/http_prometheus.c
@@ -47,7 +47,6 @@
 #include "imap/http_err.h"
 #include "imap/prometheus.h"
 
-static int prom_need_auth(struct transaction_t *txn);
 static void prom_init(struct buf *);
 static int prom_auth(const char *);
 static void prom_reset(void);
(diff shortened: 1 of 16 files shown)


dilyanpalauzov:
This fails with Prometheus, but I do not use Prometheus.  Any ideas how to handle this?

imap/http_prometheus.c
```diff
@@ -138,6 +126,9 @@ static int prom_get(struct transaction_t *txn,
     if (strcmp(txn->req_uri->path, "/metrics"))
         return HTTP_NOT_FOUND;
 
+    if (!httpd_userisadmin || !httpd_userid || httpd_userisanonymous)
+        return HTTP_UNAUTHORIZED;
+
```
elliefm, line 131: This check is breaking the `prometheus_need_auth` imapd.conf option when set to "user" or "none", which would be handled by the lines below, except that we already returned here if the user wasn't an admin.

> prometheus_need_auth: admin
> Authentication level required to fetch Prometheus metrics.
> 
> Allowed values: none, user, admin

I think you can just delete these new lines.  When I fetch your branch and do this locally, the Prometheus tests start passing again.

PR #5350, opened 2025-03-23
docs: clean up (and strip down) developer docs

rjbs:
This branch…

1. removes a bunch of pages of minimal value; the Cyrus docs is not the place to explain GitHub
2. boils down a lot of text here and there
3. removes the docs on releasing ancient (2.5-ish) versions
4. removes reference to JMAP Auth, which is long abandoned
5. removes the intersphinx link, which doesn't work offline and provides minimal value here
6. rewrites "how to run the tests" to use Docker, cyd, and dar


docsrc/imap/developer/overview.rst
```diff
@@ -1,52 +1,136 @@
...
+and test.  The other command you might want to run is ``cyd makedocs``, which
+will build the Cyrus documentation website, putting the built files at
+``/srv/cyrus-imapd/docsrc/build/html``.
+
+dar
+---
+
+Developing with just Docker and ``cyd`` is possible, but can be a pain.  You
+won't have your ssh credentials, so you can't push changes back up.  You'll be
+logged in a root, and you won't have your editor configuration.  To make
```
elliefm, line 90: typo: logged in **as** root

docsrc/imap/developer/overview.rst
```diff
@@ -1,52 +1,136 @@
...
+you can use ``dar sh`` to get a shell inside the container.
+
+When you're all done, or when you've accidentally done something to ruin or
+compromise the container, you can run ``dar prune`` to stop the container and
+remove it.  Stopping and removing the container can take a little while, if
+you've built up a lot of test output.
+
+``dar`` works by mounting your git checkout into the container where the source
+is expected.  When you build, you'll be building into that directory, meaning a
+directory on your local machine.  This means you'll send up with `foo.o` files
```
elliefm, line 131: typo: This means you'll **end** up with

docsrc/imap/developer/overview.rst
```diff
@@ -1,52 +1,136 @@
...
+in your working tree, and they'll be owned by the root user.  ``dar clean``
```
elliefm, line 132: 🤔 I wonder if we can do something about this so that these files don't wind up owned by root

docsrc/imap/developer/releasing.rst
```diff
@@ -39,16 +36,17 @@ Order of operations
...
 
 Sometimes you're releasing several new versions all at once(ish), for example
-maybe there's been a security fix that affected 2.4, 2.5 and 3.0.
+maybe there's been a security fix that affected 3.12, 3.10 and 3.8.  (Heaven
+forbid.)
 
 Github's release page will put a "Latest Release" graphic on the release with
 the newest tag (by timestamp, I think).  This means that, if you're doing new
 releases for several different versions, you want to do the oldest one first,
 and only do the release for the current-stable branch last.
```
elliefm, line 45: This information is actually out of date and has been for a while.  When you're uploading a release, github now gives you checkboxes that you can select for something like "Latest release" (default: enabled) and "Pre-release" (default: not enabled).

Only **the latest release for the stable branch** should have the "Latest Release" box ticked, it should be unticked when publishing releases from older branches.  Betas, release candidates, etc must always have the "Pre-release" box ticked and never "Latest release".

Since github now lets you specify these flags rather than inferring them from publication date and stuff, it's no longer important to do batches of releases in any particular order.

I keep meaning to update it, but finding the words to explain it is hard, and the meanings of the checkboxes seem like they would be self-evident to anyone actually doing a release anyway.

docsrc/imap/developer/releasing.rst
```diff
@@ -97,17 +95,13 @@ Pre-release testing
...
-    from each Cyrus branch, especially for 2.5.  If you're doing releases
-    regularly, you've probably got a good gut feel for which failures are a
-    problem and which ones are just "that old thing".  If you don't do
-    releases regularly, try to pull in someone who does for guidance about
-    which failures are ignorable, and which should be a source of stress.
+    from each Cyrus branch.  If you're doing releases regularly, you've
+    probably got a good gut feel for which failures are a problem and which
+    ones are just "that old thing".  If you don't do releases regularly, try to
+    pull in someone who does for guidance about which failures are ignorable,
+    and which should be a source of stress.
```
elliefm, line 104: This is no longer true since about 3.2 or 3.4 or so, we expect all Cassandane tests should pass at release time.

docsrc/conf.py
```diff
@@ -44,7 +39,6 @@
     "sphinx.ext.ifconfig",
     "sphinx.ext.mathjax",
     "sphinx.ext.todo",
-    "sphinx.ext.intersphinx",
```
elliefm, line 47: How does this play with the SASL docs?  I think cyrus-sasl might still expect intersphinx to be available for linking back to the IMAP docs.  I don't know how intersphinx works (someone else set it up and I've always left it alone), maybe it's not a problem, but maybe it is.

elliefm (commented):
Few typos, a few comments

docsrc/imap/developer/overview.rst
rjbs: I've spent a bunch of time on thinking about this so far.  There are some ideas, but nothing super promising yet. One option that I don't love is to have dar run as $UID and add that user to sudoers dynamically.

docsrc/imap/developer/releasing.rst
rjbs: Okay!  How about I merge this and come back to update that on a second pass?  I made CYR-1722.

docsrc/imap/developer/releasing.rst
rjbs: **Good!**  I'll remove this text gladly.

docsrc/conf.py
rjbs: I'm afraid the simplest way to find out is to push this and find out.  If it causes a problem, I can put this line back without using any intersphinx links.  Reading the docs, I think we're good.  They say:

* Each Sphinx HTML build creates a file named objects.inv that contains a mapping from object names to URIs relative to the HTML set’s root.
* Projects using the Intersphinx extension can specify the location of such mapping files in the [intersphinx_mapping](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_mapping) config value. The mapping will then be used to resolve both [external](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#role-external) references, and also otherwise missing references to objects into links to the other documentation.

That suggests that *every* Sphinx build has the inventory file, and the extension is to let you *use* another Sphinx site's inventory.

docsrc/imap/developer/releasing.rst
elliefm: I was actually gonna suggest something like that myself but forgot!

docsrc/conf.py
elliefm: Yeah cool, that sounds fine then. :)

PR #5355, opened 2025-03-25
Caldav.export_utf8: verify that calendar export preserves UTF-8 chars

ksmurchison:


rsto (commented):
I'm going to leave commenting on the Perl and UTF-8 string handling to rjbs to wolfsage. Other than that looks good to me.

PR #5356, opened 2025-03-25
Buf mmap fix length

brong:
This reverts the changes from https://github.com/cyrusimap/cyrus-imapd/pull/4840 which turns out to leak mmaps!  You can't just munmap a different length than you passed to mmap.

Instead, we're going to find the actual bugs and fix them as we hit them.


rsto (approved):
LGTM

PR #5359, opened 2025-03-27
use renameat and fsync(dirfd) for ALL renames

brong:
I was originally planning to just use renameat in mappedfile, but on reflection I built an API wrapper for rename, and used that everywhere (except twom; which tries not to use the rest of cyrus libs for portability)

### DIFF
--- a/lib/twom.c
+++ b/lib/twom.c
@@ -2191,8 +2191,10 @@ static int replay_cb(void *rock,
 static int tm_rename(struct twom_db *db, struct tm_file *oldfile, const char *newname)
 {
     struct stat sbuf, sbuffile;
-    char *copy = strdup(db->fname);
-    const char *dir = dirname(copy);
+    char *copyd = strdup(db->fname);
+    char *copyb = strdup(db->fname);
+    const char *dir = dirname(copyd);
+    const char *file = dirname(copyb);
     int r = 0;
     int dirfd = -1;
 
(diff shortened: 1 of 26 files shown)


brong:
This is failing tests because they use relative paths for databases; which doesn't work!  We got away with it before, but it's bogus.  *sigh*.  I guess we need to add a path resolver to this rename shenanigans.

lib/util.h
```diff
@@ -220,6 +220,9 @@ extern char *create_tempdir(const char *path, const char *subname);
  * value of remove on error. */
 extern int removedir(const char *path);
 
+/* Call rename but fsync the directory before returning success */
+extern int cyrus_rename(const char *src, const char *dest);
```
dilyanpalauzov, line 224: `extern ` is not necessary.

brong: sure but it's present on ALL THE OTHER THINGS NEARBY; so I'm gonna be consistent with the surrounding code, because it will be less weird for anyone cleaning it up later.

brong:
All the extra commits added to this PR are to support opening a dirfd for the target directory, and caching it until commit time so we don't fsync(dirfd) too often!

wolfsage (changes requested):
This is causing some cass failures; something's not quite right.

The code *looks* sane, but there's at least one bug it would seem

wolfsage:
@brong I've removed include-in-fastmail for now

wolfsage (approved):

wolfsage (changes requested):

lib/twom.c
```diff
@@ -2191,8 +2191,10 @@ static int replay_cb(void *rock,
 static int tm_rename(struct twom_db *db, struct tm_file *oldfile, const char *newname)
 {
     struct stat sbuf, sbuffile;
-    char *copy = strdup(db->fname);
-    const char *dir = dirname(copy);
+    char *copyd = strdup(db->fname);
+    char *copyb = strdup(db->fname);
+    const char *dir = dirname(copyd);
+    const char *file = dirname(copyb);
```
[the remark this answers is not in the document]
brong: Fixed

PR #5361, opened 2025-03-27
docs: remove "order of operations" for release

rjbs:
This is obsolete.  There is now a marker for "this is latest release".


rjbs:
@elliefm As per our conversation on #5350

elliefm (approved):
I don't want to hold this up picking nits in the wording when I hadn't just written it myself, and it seems fine, so approving

But, "latest release" is actually selected by default, so you have to untick it when doing patch releases for older versions.   Your wording kind of sounds like the reverse, that it's usually unticked, and you need to tick it when doing a patch release for the stable version.

PR #5363, opened 2025-03-28
prometheus: timestamp is current-as-of time, not value-last-changed time

elliefm:
This PR changes the timestamps in the promstatsd "service" report to reflect the time the values were examined and known to be correct at (i.e. the time the report was generated), rather than the time the value in question last changed.  This will stop Prometheus from whinging about old timestamps for certain statistics that almost never change.  It also updates a couple of tests to verify the new behaviour.

The "master" and "usage" reports already handled timestamps like this, and don't need to be updated.

For "gauge" statistics, which can vary both up and down, we will no longer be able to differentiate between ones whose value has not changed since the last report, and ones whose value did change but then changed back.  The difference between these cases used to be whether or not the timestamp changed, but with this PR the timestamp will always change.  We weren't observing the intermediate values anyway, so the lost detail doesn't seem to matter in any practical sense.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Prometheus.pm
+++ b/cassandane/Cassandane/Cyrus/Prometheus.pm
@@ -146,17 +146,36 @@ sub test_service_reportfile_exists
     # do something that'll get counted
     my $imaptalk = $self->{store}->get_client();
     $imaptalk->select("INBOX");
-    # and wait for a fresh report
-    sleep 3;
 
     my $fname = "$self->{instance}->{basedir}/conf/stats/service.txt";
+    my $last_timestamp = 0;
+    my $last_value = undef;
 
-    $self->assert_file_test($fname, '-f');
+    # get a few consecutive reports
+    for (1..3) {
+        sleep 3;
 
-    my $report = parse_report(scalar read_file $fname);
+        $self->assert_file_test($fname, '-f');
 
-    $self->assert(scalar keys %{$report});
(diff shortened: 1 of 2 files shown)


cassandane/Cassandane/Cyrus/Prometheus.pm
```diff
@@ -146,17 +146,36 @@ sub test_service_reportfile_exists
...
-    # and wait for a fresh report
-    sleep 3;
 
     my $fname = "$self->{instance}->{basedir}/conf/stats/service.txt";
+    my $last_timestamp = 0;
+    my $last_value = undef;
 
-    $self->assert_file_test($fname, '-f');
+    # get a few consecutive reports
+    for (1..3) {
```
rsto, line 155: Should this test be marked `_slow`?

elliefm: It's only 9 seconds, so I don't think so.  Didn't bother me in individual testing, and disappears into the results in a larger run.  But I dunno, maybe?

If we think 9 seconds is too long for a non-slow test, then I'll undo the changes to the two existing tests, and put the timestamp checks into their own test which can be marked slow.

ksmurchison: For me 9 sec when run in bulk doesn't matter

rjbs (approved):

wolfsage (approved):
Awesome, thanks

PR #5365, opened 2025-03-31
master pselect timing stats

elliefm:
Temporary PR for Fastmail...

This adds some statistics collection so we can find out how the pselect in master's main loop behaves in practice.  It'll log a line like this (wrapped for web reading) after every 10,000 pselect calls, or at master process exit:

```
2025-04-04T11:24:47.489761+11:00 debian 002441030B/master[1078003]: pselect stats:
samples=<10> timeout=<0|9.9997150000000001|9.8705120000000015|6.9219662999999994|4.7767802569057851>
n_timed_out=<3> timed_out=<1.9999999999999999e-06|1.9999999999999999e-06|1.9999999999999999e-06|1.9999999999999999e-06|0>
n_interrupted=<3> interrupted=<0.00093400000000000004|0.055022000000000001|0.011181|0.022379|0.028730199076929488>
n_ready=<4> ready=<0.00073700000000000002|0.124815|0.0012049999999999999|0.031990499999999998|0.061883393262382333> 
```

The numbers are `<min|max|median|mean|stddev>`.  `timeout` is the timeout parameter we passed to pselect (i.e. how long pselect should wait for anything to happen before returning).  `n_timed_out` and `timed_out` are how many times pselect returned due to timing out, and how long it ran for in those cases.  `n_interrupted` and `interrupted` are how many times pselect returned due to being interrupted by a signal, and how long it ran for in those cases; and `n_ready` and `ready` are how many times pselect returned due to a file descriptor becoming ready, and how long it ran in those cases.

### DIFF
--- a/master/master.c
+++ b/master/master.c
@@ -68,6 +68,7 @@
 #include <arpa/inet.h>
 #include <sysexits.h>
 #include <errno.h>
+#include <float.h>
 #include <limits.h>
 #include <math.h>
 #include <inttypes.h>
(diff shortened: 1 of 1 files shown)


ksmurchison (approved):
WFM

rjbs (approved):
Looks plausible to me.  If we intend to keep this long term, I would want this reworked to use logfmt something like `timed_out.min=1.234 timed_out.max=2.345` etc.  As a short term collection, we can certainly extract and deal with this.

elliefm:
Yeah, I don't expect this to be useful long term, I'm not even certain that it will be useful short term.  And doing it with syslog in any format sucks, this is a rapid prototype at best.

I can parse and deal with the log lines myself easily enough, I just need to have them.  I expect I'll probably want to change what I'm collecting once I've seen real data and thought about what it means.  Maybe a few iterations of that.  But eventually, once I've learned whatever I can from the data,  I expect to rip these changes out and throw them in the bin.

If we do want information like this long term, we should use a real statistics package that we can feed real time raw data to, and let it do the collection/presentation/analysis.

wolfsage (approved):

PR #5367, opened 2025-04-01
cassandane/Makefile: do not run the syntax test by default

rjbs:
We don't actually need to do a syntax test as part of make all.  It can be its own thing, run on demand.  My thinking is that it should be run by the human or robot running the tests, based on whether they think they're likely to *save* or *lose* time by compiling everything up front.  For a two-test run, `make syntax` is probably a waste.  For all the tests, it's just noise.  (On my MacBook Air M2, it takes 7s.)

The next step will be to make `cyd test` run `make syntax` iff there are no positive test specifiers given.  (That is: run it for `cyd test` and `cyd test !Foo` but not `cyd test Bar`.)

### DIFF
--- a/cassandane/Makefile
+++ b/cassandane/Makefile
@@ -41,10 +41,8 @@
 SUBDIRS := utils
 PERL := perl
 
-all clean install::
+all clean install:
 	@for dir in $(SUBDIRS) ; do cd $$dir ; $(MAKE) $@ || exit 1 ; done
-
-all:: syntax
 
 # utils/annotator.pl depends on modules installed with Cyrus, which it
 # will only be able to find when invoked by Cyrus::Instance (which sets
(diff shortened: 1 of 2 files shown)


rjbs:
This can/will be exploited by https://github.com/cyrusimap/cyrus-docker/pull/39

cassandane/Makefile
```diff
@@ -44,8 +44,6 @@ PERL := perl
 all clean install::
 	@for dir in $(SUBDIRS) ; do cd $$dir ; $(MAKE) $@ || exit 1 ; done
```
elliefm, line 46: Since the "all" target no longer appears twice, this rule above no longer needs the [double-colon](https://www.gnu.org/software/make/manual/html_node/Double_002dColon.html)

```suggestion
all clean install:
	@for dir in $(SUBDIRS) ; do cd $$dir ; $(MAKE) $@ || exit 1 ; done

```

elliefm (approved):
Looks good, but one other thing to tidy up while you're in there.  Approving now because it doesn't re-review after that change

rjbs:
Change applied!  I'll merge in not too long.

PR #5368, opened 2025-04-02
JMAPEmail: fix someInThreadHaveKeyword test for split conversations

rsto:
The test passed but actually did not test the right thing, because due to a typo it read the non-existent
"conversations_max_thread" config option instead of the correct "conversation_max_thread" option.

The Perl interpreters barfs at running a for loop with an undefined value as exit condition but does not reject the script as erroneous.

Cyrus already implements someInThreadHaveKeyword for split conversations in JMAP Sieve correctly, and the test now correctly asserts this.


ksmurchison (approved):

rsto:
Unfortunately I pushed this PR too early. There's still something to fix. Reverting to Draft.

rsto:
This is a rewrite of these tests:

```
    JMAPEmail: add regression tests for xxxInThreadHaveKeyword queries

    While we already have some tests for the someInThreadHaveKeyword
    criteria and friends, none of them systematically tested these
    criteria for JMAP Sieve and Email/query using uidsearch and
    guidsearch.

    This patch adds regression tests that assert the current behavior.
```

ksmurchison (approved):

PR #5369, opened 2025-04-03
message.c: do not split conversations for messags with '$memo' flag

rsto:
Messages with the '$memo' flag are meant to store notes associated with another email. Such messages must not cause the conversation to split, as otherwise the note and its related email would end in separate threads.


ksmurchison (approved):
Assuming CI passes LGTM

PR #5370, opened 2025-04-04
Asan CI Builds

wolfsage:
This adds an ASAN Build to all Pull Req / mainline CI runs.

Regular tests take about 17minutes, the asan build takes 22m total.

### DIFF
--- a/cassandane/Cassandane/Cyrus/TestCase.pm
+++ b/cassandane/Cassandane/Cyrus/TestCase.pm
@@ -474,6 +474,24 @@ magic(NoCheckSyslog => sub {
     my $self = shift;
     $self->{no_check_syslog} = 1;
 });
+magic(SuppressLSAN => sub {
+    my ($self, $patterns) = @_;
+
+    my @patterns = split(" ", $patterns);
+    unless (@patterns) {
+        die ":SuppressLSAN requires a list of arguments";
+    }
+
+    my ($fh, $tmp) = tempfile('lsan.suppressXXXXX', OPEN => 1,
+                              DIR => $self->{instance}->{basedir} . "/tmp");
+    for my $pattern (@patterns) {
+        print { $fh } "leak:$pattern\n";
+    }
+
+    close($fh);
+
+    $self->{lsan_suppressions} = $tmp;
+});
 magic(JmapMaxCalendarEventNotifs => sub {
     my $conf = shift;
     # set to some small number
@@ -526,6 +544,11 @@ sub _run_magic
     if (defined $sub) {
         foreach my $a (attributes::get($sub))
         {
+            my $args;
+            if ($a =~ s/\((.*)\)//) {
+                $args = $1;
+            }
+
             my $m = lc($a);
             # ignore min/max version attribution here
             next if $a =~ m/^(?:min|max)_version_/;
@@ -537,7 +560,7 @@ sub _run_magic
                 unless defined $magic_handlers{$m};
             next if $seen{$m};
             $self->{_current_magic} = "Magic attribute $a";
-            $magic_handlers{$m}->($self);
+            $magic_handlers{$m}->($self, $args);
             $self->{_current_magic} = undef;
             $seen{$m} = 1;
         }
@@ -560,6 +583,8 @@ sub _create_instances
 
     my $want = $self->{_want};
     my %instance_params = %{$self->{_instance_params}};
+
+    $instance_params{lsan_suppressions} = "$self->{lsan_suppressions}";
 
     my $cassini = Cassandane::Cassini->instance();
 
--- a/cassandane/Cassandane/Instance.pm
+++ b/cassandane/Cassandane/Instance.pm
@@ -120,6 +120,7 @@ sub new
         install_certificates => 0,
         _pid => $$,
         smtpdaemon => 0,
+        lsan_suppressions => "",
     };
 
     $self->{name} = $params{name}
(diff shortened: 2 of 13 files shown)


wolfsage:
@elliefm Would really appreciate your review here.

I've added everyone else to make sure folks are okay with 22m build times instead of 17m... 

We can maybe get that down in other ways...

ksmurchison (approved):
WFM

rjbs:
Build time increase is okay with me.  It'd be nice to trim it down, but we should generally have already cass tested locally.

rjbs (approved):

cassandane/Cassandane/Instance.pm
```diff
@@ -1213,13 +1213,21 @@ sub start_httpd {
...
     xlog "start $self->{description}: basedir $self->{basedir}";
 
+    if ($params{lsan_suppressions}) {
+        my $current = $ENV{LSAN_OPTIONS} // "";
+
+        $ENV{LSAN_OPTIONS} = "$current:suppressions=$params{lsan_suppressions}";
+
+        xlog "running with LSAN_OPTIONS=$ENV{LSAN_OPTIONS}";
+    }
+
```
elliefm, line 1230: I would have expected `lsan_suppressions` to be an argument to `Instance::new()` that's saved into the Instance object, and set up by `TestCase::_create_instances()` rather than `TestCase::_start_instances()`.  In `Instance::start()`, I'd expect to be using the object's saved value, not examining function call parameters.

A test should be able to stop and start its instances as it likes, without needing to know anything about how to invoke them.  But with this implementation, if you want to do that, you need to rustle up the right lsan_suppressions from somewhere in order to pass it to `start()`.

In particular, there's a bunch of tests that use the :NoStartInstances magic to request a test environment where the instances have been created but not yet started, so they can do other things first (usually overriding config, but a few do other things), and then start the instances themselves when they're ready.  I think these tests are going to be running without ASAN at the moment, since this PR doesn't update them to pass an lsan_suppressions parameter to `start()`, and every time we make a new test like this we're also gonna forget that detail.

elliefm (changes requested):
Broadly fine, but the way you're setting up the suppressions files won't play nicely with tests that start/stop their instances themselves.

cassandane/Cassandane/Instance.pm
elliefm: I'm pretty sure there's also at least one test that stops an instance, does some stuff, and then starts it again.  Offhand, I don't remember which test it is, but I remember this bit of juggling being tricky in some way.

rsto (approved):
👍

cassandane/Cassandane/Cyrus/TestCase.pm
```diff
@@ -471,6 +471,24 @@ magic(NoCheckSyslog => sub {
...
+magic(SuppressLSAN => sub {
+    my ($self, $patterns) = @_;
+
+    my @patterns = split(" ", $patterns);
+    unless (@patterns) {
+        die ":SuppressLSAN requires a list of arguments";
+    }
+
+    my ($fh, $tmp) = tempfile('lsan.suppressXXXXX', OPEN => 1,
+                              DIR => $self->{instance}->{basedir} . "/tmp");
```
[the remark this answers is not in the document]
wolfsage: Nah, it's kept around if you request a filename as I have done. Docs say:
```

          ($fh, $filename) = tempfile($template, UNLINK => 1);

        Return the filename and filehandle as before except that the file is
        automatically removed when the program exits (dependent on $KEEP_ALL).
        Default is for the file to be removed if a file handle is requested
        and to be kept if the filename is requested.
```

cassandane/Cassandane/Instance.pm
wolfsage: I think I just didn't know where else to put it, but your hint here helped, thanks!

I've updated it and I think it's ready for prime time

wolfsage:
@elliefm If you could give this another look I think it's ready to be merged in

elliefm (approved):
This looks good.

One thing though -- ALPN.https_h2 and ALPN.https_multi2 fail when I build and run this locally.  Looks like they're missing the suppressions that all the other ALPN tests have.  There's a patch below that fixes it.

It's a bit weird that they don't fail in CI, but they both require the nghttp2 dependency, so probably our CI builds are missing that library and these two tests get skipped.  That seems like a separate issue.

Please apply the patch (or make equivalent changes), but this doesn't require re-review afterwards, so I'm clicking Approve now.

```
diff --git a/cassandane/Cassandane/Cyrus/ALPN.pm b/cassandane/Cassandane/Cyrus/ALPN.pm
index c50677aae..5a3691290 100644
--- a/cassandane/Cassandane/Cyrus/ALPN.pm
+++ b/cassandane/Cassandane/Cyrus/ALPN.pm
@@ -359,6 +359,7 @@ sub test_https_multi
 
 sub test_https_h2
     :want_service_https :needs_component_httpd :needs_dependency_nghttp2
+    :SuppressLSAN(libcrypto.so libssl.so)
 {
     my ($self) = @_;
 
@@ -380,6 +381,7 @@ sub test_https_h2
 
 sub test_https_multi2
     :want_service_https :needs_component_httpd :needs_dependency_nghttp2
+    :SuppressLSAN(libcrypto.so libssl.so)
 {
     my ($self) = @_;
 
```

wolfsage:
> This looks good.
> 
> One thing though -- ALPN.https_h2 and ALPN.https_multi2 fail when I build and run this locally. Looks like they're missing the suppressions that all the other ALPN tests have. There's a patch below that fixes it.
> 
> It's a bit weird that they don't fail in CI, but they both require the nghttp2 dependency, so probably our CI builds are missing that library and these two tests get skipped. That seems like a separate issue.

Yeah nice catch thanks. I've applied the patch and will think about what we want to do for CI here.

PR #5371, opened 2025-04-04
Ci unroll matrix

wolfsage:
This is ready for review.

It makes the ci builds be explicit rather than using the matrix combination strategy with exceptions for different options.

It also names the builds explicitly so the names are clearer/more descriptive

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,6 +40,9 @@ jobs:
             cflags: "-Og"
             cxxflags: "-Og"
             skip-cass: 1
+          - compiler: "--gcc"
+            desc: "Non-UTC timezone"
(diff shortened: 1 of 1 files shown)


elliefm:
> separating them means it's more obvious what the related failures are.

Ooh excellent point. we could still improve the naming and organisation somewhat, but yeah I agree, let's keep them separate

elliefm:
Though... all the cassandane jobs also run `make check`, so they'll fail on unit test failures too.  In those cases it'll be obvious which thing failed, because if cunit fails then cass won't run, and the job's run time will be a few minutes instead of ~18 minutes.

I dunno I don't think we need to do anything particular here now, but if we keep thinking about it from time to time, eventually we'll have a coherent idea of how we'd like it to all hang together, and then we can make it do that.

PR #5374, opened 2025-04-08
jmap_mail_query: fix Sieve jmapquery not matching Email/query

rsto:
This fixes a couple of issues where the current Sieve jmapquery implementation evaluated Email query filters differently than Email/query. Query text containing emoji and punctuation was especially error-prone.

For testing, this patch adds test 'email_query_sieve_regression' which combines formerly separate sieve test cases into a single regression test file, and adding additional test cases.


cassandane/tiny-tests/JMAPEmail/email_query_sieve_regression
```diff
@@ -0,0 +1,250 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_query_sieve_regression
+  : needs_component_sieve : NoMunge8Bit : RFC2047_UTF8 {
+  my ($self)  = @_;
+  my $jmap    = $self->{jmap};
+  my $imap    = $self->{store}->get_client();
+  my $uuidgen = Data::UUID->new;
```
elliefm, line 9: [Just a drive by comment, didn't notice at first that this is still in draft]

The spaces between the `:` and the attribute names are jarring.  I'm a little surprised it even works, which is interesting.  I've only ever seen them written without the space, and had assumed that was important.

Our style wants the `{` on its own line, and four space indents (not two).  Was this run through perltidy or something?

rsto: Yes, I used perltidy. I am not aware of any particular style guide that we have for Perl in cyrus, and if we do I would prefer it to be documented or preferably automated. That being said, I'll manually fix the points you raised.

rsto: Updated accordingly.

ksmurchison (approved):
LGTM, especially if it works

brong (approved):
I haven't read it in really close detail but the logic looks sensible and the structures make sense.

PR #5377, opened 2025-04-09
Search non conversations mailbox

brong:
Found in production at Fastmail - if you search in a folder that has never had any messages via JMAP, it will resolve the query expression to OP_FALSE, then diligently open every single mailbox, load the index, and check every message against OP_FALSE to see if it matches.

There's a slight efficiency inside index.c at least, that it knows the message can never match and doesn't load in everything, but still - better to just bail!


rsto (approved):
LGTM. That Cassandane test looks to be testing more than is necessary, but that's certainly better than the other way round.

PR #5379, opened 2025-04-10
Calalarm cleanup replica

brong:
When we've done recent failovers at Fastmail, the calalarmd process has picked up a LOT of expunged records to check for alarms, causing it to peg for multiple minutes and delay alarms and scheduled sends until it catches up.

In testing, I discovered that the problem was more general but on masters, the running calalarmd would clean things up earlier.

The bug - when we deleted caldav_alarm.sqlite3 events, we were using dav.mailbox, which is now the uniqueid - but the events database is still by name to allow user grouping!  This PR uses the mailbox name directly instead for the delete, and the accompanying test checks for actual expected behaviour.

### DIFF
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -3776,7 +3776,7 @@ static int mailbox_update_caldav(struct mailbox *mailbox,
     /* has this record already been replaced?  Don't write anything */
     if (cdata->dav.imap_uid > new->uid) {
         /* remove associated alarms */
-        caldav_alarm_delete_record(cdata->dav.mailbox, new->uid);
+        caldav_alarm_delete_record(mailbox_name(mailbox), new->uid);
 
         r = IMAP_NO_MSGGONE;
         goto done;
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):

imap/mailbox.c
```diff
@@ -3793,7 +3793,7 @@ static int mailbox_update_caldav(struct mailbox *mailbox,
         r = caldav_delete(caldavdb, cdata->dav.rowid);
     }
     else if (cdata->dav.imap_uid == new->uid) {
-        if (new->internal_flags & FLAG_INTERNAL_EXPUNGED) {
+        if (new->internal_flags & FLAG_INTERNAL_EXPUNGED || record->system_flags & FLAG_DRAFT) {
```
ksmurchison, line 3796: nit, just to be safe, might want to put parens around the '&' arguments

ksmurchison (approved):

PR #5382, opened 2025-04-11
shush known format-nonliteral warnings

elliefm:
This PR shushes individual `-Wformat-nonliteral` warnings in places where we have no choice but to use a non-literal format string.  The intent is that we'll be able to compile with `-Wformat=2` to more aggressively detect misused printf-style functions, and only get errors from actual (new) problems.  I was also able to fix a couple of things that were detected with the higher warning level, so it's already doing its job.

In one location (master), it removes a previous suppression of `-Wformat-security`.  This line isn't actually generating a `-Wformat-security` warning, so I removed the suppression to avoid masking future problems.  promstatsd has a similar piece of code without the suppression, so I have given it the same suppressions and comment.

I have temporarily added `-Wformat=2` to Makefile.am to force the stricter warnings for all builds; **I intend to drop that commit before this PR is merged**.  It's useful in the meantime for having CI for this PR run with the extra warnings, which allowed me to detect and shush a lot of things that clang warned about but GCC didn't.  If we want `-Wformat=2` routinely enabled long term, the right place to add it is probably cyd build.

**Reviewers:** please pull this branch and build/test it locally.  If there's any build errors in your dev environment, let me know and I'll sort those out too.

### DIFF
--- a/imap/xcal.c
+++ b/imap/xcal.c
@@ -121,8 +121,12 @@ const char *icaltime_as_iso_string(const struct icaltimetype tt)
     else if (icaltime_is_utc(tt)) fmt = "%04d-%02d-%02dT%02d:%02d:%02dZ";
     else fmt = "%04d-%02d-%02dT%02d:%02d:%02d";
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-nonliteral"
+    /* format string decided above */
     snprintf(str, sizeof(str), fmt, tt.year, tt.month, tt.day,
              tt.hour, tt.minute, tt.second);
+#pragma GCC diagnostic pop
 
     return str;
 }
@@ -150,7 +154,11 @@ const char *icalvalue_utcoffset_as_iso_string(const icalvalue* value)
     if (s > 0) fmt = "%c%02d:%02d:%02d";
     else fmt = "%c%02d:%02d";
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-nonliteral"
+    /* format string decided above */
     snprintf(str, sizeof(str), fmt, sign, abs(h), abs(m), abs(s));
+#pragma GCC diagnostic pop
 
     return str;
 }
--- a/lib/times.c
+++ b/lib/times.c
@@ -537,7 +537,6 @@ static int breakdown_time_to_iso8601(const struct timeval *t, struct tm *tm,
 {
     int gmtnegative = 0;
     size_t rlen;
-    const char *datefmt = withsep ? "%Y-%m-%dT%H:%M:%S" : "%Y%m%dT%H%M%S";
 
     /*assert(date > 0); - it turns out these can happen, annoyingly enough */
     /*assert(tm->tm_year >= 69);*/
@@ -548,7 +547,9 @@ static int breakdown_time_to_iso8601(const struct timeval *t, struct tm *tm,
     }
     gmtoff /= 60;
 
-    rlen = strftime(buf, len, datefmt, tm);
+    rlen = strftime(buf, len,
+                    withsep ? "%Y-%m-%dT%H:%M:%S" : "%Y%m%dT%H%M%S",
+                    tm);
     if (rlen > 0) {
         switch(tv_precision) {
         case timeval_ms:
--- a/cunit/cyrunit.h
+++ b/cunit/cyrunit.h
@@ -59,11 +59,7 @@ extern void config_read_string(const char *s);
  * which makes it rather hard to see why an assertion failed.  So we
  * replace the macros with improved ones, keeping the same API.
  */
-/* XXX Would like to add __attribute__((format(printf, 6, 7)))
- * XXX to this so the compiler can warn if it's misused, but it looks
- * XXX like it currently gets very confused by the layers of macros
- * XXX and produces bogus warnings. :(
- */
+__attribute__((format(printf, 6, 7)))
 extern CU_BOOL CU_assertFormatImplementation(CU_BOOL bValue,
                                              unsigned int uiLine,
                                              const char strFile[],
--- a/cunit/cyrunit.h
+++ b/cunit/cyrunit.h
@@ -182,7 +182,8 @@ extern int __cunit_wrap_fixture(const char *name, int (*fn)(void));
     CU_assertFormatImplementation(!strcmp(_a?_a:"",_e?_e:""), __LINE__, \
         __FILE__, "", CU_FALSE,                                         \
         "CU_ASSERT_STRING_EQUAL(%s=\"%s\",%s=\"%s\")",                  \
-        #actual, _a, #expected, _e);                                    \
+        #actual, _a ? _a : "(null)",                                    \
+        #expected, _e ? _e : "(null)");                                 \
 } while(0)
 
 #undef CU_ASSERT_STRING_EQUAL_FATAL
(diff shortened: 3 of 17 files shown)


elliefm:
Added Do Not Merge label to remind myself (or anyone else) that the "NOMERGE Makefile.am: ..." commit needs to be dropped before this gets merged.

wolfsage:
Building with clang -O3 gives:
```
cunit/unit.c:296:33: error: format string is not a string literal [-Werror,-Wformat-nonliteral]
    vsnprintf(buf, sizeof(buf), strConditionFormat, args);
```

So we probably need:
```
CU_BOOL __attribute__((format(printf, 6, 7)))
CU_assertFormatImplementation(
```
too

elliefm:
Interesting, I wonder if that's why the attribute on that function was previously commented out.  Will give it a try.

elliefm:
@wolfsage I've fixed the format-overflow warnings from CU_ASSERT_STRING_EQUAL by explicitly handling the possibility of the arguments being null (instead of relying on the glibc behaviour, which I guess the optimiser couldn't/didn't assume), and temporarily added -O3 to AM_CFLAGS and a junk commit to twom to see what CI thinks...

elliefm:
asan and ubsan failing with -O3, but I think not for reasons related to -Wformat=2, so I'll disregard those failures for now

Makefile.am
```diff
@@ -41,10 +41,10 @@
 # OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 #
 
-WARN_CFLAGS = -Wall -Wextra -Werror
+WARN_CFLAGS = -Wall -Wextra -Wformat=2 -Werror
 
 ACLOCAL_AMFLAGS = -I cmulocal
-AM_CFLAGS = $(WARN_CFLAGS) @PERL_CCCDLFLAGS@ $(GCOV_CFLAGS)
+AM_CFLAGS = -O3 $(WARN_CFLAGS) @PERL_CCCDLFLAGS@ $(GCOV_CFLAGS)
```
dilyanpalauzov, line 47: This `-O3` should really be up to the the person compiling the software.  That person might want `-O2`, so it should not be implied by the build system.  The default for Autoconf/Automake is `-O2`, so one can rely on this default. Besides, `-O3` is already present on many other places: `libcrc32_la_CFLAGS = -O3 $(AM_CFLAGS) $(CFLAG_VISIBILITY)`, and many `__attribute__((optimize("-O3")));` in twom.c, util.c, dlist.c and others.

wolfsage: This commit is not to be merged and is only here to help this PR detect more errors.

wolfsage:
(I will try to look at this later this week!)

Makefile.am
brong, line 47: Does this force us to always compile with `-O3` which we didn't before? It reduces the debugability of our builds quite considerably to do that.

imap/xcal.c
```diff
@@ -121,8 +121,12 @@ const char *icaltime_as_iso_string(const struct icaltimetype tt)
     else if (icaltime_is_utc(tt)) fmt = "%04d-%02d-%02dT%02d:%02d:%02dZ";
     else fmt = "%04d-%02d-%02dT%02d:%02d:%02d";
```
brong, line 122: We could just do two separate sprintf statements here rather than creating `fmt` as a variable!

lib/times.c
```diff
@@ -548,7 +547,9 @@ static int breakdown_time_to_iso8601(const struct timeval *t, struct tm *tm,
     }
     gmtoff /= 60;
 
-    rlen = strftime(buf, len, datefmt, tm);
+    rlen = strftime(buf, len,
+                    withsep ? "%Y-%m-%dT%H:%M:%S" : "%Y%m%dT%H%M%S",
```
brong, line 551: Heh, yeah or we could use this pattern in a few more places.

brong (commented):
a couple of comments on this, one of which I'd like to have an answer on before making the change. Ability to attach gdb and step, or to read a core file, are very valuable!

Makefile.am
wolfsage: This is a temporary commit not to be included with the final bit

wolfsage (approved):
LGTM. If we think -wformat=2 is always worth having (and why this MR exists), why do we not want to include it in the defaults yet?

We could easily add another make check phase that includes this at least so CI tells us if we reintroduce anything...

elliefm:
@wolfsage,

> LGTM. If we think -wformat=2 is always worth having (and why this MR exists), why do we not want to include it in the defaults yet?

In the Makefile.am defaults, we set `-Wall` and `-Wextra` (and `-Werror`, but that's not a warning class, that's a behaviour).  We want everyone building Cyrus to get as much warning protection as is reasonable by default, but the compiler gets to choose what it thinks is reasonable (which won't include options that it considers experimental or unreliable).  This gives us a good balance of portability to various compiler versions while getting decent protection, without a lot of false positives.

[In current GCC](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html), `-Wall` sets `-Wformat=1` as well as a bunch of other specific format-foo warnings, but it doesn't set `-Wformat=2`.  To me, this implies that they don't think the higher level format warnings are mature or reliable enough yet to be general purpose.  Not sure where clang's at, I didn't look.

It's also worth observing that the CI gcc and clang versions do detect and warn about different things with `-Wformat=2`.  My earlier attempts passed the gcc CI builds, but I had to review the clang CI reports to track down and fix additional things that only clang noticed.  So even among the two compilers we actively support, `-Wformat=2` behaviour is still inconsistent at present.

I do think `-Wformat=2` is worth using for all FM and CI builds, where we are targeting specific compilers and versions.  But we should configure that in the FM/CI build infrastructure, not Makefile.am.  I believe we already do a similar thing for `-Wwrite-strings`, and maybe some others.

elliefm:
Forgot to finish my thought:

> we should configure that in the FM/CI build infrastructure

... and the intention of this PR is to get us into a position where we can do that

imap/xcal.c
elliefm: My read is that we deliberately split it into a `fmt` variable to make it clear that the only thing that differs between the three cases is the format string.  If each case had its own snprintf call, a reader might reasonably assume there were more differences than just that, and would have to read quite closely to notice that only the format string changes.

Given that `fmt` is always a pointer to a string literal in every code path here, I would hope that in a few years compilers are smart enough to not throw a format-nonliteral warning for this line.  In which case, using the variable is more readable long term, and shushing the not-quite-right warning for now stops it being a nuisance.

But I could be persuaded the other way -- if someone changes one of these cases to like `fmt = strdup("whatever");`, then the warning would be valid, but we wouldn't get it because we shushed it.  But I trust us to not do that, or at least to see the big #pragma block below during code review.

lib/times.c
elliefm: Yeah this pattern is fine here because it's terse -- there's only two cases, and the format strings are short.  The other spots where I left the variable and shushed the warnings had more cases and/or long format strings, so a ternary inline would've been a mess.

imap/xcal.c
brong: yeah, that seems fair

brong (approved):

rsto (commented):
Is there more work required to make CI pass?

elliefm:
@rsto 

> Is there more work required to make CI pass?

No, there is not.  CI will pass before it's merged, but that will happen once the temporary commits that force aggressive warning levels are dropped.

* ASAN and UBSAN are failing because of unrelated issues that are exposed by -O3, which should be addressed, but are out of scope for this PR.
* The fact that the regular Cassandane checks pass for both gcc and clang even with -O3 on proves that I have found and fixed all the -Wformat=2 issues -- at least for the compilers/versions used in CI.

I'd appreciate if you could fetch and build it locally, in case your environment finds something mine or CI didn't.

rsto (approved):
Builds fine on my setup:
```
gcc (Debian 12.2.0-14) 12.2.0

CFLAGS="-g -Og -fPIC -fstack-protector-all -fno-semantic-interposition -Bsymbolic -fno-omit-frame-pointer"
CXXFLAGS="-g -O0 -Wpedantic"
```

PR #5385, opened 2025-04-14
clang-format: disable formatting for JMAP and DAV API struct literals

rsto:
https://github.com/cyrusimap/cyrus-imapd/pull/5332 aims to add clang-format formatting rules for the Cyrus codebase. The JMAP and DAV code in Cyrus uses static struct literals to initialize a couple of common data structures. We want to keep these consistently formatted for every entry, but clang-format seems not be able to apply consistent struct literal formatting across nested struct lists.

This patch disables clang-formatting for the code areas that area known to produce inconsistent struct literal formats in the JMAP and DAV code.

### DIFF
--- a/imap/http_admin.c
+++ b/imap/http_admin.c
@@ -93,6 +93,7 @@ static int action_conf(struct transaction_t *txn);
 
 
 /* Namespace for admin service */
+// clang-format off
 struct namespace_t namespace_admin = {
     URL_NS_ADMIN, 1, "admin", "/admin", NULL,
     http_allow_noauth_get, /*authschemes*/0,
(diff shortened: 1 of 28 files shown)


rsto:
> > clang-format seems not be able to apply consistent struct literal formatting across nested struct lists.
> 
> Is there at llvm a bug report for this?

Not that I am aware. I am also not sure to call it a bug, more a matter of formatting preferences. What's "inconsistent" for us in this context is: nested struct literals seem to be only formatted by taking the respective literal into account, not how previous and following literals in a list were formatted. We prefer to use the same formatting for all literals, even if some of those might fit on less lines.

PR #5389, opened 2025-04-16
Add WebSocket testing capabilities to Cassandane

wolfsage:
This uses JMAP::Tester::WebSocket to provide a WebSocket testing interface.

It also provides a regular JMAP::Tester if one wants to use that too....

The big change here is that this *requires* JMAP::Tester and JMAP::Tester::WebSocket be installed (which our docker images already do).

If we're happy with that idea, then we can update the Cassandane docs to mention the new requirements. If we aren't, then we'll have to make `$self->jmap_tester` only usable if JMAP::Tester is detected, and `$self->jmap_tester_ws` only usable is JMAP::Tester::WebSocket is detected. @elliefm This is more a question for you I suppose! 

Beyond that, we may need tests using ::WebSocket specify `:needs_dependency_wslay` if we don't consider wslay support a core required build option...


rjbs:
I haven't reviewed yet, but:  I am fine with making the JMAP::Tester libraries, and wslay, required.  Part of v4 "fewer knobs" project…

cassandane/tiny-tests/JMAPContacts/ws_contact_set_after_reconstruct
```diff
@@ -0,0 +1,44 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_ws_contact_set_after_reconstruct
```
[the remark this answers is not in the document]
wolfsage: The test currently fails because this is showing the existence of an active bug, so I think the commit message is accurate?

cassandane/Cassandane/Cyrus/TestCase.pm
```diff
@@ -917,6 +917,12 @@ sub _setup_http_service_objects
 
         my $ws_base = $base =~ s/^http:/ws:/r;
 
+        # Ugh. We must load AnyEvent::Loop before JMAP::Tester::WebSocket,
+        # otherwise AnyEvent in Cassandane::Instance::notifyd will use
+        # AnyEvent::Impl::IOAsync which is ... not actually running (except
+        # when ::WebSocket makes a request) ... which will make the perl
+        # notifyd process hang and lock up the tests
```
rjbs, line 924: Software is bad, but people are worse.

cassandane/tiny-tests/JMAPContacts/ws_contact_set_after_reconstruct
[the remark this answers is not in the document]
rjbs: Oh okay!

rsto:
Should the WebSocket changes and the test that demonstrates an issue with dav_reconstruct be split into two PRs? I won't find time to look at the dav_reconstruct in this and next week.

wolfsage:
> Should the WebSocket changes and the test that demonstrates an issue with dav_reconstruct be split into two PRs? I won't find time to look at the dav_reconstruct in this and next week.

Yeah that's the plan, I'd merge just the tooling then open a branch with the broken test you can cherry-pick from or push to directly

elliefm (commented):
> we'll have to make $self->jmap_tester only usable if JMAP::Tester is detected, and $self->jmap_tester_ws only usable is JMAP::Tester::WebSocket is detected

I think that will be the right approach for now.  We do something similar already for the JMAPTestSuite and JMAPTestSuiteWS suites, see subs `list_tests` and `run_test`.

> Beyond that, we may need tests using ::WebSocket specify :needs_dependency_wslay if we don't consider wslay support a core required build option...

Yep, will need this too, and :needs_component_httpd, and :want_service_http(s), ... as annoying as all that is.  Though usually all the tests in a suite would need the same stuff, and you'd just register their needs once during new().  The per-test annotations are meant for per-test dependencies.

When we make wslay (et al) mandatory, we can rip out the test conditionalisation at the same time as we're ripping out the ifdefs and such; but as long as stuff is optional, tests that need an optional thing must declare that they need it.

elliefm:
Also, have a look at tiny-tests/JMAPCore/echo and tiny-tests/JMAPCore/echo_tls.  You might wanna add a similar pair for WS, to prove it's possible to get a working WS connection with/without TLS.

rsto (changes requested):
Please move the breaking test out of the PR.

cassandane/Cassandane/Instance.pm
```diff
@@ -3242,16 +3250,19 @@ sub new_jmaptester_for_user ($self, $user, $new_arg = undef) {
...
         unless exists $ENV{JMAP_TESTER_LOGGER};
 
-    my $jtest = Cassandane::JMAPTester->new({
+    my $jtest = $tester_class->new({
         fallback_account_id => $user->username,
         default_using => [ @DEFAULT_USING ],
         %overrides,
+        ( $tester_class eq 'Cassandane::JMAPTesterWS'
+            ? (cache_connection => 1, ws_api_uri => '') : ()
+        ),
```
rjbs: This is really up to you, but I don't like that we're checking the $tester_class, here.  I think better would be to move the setting of cache_connection and ws_api_uri into `new_jmaptester_ws_for_user` so that the generic method is truly generic, and the specialization happens in the specialization code.

cassandane/Cassandane/JMAPTesterRole.pm
```diff
@@ -0,0 +1,150 @@
+package Cassandane::JMAPTesterRole;
```
rjbs: It's more convention in our code to have X::Role::Y rather than X::YRole, and I'd prefer that, but won't withhold an approve for this little detail.

cassandane/Cassandane/TestUser.pm
```diff
@@ -19,6 +19,15 @@ has jmap => (
...
 );
 
+has jmap_ws => (
+    is => 'ro',
+    lazy => 1,
+    default => sub {
+        my ($self) = @_;
+        $self->new_jmaptester_ws;
+    }
+);
```
rjbs, line 29: It's fine, but my guess is that we don't need this.  How often will we want to cache a JMAP::Tester for WebSockets on a user, as opposed to just getting it in the test?  (Honestly, I think `->jmap` should probably go, and the test can do a new_jmaptester once and keep the result.  Based on my experience refactoring, I think this'll be better.)

rjbs (approved):
I am approving, because it'll be good to get this in place for testing WS.  That said, I was tempted to haggle on some small issues.  I think that the schema and password setting methods can and should be implemented in the role, with the HTTP/WS-specific bits being performed by method modifiers in the specific classes.  That might be it.  It's small potatoes, which is why I'm just going to click approve and probably never think about this again.  But if I do, maybe I'll ask you to review a patch to do it…

cassandane/Cassandane/TestUser.pm
wolfsage: My habit (at least in newt) is `$user->tester->request()` most of the time, which avoids more newlines and lexicals, so IMO cleaner in some ways, but I could be convinced otherwise.

cassandane/Cassandane/JMAPTesterRole.pm
wolfsage: Done

cassandane/Cassandane/Instance.pm
wolfsage: Good call. Done

wolfsage:
> I think that the schema and password setting methods can and should be implemented in the role,

Good call, done!

rjbs (approved):

PR #5391, opened 2025-04-16
Cassandane: Report on all errors, even those in cleanup, and those that are weird

wolfsage:
With these fixes:

* `die([])` or `Some::Exception::Class->throw()` will be caught/reported properly
* If syslog errors or other problems are reported in tear_down, they won't overshadow higher up errors

### DIFF
--- a/cassandane/Cassandane/Unit/FormatPretty.pm
+++ b/cassandane/Cassandane/Unit/FormatPretty.pm
@@ -157,15 +157,23 @@ sub print_errors
     $self->_print($msg);
 
     my $i = 0;
+
+    my $annotations;
+
     for my $e (@{$result->errors()}) {
         my ($test, $errors) = split(/\n/, $e->to_string(), 2);
         chomp $errors;
         my $prettytest = _prettytest($test);
         $self->_print("\n") if $i++;
         $self->_print($self->ansi([31], "$i) $prettytest") . "\n$errors\n");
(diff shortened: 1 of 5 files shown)


cassandane/Cassandane/Unit/FormatPretty.pm
```diff
@@ -157,16 +157,24 @@ sub print_errors
...
         chomp $errors;
         my $prettytest = _prettytest($test);
         $self->_print("\n") if $i++;
         $self->_print($self->ansi([31], "$i) $prettytest") . "\n$errors\n");
-        $self->_print("\nAnnotations:\n", $e->object->annotations())
-          if $e->object->annotations();
+
+        # These will always be the same since they share the same test object
+        # so we just need one of them...
+        $annotations ||= \$e->object->annotations();
```
elliefm, line 172: I feel like there was a reason this was the way it was, but I don't remember it.  It might just be that I saw the same redundancy and decided to ignore it instead of fix it.  This is one of those things where Cassandane deeply inverts the way Test::Unit wants to work, and then you end up with a mess that's hard to reason about consistently, with little bits of monkey patching everywhere...

elliefm (commented):
So I think the main change being made here is that a single test can now report multiple errors, instead of only one?  Like, perhaps the test itself fails due to an assertion, and then the post-test sanity checks fail because something was left in a state that reconstruct didn't like, and now we'll see both problems, whereas before we just saw one or the other?

PR #5392, opened 2025-04-16
64bit nanosecond jmapids bis

ksmurchison:


imap/sync_support.c
```diff
@@ -1461,34 +1461,77 @@ static void encode_annotations(struct dlist *parent,
...
-            annots = dlist_newlist(parent, "ANNOTATIONS");
-        aa = dlist_newkvlist(annots, NULL);
-        dlist_setatom(aa, "ENTRY", IMAP_ANNOT_NS "savedate");
-        dlist_setatom(aa, "USERID", "");
-        dlist_setnum64(aa, "MODSEQ", 0);
-        dlist_setnum32(aa, "VALUE", record->savedate);
-    }
+    switch (mailbox->i.minor_version) {
+    case 20:
+        if (record->internaldate.tv_sec != UTIME_OMIT) {
```
brong, line 1468: should this be tv_nsec?

ksmurchison: Yes!  Good catch

imap/conversations.c
```diff
@@ -986,8 +994,8 @@ static void conv_to_buf(conversation_t *conv, struct buf *buf, int flagcount)
         nn = dlist_newlist(n, "THREAD");
         dlist_setguid(nn, "GUID", &thread->guid);
         dlist_setnum32(nn, "EXISTS", thread->exists);
-        dlist_setnum32(nn, "INTERNALDATE", thread->internaldate);
-        dlist_setnum32(nn, "CREATEDMODSEQ", thread->createdmodseq);
+        dlist_setnum64(nn, "INTERNALDATE", thread->internaldate);
```
brong, line 997: I'd like to see this given a different name, since it's a different value (nanosec, not seconds)

imap/conversations.c
```diff
@@ -1514,7 +1522,7 @@ int _saxconvparse(int type, struct dlistsax_data *d)
             return 0;
 
         case 2:
-            rock->thread->internaldate = atol(d->data);
+            rock->thread->internaldate = atoll(d->data);
```
brong, line 1525: ditto this field which is now nanoseconds

imap/conversations.c
```diff
@@ -1799,8 +1807,12 @@ static int _thread_datesort(const void **a, const void **b)
 {
     const conv_thread_t *ta = (const conv_thread_t *)*a;
     const conv_thread_t *tb = (const conv_thread_t *)*b;
+    struct timespec ta_internaldate, tb_internaldate;
+
+    TIMESPEC_FROM_NANOSEC(&ta_internaldate, ta->internaldate);
+    TIMESPEC_FROM_NANOSEC(&tb_internaldate, tb->internaldate);
 
-    int r = (ta->internaldate - tb->internaldate);
+    int r = (ta_internaldate.tv_sec - tb_internaldate.tv_sec);
```
brong, line 1815: Do we want to sub-sort by nanoseconds as well if r == 0?  It's still going to be stable and better than GUID comparison

brong (changes requested):
There's a couple of queries around internaldate handling and naming I'd like to have resolved.

imap/append.c
```diff
@@ -910,8 +939,14 @@ static int findstage_cb(const conv_guidrec_t *rec, void *vrock)
             if (file) {
                 struct body *body = NULL;
                 r = message_parse_file(file, NULL, NULL, &body, msgpath);
-                if (!r && !strcmp(rock->guid, message_guid_encode(&body->guid)))
+                if (!r && !strcmp(rock->guid, message_guid_encode(&body->guid))) {
                     rock->fname = xstrdup(msgpath);
+                    if (!rock->dupcheck.internaldate ||
```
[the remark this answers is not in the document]
ksmurchison: No.  dupcheck.internaldate is a pointer so if we haven't assigned it, we're dealing with a non-email message

imap/conversations.c
ksmurchison: Yeah, just comparing the nanosec since epoch is stable and deterministic.  Fixed

imap/jmap_mail.c
```diff
@@ -13668,16 +13813,19 @@ static int jmap_email_set(jmap_req_t *req)
...
+            atomodseq_t(if_in_state) != old_modseq) {
             jmap_error(req, json_pack("{s:s}", "type", "stateMismatch"));
             goto done;
         }
         set.old_state = xstrdup(set.if_in_state);
     }
-    else {
-        set.old_state = modseqtoa(jmap_modseq(req, MBTYPE_EMAIL, 0));
-    }
+
```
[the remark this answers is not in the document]
wolfsage: Mm, no cass wouldn't be able to detect a problem I think. If `inInState` doesn't match then you get an error, but if it does then both of these sets should be setting the same value, so it's purely up to asan/valgrind catching this I think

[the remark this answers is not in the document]
ksmurchison: Fixed

imap/jmap_mail.c
```diff
@@ -14562,33 +14572,29 @@ static int jmap_email_copy(jmap_req_t *req)
     srcinbox = mboxname_user_mbox(copy.from_account_id, NULL);
     dstinbox = mboxname_user_mbox(req->accountid, NULL);
 
+    int r = conversations_open_user(copy.from_account_id, 0, &from_cstate);
+    if (r) {
+        syslog(LOG_ERR,
+               "jmap_email_copy: can't open source conversations: %s",
```
rsto, line 14578: Use xsyslog here and maybe replace this with something like "can't open conversations: fromAccountId=<%s> error=<%s>"?

cassandane/Cassandane/Cyrus/Conversations.pm
```diff
@@ -213,6 +213,10 @@ sub test_move_200
     my ($self) = @_;
     my %exp;
 
+    xlog $self, "Enable compactids";
```
brong: This feels like what we really want is a `:CompactIds` magic, though I see later that we're going to run into issues with it being for multiple users... should we have a config option that always turns them on for new users instead in the Cass tests?

imap/mailbox.c
```diff
@@ -2330,7 +2330,7 @@ static int _commit_one(struct mailbox *mailbox, struct index_change *change)
                    session_id(), mailbox_name(mailbox), mailbox_uniqueid(mailbox), record->uid,
                    record->modseq, flagstr,
                    message_guid_encode(&record->guid),
-                   conversation_id_encode(cstate ? cstate->version : 2, record->cid),
```
brong: Can this be called without cstate?  It used to have a guard and now doesn't

imap/conversations.c
```diff
@@ -3074,7 +3079,7 @@ EXPORTED const char *conversation_id_encode(int cstate_version,
 
     if (cid == NULLCONVERSATION) {
         buf_setcstr(&buf, "NIL");
-    } else if (cstate_version >= 2) {
+    } else if (USER_COMPACT_EMAILIDS(state)) {
```
brong, line 3090: here it is!  In the case without a cstate below, this used to be true but is now false, because it was `cstate ? cstate->version : 2` which would have returned true, but now it's got a test that requires cstate to be true.

brong (approved):
@ksmurchison I have comments around the change in behaviour for cstate being NULL, but otherwise this looks great

imap/ctl_conversationsdb.c
```diff
@@ -1156,6 +1194,7 @@ int main(int argc, char **argv)
             break;
 
         default:
+fprintf(stderr, "WTF!\n");
```
wolfsage: Debug I'm guessing

imap/conversations.c
ksmurchison: I think the old logic was incorrect.  If we don't have cstate, we can't have the use compact ids key set

cassandane/Cassandane/Cyrus/Conversations.pm
ksmurchison: What would that look like?  We can't set the key in conv.db until the user has been created

imap/append.c
```diff
@@ -266,7 +266,7 @@ EXPORTED int append_commit(struct appendstate *as)
 
     if (as->nummsg) {
         /* Calculate new index header information */
-        as->mailbox->i.last_appenddate = time(0);
+        as->mailbox->i.last_appenddate.tv_sec = time(0);
```
brong, line 269: Should this be using:
```
gettimeofday(&as->mailbox->i.last_appenddate, 0);
```
Rather than leaving the rest of the last_appenddate either zero or unassigned?

imap/append.c
```diff
@@ -910,8 +942,14 @@ static int findstage_cb(const conv_guidrec_t *rec, void *vrock)
             if (file) {
                 struct body *body = NULL;
                 r = message_parse_file(file, NULL, NULL, &body, msgpath);
-                if (!r && !strcmp(rock->guid, message_guid_encode(&body->guid)))
+                if (!r && !strcmp(rock->guid, message_guid_encode(&body->guid))) {
                     rock->fname = xstrdup(msgpath);
```
brong, line 946: Can we leak memory here if the same message exists multiple times expunged?  It looks like we either:
a) don't find the file if it's always expunged or
b) find the file but don't notice the timestamp if we hit an expunged one first.

I THINK it's that we'll leak memory by setting rock->fname multiple times.

imap/append.c
```diff
@@ -1069,9 +1130,15 @@ EXPORTED int append_fromstage_full(struct appendstate *as, struct body **body,
     }
 
     /* And make sure it has a timestamp */
-    r = msgrecord_get_internaldate(msgrec, &internaldate);
-    if (!r && !internaldate)
-        r = msgrecord_set_internaldate(msgrec, time(NULL));
+    struct timespec now;
+    if (!internaldate || internaldate->tv_nsec >= UTIME_OMIT) {
```
brong, line 1134: Wow!  I assume this is "if it's UTIME_OMIT or UTIME_NOW" based on their actual values.  Sneaky.

imap/append.c
```diff
@@ -1069,9 +1130,15 @@ EXPORTED int append_fromstage_full(struct appendstate *as, struct body **body,
...
+        clock_gettime(CLOCK_REALTIME, &now);
+        if (!internaldate)
+            internaldate = &now;
+        else
+            internaldate->tv_nsec = now.tv_nsec;
```
brong, line 1139: Is this supposed to set the nsec but NOT the sec?  That seems to be a meaningless value.

imap/conversations.c
```diff
@@ -1818,16 +1823,9 @@ static int _thread_datesort(const void **a, const void **b)
...
-
-    // if same internaldate, use createdmodseq for a stable sort
-    int64_t r2 = (ta->createdmodseq - tb->createdmodseq);
-    if (r2 < 0) return -1;
-    if (r2 > 0) return 1;
-
-    return message_guid_cmp(&ta->guid, &tb->guid);
+    if (ta->nano_internaldate < tb->nano_internaldate) return -1;
+    if (ta->nano_internaldate > tb->nano_internaldate) return 1;
+    return 0;  // should never happen
```
brong, line 1831: This is only true with the latest version, right?  Do we still need a guid_cmp to tiebreak?

imap/conversations.c
```diff
@@ -3044,23 +3075,33 @@ EXPORTED int conversations_prune(struct conversations_state *state,
...
     } else {
-        strncpy(text, "NIL", sizeof(text));
+        buf_printf(&buf, CONV_FMT, cid);
     }
 
-    return text;
+    return buf_cstring(&buf);
 }
 
 EXPORTED int conversation_id_decode(conversation_id_t *cid, const char *text)
```
brong, line 3100: is this called often enough in commands that we should optimise it by only doing a strlen once?

imap/conversations.c
```diff
@@ -3359,4 +3408,109 @@ EXPORTED int conversations_zero_modseq(struct conversations_state *state)
...
+       - to prevent strcat() from overrunning a fixed-length buffer
+       - to short-circuit an ill-fated DB fetch
+    */
+    idlen = strspn(jidrep, jmapid_alphabet);
+    if (idlen != CONV_JMAPID_SIZE || jidrep[idlen] != '\0') {
+        /* we know this J key can't exist */
+        return CYRUSDB_NOTFOUND;
+    }
+
+    strcat(key, jidrep);
```
brong, line 3435: we could use memcpy for this since the length is explicitly known.

imap/append.c
brong, line 1134: This function gets called AFTER we've done the conversationsdb "check that this internaldate isn't already in use.  I assume we're relying on the idea that nobody messed with the clock so accurately that they could possibly set a future date and have use hit the same ID - rather than calculating this BEFORE we do that check?

imap/imapd.c
```diff
@@ -7324,7 +7340,21 @@ static void cmd_create(char *tag, char *name, struct dlist *extargs, int localon
     }
 
     /* Close newly created mailbox before writing annotations */
-    mailboxid = xstrdup(mailbox_uniqueid(mailbox));
+    struct conversations_state *cstate = mailbox_get_cstate(mailbox);
+    if (is_inbox && compactids && (!minor_version || minor_version >= 20)) {
```
brong, line 7376: Do we want to have an:
```
if (!is_inbox && compactids) goto bad;
```

imap/index.c
```diff
@@ -5596,7 +5635,8 @@ MsgData **index_msgdata_load(struct index_state *state,
 #endif
                 if (!cur->savedate) {
                     /* If not snoozed in mailboxId, we use receivedAt */
-                    cur->internaldate = record.internaldate;
```
brong, line 5604: wouldn't the old form have kept working?  Copying a struct over a struct copies all the fields.

imap/jmap_blob.c
```diff
@@ -841,20 +843,24 @@ static int jmap_blob_lookup(jmap_req_t *req)
...
 
                 switch (item->typenum) {
-                case DATATYPE_MAILBOX:
-                    strarray_add(ids, uniqueid);
+                case DATATYPE_MAILBOX: {
+                    char mboxid[JMAP_MAX_MAILBOXID_SIZE];
+                    jmap_set_mailboxid(req->cstate, mbentry, mboxid);
+                    strarray_add(ids, mboxid);
                     break;
+                }
```
brong, line 851: I think the indent on this line is wrong compared the surrounding

imap/jmap_mail.c
```diff
@@ -14422,43 +14594,59 @@ static int jmap_email_copy(jmap_req_t *req)
     srcinbox = mboxname_user_mbox(copy.from_account_id, NULL);
     dstinbox = mboxname_user_mbox(req->accountid, NULL);
 
+    int r = conversations_open_user(copy.from_account_id, 0, &from_cstate);
```
brong, line 14597: This looks like it's now opening both conversations.db files when it used to only need the destination, which means that we've now got a lock inversion bug.  If you're opening more than one, you MUST open the one with the alphabetically earlier INBOX name (yeah, really).  See imap/imapd.c:
```
    struct mboxlock *oldnamespacelock = NULL;
    struct mboxlock *newnamespacelock = NULL;

    if (strcmpsafe(oldmailboxname, newmailboxname) < 0) {
        oldnamespacelock = mboxname_usernamespacelock(oldmailboxname);
        newnamespacelock = mboxname_usernamespacelock(newmailboxname);
    }
    else {
        newnamespacelock = mboxname_usernamespacelock(newmailboxname);
        oldnamespacelock = mboxname_usernamespacelock(oldmailboxname);
    }
```

We should probably do this in the outer layer of jmap_api.c

imap/lmtp_sieve.c
```diff
@@ -191,7 +191,9 @@ static int getmailboxidexists(void *sc, const char *extname)
 {
     script_data_t *sd = (script_data_t *)sc;
     const char *userid = mbname_userid(sd->mbname);
-    char *intname = mboxlist_find_uniqueid(extname, userid, sd->authstate);
+    char *intname = (*extname == JMAP_MAILBOXID_PREFIX) ?
```
brong, line 194: This code pattern is repeated enough times that it would be handy to have a MACRO for it!

imap/lmtpd.c
```diff
@@ -528,7 +528,7 @@ int deliver_mailbox(FILE *f,
     char *uuid = NULL;
     duplicate_key_t dkey = DUPLICATE_INITIALIZER;
     quota_t qdiffs[QUOTA_NUMRESOURCES] = QUOTA_DIFFS_DONTCARE_INITIALIZER;
-    time_t internaldate = 0;
+    struct timespec internaldate = { 0 };
```
brong, line 531: we don't explicitly zero this field in lots of other places, do we want a standard approach for that?

imap/mboxlist.c
```diff
@@ -979,6 +1000,116 @@ EXPORTED int mboxlist_lookup_by_uniqueid(const char *uniqueid,
     return 0;
 }
 
+/*
+ * read a single _J_mapid record from the mailboxes.db and return a pointer to it
+ */
+static int mboxlist_read_jmapid(const char *inboxid, const char *jmapid,
```
brong, line 1008: Is this a `userid` or an `inboxid`?  I don't like `inboxid` because it's unclear what it means.  We should use userid consistently for userids.

imap/mboxlist.c
```diff
@@ -433,6 +440,16 @@ static void mboxlist_id_to_key(const char *id, struct buf *key)
     buf_appendcstr(key, id);
 }
 
+static void mboxlist_jmapid_to_key(const char *userid,
```
brong, line 445: likewise we should use 'jmapid' consistently rather than different names!

imap/mboxlist.c
```diff
@@ -979,6 +1000,116 @@ EXPORTED int mboxlist_lookup_by_uniqueid(const char *uniqueid,
...
+               jmapid, cyrusdb_strerror(r));
+        r = IMAP_IOERROR;
+        break;
+    }
+
+    buf_free(&key);
+    return r;
+}
+
+EXPORTED char *mboxlist_find_jmapid(const char *jmapid,
```
brong, line 1053: Is there a reason why the argument order on this public API is different than the argument order of the functions which it calls?  userid vs jmapid.

imap/statuscache_db.c
```diff
@@ -409,7 +410,7 @@ HIDDEN void status_fill_mailbox(struct mailbox *mailbox, struct statusdata *sdat
 
     // mbentry items are also available from an open mailbox
     sdata->uidvalidity = mailbox->i.uidvalidity;
-    sdata->mailboxid = mailbox_uniqueid(mailbox);
+    // sdata->mailboxid set in imapd_statusdata()
```
brong: Is there a reason why we moved it there?  Yes it's dependent on cstate, but it will always be the same for the same mailbox, and if we store it in the statuscache then we can read it without needing to recalculate it.

Part of the point of statuscache (maybe it's not needed so much any more!) was to avoid opening so many things to get status information.

brong:
@murch I have added three commits on top of this PR, adding J key split and also J key conversion on startup with the new version, as well as support for reading these values in the Mbentry Perl code.

imap/ctl_mboxlist.c
```diff
@@ -1245,144 +1245,80 @@ static void do_verify(void)
...
-            deleted_mbentry =
-                hash_del(mbentry->jmapid, &frock->deleted_mbentry_by_jmapid);
-            if (deleted_mbentry) {
-                ptrarray_add(&frock->deleted_mbentry_to_fix, deleted_mbentry);
-            }
-        }
+        // already got a record, we're good!
+        int res = mboxlist_lookup_by_jmapid(userid, mbentry->jmapid, NULL, NULL);
+        mbname_free(&mbname);
+        if (!res) return 0;
```
wolfsage, line 1273: Does this leek byunqid?

imap/ctl_mboxlist.c
```diff
@@ -1245,144 +1245,80 @@ static void do_verify(void)
...
-    }
+        byunqid->foldermodseq = mailbox_modseq_dirty(mailbox);
+
+        buf_putc(&jmapid, JMAP_MAILBOXID_PREFIX);
+        MODSEQ_TO_JMAPID(&jmapid, byunqid->foldermodseq);
+        mailbox_set_jmapid(mailbox, buf_cstring(&jmapid));
+
+        free(byunqid->jmapid);
+        byunqid->jmapid = buf_release(&jmapid);
+        buf_free(&jmapid);
```
wolfsage, line 1293: this buf_free isn't needed is it? buf_release should have done that?

PR #5393, opened 2025-04-17
WIP: Install EV

wolfsage:


wolfsage:
I noticed AnyEvent was using the PurePerl Implementation, I added this to see if cass would be any faster with the EV Impl, but seems like not.

PR #5400, opened 2025-04-28
mboxlist_update_intermediaries(): DO NOT reuse cmodseqs

ksmurchison:


rsto (commented):
Could you please describe shortly in the commit message why that change improves things? If it is fixing a bug, what was broken?

ksmurchison:
> Could you please describe shortly in the commit message why that change improves things? If it is fixing a bug, what was broken?

Actually maybe this should just be a commit on the jmapid-bis branch since that's where things break.

ksmurchison:
Moved this commit to jmapids-bis.  Closing this PR.

PR #5403, opened 2025-04-30
jmap_contact.c: calculate hash table sizes from vCard

rsto:
This patch includes two improvements when converting vCards to JMAP. It uses an heuristic to determine the size of hash tables that are used for group label lookups, rather than hardcoding a predefined value. It also caches X-ABADR property values for translation of country codes.

This change does not change or add JMAP Contact features or elements, so no Cassandane test is added.

### DIFF
--- a/imap/jmap_contact.c
+++ b/imap/jmap_contact.c
@@ -1613,6 +1613,21 @@ static json_t *jmap_contact_from_vcard(const char *userid,
     json_t *online = json_array();
     int emailIndex = 0, defaultEmailIndex = -1;
 
+    // Generate lookup table for Apple's X-ABADR property values.
+    hash_table abadr_by_group = HASH_TABLE_INITIALIZER;
+    size_t nabadr = 0;
+    for (struct vparse_entry *it = card->properties; it; it = it->next) {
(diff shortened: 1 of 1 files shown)


ksmurchison (approved):
LGTM

imap/jmap_contact.c
```diff
@@ -1679,15 +1698,7 @@ static json_t *jmap_contact_from_vcard(const char *userid,
...
-                // O(n^2) n is the (presumably small) count of VCARD props
-                struct vparse_entry *iter;
-                for (iter = card->properties; iter; iter = iter->next) {
-                    if (!strcasecmpsafe(iter->group, entry->group) &&
-                            !strcasecmp(iter->name, VCARD_APPLE_ABADR_PROPERTY)) {
-                        countrycode = iter->v.value;
-                        break;
-                    }
-                }
+                countrycode = hash_lookup(entry->group, &abadr_by_group);
```
wolfsage, line 1701: @rsto

The previous code would compare the group names case insensitively, the new code looks for an exact match.

Is that a problem? Does the space say ITEM1 and item1 are the same or different?

PR #5408, opened 2025-05-05
Cass mailbox version

wolfsage:
This is somewhat a draft for the time being


wolfsage:
This is now included in https://github.com/cyrusimap/cyrus-imapd/pull/5392/, clsoing

PR #5409, opened 2025-05-05
slowio: Use CLOCK_PROCESS_CPUTIME_ID instead of CLOCK_MONOTONIC

wolfsage:
On systems with slower disks or heavy load, the cyrus processes may not get much CPU time. This can skew the slowio logic when it's trying to determine the rate of speed at which data is being read or written.

For example, if the process has had only .0008 seconds of CPU time, but it's been waiting on disk IO for .004s, and it's processed 16384 bytes, it will calculate:

    16384 / .004 = 4,096,000 bytes/sec

...but the threshold to delay is 6,250,000 bytes/sec.

The CPU time would have caused a delay since:

    16384 / .0008 = 20,480,000 bytes/sec.

This seems reliable to "fix" the tests on my boxes which have slow disks,


cunit/slowio.testc
```diff
@@ -112,6 +112,25 @@ static void test_limit2(void)
     }
 }
 
+void busywait_for(struct timespec *howlong)
```
elliefm, line 115: Can you make this function static, it doesn't need to be visible to other modules.  Like this:

```suggestion
static void busywait_for(struct timespec *howlong)
```

cunit/slowio.testc
```diff
@@ -112,6 +112,25 @@ static void test_limit2(void)
...
+
+    clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &start);
+
+    for (long long int i = 0; i < max_tries; i++) {
+      struct timespec now;
+      clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &now);
+
+      if (now.tv_sec - start.tv_sec > howlong->tv_sec &&
+          now.tv_nsec - start.tv_nsec > howlong->tv_nsec
+      ) {
```
elliefm, line 128: This smells like it might have weird edge cases.  I don't _see_ any, I just remember that timespec comparisons are tricky.  The safe thing to do would be to convert `now`, `start`, and `howlong` to doubles, and then compare those.  This is how `slowio_maybe_delay_impl()` already does it, so look there for an example.

We do have some utility functions along those lines, except they're for timeval, not timespec, which is annoying.

Style nit:  I like operators at wrapping points to be at the start of the next line, not the end of the previous.  It puts the (IMO) most important detail -- how do these lines relate to one another -- on the left hand side where you see it first.  Our WIP clang-format rules should format this like this:

```suggestion
      if (now.tv_sec - start.tv_sec > howlong->tv_sec
          && now.tv_nsec - start.tv_nsec > howlong->tv_nsec)
      {
```

cunit/slowio.testc
```diff
@@ -112,6 +112,25 @@ static void test_limit2(void)
     }
 }
 
+void busywait_for(struct timespec *howlong)
+{
+    struct timespec start;
+    long long int max_tries = LLONG_MAX;
```
elliefm, line 118: I prefer to use stdint.h types these days for general integers that are more complicated than just `int` or `unsigned` (you might need to `#include <stdint.h>` if it's not already):

```suggestion
    int64_t max_tries = INT64_MAX;
```

If the intention was not specifically "64-bit integer", but more like "the biggest integer the platform has", look at `intmax_t`.  It's usually 64 bits too, but the different name communicates that we don't care about the exact number of bits so long as it's very big.

elliefm (changes requested):
This is really interesting, good find.  I think I was using CLOCK_MONOTONIC because last time I looked closely at clock_gettime, it sounded like CLOCK_REAL and CLOCK_MONOTONIC were the only portable choices.

Reading again today, it sounds like CLOCK_PROCESS_CPUTIME_ID is actually specified by POSIX, so it could be fine to use as is.  Looks like there's a `_POSIX_CPUTIME` symbol that will be defined if it's available, so if some BSD or whatever users complain about it, we could add a configure check to choose which to use.  But I think it's fine to just use for now, and we can stuff around with a configure check later if it actually comes up.

I've flagged a couple of minor style nits, but the only detail I'm actually concerned about is the timespec comparison in the busywait_for.

cunit/slowio.testc
elliefm: Come to think of it, once busywait_for is using doubles for the difference and comparison, `howlong` itself might as well just be a double.  The place busywait_for is called from was using a struct timespec because that was needed by nanosleep, but since we're not using nanosleep, we should use something sane, and keep the timespec mess encapsulated.

wolfsage:
@elliefm - I think I've addressed the feedback. Take another look? Thanks

elliefm (approved):
Yeah nice, thanks

cunit/slowio.testc
```diff
@@ -112,6 +113,26 @@ static void test_limit2(void)
...
+        clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &now);
+
+        double elapsed = (double)(now.tv_sec - start.tv_sec);
+        elapsed += (double)(now.tv_nsec - start.tv_nsec) / 1000000000.0;
+
+        if (elapsed > howlong) {
+            break;
+        }
+    }
+}
```
elliefm, line 134: The less code there is, the more I'll overthink it...

Something like this might get us portability to hypothetical no-CPUTIME platforms without too much mess.  Keep the new busywait_for API, but if we don't have CPUTIME, just use nanosleep after all.  Still not convinced this is worth doing until we have evidence that such a platform exists, though.

```suggestion
static void busywait_for(double howlong)
{
#ifdef _POSIX_CPUTIME
    struct timespec start;
    int64_t max_tries = INT64_MAX;

    clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &start);

    for (int64_t i = 0; i < max_tries; i++) {
        struct timespec now;
        clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &now);

        double elapsed = (double)(now.tv_sec - start.tv_sec);
        elapsed += (double)(now.tv_nsec - start.tv_nsec) / 1000000000.0;

        if (elapsed > howlong) {
            break;
        }
    }
#else
    struct timespec delay;
    delay.tv_sec = floor(howlong);
    delay.tv_nsec = (howlong - delay.tv_sec) * 1000000000;
    nanosleep(&delay, NULL);
#endif
}
```

elliefm: I don't know why it's formatting the suggestion as an entire replacement of the function instead of just inserting a few lines 🤷‍♀️

wolfsage:
I set up NixOS on a VM and I can reproduce this failure.

After some digging I tracked (the first part of the problem) down to clock_gettime with CLOCK_PROCESS_CPUTIME_ID returning identical values occasionally when called multiple times. This is.. highly surprising to me.

This program shows the issue on NixOS:
```
#include <stdio.h>
#include <time.h>
#include <stdlib.h>

int main(void) {
  struct timespec last = { 0 };
  long int c = 0;
  long int was_same = 0;

  while (1) {
    struct timespec now;
    if (clock_gettime(CLOCK_PROCESS_CPUTIME_ID, &now) == -1) {
      perror("clock_gettime");
      exit(1);
    }

    if (now.tv_sec = last.tv_sec && now.tv_nsec == last.tv_nsec) {
      was_same++;
      continue;
    }

    if (was_same) {
      printf("repeated %ld times\n", was_same);
      was_same = 0;
    }

    last = now;
    c++;
  }

  return 0;
}
```

When run on Ubuntu or Debian, I get no output, but on NixOS, I *constantly* get output that clock_gettime returned an identical value for 1-10 times or more.

This seems very strange to me since it's measuring CPU time, and the program is definitely using CPU time to loop around and check the clock again.

Are nanoseconds too small of a precision on NixOS somehow because the operations are happening too fast?? Is there some bug in the newer linux Kernel, or in NixOS?

I don't know...

We may want to roll back this fix until we find something better. @elliefm / @rjbs thoughts?

wolfsage:
Okay this also behaves strangely on CentOS... think we should roll it back perhaps

wolfsage:
I'd still like to fix the underlying issue :)

wolfsage:
I've reverted it for now in https://github.com/cyrusimap/cyrus-imapd/pull/5470

Cheers

elliefm:
[I left [this comment](https://github.com/cyrusimap/cyrus-imapd/pull/5470#issuecomment-2960842118) on the PR for the revert before I saw the discussion already happening here, so here it is again.]

It's weird that returning the same value multiple times leads to it breaking?  Eyeballing it, it looks like it would only be a problem if it returned the same value INT64_MAX times in a row, which seems bonkers, that's a big number.  Are the affected platforms building with optimisations on, and the loop is getting compiled out entirely or something?

I wonder if it would be enough to add something like this inside the loop to force a real(ish) delay:

```
// need to initialise prev=start before entering the main loop
    clock_gettime(..., &now);
    while (now == prev) {
        nanosleep((struct timespec){ 0, 1 }, NULL); // 1ns
        clock_gettime(..., &now);
    }
    prev = now;

    // do the elapsed calculations with start and now as usual
```

nanosleep(3) says:
> If the interval specified in req is not an exact multiple of the granularity underlying clock (see time(7)), then the interval will be rounded up to  the next multiple. Furthermore,  after  the sleep completes, there may still be a delay before the CPU becomes free to once again execute the calling thread.

So a 1ns sleep should be "whatever is the finest non-zero granularity available".  Even though the time spent asleep won't increase the cputime, I think nanosleep's own overhead might cause cputime to tick at least once by the time it returns?  But if it not, it'll spin in the inner nanosleep loop until it does.

It's not clear to me whether CLOCK_PROCESS_CPUTIME_ID counts user cputime or system cputime or both.  I had assumed both, but if it's actually only user time, then the nanosleep overhead won't be counted either.  Maybe we need to use getrusage(2) instead of clock_gettime, and count both the `ru_utime` and `ru_stime` towards `elapsed`.

I dunno it's strange.  I wonder what's really going on.

elliefm:
#5471 brings this back but with (hopefully) fixes for the two bugs...

PR #5413, opened 2025-05-06
Sync rename lock order

brong:
This fixes a couple issues seen in Fastmail production with renames

One issue is rename A -> C -> B; depending which user is the initiating userid in sync_do_user, it locks that one first; leading to lock inversion issues.

Also, the tombstone check can wind up adding same folder, same partition to the rename list, which the replica will reject!

### DIFF
--- a/cassandane/tiny-tests/Replication/rename_old_synccache
+++ b/cassandane/tiny-tests/Replication/rename_old_synccache
@@ -1,0 +1,82 @@
+#!perl
+use Cassandane::Tiny;
+
+#
+# Test replication of mailbox only after a rename
+#
+sub test_rename_old_synccache
+    :NoAltNameSpace :AllowMoves :Replication :SyncLog :DelayedDelete :SyncCache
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):
Looks sane to me, and if it fixes our problem, even better

cassandane/tiny-tests/Replication/rename_old_synccache
```diff
@@ -0,0 +1,82 @@
...
+    $admintalk = $self->{adminstore}->get_client();
+    $admintalk->rename("user.dest", "user.third");
+
+    # replication works again
+    $self->run_replication(rolling => 1, inputfile => $synclogfname);
+    unlink($synclogfname);
+    $self->check_replication('third');
+
+    # rename back sync-cache having missed the renames
+    rename("$cachefname.junk", $cachefname);
```
rjbs, line 65: Nice / gross. :-)

PR #5415, opened 2025-05-07
cassandane: use master_pid_file from config so tests can override it

elliefm:
The master pid file is configurable in imapd.conf, but Cassandane currently forces it to a hardcoded value via the '-p' argument.  This PR updates Cassandane to set and use the value from imapd.conf.  In practice, nothing really changes here, except that now tests will be able to override it if they need to.

I expect to make use of this soon, but I'd like to get it through CI and landed before that.

### DIFF
--- a/cassandane/Cassandane/Config.pm
+++ b/cassandane/Cassandane/Config.pm
@@ -103,6 +103,7 @@ sub default
             configdirectory => '@basedir@/conf',
             syslog_prefix => '@name@',
             sievedir => '@basedir@/conf/sieve',
+            master_pid_file => '@basedir@/run/master.pid',
             master_ready_file => '@basedir@/master.ready',
             defaultpartition => 'default',
             defaultdomain => 'defdomain',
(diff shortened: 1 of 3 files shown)


rsto (approved):

ksmurchison (approved):

wolfsage (approved):
Cool. (Seems like master.ready should be in /run? But that's not necessary for this)

elliefm:
@wolfsage Good eye, but it's more subtle than that. Cassandane uses the run subdirectory specifically for storing pid files of processes it manages (which includes master, but excludes everything master manages itself).  During shutdown, I'll loop over every file in that directory, expecting to read a pid from it, which it will then kill (see the loop around `_list_pid_files` in `Instance::stop()`).  The ready file isn't a pid file, nor does it belong to Cassandane, so it's separate.

elliefm:
master_pid_file was new in 3.10 so this can't be backported to 3.8

PR #5416, opened 2025-05-07
imapd.c: Add support for STATUS APPENDLIMIT (RFC 7889)

ksmurchison:
Also, sorts STATUS items by RFC number

The omission was noted by Daniel Eggert @ Apple.

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -9531,26 +9531,29 @@ static int parse_statusitems(unsigned *statusitemsp, const char **errstr)
         else if (!strcmp(arg.s, "uidvalidity")) {
             statusitems |= STATUS_UIDVALIDITY;
         }
-        else if (!strcmp(arg.s, "mailboxid")) {
-            statusitems |= STATUS_MAILBOXID;
-        }
         else if (!strcmp(arg.s, "unseen")) {
             statusitems |= STATUS_UNSEEN;
         }
-        else if (!strcmp(arg.s, "size")) {
-            statusitems |= STATUS_SIZE;
-        }
-        else if (!strcmp(arg.s, "highestmodseq")) {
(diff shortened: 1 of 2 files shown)


imap/imapd.c
```diff
@@ -13933,9 +13940,9 @@ static int list_data_remote(struct backend *be, char *tag,
                 /* print status items */
                 const char *status_items[] = {
                     /* XXX  MUST be in same order as STATUS_* bitmask */
-                    "messages", "recent", "uidnext", "uidvalidity",
-                    "unseen", "highestmodseq", "size", "mailboxid",
-                    "deleted", "deleted-storage", "",
+                    "messages", "recent", "uidnext", "uidvalidity", "unseen",
+                    "highestmodseq", "appendlimit", "size", "mailboxid",
+                    "deleted", "deleted-storage", "", "", "",
```
brong, line 13945: what's this "empty string" stuff?  It's going to cause:

```
                    prot_printf(be->out, "%c%s", c, status_items[j]);
                    c = ' ';
```

Which will mean if any of these appear there'll be multiple spaces in the result, which probably breaks parsing.

brong (approved):
This looks fine, the "" in status_items was already there; but seems bogus

imap/imapd.c
ksmurchison: @brong The strings need to match the STATUS_* bits and we put the non-standard ones as the last bits, have 3 open bits (11-13).  See imapd.h

ksmurchison: I confirmed that the placeholders are required when trying to proxy 
`LIST "" % RETURN (STATUS (CREATEDMODSEQ))`

I added a comment next to the placeholders

PR #5417, opened 2025-05-07
JMAPCore tests: assert that all mail accountCapabilities are present

rjbs:


rjbs:
I've added a failing test, which can be made better when we have implemented more of this.

I also added a `assert_cmp_deeply` because Test::Deep is great, and we should have it.

ksmurchison:
Superseded by #5422

PR #5414, opened 2025-05-06
Remove support for OpenSSL < 3.0

ksmurchison:

PR #5418, opened 2025-05-08
imapd.conf: Deprecate the 'tls_server_dhparam' option.

ksmurchison:
This is no longer needed by OpenSSL >= 3.0.

As noted by Dilyan Palauzov

### DIFF
--- a/lib/imapoptions
+++ b/lib/imapoptions
@@ -3174,9 +3174,8 @@ product version in the capabilities
    Two certificates can be set, e.g RSA and EC, if the filenames are separated with
    comma without spaces. */
 
-{ "tls_server_dhparam", NULL, STRING, "3.1.4" }
-/* File containing the DH parameters belonging to the certificate in
-   tls_server_cert. Used by OpenSSL before version 3.0. */
+{ "tls_server_dhparam", NULL, STRING, "UNRELEASED", "UNRELEASED" }
+/* Deprecated. No longer used. */
 
 { "tls_server_key", NULL, STRING, "3.1.8" }
 /* File containing the private key belonging to the certificate in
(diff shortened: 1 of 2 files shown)


lib/imapoptions
```diff
@@ -3174,9 +3174,8 @@ product version in the capabilities
    Two certificates can be set, e.g RSA and EC, if the filenames are separated with
    comma without spaces. */
 
-{ "tls_server_dhparam", NULL, STRING, "3.1.4" }
-/* File containing the DH parameters belonging to the certificate in
-   tls_server_cert. Used by OpenSSL before version 3.0. */
+{ "tls_server_dhparam", NULL, STRING, "3.1.4", "3.12.0" }
+/* Deprecated. No longer used. */
```
elliefm, line 3178: This needs to be:

```suggestion
{ "tls_server_dhparam", NULL, STRING, "UNRELEASED", "UNRELEASED" }
/* Deprecated. No longer used. */
```

* use "UNRELEASED" for the version numbers because this change isn't in any release yet.  Updating these fields with a real number is part of the release process.
* deprecating the option counts as a modification, so the last_modified field needs to be bumped at the same time as the deprecated_since field is added.  This is important because `cyr_info -s <old_version>` lists options that have changed since _old_version_, which we tell people to do during upgrades to figure out what they want to change in their imapd.conf

elliefm (changes requested):

elliefm:
This follows from #5414

lib/imapoptions
ksmurchison: Ahh!  Thanks. Wasn't aware. Will fix in the morning. Otherwise is this good to go?

elliefm: Yep

rsto (approved):
LGTM, as long as elliefm's review comments are resolved

lib/imapoptions
ksmurchison: Fixed

PR #5421, opened 2025-05-08
imapd.c: allow setting mailbox version with CREATE param

ksmurchison:
- If creating a v19 (or earlier) INBOX, a v1 conv.db will be created
- CREATEs w/o an explicit version will use the version of the parent

This was cherry-picked (and modified) from 64bit-nanosecond-jmapids-bis branch

### DIFF
--- a/cassandane/Cassandane/Cyrus/Create.pm
+++ b/cassandane/Cassandane/Cyrus/Create.pm
@@ -45,7 +45,9 @@ use Data::Dumper;
 use lib '.';
 use base qw(Cassandane::Cyrus::TestCase);
 use Cassandane::Util::Log;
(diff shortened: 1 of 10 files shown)


rsto (approved):
LGTM - but is there no need to add this in jmap_mailbox.c for renames, too?

ksmurchison:
> LGTM - but is there no need to add this in jmap_mailbox.c for renames, too?

With mailboxes-by-uuid most renames are simply a change to mbentry and not a creation of a new mailbox.  In the case where we do have to create a new mailbox, mailbox_rename_copy() uses the version of the source mailbox as the version for creating the new mailbox.

PR #5422, opened 2025-05-08
jmap_mailbox.c: add maxMailboxDepth & maxSizeMailboxName to accountCapabilities

ksmurchison:
- Mailbox-specfic limits moved from jmap_mail.c to jmap_mailbox.c
- maxSizeMailboxName enforced in _mboxset_args_parse()
- emailsListSortOptions renamed to emailQuerySortOptions

### DIFF
--- a/cassandane/Cassandane/Unit/TestCase.pm
+++ b/cassandane/Cassandane/Unit/TestCase.pm
@@ -523,6 +523,27 @@ sub assert_not_file_test
                   "'$path' unexpectedly passed '$test_type' test");
 }
 
+sub assert_cmp_deeply
+{
+    my ($self, $have, $want, $desc) = @_;
+    $desc ||= "deep comparison matched";
+
+    require Test::Deep;
+    my ($ok, $stack) = Test::Deep::cmp_details($want, $have);
+
+    if ($ok) {
+        return $self->assert(1, $desc);
+    }
+
+    my ($package, $filename, $line) = caller;
+
+    my $diag = join qq{\n},
+               "deep comparison failed at $filename, line $line:\n",
+               Test::Deep::deep_diag($stack);
+
+    return $self->assert(0, $diag);
+}
+
 sub new_test_url
 {
     my ($self, $content_or_app) = @_;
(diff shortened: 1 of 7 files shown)


cassandane/Cassandane/Unit/TestCase.pm
```diff
@@ -523,6 +523,27 @@ sub assert_not_file_test
                   "'$path' unexpectedly passed '$test_type' test");
 }
 
+sub assert_cmp_deeply
+{
+    my ($self, $have, $want, $desc) = @_;
+    $desc ||= "deep comparison matched";
+
+    require Test::Deep;
```
rjbs, line 531: This effectively make Test::Deep a prereq of Cass, but I think this is okay.  We have basically agreed that JMAP::Tester is a prereq, and it requires it.  So, I think nothing needs changing, just calling it out.

rjbs (approved):
Putting aside any subtle C issues, looks good.

imap/jmap_mailbox.c
```diff
@@ -152,9 +152,17 @@ HIDDEN void jmap_mailbox_init(jmap_settings_t *settings)
...
-                                      __attribute__((unused)))
+HIDDEN void jmap_mailbox_capabilities(json_t *account_capabilities,
+                                      int mayCreateTopLevel)
 {
+    json_t *email_capa =
+        json_object_get(account_capabilities, JMAP_URN_MAIL);
+
+    json_object_set_new(email_capa, "maxMailboxDepth", json_null());
+    json_object_set_new(email_capa, "maxSizeMailboxName",
+                        json_integer(MAX_MAILBOX_NAME / 2));
```
rsto, line 163: Maybe it's too early morning for me, but why do we only allow half of MAX_MAILBOX_NAME?

rsto (approved):

imap/jmap_mailbox.c
ksmurchison: So MAX_MAILBOX_NAME is the limit on the entire mailbox name (e.g. "user.userid.foo.bar"), not just each individual name segment (e.g. "foo").  Since JMAP never exposes "user.userid", we have to set our limit less than MAX_MAILBOX_NAME.  And Rik did some research and founf that we have at least one user with a mailhbox name segment of 247 chars.  So we chose MAX_MAILBOX_NAME/2 to accommodate this nonsense.

rsto: I figured it must be something like that. WFM.

PR #5429, opened 2025-05-12
jmap_mail.c: move maxKeywordsPerEmail into extension capabilities

ksmurchison:


rsto (changes requested):
CI fails, presumably because some tests use the wrong JMAP request `using` argument.

rsto (approved):

PR #5432, opened 2025-05-13
Cassandane: silence TLS session cache leaks detected by valgrind

ksmurchison:
With this in place we can remove the :SuppressLSAN() magic from:

ALPN.pm
Alpaca.pm
MboxEvent.pm
StartTLS.pm
JMAPCore/echo_tls

### DIFF
--- a/cassandane/Cassandane/Cyrus/ALPN.pm
+++ b/cassandane/Cassandane/Cyrus/ALPN.pm
@@ -64,6 +64,7 @@ sub new
     my $config = Cassandane::Config->default()->clone();
     $config->set(tls_server_cert => '@basedir@/conf/certs/cert.pem',
                  tls_server_key => '@basedir@/conf/certs/key.pem',
+                 tls_session_timeout => 0,
                  allowstarttls => 'on',
                  httpmodules => 'caldav');
 
--- a/cassandane/Cassandane/Cyrus/TestCase.pm
+++ b/cassandane/Cassandane/Cyrus/TestCase.pm
@@ -424,6 +424,7 @@ magic(TLS => sub {
     my $self = shift;
     $self->config_set(tls_server_cert => '@basedir@/conf/certs/cert.pem');
     $self->config_set(tls_server_key => '@basedir@/conf/certs/key.pem');
+    $self->config_set(tls_session_timeout => 0);
     $self->want('install_certificates');
     $self->want_services('imaps');
 });


elliefm:
Oh, interesting!

Could you add a commit that removes those suppressions please?  Perhaps just temporarily (see what @wolfsage thinks).  It would be good to see if CI passes with the suppressions gone.

ksmurchison:
> Oh, interesting!
> 
> Could you add a commit that removes those suppressions please? Perhaps just temporarily (see what @wolfsage thinks). It would be good to see if CI passes with the suppressions gone.

Sure.  I will remove the suppressions in this PR tomorrow and see if CI is happy.

wolfsage:
So the problem is... with session caching enabled, at shutdown time, the cached values aren't being cleaned up, so we get a leak. Is that accurate?

If that's the case, this doesn't feel like a fix, but avoiding the problem in testing only? Or have I misunderstood.

I was hoping https://github.com/cyrusimap/cyrus-imapd/pull/5433/files was the real fix for these

ksmurchison:
> So the problem is... with session caching enabled, at shutdown time, the cached values aren't being cleaned up, so we get a leak. Is that accurate?

Yes.


> If that's the case, this doesn't feel like a fix, but avoiding the problem in testing only? Or have I misunderstood.

It's avoiding the problem by disabling the problematic code thereby allowing us to remove an ASAN exception/override. 

I tried fixing this once before and had a PR that preventing valgrind from complaining but it caused issues with other deployments, the nature of which I don't remember. 

> I was hoping https://github.com/cyrusimap/cyrus-imapd/pull/5433/files was the real fix for these 

Nope. That fixes another leak which gets lost in the TLS session cache leak noise.

wolfsage:
Okay so to summarize this is hiding a problem that real world default configuration exposes. The SuppressLSAN tells us we know about the bug, but this change would hide that fact. I think we should either...

 - Leave the SuppressLSAN in place until the problem is fixed
 - Merge this, but update the commit message with more detail and add a TODO or XXX where we're setting the config param to explain the bug and that it needs fixing
 -  Merge this, but update the commit message with more detail and create a ticket to track the bug so we don't forget about it

Thoughts? Removing SuppressLSAN is certainly appealing incase it's hiding some other bugs that crop up, so I support that

ksmurchison:
I think I agree that we should leave the SuppressLSAN and find a way to fix the actual leak.

I have found that the SuppressLSAN is masking other leaks, but I'm working on those.

rsto (commented):
I agree with Ken to find a way to fix the actual leak.

ksmurchison:
Superseded by #5437

PR #5434, opened 2025-05-14
twom: fix stringop-truncation warning with -O3

elliefm:
When x has no key, `KEYLEN(x)` returns 0, but `KEYPTR(x)` returns `""` rather than NULL.  This means the `if (key)` check is always true, and we might call strncpy with len=0. The optimiser notices this and complains.  Shush it by checking len too.

We're just producing (maybe-truncated) human readable output here, so a possibly-empty key isn't concerning.

### DIFF
--- a/lib/twom.c
+++ b/lib/twom.c
@@ -2988,7 +2988,7 @@ static int twom_txn_dump(struct twom_txn *txn, int detail)
             const char *key = KEYPTR(loc->file->base + parent_offset);
             size_t len = KEYLEN(loc->file->base + parent_offset);
             if (len > 79) len = 79;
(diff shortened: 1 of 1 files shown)


elliefm:
@wolfsage mentioned the issue and workaround to me a few weeks ago.  I finally got around to staring closely at it for a while, and have decided it's actually a good fix.

brong (approved):

brong:
Yeah, clearly good - I'm just gonna merge this

wolfsage:
Awesome, unless something else has snuck in this will let me move forward with adding a few more CI make check phases with a few different compilation options to catch more problems!

PR #5437, opened 2025-05-15
tls.c: don't leak SSL_SESSIONs when caching

ksmurchison:
Let's get this on future to make sure that it doesn't break anything before we actually merge.


wolfsage:
Wild! I think I understand - but could you update the commit message to describe the problem and how this fixes it?

Detailed commit messages are good!

ksmurchison:
> Wild! I think I understand - but could you update the commit message to describe the problem and how this fixes it?
> 
> Detailed commit messages are good!

Done.

elliefm (approved):

PR #5440, opened 2025-05-16
Email/query: Always track collapsed matches so query cache is consistent

wolfsage:
When you "Email/query" with collapseThreads = true, we store in the cache info necessary for both collapsed and uncollapsed results.

When you "Email/query" with collapseThreads = false, we were not storing info necessary for collapsed threads results, so subsequent requests that hit the cache but have collapseThreads = true would return 0 items.

Fix that by always computing both


rsto (approved):
On second thought, I think this is the better fix, as it has a much smaller chance of introducing new bugs.

PR #5442, opened 2025-05-19
configure.ac: automatically use libcap if it's available

elliefm:
This switches Cyrus from "use libcap if it's requested" to "use libcap if it's available, unless told not to".  I believe libcap is Linux-only, so we can't mandate it for all builds.

The interesting thing that libcap gets us, according to https://github.com/cyrusimap/cyrus-imapd/commit/2e9a36cd4, is the ability for master to drop privileges while retaining the ability to bind ports.  I think this means new services  or dead services can be restarted on a running Cyrus with just a SIGHUP to master, rather than needing a whole restart.

But the actual ramifications are unknown to me; I don't believe any of our tests exercise anything that would be different with or without it.  Cassandane exclusively uses high port numbers and never starts master as root, for example.

**Fastmail:** Do our builds currently configure with `--with-libcap`?  If so, then that means this feature must work fine, because we're already using it.  If not, this PR will probably make us start using it, unless we add `--without-libcap` or are missing the `libcap-dev` package.

Marking this "Do Not Merge" until we understand the ramifications a bit better...


ksmurchison (approved):
Looks sane

rsto (approved):

rjbs:
@elliefm We don't yet compile with libcap, but let's start doing so.  We can shepherd this into production on the usual ladder, no special work from you other than telling me when it's (a) merged or (b) ready for inclusion.

PR #5447, opened 2025-05-22
http_caldav.c: add support for bulk export (GET) noDeclined query param

ksmurchison:
If supplied, this param will cause events DECLINED by the user to be ommitted from the export

### DIFF
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -1672,6 +1672,8 @@ static int export_calendar(struct transaction_t *txn)
     modseq_t syncmodseq = 0;
     int unbind_flag = -1, unchanged_flag = -1;
     struct caldav_db *caldavdb = NULL;
+    bool no_declined = !!hash_lookup("noDeclined", &txn->req_qparams);
+    strarray_t schedule_addresses = STRARRAY_INITIALIZER;
 
     /* Check requested MIME type:
        1st entry in caldav_mime_types array MUST be default MIME type */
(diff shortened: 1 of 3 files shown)


imap/http_caldav.c
```diff
@@ -1672,6 +1672,8 @@ static int export_calendar(struct transaction_t *txn)
     modseq_t syncmodseq = 0;
     int unbind_flag = -1, unchanged_flag = -1;
     struct caldav_db *caldavdb = NULL;
+    unsigned no_declined = !!hash_lookup("noDeclined", &txn->req_qparams);
```
rsto, line 1675: We agreed to set C11 as the base line, so we can just as well use type `bool` here.

rsto (approved):
LGTM - one nit

dilyanpalauzov:
There is no documentation for this parameter.

imap/http_caldav.c
```diff
@@ -1672,6 +1672,8 @@ static int export_calendar(struct transaction_t *txn)
     modseq_t syncmodseq = 0;
     int unbind_flag = -1, unchanged_flag = -1;
     struct caldav_db *caldavdb = NULL;
+    bool no_declined = !!hash_lookup("noDeclined", &txn->req_qparams);
```
wolfsage, line 1675: nit: we support 'managed-id', so maybe this should be 'no-declined' for consistency or just 'declined' with 'declined=0' disabling including declined items, but I don't feel too strongly about it

wolfsage (approved):

ksmurchison:
@rjbs Any thoughts on this?

rjbs:
@ksmurchison Nope.  How about we label it `include-in-fastmail`, get it in a build, reimplement the export in terms of this, and make sure it works?

imap/http_caldav.c
ksmurchison: managed-id is all lowercase because that's how its defined in RFC 8607.  I have no strong preference for what we name the new thing.

ksmurchison:
> @ksmurchison Nope. How about we label it `include-in-fastmail`, get it in a build, reimplement the export in terms of this, and make sure it works?

@rjbs do you care what we call the new query param, or what case we use (I don't)?

rjbs:
@ksmurchison No, I have no feelings about the name other than "we might think of a better one later if we fiddle with it". :)

PR #5448, opened 2025-05-22
httpd.c: free txn and ws contexts when receiving shutdown signal

ksmurchison:


elliefm:
I believe at the moment our CI tests don't build with nghttp2, which means the Cassandane ASAN run won't detect h2 leaks because the h2 tests get skipped.  If you haven't already done so, might be worth running ALPN.https_h2 and ALPN.https_multi2 under valgrind, in case there's any leaks still hiding in the h2 shutdown code.

elliefm:
Weird, I just pulled the branch and tried running the ALPN suite locally under valgrind, and almost all of the tests are failing with a leak from tls_start_servertls:1160.  Didn't we fix that?  The only ones that don't leak are the foo_bad ones, where the initial TLS handshake fails and so no TLS connection is ever established.

Just tried again with the master branch, and got the same failures, so I guess we didn't fix this yet.  We don't have a suppression for it in vg.supp though, so I thought it must be fixed.  Not sure what's going on.

rsto (approved):
Can't say anything about the valgrind suppressions, but cleaning up txn before shutdown looks good to me.

ksmurchison:
> Weird, I just pulled the branch and tried running the ALPN suite locally under valgrind, and almost all of the tests are failing with a leak from tls_start_servertls:1160. Didn't we fix that? The only ones that don't leak are the foo_bad ones, where the initial TLS handshake fails and so no TLS connection is ever established.
> 
> [time passes]
> 
> Just tried again with the master branch, and got the same failures, so I guess we didn't fix this yet. We don't have a suppression for it in vg.supp though, so I thought it must be fixed. Not sure what's going on.

We do have a fix $5437

I haven't committed it yet because I wanted to test it further and make sure that it doesn't break anything like I did with the last "fix".  But having said that, I don't think we ever have clients connecting to Cyrus using HTTPS.

elliefm:
Oh right, that's why I thought we had a fix already, okay!

elliefm (approved):
I pulled this locally again, and also cherry picked the TLS leak fix from #5437, and now all the ALPN tests pass under valgrind.  Thanks!

Issue #5450, opened 2025-05-24
Fail to build with GCC 15


guimard:
Hi,

following [Debian #1096496](https://bugs.debian.org/1096496), I can confirm that we have the same bug with last cyrust version (3.12):
> ```
> gcc -DHAVE_CONFIG_H -I.   -I. -I./lib -I. -I./lib -DLIBEXEC_DIR=\"/usr/lib/cyrus/bin\" -DSBIN_DIR=\"/usr/lib/cyrus/bin\" -DSYSCONF_DIR=\"etc\" -DHAVE_CONFIG_H      -I/usr/include/libxml2               -I/usr/include/postgresql -Wdate-time -D_FORTIFY_SOURCE=2 -I/usr/include -fPIC  -g -O2 -Werror=implicit-function-declaration -ffile-prefix-map=/build/reproducible-path/cyrus-imapd-3.10.1=. -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -Wall -Wextra -g -fno-strict-aliasing -pipe -O2 -DPCRE2POSIX_SHARED  -c -o imap/defaultalarms.o imap/defaultalarms.c
> imap/zoneinfo_db.c:308:14: error: conflicting types for ‘zoneinfo_find’; have ‘int(const char *, int,  time_t,  int (*)(void), void *)’ {aka ‘int(const char *, int,  long int,  int (*)(void), void *)’}
>   308 | EXPORTED int zoneinfo_find(const char *find, int tzid_only, time_t changedsince,
>       |              ^~~~~~~~~~~~~
> In file included from imap/zoneinfo_db.c:64:
> imap/zoneinfo_db.h:99:12: note: previous declaration of ‘zoneinfo_find’ with type ‘int(const char *, int,  time_t,  int (*)(const char *, int,  struct zoneinfo *, void *), void *)’ {aka ‘int(const char *, int,  long int,  int (*)(const char *, int,  struct zoneinfo *, void *), void *)’}
>    99 | extern int zoneinfo_find(const char *find, int tzid_only, time_t changedsince,
>       |            ^~~~~~~~~~~~~
> imap/zoneinfo_db.c: In function ‘zoneinfo_find’:
> imap/zoneinfo_db.c:321:16: error: assignment to ‘int (*)(const char *, int,  struct zoneinfo *, void *)’ from incompatible pointer type ‘int (*)(void)’ [-Wincompatible-pointer-types]
[...]

PR #5451, opened 2025-05-24
Fix zoneinfo_db code for gcc 15

guimard:
Fixes: #5450


dilyanpalauzov:
This solves a compiler error with GCC 15, which error appears in Cyrus 2.5 and all later versions.

PR #5455, opened 2025-05-29
Cyrus indexfile tests

wolfsage:


ksmurchison (approved):

wolfsage:
This works with v19 indexes. v20 indexes will be different and need adjustments. I'll likely update this test to work with the 64bit jmap tests and merge it over there instead.

PR #5456, opened 2025-05-30
docs: update release process documentation

elliefm:
The way we update the website has changed as of earlier this year, so I've updated the documentation to describe the new process.  I've also updated a handful of other details while I was at it, so the documented process more closely reflects what I actually do.

### DIFF
--- a/docsrc/imap/developer/releasing.rst
+++ b/docsrc/imap/developer/releasing.rst
@@ -37,14 +37,14 @@ Release notes
 
 Write up the release notes, and add them to the appropriate location under
 ``docsrc/imap/download/release-notes/``.  They will not yet be linked up.
+Build the documentation locally with ``make doc-html``, then check the new
+release notes file in your brower to make sure it's correct and complete.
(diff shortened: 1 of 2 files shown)


docsrc/imap/developer/releasing.rst
```diff
@@ -132,20 +133,25 @@ The website is built from an amalgamation of documentation from:
...
+branch but it's not hard to figure out how to massage it.
+
+For cyrus-sasl, manually update its docsrc/conf.py with the correct cyrus-imapd
+version numbers, commit it with a similar message, and push it.  If in doubt,
+have a look at previous commits that modified this file for inspiration.  If
+you don't have push access to the cyrus-sasl repository, check with someone who
+does.
 
-This step often gets forgotten, so if you actually follow it, and notice
+This step sometimes gets forgotten, so if you actually follow it, and notice
```
rjbs, line 154: This change made me chuckle. :)

rjbs (approved):
This all makes sense to me.  Some of this is fallout from things I did, so thanks for dealing with that!

PR #5458, opened 2025-06-02
mailbox.c: read/write index files based on a version-specific template

ksmurchison:
This eliminates testing multiple conditionals for each read/write

Note that this is actually a single-commit PR against my 64bit-nanosecond-jmapids-bis branch.  Only look at the last commit.

### DIFF
--- a/imap/index_file.c
+++ b/imap/index_file.c
@@ -444,7 +444,7 @@ static const index_file_template_t v20_template =
     { 192, 144, MAILBOX_OPT_VALID, v20_hdr_fields, v20_rec_fields, crc_field };
 
 
-EXPORTED const index_file_template_t *index_files_by_version[MAILBOX_MINOR_VERSION+1] = {
+EXPORTED const index_file_template_t *const index_files_by_version[] = {
     &v00_template, // "base" header fields
     NULL,
     NULL,
@@ -467,6 +467,8 @@ EXPORTED const index_file_template_t *index_files_by_version[MAILBOX_MINOR_VERSI
     &v19_template,
     &v20_template
 };
+EXPORTED const size_t n_index_files = sizeof(index_files_by_version)
+                                      / sizeof(index_files_by_version[0]);
 
 EXPORTED const char *index_file_read_fields(const char *bufp, void *base,
                                             const index_field_t *fields)
--- a/cunit/index_file.testc
+++ b/cunit/index_file.testc
@@ -212,6 +212,11 @@ static void test_existent_mailbox_versions(void)
     }
 }
 
+static void test_no_missing_versions(void)
+{
+    CU_ASSERT_EQUAL(n_index_files, MAILBOX_MINOR_VERSION + 1);
+}
+
 #define dump_offset(o, f) do { \
     fprintf(stderr, "%s.%s: " OFF_T_FMT "\n", #o, #f, offsetof(o, f)); \
 } while(0)
--- a/imap/index_file.h
+++ b/imap/index_file.h
@@ -70,7 +70,8 @@ typedef struct index_file_template {
     const index_field_t *crc_field;
 } index_file_template_t;
 
-extern const index_file_template_t *index_files_by_version[];
+extern const index_file_template_t *const index_files_by_version[];
+extern const size_t n_index_files;
 
 extern const char *index_file_read_fields(const char *bufp, void *base,
                                           const index_field_t *fields)


rsto (commented):
I think this is a great improvement, I find this implementation much more readable than the current one. The one thing I struggled with are the trailing comment lines in lines such as
```
   { 8, 8, offsetof(struct index_header, quota_mailbox_used)      }, /*  32 */ \
```
where the trailing comment indicates the disk offsets. I suggest to either remove these comments or to indicate first what they represent.

The reason why I only comment, not approve this PR is because when I reviewed it, it did not pass CI (but it might now). More importantly, my concern is that even if it passes CI then it only validated that parsing succeeded for the latest Cyrus index header version. Is that good enough, or should this also validate legacy index header versions?

lib/util.h
```diff
@@ -94,20 +94,27 @@ extern const char CYRUS_VERSION[];
...
+#define UINT64_NANOSEC_FMT ".%.9lu"
+#elif ULLONG_MAX == BIT64_MAX
+#define BIT64_FMT          "%016llx"
+#define UINT64_FMT         "%llu"
+#define UINT64_LALIGN_FMT  "%-*llu"
+#define UINT64_NANOSEC_FMT ".%.9llu"
 #else
-#error dont know what to use for bit32
+#error dont know what to use for bit64
 #endif
```
elliefm: (@dilyanpalauzov mentioned this also)

You should be able to remove the conditionalisation here by just defining these in terms of PRIu64 et al, which are defined in `<inttypes.h>`.  Something like:

```
#define BIT64_FMT "%016" PRIx64 /* n.b. x rather than u */
#define UINT64_FMT "%" PRIu64 /* XXX this is sorta redundant though */
#define UINT64_LALIGN_FMT "%-*" PRIu64
#define UINT64_NANOSEC_FMT ".%.9" PRIu64
```

ksmurchison: Done on 64bit-nanosecond-jmapids-bis branch and then rebase this fork

lib/util.h
```diff
@@ -94,20 +95,18 @@ extern const char CYRUS_VERSION[];
...
-typedef unsigned int bit32;
-#elif ULONG_MAX == BIT32_MAX
-typedef unsigned long bit32;
-#elif USHRT_MAX == BIT32_MAX
-typedef unsigned short bit32;
-#else
-#error dont know what to use for bit32
-#endif
+#define BIT64_FMT          "%016" PRIx64
+#define UINT64_FMT         "%" PRIu64
```
dilyanpalauzov, line 103: https://github.com/cyrusimap/cyrus-imapd/pull/3748 has removed `INT_FMT` ⇔ "%lld"/ "%ld" in favour of `"%" PRId64`.

I think that using PRIu64 instead of UINT64_FMT in the code will allow people, which do not have deep understanding about the Cyrus internals, to understand the code.  For this reason I do not think that adding `UINT64_FMT` is a good idea.  Using `UINT64_FMT` instead of `"%" PRIu64` does not safe any bytes.

elliefm: But it does allow code that's using BIT64_FMT, UINT64_FMT, UINT64_LALIGN_FMT, and UINT64_NANOSEC_FMT nearby to use them all the same way, which will help avoid mistakes compared to sometimes needing the "%" but sometimes not.

It's also consistent with the rest of our custom format macros (SIZE_T_FMT, MODSEQ_FMT, etc), which embed the "%".

I don't love having a UINT64_FMT defined when the standard already provides that in PRIu64.  But in context, I don't hate it either.

ksmurchison: Agreed.  I like the consistency of having these macros.  In fact, I'd be in favor of bringing back

`#define INT64_FMT "%" PRId64`

imap/index_file.c
```diff
@@ -0,0 +1,546 @@
...
+            *((bit32 *) bufp) = htonl(*((bit32 *) datap));
+            break;
+
+        case '8':
+            if (fields->disk_size == 8)
+                align_htonll(bufp, *((bit64 *) datap));
+            else {
+                /* this field is 64b in memory but stored as 32b on disk */
+                bit32 b32 = *((bit64 *) datap);
+                *((bit32 *) bufp) = htonl(b32);
```
elliefm, line 543: Do we want to assert that we're not losing data by truncating to 32 bits?  Not sure if it matters for these fields.

Perhaps something like `assert(*(bit64 *) datap >> 32 == 0);` or `assert((bit64) b32 == *(bit64 *) datap);`

imap/index_file.c
```diff
@@ -0,0 +1,546 @@
...
+                *((bit32 *)(bufp)) = htonl(ts->tv_sec);
+            break;
+        }
+
+        case 'E':
+            break;
+
+        default:
+            // should never get here
+            assert(0);
```
elliefm: Maybe `fatal(EX_SOFTWARE, "unknown field data type");` or whatever the syntax is, rather than an assert.  The utility of assert is the conditional expression, but in this case it's unconditional, and "0" isn't a helpful message.

imap/index_file.c
```diff
@@ -0,0 +1,546 @@
...
+    off_t offset;     // offset to data type in storage struct
+};
+
+#define NULL_FIELD { 0, 0, -1 }
+
+
+/*
+ * Index file header fields by version
+ *
+ * Comment at the end of each field is the offset of the field in header buffer
```
elliefm: I think I would put these expected offsets into an actual field in the structs, rather than as comments at eol.  That opens the possibility for the code (or a unit test) to validate the computed value against the expected value, so we can detect and correct any mistakes.  Can't unit test a comment!  Perhaps something like:

```
struct index_field {
    size_t disk_size;
    char data_type;
    off_t offset;
    unsigned expected_offset;
};

static const index_field_t v00_header_fields[] = {
    { 4, '4', offsetof(struct index_header, generation_no), 0 },
    { 4, '4', offsetof(struct index_header, format),        4 },
    // and so on
```

Then later somewhere, something like:

```
        assert(fields->offset >= 0 && (unsigned) fields->offset == fields->expected_offset);
        void *datap = (void *) ((off_t) base + fields->offset);
```

I might have misunderstood exactly which offset the comments are describing, but I think the general point stands.

The main objection I'd anticipate here is the increased binary size from the extra 4 bytes in every one of these structs, but we're already wasting 7 bytes because of the char in between size_t and off_t.  If we're really that worried about it, we could reorder the struct to put the expected offset into the wasted space.  That would impede readability of the field definitions, but we could solve that by using a macro that takes its arguments in a readable order and translates them into the memory-efficient struct order.

elliefm (commented):
A few notes on index_file.c, I haven't had a chance to review the rest yet, but I need to step away

rsto (commented):
I agree with @elliefm, I found the expected offsets in the comment confusing during my first review and adding them to the struct would make them more readable and will help with error reporting.

imap/index_file.c
ksmurchison: Done

imap/index_file.c
ksmurchison: I removed the comments with the offsets.  They served no purpose other than double checking my work as I proceeded.  I did add some sanity checks to make sure that we have read/written the proper amount of data for the header and records.

imap/index_file.c
ksmurchison: Done

rjbs (commented):
I'm going to leave the review this to the rest of the gang.

imap/index_file.c
```diff
@@ -0,0 +1,580 @@
...
+};
+
+#define NULL_FIELD      { 0, 0, -1 }
+#define EMPTY_FIELD(N)  { N, 'E', 0 }
+
+
+/*
+ * Index file header fields by version
+ *
+ * Comment at the end of each field is the offset of the field in header buffer
```
wolfsage: I think this comment no longer applies?

imap/index_file.c
```diff
@@ -0,0 +1,580 @@
...
+    {  8, 'T', offsetof(struct index_header, last_repack_time)     },
+    {  4, '4', offsetof(struct index_header, recentuid)            },
+    NULL_FIELD
+};
+
+
+/*
+ * Index file record fields by version
+ *
+ * Comment at the end of each field is the offset of the field in record buffer
```
wolfsage: I think this comment also doesn't make sense anymore?

ksmurchison: Fixed

imap/index_file.c
ksmurchison: Fixed

imap/index_file.c
```diff
@@ -0,0 +1,585 @@
...
+    { 160, 112, MAILBOX_OPT_VALID, v18_hdr_fields, v16_rec_fields, crc_field };
+
+static const index_file_template_t v19_template =
+    { 160, 112, MAILBOX_OPT_VALID, v19_hdr_fields, v16_rec_fields, crc_field };
+
+static const index_file_template_t v20_template =
+    { 192, 144, MAILBOX_OPT_VALID, v20_hdr_fields, v20_rec_fields, crc_field };
+
+
+const index_file_template_t *index_files_by_version[MAILBOX_MINOR_VERSION+1] = {
```
elliefm:
```suggestion
EXPORTED const index_file_template_t *index_files_by_version[MAILBOX_MINOR_VERSION+1] = {
```

elliefm (approved):
One nit, but this is looking really good.

I think I've thought of a way to unit test those index_file arrays, to make sure the sizes/offsets are coherent without needing manual eyeball verification.  Want me to have a go at it, or try to describe it so that you can?

If we don't do it now, and merge this as is, then I'll retrofit tests later when I get a chance

imap/index_file.c
ksmurchison: Fixed.

ksmurchison:
> One nit, but this is looking really good.
> 
> I think I've thought of a way to unit test those index_file arrays, to make sure the sizes/offsets are coherent without needing manual eyeball verification. Want me to have a go at it, or try to describe it so that you can?

@elliefm If you have an idea, go ahead and run with it!

elliefm:
Okay! I'll have a bang on it today and see if I can pull something together.

imap/index_file.c
```diff
@@ -0,0 +1,585 @@
...
+    {  4, '4', offsetof(struct index_header, last_uid)             },
+    {  4, '4', offsetof(struct index_header, uidvalidity)          },
+    {  4, '4', offsetof(struct index_header, header_file_crc)      },
+    {  4, '4', offsetof(struct index_header, synccrcs.basic)       },
+    {  8, 'T', offsetof(struct index_header, recenttime)           },
+    {  8, 'T', offsetof(struct index_header, pop3_show_after)      },
+    {  4, '4', offsetof(struct index_header, synccrcs.annot)       },
+    {  4, '4', offsetof(struct index_header, unseen)               },
+    {  8, '8', offsetof(struct index_header, createdmodseq)        },
+    {  8, '4', offsetof(struct index_header, quota_deleted_used)   },
```
elliefm: Is this a typo?  disk_size = 8, but data_type is `'4'`.

The read and write functions have special behaviour for `'8'` fields that are 4 bytes on disk, but not for `'4'` fields that are 8 bytes on disk.

In v19.... above, quota_deleted_used is `8, '8', ...` like I'd expect.

imap/index_file.c
```diff
@@ -0,0 +1,585 @@
...
+
+
+/*
+ * changes_epoch is initialized to last_repack_time until v17
+ * (both values read from same buffer offset)
+ */
+#define V12_HEADER_FIELDS                                               \
+    V08_HEADER_FIELDS,                                                  \
+    {  8, '8', offsetof(struct index_header, deletedmodseq)        },   \
+    {  4, '4', offsetof(struct index_header, exists)               },   \
```
elliefm, line 126: V12_HEADER_FIELDS has deletedmodseq included twice: once here, once via V08_HEADER_FIELDS

elliefm: Oh I see, the one in V08 uses no disk bytes. Hmm.

imap/index_file.c
ksmurchison: Yep.  Fixed

elliefm:
@ksmurchison I've added tests, plus a couple of small changes to index_file to make it testable.  Could you stare closely at the tests please?  They all pass, but are they correct and complete?

ksmurchison:
I ran some timing tests on buf_to_index_header/record() and index_header/record_to_buf() on both master and on this branch and the speed difference is general +/- 1 microsec

Issue #5464, opened 2025-06-04
sivtest GSSAPI authentication issue


pzirnik:
I am using sivtest (3.8.5) to test GSSAPI authentication against dovecot24. However the authentication does hang as sivtest looks like to expect more data from the server, which never do arrive. When the server side then finally shutdown the connection sivtest does error out with a base64 decoding error. My guess is sivtest does expect the length of the transmission in curly brackets, which never send by the server. I am unsure if this is a sivtest issue or issue on the server side. As sivtest probably is a all purpose testing utility i think it is worth to report the issue.
Here is what i get.

```
imat@speedy:~> sivtest -p 4190 -u imat -m GSSAPI manitou.disconnected.homeip.net
S: "IMPLEMENTATION" "Dovecot Pigeonhole"
S: "SIEVE" "fileinto reject envelope encoded-character vacation subaddress comparator-i;ascii-numeric relational regex imap4flags copy include body variables enotify environment mailbox date index ihave duplicate mime foreverypart extracttext"
S: "NOTIFY" "mailto"
S: "SASL" "GSSAPI"
S: "STARTTLS"
S: "VERSION" "1.0"
S: OK "Dovecot ready."
C: AUTHENTICATE "GSSAPI" {1076+}
[...]

PR #5465, opened 2025-06-05
sivtest: remove quotes if needed

pzirnik:
cleaned up the patch from issue#5464, please review.

### DIFF
--- a/imtest/imtest.c
+++ b/imtest/imtest.c
@@ -895,7 +895,17 @@ static imt_stat getauthline(struct sasl_cmd_t *sasl_cmd, char **line, int *linel
     else if (sasl_cmd->cont) {
         str += strlen(sasl_cmd->cont); /* jump past the continuation */
     }
-    else {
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
I am not sure how will this work with the other protocols supported by imtest (which installs as symlinks), but I think it is okay.  I also guess that modifying `str` is okay, which can be verified by running `sivtest` under valgrind with many responses by the server (I have not verified it).

Why doesn't `str` jump to the byte after the first `"`?

You could add a check on the last `else` whether the first byte is `{` and emit an error, if it is not.

N.B. Progress here does not depend on me and opinions, opposite to what I said, are possible.

dilyanpalauzov:
Will this work? : 

Closes: https://github.com/cyrusimap/cyrus-imapd/issues/5464.

dilyanpalauzov:
> Will this work? :
> 
> Closes: #5464.

It does not work, at least not, when it is from me.

PR #5466, opened 2025-06-06
imtest: fix output for `C: {0+}`

dilyanpalauzov:
Before this change the output of `sivtest` contained
```
{0+}
C:
```
now it is
```
C: {0+}
```

### DIFF
--- a/imtest/imtest.c
+++ b/imtest/imtest.c
@@ -1115,7 +1115,7 @@ static int auth_sasl(struct sasl_cmd_t *sasl_cmd, const char *mechlist)
         } else if (sendliteral) {
             /* If we had no response, we still need to send the
                empty literal in this case */
-            printf("{0+}\r\nC: ");
+            fputs("C: {0+}", stdout);
             prot_printf(pout, "{0+}\r\n");
         } else if (!initial_response) {
             printf("C: ");


imtest/imtest.c
```diff
@@ -1115,7 +1115,7 @@ static int auth_sasl(struct sasl_cmd_t *sasl_cmd, const char *mechlist)
         } else if (sendliteral) {
             /* If we had no response, we still need to send the
                empty literal in this case */
-            printf("{0+}\r\nC: ");
+            printf("C: {0+}");
```
elliefm: I don't know this code very well, but from looking at the other sections it looks like the "C: " needs to be conditional on initial_response.  I'm not sure if initial_response is already known to be false in this case.  Does something like this make sense?

```suggestion
            printf("%s{0+}", initial_response ? "" : "C: ");
```

Otherwise, `printf` with no format arguments is a bad smell, though I realise this line was already like that.  It's better to use `fputs`, like this:

```suggestion
            fputs("C: {0+}", stdout);
```

dilyanpalauzov: The output of imtest is supposed to be lines starting with `S:` or `C:` followed by the information sent from the `S`erver or the `C`client.

Without the proposed change, the output is
```
…
C: AUTHENTICATE "GSSAPI" {1064+}
YIIDGQ… (cut) … aku4=
S: {208}
S: YIGZBgkqh … (cut) …BD
{0+}
C: 
S: {44}
S: BQQF/wAMAAAAAAAAGGK+YwEAAAAdt7Y/I8rfrO6HD/Y=
C: {44+}
C: BQQE/wAMAAAAAAAAJ5iecwEAAAARfnm6UdgxhZSsoV0=
S: OK
Authenticated.
Security strength factor: 256
```
With the proposed change the output is:
```
C: AUTHENTICATE "GSSAPI" {1064+}
YIIDGQYJ… (cut) … mzA=
S: {208}
S: YIGZBgk… (cut) … X1GxQ3
C: {0+}
S: {44}
S: BQQF/wAMAAAAAAAAFX2f4gEAAAB8WNIfq//SCcJqZmc=
C: {44+}
C: BQQE/wAMAAAAAAAAF9tNgwEAAACZjazpQRA2qynveX0=
S: OK
Authenticated.
Security strength factor: 256
```
So this change ensures that `C:` is printed before the data, sent by the client.

`initial_response` clarifies only, if `C:` is emitted, but the current modified snippet prints `C:` unconditionally.  I have not looked at how `initial_response` is used, because I do not consider the way it is used here as problematic.

I could change printf → fputs, if this is the only obstacle.

dilyanpalauzov: > I don't know this code very well, but from looking at the other sections it looks like the "C: " needs to be conditional on initial_response. I'm not sure if initial_response is already known to be false in this case. Does something like this make sense?

Can you provide a real-world example with `imtest` or `sivtest`, where on that line it would be reasonable to hide `C:` by utilizing `initial_response`?

dilyanpalauzov: i replaced `printf()` with `fputs()`.

elliefm: I don't really use this tool, so I don't know of any invocation where the unconditional `C:` would cause problems, I'm only looking at what's different in the nearby code.  My understanding (perhaps incorrect) is that `initial_response` tracks whether for some reason we have already sent `C:` for this line, in which case we try to not send it again.  I don't know why we would have already sent `C:` for this line though.

You know this tool better than I do, and the change you've made is clearly better than what was already there.  So if you're not worried about `initial_response`, neither am I.  We can always fix it later if we do find a case where it's a problem.

PR #5467, opened 2025-06-06
Cassandane: prevent reloading tests in tiny test loader

rjbs:
We can't just turn on fatal warnings here, because they're lexical in scope and won't propagate down into the "do".  This means it's time to switch from "do"-ing the file to reading its content and eval(bytes)-ing them.  Good, this also eliminates any question about @INC and "do" playing together!

Note the introduction of the magic "#line 1" comment.  Because we're not using "do", we need to tell Perl to act like it really knows where the content comes from.


rjbs:
Discussed this with alh over Zoom and merging based on that discussion.

PR #5468, opened 2025-06-08
databases.rst: write text on un/shared annotations

dilyanpalauzov:
Describe how to set the `userid` part of the key in annotations.db.

This is related to https://github.com/cyrusimap/cyrus-imapd/pull/5071.

### DIFF
--- a/docsrc/imap/concepts/deployment/databases.rst
+++ b/docsrc/imap/concepts/deployment/databases.rst
@@ -69,10 +69,27 @@ Annotations (annotations.db)
 ----------------------------
 
 This database contains mailbox and server annotations, including WebDAV
-properties. The database is indexed by mailbox name (empty for server
-annotations) + annotation name + userid (empty for shared annotations) and each
-data record contains the value size, value data, content-type of the data and
-timestamp of the record. The format is each record is as follows::
+properties. The database key starts with the cached mailbox name, UUID, or the
+empty string for server annotations, and is followed by the annotation name,
+then the userid.  Each data record contains the value size, value data,
+content-type of the data and timestamp of the record.
+
+For annotations, including WebDAV properties, where the same value is retrieved
+for all users and the value cannot be set per user, the `userid` part of the
+database key is `[.OwNeR.]` or the empty string.
+
+When `userid` is the empty string, the `\\0` after it is still present.
+
+For annotations, where each user can set a distinct value, overwriting the
+value of the mailbox owner, when the owner sets the value, it again uses
+the empty string or `[.OwNeR.]` for `userid`. Others use their authentication id
+as `userid`.  Handling such values is facilitated by functions from `annotate.h`
+containing `mask` in their name.
+
+When the authenticated user is administrator, `userid` is the empty string
+or `[.OwNeR.]`.
+
+The format is each record is as follows::
 
     Key: <Mailbox Name>\0<Annotation Name>\0<Userid>\0
 


elliefm (approved):

elliefm:
As documentation, this seems fine to me, but I do not know if the content is correct.  Looks like this PR is part of an ongoing discussion with @rsto.

docsrc/imap/concepts/deployment/databases.rst
```diff
@@ -70,9 +70,24 @@ Annotations (annotations.db)
...
-data record contains the value size, value data, content-type of the data and
-timestamp of the record. The format is each record is as follows::
+annotations) + annotation name + userid and each data record contains the
+value size, value data, content-type of the data and timestamp of the record.
+
+For annotations, including WebDAV properties, where the same value is retrieved
+for all users and the value cannot be set per user, the `userid` part of the
+database key is the empty string.
+
+For annotations, where each user can set a distinct value, overwriting the
```
rsto, line 83: This doesn't sound right. In terms of database records, these are the possible cases:

- the record shares its value across users: in this case the userid is not set
- the record maintains it value per user: in this case the userid is the special string `[.OwNeR.]` for the mailbox owner, or the userid (authentication id) for any other user

The `mask`-annotation functions sets the shared value if the authenticated user is an admin.

BTW: while it's not part of this PR, the key format description on line 92 isn't correct for mailboxes indexed by uuid anymore. Skimming this whole document, it looks like a couple of other parts of the documentation also are out of date.

dilyanpalauzov: > • the record maintains it value per user: in this case the userid is the special string `[.OwNeR.]` for the mailbox owner, or the userid (authentication id) for any other user

Sometimes this is correct, but sometimes it does not match reality.  Let’s take as an example the proppatch_caltransp() function for the schedule-calendar-transp PROP_PERUSER property.  It calls `proppatch_todb(prop, set, pctx, propstat, (void *) value);` → `proppatch_todb_internal(prop, set, pctx, propstat, value, 1)` → `annotate_state_writemask(astate, buf_cstring(&pctx->buf), httpd_userisadmin ? "" : httpd_userid, &value)`.  That is:
```c
int annotatemore_writemask(const char *mboxname, const char *entry, const char *userid, const struct buf *value) {
    if (mboxname_userownsmailbox(userid, mboxname))
        return annotatemore_write(mboxname, entry, "", value);
    return annotatemore_write(mboxname, entry, userid, value);
}
```
So when the userid is the mailbox owner or is admin ⇔ httpd_userisadmin, is called `annotatemore_write(mboxname, entry, "", value);` → `write_entry(mailbox, /*uid*/0, entry, userid ⇔ "", value, /*ignorequota*/1, /*silent*/0, NULL, /*maywrite*/1);`. `write_entry()` calls `make_key()` and the latter can substitute userid with `[.OwNeR.]`, but it does not substitute `""` with `[.OwNeR.]`.  So in the annotations.db database is stored `""`.

Test on commit e6d7b8a5a1 with
```
curl -XPROPPATCH -Hcontent-type:application/xml -uggg@fff:ggg --data-binary @- <<EOF http://127.0.0.3/dav/calendars/user/ggg@fff/Default
<propertyupdate xmlns="DAV:">
  <set>
    <prop>
      <schedule-calendar-transp xmlns="urn:ietf:params:xml:ns:caldav"><transparent/></schedule-calendar-transp>
    </prop>
  </set>
</propertyupdate>
EOF
curl -XPROPPATCH -Hcontent-type:application/xml -up@fff:p --data-binary @- <<EOF http://127.0.0.3/dav/calendars/user/ggg@fff/Default
<propertyupdate xmlns="DAV:">
  <set>
    <prop>
      <schedule-calendar-transp xmlns="urn:ietf:params:xml:ns:caldav"><transparent/></schedule-calendar-transp>
    </prop>
  </set>
</propertyupdate>
EOF
```
And then
```
$ cvt_cyrusdb /etc/cyrus/annotations.db twoskip /etc/cyrus/x.db flat
Converting from /etc/cyrus/annotations.db (twoskip) to /etc/cyrus/x.db (flat)
```
Finally
```
$ grep -ia owner x.db

0eff8284-a63c-4983-bc1b-5fc6f3f44d28/vendor/cmu/cyrus-jmap/defaultalerts[.OwNeR.]       DEFAULTALARMS %(WITH_TIME %(CONTENT "" GUID 0000000000000000000000000000000000000000 ATAG NIL) WITH_DATE %(CONTENT "" GUID 0000000000000000000000000000000000000000 ATAG NIL))text/plain
8318adac-4794-4d25-9d8f-b82502fd70c8/vendor/cmu/cyrus-jmap/defaultalerts[.OwNeR.]       DEFAULTALARMS %(WITH_TIME %(CONTENT "" GUID 0000000000000000000000000000000000000000 ATAG NIL) WITH_DATE %(CONTENT "" GUID 0000000000000000000000000000000000000000 ATAG NIL))text/plain
a8fde1a0-b596-4277-985b-faf3a2d84045/vendor/cmu/cyrus-jmap/defaultalerts[.OwNeR.]       DEFAULTALARMS %(WITH_TIME %(CONTENT "" GUID 0000000000000000000000000000000000000000 ATAG NIL) WITH_DATE %(CONTENT "" GUID 0000000000000000000000000000000000000000 ATAG NIL))text/plain

$ grep -ia trans x.db 
0eff8284-a63c-4983-bc1b-5fc6f3f44d28/vendor/cmu/cyrus-httpd/<urn:ietf:params:xml:ns:caldav>schedule-calendar-transpp@fff
                transparenttext/plain
0eff8284-a63c-4983-bc1b-5fc6f3f44d28/vendor/cmu/cyrus-httpd/<urn:ietf:params:xml:ns:caldav>schedule-calendar-transp
        transparenttext/plain
```
So `[.OwNeR.]` is used for  `defaultalerts`, but not when setting `<urn:ietf:params:xml:ns:caldav>schedule-calendar-transp`.  To verify this I pass the output of the last commad over hexdump -C and then make sure that after nsp follow `ff 80 ff 80`, so the userid part is the empty string:
```
00000140  61 6e 73 70 ff 80 ff 80  09 ff 80 ff 80 ff 80 0b  |ansp............|
```

In any case it would be good to agree how this is supposed to work and somebody could write in the future a migration tool which applies consistently the `userid` part on all database keys.

rsto: > Sometimes this is correct, but sometimes it does not match reality. 

It is correct insofar as it describes the actual data format on disk: each record (=a key-value pair) only can be either a shared record (by not having a userid set), or a per-user record (by having the user id set, possibly to the magic `[.OwNeR.]` string).

I do however agree that starting from that it soon gets confusing: each caller of the API has to decide for each WebDAV property (or other annotations) if they should be stored "masked" or "not masked", and similarly how they should be read. That's the discrepancy that you see between e.g. the `schedule-calendar-transp` property and the `defaultalerts` annotation. Then there's even the chance that the same annotation could be written masked, but read unmasked, or the other way (which would almost certainly be a bug, but the API doesn't prevent that).

Adding to that, a property can have multiple records in the database, e.g. for each user and an additional one for a shared value of the owner.

> In any case it would be good to agree how this is supposed to work and somebody could write in the future a migration tool which applies consistently the `userid` part on all database keys.

I'm not sure if we'll ever come up with a scheme where the userid is set consistently, if that would imply making all properties either masked or non-masked properties. But it certainly would be great to look at the complete list of annotations written in the Cyrus codebase and for each determine if its current read/write mode (masked or non-masked) is actually the right one.

dilyanpalauzov: Please suggest how the documentation should be improved.

rsto: I won't find time for that anytime soon.

dilyanpalauzov:
> BTW: while it's not part of this PR, the key format description on line 92 isn't correct for mailboxes indexed by uuid anymore. 

For a pre-UUID annotations.db I have the entries starting with

```
0ed384eb-6c22-4420-b783-1e93e67633fb<FF><80>/vendor/cmu/cyrus-httpd/<http://apple.com/ns/ical/>calendar-order  >
0ed384eb-6c22-4420-b783-1e93e67633fb<FF><80>/vendor/cmu/cyrus-httpd/<urn:ietf:params:xml:ns:caldav>supported-ca>
0f20cbb2-c211-4c6b-80e3-ec0d64b33481<FF><80>/vendor/cmu/cyrus-httpd/<DAV:>displayname<FF><80><FF><80>   <FF><80>
0pqaqovm4qg9a7maq6vvfe9u<FF><80>/vendor/cmu/cyrus-httpd/<DAV:>displayname<FF><80><FF><80>       <FF><80><FF><80>
0pqaqovm4qg9a7maq6vvfe9u<FF><80>/vendor/cmu/cyrus-httpd/<http://apple.com/ns/ical/>calendar-color<FF><80><FF>  >
0pqaqovm4qg9a7maq6vvfe9u<FF><80>/vendor/cmu/cyrus-httpd/<urn:ietf:params:xml:ns:caldav>calendar-description<FF>>
0pqaqovm4qg9a7maq6vvfe9u<FF><80>/vendor/cmu/cyrus-httpd/<urn:ietf:params:xml:ns:caldav>supported-calendar-compo>
0sahaud53vv3clb5pdgxr072<FF><80>/vendor/cmu/cyrus-httpd/<DAV:>displayname<FF><80><FF><80>       <FF><80><FF><80>
10bc4099-9344-4bad-a2b5-a2bced9ca8a7<FF><80>/vendor/cmu/cyrus-httpd/<DAV:>displayname<FF><80><FF><80>   <FF><80>
10bc4099-9344-4bad-a2b5-a2bced9ca8a7<FF><80>/vendor/cmu/cyrus-httpd/<http://apple.com/ns/ical/>calendar-color  >
1c4fe6665815b9a1<FF><80>/vendor/cmu/cyrus-httpd/<http://apple.com/ns/ical/>calendar-color<FF><80><FF><80>      >
2724f0314e1c705c<FF><80>/specialuse<FF><80>[.OwNeR.]<FF><80>    <FF><80><FF><80><FF><80>^F<FF><80><FF><80><FF> >
273c64535850c0a4<FF><80>/vendor/cmu/cyrus-httpd/<DAV:>displayname<FF><80><FF><80>       <FF><80><FF><80><FF><80>
273c64535850c0a4<FF><80>/vendor/cmu/cyrus-httpd/<http://apple.com/ns/ical/>calendar-color<FF><80><FF><80>      >
```
so this does not look like a mailbox name as the start of the key. (Lines are cut at the end)  What are `0pqaqovm4qg9a7maq6vvfe9u` and `1c4fe6665815b9a1` if it they are neither mailbox names nor UUIDs?

dilyanpalauzov:
> What build version is this?

I use 3.8, but the database is very old, let’s it was filled with data since 2003.

rsto:
these are ancient mailbox ids, the original ids were just a hash of the mailbox name

dilyanpalauzov:
I have changed the text to state, that as `userid` in the database key both `[.OwNeR.]` and the empty string can be used, leaving unspecified what should be used when. (Because it is not clear for me).

As for the first component of the database key, I have written that it can be UUID, cache of mailbox name, or emtpy for server-wide annotations.

PR #5471, opened 2025-06-11
slowio: use cputime, second try

elliefm:
Same as #5409, but hopefully with the NixOS/CentOS issue fixed?

Turns out there was an edge-case bug in my original slowio implementation, which was exposed on some platforms once we switched from MONOTONIC to CPUTIME.

### DIFF
--- a/lib/slowio.c
+++ b/lib/slowio.c
@@ -50,6 +50,7 @@
 
 static struct slowio slowio_read = { 0 };
 static struct slowio slowio_write = { 0 };
+static double slowio_min_elapsed = NAN;
 
 EXPORTED void slowio_reset_impl(void)
 {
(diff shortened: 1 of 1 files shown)


elliefm:
If the slowio_min_elapsed fix alone is _not_ enough to solve the problem, I still think I should drop the first commit.  But if that's the case, then I'd think the next problem might be the design of the limit2 test, rather than the code being tested.

wolfsage (approved):
At least on my NixOS vm, it seems like [WIP slowio: can't use nanosleep when measuring cpu time](https://github.com/cyrusimap/cyrus-imapd/pull/5471/commits/46919964ab666fb29f8dcd885d09741d349071ac) is all that's needed on top of the original patch to get it working reliably

wolfsage:
I'd drop the second commit and reword the WIP and then merge maybe

elliefm:
I'd rather prefer to keep the second commit and drop the first, if that arrangement would solve the NixOS problem.

That is to say:

I think the second commit, [WIP slowio: definitely must delay when elapsed is zero](https://github.com/cyrusimap/cyrus-imapd/pull/5471/commits/221dd358a4e34bf39e2c33135f618976f2fa125c), fixes a real (if hypothetically rare) bug in the slowio implementation.  We ought to fix this bug regardless of whether it helps this situation specifically, so I want to keep this commit.  I suspect it might actually be the correct fix for the NixOS issue too.

Whereas I think the first commit, [WIP slowio: can't use nanosleep when measuring cpu time](https://github.com/cyrusimap/cyrus-imapd/pull/5471/commits/46919964ab666fb29f8dcd885d09741d349071ac), is a crude workaround for the cunit tests failing on NixOS, but which makes slowio behave badly for everyone in the situation where it matters (Cassandane).  I'll revert (but not yet drop) this commit now, so we can easily see the other one's behaviour in isolation.

ksmurchison (approved):
LGTM

PR #5254, opened 2025-02-04
Revert "message.c: reject to parse MIME headers with NUL byte"

rsto:
This reverts commits f56c25f7fa5eb77668e8d3ba706ca87d0c93ad56 and 0f137e7ab8dfbfaa14b0113cc8def0cc9eb2d94d to avoid reconstruct failing for messages that contain NUL bytes. We'll come up with a proper fix in https://github.com/cyrusimap/cyrus-imapd/pull/5253

PR #5472, opened 2025-06-12
jmap_mail.c: use safe string comparison for body part types

rsto:
The MIME parser in almost all cases guarantees the body part media type to be set, but it does not for invalid MIME messages that already are on disk. This is a known issue that needs to get fixed in "reconstruct" first.

Also see https://github.com/cyrusimap/cyrus-imapd/pull/5254


ksmurchison:
Agree with @wolfsage about the missing cases

PR #5480, opened 2025-06-18
message_get_deliveredto(): only look for X-Delivered-To header

ksmurchison:
Per CYR-1922


rsto:
> Can we do anything to make Xapian ignore that data in the index or are both headers indexed under the same key?

It's the same key, only a full reindex will do the trick.

PR #5482, opened 2025-06-23
imapd: fix typo in SASL-IR capability

### LINKED ISSUE #5481: RFC 4959/SASL-IR capability broken, causing additional IMAP4rev1 round trips for some AUTHENTICATE schemes
Greetings,

(I am the [fetchmail](https://www.fetchmail.info/) maintainer, but fetchmail apparently is unaffected when polling from Cyrus IMAP. I saw this by chance reading protocol traces when testing.)

I recently figured that one of my typical IMAP servers greeted me like this:

`* OK [CAPABILITY IMAP4rev1 IMAP4rev2 ENABLE ID LITERAL+ LOGINDISABLED SASL_IR STARTTLS] unimail Cyrus IMAP 3.12.0 server ready`

note the `SASL_IR`, which is not compliant to [RFC4959](https://datatracker.ietf.org/doc/html/rfc4959) and should be `SASL-IR` instead.

This might have been broken since 3.10, and particulary I find `SASL_IR` only in once place in the source:

```text
$ rg -wC2 SASL_IR
imap/imapd.c
449-    { "REPLACE",               CAPA_POSTAUTH,           { 0 } }, /* RFC 8508 */
450-    { "RIGHTS=kxten",          CAPA_POSTAUTH,           { 0 } }, /* RFC 4314 */
451:    { "SASL_IR",               CAPA_PREAUTH,            { 0 } }, /* RFC 4959 */
452-    { "SAVEDATE",              CAPA_POSTAUTH,           { 0 } }, /* RFC 8514 */
453-    { "SEARCH=FUZZY",          CAPA_POSTAUTH,           { 0 } }, /* RFC 6203 */
```

and that particular line 451 traces back to commit cbe7d93de5e3037e0bed3def3f4661dab4efc0df, which is part of branch imap_capa_response, and happened roughly 330 commits before cyrus-imapd-3.10.0-alpha0.  To me looks like a simple typo when restructuring the IMAP CAPABILITY support, which only has a tiny impact in that it requires some IMAP4rev1 clients to do one extra round trip when they used `<tag> AUTHENTICATE <method>`, but not on, for instance, LOGIN.

Please fix said line to read `"SASL-IR"` instead, and re-test if SASL AUTHENTICATE client-side commands then work.

I suppose not many parties noticed because IMAP4rev2 clients are unaffected and IMAP4rev1 clients just suffer one additional AUTHENTICATE round trip because they can't provide client data on the AUTHENTICATE request already and many networks are fast these days.

FreeBSD downstream report with patch: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=287657

elliefm:
Fixes #5481

I don't see a good way to test this.  It is possible to coerce Cassandane and Mail::IMAPTalk into using AUTHENTICATE instead of LOGIN, but then Mail::IMAPTalk doesn't expose the details of the SASL transaction, so I can't interrogate whether the initial response was actually used or not.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Simple.pm
+++ b/cassandane/Cassandane/Cyrus/Simple.pm
@@ -47,6 +47,7 @@ use DateTime;
 use lib '.';
 use base qw(Cassandane::Cyrus::TestCase);
 use Cassandane::Util::Log;
+use Cassandane::Util::Slurp;
 
 $Data::Dumper::Sortkeys = 1;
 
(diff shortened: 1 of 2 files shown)


ksmurchison (dismissed):

elliefm:
Found a way to test it by invoking `imtest` and parsing the output...

PR #5483, opened 2025-06-23
imapparse.c: don't prot_ungetc('N') before parsing "IL"

ksmurchison:
- first, we already know we have the 'N' so no need to parse it again
- second, if we actually have "NIL" we zero-out the buffer, and if we don't have "NIL", we fail and ignore the string
- last, and most importantly, if we have to refill the prot_stream buffer between parsing the 'N' and the rest of the string, trying to ungetc the 'N' will result in a fatal()

The fatal() is an edge case seen in the wild while using compression and can be reproduced with the ID.cmd_id_nil_with_compress test

This commit also cleans up the NIL parsing code to use buffer accessor functions and NOT use buffer internals directly


cassandane/Cassandane/Cyrus/ID.pm
```diff
@@ -96,4 +96,23 @@ sub test_cmd_id
     }
 }
 
+sub bogus_test_cmd_id_nil_with_compress
```
elliefm: This test is disabled, is that something temporary from testing that accidentally got left behind?  Is it a negative test that fails when it's working correctly?  Is it just really slow because of the 10,000? It'd be nice for this to be a real test so it can detect regressions for us in the future.  If it's just slow you could mark it so:

```suggestion
sub test_cmd_id_nil_with_compress_slow
```

I'm about to fetch this branch and see for myself, so I'll probably be back with another suggestion once I see what's really going on

elliefm: The test runs and passes just fine for me locally, doesn't even take that long (a few seconds), so I think just enable it normally:

```suggestion
sub test_cmd_id_nil_with_compress
```

elliefm (approved):
Looks good, but please enable the cmd_id_nil_with_compress test by removing the "bogus_" from the name and let CI go again before merging.

PR #5487, opened 2025-06-25
reconstruct.c: implement use of --dry-run (-n) with --set-version (-V)

ksmurchison:
reconstruct will still report repack errors,
but the new cyrus.index will not be written to disk

### DIFF
--- a/cassandane/Cassandane/Cyrus/Reconstruct.pm
+++ b/cassandane/Cassandane/Cyrus/Reconstruct.pm
@@ -580,6 +580,57 @@ sub test_reconstruct_uniqueid_from_header_uuidmb
     $self->assert_str_equals("user\x1fcassandane", $hash->{N});
 }
 
+sub test_downgrade_dryrun
+{
+    my ($self) = @_;
+
+    my $talk = $self->{store}->get_client();
+
+    xlog $self, "Add two messages";
+    $self->make_message('Message A');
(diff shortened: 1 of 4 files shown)


rjbs (approved):
LGTM.  Do we want an assertion about the exit status of `reconstruct` in "worked but dry run" mode?  (It should be zero.)  But seems correct to me

elliefm:
> LGTM. Do we want an assertion about the exit status of `reconstruct` in "worked but dry run" mode? (It should be zero.) But seems correct to me

Do you mean in the tests? `run_command`'s (not-obvious*) default is to die if the child does anything other than return zero, so you only need an explicit assertion after it if you're doing something tricky.

(* The defaults come from `%default_command_handlers`, which are installed by `_add_child()` if no custom handlers were provided.)

ksmurchison:
> LGTM. Do we want an assertion about the exit status of `reconstruct` in "worked but dry run" mode? (It should be zero.) But seems correct to me

@rjbs I'm not following your question.

rjbs:
@ksmurchison I mean:  We have chosen that in a dry run, we exit 0.  In the future, we might have a bug where the program crashes, but the data file is untouched, so the test passes.

I've done some more reading of `run_command`, though, and I think that we'd throw an exception of the run command exited nonzero, so we should be good here!

PR #5489, opened 2025-06-26
imapd.c: split logging of client use of ANNOTATE (message) and GET/SETANNOTATION (mailbox)

ksmurchison:
Also, logs ANNOTATE use with APPEND, STORE, SEARCH, SELECT


rjbs:
Looks good.  I think probably this should've been >1 commit, as part is "fix annotate/metadata" and one is "set behavior mask closer to detection".  But that's kind whatever.

**I think you should rename `annotatemore` to `annotate_old` or `annotate_obs` so we can remember that it's the old/obsolete behavior without looking up the core/more distinction.**

Approved as is, though.

rjbs (approved):
(clicked approve, but would kinda like one rename)

Issue #5490, opened 2025-06-26
`messageContent` missing in LMTP `MessageNew` events causes delivery failure when notifications are enabled


MarioHoberg:
## Summary

When messages are delivered via **LMTP** and the `MessageNew` event is emitted from `append_fromstage_full()`, the `messageContent` is not extracted. This results in **fatal mail delivery failures** when event notifications are configured to include the `messageContent` field.

## Version

The bug was observed on

- Cyrus IMAP 3.12.0 (from Debian experimental)
- Cyrus IMAP 3.8.1-1~bpo12+1 (Debian backports)
- Cyrus IMAP 3.6.1-4+deb12u3 (Debian bookworm)
- Cyrus IMAP 3.2.6-2+deb11u4 (Debian bullseye)

## Symptoms

- Mail delivery via LMTP fails **entirely** when event notifications are enabled and configured to include `messageContent`.
- Syslog output includes the following fatal assertion:
  ```
  cyrus/lmtp[3634436]: FATAL: Internal error: assertion failed: imap/mboxevent.c: 743: filled_params(type, event)
  ```

## Root Cause

- The function `mboxevent_extract_content_msgrec()` is responsible for extracting the `messageContent` field used in notifications.
- It is currently **only** invoked in `append_fromstream()`.
- However, LMTP delivery uses `append_fromstage_full()`, which emits `MessageNew` events but **does not extract message content**.
- When the `messageContent` field is expected by notification consumers, its absence causes the assertion in `filled_params()` to fail, **terminating delivery**.

## Minimal Patch

To resolve the issue, insert the following block into `append_fromstage_full()` after the message file is written:

```diff
diff --git a/imap/append.c b/imap/append.c
index 1340f2854..3db7380c8 100644
--- a/imap/append.c
+++ b/imap/append.c
@@ -1195,6 +1195,14 @@ out:
     }
 
     /* finish filling the event notification */
+    destfile = fopen(fname, "r");
+    if (!destfile) {
+      syslog(LOG_ERR, "IOERROR: opening message file %s: %m", fname);
+    } else {
+      /* Extract messageContent for MessageNew notifications */
+      mboxevent_extract_content_msgrec(mboxevent, msgrec, destfile);
+      fclose(destfile);
+    }
     /* XXX avoid to parse ENVELOPE record since Message-Id is already
      * present in body structure ? */
     mboxevent_extract_msgrecord(mboxevent, msgrec);

```

[...]

PR #5491, opened 2025-06-26
Provide missing messageContent for new messages delivered via LMTP

MarioHoberg:
This PR adresses issue #5490 

## ✅ Summary

This PR ensures that `messageContent` is included in `MessageNew` events emitted during LMTP delivery (`append_fromstage_full()`). Without this, systems configured to send event notifications that include `messageContent` may experience fatal delivery failures.

## 🐛 Problem

Currently, `messageContent` is only extracted in `append_fromstream()`, not in `append_fromstage_full()`, which is used during LMTP delivery. If notifications are enabled and expect `messageContent`, delivery fails with:

```
FATAL: Internal error: assertion failed: imap/mboxevent.c: 743: filled_params(type, event)
```

## 🔧 Fix

This patch opens the newly written message file and calls `mboxevent_extract_content_msgrec()` inside `append_fromstage_full()` to ensure `messageContent` is populated in the event structure, mirroring the behavior in `append_fromstream()`.

## 🧪 Tested

Tested locally with LMTP delivery and notification configuration requiring `messageContent`. Messages are now delivered successfully and the content is included in the emitted event.

### DIFF
--- a/imap/append.c
+++ b/imap/append.c
@@ -1195,13 +1195,13 @@ out:
     }
 
     /* finish filling the event notification */
-    destfile = fopen(fname, "r");
-    if (!destfile) {
+    FILE *msgfile = fopen(fname, "r");
+    if (!msgfile) {
       syslog(LOG_ERR, "IOERROR: opening message file %s: %m", fname);
     } else {
       /* Extract messageContent for MessageNew notifications */
-      mboxevent_extract_content_msgrec(mboxevent, msgrec, destfile);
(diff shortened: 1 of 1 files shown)


MarioHoberg:
Sorry, I messed up with the differend versions. The last commit was successfully built and tested with master.

dilyanpalauzov:
As you can see on the other lines, the indentation is four spaces, you use two spaces.

PR #5493, opened 2025-06-30
cassandane: option to suppress success output (only show errors and failures)

elliefm:
When cassandane tests fail in CI, it can be hard to see which ones failed:

* at the "run cassandane quietly" stage, the failing tests are lost in a sea of OKs
* at the "rerun cassandane failures noisily" stage, they're lost in a sea of detailed error logging

This PR adds a "no_ok" formatter parameter for suppressing success results, and implements it for FormatPretty (i.e. `pretty` and `prettier`).  Formatters that don't recognise it (`tap`, `xml`) will ignore it.  Call `testrunner.pl` with `--no-ok` to enable this behaviour.

`cyd test` doesn't yet know about `--no-ok` and will reject it as an invalid argument, so to use the new testrunner option with cyd, for now you'll need to smuggle it through as a non-option argument.  The main.yml workflow has been updated to do this for the "run cassandane quietly" step.

**Deliberately fails CI**, due to the "JUNK force a failure and an error" commit, which I will drop before merging.  View the CI reports and check the "run cassandane quietly" output: it's now very easy to see exactly which tests failed, because there isn't ~2k lines of "[ OK ] Some.other_test" cluttering things up anymore.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,7 +65,7 @@ jobs:
       id: cass1
       continue-on-error: true
       # We haven't figured out how to make Test::Core work in actions yet.
-      run: cyd test --slow "!Test::Core" --format prettier "${{ matrix.skip }}"
+      run: cyd test --no-ok --slow "!Test::Core" --format prettier "${{ matrix.skip }}"
     - name: rerun cassandane failures noisily
       if: ${{ steps.cass1.outcome == 'failure' }}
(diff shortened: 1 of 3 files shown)


elliefm:
Will hold off on merging this till https://github.com/cyrusimap/cyrus-docker/pull/47 lands, because that'll let me make the github-actions change a bit nicer immediately, instead of needing to come back and tidy it up later

PR #5497, opened 2025-07-02
sync_support.c: don't error out on missing tombstone

ksmurchison:


rjbs (approved):

imap/sync_support.c
```diff
@@ -6909,9 +6909,15 @@ static int do_folders(struct sync_client_state *sync_cs,
...
-            /* XXX copy the missing local mailbox back from the replica? */
+            if (mboxname_isdeletedmailbox(rfolder->name, NULL)) {
+                /* not an error if tombstone for DELETED.* is missing */
+                r = 0;
+            }
+            else {
+                syslog(LOG_NOTICE, "SYNCNOTICE: no tombstone for deleted mailbox %s (%s)",
+                       rfolder->name, error_message(r));
+                /* XXX copy the missing local mailbox back from the replica? */
+            }
```
wolfsage, line 6915: The problem with the previous code is that 'r' might not get reset below if we don't hit this:
```
            /* Found unprocessed item which should rename cleanly */
            r = folder_rename(sync_cs, item->oldname, item->newname, item->part,
                              item->uidvalidity);
```
... so this notice is seemingly informational, but can cause the caller to assume failure randomly based on if later work is attempted or not.

With this fix, we now set r to 0 for a specific known case, but we still let what looks like something that should be a failure maybe get reset to 0 by later code in the same block...

Should this instead be doing `goto bail` and using log_err? Or should it always set r to 0 and only log notice if we think it's unexpected?

wolfsage (changes requested):

imap/sync_support.c
ksmurchison: @brong you know this code better then anyone.  What do we want here?

brong: There's more as well, you'd also have to have no master_folders, BUT - you're right, we should just set r=0 and only log if it's not in the DELETED namespace.

brong (changes requested):
yeah, let's set `r = 0` in all cases, there was a legit potential bug where there's nothing on the master and it would have errored out.

imap/sync_support.c
```diff
@@ -6909,9 +6909,13 @@ static int do_folders(struct sync_client_state *sync_cs,
...
         else {
             mboxlist_entry_free(&tombstone);
-            syslog(LOG_NOTICE, "SYNCNOTICE: no tombstone for deleted mailbox %s (%s)",
-                               rfolder->name, error_message(r));
-            /* XXX copy the missing local mailbox back from the replica? */
+            r = 0;
+
+            if (!mboxname_isdeletedmailbox(rfolder->name, NULL)) {
+                syslog(LOG_NOTICE, "SYNCNOTICE: no tombstone for deleted mailbox %s (%s)",
+                       rfolder->name, error_message(r));
```
wolfsage, line 6913: You set `r = 0` above so this `error_message(r)` won't be accurate. I think you want to reverse the order

ksmurchison: Fixed

PR #5501, opened 2025-07-08
Cassandane/Cyrus/ID: Don't lock up the test if tcp buffers fill

wolfsage:


wolfsage:
Closing in favor of https://github.com/cyrusimap/cyrus-imapd/pull/5505

PR #5502, opened 2025-07-08
cyrusdb.c: dump binary data values as hex

ksmurchison:
Did this mostly to see conv.db G record values

### DIFF
--- a/lib/cyrusdb.c
+++ b/lib/cyrusdb.c
@@ -40,6 +40,7 @@
  */
 
 #include <config.h>
+#include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


lib/cyrusdb.c
```diff
@@ -491,9 +492,26 @@ static int print_cb(void *rock,
                     const char *data, size_t datalen)
 {
     FILE *f = (FILE *)rock;
+    bool is_binary = 0;
```
rsto: If we already use bool, let's also use `true` or `false`, rather than their integer representations.

lib/cyrusdb.c
```diff
@@ -491,9 +492,26 @@ static int print_cb(void *rock,
...
-    fprintf(f, "%.*s\t%.*s\n", (int)keylen, key, (int)datalen, data);
+    for (i = 0; i < datalen; i++) {
+        if (!isprint(data[i])) {
+            is_binary = 1;
+            break;
+        }
+    }
+
+    fprintf(f, "%.*s\t", (int)keylen, key);
+    if (is_binary) {
```
rsto, line 506: I don't have an immediate need for this, so if you prefer to see either everything as printable ASCII or nothing, this is good for me. But maybe it might be useful to just display the non-printable characters as hex?

lib/cyrusdb.c
ksmurchison: Good call.  Fixed.

lib/cyrusdb.c
ksmurchison: In the case of G records, the data is concatenated uint32/uint64 values.  I think printing some part of those as ascii chars wouldn't be helpful, e.g:  45 'a' b4 5e

wolfsage: Really what'd be nice is hexdump style display:

`38 33 38 34 00 00 13 1a  25 2b                    |8384....%+|`

but I can live with this :)

Thanks

PR #5503, opened 2025-07-09
cass support systemd-coredump

elliefm:


cassandane/Cassandane/Instance.pm
```diff
@@ -1638,8 +1644,15 @@ sub find_cores
...
 {
     my ($self) = @_;
-    my $coredir = $self->{basedir} . '/conf/cores';
 
+    my $cassini = Cassandane::Cassini->instance();
+    my $coredir = $cassini->val('cassandane', 'core_directory', undef);
+
+    # core_directory with absolute path may include unrelated systemwide core
+    # files, so for now we can't error on unexpected core files existing
+    return if $coredir && substr($coredir, 0, 1) eq '/';
```
brong, line 1653: Can we just error out with a "you can't run tests when there's crap sitting in your core directory" instead?  Or have every test read the core list at the start and compare it at the end (though of course that means that tests could fail incorrectly due to a different test dumping core, or something else on your system).

This seems more likely to miss things coring if we don't error out when we get core dumps.

brong (changes requested):
I'd prefer to require setting kernel pattern; or some other level of "require a clean slate in the environment you're in" here.  This seems too brittle.

cassandane/Cassandane/Instance.pm
elliefm: Typed my proper comment in the wrong box, but yeah this is bogus as it is now.  This is just the point where I stalled out, because what I actually wanted doesn't seem possible!

PR #5505, opened 2025-07-09
ID.pm: better test checking parsing of NIL on prot buffer boundary

ksmurchison:
test_cmd_id_nil_with_compress would lock up if tcp buffers fill

### DIFF
--- a/cassandane/Cassandane/Cyrus/ID.pm
+++ b/cassandane/Cassandane/Cyrus/ID.pm
@@ -96,23 +96,27 @@ sub test_cmd_id
     }
 }
 
-sub test_cmd_id_nil_with_compress
+sub test_cmd_id_nil_cant_unget
 {
     my ($self) = @_;
 
     # Purge any syslog lines before this test runs.
     $self->{instance}->getsyslog();
 
-    my $imaptalk = $self->{store}->get_client(UseCompress => 1);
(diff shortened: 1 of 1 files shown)


cassandane/Cassandane/Cyrus/ID.pm
```diff
@@ -96,23 +96,27 @@ sub test_cmd_id
...
+    # in the prot buffer.
+    # This will require a new read to get the rest of the command,
+    # but will also prohibit calling prot_ungetc('N').
+    # Successful execution of the command will verify that we have fixed the
+    # parsing issue.
+    # If the previous bug returns, imapd will fatal() attempting to prot_unget()
+    my $x = 'x' x 1014;
+    $imaptalk->{CmdId} = 'XX';
+    $imaptalk->_imap_cmd('ID', 0, {},
+                         "(\"a\" \"$x\" \"b\" \"$x\" \"c\" \"$x\" \"d\" \"$x\" \"e\" NIL)" );
```
wolfsage: One nit: This'll be much more readable as:
```qq{("a" "$x" "b" "$x" "c" "$x" "d" "$x" "e" NIL)}```

ksmurchison: Thx.  I knew there had to be a better way..  FIxed.

PR #5506, opened 2025-07-10
mailbox: cache conversationsdb state, including lack of

brong:
We used to do a check if conversations were enabled, which required a mailboxes.db lookup, for EVERY record in some tight loops.  This caches the value while still meaning that release commands also release the conversation state

### DIFF
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -2711,7 +2711,7 @@ EXPORTED void mailbox_unlock_index(struct mailbox *mailbox, struct statusdata *s
         }
     }
 
-    if (mailbox->cstate_flag == 3) {
+    if (mailbox->cstate_flag == CSTATE_FLAG_LOCAL) {
         int r = conversations_commit(&mailbox->cstate_value);
         if (r) {
             xsyslog(LOG_ERR, "IOERROR: Error committing to conversations database",
@@ -2719,8 +2719,8 @@ EXPORTED void mailbox_unlock_index(struct mailbox *mailbox, struct statusdata *s
                     mailbox_name(mailbox), error_message(r));
         }
     }
+    mailbox->cstate_flag = CSTATE_FLAG_UNSET;
     mailbox->cstate_value = NULL;
-    mailbox->cstate_flag = 0;
 
     /* release caches */
     int i;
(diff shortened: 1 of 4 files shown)


imap/mailbox.c
```diff
@@ -4394,29 +4398,44 @@ EXPORTED int mailbox_add_sieve(struct mailbox *mailbox)
...
 
+// because we use mailbox names in conversations_get_mbox and mailbox_has_conversations_full,
+// those require mailboxes.db lookups - but we want this particular function to be usable
+// in tight inner loops iterating over the mailbox, so we cache.
+//
+// We have 4 possible values for the flag:
+//  0: we haven't looked up converstaions
+//  1: we looked, this mailbox doesn't have conversations
+//  2: we looked, someone else already opened the state, we cached a copy
+//  3: we opened the state and are responsible for closing it
```
rjbs, line 4409: Feels like this should be an enum.  `CSTATE_UNKNOWN` and `CSTATE_NONE` and `CSTATE_BORROWED` and `CSTATE_OWNED` mean never searching to look up what `3` means.

imap/mailbox.c
```diff
@@ -6772,13 +6792,13 @@ HIDDEN int mailbox_rename_nocopy(struct mailbox *oldmailbox,
...
-            struct conversations_state *tempcs = oldmailbox->local_cstate;
+            struct conversations_state *tempcs = oldmailbox->cstate_value;
             mbentry_t *tempmb = oldmailbox->mbentry;
-            oldmailbox->local_cstate = newcstate;
+            oldmailbox->cstate_value = newcstate;
             oldmailbox->mbentry = newmbentry;
             r = mailbox_add_conversations(oldmailbox, oldmailbox->silentchanges);
             oldmailbox->mbentry = tempmb;
-            oldmailbox->local_cstate = tempcs;
+            oldmailbox->cstate_value = tempcs;
```
rjbs, line 6801: I'm not 100% sure I follow this and will ask you out loud.

elliefm:
This _might_ have a side effect of also fixing that pop3d crash that we saw the other week.  As I recall, it asserted on something while trying to open the conversations state in order to do the logout expunge.

imap/mailbox.c
brong: I have done this.

imap/mailbox.c
brong: I explained this one out loud :)  The `cstate_flag` is guaranteed non-zero at this point by a call further above, so any calls to mailbox_get_cstate{,_full} inside the add_conversations above were guaranteed to read the value (even if NULL) of newcstate, and then we put the existing value (even if NULL) back.  This is actually safer for some weird configurations than before when we used presence of the pointer rather than a separate flag to know if we had already tried to load conversations.

PR #5499, opened 2025-07-04
cunit/ical_support: verify the last byte of http_cal_abook_admin.js is 0

dilyanpalauzov:
The content of http_cal_abook_admin.js  is used in list_addressbooks():
```c
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wformat-nonliteral"
    /* format string comes from http_cal_abook_admin_js.h */
    buf_printf(body, http_cal_abook_admin_js, CYRUS_VERSION, http_cal_abook_admin_js_len);
#pragma GCC diagnostic pop
```
In the past I have seen compiler warnings, when http_cal_abook_admin_js does not end with `\0`.

With this change, if http_cal_abook_admin.js does not end in `\0`, then `make check` fails.

PR #5507, opened 2025-07-11
build deps files for cunit tests

elliefm:
This addresses an issue observed in #5499, that we don't build deps files for our cunit tests.  This meant the cunit tests wouldn't necessarily recompile after things they were trying to test had changed.  The change updates the custom rule used for building tests' object files, which had been missing some details, to do the same things as the builtin rule used for normal object files.

It doesn't (yet) address the specific issue of ical_support.testc not recompiling after http_cal_abook_admin.js changes.  I'm still missing a detail somewhere... I've tried a few obvious things, and a few less obvious things, but so far none have made the difference I was looking for.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -695,7 +695,9 @@ endif
 
 .testc.o:
 	$(AM_V_at)$(CUNIT_PL) --generate-wrapper $<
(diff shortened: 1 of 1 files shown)


elliefm:
#5512 supersedes this

PR #5508, opened 2025-07-11
don't skip rights check when renaming intermediates

### LINKED ISSUE #5495: Check Bypass Vulnerability in mboxlist_renamemailbox
Hello, I have been conducting a security audit of the `cyrus-imapd` source code and believe I have discovered a security vulnerability.

### **Version**
`cyrus-imapd-3.13.0-alpha0-792-g947b9b5ff`

### **Summary**

An access control bypass vulnerability exists in the `mboxlist_renamemailbox()` function in `imap/mboxlist.c`. An authenticated user, without possessing the necessary 'delete' permission (`ACL_DELETEMBOX`) on a target mailbox, can successfully rename (effectively moving or deleting) a mailbox of type `MBTYPE_INTERMEDIATE`.

This vulnerability allows a low-privileged user to disrupt the shared mailbox hierarchy or move a shared folder into their private namespace, leading to a denial of service for other legitimate users.

### **Root Cause Analysis**

The root cause of this vulnerability lies within the control flow of the `mboxlist_renamemailbox()` function in `imap/mboxlist.c`. While the function correctly checks for `ACL_DELETEMBOX` permission on standard mailboxes, it contains a flawed special-case for mailboxes of type `MBTYPE_INTERMEDIATE`. When handling these intermediate mailboxes, the code enters a logic branch that, after checking the destination, executes a `goto dbupdate;` statement. This jump bypasses the crucial permission check that should have been performed on the source mailbox. The execution flow then proceeds to the `dbupdate:` label, which calls the lower-level `mboxlist_update_entry_full` function to perform the database write. Because this lower-level function implicitly trusts its caller to have already completed all security validations, the unauthorized rename operation is executed, creating a clear access control bypass.

### **Comparison with Similar Vulnerability & Proposed Fix**

I believe the underlying pattern of this bug—a special case bypassing a general security check—is logically similar to the flaw fixed in **CVE-2021-32056**.

  * In **CVE-2021-32056**, the permission check for server annotations was bypassed because it was located inside an `if (mailbox)` block.
  * In this new vulnerability, the permission check for intermediate mailboxes is bypassed because of a `goto` statement.

The essence of both vulnerabilities is the failure to enforce a universal security rule at a common checkpoint. Instead, the code branches into special-case logic too early, allowing the security policy to be circumvented.

Referencing the fix for CVE-2021-32056 (which enforces checks before conditional logic), I believe the following patch will correctly resolve this vulnerability:

```diff
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -58,6 +58,20 @@
 
     assert_namespacelocked(mbentry->name);
     assert_namespacelocked(newname);
+
+    /*
+     * IMPORTANT SECURITY CHECK
+     *
+     * This check must happen before any special-casing (like intermediate
+     * mailboxes) to prevent ACL bypass. Renaming a mailbox is conceptually
+     * a DELETE operation on the old name, thus requires 'd' rights.
+     */
+    myrights = cyrus_acl_myrights(auth_state, mbentry->acl);
+    if (!isadmin && !(myrights & ACL_DELETEMBOX)) {
+        r = (myrights & ACL_LOOKUP) ?
+            IMAP_PERMISSION_DENIED : IMAP_MAILBOX_NONEXISTENT;
+        return r;
+    }
 
     /* special-case: intermediate mailbox */
     if (mbentry->mbtype & MBTYPE_INTERMEDIATE) {
@@ -77,18 +91,6 @@
         goto dbupdate;
     }
 
-    myrights = cyrus_acl_myrights(auth_state, mbentry->acl);
-
-    /* check the ACLs up-front */
-    if (!isadmin) {
-        if (!(myrights & ACL_DELETEMBOX)) {
-            r = (myrights & ACL_LOOKUP) ?
-                IMAP_PERMISSION_DENIED : IMAP_MAILBOX_NONEXISTENT;
-            return r;
-        }
-    }
-
     /* 1. open mailbox */
     r = mailbox_open_iwl(oldname, &oldmailbox);
     if (r) return r;

```

This patch remediates the vulnerability by reordering the function's logic. It ensures that the mandatory `ACL_DELETEMBOX` permission check is performed unconditionally at the beginning of the function, before any special-case branching. This makes the security check impossible to bypass.

elliefm:
This updates mboxlist_renamemailbox to do the rights check before the special-case intermediates handling, based on the patch in #5495.  Also adds a way for Cassandane tests to create intermediate mailboxes, and a regression test for the bug.

Fixes #5495

### DIFF
--- a/imap/mboxlist.c
+++ b/imap/mboxlist.c
@@ -2735,6 +2735,17 @@ EXPORTED int mboxlist_renamemailbox(const mbentry_t *mbentry,
     assert_namespacelocked(mbentry->name);
     assert_namespacelocked(newname);
 
+    myrights = cyrus_acl_myrights(auth_state, mbentry->acl);
+
+    /* 0. check the ACLs up-front */
+    if (!isadmin) {
+        if (!(myrights & ACL_DELETEMBOX)) {
+            r = (myrights & ACL_LOOKUP) ?
+                IMAP_PERMISSION_DENIED : IMAP_MAILBOX_NONEXISTENT;
+            return r;
+        }
+    }
+
     /* special-case: intermediate mailbox */
     if (mbentry->mbtype & MBTYPE_INTERMEDIATE) {
         r = mboxlist_create_namecheck(newname, userid, auth_state,
@@ -2750,17 +2761,6 @@ EXPORTED int mboxlist_renamemailbox(const mbentry_t *mbentry,
 
         /* skip ahead to the database update */
         goto dbupdate;
-    }
-
-    myrights = cyrus_acl_myrights(auth_state, mbentry->acl);
-
-    /* check the ACLs up-front */
-    if (!isadmin) {
-        if (!(myrights & ACL_DELETEMBOX)) {
-            r = (myrights & ACL_LOOKUP) ?
-                IMAP_PERMISSION_DENIED : IMAP_MAILBOX_NONEXISTENT;
-            return r;
-        }
     }
 
     /* 1. open mailbox */


ksmurchison (approved):

elliefm:
Backporting: the commits rebase cleanly onto 3.12, but the new test throws an error while trying to force a mailbox into intermediate status.  I guess something is different from master and the setup will need to be massaged.

PR #5510, opened 2025-07-15
pop3d: use read-lock at start and then close and re-open mailbox to write

brong:
This avoids a lock inversion risk, means we only need a read-lock to slurp the initial data (because nothing ever writes then) and also means that we look up records by their UID rather than recno throughout.

### DIFF
--- a/imap/pop3d.c
+++ b/imap/pop3d.c
@@ -722,62 +722,23 @@ EXPORTED void fatal(const char* s, int code)
     shut_down(code);
 }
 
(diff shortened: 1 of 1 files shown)


brong:
@rjbs sorry this one took so long to write, I wrote test cases too to make sure the seen and delete stuff works still

rjbs (approved):
Through long, tedious reference to mailbox.c, I have come to believe this is correct.  ;)

PR #5511, opened 2025-07-16
fix Util::Words::random_word() to not repeat words

elliefm:
Looks like it was always intended to not repeat words (unless it ran out), but there were two bugs in that, causing occasional test failures like this:

```
1) Shared.subscribe_vd_prefix2
 Perl exception: Cannot create user.jim.perl@example.com: IMAP Command : 'create' failed. Response was : no - Mailbox already exists
 at Cassandane/Instance.pm line 1099.
```

### DIFF
--- a/cassandane/Cassandane/Util/Words.pm
+++ b/cassandane/Cassandane/Util/Words.pm
@@ -40,7 +40,7 @@
 package Cassandane::Util::Words;
 use strict;
 use warnings;
-use List::Util qw(uniq);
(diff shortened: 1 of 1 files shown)


rjbs (approved):

wolfsage (approved):
It means we end up with <2048 words likely, but that's okay I think. And we *might* repeat when we get close to the end / reshuffle, but I bet that's rare (or never happens?)

This is definitely an improvement, thanks

PR #5514, opened 2025-07-18
cass: pop3 tests: Wait for QUIT to finish its work

wolfsage:
... otherwise we may call status before pop3 has committed its changes and the test will fail.


elliefm (approved):

elliefm:
Does this fix Pop3.dele failing sometimes?

PR #5515, opened 2025-07-18
global.c: only start logging debug toggle after initial setup

rjbs:
Every single program that loads config with debug off will currently log "debug logging turned off", which feels sort of ironic.  Also, it's annoying for processes that might run *really* often, like `mbpath`.  Let's set the debug level quietly to whatever was configured, then log when it's toggled during runtime.

### DIFF
--- a/imap/global.c
+++ b/imap/global.c
@@ -259,11 +259,11 @@ EXPORTED int cyrus_init(const char *alt_config, const char *ident, unsigned flag
         openlog(config_ident, syslog_opts, SYSLOG_FACILITY);
     }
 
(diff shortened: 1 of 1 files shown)


rjbs:
Just a bit funnier than this:  In the previous behavior, we'd register the callback, then load the config.  Loading the config was a "toggle", so it logged.  That was what I intended to fix.  But *also*, we could toggle before the syslog prefix was set, which meant it might or might not go to the right place, depending on config file ordering?

Anyway, this will reduce noise.

ksmurchison (approved):
LGTM

PR #5520, opened 2025-07-22
bench/cyrdbbench.c: typo zeroskip -> twoskip

dilyanpalauzov:
Defect since 7aff6fc48b48bbe6b70c75463f2e.


elliefm:
Thanks.  Do you mind also updating the usage message while you're in there, around line 250 or so:

```
    printf("                       Available Cyrus DB's: twoskip\n");
```

It ought to say "twoskip, twom" or something instead of just "twoskip"

PR #5521, opened 2025-07-22
CalDAV: start and end calendar-user-address-set with preferred href

rsto:
This changes the PROPFIND return value for `calendar-user-address-set`. In addition to returning the default calendar user address at the *end* of the list of hrefs, this patch also sets a XML href node at the *start* of the list, effectively duplicating the preferred href in the value.

CalDAV clients interpret the value of `calendar-user-address-set` differently and Thunderbird recently changed its logic from picking the *last* entry to picking the *first*. This patch is meant to allow interoperating with both older and newer Thunderbird clients, as well as Apple and others, presuming that clients ignore any href other than the one they are hard-coded to pick in the list.

On a tangent, this PR changes the value format for the calendar-user-address-set CalDAV annotation. It reverts the format back to its version before the implementation of the defaultParticipantIdentityId of the (now obsolete) JMAP CalendarPreferences object. This PR adds regression tests for calendar-user-address-set, including the quirks specific to the Cyrus implementation.

### DIFF
--- a/cassandane/tiny-tests/Caldav/calendar_user_address_set_principal
+++ b/cassandane/tiny-tests/Caldav/calendar_user_address_set_principal
@@ -17,6 +17,7 @@ sub test_calendar_user_address_set_principal
         { href => 'mailto:bar@example.com' },
     ]);
     $self->assert_address_set([
+        { href => 'mailto:bar@example.com' },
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):
LGTM

PR #5522, opened 2025-07-23
jmap_email.c: properly allow /copy from a read-only mailbox

ksmurchison:


cassandane/tiny-tests/JMAPEmail/email_copy
```diff
@@ -177,6 +177,41 @@ sub test_email_copy
...
+    $res = $jmap->CallMethods([
+        ['Email/set', {
+            destroy => [ $copiedEmailId ]
+         }, 'R1'],
+    ]);
+
+    xlog $self, "share other user mailbox read-only";
+    $admintalk->setacl("user.other", "cassandane", "lrs") or die;
+
+    xlog $self, "move email back (again); copy will succeed, destroy will fail";
```
rsto: Is that a move? It looks like regular copy, since destroy will fail.

ksmurchison: I'm going to change the wording to "attempt to move"

PR #5525, opened 2025-07-28
dynarray append improvements

elliefm:
To use `dynarray_append()`, you need to already have the object you want to append to the array.  If you don't, you need to construct one in scratch space somewhere, so it can be copied into place by `dynarray_append()`, and then (shallowly) discard the scratch space copy.

This PR adds `dynarray_append_empty()`, which allocates a new, empty, slot in the array, and gives you back a pointer to it so you can construct its value in place.  For trivial objects the difference is academic, but for complex objects it avoids the memcpy and simplifies the memory management.

It also updates `dynarray_append()` to return the index of the new element.  This will let you know where the thing you added is, which will be useful for more complicated data models using dynarrays as backing store.  I've checked the existing callers and none of them benefit from this detail, so they haven't been updated to use it, they just ignore the return value.

The API for `dynarray_append_empty()` is a little funny, because `void **` does not benefit from the same implicit casting behaviours as `void *` does -- see the cunit test for examples of what it's like to call it.  I thought about making it just return the pointer instead of using an out parameter for it, but I also needed the index, and it felt better to return the index like `dynarray_append()` does.

I'll need this for #5317 eventually, but it stands alone, and it's simpler to review it separately.

### DIFF
--- a/cunit/dynarray.testc
+++ b/cunit/dynarray.testc
@@ -41,6 +41,37 @@ static void test_basic(void)
     CU_ASSERT_PTR_NULL(da);
 }
 
+static void test_append_empty(void)
+{
+    struct dynarray *da = dynarray_new(sizeof(uint32_t));
+    uint32_t *valp1, *valp2;
+    int index;
+
+    CU_ASSERT_EQUAL(dynarray_size(da), 0);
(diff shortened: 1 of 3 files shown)


rsto (approved):
Looks good to me, will come in handy!

PR #5526, opened 2025-07-28
lmtpd.c: don't fuzzy match mailboxes at a lower depth

ksmurchison:


rjbs:
I suspect we want one more case here:  If `INBOX.INBOX.exists` exists, delivery to `cassandane+inbox.inbox` should deliver to `INBOX` and not `INBOX.INBOX.`

rjbs (changes requested):
Sorry, I have to repeat myself to make GitHub happy.  I think we want a test about never delivering to INBOX.INBOX, even if it seems targeted.

imap/lmtpd.c
```diff
@@ -455,7 +455,7 @@ EXPORTED int fuzzy_match(mbname_t *mbname)
         int i;
         const strarray_t *newboxes = mbname_boxes(frock.result);
         mbname_truncate_boxes(mbname, 0);
-        for (i = 0; i < strarray_size(newboxes); i++)
+        for (i = 0; i < frock.depth; i++)
```
rsto, line 458: Why is this change necessary?

ksmurchison: Because without it, an address like cassandane+inbox would match mailbox user.cassandane.INBOX.foo, when user.cassandane.INBOX doesn't exist.  Basicially, we don't want to deliver to a deeper mailbox just because the plus address matches a prefix of that mailbox

rsto: OK, thanks

ksmurchison:
@rjbs I think I resolved your change request.  Need a re-review

ksmurchison:
@rjbs are you ok with this?

PR #5527, opened 2025-07-29
conversations.c: NFC-normalize conversion subjects

rsto:
This changes conversion subject normalization to use Unicode NFC normalization. Before, it left non-ASCII bytes in the subject as-is. This led to replies to emails with non-ASCII subjects sometimes not getting assigned to their threads.

This patch bumps the version of B records to 2. It preserves the former subject normalization code for backwards compatibility with existing on-disk conversation records of version 0 or 1.


ksmurchison (approved):
LGTM

PR #5529, opened 2025-07-29
mailbox.c, mboxlist.c: add jmapids to auditlog

ksmurchison:
Do not merge/deploy until FM tooling has been updated

### DIFF
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -2346,9 +2346,12 @@ static int _commit_one(struct mailbox *mailbox, struct index_change *change)
         struct conversations_state *cstate = mailbox_get_cstate(mailbox);
         char flagstr[FLAGMAPSTR_MAXLEN];
         char emailid[JMAP_MAX_EMAILID_SIZE];
+        char threadid[JMAP_THREADID_SIZE];
 
         jmap_set_emailid(cstate, &record->guid,
                          0, &record->internaldate, emailid);
+
+        jmap_set_threadid(cstate, record->cid, threadid);
 
         flags_to_str(record, flagstr);
         if (change->flags & CHANGE_ISAPPEND)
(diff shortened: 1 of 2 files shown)


wolfsage (approved):
This seems right, but from my understanding only logs jmap ids of mailboxes, not the new ids of messages in those mailboxes?

Before we could go from a Mabcd to a GUID but now that's harder - is the jmap id of a message ever logged anywhere?

imap/mailbox.c
```diff
@@ -7264,10 +7297,11 @@ HIDDEN int mailbox_rename_copy(struct mailbox *oldmailbox,
     if (r) goto fail;
 
     if (config_auditlog)
-        syslog(LOG_NOTICE, "auditlog: rename sessionid=<%s> "
-                           "oldmailbox=<%s> newmailbox=<%s> uniqueid=<%s>",
-                           session_id(),
-                           mailbox_name(oldmailbox), newname, mailbox_uniqueid(newmailbox));
+        xsyslog(LOG_NOTICE, "auditlog: rename",
+                "sessionid=<%s> oldmailbox=<%s> newmailbox=<%s> "
```
wolfsage: Most of the other spaces between joined lines are at the start of the following line, not the end of the current line - we should probably stick to one or the other?

cassandane/Cassandane/Cyrus/Delivery.pm
```diff
@@ -623,7 +623,7 @@ sub test_auditlog_size
     xlog $self, "Check the correct size was auditlogged";
     if ($self->{instance}->{have_syslog_replacement}) {
         my @appends = $self->{instance}->getsyslog(
-            qr/auditlog: append .* uid=<1>/);
+            qr/auditlog: append: .* uid=<1>/);
```
rsto, line 626: nit: is this going to break any existing log analysis tooling?

rjbs: We should drop this colon, it's not like any(?) other auditlog.

rjbs: Wait, this is weird.  Why is there a `:` here when there is no `:` in the actual append log?  I am confused.

rjbs: It's all clear to me now.  This is a function of moving from syslog to xsyslog.  I will now re-audit out log parsing code.

rjbs:
I added a commit (reviewed by @ksmurchison) to include the threadid prefix in the `cid=<...>` bits.

PR #5532, opened 2025-07-29
New user namespace locking code

brong:
This does a bunch of things:

- check that we never have a global annotations db lock while we're locking any user, or when we're dropping to read-only long mailbox actions
- adds a dual-user locking version of namespace locks, which always takes them in the right order (and checks that they're not already locked incorrectly!)
- updates JMAP callers to double-lock whenever fromAccountId is set, removing a bunch of special-case code for /copy operations
- makes more /get operations reliably read-only so they can only take SHARED locks

### DIFF
--- a/cassandane/Cassandane/Instance.pm
+++ b/cassandane/Cassandane/Instance.pm
@@ -2853,6 +2853,14 @@ sub install_sieve_script
     die "Symlink does not exist: $sieved/defaultbc" if not -l "$sieved/defaultbc";
 
     xlog "Sieve script installed successfully";
+
+    if ($params{upgrade}) {
+        my $srv = $self->get_service('imap');
+        my $adminstore = $srv->create_store(username => 'admin');
+        my $adminclient = $adminstore->get_client();
+        $adminclient->_imap_cmd('UPGRADESIEVE', 0, '', $user);
+        xlog "Sieve upgraded to mailbox version";
+    }
 }
 
 sub install_old_mailbox
--- a/cunit/annotate.testc
+++ b/cunit/annotate.testc
@@ -12,6 +12,7 @@
 #include "imap/annotate.h"
 #include "imap/append.h"
 #include "imap/mboxlist.h"
+#include "imap/user.h"
 #include "imap/imap_err.h"
 
 #define DBDIR           "test-dbdir"
@@ -2212,7 +2213,7 @@ static int set_up(void)
     mboxlist_init();
     mboxlist_open(NULL);
 
-    struct mboxlock *namespacelock = mboxname_usernamespacelock(MBOXNAME1_INT);
+    user_nslock_t *user_nslock = user_nslock_lockmb_w(MBOXNAME1_INT);
 
     /* XXX API abuse? perhaps this should just call mboxlist_create,
      * XXX rather than the low level APIs. */
@@ -2239,9 +2240,9 @@ static int set_up(void)
         return r;
     mailbox_close(&mailbox);
 
-    mboxname_release(&namespacelock);
+    user_nslock_release(&user_nslock);
 
-    namespacelock = mboxname_usernamespacelock(MBOXNAME2_INT);
+    user_nslock = user_nslock_lockmb_w(MBOXNAME2_INT);
 
     r = mailbox_create(MBOXNAME2_INT, /*mbtype*/0, /*version*/0, PARTITION, ACL,
                        makeuuid(), /*jmapid*/0,
@@ -2266,9 +2267,9 @@ static int set_up(void)
         return r;
     mailbox_close(&mailbox);
 
-    mboxname_release(&namespacelock);
+    user_nslock_release(&user_nslock);
 
-    namespacelock = mboxname_usernamespacelock(MBOXNAME3_INT);
+    user_nslock = user_nslock_lockmb_w(MBOXNAME3_INT);
 
     r = mailbox_create(MBOXNAME3_INT, /*mbtype*/0, /*version*/0, PARTITION, ACL,
                        makeuuid(), /*jmapid*/0,
@@ -2290,7 +2291,7 @@ static int set_up(void)
 
     mailbox_close(&mailbox);
 
-    mboxname_release(&namespacelock);
+    user_nslock_release(&user_nslock);
 
     old_annotation_definitions =
         imapopts[IMAPOPT_ANNOTATION_DEFINITIONS].val.s;
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -2162,7 +2162,7 @@ EXPORTED int caldav_alarm_process(time_t runtime, time_t *intervalp, int dryrun)
         int num_user_records = 0;
         int skip_user = 0;
         char *userid = NULL;
-        struct mboxlock *nslock = NULL;
+        user_nslock_t *user_nslock = NULL;
         for (i = 0; i < rock.list.count; i++) {
             struct caldav_alarm_data *data = ptrarray_nth(&rock.list, i);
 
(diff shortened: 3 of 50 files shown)


cunit/annotate.testc
```diff
@@ -2212,7 +2213,7 @@ static int set_up(void)
     mboxlist_init();
     mboxlist_open(NULL);
 
-    struct mboxlock *namespacelock = mboxname_usernamespacelock(MBOXNAME1_INT);
+    unslock_t *unslock = unslock_lockmb(MBOXNAME1_INT);
```
rjbs: I may be the odd one out, but all I can see here is `un-slock`.  If nothing else, a `_` might go a long way.

elliefm: Same, I'd like to see `uns_lock` or something.  `unslock` looks too much like `unlock`, perhaps the opposite of some hypothetical `slock`

rsto: While we are it, I would prefer this even less opaquely named:

`user_nslock` rather than `unslock` would go a long way with:
1. it's consistent with using the file name of the header that defines this function as a prefix (`user.h` in this case)
2. it's clearer to me to understand it's a user namespace lock when looking at the function name

ksmurchison: > While we are it, I would prefer this even less opaquely named:
> 
> `user_nslock` rather than `unslock` would go a long way with:
> 
>     1. it's consistent with using the file name of the header that defines this function as a prefix (`user.h` in this case)
> 
>     2. it's clearer to me to understand it's a user namespace lock when looking at the function name

Agreed

brong: Fair and legit, I don't love the existing name either, will fix.

rjbs (approved):
I am not the only person who should give this a +1, as my knowledge of this stuff is not sufficient for an expert review — but having been walked through it and having read it, I believe it is sensible.

brong:
I have fixed up the naming as requested, and walked Rik through a description, but additional reviews would be very appreciated!

imap/user.h
```diff
@@ -93,10 +100,16 @@ char *user_hash_xapian(const char *userid, const char *root);
 char *user_hash_xapian_byname(const mbname_t *mbname, const char *root);
 char *user_hash_xapian_byid(const char *mboxid, const char *root);
 
+user_nslock_t *user_nslock_full1(const char *a, int locktype);
```
rsto: Rename `a` to `userid`.

imap/user.h
```diff
@@ -93,10 +100,16 @@ char *user_hash_xapian(const char *userid, const char *root);
...
+user_nslock_t *user_nslock_full2(const char *a, const char *b, int locktype);
```
rsto: Rename `a` and `b` to `userid1` and `userid2` (or similarly). The implementation already names these `userida` and `useridb`. Personally, I find the missing underscore between the letters hard to read, but we aren't consistent with that in the codebase (and neither am I!).

imap/user.h
rsto: How about renaming `user_nslock_full1` to `user_nslock_lock` and renaming the existing `user_nslock_lock` macro to either `user_nslock_lock_w` or `user_nslock_lock_excl`?

imap/user.h
rsto: How about renaming `user_nslock_full2` to `user_nslock_lock2`?

imap/user.h
```diff
@@ -93,10 +100,16 @@ char *user_hash_xapian(const char *userid, const char *root);
...
+user_nslock_t *user_nslock_fullmb(const char *mba, const char *mbb, int locktype);
```
rsto: The implementation names this`mboxnamea` and `mboxnameb` which is more obvious. Without having looked at the code, is this always going to be a mailbox name, could it also be a mailbox uuid? In this case, should these arguments rather be mbentry?

cassandane/Cassandane/Instance.pm
```diff
@@ -2853,6 +2853,14 @@ sub install_sieve_script
     die "Symlink does not exist: $sieved/defaultbc" if not -l "$sieved/defaultbc";
 
     xlog "Sieve script installed successfully";
```
rsto, line 2855: Does this commit really belong to that pull request? Is it related to the lock changes?

imap/imapd.c
```diff
@@ -14102,6 +14113,20 @@ static int reset_saslconn(sasl_conn_t **conn)
     return SASL_OK;
 }
 
+static void cmd_upgradesieve(char *tag, char *userid)
```
rsto, line 14116: Also, is this related to the lock API changes?

imap/jmap_sieve.c
```diff
@@ -307,11 +307,19 @@ static int jmap_sieve_get(jmap_req_t *req)
         goto done;
     }
 
-    r = sieve_ensure_folder(req->accountid, &mailbox, /*silent*/0);
+    char *mboxname = sieve_mboxname(req->accountid);
```
rsto, line 310: Also: is this related to the lock changes?

imap/caldav_alarm.c
```diff
@@ -2179,20 +2179,20 @@ EXPORTED int caldav_alarm_process(time_t runtime, time_t *intervalp, int dryrun)
...
-                nslock = user_namespacelock_full(userid, LOCK_NONBLOCKING);
+                user_nslock = user_nslock_full1(userid, LOCK_NONBLOCKING);
             }
             mbname_free(&mbname);
 
             if (skip_user) continue;
 
             // if we failed to lock the user, or have done too many for this user, skip
-            if (!nslock || ++num_user_records > MAX_CONSECUTIVE_ALARMS_PER_USER) {
+            if (user_nslock || ++num_user_records > MAX_CONSECUTIVE_ALARMS_PER_USER) {
```
rsto: Did we loose the negation here, should this be `!user_nslock`?

imap/jmap_blob.c
```diff
@@ -285,28 +283,6 @@ static int jmap_blob_copy(jmap_req_t *req)
...
-    // now we can open the cstate
-    r = conversations_open_user(req->accountid, 0, &req->cstate);
-    if (r) {
-        syslog(LOG_ERR, "jmap_email_copy: can't open converstaions: %s",
-                        error_message(r));
-        jmap_error(req, jmap_server_error(r));
-        goto cleanup;
-    }
-
     /* Check if we can upload to toAccountId */
```
rsto: While we are at it, we could get rid of this stale comment. There was an `toAccountId` argument in early drafts of  Blob/copy, but the final API uses `accountId` and `fromAccountId`

rsto (commented):
Generally, this looks good.
I only found one potential logic error. See my comment for the `caldav_alarm_process` function.
I also have some question/suggestions about naming the new user_nslock API.
I am not sure how the Sieve upgrade code relates to the user_nslock API changes.
Lastly, I think the commit messages could be clarified:
- `b780fcb: annotate: way to check for locks`: I don't know what this means.
- `48bbbc4: Also changes some detail in JMAP locking`: this is vague. Either remove it from the commit message or describe the changes.

 Before approving, here are my comments on
-  naming the new user_nslock API.
- a potential logic error in `caldav_alarm_process`

imap/jmap_admin.c
```diff
@@ -185,15 +184,6 @@ static int rewrite_calevent_privacy(const char *userid, void *vrock)
...
 
-    namespacelock = user_namespacelock(userid);
-    if (!namespacelock) {
-        json_t *err = jmap_server_error(IMAP_INTERNAL);
-        json_object_set_new(err, "description",
-                json_string("can not lock namespace"));
-        json_object_set_new(rock->not_rewritten, userid, err);
-        goto done;
-    }
-
```
ksmurchison, line 196: Do we not need locking here?

ksmurchison (commented):
I added one comment about some code where a lock was removed.

Also, I second most, if not all, of rsto's comments.

imap/caldav_alarm.c
brong: Funny enough, guess what!  Calalarmd failed to process anything because... this negation was lost!

imap/user.h
brong: Yes, that's consistent with the existing locking APIs; that's reasonable.

imap/user.h
brong: They're always mboxnames, never UUIDs.

Often we don't have an entry (though sometimes we do).  We can revisit this as part of making more things take an mbentry in general.

imap/imapd.c
brong: Yes - because we switched the JMAP API to use wrapping locks, and hence we need to know if /get will need a write lock. More specifically, we never want it to.  But fricking Cassandane creates sieve scripts by putting them directly on the filesystem, and a bunch of tests then assumed that a Sieve/get would upgrade the mailbox (because it used to).  So this moves the upgrade out into a command so otherwise reasonable Cassandane tests can keep running.

I don't love it, but I also didn't see another way short of adding a whole new Cassandane sieve handler.

imap/jmap_blob.c
brong: changed the text

imap/jmap_admin.c
brong: Actually quite the opposite, I think for these obnoxiously cross-user functions, we need to not take the global lock at all!  Interesting catch.

The mailbox opening operations also take user locks, which is all we need in this case because we're not trying to create a state object for which the user_nslock would protect us from concurrent changes.

ksmurchison (approved):

PR #5535, opened 2025-07-30
imapoptions: remove sortcache_db

rjbs:
This was only used by XCONVMULTISORT, which is gone.

This also removes a test that was looking for a bug in XCONVMULTISORT and Cassandane support for sortcache_db.


lib/imapoptions
```diff
@@ -2846,10 +2846,6 @@ product version in the capabilities
    successfully authenticate.  Otherwise lmtpd returns permanent failures
    (causing the mail to bounce immediately). */
 
-{ "sortcache_db", "twoskip", STRINGLIST("skiplist", "twom", "twoskip"), "UNRELEASED" }
-/* The cyrusdb backend to use for caching sort results (currently only
-   used for xconvmultisort). */
-
```
elliefm: You need to mark this option as deprecated, not delete it.  You do that by adding another version field, also containing "UNRELEASED", after the first one, like this:

```
- { "sortcache_db", "twoskip", STRINGLIST("skiplist", "twom", "twoskip"), "UNRELEASED" }
+ { "sortcache_db", "twoskip", STRINGLIST("skiplist", "twom", "twoskip"), "UNRELEASED", "UNRELEASED" }
```

and replace the description in the comment with something saying it's deprecated, look around in the file for an example.

If this weren't already marked "UNRELEASED" due to the "twom" change previously, you would still update both fields to "UNRELEASED", since the act of deprecating it is also changing it.

Marking it as deprecated means that conf lint etc can warn or complain about it properly.

rjbs: Done!  Also, I got config2header in my eyes!

elliefm (approved):

PR #5536, opened 2025-07-30
charset.c: discard empty escape sequences from ISO-2022-JP text

rsto:
This preprocesses text encoded in ISO-2022-JP before handing it over to libicu for conversion. The character encoding is defined in RFC 1468 and contains segments of both one-byte and two-byte character sequences. A segment starts with an escape sequence that indicates the encoding of the following text bytes.

Escape sequences must be followed by at least one text character, but some MIME messages violate this and instead concatenate multiple escape sequences. For every such sequence, libicu inserts a REPLACEMENT character. This is bad user experience and other character encoding libraries, such as the one for Java, Go or the iconv utility ignore such empty sequences.

This patch tweaks the charset pipeline by discarding all but the last in a run of escape sequences before passing the bytes to the libicu-based charset converters. As a result, the converted text does not contain REPLACEMENT characters, as if those empty escape sequences never were part of the byte stream.

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_get_bodyvalues_iso2022jp
+++ b/cassandane/tiny-tests/JMAPEmail/email_get_bodyvalues_iso2022jp
@@ -1,0 +1,59 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_get_bodyvalues_iso2022jp
+{
+    my ($self) = @_;
+    my $jmap = $self->{jmap};
+    my $imap = $self->{store}->get_client();
+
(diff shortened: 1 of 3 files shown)


rsto:
> I'm not sure how readily this code can be made more generic, given the difference even in _length_ of shift code. It may never be needed, but I thought I'd call out the related possible problem here.

Thanks for pointing this out. I don't know if any of these encoding actually are used in MIME messages, as opposed to ISO-2022-JP where we know for sure. I'd rather wait tweaking our charset code for these encodings until we know that this is required.

PR #5538, opened 2025-07-31
Add test showing moving a message before 19 -> 20 breaks J keys

wolfsage:
You'll get 2 J keys for the same blob - one for the expunged message!


wolfsage:
This is in https://github.com/cyrusimap/cyrus-imapd/pull/5539 now

PR #5539, opened 2025-07-31
Internaldate and j keys

brong:


lib/util.h
```diff
@@ -158,6 +158,7 @@ extern const unsigned char convert_to_uppercase[256];
 #define TIMESPEC_FROM_NANOSEC(ts, nanosec) {    \
         (ts)->tv_sec  = (nanosec) / 1000000000; \
         (ts)->tv_nsec = (nanosec) % 1000000000; \
+        if (!(ts)->tv_nsec) (ts)->tv_nsec = UTIME_OMIT; \
```
wolfsage: 0 nanoseconds is a plausible value (unless we prevent it) - won't that mean we'll incorrectly use UTIME_OMIT and cause strange bugs?

imap/mailbox.c
```diff
@@ -5510,31 +5510,22 @@ struct find_dup_rock {
...
-
-        if (mbtype_isa(mbentry->mbtype) == MBTYPE_EMAIL) {
-            // found a non-expunged duplicate email; use its internaldate
-            struct timespec internaldate;
-            TIMESPEC_FROM_NANOSEC(&internaldate, rec->nano_internaldate);
-            if (internaldate.tv_nsec < UTIME_OMIT) {
-                frock->internaldate = internaldate;
-                ret = CYRUSDB_DONE;
-            }
+        !(rec->foldernum == frock->foldernum && rec->uid == frock->uid)) {
```
wolfsage: I thought MBTYPE_EMAIL was here to prevent matching uploaded blobs? If not that, why *was* it here?

lib/util.h
brong: It's pretty unlikely!  We might need to just ensure we never write a zero either though.  This is certainly the less absolutely necessary component here.

imap/mailbox.c
brong: Just an assumption that we don't need S (email_id) IDs for anything else I think, but this wasn't consistent throughout.  We still created them inside conversations.c, and the logic for various paths into conversations.db and detecting GUID matches was inconsistent.

lib/util.h
```diff
@@ -153,7 +153,7 @@ extern const unsigned char convert_to_uppercase[256];
 #endif
 
 #define TIMESPEC_TO_NANOSEC(ts)                                             \
-        ((uint64_t) (ts)->tv_sec * 1000000000 + (ts)->tv_nsec)
+        ((uint64_t) (ts)->tv_sec * 1000000000 + ((ts)->tv_nsec > 1000000000 ? 0 : (ts)->tv_nsec))
```
rjbs: imho, worth a comment or note in commit message "values over 1e9 should never happen, and if they do, they're probably UTIME_OMIT or other magic value"

lib/util.h
rjbs: Messages upgraded from disk use the top X bits of the digest, so can be some approximation of 1e9, so it's a 1-in-a-billion chance.  Messages appended from IMAP use the ns fraction of a realtime clock reading, which I believe has only microsecond accuracy.  Statistically, we'll see a perfect zero once in a million deliveries.

We have multiples of that many deliveries per day.  I think we should bump 0ns to 1ns (or some similar nudge).  It won't increase conflicts, because inside a user we surely(?) can't deliver more than 1 message per ns.  On the other hand, it'll eliminate the semi-predicate problem here.

imap/mailbox.c
```diff
@@ -5715,104 +5698,18 @@ static int _mailbox_index_repack(struct mailbox *mailbox,
...
-                    buf_printf(&buf, UINT64_FMT, copyrecord.internaldate.tv_nsec);
-                    repack->crcs.annot ^=
-                        crc_annot(record->uid,
-                                  IMAP_ANNOT_NS "internaldate.nsec", "", &buf);
-                }
+                copyrecord.internaldate.tv_nsec = nsec;
             }
             buf_reset(&buf);
             r = annotate_state_writesilent(astate, IMAP_ANNOT_NS "internaldate.nsec", "", &buf);
-            if (r) goto done;
```
wolfsage: Why can we safely ignore this error?

imap/mailbox.c
```diff
@@ -5715,104 +5698,18 @@ static int _mailbox_index_repack(struct mailbox *mailbox,
...
-
-                if (mailbox->i.minor_version < 20) {
-                    // add virtual annotation to old crc
-                    buf_reset(&buf);
-                    buf_printf(&buf, UINT64_FMT, copyrecord.internaldate.tv_nsec);
-                    repack->crcs.annot ^=
-                        crc_annot(record->uid,
-                                  IMAP_ANNOT_NS "internaldate.nsec", "", &buf);
-                }
+                copyrecord.internaldate.tv_nsec = nsec;
```
rjbs, line 5680: If the annotation contained a 0, we will end up with tv_nsec 0.  Later, under "calculate nanoseconds for internaldate if not yet present or re-calculating" comment, we only compute a new tv_nsec if we saw UTIME_OMIT.

So:

1. is there any place where we set tv_nsec on unannotated messages to (say) UTIME_OMIT or 0?  
2. should we set/leave it OMIT if the annotation contains 0?

lib/util.h
wolfsage: I'm *assuming* the only case this would be > 0 is for UTIME_OMIT?

But that's not portable! *BSD like systems (mac, etc) at least  use negative values for tv_nsec UTIME_OMIT, so this will be horribly wrong

imap/mailbox.c
```diff
@@ -5510,31 +5510,22 @@ struct find_dup_rock {
...
-            TIMESPEC_FROM_NANOSEC(&internaldate, rec->nano_internaldate);
-            if (internaldate.tv_nsec < UTIME_OMIT) {
-                frock->internaldate = internaldate;
-                ret = CYRUSDB_DONE;
-            }
+        !(rec->foldernum == frock->foldernum && rec->uid == frock->uid)) {
+        // found a duplicate email; use its internaldate
+        struct timespec internaldate;
+        TIMESPEC_FROM_NANOSEC(&internaldate, rec->nano_internaldate);
+        if (internaldate.tv_nsec < UTIME_OMIT) {
```
wolfsage: This comparison isn't portable - UTIME_OMIT can be negative on bsd-like systems and maybe other flavors of linux

lib/util.h
rjbs: After alh pointed out that UTIME_OMIT has worse (imho) value elsewhere, we could/should write SAFE_NS, which is `0 < x < 1e9` and it'll sort out zeroes and things like negative UTIME_OMIT at one go.

wolfsage:
Looks like https://github.com/cyrusimap/cyrus-imapd/pull/5539/commits/cc77383e606c783a3526801fd6a264a88d3a1de9 should be dropped and rolled into 36f033ec1d1 since they both change that line and the first no longer has any meaning (and is confusing to boot!)

Beyond that, it would be great if you went through each commit message and expanded on them.

For example - https://github.com/cyrusimap/cyrus-imapd/pull/5539/commits/621c0a8a3307c11c6e21119bb4bc21a715364df6 says "mailbox: fix basecid to match between v19 and v20" but adds this:
```
 /* but then if it's identical to CID, zero it */
    if (record->basecid == record->cid)
        record->basecid = 0;
    if (!record->basecid)
        record->internal_flags &= ~FLAG_INTERNAL_SPLITCONVERSATION;
```

And I don't really see any explanation or understand the rational for zeroing out identical cids, or how this achieves the goal of the commit message.

It'd be great if all of the commits had some discussion about what's changing, why it's changing, how it achieves the goal of the commit short desc, etc...

imap/mailbox.c
brong: yeah, that was a copy-paste error, fixed.

lib/util.h
brong: Right: what we want is a tristate:
```
0 => 1
1-999999999 => self
else => 0
```

And then on the read end (as it already does):

```
0 => omit
else => self
```

imap/mailbox.c
brong: we'll make this a macro

brong:
Rebased and pushed - there's a bunch of significant changes, so probably worth reading from the start - I tried to make the commits small and sane though the big one that changes how conversations_adjust_internaldate works and is called is still chunky.

One change, I totally removed UTIME_OMIT.  It meant the same as 0 to every place it was called, so why not just use 0 instead, particularly since that make utimes work.

imap/conversations.c
```diff
@@ -3494,12 +3544,13 @@ EXPORTED void conversations_adjust_internaldate(struct conversations_state *csta
                 buf_cstring(&jidrep), my_guid, existing_guid);
 
         // try the next nanosecond */
-        internaldate->tv_nsec = (internaldate->tv_nsec + 1) % 1000000000;
+        internaldate->tv_nsec++;
```
ksmurchison, line 3585: This seems wrong.  If the 29 bits of GUID gets us close to 999999999 and after a few iterations we increment to 1000000000, we will hit the check at the top of the loop and start with 29 bits of GUID again.  I think we want to set tv_nsec=1 if we get to 1000000000.

imap/jmap_util.c
```diff
@@ -1312,20 +1312,20 @@ EXPORTED void jmap_set_emailid(struct conversations_state *cstate,
                                uint64_t nanosec, struct timespec *ts,
                                char *emailid)
 {
-    // initialize a struct buf with char emailid[JMAP_MAX_EMAILID_SIZE]
-    struct buf buf = { emailid, 0, JMAP_MAX_EMAILID_SIZE, 0 };
+    // always zero out the entire emailid buffer
```
ksmurchison: What is actually wrong with the existing code?

imap/conversations.c
rjbs: 2**29-1 is 536,870,911.  The loop will give up after 1,000 iterations, but could go for 400,000,000+ before capping 1e9 ns.  I think this is safe?

ksmurchison: Yeah, right.  I stand corrected.

imap/jmap_util.c
brong: I now pass it to hashset_add:
```        /* Have we seen this message already? */
        if (!hashset_add(rrock->seen_emails, emailid))
            continue;
```
OOH, should update this too:

```
rrock->seen_emails = hashset_new(MESSAGE_GUID_SIZE);
```
That needs to be JMAP_MAX_EMAILID_SIZE

Anyway - we were just zeroing out the final byte, so the hashset could be getting uninitialised data.

imap/mailbox.c
```diff
@@ -5728,19 +5728,20 @@ static int _mailbox_index_repack(struct mailbox *mailbox,
...
-            repack->crcs.basic ^= crc_basic(&repack->newmailbox, &copyrecord);
-            repack->crcs.annot ^= crc_virtannot(&repack->newmailbox, &oldrecord);
-            repack->crcs.annot ^= crc_virtannot(&repack->newmailbox, &copyrecord);
-
-            // preload cache so we don't load it on newmailbox
-            r = mailbox_cacherecord(mailbox, &copyrecord);
-            if (r) goto done;
+            int dirty = oldrecord.internaldate.tv_sec != copyrecord.internaldate.tv_sec
+                     || oldrecord.internaldate.tv_nsec != copyrecord.internaldate.tv_nsec;
+
```
ksmurchison, line 5728: Couldn't this also do struct->struct comparison like 
https://github.com/cyrusimap/cyrus-imapd/pull/5539/commits/2af6374c1914ab0744889052ba7fca802ab35bd2

imap/mailbox.c
```diff
@@ -5738,6 +5738,10 @@ static int _mailbox_index_repack(struct mailbox *mailbox,
                 repack->crcs.basic ^= crc_basic(&repack->newmailbox, &copyrecord);
                 repack->crcs.annot ^= crc_virtannot(&repack->newmailbox, &oldrecord);
                 repack->crcs.annot ^= crc_virtannot(&repack->newmailbox, &copyrecord);
+                repack->newmailbox.i.synccrcs.basic ^= crc_basic(&repack->newmailbox, &oldrecord);
+                repack->newmailbox.i.synccrcs.basic ^= crc_basic(&repack->newmailbox, &copyrecord);
+                repack->newmailbox.i.synccrcs.annot ^= crc_virtannot(&repack->newmailbox, &oldrecord);
+                repack->newmailbox.i.synccrcs.annot ^= crc_virtannot(&repack->newmailbox, &copyrecord);
             }
```
ksmurchison, line 5752: Do we actually want to calculate all 4 of these CRCs twice, or should we assign them to a variable and then XOR that value against each of the two destination values?

ksmurchison (approved):
Approved, but noted a couple of nits.

imap/mailbox.c
brong: Yeah, we could do that!  In fact, we only need to calculate the CRC at all if that part has changed, so I've made it more efficient for every case.

imap/mailbox.c
brong: Actually, I think the mistake was removing the nsec check on reconstruct.  I removed it for bugs but I think they're all solved, I'll add it back.

You can't just == a struct.  You can copy it with =, but comparison would have to be `memcmp(a, b, sizeof(struct))` which is just fugly.  At least in non-C++.

PR #5543, opened 2025-08-04
Cyr 2034 emailsubmission ids downgrade

ksmurchison:


imap/jmap_mail_submission.c
```diff
@@ -969,31 +969,60 @@ static json_t *fetch_submission(jmap_req_t *req, message_t *msg)
...
+                        json_object_set_new(onsend, "moveToMailboxId",
+                                            json_string(mboxid));
+                        mboxlist_entry_free(&mbentry);
+                    }
+                }
+            }
+        }
+        else if (*id == JMAP_EMAILID_PREFIX) {
+            /* Rewrite to use GUID-based emailId */
+            char guidrep[2*MESSAGE_GUID_SIZE+2] =
```
wolfsage, line 1004: I feel like this is the first +2 I've seen. What's the extra byte space for?

wolfsage: Oh, we don't usually use this to store it with the M prefix? I get it.

ksmurchison: For the prefix

imap/jmap_mail_submission.c
```diff
@@ -969,31 +969,60 @@ static json_t *fetch_submission(jmap_req_t *req, message_t *msg)
...
+            r = conversations_jmapid_guidrep_lookup(req->cstate,
+                                                    id + 1, guidrep + 1);
             if (!r) {
-                char emailid[JMAP_MAX_EMAILID_SIZE];
-                jmap_set_emailid(req->cstate, NULL,
-                             internaldate, NULL, emailid);
-                json_object_set_new(sub, "emailId", json_string(emailid));
+                json_object_set_new(sub, "emailId",
+                                    json_pack("s%", guidrep,
+                                              JMAP_LEGACY_EMAILID_SIZE-1));
```
wolfsage, line 1012: Could json_pack("s%", guidrep, JMAP_LEGACY_EMAIL_ID_SIZE-1) instead be json_string(guidrep)?

ksmurchison: No, because we only use the first 24 chars of the GUID, not all 40.  I made that mistake myself and had to adjust

wolfsage: Ha! Right, thanks

rsto (approved):

PR #5544, opened 2025-08-05
Twom cursor speed

brong:


lib/twom.c
```diff
@@ -2799,6 +2812,10 @@ int twom_txn_begin_cursor(struct twom_txn *txn,
     struct twom_cursor *cur = (struct twom_cursor *)twom_zmalloc(sizeof(struct twom_cursor));
     cur->txn = txn;
     if (flags & TWOM_ALWAYSYIELD) cur->alwaysyield = 1;
+    if ((flags & TWOM_CURSOR_PREFIX) && prefix && prefixlen) {
+        cur->prefix = strdup(prefix);
+        if (cur->prefix) cur->prefixlen = prefixlen;
```
ksmurchison: if (cur->prefix) seems redundant since we already know its non-NULL

brong: we used strdup - which isn't xstrdup, because twom doesn't have all the nice stuff

ksmurchison: Oh, so you're afraid that strdup() might fail.  OK

lib/twom.c
```diff
@@ -2713,6 +2715,15 @@ int twom_cursor_next(struct twom_cursor *cur,
         if (!ptr) return TWOM_IOERROR;
     }
 
+    // have we left our prefix?
+    if (cur->prefixlen) {
+        size_t keylen = KEYLEN(ptr);
+        if (keylen < cur->prefixlen) return TWOM_DONE;
+        const char *key = KEYPTR(ptr);
+        if (txn->file->compar(key, cur->prefixlen, cur->prefix, cur->prefixlen))
+            return TWOM_DONE;
```
ksmurchison, line 2724: Shouldn't this be txn->file->compar(key, keylen, cur->prefix, cur->prefixlen))

brong: No - we're only comparing the substring which is the prefix (see the removed code below)

PR #5545, opened 2025-08-05
twom: don't exit on first DELETED pointer

brong:


rjbs (approved):

brong:
I have merged it, I'll do tests later - easy to add to cunit, but I want upstream to not be broken

PR #5551, opened 2025-08-06
cassandane: use twom for most db backends

rjbs:
...and make it easy to test with twoskip (or something else) instead

### DIFF
--- a/cassandane/Cassandane/Config.pm
+++ b/cassandane/Cassandane/Config.pm
@@ -96,6 +96,8 @@ sub new
 
 sub default
 {
+    my $default_backend = $ENV{CASSANDANE_DEFAULT_DB} // 'twom';
+
     if (!defined($default)) {
         $default = Cassandane::Config->new(
             admins => 'admin mailproxy mupduser repluser',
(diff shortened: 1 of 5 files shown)


cassandane/Cassandane/Config.pm
```diff
@@ -121,24 +123,27 @@ sub default
...
+            sync_cache_db => $default_backend,
+            tlscache_db => 'twoskip', # deprecated, does not allow twom
+            tls_sessions_db => $default_backend,
             userdeny_db => 'flat',
-            zoneinfo_db => 'twoskip',
+            zoneinfo_db => $default_backend,
+
+            # This, because some tests include tarballs of state which might
+            # not be in $default_backend format. -- rjbs, 2025-08-06
+            cyrusdb_autoconvert => 'yes',
```
elliefm: Not sure I love enabling autoconvert.  Though the fact you apparently haven't had to update any tests to be aware that autoconvert is now on-by-default suggests we might be lacking tests around autoconvert...

Rather than enabling autoconvert, I think I would update the tests that use those tarballs to be explicit about the db format they expect.  Have a look at some tests with the :NoStartInstances attribute for examples, since the most common use of that attribute is so the individual test can override config before stuff starts up.  But if there's whole suites using those tarballs, just override the config in `new`, not each test.

elliefm (approved):

PR #5552, opened 2025-08-06
Cassandane: put a "description" file in most instance base dirs

rjbs:
After a test run, sometimes I want to very quickly find the main instance for the Foo.bar_baz test.  This is a bit of a pain.

This commit should make it a bit less of a pain.  When the test represented by &Cassandane::Cyrus::Foo::test_bar_baz is run, the /tmp/cass/XXXXX directory for its main instance will now have a `description` file that contains "Foo.bar_baz".

We can improve upon this in the future, but this should help to start.

### DIFF
--- a/cassandane/Cassandane/Cyrus/TestCase.pm
+++ b/cassandane/Cassandane/Cyrus/TestCase.pm
@@ -668,7 +668,10 @@ sub _create_instances
         $instance_params{old_jmap_ids} = $self->{old_jmap_ids}
             if exists $self->{old_jmap_ids};
 
-        $instance_params{description} = "main instance for test $self->{_name}";
+        my $class = ref $self;
+        my $name  = $self->{_name} =~ s/^test_//r;
+        $instance_params{description} = "main instance for test $class.$name";
+
         $self->{instance} = Cassandane::Instance->new(%instance_params);
         $self->{instance}->add_services(@{$want->{services}});
         $self->{instance}->_setup_for_deliver()
@@ -699,7 +702,9 @@ sub _create_instances
                 $replica_params{installation} = 'replica';
             }
 
-            $replica_params{description} = "replica instance for test $self->{_name}";
+            my $class = ref $self;
+            my $name  = $self->{_name} =~ s/^test_//r;
+            $replica_params{description} = "replica instance for test $class.$name";
             $self->{replica} = Cassandane::Instance->new(%replica_params,
                                                          setup_mailbox => 0);
             my ($v) = Cassandane::Instance->get_version($replica_params{installation});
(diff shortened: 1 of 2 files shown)


cassandane/Cassandane/Cyrus/TestCase.pm
```diff
@@ -668,7 +668,10 @@ sub _create_instances
         $instance_params{old_jmap_ids} = $self->{old_jmap_ids}
             if exists $self->{old_jmap_ids};
 
-        $instance_params{description} = "main instance for test $self->{_name}";
+        my $class = ref $self;
+        my $name  = $self->{_name} =~ s/^test_//r;
+        $instance_params{description} = "main instance for test $class.$name";
+
```
elliefm, line 674: Could you update the other places that set `$foo_params{description}` to also use the same `$class.$name` logic please.  I expect you'll find one for replica and a couple for murder

cassandane/Cassandane/Instance.pm
```diff
@@ -336,7 +338,9 @@ sub _make_unique_instance_info
         last if mkdir($basedir);
         die "Cannot create $basedir: $!" if ($! != EEXIST);
     }
-    return _make_instance_info($name, $basedir);
+    my $info = _make_instance_info($name, $basedir);
+
+    return $info;
```
elliefm: What difference does this make?  Is it just housekeeping?

cassandane/Cassandane/Instance.pm
```diff
@@ -720,10 +725,21 @@ sub _build_skeleton
...
+    if ($self->{description}) {
+      open my $desc_fh, '>', "$basedir/description"
+        or die "Can't write to $basedir/description: $!";
+
+      $desc_fh->say($self->{description})
+        or die "Error writing to $basedir/description: $!";
+
+      close $desc_fh
+        or die "Error closing $basedir/description: $!";
+    }
```
elliefm: nit: we use 4-space indents, not 2

wolfsage (approved):
+1 to ellie's comments

Thanks for this!

elliefm:
For what it's worth, if you put "cleanup = yes" in the `[cassandane]` section of your cassandane.ini, then cassandane will clean up all the basedirs from successful tests after it finishes.  Only the basedirs from failed tests will be left behind.  When you're actively debugging something by running one test that fails, you don't need to find its basedirs amongst a sea of others, they'll be the only ones there.

If you set this, you can invoke testrunner with `--no-cleanup` to turn it off for one run, perhaps because you want to examine the basedirs of a test that unexpectedly succeeds.

cassandane/Cassandane/Cyrus/TestCase.pm
rjbs: Done!

cassandane/Cassandane/Instance.pm
rjbs: I reverted this locally.  It's historical nonsense.  When I was working on this, I thought I had to put some code in here to munge $info, but it turns out that this is the code I needed to put into `_build_skeleton`.

cassandane/Cassandane/Instance.pm
rjbs: This is going to be the death of me.

elliefm (approved):

PR #5554, opened 2025-08-07
email_legacy_ids: add Thread/get request

ksmurchison:
This test will fail until #5553 lands on master


brong (approved):
Yeah, good to have a test for this!

PR #5555, opened 2025-08-07
imapd.c: release index before sending COPY response

ksmurchison:
If we don't release the lock, mboxlist_createmailboxcheck() will assert() when calling conversations_open_mbox()


brong (approved):
Good!  We always want to unlock before replying anyway :)

PR #5557, opened 2025-08-07
Dav lock move

brong:
One more assertion failure found in FM production.  Needs to be an event with per-user alarms (or other per-user properties) so the annotations database is opened before the second user is locked for us to catch it reliably, but this was a lock inversion


ksmurchison (approved):
Approved pending passing CI

PR #5559, opened 2025-08-07
logging: log good and bad logins with logfmt

rjbs:
This is the first big push toward logfmt-based logging, chosen not because it's the most important thing to convert, but because it was a fairly wide-ranging update with a limited subject matter.

One next step is to write down what a "login.good" and "login.bad" event should look like and stick it in notes.  Even without that, this makes logins easy to machine-parse.

Special note, the loss of `popd_apop_chal` in popd login failure logging.  I think this is acceptable, largely because I don't know what anybody would want to do with the challenge in their logs!

### DIFF
--- a/cassandane/Cassandane/Cyrus/Simple.pm
+++ b/cassandane/Cassandane/Cyrus/Simple.pm
@@ -231,9 +231,9 @@ sub test_toggleable_debug_logging
 
     # find our imapd pid from syslog
     my $loginpat = qr{
-        \bimap\[(\d+)\]:\slogin:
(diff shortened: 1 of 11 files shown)


elliefm:
> I think this is pretty good but makes it seem like what we really want is to have a function for "create a login event log entry" that takes optional parameters of the various different things we log and spits out a unified format.

I'm doing basically this with audit logs at the moment, and meant to follow up with the login logs after.  I'd forgotten this PR existed; maybe this comment is what put the idea in my mind for audit logs.  I won't touch login logs till after this lands, but I probably will after that.

PR #5561, opened 2025-08-11
Cyr expire safety

brong:
This fixes an issue where renamed users would have their per-user data wiped a week later!  Ouch.

### DIFF
--- a/cassandane/tiny-tests/FastMail/cyr_expire_rename_safe
+++ b/cassandane/tiny-tests/FastMail/cyr_expire_rename_safe
@@ -1,0 +1,106 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_cyr_expire_rename_safe
+    :DelayedDelete :NoMailboxLegacyDirs
+{
+    my ($self) = @_;
+
+    my $adminstore = $self->{adminstore};
+    my $admintalk = $adminstore->get_client();
+
+    my $inbox = "user.magicuser";
+    my $subfolder = "$inbox.foo";
+
+    $admintalk->create($inbox);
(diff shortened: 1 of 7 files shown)


cassandane/tiny-tests/FastMail/cyr_expire_rename_safe
```diff
@@ -0,0 +1,106 @@
...
+    # Rename the user
+    $admintalk->rename($inbox, "user.manifold");
+
+    my $basedir = $self->{instance}{basedir};
+    open(FH, "-|", "find", $basedir);
+    my @files = grep { m{/uuid/} } <FH>;
+    close(FH);
+
+    xlog $self, "files for the inbox still exist";
+    $self->assert(scalar grep { m{$inboxid} } @files);
```
elliefm, line 54: Nit: the error messages when plain `assert()` fails are pretty useless, it's just like "assertion failed at line X", it's not like in C where you get the failed expression and can reason about it without even looking at the code.  If you must use it, it's helpful to provide a descriptive message in the (optional) second argument, which is how all the more specific assertions are implemented.

That said, there isn't a good more-specific assertion for this case.  assert_matches could be wrestled into service like `$self->assert_matches(qr{$inboxid}, "@files");`, but it's not quite the same thing.  I would probably do something like `$self->assert_num_gt(0, scalar grep ...);` or `$self->assert_num_equals(1, scalar grep ...);` if I needed the match-each-filename-distinctly behaviour.

elliefm (approved):
Looks fine to me.  A nit about the use of plain "assert" in the tests, but it's not a blocker

PR #5562, opened 2025-08-11
TestCase: fix warning when no lsan_suppressions

elliefm:
Fixes:

> Use of uninitialized value in string at Cassandane/Cyrus/TestCase.pm line 602, <GEN2> line 1.


wolfsage (approved):
Can't recall why I explicitly stringified that... weird!

PR #5563, opened 2025-08-11
jmap_mail.c: free jmapseen context when refcount goes zero

rsto:
Before, it freed the context any time decref got called and the refcount got zero or already was zero.


rsto:
Ready for review.

PR #5564, opened 2025-08-11
index.c: reusable buffers for mmaped messages and fetch item values

ksmurchison:
This eliminates allocating/freeing buffers for every message for which we are fetching items.

This will have to be rebased after #5561 is merged

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -154,7 +154,8 @@ static int index_store_annotation(struct index_state *state, uint32_t msgno,
                            msgrecord_t *mrw, struct storeargs *storeargs,
                            int *dirty);
 static int index_fetchreply(struct index_state *state, uint32_t msgno,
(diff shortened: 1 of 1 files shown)


wolfsage:
Would you mind updating the commit message to explain why you're making this change? It's not clear to me.

I'm guessing performance, but not sure

ksmurchison:
> This is half way to better :) it's clearly better, but I'd almost be tempted to hoist those buffers even a layer higher into an object passed in to this function, because it gets called for every message in a fetch reply!

I did just that.  Please re-review

rsto (approved):
Looks good to me.

rjbs (approved):
It'd be cool to see measurements on this, but I'm not sure how much work they'd be to get.  It'd be cool to live in a world where it was easy. :)

PR #5567, opened 2025-08-12
Add Mailbox/touch and Calendar/touch to bump state

ksmurchison:


rjbs:
Looks good to me, and we were *just* talking about a use for this.

I think the test could be easily made a lot stronger.  We want to assert that only the touched item is considered changed.  The current test asserts that the collection state changed.  If we do a Mailbox/changes since old-state, we can assert that the only change is in the touched item.  Should be easy, and a much stronger assertion!

imap/jmap_api.c
```diff
@@ -2791,6 +2794,105 @@ HIDDEN json_t *jmap_parse_reply(struct jmap_parse *parse)
...
+    json_t *arg;
+
+    // assume that we are given an initialized jmap_touch struct
+    touch->not_touched = json_array();
+
+    json_object_foreach(jargs, key, arg) {
+        if (!strcmp(key, "accountId")) {
+            /* already handled in jmap_api() */
+        }
+
```
rsto, line 2815: nit: why the empty lines here and below? Is that something we do in the other Foo parse methods?

imap/jmap_api.c
```diff
@@ -2791,6 +2794,105 @@ HIDDEN json_t *jmap_parse_reply(struct jmap_parse *parse)
...
+        else if (!strcmp(key, "ids") && json_array_size(arg) &&
+                 jmap_parse_strings(arg, parser, "ids")) {
```
rsto, line 2817: nit: wrong indentation

imap/jmap_api.c
```diff
@@ -2791,6 +2794,105 @@ HIDDEN json_t *jmap_parse_reply(struct jmap_parse *parse)
...
+
+HIDDEN json_t *jmap_touch_reply(struct jmap_touch *touch)
+{
+    json_t *res = json_object();
+
+    json_object_set_new(res, "state", json_string(touch->state));
+    if (json_array_size(touch->not_touched))
+        json_object_set(res, "notTouched", touch->not_touched);
+    else
+        json_object_set_new(res, "noTtouched", json_null());
```
rsto: typo: `noTtouched` -> `notTouched` . which suggests this codepath could be tested?

imap/jmap_api.c
```diff
@@ -2791,6 +2794,105 @@ HIDDEN json_t *jmap_parse_reply(struct jmap_parse *parse)
...
+    if (err) {
+        jmap_error(req, err);
+        goto done;
+    }
+
+    json_array_foreach(touch.ids, i, id) {
+        mbentry_t *mbentry = NULL;
+        struct mailbox *mailbox = NULL;
+
+        id_to_mbentry(req, json_string_value(id), &mbentry);
```
rsto, line 2870: return value of callback isn't checked, because the callback is expected to set mbentry to NULL on failure. If the `int` is supposed to be a regular Cyrus error and differs from NOT_FOUND, then report it with `error_message` on syslog?

imap/jmap_calendar.c
```diff
@@ -2314,6 +2321,22 @@ static int jmap_calendar_set(struct jmap_req *req)
     return r;
 }
 
+static int calid_to_mbentry(jmap_req_t *req, const char *id,
+                            mbentry_t **mbentry)
+{
+    char *mboxname = caldav_mboxname(req->accountid, id);
+    int r = mboxlist_lookup(mboxname, mbentry, NULL);
+
```
rsto, line 2329: nit: unnecessary empty line

imap/jmap_mailbox.c
```diff
@@ -4434,3 +4441,23 @@ static int jmap_mailbox_changes(jmap_req_t *req)
...
     return 0;
 }
+
+static int mboxid_to_mbentry(jmap_req_t *req, const char *id,
+                             mbentry_t **mbentry)
+{
+    int r;
+
+    if (USER_COMPACT_EMAILIDS(req->cstate)) {
+        r = mboxlist_lookup_by_jmapid(req->accountid, id, mbentry, NULL);
```
rsto, line 4453: nit: could use `return` in both conditional blocks, rather than `r` variable

rsto (changes requested):
Found one typo in the JMAP response fields, also think that error handling in `jmap_touch_exec` could be slightly improved. Other than that just a couple of nits regarding code style.

imap/jmap_api.c
ksmurchison: Yep, every other method parsing function has the same spacing.

imap/jmap_api.c
ksmurchison: Fixed.

imap/jmap_mailbox.c
ksmurchison: Fixed.

imap/jmap_calendar.c
ksmurchison: Fixed

imap/jmap_api.c
ksmurchison: Fixed

imap/jmap_api.c
ksmurchison: Fixed

rsto: I meant syslogging it here, rather than in each callback separately. But let's not get hung up on it.

ksmurchison: I decided to syslog it in the callbacks because I could have more context in thge log message - lookup by jmapid, uniqueid, or name (calendar)

rsto: Right, that makes sense.

PR #5568, opened 2025-08-13
ctl_conversationsdb.c: add --force

rjbs:
This will bypass the short-circuiting "already on latest version" behavior of -U.


imap/ctl_conversationsdb.c
```diff
@@ -1124,6 +1122,7 @@ int main(int argc, char **argv)
         { "recursive", no_argument, NULL, 'r' },
         { "undump", no_argument, NULL, 'u' },
         { "upgrade", no_argument, NULL, 'U' },
+        { "force", no_argument, NULL, 'f' },
```
wolfsage: The --help usage needs updating too (in this same file) to mention --force and how -U / -R behave or don't with/without it

rjbs: Fixed.

brong (approved):

PR #5570, opened 2025-08-13
lmtp_sieve.c: auditlog successful vacation autoresponses

ksmurchison:


imap/lmtp_sieve.c
```diff
@@ -2120,6 +2120,14 @@ static int send_response(void *ac, void *ic,
...
         prometheus_increment(CYRUS_LMTP_SIEVE_AUTORESPOND_SENT_TOTAL);
 
+        syslog(LOG_INFO, "sieve autoresponded: %s to: %s",
+               md->id ? md->id : "<nomsgid>", src->addr);
+        if (config_auditlog) {
+            syslog(LOG_NOTICE,
+                   "auditlog: vacation response sessionid=<%s> message-id=%s target=<%s> userid=<%s>",
+                   session_id(), md->id ? md->id : "<nomsgid>", src->addr, ctx->userid);
+        }
+
```
wolfsage, line 2141: Nice thanks. I think it'd be useful to add in the from address since those can be configured by the sieve. What do you think?

wolfsage: I think we'd also want the outmsgid so we can tie this to other logs easily

wolfsage (changes requested):

rjbs (commented):
I agree with Matthew: nice, and 1-2 more bits of data make it more useful.

imap/lmtp_sieve.c
```diff
@@ -2120,6 +2120,14 @@ static int send_response(void *ac, void *ic,
 
         prometheus_increment(CYRUS_LMTP_SIEVE_AUTORESPOND_SENT_TOTAL);
 
+        syslog(LOG_INFO, "sieve autoresponded: %s to: %s",
+               md->id ? md->id : "<nomsgid>", src->addr);
+        if (config_auditlog) {
+            syslog(LOG_NOTICE,
```
rsto: might want to use xsyslog here, didn't we deprecate syslog in favor of it?

imap/lmtp_sieve.c
ksmurchison: Done;

imap/lmtp_sieve.c
ksmurchison: Done.

imap/lmtp_sieve.c
```diff
@@ -2120,6 +2123,24 @@ static int send_response(void *ac, void *ic,
...
+                    "in.msgid=%s out.msgid=%s from=%s to=<%s>",
+                    md->id ? md->id : "", outmsgid, fromaddr, src->addr);
+        if (config_auditlog) {
+            xsyslog(LOG_NOTICE, "auditlog: vacation response",
+                    "sessionid=<%s> userid=<%s> in.msgid=%s"
+                    " out.msgid=%s from=%s to=<%s>",
+                    session_id(), ctx->userid, md->id ? md->id : "",
+                    outmsgid, fromaddr, src->addr);
+
+            buf_free(&buf);
```
wolfsage: buf_free(&buf) needs to be lifted a block or we'll leak memory if there's no auditlog

ksmurchison: Yep.  Good catch!  Fixed.

imap/lmtp_sieve.c
```diff
@@ -2107,6 +2107,9 @@ static int send_response(void *ac, void *ic,
     smtpclient_close(&sm);
 
     if (r == 0) {
+        const char *fromaddr = src->fromaddr;
```
rsto: nit: rather than introducing two variables just for logging, this could be reduced to one `struct buf fromaddr`. Initializing that with `buf_init_ro_cstr(src-fromaddr)` is cheap and will not require the reader to follow the state of `fromaddr`.

ksmurchison: Good idea.

PR #5571, opened 2025-08-14
smtpclient and others: Fix smtp typos

wolfsage:
And fix vacation_multiple test which relied on that typo to not fail. Without an smtp daemon running it was logging two messages about smtpclient, but one of them was spelled wrong so the test never picked up on it.


elliefm (approved):
yikes, good catch

PR #5575, opened 2025-08-15
IMAP: use jmap_set_*id() functions rather than building ids "by hand"

ksmurchison:


imap/imapd.c
```diff
@@ -7382,7 +7383,7 @@ static void cmd_create(char *tag, char *name, struct dlist *extargs, int localon
     }
 
     /* Close newly created mailbox before writing annotations */
-    struct conversations_state *cstate = mailbox_get_cstate(mailbox);
+    struct conversations_state *cstate = mailbox_get_cstate(mailbox);
```
elliefm: nit: trailing space added

elliefm (commented):
I think this might not build with --disable-jmap (I think jmap_util won't be available?). But I just found that the master branch doesn't build like that either.

imap/statuscache_db.c
```diff
@@ -433,8 +434,8 @@ HIDDEN void status_fill_mailbox(struct mailbox *mailbox, struct statusdata *sdat
 {
     assert(mailbox);
     assert(sdata);
-    static char static_uniqueid[101];
-    static char static_mailboxid[101];
+    static char static_uniqueid[UUID_STR_LEN];
```
wolfsage, line 437: Huh. Why was this using such big buffers before?

imap/index.c
```diff
@@ -313,12 +313,14 @@ EXPORTED int index_open_mailbox(struct mailbox *mailbox, struct index_init *init
     int r;
     struct index_state *state = xzmalloc(sizeof(struct index_state));
     struct conversations_state *cstate = mailbox_get_cstate(mailbox);
+    char *mailboxid = xzmalloc(JMAP_MAX_MAILBOXID_SIZE);
```
rsto: Why allocate and assign a local variable here, rather than allocating and assigning it to `state->mailboxid` directly? Or could `struct index_state` even get a `char mailboxid[JMAP_MAX_MAILBOXID_SIZE]` field, avoiding the allocations at all?

imap/imapd.c
```diff
@@ -5810,7 +5811,7 @@ static int parse_fetch_args(const char *tag, const char *cmd,
         fa->authstate = imapd_authstate;
     }
     if (config_getswitch(IMAPOPT_CONVERSATIONS)
-        && (fa->fetchitems & (FETCH_MAILBOXIDS|FETCH_MAILBOXES))) {
+        && (fa->fetchitems & (FETCH_MAILBOXIDS|FETCH_MAILBOXES|FETCH_EMAILID|FETCH_THREADID))) {
```
rsto, line 5814: This looks like a functional change but the commit message suggests it's just internal code that gets refactored. I guess this commit should have been split into the part that only sets mailboxid differently, then another commit add these two new `FETCH` fields. But since this PR is approved already, let's just go with it as-is.

imap/statuscache_db.c
ksmurchison: I don't know.  That code predates compact objectids, where I added the size macros

imap/imapd.c
ksmurchison: Fixed

imap/index.c
ksmurchison: Assigned it directly

imap/imapd.c
ksmurchison: Its not a functional change.  We already have these fetch fields, this just preloads conv state when these fields are requested, so we know whether to return new/old ids.

PR #5582, opened 2025-08-19
imapd.c: unconditionally reject bare LF and NUL bytes in APPEND

rsto:
This refactors message_copy_strict to always reject bare newlines and zero bytes. Before, the boolean `allow_null` argument allowed to leniently accept such data if the data was uploaded in an UTF-8 literal or as a binary literal.

Any other invocation of message_copy_strict already had disabled the `allow_null` argument. Now APPEND unconditially does, too.


rsto:
So, CI fails because it expects us to accept NUL bytes.

We can tweak this patch to only reject bare LFs, but the way we currently deal with NUL bytes regardless looks wrong: We accept NUL on ingest and then don't deal with it properly on read. We should either reject NUL in APPEND (as RFC 3516 allows us to), or rewrite the CTE of the data (as RFC 3516 seemingly allows us to, too. But didn't we conclude that rewriting data during ingest is bad?), or we keep accepting NUL as-is but then have to deal with the fallout. Thoughts?

PR #5583, opened 2025-08-19
jmap_contact.c: use compactids for AddressBooks

ksmurchison:


imap/jmap_api.c
```diff
@@ -2254,7 +2254,10 @@ HIDDEN json_t *jmap_copy_reply(struct jmap_copy *copy)
 
 /* Foo/query */
 
-HIDDEN jmap_filter *jmap_buildfilter(json_t *arg, jmap_buildfilter_cb *parse)
+HIDDEN jmap_filter *jmap_buildfilter(jmap_req_t *req,
```
rsto: Why not plumb `req` through in `rock` in those callbacks that actually need it?

ksmurchison: Yeah, I can do that

rsto (approved):

wolfsage (changes requested):
I need more brain for this (later this week perhaps), but looking at tests and trying something I noticed that if you create a ContactCard and don't specify an addressBookIds, the created response includes this:
```
            "addressBookIds": {
              "Default": true
            },
```

while a subsequent get of the contact card has:
```
            "addressBookIds": {
              "R0k": true
            }

```

The data in the /set created response looks wrong

ksmurchison:
> I need more brain for this (later this week perhaps), but looking at tests and trying something I noticed that if you create a ContactCard and don't specify an addressBookIds, the created response includes this:
> 
> ```
>             "addressBookIds": {
>               "Default": true
>             },
> ```
> 
> while a subsequent get of the contact card has:
> 
> ```
>             "addressBookIds": {
>               "R0k": true
>             }
> ```
> 
> The data in the /set created response looks wrong

Yep.  Good catch

wolfsage:
As far as eyeballing goes everything but the final commit (xxxContactCardId - I did not review because I believe this is all in flux) looks good to me. It'd be nice to get @brong eyes on it thought

imap/carddav_db.c
```diff
@@ -663,47 +694,41 @@ EXPORTED strarray_t *carddav_getuid2groups(struct carddav_db *carddavdb,
...
+                                       const char *vcard_uid, unsigned kind,
+                                       const mbentry_t *othermb)
 {
     int r = 0;
     int isshared = 0;
-    if (!strncmpsafe(group, "shared/", 7)) {
+    if (!strncmpsafe(vcard_uid, "shared/", 7)) {
         assert(!mbentry); // no mailbox filter on shared groups
         if (!carddavdb->db->attached) return NULL;
-        if (!*group) return NULL; // can't just be "shared/"
```
brong, line 704: looks like a bug was fixed here too!

imap/jmap_util.c
```diff
@@ -1346,6 +1346,38 @@ EXPORTED void jmap_set_threadid(struct conversations_state *cstate,
...
+    if (USER_COMPACT_EMAILIDS(cstate)) {
+        strlcpy(abookid, mbentry->jmapid, JMAP_MAX_ADDRBOOKID_SIZE);
+        // replace MAILBOXID prefix with ADDRBOOK prefix
+        abookid[0] = JMAP_ADDRBOOKID_PREFIX;
+    }
+    else {
+        // last segment of mailbox name
+        mbname_t *mbname = mbname_from_intname(mbentry->name);
+        const strarray_t *boxes = mbname_boxes(mbname);
+        const char *id = strarray_nth(boxes, boxes->count-1);
```
brong: This could just be `strarray_nth(boxes, -1)` - it does negative indexing like Perl does

imap/mboxlist.c
```diff
@@ -1106,6 +1106,7 @@ EXPORTED int mboxlist_lookup_by_jmapid(const char *inboxid, const char *jmapid,
 
     if (entryptr) {
         *entryptr = entry;
+        if (!entry->jmapid) entry->jmapid = xstrdup(jmapid);
```
brong, line 1110: Is this possible?  I can't see this being WRONG, but what entries will be found with `lookup_by_jmapid` but then not have a jmapid?

cassandane/tiny-tests/JMAPEmail/email_query_fromcontactcarduid_shared
```diff
@@ -0,0 +1,143 @@
...
+        }, 'R1']
+    ], $using);
+    $self->assert_num_equals(3, scalar @{$res->[0][1]{ids}});
+    my $emailIds = $res->[0][1]{ids};
+
+    xlog "Assert Email/query";
+    $res = $jmap->CallMethods([
+        ['Email/query', {
+            filter => {
+                fromContactCardUid => "shared/$contactUid",
```
rjbs: This `shared/` is bogus.  The spec (text of CYR-2203) just says "look in all AddressBooks visible to the user".  Putting "shared/" on the front shouldn't be needed, and creates ambiguity.  (Imagine the joker who creates a card with the uid "shared/foo"!)

ksmurchison: Fixed

imap/mboxlist.c
ksmurchison: We are looking up an entry by its J key, but the entry value might not actually include the jmapid

imap/jmap_util.c
ksmurchison: Fixed

PR #5584, opened 2025-08-20
cass: Add ContactCard test showing adding a member to an empty group …

wolfsage:
…crashes

I've seen crashes after adding a contact to an empty group after removing all previous contacts, but this test is actually triggering the crash on the *first* add, which surprised me...


wolfsage:
This was pulled into https://github.com/cyrusimap/cyrus-imapd/pull/5596/files

PR #5585, opened 2025-08-20
make OpenSSL mandatory

elliefm:
**Reviewers:** the repetitive noise is bundled up in the "various: remove ssl conditionalisation" commit; the interesting parts have their own commits.

This removes the `--with-openssl` configure option and all cruft associated with it.  OpenSSL is now required, and is detected exclusively using pkg-config.


wolfsage:
This seems right.

The change suggests that --with-openssl=/path/to/some/specific/ver used to work, but we're ripping out --with-openssl. Do we need to document how to specify a specific installation of openssl going forward or is that "common knowledge of using pkg-config"?

elliefm:
Oh, here is where I read that linking against multiple versions of OpenSSL can cause problems: https://www.cyrusimap.org/imap/reference/faqs/install-linkerwarnings.html

elliefm:
Yeah I expect so, but I wasn't sure how to prove that to myself, so I left it conditional for now.  The amount of conditionalised code isn't much, and the conditionalisation isn't tricky to reason about.  I can chase it further if we want, or are we happy with it being left as is?

rsto (approved):
Great

elliefm:
I've spun the TLS ALPN question off into a separate task, so I'll leave it as is in this PR, and revisit it later

rjbs:
What we want is to eliminate needless options, to reduce the number of possible configurations producing bug reports.  For most features, if they are to be there at all, I'd like to see them simply _there_.  Setting up a development environment is, in my experience and opinion, easy enough that it doesn't form a strong argument to allowing more configurations to exist in the wild.  (Even moreso if you develop with cyrus-docker.)

PR #5586, opened 2025-08-20
message.c: parse multipart bodies containing bare newlines

rsto:
Bare newlines are forbidden except for "binary" Content-Transfer encoding but APPEND leniently accepts such messages. This patch is to parse them appropriately, irrespective if APPEND might become more stricter for new messages in the future.


ksmurchison (approved):
LGTM

rsto:
Updated to also bump the mailbox cache minor version.

PR #5587, opened 2025-08-21
xapian_wrap.c: do not stem Italian and Irish Gaelic elisions

rsto:
Italian and Irish Gaelic both elide articles and other elements using apostrophes. This patch refactors the French contraction stemmer to a generic elision stemmer.


rsto:
Huh, CI catched a mismatch between my `aarch64` build environment and `amd64` platforms.

rsto:
Alright, CI passed, this is ready for review for real. Note that in addition to the new test for Italian elisions, it also passed the already existing French contraction test (`JMAPEmail.email_query_language_french_contractions`). Irish Gaelic elisions are a functional subset of Italian and French, no test added.

dilyanpalauzov:
I think using in `get_stemmer()` std::vector or even array instead of std::list is a better choice, as std::list uses non-contiguous memory for the elements - https://www.geeksforgeeks.org/cpp/when-to-use-list-instead-of-vector-in-cpp/ .  Using in the constructor of ElisionStemmer references instead of pointers is better C++ as it shows more evident where memory is allocated and deallocated.  It also allows, as a matter of preference, in-place initialization, instead of using separate variable.  Something like:
```c++
explicit ElisionStemmer(const std::vector<std::string> &elisions, Xapian::Stem stem)
…
static const std::list<std::string> elisions = {
            "c",     "l",      "un",    "m",    "t",    "s",    "v",
            "d",     "all",    "dall",  "dell", "nell", "sull", "coll",
            "pell",  "gl",     "agl",   "dagl", "degl", "negl", "sugl",
            "nient", "nessun", "quest", "quell"
};
stem = Xapian::Stem{ new ElisionStemmer(elisions, stem) }
```

rsto:
> I think using in `get_stemmer()` std::vector or even array instead of std::list is a better choice, as std::list uses non-contiguous memory for the elements - https://www.geeksforgeeks.org/cpp/when-to-use-list-instead-of-vector-in-cpp/ . 

Thanks, I was not aware of that. I agree that vector or array it is a better choice then, I'll the update code.

> Using in the constructor of ElisionStemmer references instead of pointers is better C++

I just write enough C++ to get the Xapian wrapper working and make `clang-tidy` happy, so I appreciate any suggestions how to make the code more idiomatic C++ code. That being said, my first intuition was to use references in the constructor but this blog post by a C++ veteran made me use pointers instead: https://herbsutter.com/2020/02/23/references-simply/

Issue #5573, opened 2025-08-14
squatter deleted all of my users' mail when I told it to compact


jikamens:
This happened with 3.4.8. I not know if it is still an issue in 3.12.1 and I can't afford to find out because I don't have a test server to try it out on and I can't afford to break my family's email again like I did last night when I ran into this the first time.

I had both xapian storage and message storage configured to go to the same place, /var/spool/imap, in imapd.conf:

```
defaultpartition: default
partition-default: /var/spool/imap
defaultsearchtier: default
defaultsearchpartition-default: /var/spool/imap
```

This worked fine until I got it into my head to compact all of my xapian indexes because they were taking up a ridiculous amount of space:

`sudo -u cyrus -H squatter -z default -t default`

When this command finished running, most of my users' mailboxes had been completely wiped out. Except one mailbox was untouched, I really don't understand why.

It would appear that squatter decided all the messages and cyrus.* files in all of the mailboxes were superfluous files so it should just remove them or something.

Fortunately I had a backup from the previous night, so I was able to restore most people's emails, but it caused a lot of grief for the people in my family.

If it's not safe to store the xapian files in the same hierarchy as the message files, then imapd shouldn't be allowed to start if it's configured that way. Or maybe squatter shouldn't delete files that it has nothing to do with? Or _something_ other than what actually happened.

elliefm:
Ouch, that sucks.  I'm glad you were able to rescue it.

> If it's not safe to store the xapian files in the same hierarchy as the message files, then imapd shouldn't be allowed to start if it's configured that way.

I agree.  I've opened #5590 to address this.

PR #5590, opened 2025-08-22
config partition sanity checker

elliefm:
#5573 reports that, when Cyrus is configured with mail and search indexes being stored in the same directory, like this:

```
defaultpartition: default
partition-default: /var/spool/imap
defaultsearchtier: default
defaultsearchpartition-default: /var/spool/imap
```

... then compacting the xapian indexes will remove all the mail and metadata files.  Seems like xapian, at least, expects total control over the directories it's configured to use.  Given that, Cyrus shouldn't be allowed to start up in this configuration.

I don't think it makes sense for _any_ of the `*partition-*` settings to share directories with one another.  Even if some combination doesn't currently break, I don't believe we expect these to overlap, so that combination might break in the future.

This PR adds a sanity check to make sure that no partitions are sharing directories, nor are descendants of one another.  If this sort of sharing is detected, `master` and `squatter` will now refuse to start, as will any other cyrus service or tool that requests partition data (via the `CONFIG_NEED_PARTITION_DATA` argument to `cyrus_init()`).  Details of the collisions are logged to syslog.  `cyr_info conf-lint` will detect and report the bad entries in the same way it reports other bad entries.

The sanity checking is based on string comparisons of the configured paths; actual filesystem details are not examined.  So it's probably possible to subvert the checks using symlinks or other trickiness, but I don't think I care -- just don't do that.

For implementation reasons, if multiple partitions are configured as descendants of the same other partition, only the parent and one of the descendants can be reported.  That's enough to prevent starting up with a bad configuration, but a really bad configuration might need several rounds of test/fix to identify and resolve all the issues.

**If people already have configuration like this, what they can do to fix it prior to (or after) upgrading?**  Presumably rearrange their storage, then update their imapd.conf to describe the new layout.  Hopefully they already know how to do this.

### DIFF
--- a/lib/libconfig.c
+++ b/lib/libconfig.c
@@ -1393,10 +1393,17 @@ static void collect_partitions(const char *key, const char *value, void *rock)
     }
 }
 
-static void check_no_dups(const char *value, void *foo, void *rock)
+struct check_no_dups_rock {
+    FILE *user_output;
+    int *found_bad;
+};
+
+static void check_no_dups(const char *value, void *vpkeys, void *vprock)
 {
-    const strarray_t *keys = foo;
-    int *found_bad = rock;
+    const strarray_t *keys = vpkeys;
+    struct check_no_dups_rock *rock = vprock;
+    FILE *user_output = rock->user_output;
+    int *found_bad = rock->found_bad;
 
     if (strarray_size(keys) > 1) {
         char *joined_keys = NULL;
@@ -1407,11 +1414,31 @@ static void check_no_dups(const char *value, void *foo, void *rock)
                          value, joined_keys);
         free(joined_keys);
 
+        if (user_output) {
+            int i, n;
+
+            for (i = 0, n = strarray_size(keys); i < n; i++) {
+                fprintf(user_output, "%s: %s\n", strarray_nth(keys, i), value);
+            }
+        }
+
         (*found_bad) ++;
     }
 }
 
-static int check_no_subdirs(const hash_table *by_value)
+static void dump_kv(FILE *user_output,
+                    hash_table *by_value,
+                    const char *value)
+{
+    const strarray_t *keys;
+
+    keys = hash_lookup(value, by_value);
+    if (keys && strarray_size(keys)) {
+        fprintf(user_output, "%s: %s\n", strarray_nth(keys, 0), value);
+    }
+}
+
+static int check_no_subdirs(hash_table *by_value, FILE *user_output)
 {
     strarray_t *all_values;
     const char *prev, *value;
@@ -1428,10 +1455,14 @@ static int check_no_subdirs(const hash_table *by_value)
             && 0 == strncmp(prev, value, prev_len)
             && value[prev_len] == '/')
         {
+            /* XXX only logs first example, and no keys... */
             xsyslog(LOG_ERR, "disk path is a prefix of others",
                              "path1=<%s> path2=<%s>",
                              prev, value);
-            /* XXX only reports first example, and no keys... */
+            if (user_output) {
+                dump_kv(user_output, by_value, prev);
+                dump_kv(user_output, by_value, value);
+            }
             found_bad++;
         }
 
@@ -1449,7 +1480,7 @@ static void wrap_strarray_free(void *vp)
     strarray_free((strarray_t *) vp);
 }
 
-EXPORTED int config_partition_sanity(void)
+EXPORTED int config_partition_sanity(FILE *user_output)
 {
     hash_table by_value = HASH_TABLE_INITIALIZER;
     int found_bad = 0;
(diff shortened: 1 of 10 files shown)


lib/libconfig.c
```diff
@@ -1307,3 +1311,193 @@ EXPORTED void config_toggle_debug(void)
     config_debug = !config_debug;
     if (config_toggle_debug_cb) config_toggle_debug_cb();
 }
+
+static const unsigned char cmpstringp_path_lookup[] = {
+    0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07,
+    0x08, 0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f,
+    0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17,
+    0x18, 0x19, 0x1a, 0x1b, 0x1c, 0x1d, 0x1e, 0x1f,
+    0x21, 0x22, 0x23, 0x24, 0x25, 0x26, 0x27, 0x28,  /* interesting */
```
rsto, line 1320: what is the interesting bit?

lib/libconfig.c
```diff
@@ -1307,3 +1311,193 @@ EXPORTED void config_toggle_debug(void)
...
+
+        prev = value;
+    }
+
+    strarray_free(all_values);
+
+    return found_bad;
+}
+
+EXPORTED int config_partition_sanity(FILE *user_output)
```
rsto: nit: outside this PR, I don't think I would understand what this function is supposed to do. Maybe change its name to something actionable (e.g. `check_..`) and write a one-liner that explains what it does?

lib/libconfig.c
ksmurchison: Looks like SP has been moved to sort after /.  Is that correct?

ksmurchison (approved):
Code is approved.  I agree with rsto that it could use some more commentary.

lib/strarray.c
```diff
@@ -67,7 +67,7 @@ EXPORTED void strarray_fini(strarray_t *sa)
     sa->alloc = 0;
 }
 
-EXPORTED void strarray_free(strarray_t *sa)
+EXPORTED void strarray_free(void *sa)
```
wolfsage: I don't think we should do this. It removes compile-time protection that prevents us from mistakenly passing junk into strarray_free like `&sa` instead of `sa` for example.

Rather, we should do what we do elsewhere:
```
imap/jmap_mail_query.c:    free_hash_table(&cfilter->contactgroups, (void(*)(void*))strarray_free);
```

lib/libconfig.c
elliefm: The comment here only highlights which lines are actually different from a straight 00-ff order; the details are explained below, on the function where the lookup table is used.

> Looks like SP has been moved to sort after /. Is that correct?

Not quite...

The 0x20 ... 0x2f rows have been rearranged to be 0x21 ... 0x2f, 0x20.  This means when you look up index `/` (0x2f), the value retrieved is `0x20`, which is the first printable.  When you look up any other other index in the 0x2_ range (punctuation), the value you get back is one higher than it would otherwise have been, to compensate for 0x2f having been changed.  This means `/` sorts before all other printable characters, but otherwise the relative sorting is unchanged.  It does look like SP was moved, but this table is values _retrieved_, not the input chars, so actually it is `/` that was moved.

The comment on the function below says:

```
* Treats '/' (0x2f) as lower than other printables so that
 * file paths sort pre-order, depth-first.
 * XXX This should probably be in lib/bsearch.c, but that's in
 * XXX libcyrus, which we don't have here.
 */
static int cmpstringp_path(const void *aa, const void *bb)
{
```

Is this not sufficient?

lib/strarray.c
elliefm: I'm not sure that casting a function pointer type to a different function pointer type isn't undefined behaviour tbh.  If we do it elsewhere I guess it works... but what I would actually do, if not changing `strarray_free()`, is write a local wrapper function with the right signature and call `strarray_free()` from that.

elliefm:
> I don't _think_ it accounts for symlinks. Should it? Or should the docs specify it has this weakness if not?

I addressed that in the PR description already:

> The sanity checking is based on string comparisons of the configured paths; actual filesystem details are not examined. So it's probably possible to subvert the checks using symlinks or other trickiness, but I don't think I care -- just don't do that.

If some sysadmin knows they can't configure multiple partitions to the same path (because Cyrus is refusing to start, or `cyr_info conf-lint` is complaining), but they manage to fudge things enough to trick the sanity checker into allowing it, the consequences are all theirs.

lib/libconfig.c
elliefm: In my mind, `foo_sanity` is clearly a sanity checker for `foo`, so it makes sense to me that `config_partition_sanity` is a sanity checker for partitions, provided by the config module.  But I had a look, and this isn't an idiom we use elsewhere in Cyrus, so I'll rename it to something like `config_check_partitions`.

lib/strarray.c
wolfsage: I think this legal because you're casting the incoming argument to void* which is legal?

-fsanitize=undefined does not find it with both gcc and clang, and scan-build does not complain about it

elliefm: Ahh yeah, that makes sense, though I still think it smells bad.  I've dropped the change to strarray_free, and then made a local wrapper function rather than casting.  This isn't performance-critical code, so the extra function call overhead isn't a concern.

PR #5591, opened 2025-08-25
xapian_wrap.cpp: use vector instead of list for elisions

rsto:
C++ <list> is a linked list but we want elisions to be in contiguous memory for efficiency.


rsto:
This is a follow-up to https://github.com/cyrusimap/cyrus-imapd/pull/5587

ksmurchison (approved):

PR #5594, opened 2025-08-26
cyr_dbtool: better handling of transaction AND batch

brong:
This fixes transaction and batch support.  Right now, '-T' is not compatible with batch; so we can't take a shared lock.

### DIFF
--- a/imap/cyr_dbtool.c
+++ b/imap/cyr_dbtool.c
@@ -75,6 +75,7 @@ static char stack[STACKSIZE+1];
 static int outfd;
 
 static struct db *db = NULL;
+static struct txn **tidp = NULL;
 
 static int read_key_value(char **keyptr, size_t *keylen, char **valptr, size_t *vallen) {
   int c,res,inkey;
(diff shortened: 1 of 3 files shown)


imap/cyr_dbtool.c
```diff
@@ -434,18 +433,22 @@ int main(int argc, char *argv[])
     } else if (!strcmp(action, "repack")) {
         r = cyrusdb_repack(db);
     } else if (!strcmp(action, "damage")) {
-        cyrusdb_store(db, "INVALID", 7, "CRASHME", 7, &tid);
-        assert(!tid);
+        cyrusdb_store(db, "INVALID", 7, "CRASHME", 7, tidp);
+        assert(!*tidp);
```
rsto, line 519: `tidp` is initialized to `NULL`, so this could become a NULL pointer dereference

rsto (changes requested):

imap/cyr_dbtool.c
brong: Hmm... yeah, I guess you need to explicitly set the transaction in that case.  I'll patch it.

rsto:
base64 additions look good. You might want to also update the man page once it's clear we are going to keep the new options.

brong:
> base64 additions look good. You might want to also update the man page once it's clear we are going to keep the new options.

Done!  Also switched Cassandane to use it.

PR #5595, opened 2025-08-26
search_query.c: require top-level message for header query results

rsto:
A result for a query including a header search criteria might be a false positive if it matches a message that is both a top-level Cyrus message and a message/rfc822 attachment in another message.

This patch excludes search results having part ids if the subclause requires at least one header criteria to match.


ksmurchison (approved):
Looks sane

rsto:
Tagged this `include-in-fastmail`, let's merge this later.

PR #5598, opened 2025-08-29
jmap_contact.c: Unescape value of unknown property types

ksmurchison:
This prevents exponentially increasing backslashes when cards are updated


ksmurchison:
> Typo in commit message: `increasng`

Fixed

wolfsage (approved):
Seems right, though I'm surprised the escaping has to happen here and not in libicalvcard

PR #5601, opened 2025-08-30
jmap_contact.c: add prodid to ContactCard if not supplied in /set{cre…

ksmurchison:
…ate}

### DIFF
--- a/imap/jmap_contact.c
+++ b/imap/jmap_contact.c
@@ -11503,6 +11503,12 @@ static int _card_set_create(jmap_req_t *req,
     media = json_deep_copy(json_object_get(jcard, "media"));
     keys = json_deep_copy(json_object_get(jcard, "cryptoKeys"));
 
+    if (!json_object_get(jcard, "prodid")) {
+        /* set the PRODID */
+        vcardcomponent_add_property(card, vcardproperty_new_prodid(_prodid));
+        json_object_set_new(item, "prodid", json_string(_prodid));
+    }
+
     if (!json_object_get(jcard, "created")) {
         /* set the CREATED time */
         char datestr[ISO8601_DATETIME_MAX+1] = "";
--- a/cassandane/tiny-tests/JMAPContacts/card_set_create_no_prodid
+++ b/cassandane/tiny-tests/JMAPContacts/card_set_create_no_prodid
@@ -1,0 +1,48 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_card_set_create_no_prodid
+    :needs_dependency_icalvcard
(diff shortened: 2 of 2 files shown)


imap/jmap_contact.c
```diff
@@ -11503,6 +11503,12 @@ static int _card_set_create(jmap_req_t *req,
     media = json_deep_copy(json_object_get(jcard, "media"));
     keys = json_deep_copy(json_object_get(jcard, "cryptoKeys"));
 
+    if (!json_object_get(jcard, "prodid")) {
```
wolfsage: The property name should be `prodId`. No other tests break because nothing is checking strictly that a vcard looks right if a prodId is given I think.

That is, if you adjust the test above:
```
diff --git a/cassandane/tiny-tests/JMAPContacts/card_set_create_no_prodid b/cassandane/tiny-tests/JMAPContacts/card_set_create_no_prodid
index 6e1853671a..1c344c945f 100644
--- a/cassandane/tiny-tests/JMAPContacts/card_set_create_no_prodid
+++ b/cassandane/tiny-tests/JMAPContacts/card_set_create_no_prodid
@@ -28,7 +28,8 @@ sub test_card_set_create_no_prodid
                     '@type' => 'Card',
                     version => '1.0',
                     kind => 'individual',
-                    name => { full => $name }
+                    name => { full => $name },
+                    prodId => "Oh No Software",
                 }
             }
         }, 'R1']

```

when you run it it still passes but we see a card with two prodIds:
```
>>>>>>>> HTTP/1.1 200 OK
BEGIN:VCARD
VERSION:4.0
UID;VALUE=TEXT:urn:uuid:dcb83c04-909c-4713-b10d-44f92e1f6181
PRODID:-//CyrusIMAP.org//Cyrus 3.13.0-alpha0-1197-g835b1d978f//EN
KIND:INDIVIDUAL
FN:Mr. John Q. Public\, Esq.
PRODID:Oh No Software
CREATED:20250902T170238Z
END:VCARD
```

The code is also adding `prodid` to the created response instead of `prodId`.

We should probably update this test to check the created response, and update one of the other tests that allows passing a prodId to make sure the resulting card looks right and that a ContactCard/get looks right too (if we don't already have such a test...)

wolfsage (changes requested):

imap/jmap_contact.c
wolfsage: (We should probably update the casing in the commit message too)

ksmurchison: All fixed

wolfsage (approved):

PR #5603, opened 2025-09-02
changes: start a changes entry for compact ids

rjbs:
This is more free-form text than most changes files, but I think it's probably about right.

It claims that new users will be created compact by default, which we'll want to make true before claiming it.

### DIFF
--- a/changes/next/compact-object-ids
+++ b/changes/next/compact-object-ids
@@ -1,0 +1,46 @@
+Description:
+
+New compact object ids for Email, Mailbox, and Thread types
+
+Documentation:
+
+There isn't any.  (Should there be?)
+
+
+Config changes:
+
+None.
+
+
+Upgrade instructions:
+
+This version of Cyrus introduces "compact" object ids for email, mailbox, and
+thread objects.  These forms can be packed into 64-bits for more efficient
+client-side caching.  Their ASCII forms are also much shorter, for reduced
+bandwidth use.
+
+All users created in this version of Cyrus IMAP will have compact ids enabled
+by default.
+
+It is not strictly required to switch all users to compact object ids to
+upgrade to this version of Cyrus, but it will very likely be required before
+the next version.
+
+To switch a user to compact object ids, you must:
+
+* reconstruct each user using `reconstruct -u -T -V max`
+* upgrade the user's conversations database with `ctl_conversationsdb -U -R`
+
+In a replicated environment, these should be performed on the primary first and
+the replicas afterward.  Then, only on the primary:
+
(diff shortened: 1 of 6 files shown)


changes/next/compact-object-ids
```diff
@@ -0,0 +1,33 @@
...
+client-side caching.  Their ASCII forms are also much shorter, for reduced
+bandwidth use.
+
+It is not strictly required to upgrade to compact object ids to upgrade to this
+version of Cyrus, but it will very likely be required before the next version.
+
+To upgrade, you must reconstruct each user using `reconstruct -u -r -T -V max`,
+followed by `ctl_conversationsdb -I on -v`.
+
+Upon upgrade, all newly created users will have compact ids by default.
```
elliefm: There's some overloaded uses of "upgrade" in here:

* upgrading cyrus-imapd to a version with this feature
* upgrading a user account from legacy object ids (do we have a name for this?) to compact object ids

Maybe we want to use something like "convert" instead of "upgrade" for the reconstruct/ctl_conversationsdb parts?  Also, I think it might be clearer overall if the order of the last three paragraphs were reversed... perhaps something like (pseudoprose):

> This version of Cyrus introduces [...]
> 
> New users created on this version will have compact ids automatically.  [I believe this can't be changed, so it's not by default, it just is.  If it can be changed, document how?]
> 
> To convert existing users to compact ids, [reconstruct/ctl_conversationsdb steps]
> 
> It is not strictly required [...]

We also need an order of operations for upgrading a murder deployment, if there are ordering dependencies.  Can an older mupdate server interact sanely with an upgraded/converted backend?  Or does mupdate need to know about compact ids before its backends can start using them, and therefore must be upgraded first?  Might want to extend the MurderIMAP/MurderJMAP cassandane tests to definitely exercise some compact id cases, and then run them in a two-versions setup with each of 3.12 and master on either side.

elliefm (commented):
Thanks, this is very heartening to see!

changes/next/compact-object-ids
rjbs: Thanks, @elliefm, I was going to ask for your feedback on the general shape of this, and you have preempted my question with just the kind of answer I was hoping for!

rjbs: @elliefm I've rewritten the document a bit to try to be clearer.  I haven't addressed murder, though, and need expert input.

elliefm: Those changes look great, thanks!  When I do the release, I will feel very comfortable just pasting straight into the release docs, without feeling like I need to (learn and) rewrite the details myself.

Thanks for including the replication order of operations, I hadn't thought of that myself yet!

PR #5605, opened 2025-09-05
jmap_contact.c: handle all data types for vCardProps

ksmurchison:
Note that we can translate structured X- props from JSContact
to vCard because we have enough context to do so.
Unfortunately, we lack enough context to do the reverse

### DIFF
--- a/cassandane/tiny-tests/JMAPContacts/card_set_vcardprops
+++ b/cassandane/tiny-tests/JMAPContacts/card_set_vcardprops
@@ -1,0 +1,243 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_card_set_vcardprops
+    :needs_dependency_icalvcard :MailboxLegacyDirs
+{
+    my ($self) = @_;
+    my $jmap = $self->{jmap};
+
+    my $service = $self->{instance}->get_service("http");
+    $ENV{DEBUGDAV} = 1;
+    my $carddav = Net::CardDAVTalk->new(
+        user => 'cassandane',
+        password => 'pass',
+        host => $service->host(),
+        port => $service->port(),
+        scheme => 'http',
+        url => '/',
+        expandurl => 1,
+    );
+
+    my $prodid = '-//Example Corp.//CardDAV Client//EN';
+    my $name = 'Mr. John Q. Public, Esq.';
+
+    xlog $self, "Create card with vCard X- properties";
+    my $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            create => {
+                "1" => {
+                    '@type' => 'Card',
+                    version => '1.0',
+                    prodId => $prodid,
+                    kind => 'individual',
+                    name => { full => $name },
+                    vCardProps => [
+                        [
+                            'x-text',
+                            {},
+                            'text',
+                            'baz',
+                        ],
+                        [
+                            'x-text-multi',
+                            {},
+                            'text',
+                            'foo', "", 'bar',
+                        ],
+                        [
+                            'x-uri',
+                            {},
+                            'uri',
+                            'ldap://ldap.example.com/cn=foo',
+                        ],
+                        [
+                            'x-date',
+                            {},
+                            'date',
+                            '--04-12',
+                        ],
+                        [
+                            'x-time-utc',
+                            {},
+                            'time',
+                            '12:30:00Z',
+                        ],
+                        [
+                            'x-date-time',
+                            {},
+                            'date-time',
+                            '2013-08-15T09:45:00+01:00',
+                        ],
+                        [
+                            'x-time-standalone',
+                            {},
+                            'date-and-or-time',
+                            'T12:30',
+                        ],
+                        [
+                            'x-timestamp',
+                            {},
+                            'timestamp',
+                            '2013-02-14T12:30:00Z',
+                        ],
+                        [
+                            'x-boolean',
+                            {},
+                            'boolean',
+                            JSON::true,
+                        ],
+                        [
+                            'x-integer-min',
+                            {},
+                            'integer',
+                            -9223372036854775808,
+                        ],
+                        [
+                            'x-integer-max',
+                            {},
+                            'integer',
+                            9223372036854775807,
+                        ],
+                        [
+                            'x-integer-multi',
+                            {},
+                            'integer',
+                            1, 2, 3, 4, 5,
+                        ],
+                        [
+                            'x-float-min',
+                            {},
+                            'float',
+                            -9007199254740992.,
+                        ],
+                        [
+                            'x-float-max',
+                            {},
+                            'float',
+                            9007199254740991.,
+                        ],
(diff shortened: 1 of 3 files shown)


cassandane/tiny-tests/JMAPContacts/card_set_vcardprops
```diff
@@ -0,0 +1,209 @@
...
+        port => $service->port(),
+        scheme => 'http',
+        url => '/',
+        expandurl => 1,
+    );
+
+    my $prodid = '-//Example Corp.//CardDAV Client//EN';
+    my $name = 'Mr. John Q. Public, Esq.';
+
+    xlog $self, "Create card with vCard X- property that needs escaping";
```
wolfsage: nit: I think this description is copied from another test and not quite accurate here

cassandane/tiny-tests/JMAPContacts/card_set_vcardprops
```diff
@@ -0,0 +1,209 @@
...
+                "1" => {
+                    '@type' => 'Card',
+                    version => '1.0',
+                    prodId => $prodid,
+                    kind => 'individual',
+                    name => { full => $name },
+                    vCardProps => [
+                        [
+                            'x-text',
+                            { },
```
wolfsage: nit: extra space in this one {} block

cassandane/tiny-tests/JMAPContacts/card_set_vcardprops
```diff
@@ -0,0 +1,209 @@
...
+                    ]    
+                }
+            }
+        }, 'R1']
+    ]);
+
+    my $id = $res->[0][1]{created}{1}{id};
+    my $href = $res->[0][1]{created}{1}{'cyrusimap.org:href'};
+
+    xlog $self, "Verify that X- property is escaped";
```
wolfsage: nit: This comment also not quite useful here

imap/jmap_contact.c
```diff
@@ -6678,6 +6679,51 @@ static void jscomps_from_vcard(json_t *obj, vcardproperty *prop,
...
+    do {
+        json_t *jval = NULL;
+
+        switch (vkind) {
+        case VCARD_BOOLEAN_VALUE:
+            jval = json_boolean(!strcasecmp("TRUE", val));
+            break;
+
+        case VCARD_INTEGER_VALUE:
+            jval = json_integer(atoi(val));
```
wolfsage: https://datatracker.ietf.org/doc/html/rfc6350#section-4.5 says this is a 64-bit signed integer. If so, we can't use atoi here

wolfsage (changes requested):
Some starting feedback, will look more later

imap/jmap_contact.c
```diff
@@ -6678,6 +6679,51 @@ static void jscomps_from_vcard(json_t *obj, vcardproperty *prop,
...
+            break;
+
+        case VCARD_FLOAT_VALUE:
+            jval = json_real(atof(val));
```
wolfsage: I also wonder what are we supposed to do here if the provided values don't convert cleanly. This could lie and return 0.0 for bad values and then we can't round trip properly between different protocols...

cassandane/tiny-tests/JMAPContacts/card_set_vcardprops
ksmurchison: Fixed

cassandane/tiny-tests/JMAPContacts/card_set_vcardprops
ksmurchison: Fixed

imap/jmap_contact.c
ksmurchison: Fixed

imap/jmap_contact.c
ksmurchison: I think its reasonable to assume that libicalvcard will reject properties with bad values on ingest, so we can avoid doing any error detection when converting.

wolfsage: I'm not entirely sure what you mean. We can put bad values in through a caldav PUT.

For example, if I put 2^64 in, JMAP/get returns LLONG_MAX instead of the value I gave it.

Do we care / is there anything we could even do?

imap/jmap_contact.c
```diff
@@ -7436,6 +7490,9 @@ static void jsprop_from_vcard(vcardproperty *prop, json_t *obj,
                 break;
             }
         }
+        else {
```
brong: I probably would have inverted this and put it at the top - `if (type) {  } else { }`

brong: It looks like we're overwriting the 'vkind = vcardparameter_kind_value_kind(param_kind, &is_multivalued);' up above in all cases now.  Is that OK?

ksmurchison: Fixed.  Used separate variables

imap/jmap_contact.c
ksmurchison: Yeah, ok.  That's an issue with the libicalvcard parser which will have to be resolved there.

cassandane/tiny-tests/JMAPContacts/card_set_vcardprops
ksmurchison: Fixed

cassandane/tiny-tests/JMAPContacts/card_set_vcardprops
```diff
@@ -0,0 +1,243 @@
...
+    $self->assert_matches(qr|X-DATE;VALUE=DATE:--04-12|, $card);
+    $self->assert_matches(qr|X-TIME-UTC;VALUE=TIME:12:30:00Z|, $card);
+    $self->assert_matches(qr|X-DATE-TIME;VALUE=DATE-TIME:2013-08-15T09:45:00\+01:00|,
+                          $card);
+    $self->assert_matches(qr|X-TIME-STANDALONE;VALUE=DATE-AND-OR-TIME:T12:30|,
+                          $card);
+    $self->assert_matches(qr|X-TIMESTAMP;VALUE=TIMESTAMP:2013-02-14T12:30:00Z|,
+                          $card);
+    $self->assert_matches(qr|X-BOOLEAN;VALUE=BOOLEAN:TRUE|, $card);
+    $self->assert_matches(qr|X-INTEGER-MIN;VALUE=INTEGER:-9223372036854775808|,
```
wolfsage: Thanks for adding these! One final complaint: These kind of checks are a little too lax, it doesn't tell us if there's somehow junk after the value in the card output, it'd be good to ensure we get whatever it is we expect to be next (either \r\n, or a ;, or whatever the separator is). But I see we do this everywhere so I don't think it's necessary to fix here, but something we should fix and keep in mind when writing future tests

ksmurchison: Done.  I will merge when CI completes

PR #5606, opened 2025-09-08
speed up slowio unit tests

elliefm:
This PR substantially extends the `cunit/unit-timeofday.c` time mocking functionality, and then updates the slowio unit tests to run in accelerated time instead of real time.

There's a new `meta-timeofday.testc` suite that tests the unit-timeofday mocked time functionality, which adds about 9 seconds of runtime to `make check`, but the `slowio.testc` tests are now about 27 seconds faster, so the net benefit is decent.

**There might be compiler/platform-specific bugs in the libc interception code.**  Please fetch this branch and test it locally!


lib/util.h
```diff
@@ -558,4 +558,18 @@ enum logfmt_type {
...
+ * tests can mock cyrus_gettime if they need to fake the passing of time.
+ *
+ * We need this shim because, unlike gettimeofday, clock_gettime itself is
+ * not a weak alias, so it can't be overridden directly.
+ */
+static int wrap_clock_gettime(clockid_t id, struct timespec *ts)
+{
+    return clock_gettime(id, ts);
+}
+__attribute__((weak, alias("wrap_clock_gettime"), visibility("default")))
```
wolfsage, line 572: TIL weak aliases. Very neat

cunit/unit-timeofday.c
```diff
@@ -234,124 +248,152 @@ static int real_gettimeofday(struct timeval *tv, ...)
 #endif
 }
 
-/* provide a function to hide the libc weak alias */
-EXPORTED int gettimeofday(struct timeval *tv, ...)
+EXPORTED time_t real_time(time_t *tp)
 {
-    to_timeval(transform(now()), tv);
-    return 0;
+    /* XXX can't find the name of the real function to wrap, so mimic it... */
```
wolfsage, line 253: I don't understand this comment, can you explain?

Also why is it a TODO?

cunit/unit.h
```diff
@@ -217,6 +218,74 @@ extern int __cunit_wrap_fixture(const char *name, int (*fn)(void));
...
+        "CU_ASSERT_DOUBLE_EQUAL(%s=%.17g,%s=%.17g,%s=%.17g)",           \
+        #actual, _a, #expected, _e, #granularity, _g);                  \
+} while (0)
+
+#undef CU_ASSERT_DOUBLE_EQUAL_FATAL
+#define CU_ASSERT_DOUBLE_EQUAL_FATAL(actual,expected,granularity) do {  \
+    double _a = (actual), _e = (expected), _g = (granularity);          \
+    CU_assertFormatImplementation((fabs(_a - _e) <= fabs(_g)),          \
+        __LINE__, __FILE__, "", CU_TRUE,                                \
+        "CU_ASSERT_DOUBLE_EQUAL(%s=%.17g,%s=%.17g,%s=%.17g)",           \
```
wolfsage: The same output message has been copied for these 4 `define`s but they should all be distinct and match the function they are from

wolfsage (changes requested):
I think the CU_ASSERT* override defines need their error messages adjusted, and I had one question about a new TODO item. Otherwise seems good and works on my hardware at least

cunit/unit-timeofday.c
elliefm: What I actually want to do here is call through to the real glibc function, like what `real_gettimeofday()` and such do, with the big mess of ifdefs to choose the correct internal name.

But for this one and one other (iirc), I couldn't find the right name to invoke the real function after a reasonable amount of exploration, so I gave up and just implemented something equivalent instead.

cunit/unit.h
elliefm: Oh, good eye.  I thought I'd noticed that myself and fixed it already, but maybe I lost the fix later while rebasing.  Fixed again!

wolfsage (approved):

rjbs:
@elliefm re: valgrind:
```
configure:30215: checking for valgrind/valgrind.h
configure:30215: gcc -c  -g -fPIC -W -Wall -Wextra -Werror -Wwrite-strings -Wformat=2    conftest.c >&5
configure:30215: $? = 0
configure:30215: result: yes
```

rjbs:
As you suggested, this is on my Docker container, running Debian, under OrbStack, on macOS.

I don't have this problem with Docker on native Linux.  There may be a problem or slowdown in the hypervisory layer.

rjbs:
(I'm not running the cunit tests with VG=1 fwiw)

PR #5608, opened 2025-09-08
mailbox: all system users should use internal seen

brong:
This includes admin, anyone, anonymous, etc.


brong:
@ksmurchison I'm sending this for re-review because I added a config option along with the docs.  When I considered the upgrade path, I realised people might have admin tools that do weird stuff!  Dammit.

PR #5610, opened 2025-09-09
jmap_api.c, jmap_contact.c: implement applyEmptyUpdates for cards

ksmurchison:


brong (approved):
Content of the changes approved.  The commit message and/or PR could contain some more detail about why we're making this change and how it's supposed to be used.

I spent some time understanding the 'vcard' object just being kept rather than converted, destroyed, recreated - but it makes sense.

PR #5614, opened 2025-09-12
add traceid for logging

elliefm:
**Reviewers:** I'd suggest reading this commit-by-commit.

This PR adds an API for managing trace id's, with an LMTP capability and command such that a client (MTA) can provide a trace id to associate with the delivery.  `xsyslog()` and `xsyslog_ev()`  now automatically log the sessionid and traceid if they are set, and most LMTP logging has been converted to use `xsyslog()` to take advantage of this.  Audit logging has not been changed; I expect to revisit that separately later.

There's also some cleanup and restructure of the existing sessionid code, in support of these changes.

### DIFF
--- a/changes/next/cyr-1402-traceid
+++ b/changes/next/cyr-1402-traceid
@@ -1,0 +1,26 @@
+Description:
+
+* Fixed: lmtpd generated a new session id after LHLO response
+* Adds lmtp `TRACE` capability and `TRACE trace-id` command.  The supplied
+  trace-id will be included in all log lines related to this delivery, until
+  the next `RSET`.
+
+
+Documentation:
+
+None
+
+
+Config changes:
+
+None
+
+
+Upgrade instructions:
+
+Nothing required.
+
+
+GitHub issue:
+
+None


cassandane/Cassandane/Cyrus/Annotator.pm
```diff
@@ -504,9 +504,13 @@ sub test_log_missing_acl
         $self->{store}->set_folder('Other Users.other');
         $self->make_message("Email", body => "set_flag $_->{flag}\r\n",
             store => $self->{store}) or die;
-        my $wantLog = "could not write flag due missing ACL: "
-                    . "flag=<\\$_->{flag}> need_rights=<$_->{need_rights}>";
-        $self->assert_syslog_matches($self->{instance}, qr{$wantLog});
+        my $wantLog = qr{
+            could\snot\swrite\sflag\sdue\smissing\sACL:
+            (?:\s\w+=<[^>]*>)*
```
elliefm: Oh, this won't skip past `t.id=<...>` because `.` is not `\w`!  These will need to be `\s[^=]+=...` probably...

imap/lmtp_sieve.c
```diff
@@ -2258,7 +2273,7 @@ static int jmapquery(void *ic, void *sc, void *mc, const char *json)
             priority = LOG_ERR;
         }
 
-        syslog(priority, "sieve: jmapquery error: %s", errstr);
+        xsyslog(priority, "jmapquery error", "error=<%s>", errstr);
```
wolfsage: dropping sieve from this one fields weird?

imap/lmtp_sieve.c
```diff
@@ -524,8 +525,7 @@ static int send_rejection(const char *userid,
         r = smtpclient_send(sm, &sm_env, &msgbuf);
     }
     if (r) {
-        syslog(LOG_ERR, "sieve: send_rejection: SMTP error: %s",
-                error_message(r));
+        xsyslog(LOG_ERR, "SMTP error", "error=<%s>", error_message(r));
```
rjbs: This has lost the strings `sieve` and `send_rejection`, both of which were kind of useful in contextualizing what happened.  On the other hand, this is the only "SMTP error" in the whole source.  I think "sieve" has been retained enough places that, unless it was a mistake, "sieve SMTP error" would save some future debugging time.

(I reckon much of this may change in the future as we xsyslog_ev this, but…)

rjbs (approved):
This looks great.  I made one small request for change which I can live without, or can adjust in the future when we ev-ify some of this.  Or maybe don't need.  But I think it's a good suggestion. :)

lib/util.c
```diff
@@ -2200,12 +2201,23 @@ EXPORTED void xsyslog_fn(int priority, const char *description,
...
 
     buf_appendcstr(&buf, description);
     buf_appendmap(&buf, ": ", 2);
+    if (session_have_id()) {
+        buf_appendmap(&buf, "sessionid=<", 11);
+        buf_appendcstr(&buf, session_id());
+        buf_appendmap(&buf, "> ", 2);
+    }
+    if (traceid) {
+        buf_appendmap(&buf, "t.id=<", 6);
```
rjbs: This is 100% my fault, but I got this wrong in writing up the issue.  This should be `r.tid` to match the Fastmail standard.  Almost doesn't matter, since we'll rename a bunch of fields when we eventify things.

But probably 10m to change this, tops, so I am requesting this change.

imap/lmtp_sieve.c
elliefm: This is converting from syslog to xsyslog, so it'll have the automatic extra context that xsyslog provides.  Since it's LOG_ERR (i.e. not LOG_NOTICE or LOG_INFO), xsyslog will add `func=<send_rejection>` automatically.  This is the only `send_rejection()` function in the whole source.

The usual format of xsyslogged lines is `[prelude] description: key=<value>...`.  I dropped the "sieve:" because I've been trying not to embed `:` characters into the descriptions, in case it confuses parsers.  Though I don't know that it matters.

I thought xsyslog also logged the source filename (in this case, that would be "lmtp_sieve.c"), but it turns out it doesn't.  That would not be hard to add (as a separate task) if we think it's useful.  But it also seems obvious, so perhaps the reason it's not just there already is that we already considered it and decided against it.

lib/util.c
elliefm: Yeah great, will do

imap/lmtp_sieve.c
elliefm: I said the same thing in my reply to Rik's comment, but I've been trying to not have `:` embedded in the xsyslog description string in case it confuses parsers that expect a single `description:` chunk in the log line.  I don't know if that matters.

imap/lmtp_sieve.c
elliefm: This is now "sieve SMTP error"

lib/util.c
elliefm: Done!

imap/lmtp_sieve.c
elliefm: This is now "sieve jmapquery error"

rjbs (approved):

PR #5615, opened 2025-09-12
imapd: treat 'replicaonly' the same as 'readonly'

brong:
It's never going to be a good idea to run commands that can't be run in readonly when you're on a replica server.


brong:
CI now passes with some test improvements :)

ksmurchison (approved):

PR #5618, opened 2025-09-19
add an auditlog api

elliefm:
**Reviewers:** I suggest reading this commit-by-commit.

This adds an auditlog API, providing a handful of real (type-checked!) functions for the various patterns we auditlog, and updates all existing audit logging to use it.  I think we will find it is much easier to use and maintain than syslog or xsyslog were.

The output should be mostly unchanged, except that:

* trace-id is now audit logged as `r.tid=<...>` if it was set  (see #5614)
* the `unlink` and `unlinkarchive` lines from `mailbox_cleanup_uid()` are now `remove-spool` and `remove-archive` respectively, so they don't get conflated with the `unlink` lines from `mailbox_commit()`
* some fields may be output in a different order from before
* some extra fields may be output that weren't before
* some inconsistencies have been brought into line

I thought I was going to implement this as macros around xsyslog, but that turned out to be a silly direction.  Instead, the `auditlog_foo()` functions are implemented in terms of a lightweight internal API that will be easy to modify or replace if we want to change the output format in the future.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Delivery.pm
+++ b/cassandane/Cassandane/Cyrus/Delivery.pm
@@ -623,7 +623,7 @@ sub test_auditlog_size
     xlog $self, "Check the correct size was auditlogged";
     if ($self->{instance}->{have_syslog_replacement}) {
         my @appends = $self->{instance}->getsyslog(
-            qr/auditlog: append: .* uid=<1>/);
+            qr/auditlog: append .* uid=<1>/);
         $self->assert_num_equals(1, scalar @appends);
 
         # delivery will add some headers, so it will be larger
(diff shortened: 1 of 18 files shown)


elliefm:
This probably wants #5614 to have already landed

ksmurchison (approved):
This looks great.  Thanks for this!

cassandane/Cassandane/Cyrus/Delivery.pm
```diff
@@ -623,7 +623,7 @@ sub test_auditlog_size
     xlog $self, "Check the correct size was auditlogged";
     if ($self->{instance}->{have_syslog_replacement}) {
         my @appends = $self->{instance}->getsyslog(
-            qr/auditlog: append: .* uid=<1>/);
+            qr/auditlog: append .* uid=<1>/);
```
rjbs, line 626: Fortunately we just made Fastmail log parsing agnostic about whether this `:` is required. :-D

PR #5621, opened 2025-09-23
MUPDATE: add MAILBOX-EXTENDED capability and functionalty

ksmurchison:
This extends the ACTIVATE command to accept " (JMAPID xxx)" as its last argument,
and extends the MAILBOX response to return " (JMAPID xxx)" as its last argument (iff the client has sent ENABLE MAILBOX-EXTENDED)

This also needs a changes file and some documentation once the guts are approved.

### DIFF
--- a/imap/ctl_mboxlist.c
+++ b/imap/ctl_mboxlist.c
@@ -113,6 +113,7 @@ struct mb_node
 {
     char mailbox[MAX_MAILBOX_BUFFER];
     char location[MAX_MAILBOX_BUFFER];
+    char jmapid[JMAP_MAX_MAILBOXID_SIZE];
     char *acl;
     struct mb_node *next;
 };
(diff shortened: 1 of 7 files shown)


imap/mupdate-client.c
```diff
@@ -88,6 +89,17 @@ static struct protocol_t mupdate_protocol =
...
+/* We're really only looking for an OK or NO or BAD here -- and the callback
+ * is never called in those cases.  So if the callback is called, we have
+ * an error! */
+static int mupdate_scarf_one(struct mupdate_mailboxdata *mdata __attribute__((unused)),
+                             const char *cmd,
+                             void *context __attribute__((unused)))
+{
+    syslog(LOG_ERR, "mupdate_scarf_one was called, but shouldn't be.  Command received was %s", cmd);
+    return -1;
+}
```
elliefm, line 101: Might be nice to assert in here that it is actually an error.  If someone changes the behaviour elsewhere, such that this starts getting called for OK/etc, the subsequent assertion failure will remind them to update this as well, whereas the log line might go unnoticed, and the -1 return might be hard to debug.

elliefm (commented):
Looks okay so far.

Is it possible to test with/without the new capability in some way?

I think a cross-version Cassandane run of the existing Murder tests, with a 3.12 build and a master build, would cover that.  But CI doesn't do cross-version testing (yet...).

imap/ctl_mboxlist.c
```diff
@@ -219,7 +223,7 @@ static int pop_mupdate_cb(const mbentry_t *mbentry, void *rockp)
             /* No need to update mupdate NOW, we'll get it when we
              * untag the mailbox */
             skip_flag = 1;
-        } else if (act_head->acl) {
+        } else if (act_head->acl && mbentry->jmapid) {
```
wolfsage, line 226: I think this change is "If it doesn't have a jmapid, we don't want to skip so that updating it gives it one"?

wolfsage (commented):
I see some murder tests in cassandane, would it be hard to add tests for this?

imap/ctl_mboxlist.c
ksmurchison: Correct

PR #5624, opened 2025-09-25
http_caldav.c: place a limit on recurrence expansion

ksmurchison:
This could probably use a battery of tests to check all combinations of RRULE frequencies and BY* parts, but I have higher priority tasks.

### DIFF
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -18,6 +18,7 @@
 
 #include <sysexits.h>
 #include <syslog.h>
+#include <float.h>
 
 #include <libical/ical.h>
 #include <libxml/tree.h>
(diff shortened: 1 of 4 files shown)


imap/http_caldav.c
```diff
@@ -5446,11 +5447,292 @@ static int expand_cb(icalcomponent *comp,
...
+    case ICAL_YEARLY_RECURRENCE:
+    case ICAL_MONTHLY_RECURRENCE:
+    case ICAL_WEEKLY_RECURRENCE:
+    case ICAL_DAILY_RECURRENCE:
+        switch (rrule->freq) {
+        case ICAL_YEARLY_RECURRENCE:
+            period = interval * AVE_DAYS_PER_YEAR;
+
+            if (bysetpos)       nrecur = bysetpos;
+            else if (byyearday) nrecur = byyearday;
```
rjbs: Around here, I get to "I need to know more about how the `icalrecurrencetype` type works.  The actual BYSETPOS value can be negative, but maybe `.size` means something else.  Reading icalrecur.h didn't help much.

I think I'll want to get some time from you to ask some questions, then I'll review again and write that battery of tests.

rjbs:
@ksmurchison I *finally* got back to my test update for this.  All the tests I wanted to throw at it passed.  But then I went further:

1. made it handle multiple RRULEs (ugh)
2. extracted the predictor out into ical_support
3. …so we could use it for JMAP also
4. and while I was in there, enforced the jmap maximum query duration

rjbs (approved):

PR #5625, opened 2025-09-29
refactor logfmt code into reusable module

elliefm:
**Reviewers:** I suggest reading this commit-by-commit.

This PR pulls the guts of `xsyslog_ev()` into a reusable low-level logfmt API, and updates `xsyslog_ev()` to use it.  It also updates the implementation to match the current Fastmail spec (which will be published somewhere TBD), and adds support for utf8 value strings.

The general premise is that you initialise a logfmt object, push key-value pairs into it, extract a cstring that can be syslogged, and then clean up the object.

You usually wouldn't use this module directly for routine logging.  Rather, it's for building domain-specific higher level logging functions on top of.   #5637 (draft) will add one such for login logs; I expect to update auditlog to use this internally too.  For one-shot logging, use `xsyslog_ev()`, or make your own local logfmt-based logging function and call that.

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -1560,6 +1560,9 @@ EXPORTED void _xsyslog_ev(int saved_errno, int priority, const char *event,
         case LF_S:
             logfmt_push(&lf, name, arg->data[i].s);
             break;
+        case LF_UTF8:
+            logfmt_push_utf8(&lf, name, arg->data[i].s);
+            break;
         case LF_RAW:
             logfmt_push(&lf, name, arg->data[i].s);
             free((char *)arg->data[i].s);
--- a/cunit/xsyslog_ev.testc
+++ b/cunit/xsyslog_ev.testc
@@ -294,6 +294,84 @@ static void test_lf_s(void)
     CU_ASSERT_SYSLOG(/*all*/0, 1);
 }
 
+static void test_lf_utf8(void)
+{
+    const char *third = "three";
+
+    // simple
+    CU_SYSLOG_MATCH("event=\"lf_utf8 test\" first=one second=two third=three");
+    xsyslog_ev(LOG_ERR, "lf_utf8 test",
+        lf_utf8("first", "one"), lf_utf8("second", "two"), lf_utf8("third", third));
+    CU_ASSERT_SYSLOG(/*all*/0, 1);
+
+    // empty string
+    CU_SYSLOG_MATCH("event=\"lf_utf8 test\" first=\"\"");
+    xsyslog_ev(LOG_ERR, "lf_utf8 test", lf_utf8("first", ""));
+    CU_ASSERT_SYSLOG(/*all*/0, 1);
+
+    // spaces
+    CU_SYSLOG_MATCH("event=\"lf_utf8 test\" first=\"with spaces\"");
+    xsyslog_ev(LOG_ERR, "lf_utf8 test", lf_utf8("first", "with spaces"));
+    CU_ASSERT_SYSLOG(/*all*/0, 1);
+
+    // tabs
+    CU_SYSLOG_MATCH("event=\"lf_utf8 test\" first=\"with\\\\ttabs\"");
+    xsyslog_ev(LOG_ERR, "lf_utf8 test", lf_utf8("first", "with\ttabs"));
+    CU_ASSERT_SYSLOG(/*all*/0, 1);
+
+    // quotes
+    CU_SYSLOG_MATCH("event=\"lf_utf8 test\" first=\"\\\\\"quotes\\\\\"\"");
+    xsyslog_ev(LOG_ERR, "lf_utf8 test", lf_utf8("first", "\"quotes\""));
+    CU_ASSERT_SYSLOG(/*all*/0, 1);
+
+    // \r and \n
+    CU_SYSLOG_MATCH("event=\"lf_utf8 test\" first=\"\\\\r\\\\n\"");
+    xsyslog_ev(LOG_ERR, "lf_utf8 test", lf_utf8("first", "\r\n"));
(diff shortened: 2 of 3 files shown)


lib/logfmt.c
```diff
@@ -131,3 +137,62 @@ EXPORTED char *logfmt_escape(const char *val)
...
+
+    buf_free(&formatted);
+}
+
+EXPORTED void logfmt_push_session(struct logfmt *lf)
+{
+    const char *traceid = trace_id();
+
+    if (session_have_id()) {
+        logfmt_push(lf, "session_id", session_id());
```
elliefm: @rjbs: I spelled this `"session_id"` because that's what you used for the logfmt login logs recently, but it's different from the `"sessionid"` spelling that we use elsewhere.  I'm assuming your use of underscore was a deliberate choice to bring Cyrus logfmt in line with other FM stuff, but if it was actually just an accident and `"sessionid"` would be correct, please lmk! :)

lib/logfmt.c
```diff
@@ -156,6 +162,113 @@ EXPORTED void logfmt_escape(struct buf *buf, const char *val)
...
+    need_dquotes = val_needs_dquotes(utf8val);
+
+    buf_reset(buf);
+    buf_ensure(buf, strlen(utf8val) + 2 * need_dquotes);
+    if (need_dquotes) buf_putc(buf, '"');
+
+    while ((uint32_t) U_SENTINEL != (cp = utext_next32(&ut))) {
+        uint8_t scratch[8] = {0};
+        unsigned i, len = 0;
+        bool is_error = false;
```
rsto: nit: the argument types for `U8_APPEND` are `UBool` for `isError`. The 2nd and 3rd numeric arguments for buffer length and capacity are `int32_t`

rsto (approved):
LGTM

lib/logfmt.c
elliefm: Where did you see that?  Whatever documentation I was looking at last week showed it as a macro with no particular types, but this is my first time working with ICU, perhaps I was looking in the wrong place?

rsto: I looked here https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/utf8_8h.html#a8a013997e3d13c501ce73bc45e662881

elliefm: Ahh, I think that's the same documentation I looked at, but I guess I got distracted looking at the macro implementation, and didn't notice the types specified below.  I guess the types are specified in case it's ever implemented as a function rather than a macro.  I'll update to match.  Good catch, thanks!

lib/util.c
```diff
@@ -1532,76 +1532,70 @@ EXPORTED char *modseqtoa(modseq_t modseq)
...
-        case LF_LLU: buf_printf(&buf, "%llu", arg->data[i].llu); break;
-        case LF_ZD:  buf_printf(&buf, "%zd",  arg->data[i].zd);  break;
-        case LF_ZU:  buf_printf(&buf, "%zu",  arg->data[i].zu);  break;
-        case LF_LLX: buf_printf(&buf, "%llx", arg->data[i].llu); break;
-        case LF_F:   buf_printf(&buf, "%f",   arg->data[i].f);   break;
-        case LF_M: {
-            char *escaped_errno = logfmt_escape(strerror(saved_errno));
-            buf_appendcstr(&buf, escaped_errno);
-            free(escaped_errno);
+        case LF_C:
```
wolfsage: Nit: Format wise I think keeping these all on one line was way easier to scan/read, but I won't make a stink about it

lib/logfmt.c
```diff
@@ -0,0 +1,344 @@
...
+            break;
+        case '\"': /* 0x22 */
+            buf_appendmap(buf, "\\\"", 2);
+            break;
+        case '\\': /* 0x5C */
+            buf_appendmap(buf, "\\\\", 2);
+            break;
+
+        /* escaped-hex, eightbit */
+        case 0x00 ... 0x08:
```
wolfsage, line 146: This is not your bug, but I realize as written "logfmt_escape_bytestring" cannot be used with data with embedded nulls.

Don't know if you want to think of a solution for that yet or we can do something about it when we come to it. (For now you can use lf_raw(...) with your own formatting at least).

But `lf_b("key", bytes, bytelen)` could be useful ...

elliefm: Already thought about this a bunch, and decided it wasn't worth pursuing until we're actually trying to log binary data.

At that point, we'd probably add `logfmt_push_binary()` and `lf_b()` implementations, and I suspect probably force the binary data into base64 anyway, in which case we still wouldn't need to deal with embedded nulls in the escape functions.  But if we didn't base64, then we'd tweak the escape functions to take a length parameter.  Easy enough, just no benefit right now.

lib/logfmt.c
rjbs: Let's use `sessionid` as we did, my introduction of `_` was not intentional.

lib/util.c
elliefm: Hmm.  I thought they became too long for one line, but it doesn't look like it now.  Maybe they became too long so I wrapped them in an intermediate step which later got squashed out.  I'll try collapsing them back to one line and see how it looks.  If nothing else the diff should be easier to read if they remain as single lines.

elliefm: Ahh yeah the lines are still too long... but only by two chars in the worst case.  I could fix that by renaming `log_push_fmt()` to like `log_pushf()`.  That'd be consistent with the usual `printf` naming scheme.

elliefm: I've done this, it looks better, thanks for the prompt

lib/logfmt.c
elliefm: Done

PR #5626, opened 2025-09-29
util.c: do not segv on null string

rjbs:
Before this commit:

    xsyslog_ev(...,
               LF_S("key", null_string));

would produce a segmentation violation.  This commit changes it to format the string as "~null~", mirroring our Perl implementation's "~missing~".  We can refine this later, but this is better than a segv.


wolfsage:
Looks like I copied the code from an older version of some of the initial experiments before it was rewritten to use a buf

elliefm:
I'm working in this space myself already, so I'll make the buffer usage more cromulent while I'm in there

PR #5628, opened 2025-09-30
jmap_mail_query.c: don't hardcode user prefix for shared addrbook

ksmurchison:
mailbox pattern.  Instead use prefix in the given namespace

### DIFF
--- a/imap/jmap_mail_query.c
+++ b/imap/jmap_mail_query.c
@@ -148,8 +148,10 @@ static mbentry_t *_get_sharedaddressbook(const char *userid,
 
     strarray_t patterns = STRARRAY_INITIALIZER;
     struct buf pattern = BUF_INITIALIZER;
-    buf_setcstr(&pattern, "user");
-    buf_putc(&pattern, namespace->hier_sep);
+
+    assert(namespace->hier_sep && namespace->prefix[NAMESPACE_USER][0]);
+
+    buf_setcstr(&pattern, namespace->prefix[NAMESPACE_USER]);
     buf_putc(&pattern, '*');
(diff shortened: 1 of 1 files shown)


imap/jmap_mail_query.c
```diff
@@ -148,8 +148,7 @@ static mbentry_t *_get_sharedaddressbook(const char *userid,
 
     strarray_t patterns = STRARRAY_INITIALIZER;
     struct buf pattern = BUF_INITIALIZER;
-    buf_setcstr(&pattern, "user");
-    buf_putc(&pattern, namespace->hier_sep);
+    buf_setcstr(&pattern, namespace->prefix[NAMESPACE_USER]);
```
rsto, line 154: Is this safe to assume to be not NULL for all namespaces? Maybe either assert or fall back to setting "user" in case of NULL?

ksmurchison: The same could be asked about namespace->hier_sep.  If its \0, I'm guessing that all hell will break loose

rsto: Yep, that's what I thought, too. But now that we do update that code, should we add an assert?

ksmurchison: The prefixes are char arrays (not char*), so are we thinking 

`assert(namespace->hier_sep && *namespace->prefix[NAMESPACE_USER]);`

rsto: Yeah, I think that's good! It should catch all cases where we might only partially initialize the namespace structs.

ksmurchison: Done

PR #5629, opened 2025-09-30
http_caldav.c: time-range in free-busy-query REPORT is mandatory

ksmurchison:


rsto (commented):
I'm all for making Cyrus more spec-compliant, but I'd like to understand why we are doing this change. Is this to prevent a bug with an existing client? On the flip side, is this change likely to break interoperability with some CalDAV clients which we currently are dealing with fine?
Whatever the answer is, maybe add some comment or commit message that explains the motivation for this change?

ksmurchison:
@rjbs I don't recall why we decided to do this?  Was it just to match the spec?

dilyanpalauzov:
> The request body MUST be a CALDAV:free-busy-query XML element … , which MUST contain exactly one CALDAV:time-range XML element ….

I think the current code requires AT LEAST ONE `CALDAV:time-range` XML element.  It also allows in the first `time-range` element to set start, in the second `time-range` element to set end, and then assumes that only one `time-range` element was present, which contained both start and end.

In addition after `start = xmlGetProp(node, BAD_CAST "start");` also the namespace of the attribute can be enforced.

imap/http_caldav.c
```diff
@@ -8029,14 +8051,26 @@ static int report_fb_query(struct transaction_t *txn,
...
+
+        fbfilter.start =
+            icaltime_from_timet_with_zone(caldav_epoch, 0, utc_zone);
+    }
+    else if (icaltime_is_null_time(fbfilter.end)) {
+        fbfilter.end =
+            icaltime_from_timet_with_zone(caldav_eternity, 0, utc_zone);
+    }
+
+    if (!is_valid_timerange(fbfilter.start, fbfilter.end)) {
```
dilyanpalauzov, line 8070: I think here `else if` can be used - `is_valid_timerange(fbfilter.start, fbfilter.end)` is executed only if the `time-range` filter contained both `start` and `end`.

PR #5630, opened 2025-10-01
JMAP: remove Foo/touch methods

ksmurchison:
These methods turned out to be not as useful as initially intended


rsto:
CI does not pass. I'll re-review when it does.

rsto (approved):

PR #5631, opened 2025-10-02
JMAP: bump modseq when "updating" externally stored properties

ksmurchison:
This will allow the server to signal to the client via /changes that it needs to sync the resource

### DIFF
--- a/imap/jmap_calendar.c
+++ b/imap/jmap_calendar.c
@@ -940,32 +940,32 @@ static const jmap_property_t calendar_props[] = {
     {
         "syncedFrom",
         JMAP_CALENDARS_EXTENSION,
-        0
+        JMAP_PROP_EXTERNAL
     },
     {
         "isEventsPublic",
         JMAP_CALENDARS_EXTENSION,
-        0
+        JMAP_PROP_EXTERNAL
     },
     {
         "isFreeBusyPublic",
         JMAP_CALENDARS_EXTENSION,
-        0
+        JMAP_PROP_EXTERNAL
     },
     {
         "eventsUrl",
         JMAP_CALENDARS_EXTENSION,
-        JMAP_PROP_SERVER_SET
+        JMAP_PROP_SERVER_SET | JMAP_PROP_EXTERNAL
     },
     {
         "freeBusyUrl",
         JMAP_CALENDARS_EXTENSION,
-        JMAP_PROP_SERVER_SET
+        JMAP_PROP_SERVER_SET | JMAP_PROP_EXTERNAL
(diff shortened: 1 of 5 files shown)


imap/jmap_calendar.c
```diff
@@ -1201,6 +1201,7 @@ struct setcalendar_props {
     int isVisible;
     int isSubscribed;
     int transp;
+    bool ext_props; // request is setting externally-stored properties
```
rsto: This could be more clearly named as a boolean, e.g. "sets_ext_props" or "has_ext_props"

imap/jmap_calendar.c
```diff
@@ -1276,6 +1277,17 @@ static void setcalendar_parseprops(jmap_req_t *req,
...
         props->transp = -1;
         props->comp_types = -1;
+
+        /* Any externally-stored properties? */
+        const char *key;
+        json_t *val;
+        json_object_foreach(arg, key, val) {
+            const jmap_property_t *prop = jmap_property_find(key, calendar_props);
+            if (prop && (prop->flags & JMAP_PROP_EXTERNAL)) {
+                props->ext_props = 1;
```
rsto: nit: here and below the `bool` value is set using numerics, rather than `true` and `false`. We don't have a guideline for this, but we might want to always use `true` and `false` to make the value type of the assigned variable clear. Also, in C23 they became keywords!

rsto (commented):
Initial review while CI runs: this looks good feature-wise but I would prefer some minor code updates.

imap/jmap_calendar.c

imap/jmap_calendar.c
ksmurchison: Fixed

imap/jmap_mailbox.c
```diff
@@ -2790,29 +2804,42 @@ static void _mbox_update(jmap_req_t *req, struct mboxset_args *args,
...
+            if (args->is_seenshared >= 0) {
+                newopts = mbox->i.options;
+                if (args->is_seenshared) newopts |= OPT_IMAP_SHAREDSEEN;
+                else newopts &= ~OPT_IMAP_SHAREDSEEN;
+
+                /* only mark dirty if there's been a change */
+                if (mbox->i.options != newopts) {
+                    mailbox_index_dirty(mbox);
+                    mbox->i.options = newopts;
+                    bump_modseq = 1;
```
rsto: nit: forgot to rename this to `true`

ksmurchison: Fixed

PR #5633, opened 2025-10-02
Cyr 2196 abort if conversations open

brong:
This ensures that we catch the "calalarmd holds conversations open" bug if it happens again!

I have slipstreamed in my two "make it compile on Ubuntu 24.04 on WSL" commits as well.

### DIFF
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -246,7 +246,9 @@ EXPORTED void flags_to_str(const struct index_record *record, char *flagstr)
 
 EXPORTED int open_mailboxes_exist()
 {
-    return open_mailboxes ? 1 : 0;
+    if (open_mailboxes) return 1;
+    if (open_conversations_exist()) return 1;
+    return 0;
 }
 
 EXPORTED int open_mailboxes_namelocked(const char *userid)
--- a/imap/message.c
+++ b/imap/message.c
@@ -1226,7 +1226,7 @@ EXPORTED void message_parse_string(const char *hdr, char **hdrp)
     while ((he = strchr(he, '\n'))!=NULL) {
         if (he > *hdrp && he[-1] == '\r') {
             he--;
-            memmove(he, he+2, strlen(he+2)+1);
+            if (he) memmove(he, he+2, strlen(he+2)+1);
         }
         else {
             memmove(he, he+1, strlen(he+1)+1);
--- a/imap/conversations.c
+++ b/imap/conversations.c
@@ -369,6 +369,11 @@ EXPORTED size_t conversations_estimate_emailcount(struct conversations_state *st
     return count;
 }
 
+EXPORTED int open_conversations_exist(void)
+{
+    return open_conversations ? 1 : 0;
+}
+
 EXPORTED int conversations_open_path_version(const char *fname,
                                              const char *userid, int shared,
                                              struct conversations_state **statep,
--- a/imap/conversations.h
+++ b/imap/conversations.h
@@ -237,6 +237,8 @@ extern void conversations_set_suffix(const char *suff);
 extern char *conversations_getmboxpath(const char *mboxname);
 extern char *conversations_getuserpath(const char *username);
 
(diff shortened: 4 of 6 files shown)


imap/mailbox.c
```diff
@@ -246,7 +246,9 @@ EXPORTED void flags_to_str(const struct index_record *record, char *flagstr)
 
 EXPORTED int open_mailboxes_exist()
 {
-    return open_mailboxes ? 1 : 0;
+    if (open_mailboxes) return 1;
+    if (open_conversations_exist) return 1;
+    return 0;
```
ksmurchison, line 251: This should be `if (open_conversations_exist()) return 1;`

ksmurchison (changes requested):
I assume this is why CI fails to compile

imap/mailbox.c
brong: Oh yeah!

elliefm:
@brong I cancelled the CI for this because it had been running for 3h and wasn't getting anywhere.  Looks like something was broken in a way that confused Cassandane really badly.  Hundreds of [ERROR] and [FAILED] tests, and then it hung on something and never managed to clean up and exit.

elliefm:
Will pull this branch now and see if I can repro locally...

elliefm:
Can reproduce locally (Ubuntu 24.04, Intel):

Most (all?) of the FastMail and JMAPFoo tests error like this:

> JMAP request for cassandane failed (http://127.0.0.1:29224/jmap/): 500 Internal Server Error: Fatal error: Internal error: assertion failed: imap/jmap_api.c: 397: !open_mailboxes_exist()

In JMAPTestSuiteWS, a bunch of the Email:get:* tests hang, so that explains why the CI run didn't finish.  I assume this has the same underlying cause, and the hang is some part of JMAP-TestSuite or its cassandane wrapper not coping well with the failure.

imap/message.c
```diff
@@ -1224,7 +1224,7 @@ EXPORTED void message_parse_string(const char *hdr, char **hdrp)
     /* Un-fold header (overlapping buffers, use memmove) */
     he = *hdrp;
     while ((he = strchr(he, '\n'))!=NULL) {
-        if (he > *hdrp && he[-1] == '\r') {
+        if (he-1 > *hdrp && he[-1] == '\r') {
```
ksmurchison: I'm thinking this should be 

`if (he-1 >= *hdrp && ...`

Or maybe just change the original to 

`if (he > *hdrp && *(he-1) == '\r')`

ksmurchison:
I've confirmed that its the change to 

> Can reproduce locally (Ubuntu 24.04, Intel):
> 
> Most (all?) of the FastMail and JMAPFoo tests error like this:
> 
> > JMAP request for cassandane failed (http://127.0.0.1:29224/jmap/): 500 Internal Server Error: Fatal error: Internal error: assertion failed: imap/jmap_api.c: 397: !open_mailboxes_exist()
> 
> In JMAPTestSuiteWS, a bunch of the Email:get:* tests hang, so that explains why the CI run didn't finish. I assume this has the same underlying cause, and the hang is some part of JMAP-TestSuite or its cassandane wrapper not coping well with the failure.

I've confirmed that its the change to open_mailboxes_exist() that causes the failures, not the change to message_parse_string().  I do believe that there is a logic change in the latter function that needs to be resolved (see below).

imap/message.c
brong: The issue isn't this, it's that the pointer math could theoretically return zero and memmove is unhappy about that analysis.  The only way this commit could change anything is if a string STARTED with "<anychar>\n" then it would fall through to the earlier wrapping logic and not remove the \r.

ksmurchison: Do we still want to fix this logic?  `he > *hdrp` should be `he-1 >= *hdrp`, correct?

brong: I fixed it with a different approach.

ksmurchison (approved):

PR #5634, opened 2025-10-03
move (most of) struct buf to its own module

elliefm:
This resolves some annoying header dependency issues, and will probably allow resolving some others.  Specifically, it allows headers that require `struct buf` to just include buf.h, without needing to pull in all of util.h and its dependencies.

util.h now includes buf.h, so in normal usage nothing much changes.

There's a few buf_foo functions I left behind in util.h.  These ones depend on data types that are only available with util.h and/or config.h.

@dilyanpalauzov started something similar at #5108, but that PR only moves the header declarations.  This PR moves the implementations too, and updates the libcyrus examples to match.

libconfig.h and message_guid.h are now installed, and I've moved `dir_hash_c()` and `dir_hash_b()` to libconfig so they can used without util.h, at Dilyan's request.

### DIFF
--- a/lib/util.h
+++ b/lib/util.h
@@ -224,19 +224,6 @@ int strcmpnull(const char *a, const char *b);
 extern keyvalue *kv_bsearch (const char *key, keyvalue *kv, int nelem,
                                int (*cmpf)(const char *s1, const char *s2));
 
-/* Examine the name of a file, and return a single character
- *  (as an int) that can be used as the name of a hash
- *  directory.  Caller is responsible for skipping any prefix
- *  of the name.
- */
-extern int dir_hash_c(const char *name, int full);
-/*
- * Like dir_hash_c() but builds the result as a single-byte
- * C string in the provided buffer, and returns the buffer,
- * which is sometimes more convenient.
- */
-extern char *dir_hash_b(const char *name, int full, char buf[2]);
-
 /*
  * create an [unlinked] temporary file and return the file descriptor.
  */
--- a/lib/cyrusdb_quotalegacy.c
+++ b/lib/cyrusdb_quotalegacy.c
@@ -87,6 +87,7 @@
 #include "cyrusdb.h"
 #include "hash.h"
 #include "map.h"
+#include "libconfig.h"
 #include "libcyr_cfg.h"
 #include "cyr_lock.h"
 #include "util.h"
--- a/lib/libconfig.c
+++ b/lib/libconfig.c
@@ -1510,3 +1510,46 @@ EXPORTED int config_check_partitions(FILE *user_output)
     free_hash_table(&by_value, &wrap_strarray_free);
     return 0 - found_bad;
 }
+
+/* Examine the name of a file, and return a single character
+ * (as an int) that can be used as the name of a hash
+ * directory.  Stop before the first dot.  Caller is responsible
+ * for skipping any prefix of the name.
+ */
+EXPORTED int dir_hash_c(const char *name, int full)
+{
+    int c;
+
+    if (full) {
+        unsigned char *pt;
+        uint32_t n;
+        enum {
+            DIR_X = 3,
+            DIR_Y = 5,
+            DIR_P = 23,
+            DIR_A = 'A'
+        };
+
+        n = 0;
+        pt = (unsigned char *)name;
+        while (*pt && *pt != '.') {
+            n = ((n << DIR_X) ^ (n >> DIR_Y)) ^ *pt;
+            n &= UINT32_MAX;
+            ++pt;
+        }
+        c = DIR_A + (n % DIR_P);
+    }
+    else {
+        c = tolower(*name);
+        if (!Uisascii(c) || !Uislower(c)) c = 'q';
+    }
+
+    return c;
+}
+
+EXPORTED char *dir_hash_b(const char *name, int full, char buf[2])
+{
+    buf[0] = (char)dir_hash_c(name, full);
+    buf[1] = '\0';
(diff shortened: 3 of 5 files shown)


dilyanpalauzov:
Can also the function declaration `int dir_hash_c(const char *name, int full)` be put under /usr/include/cyrus, e.g in buf.h?

elliefm:
> Can also the function declaration `int dir_hash_c(const char *name, int full)` be put under /usr/include/cyrus, e.g in buf.h?

It's not really anything to do with `struct buf`, so I'd prefer to put it somewhere else.  I'll have a think about where...

Looks like most uses of this function are like `dir_hash_c(..., config_fulldirhash)`, which is to say, you always need to use it knowing how you're configured.  So I think a good spot to stash it will be in libconfig, especially since that will soon be an installed header.

elliefm:
> Apart from splitting struct buf to a separate, installable file buf.h, #5108 does also on make install install lib/libconfig.h and imap/message_guid.h. For me this is useful.

Ah thanks, I saw those changes in your PR and was wondering what they were about.  I'll do the same.

lib/util.h
```diff
@@ -319,80 +307,14 @@ extern int64_t now_ms(void);
...
-const char *buf_base(const struct buf *buf);
-void buf_reset(struct buf *buf);
-void buf_truncate(struct buf *buf, ssize_t len);
-void buf_setcstr(struct buf *buf, const char *str);
-void buf_setmap(struct buf *buf, const char *base, size_t len);
-void buf_copy(struct buf *dst, const struct buf *src);
-void buf_append(struct buf *dst, const struct buf *src);
-void buf_appendcstr(struct buf *buf, const char *str);
-void buf_appendoverlap(struct buf *buf, const char *str);
 void buf_appendbit32(struct buf *buf, bit32 num);
```
rsto: I think the appendbit32/64 should go to `buf.h` as well. This could be accomplished by changing their argument types to `uint32_t` and `uint64_t`. I understand that for `bit64` this will require to decide where to define the shared type definition. As an alternative, I wonder if we could just get rid of the `bit32` and `bit64` type aliases. I never found them particularly useful and I always have to look up their definition to understand their signedness.

lib/util.h
```diff
@@ -319,80 +307,14 @@ extern int64_t now_ms(void);
...
-                __attribute__((format(printf, 2, 0)));
-void buf_printf(struct buf *buf, const char *fmt, ...)
-                __attribute__((format(printf, 2, 3)));
-void buf_replace_buf(struct buf *buf, size_t offset, size_t length,
-                     const struct buf *replace);
-int buf_replace_all(struct buf *buf, const char *match,
-                    const char *replace);
-int buf_replace_char(struct buf *buf, char match, char replace);
 #ifdef ENABLE_REGEX
 int buf_replace_all_re(struct buf *buf, const regex_t *,
```
rsto, line 316: Should the `appendbitXX` functions get refactored to `buf.h` then these seem to be the only ones left "stealing" the `buf_` prefix. We might just rename them, we have tons functions in `charset.h` et al that take a struct buf as argument for the buffer and which do not get prefixed with `buf_`.

rsto (changes requested):
I'm very much in favor of this PR but before approval I would like to first agree on if and how to get rid of the few remaining `buf_` functions in util.h. I submit my review as "Request changes" in the definition that Github gives: "feedback that must be addressed before merging." I do not imply that these functions have to move for my approval.

lib/util.h
elliefm: I think the bit32 and bit64 typedefs were useful before C99 standardised the fixed width types, but I don't think they're so useful anymore.  It wouldn't bother me if we dropped them and just started using uint32_t and uint64_t directly.  But it would be a pretty intrusive find-and-replace change, which I think should be in a standalone PR for ease of code review.

If we assume we'll be doing that at some point, then right now we could either:

1. move the buf_appendbit functions to buf.h, changing their argument types to uint32_t and uint64_t to make it work; or
2. leave them in util.h for now, and move them as part of the greater bit32 change

Do you have an opinion?  I think I prefer 1, but I could go either way.

lib/util.h
elliefm: So, the problem with the `buf_replace_foo_re()` functions is that their declarations are dependent on the `regex_t` typedef.  This is only available when `config.h` is included, because we need to know which regex engine's header to get it from.  Different regex engines typedef it to different types, so we can't just predeclare it ourselves as a workaround without still needing to know which engine.

We don't (and can't) install the `config.h` header, which means we can't install any headers that depend on it.  This is why `util.h` can't just be installed itself.  Large parts of libcyrus.so are broken at the moment because of this (in the sense that they can't be used from outside the cyrus-imapd source tree and build system).  Fixing it all properly would be highly intrusive and a lot of work, but it's not a priority for anyone, so instead I just fix little bits at a time as best I can when they get in the way, like this buf.h work.

The `buf_replace_foo_re()` functions are analogous to the other `buf_replace_foo()` functions, so I think they should keep stealing the `buf_` prefix to preserve that analogy.  One day, maybe, our infrastructure might be fixed enough that they can rejoin their friends.  For now, I'll add a comment explaining why they're hiding in util.h instead of buf.h.  How's that?

lib/util.h
rsto: I agree, I also prefer option 1.

lib/util.h
rsto: Also agreed.

rsto (approved):
Looks good to me!

PR #5636, opened 2025-10-03
Cass: test transformation of vCard group properties (KIND,MEMBER)

ksmurchison:
This will fail CI until our libicalvcard build includes https://github.com/libical/libical/pull/1063


rsto (commented):
I _think_ this is doing the right thing, but before changing the current logic, I would like us to first get a complete understanding of what current Cyrus serves to CardDAV clients for all variants of UID and MEMBER values and v3/v4.

It's very likely the logic proposed in this PR (and libicalvcard) is what we will want to do, but we should be sure to fully cover all required changes in one go. Maybe I've been missing some discussion where we agreed on these already, in that case my apologies. Let's sync on this in today's Cyrus developer meeting.

ksmurchison:
This is now part of #5583

PR #5637, opened 2025-10-08
add a loginlog api

elliefm:


elliefm:
This needs #5634 and #5625 to land first

rjbs:
I gave this a fairly high-level review, rather than any sort of close scrutiny, which I think is not what you want from me here. ;-)  Looks good, and hopefully a nice building block for future work on this!

rjbs (approved):

elliefm:
@wolfsage 

> With this, we _could_ get rid of the beautify_* functions (unless they are exposed as part of the libcyrus api...) if you wanted to delete more code

I was actually very excited to remove them (they're awful IMO), but I felt like it might be a distraction during code review, so I decided to keep the scope narrow and left them alone.  We can always throw them out later (if someone makes a PR to do this, I'll approve it).

They do indeed look like they might be part of the libcyrus API, but they're not, because the only header that knows about them is util.h, and util.h isn't installed.  That's it's own mess, and is the reason we recently moved the struct buf functions out to their own header in #5634.

PR #5638, opened 2025-10-09
auditlog: add and use auditlog_message_touched()

ksmurchison:
This reintroduces "oldflags" in the "touched" log message, which was lost with auditlog_message()

### DIFF
--- a/imap/auditlog.c
+++ b/imap/auditlog.c
@@ -254,6 +254,7 @@ EXPORTED void auditlog_mboxname(const char *action,
 
 EXPORTED void auditlog_message(const char *action,
                                struct mailbox *mailbox,
+                               const struct index_record *oldrecord,
                                const struct index_record *record,
                                const char *message_id)
 {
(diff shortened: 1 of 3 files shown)


imap/auditlog.c
```diff
@@ -252,10 +252,11 @@ EXPORTED void auditlog_mboxname(const char *action,
...
 
-EXPORTED void auditlog_message(const char *action,
-                               struct mailbox *mailbox,
-                               const struct index_record *record,
-                               const char *message_id)
+static void _auditlog_message(const char *action,
+                              struct mailbox *mailbox,
+                              const struct index_record *record,
+                              const struct index_record *oldrecord,
+                              const char *message_id)
```
elliefm: I think it would be fine for this function to still be the public entry point after the change, and just update the other callers of `auditlog_message()` to also pass a NULL `oldrecord`.  We already have a similar thing happening in `auditlog_mailbox()`.  But I would switch the order of the `oldrecord` and `record` arguments, for consistency with that.

To be honest I'm not sure why I didn't just do it that way in the first place.  Maybe I confused something in my notes and didn't notice that I'd lost that detail.  Sorry!

ksmurchison: Done

elliefm (approved):

rsto (approved):

PR #5642, opened 2025-10-10
imapd.c: only an admin can delete a user's INBOX

ksmurchison:
so don't delete the sub-mailboxes either

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -7571,7 +7571,7 @@ static void cmd_delete(char *tag, char *name, int localonly, int force)
     delete_user = mboxname_isusermailbox(mbname_intname(mbname), 1);
     if (!r && !localonly && delete_user) {
         if (!isadmin) {
-            r = IMAP_PERMISSION_DENIED;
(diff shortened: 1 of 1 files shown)


imap/imapd.c
```diff
@@ -7564,22 +7564,28 @@ static void cmd_delete(char *tag, char *name, int localonly, int force)
...
+        else {
+            const char *userid = mbname_userid(mbname);
+            if (userid) {
+                r = mboxlist_usermboxtree(userid, NULL, delmbox, NULL,
+                                          MBOXTREE_INTERMEDIATES|MBOXTREE_SKIP_ROOT);
+            }
         }
     }
 
     mboxevent = mboxevent_new(EVENT_MAILBOX_DELETE);
```
wolfsage, line 7585: Feels like this should be inside of `if (!r) {...` so we don't create/immediately free it

brong: Yeah maybe but - meh, it's in the error path!  I really don't care.

We should probably restructure the whole function to be "goto done" and have the cleanup there.

PR #5646, opened 2025-10-15
cunit/vg.supp: update for new shape of backend.testc sasl leak

elliefm:
The old suppression stopped working for me, I guess Ubuntu build of cyrus-sasl isn't using dlopen anymore.  I've added a new suppression that works for me, but left the existing one there because Debian/CI probably still needs it.


rsto (approved):
Works for me.

PR #5651, opened 2025-10-21
xapian_wrap.cpp: query numbers in multiple formats

rsto:
We currently do not normalize numbers before indexing, so depending on the locale of the text the same number might
be indexed differently. For decimals, either the '.' or ',' characters might separate the whole number part from the fraction. Similarly, thousands in the whole number might be separated with either of these characters.

This updates the query handler to attempt parsing a query term as a number, optionally ending with a wildcard character. If it determines the input to be a number, then it queries for all typical representations of that number.

### DIFF
--- a/imap/xapian_wrap.cpp
+++ b/imap/xapian_wrap.cpp
@@ -2132,7 +2132,7 @@ static Xapian::Query *xapian_query_new_numbermatch(enum search_part partnum,
 
     // Format number without thousands separator.
     if (frac.size()) {
-        std::string term = barenum;
+        std::string term = prefix + barenum;
         size_t n = term.size();
         term.append(frac);
         term[n] = '.';
@@ -2141,13 +2141,13 @@ static Xapian::Query *xapian_query_new_numbermatch(enum search_part partnum,
         terms.push_back(term);
     }
     else {
-        terms.push_back(barenum);
+        terms.push_back(prefix + barenum);
     }
 
     // Format number with thousands separator.
     if (barenum.size() > 3) {
         for (char sep : { '.', ',' }) {
-            std::string term;
+            std::string term = prefix;
             size_t count = 0;
             for (size_t i = barenum.size(); i > 0; --i) {
                 if (count == 3) {
--- a/cassandane/tiny-tests/JMAPEmail/email_query_number
+++ b/cassandane/tiny-tests/JMAPEmail/email_query_number
@@ -2,10 +2,29 @@
 use Cassandane::Tiny;
 
 sub test_email_query_number
+    :SearchAttachmentExtractor :JMAPExtensions
 {
     my ($self) = @_;
     my $jmap   = $self->{jmap};
     my $imap   = $self->{store}->get_client();
+
+    my $uri = URI->new($self->{instance}->{config}->get('search_attachment_extractor_url'));
+
+    # Start a dummy attachment extractor server.
+    my $handler = sub {
+        my ($conn, $req) = @_;
+        if ($req->method eq 'HEAD') {
+            my $res = HTTP::Response->new(204);
+            $res->content("");
+            $conn->send_response($res);
+        } else {
+            my $res = HTTP::Response->new(200);
+            $res->header("Keep-Alive" => "timeout=1");  # Force client timeout
+            $res->content("You got 76543 things.");
+            $conn->send_response($res);
+        }
+    };
+    $self->{instance}->start_httpd($handler, $uri->port());
 
     my $mime = <<'EOF';
 From: from@example.com
@@ -13,9 +32,27 @@ To: to@example.com
 Subject: test 9876
 Date: Mon, 13 Apr 2020 15:34:03 +0200
 MIME-Version: 1.0
-Content-Type: text/plain
+Content-Type: multipart/mixed; boundary="7435bd413d994b60"
+
+--7435bd413d994b60
+Content-Type: text/plain;charset=utf-8
 
 The total amount is $1,234.56 USD.
+
+--7435bd413d994b60
+Content-Disposition: attachment; filename="attachment.rtf"
+Content-Type: application/rtf; name="attachment.rtf"
+Content-Transfer-Encoding: base64
+
+e1xydGYxXGFuc2lcYW5zaWNwZzEyNTJcY29jb2FydGYyODIwClxjb2NvYXRleHRzY2FsaW5nMFxjb2
+NvYXBsYXRmb3JtMHtcZm9udHRibFxmMFxmc3dpc3NcZmNoYXJzZXQwIEhlbHZldGljYTtcZjFcZnN3
+aXNzXGZjaGFyc2V0MCBIZWx2ZXRpY2EtQm9sZDt9CntcY29sb3J0Ymw7XHJlZDI1NVxncmVlbjI1NV
+xibHVlMjU1O30Ke1wqXGV4cGFuZGVkY29sb3J0Ymw7O30KXG1hcmdsMTQ0MFxtYXJncjE0NDBcdmll
+d3cxMTUyMFx2aWV3aDg0MDBcdmlld2tpbmQwClxwYXJkXHR4NzIwXHR4MTQ0MFx0eDIxNjBcdHgyOD
+gwXHR4MzYwMFx0eDQzMjBcdHg1MDQwXHR4NTc2MFx0eDY0ODBcdHg3MjAwXHR4NzkyMFx0eDg2NDBc
+cGFyZGlybmF0dXJhbFxwYXJ0aWdodGVuZmFjdG9yMAoKXGYwXGZzMjQgXGNmMCBZb3UgZ290IApcZj
+FcYiBcdWwgNzY1NDMKXGYwXGIwIFx1bG5vbmUgIHRoaW5ncy59
+--7435bd413d994b60--
 EOF
     $mime =~ s/\r?\n/\r\n/gs;
     $imap->append('INBOX', $mime) || die $@;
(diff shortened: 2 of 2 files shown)


imap/xapian_wrap.cpp
```diff
@@ -2012,17 +2012,185 @@ static Xapian::Query *xapian_query_new_match_word_break(const xapian_db_t *db,
...
 static Xapian::Query *xapian_query_new_match_internal(const xapian_db_t *db,
                                                       enum search_part partnum,
                                                       const char *str,
                                                       int convert_flags)
 {
     charset_t utf8 = charset_lookupname("utf-8");
-    char *mystr = charset_convert(str, utf8, convert_flags);
+    struct buf buf = BUF_INITIALIZER;
+    buf_init_ro_cstr(&buf, str);
+    buf_trim(&buf);
```
wolfsage, line 2187: There's no tests for querying a number with leading whitespace, or trailing whitespace. If you add some, two bugs appear:
1. buf_truncate() segfaults on a buf with no alloc since it does:
```
    // Imagine buf->len = 7, and we're removing a space so len = 6. buf->alloc = 0
    if ((size_t)len > buf->alloc) {                        // True
        /* grow the buffer and zero-fill the new bytes */
        size_t more = len - buf->len;                  // More = -1, but unsigned, so UINT_MAX!
        buf_ensure(buf, more);                                            
        memset(buf->s + buf->len, 0, more);
    }

```
2. If we fix that by doing what buf_replace_buf does:
```
diff --git a/lib/buf.c b/lib/buf.c
index f8e1c9979f..797cafe3f3 100644
--- a/lib/buf.c
+++ b/lib/buf.c
@@ -228,6 +228,10 @@ EXPORTED void buf_truncate(struct buf *buf, ssize_t len)
         len = buf->len + len;
         if (len < 0) len = 0;
     }
+
+    /* we need buf to be a writable C string now please */
+    buf_cstring(buf);
+
     if ((size_t)len > buf->alloc) {
         /* grow the buffer and zero-fill the new bytes */
         size_t more = len - buf->len;

```

snippet generation fails because `xapian_snipgen_add_match(xapian_snipgen_t *snipgen, const char *match)` needs to trim spaces from match as well

rsto: Bah, one would think we've found all bugs in `buf` already. Thanks! I added a bug fix to this PR.

wolfsage: Heh, cool

imap/xapian_wrap.cpp
```diff
@@ -2012,17 +2012,185 @@ static Xapian::Query *xapian_query_new_match_word_break(const xapian_db_t *db,
...
+
+    // Format number without thousands separator.
+    std::string term = barenum;
+    if (frac.size()) {
+        size_t n = term.size();
+        term.append(frac);
+        term[n] = '.';
+        terms.push_back(term);
+        term[n] = ',';
+        terms.push_back(term);
```
dilyanpalauzov, line 2141: As this line is repeated in the `then` and `else` clauses, it can be put after the `if () {}`.

Or the copy constructor from `std::string term = barenum;` can be moved only in the `then` clause, avoiding the copy in the `else` case.

imap/xapian_wrap.cpp
```diff
@@ -2012,17 +2012,185 @@ static Xapian::Query *xapian_query_new_match_word_break(const xapian_db_t *db,
...
+                term.append(frac);
+                term[n] = sep == '.' ? ',' : '.';
+            }
+            terms.push_back(term);
+        }
+    }
+
+    // Build query from terms.
+    Xapian::Query q;
+    for (size_t i = 0; i < terms.size(); i++) {
```
dilyanpalauzov: There is C++ syntax for this, something like
```c++
for(const std::string &s: terms)
       q |= is_wildcard ? Xapian::Query(Xapian::Query::OP_WILDCARD, s) : Xapian::Query(s);
```

imap/xapian_wrap.cpp
```diff
@@ -2368,6 +2544,7 @@ struct xapian_snipgen
     Xapian::Database *memdb;
     std::vector<std::string> *loose_terms;
     std::vector<std::string> *queries;
+    std::vector<Xapian::Query> *numbers;
```
dilyanpalauzov, line 2547: Why not removing the `*` and assuming that `numbers` is always present - initially empty vector.  It does not need them initialization, or any elimination in the destructor.

imap/xapian_wrap.cpp
```diff
@@ -2451,17 +2629,35 @@ static Xapian::Query xapian_snipgen_build_query(xapian_snipgen_t *snipgen, Xapia
...
+    buf_init_ro_cstr(&buf, match);
+    buf_trim(&buf);
+    match = buf_cstring(&buf);
+    size_t len = buf_len(&buf);
     bool is_query = len > 1 && ((match[0] == '"' && match[len-1] == '"') ||
                                 (strchr(match, '*') != NULL));
 
-    if (is_query) {
+    Xapian::Query *q = xapian_query_new_numbermatch(SEARCH_PART_BODY, match);
+    if (q) {
```
dilyanpalauzov: These two lines could be written as one:
```c++
if (Xapian::Query *q = xapian_query_new_numbermatch(SEARCH_PART_BODY, match)) {
```

imap/xapian_wrap.cpp
rsto: Right, thanks. I moved the constructor into the `then` clause.

imap/xapian_wrap.cpp
rsto: Thanks, done.

imap/xapian_wrap.cpp
rsto: Thanks, done.

imap/xapian_wrap.cpp
rsto: I agree, but I'd rather want to keep this consistent for now. For real, the pointers should get removed from all the fields and the snippet generation rewritten.

dilyanpalauzov: This is https://github.com/cyrusimap/cyrus-imapd/pull/5128.

imap/xapian_wrap.cpp
```diff
@@ -2012,17 +2012,185 @@ static Xapian::Query *xapian_query_new_match_word_break(const xapian_db_t *db,
...
+ * This function attempts to parse the term 'str' as a number,
+ * optionally ending with a wildcard character. If it determines
+ * the input to be a number, then it returns a query that matches
+ * all typical representations of that number. Otherwise it
+ * returns NULL.
+ */
+static Xapian::Query *xapian_query_new_numbermatch(enum search_part partnum,
+                                                   const char *str)
+{
+    std::string prefix(get_term_prefix(partnum));
```
dilyanpalauzov, line 2034: If this is moved after the early return 4 lines lower, then sometimes constructing and destructing a string object can be avoided.

imap/xapian_wrap.cpp
```diff
@@ -2012,17 +2012,185 @@ static Xapian::Query *xapian_query_new_match_word_break(const xapian_db_t *db,
...
+            }
+        }
+    }
+    else {
+        // No thousands separator detected, the whole number must
+        // not be empty and must only contain digits.
+        if (num.empty()) {
+            return NULL;
+        }
+        for (size_t i = 0; i < num.size(); ++i) {
```
dilyanpalauzov: I guess C++ syntax can be used here:
```c++
for (const auto& it: num)
  if (!isdigit(it)) return NULL;
```

imap/xapian_wrap.cpp
```diff
@@ -2012,17 +2012,185 @@ static Xapian::Query *xapian_query_new_match_word_break(const xapian_db_t *db,
...
+
+    // Format number without thousands separator.
+    if (frac.size()) {
+        std::string term = barenum;
+        size_t n = term.size();
+        term.append(frac);
+        term[n] = '.';
+        terms.push_back(term);
+        term[n] = ',';
+        terms.push_back(term);
```
dilyanpalauzov, line 2141: To what I remember from C++ the last line:
```c++
 terms.push_back(term);
```
means that the char* array of `term` is copied into `terms` and then`term` and its char* array are destroyed, while
```c++
terms.push_back(std::move(term))
```
will move the char* array of `term` into `terms` - that array is not copied and the origin is immediately afterwards destroyed as it goes out of scope.

imap/xapian_wrap.cpp
```diff
@@ -2451,17 +2629,34 @@ static Xapian::Query xapian_snipgen_build_query(xapian_snipgen_t *snipgen, Xapia
         }
     }
 
+    if (snipgen->numbers) {
+        for(size_t i = 0; i < snipgen->numbers->size(); ++i) {
```
dilyanpalauzov, line 2633: Here C++ syntax can be used, something like:
```c++
for (const auto &it : *(snipgen->numbers)) q |= it;
```
and instead of `auto` it is possible to write `Xapian::Query`.  Likewise 5 lines above.

imap/xapian_wrap.cpp
```diff
@@ -2012,17 +2012,185 @@ static Xapian::Query *xapian_query_new_match_word_break(const xapian_db_t *db,
...
+            for (size_t j = i + 1; j < i + 4; ++j) {
+                if (!isdigit(num[j])) return NULL;
+            }
+        }
+    }
+    else {
+        // No thousands separator detected, the whole number must
+        // not be empty and must only contain digits.
+        if (num.empty()) {
+            return NULL;
```
dilyanpalauzov, line 2119: On the other places around single `return NULL;` there are no brackets.

cunit/buf.testc
```diff
@@ -9,6 +9,9 @@
 
 #include <sys/mman.h>
 
+#define CU_ASSERT_MEM_EQUAL(actual, expected, sz) \
+    CU_ASSERT_EQUAL(memcmp(actual, expected, sz), 0)
+
```
elliefm, line 14: I missed this earlier, but actually there's already a `CU_ASSERT_MEMEQUAL`, whoops

PR #5652, opened 2025-10-21
Update xxhash to upstream commit 3ba60326b8...

brong:
There's been a bunch of upstream changes, and this might help us with compiling on GCC 14.2.1

### DIFF
--- a/.fixme_ignore
+++ b/.fixme_ignore
@@ -1,0 +1,1 @@
+lib/xxhash.h


brong:
> I would wrap this in a `#IFDEF DEBUG` or whatever gcc defines when you run `gcc -Og`, since we probably want to get the performance benefit from the inlining in non-debug builds (assuming we run non-debug builds at least some of the time?)
> 
> Also, what's the difference between inline hints (XXH_NO_INLINE_HINTS) and "inline all" (XXH_INLINE_ALL)?

It looks like gcc doesn't define anything specific when you set -Og, however here's what the docs say at https://github.com/Cyan4973/xxHash

XXH_NO_INLINE_HINTS: By default, xxHash uses __attribute__((always_inline)) and __forceinline to improve performance at the cost of code size. Defining this macro to 1 will mark all internal functions as static, allowing the compiler to decide whether to inline a function or not. This is very useful when optimizing for smallest binary size, and is automatically defined when compiling with -O0, -Os, -Oz, or -fno-inline on GCC and Clang. It may also be required to successfully compile using -Og, depending on compiler version.

I think letting the compiler decide seems fine.  We'll see if it causes performances changes that justify putting a guard around it.

PR #5657, opened 2025-10-24
jmap_contact.c: handle VALUE=X for media properties

ksmurchison:
This fixes a crasher where it was assumed that the value would be a URI and we would fetch the data from the wrong field in the vcardvalue, resulting in strdup()'ing a NULL string

### DIFF
--- a/imap/jmap_contact.c
+++ b/imap/jmap_contact.c
@@ -6117,9 +6117,12 @@ static char *_value_to_uri_blobid(vcardproperty *prop,
                                   vcardproperty_version version,
                                   char **type, char **blobid)
 {
+    const char *val;
+
+    val = vcardvalue_as_vcard_string(vcardproperty_get_value(prop));
+
     if (!blobid && version == VCARD_VERSION_40) {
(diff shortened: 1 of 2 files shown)


imap/jmap_contact.c
```diff
@@ -6117,8 +6117,16 @@ static char *_value_to_uri_blobid(vcardproperty *prop,
...
                                   char **type, char **blobid)
 {
+    const vcardvalue *val;
+    const char *uri;
+
+    val = vcardproperty_get_value(prop);
+    if (vcardvalue_isa(val) == VCARD_X_VALUE)
+        uri = vcardvalue_get_x(val);
+    else
+        uri = vcardvalue_get_uri(val);
```
wolfsage: This solves VALUE=X, but you could have VALUE=date:20151008 and then we'll crash because uri is null and callers expect it not to be.

Makes me thing this should never return null and xstrdupnull below is a mistake?

ksmurchison: Changed this to use vcardvalue_as_vcard_string().  The only issue is that unsafe chars will be escaped, but GIGO.

PR #5659, opened 2025-10-24
xstats.c: remove XSTATS, the unused metrics system

rjbs:
XSTATS is not documented, and it's not in use by Fastmail.  We now offer a Prometheus endpoint, which should be the focus of future observability work.

### DIFF
--- a/changes/next/cyr-1355-remove-xstats
+++ b/changes/next/cyr-1355-remove-xstats
@@ -1,0 +1,24 @@
+Description:
+
+The undocumented XSTATS metrics system has been removed.
+
+
+Documentation:
+
+There is none.
+
+
+Config changes:
+
+None.
+
+
+Upgrade instructions:
+
+If you somehow had found XSTATS and were using it, stop!  Look at the
+Prometheus metrics instead.
+
+
+GitHub issue:
+
+https://github.com/cyrusimap/cyrus-imapd/pull/5659


changes/next/cyr-1355-remove-xstats
```diff
@@ -0,0 +1,24 @@
...
+
+Upgrade instructions:
+
+If you somehow had found XSTATS and were using it, stop!  Look at the
+Prometheus metrics instead.
+
+
+GitHub issue:
+
+https://github.com/cyrusimap/cyrus-imapd/pull/5659
```
elliefm, line 24: FYI This field doesn't need to be self-referential, it's just if there was an actual github issue that this PR addresses (so we can write "Fixed #xxxx: blah" instead of just "Fixed: blah" or whatever).  Or I guess if we're retroactively adding a changes file for a previous change that was lacking one.  But usually the PR number will be in the merge commit that added the changes file, if we need it, so it doesn't need to be here too.

elliefm (approved):

changes/next/cyr-1355-remove-xstats
rjbs: I'm going to leave this here, but will be open to hear further objection.  My thinking is:

At some point in time, we expect a person to collect up the changes/next files and combine them into a larger changelog.  Saving them the trouble of looking at the blame for the file to find its commit, and then find that commit's pull commit, to then find the PR… seems kind.  (Maybe we don't plan to include these links in the final changelog either, but that'll need establishing.)

PR #5660, opened 2025-10-29
example_libcyrus.c: exercise parseaddr functions

elliefm:
This PR updates `doc/examples/libcyrus/example_libcyrus.c` to exercise the parseaddr functions, which are cyrus-imapd's implementation of an RFC822 address parser.

The existing cassandane LibCyrus.example_libcyrus test compiles, links, and runs this example to verify that it's good.  The fact that it still passes proves that the address parser is usable from outside cyrus-imapd by an application that links against libcyrus.so.

### DIFF
--- a/doc/examples/libcyrus/example_libcyrus.c
+++ b/doc/examples/libcyrus/example_libcyrus.c
@@ -263,6 +263,32 @@ void test_mappedfile(void)
     puts("mappedfile ok");
 }
 
+void test_parseaddr(void)
+{
+    const char *str = "Fred Bloggs <fbloggs@example.net>,"
+                      " Sarah Jane Smith <sjsmith@example.net>";
+
+    struct address *list = NULL;
+    parseaddr_list(str, &list);
+    parseaddr_free(list);
+
+    const struct address *addr;
+    struct address_itr iter;
+    address_itr_init(&iter, str, 0);
+    for (addr = address_itr_next(&iter);
+         addr != NULL;
+         addr = address_itr_next(&iter))
+    {
+        char *x = address_get_all(addr, 1);
+
+        (void) x;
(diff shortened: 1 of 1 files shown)


doc/examples/libcyrus/example_libcyrus.c
```diff
@@ -263,6 +263,32 @@ void test_mappedfile(void)
...
+    const struct address *addr;
+    struct address_itr iter;
+    address_itr_init(&iter, str, 0);
+    for (addr = address_itr_next(&iter);
+         addr != NULL;
+         addr = address_itr_next(&iter))
+    {
+        char *x = address_get_all(addr, 1);
+
+        (void) x;
```
rsto, line 284: Is this some "make-toolchain-happy" trickery?

elliefm: Yeah -- this `(void) x;` statement "uses" `x` without actually doing anything with it (in non-example code, you would do something useful here).  Without it, the compiler _might_ detect that x is set-but-not-used (and warn about it), and the optimiser _might_ then remove the call to `address_get_all()` by dead-code-elimination. 

The point is to actually call into the shared library function at runtime, to prove that it's possible to do so.  If the call was eliminated during compilation, runtime success proves nothing.

elliefm: The reason we need this here, but not earlier in the `parseaddr_list(); parseaddr_free();` step, is because the compiler might recognise the `free(x)` call as not doing anything with the value.  `parseaddr_free()` otoh is too complicated for it to make the same inference, so `list` always looks used (for now -- if this changes, it'll start warning about set-but-not-used, and then I'll add a `(void) list;` for that one too).

PR #5662, opened 2025-10-29
Document the replication protocol

brong:


doc/internal/replication_protocol.md
```diff
@@ -0,0 +1,1734 @@
+# Cyrus Replication Protocol
+
+## Introduction
+
+The Cyrus Replication protocol in versions 2.4+ was created to
+replace the earlier replication protocol built by David Carter
+at Cambridge University.  It's based on the same underlying
+principle of "edge trigger to say that something changed"
+followed by "resolve the differences to update the replica".
```
rsto: I find this sentence confusing. I first thought the principle is "edge trigger..." and had no idea what that means, then I came to think that the two quotes together form the principle but then I had to think what that principle actually means.

I think this should get rephrased in simpler terms. It seems to mean that the protocol consists of three stages: first, the replica is informed that state changed, then the difference between the master and replicate state is determined, finally the replica is updated to match the master state.

doc/internal/replication_protocol.md
```diff
@@ -0,0 +1,1734 @@
...
+* **replica** — the server that receives updates from the master.  In normal operation it is read-only (or at least not authoritative).
+* **sync\_client** — the process on the master that drives replication.  It reads the master's state, compares it with the replica, and sends the necessary APPLY commands.
+* **sync\_server** — the process on the replica that accepts replication commands.  It can run as a standalone daemon (via `cyrus-master`) or be embedded in `imapd`.
+* **channel** — a named replication stream.  Multiple channels allow a single master to replicate to several replicas, each with its own sync\_log.
+* **rolling replication** — continuous, event-driven replication.  sync\_client runs as a daemon, tailing the sync\_log and replicating changes within seconds.
+* **one-shot replication** — a single invocation of sync\_client to sync specific users or all users, then exit.  Typically run by an administrator or from a script.
+
+### Design Philosophy
+
+The protocol follows an **edge-triggered** design: the master logs
```
rsto: OK, this sections makes the principle stated in the Introduction much clearer! Unless you update the Introduction with a simpler explanation for the design, you might consider moving this text up to the Introduction.

doc/internal/replication_protocol.md
```diff
@@ -0,0 +1,1734 @@
...
+the *name* of the object that changed (a user, a mailbox, a quota
+root, etc.) but not what the change was.  The sync\_client then
+fetches the current state from both sides, computes the delta, and
+applies it.  This makes the protocol self-healing — if a sync is
+missed or fails partway through, the next attempt will still
+converge to the correct state.
+
+## The DList Format
+
+The Cyrus Replication protocol uses a format called 'DList'.
```
rsto: nit: "uses" is a bit ambiguous. Use "The ... protocol wire format is 'DList'"?

doc/internal/replication_protocol.md
```diff
@@ -0,0 +1,1734 @@
...
+### Types
+
+```
+dlist = dlist-atom / dlist-flag / dlist-num / dlist-hex /
+        dlist-list / dlist-kvlist / dlist-file
+```
+
+#### atom
+
+An atom is actually a sequence of any character other than '\0',
```
rsto: nit: remove "actually"?

doc/internal/replication_protocol.md
```diff
@@ -0,0 +1,1734 @@
...
+the NULL byte.  Character encoding is not specified, but it can
+contain 8 bit characters, and is probably utf8.
```
rsto: is "probably utf8" relevant here if it' not a given that the data actually *is* UTF-8?

doc/internal/replication_protocol.md
```diff
@@ -0,0 +1,1734 @@
...
+On the wire, this is encoded as an astring.
+
+```
+dlist-atom = astring
+               ; from rfc3501
+```
+
+#### flag
+
+Flag is a horrible special case of atom to allow \word to be
```
rsto: nit: Since the prompt was to produce an IETF-like document, the "horrible" is not the right tone. I'd just remove it, unless you intend to keep the documentation style colloquial.

doc/internal/replication_protocol.md
```diff
@@ -0,0 +1,1734 @@
...
+just to make it easier to read.
+
+```
+dlist-flag = flag
+               ; from rfc3501
+```
+
+#### num32/num
+
+Both stored as 64 bit integers internally, and sent as decimal
```
rsto: nit: I'm only going to say this once for the whole document. The style is inconsistent, some paragraphs are fully constructed sentences then paragraphs like this read more like work-in-progress developer notes than fully fleshed documentation.

doc/internal/replication_protocol.md
```diff
@@ -0,0 +1,1734 @@
...
+dlist-hex = 8HEXDIG | 16HEXDIG
+```
+
+#### map
+
+Like atom, but can contain NULL.  All values are parsed off
+the wire as 'map' type and then converted on demand into the
+requested type.
+
+NOTE: this is maybe bogus and should be using literal8 from RFC3516
```
rsto: I don't understand. Does this NOTE apply to the documentation or the implementation?

rsto (changes requested):
I think this document is a good start, and the amount of further work on it depends on what this document is expected to be used for. I had started reading this as if it's supposed to be an IETF document, but as you'll see I soon stopped highlighting every nit.

I think this document first needs some structural work: the document contains both the protocol design as well as its wire format and methods. I suggest to move everything starting from the "The Sync Algorithm" section up before the "The DList Format" section. The latter seems to contain all the nitty gritty about the wire format. Without a clear understanding of the design, this looks like too much detail.

The "The Sync Algorithm" section also could internally be reorganized, e.g. wouldn't the sections about mailbox and GUID conflict resolution get organized in a "Conflict Resolution" section?

Other than structural changes, it might make sense to clarify what this document is meant to be. Is it a developer note for Cyrus internals? Is it meant for anyone knowing nothing about Cyrus internal to be able write a sync client or server? For the latter, I think this needs more work. For the former, it's already quite close!

rsto (approved):

doc/internal/replication_protocol.md
brong: I did a fairly significant rewrite based on this whole review, and this got rewritten as part of that.  It's now clearer about the way that replication works.

PR #5663, opened 2025-10-30
Cassandane: Add test showing NOT:body with findMatchingParts fails

wolfsage:
With an email body containing "Foo Bar Baz" and an Email/query like:

    "filter": {
        "operator": "AND",
        "conditions": [
            { "inMailbox": "<inbox id>" },
            { "text": "\"Foo Bar\"" },
            {
                "operator": "NOT",
                "conditions": [
                    { "body": "Baz" }
                ]
            }
        ]
    }

We find 0 messages as expected.

However, if we use 'https://cyrusimap.org/ns/jmap/mail' and add in:

    "findMatchingParts": true

we incorrectly find a message!

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_query_not_body_find_matching_parts
+++ b/cassandane/tiny-tests/JMAPEmail/email_query_not_body_find_matching_parts
@@ -1,0 +1,43 @@
+#!perl
+use Cassandane::Tiny;
+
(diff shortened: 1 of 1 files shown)


wolfsage:
Keeping this around as a "known bug" but fixing will take some time, so closing for now.

PR #5664, opened 2025-10-31
xsyslog et al: also log file/line when logging caller

elliefm:
Context: for non-informational log priorities (i.e. errors and debug), `xsyslog()` automatically logs the calling function name as `func=<x>`, and the stringified value of `errno` (if any) as `syserror=<y>`.  But we realised recently (sorry, don't remember where) that function name alone can be ambiguous.

This PR:

* updates the original `xsyslog()` to log file name and line number too,  as `source=<file:line>`.
* updates `xsyslog_ev()` to automatically log the same diagnostics as `xsyslog()`, as `sys.error`, `caller.file`, `caller.line`, and `caller.func`.  I chose "sys.error" so that "error" remains available to use for some caller-chosen error string.
* removes the `lf_m()` macro, since this is now automatic

`logfmt_push_caller()` is now available for use by higher level logging functions, but no others currently need it.

### DIFF
--- a/cunit/xsyslog_ev.testc
+++ b/cunit/xsyslog_ev.testc
@@ -1,8 +1,9 @@
 #include "cunit/unit.h"
 #include "lib/util.h"
 
-#include "lib/sessionid.h"
 #include "lib/libconfig.h"
+#include "lib/logfmt.h"
+#include "lib/sessionid.h"
 
 #include <syslog.h>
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):

wolfsage:
I do recall that conversation, okay. I can live with it I suppose, cheers

PR #5668, opened 2025-11-03
implement draft-ietf-mailmaint-objectid-bis

brong:
NOTE: this implementation uses the bogus userid which isn't a legit JMAPID

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -4945,14 +4945,13 @@ static void cmd_select(char *tag, char *cmd, char *name)
      * optionally scoped by ACCOUNTID */
     char *intname = NULL;
     if (objid.mailboxid) {
-        mbentry_t *id_mbentry = NULL;
         if (mboxlist_lookup_by_jmapid(imapd_userid, objid.mailboxid,
-                                      &id_mbentry, NULL) == 0
+                                      &mbentry, NULL) == 0
             || mboxlist_lookup_by_uniqueid(objid.mailboxid,
-                                           &id_mbentry, NULL) == 0) {
+                                           &mbentry, NULL) == 0) {
             int accountid_ok = 1;
             if (objid.accountid) {
-                char *found_userid = mboxname_to_userid(id_mbentry->name);
+                char *found_userid = mboxname_to_userid(mbentry->name);
                 if (found_userid) {
                     accountid_ok = !strcmp(found_userid, objid.accountid);
                     free(found_userid);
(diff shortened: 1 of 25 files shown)


ksmurchison (commented):
Looks correct but probably needs some tests, or just leverage the existing ones

imap/imapd.c
```diff
@@ -4882,7 +4949,37 @@ static void cmd_select(char *tag, char *cmd, char *name)
         name = normalize_mboxname(name, NULL);
     }
 
-    char *intname = mboxname_from_external(name, &imapd_namespace, imapd_userid);
+    /* ID-based mailbox selection: try to find mailbox by MAILBOXID,
+     * optionally scoped by ACCOUNTID */
+    char *intname = NULL;
+    if (objid.mailboxid) {
+        mbentry_t *id_mbentry = NULL;
```
ksmurchison: I think we can use the mbentry_t scoped to cmd_select() and not do a second lookup below by name.

imap/imapd.c
```diff
@@ -4882,7 +4949,37 @@ static void cmd_select(char *tag, char *cmd, char *name)
...
+            if (accountid_ok) {
+                intname = xstrdup(id_mbentry->name);
+            }
+            mboxlist_entry_free(&id_mbentry);
+        }
+    }
+    free(objid.mailboxid);
+    free(objid.accountid);
+
+    if (!intname)
```
ksmurchison: Given my comment above, this conditional could be

```
if (!mbentry) {
    r = mlookup()
```

imap/imapd.c
```diff
@@ -4785,6 +4805,48 @@ static int parse_select_params(char *tag, char *cmd, int allowdeleted,
...
+                /* check for optional (MAILBOXID id ACCOUNTID id) */
+                if (c == ' ') {
+                    c = prot_getc(imapd_in);
+                    if (c == '(') {
+                        /* parse key-value pairs */
+                        for (;;) {
+                            c = getword(imapd_in, &arg);
+                            if (!arg.s[0]) break;
+                            ucase(arg.s);
+                            if (c != ' ') break;
```
ksmurchison, line 4787: Should we move this check above ucase()?  No point in upper-casing a string that we won't be using.

imap/imapd.c
```diff
@@ -5814,8 +5918,10 @@ static void cmd_fetch(char *tag, char *sequence, int usinguid)
     if (fetchargs.fetchitems & FETCH_ANNOTATION)
         client_behavior_mask |= CB_ANNOTATE;
 
-    if (fetchargs.fetchitems & (FETCH_EMAILID | FETCH_THREADID))
```
ksmurchison, line 5778: We still want this check to support the individual OBJECTID items, correct?

dilyanpalauzov:
I suggest mentioning in docsr/rfc-support.rst that `draft-ietf-mailmaint-objectid-bis` is implemented, e.g. by adding a Note after RFC 8474.  An example Note can be found after RFC 6154.

ksmurchison:
@brong In an effort to move this towards merging, I rebased on master and added commits to resolve all of my previous comments.

My only outstanding question is whether FETCH (OBJECTID EMAILID) or STATUS (OBJECTID MAILBOXID), should return both items, or just return the compound OBJECTID.  In other words, should OBJECTID supersede any of the separate FETCH/STATUS items?

ksmurchison (approved):

brong:
> @brong In an effort to move this towards merging, I rebased on master and added commits to resolve all of my previous comments.
> 
> My only outstanding question is whether FETCH (OBJECTID EMAILID) or STATUS (OBJECTID MAILBOXID), should return both items, or just return the compound OBJECTID. In other words, should OBJECTID supersede any of the separate FETCH/STATUS items?

It should return both; if you asked for both, you get both!

PR #5672, opened 2025-11-07
convert auditlogs to logfmt

elliefm:
**Reviewers:** I suggest reading this commit by commit.

This PR converts auditlogs to logfmt, by updating the existing auditlog module to call into the new logfmt API.  The auditlog API is mostly unchanged, just the internals are implemented differently.

Key differences:
* field names have changed to match the logfmt style
* new data type-specific `logfmt_push_mailbox()`, `logfmt_push_mbname()` and `logfmt_push_record()` functions allow easy and consistent logging of `struct mailbox`, `mbname_t`, and `struct index_record` objects
* logged lines for iMIP, sieve, and duplicate elimination actions have had their sub-action promoted to the top level (that is: `auditlog: duplicate action=<foo>` is now `event=auditlog.duplicate.foo`, and similarly for iMIP and sieve)

`logfmt_push_mailbox()` and `logfmt_push_mbname()` always log mailbox names in the admin namespace.  It has long been our intention to use this namespace consistently across our logging and command-line tools, so that no translation is needed during admin workflows (whether using command line tools, cyradm, or direct IMAP access); this is a step towards that.  This detail could be split into a separate change if we're not ready to press forward in that direction.

**Fastmail:** our middleware will need updates to accommodate these changes.  Don't merge until that's done.

### DIFF
--- a/changes/next/cyr-2335-logfmt-auditlog
+++ b/changes/next/cyr-2335-logfmt-auditlog
@@ -1,0 +1,23 @@
+Description:
+
+Audit logs have been changed to logfmt
+
+
+Documentation:
+
+https://metacpan.org/pod/Log::Fmt#The-logfmt-text-format
+
+
+Config changes:
+
+None
+
+
+Upgrade instructions:
+
+Deployments that monitor or parse auditlogs will need to update their tooling
+
+
+GitHub issue:
+
+None


elliefm:
@rjbs, @robmueller: Thanks for the comments!

I've made these changes:

* mb -> mbox
* mb.mboxid -> mbox.mailboxid
* msg.uid -> msg.imapuid
* r.client -> r.clienthost (matches login logs)
* sessionid -> r.sessionid

The sessionid -> r.sessionid change is in its own commit, so it'll be easy to postpone that one detail for later, if it comes to that.  I've also separated the "log mailbox names in admin namespace" change into a separate commit from the more general "log mailbox names consistently" change, for similar reasons.

> What's the _remote_ sessionid?

This is the sessionid at the other end when Cyrus proxies a connection to somewhere else.  Not sure if this happens in practice in our environment; it'll show up as `auditlog: proxy [...]` in our current logs if it does.  With these changes, it'll be `event=auditlog.proxy [...]`, so I think "remote.sessionid" is clear in situ, even though it's unclear when it's just in a list of field names.

> FYI we have these fields for client request details we could reuse.

Yeah, I chose all these field names in consideration of the names already in use from the logfmt.mdown spec.  Very few of them matched up exactly though, especially none of the hostname/ip address ones (since in Cyrus, we carry "`hostname [ip]"` around in a single clienthost string, not separate fields).

elliefm:
I need a changes file for this too...

rsto (approved):
LGTM, I like it! One nit: I wish the "logfmt: sessionid -> r.sessionid" commit message would have been rewritten into "logfmt: rename sessionid to r.sessionid" for clarity.

PR #5673, opened 2025-11-07
imapd.c: can't keep conv.db open across commands for NOTIFY SELECTED

ksmurchison:
so open and close it as needed when pushing FETCH responses.

As a side effect, this commit moves convstate out of fetchargs

### DIFF
--- a/cassandane/Cassandane/Cyrus/Notify.pm
+++ b/cassandane/Cassandane/Cyrus/Notify.pm
@@ -142,7 +142,7 @@ sub test_bad
 }
 
 sub test_message
-    :needs_component_idled :min_version_3_9
+    :needs_component_idled :Conversations
 {
     my ($self) = @_;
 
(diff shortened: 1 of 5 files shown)


imap/imapd.c
```diff
@@ -15691,9 +15693,22 @@ static void push_updates(int idling)
                 notify_event_groups->selected.fetchargs.fetchitems) {
                 const char *uidset =
                     json_string_value(json_object_get(msg, "uidset"));
+                struct fetchargs *fa = &notify_event_groups->selected.fetchargs;
 
-                index_fetch(imapd_index, uidset ? uidset : "*", 1,
-                            &notify_event_groups->selected.fetchargs, NULL);
+                /* Open conv.db if we need it */
+                if (fa->fetchitems & (FETCH_MAILBOXIDS|FETCH_MAILBOXES)) {
```
wolfsage: Huh, why doesn't this need to open convs for `FETCH_EMAILID|FETCH_THREADID` too?

ksmurchison: It can get them from the index record.

cassandane/Cassandane/Cyrus/Notify.pm
```diff
@@ -142,7 +142,7 @@ sub test_bad
 }
 
 sub test_message
-    :needs_component_idled :min_version_3_9
+    :needs_component_idled :min_version_3_9 :Conversations
```
rsto: Might also remove the `min_version_3_9`, we decided not to use these anymore AFAIK

ksmurchison: Fixed

brong (approved):

PR #5674, opened 2025-11-07
add backupcyrus daemon

brong:
Making this a separate PR.  This is a format used in-house at Fastmail, along with the perl library that's also being added in this PR.

This replicates the old perl backupcyrusd that Fastmail has been running for many years, speaking a custom protocol.  It's very efficient, though it isn't using modseq and synccrc, it's more based on noticing changes to individual files.  We were going to use the other backup code instead, but that was never finished so we're still running this protocol in production at Fastmail.

This code moves from using a perl daemon with its own locking logic into having the backups from inside Cyrus, where the locking is all using the standard APIs.

Efficiency-wise, we COULD avoid opening mailboxes and just use statuscache instead for fmultistatus; but it's still pretty fast!

### DIFF
--- a/docsrc/imap/reference/manpages/systemcommands/backupcyrusd.rst
+++ b/docsrc/imap/reference/manpages/systemcommands/backupcyrusd.rst
@@ -21,11 +21,11 @@ Description
 ===========
 
 **backupcyrusd** is server which speaks the backup protocol used at Fastmail.
-It implements NO AUTHENTIATION so MUST only listen on safe internal networks
+It implements NO AUTHENTICATION so MUST only listen on safe internal networks
 and be firewalled from any untrusted users.
 
 It accepts commands on its standard input and responds on its standard output.
-t MUST invoked by :cyrusman:`master(8)` with those descriptors attached to a
+It MUST invoked by :cyrusman:`master(8)` with those descriptors attached to a
 remote client connection.
 
 Options


imap/backupcyrusd.c
```diff
@@ -0,0 +1,757 @@
...
+{
+    uint64_t bytes_in = 0;
+    uint64_t bytes_out = 0;
+
+    in_shutdown = 1;
+
+    libcyrus_run_delayed();
+
+    proc_cleanup(&proc_handle);
+
```
wolfsage: Feels weird to duplicate all of this code from bcd_reset, should we refactor that somehow?

imap/backupcyrusd.c
```diff
@@ -0,0 +1,757 @@
...
+        return;
+    }
+    size_t len = strlen(base);
+    struct message_guid guid = MESSAGE_GUID_INITIALIZER;
+    message_guid_generate(&guid, base, strlen(base));
+    // synthetic inode
+    char guidbuf[MESSAGE_GUID_SIZE];
+    message_guid_export(&guid, guidbuf);
+    long unsigned ino = ntohl(*((uint32_t *)(guidbuf)));
+    long unsigned mtime = 0;
```
wolfsage, line 342: Why is mtime always just 0?

imap/backupcyrusd.c
```diff
@@ -0,0 +1,757 @@
...
+ *  "NO $file errormsg" if there is an error reading the file.
+ */
+static void send_file(const char *name, const char *mboxname, const char *fname, int do_data)
+{
+    struct stat sbuf;
+
+    if (!fname || stat(fname, &sbuf)) {
+        if (!do_data) return; // quiet if no file
+        prot_printf(bcd_out, "NO no such file ");
+        puturistring(bcd_out, name);
```
wolfsage, line 376: Here and below we send 'no such file ' using the `name` parameter, not the `fname` parameter, is that intentional?

wolfsage (changes requested):
Looks plausible.

Could we get some automated tests added either through cassandane talking to the daemon or using cunit that makes sure each API returns sensible data?

By exercising the code in cassandane we'll be able to see (through asan/ubsan) if we've done something silly with pointers, etc.., so it's nice to have as much coverage as possible and we should avoid adding new hunks of features without tests to prove them out

imap/backupcyrusd.c
brong: Because it's a synthetically generated file, so the inode is where the change is shown.  This duplicates code in the perl version.

imap/backupcyrusd.c
brong: yes, the name is what was requested, the fname is just the local file.  Not that it matters - it's just for debugging, but the protocol shouldn't (at least in theory) care about the local storage layout.

imap/backupcyrusd.c
brong: Seems reasonable, though I'm not too concerned.  This is copy-paste from another daemon which did the same thing.

brong:
@wolfsage this now has tests, which indeed found memory leaks!  Leaks are fixed too

docsrc/imap/reference/manpages/systemcommands/backupcyrusd.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Description
+===========
+
+**backupcyrusd** is server which speaks the backup protocol used at Fastmail.
+It implements NO AUTHENTIATION so MUST only listen on safe internal networks
+and be firewalled from any untrusted users.
+
+It accepts commands on its standard input and responds on its standard output.
+t MUST invoked by :cyrusman:`master(8)` with those descriptors attached to a
```
wolfsage: nit: "t MUST invoked by" is missing something ...

cassandane/tiny-tests/FastMail/backupcyrusd
```diff
@@ -0,0 +1,45 @@
...
+    my $talk = $self->{store}->get_client();
+    $talk->subscribe("INBOX");
+    $self->make_message("Email 3") or die;
+
+    my ($newversion, $newsize) = Cyrus::Backup::BackupUser($service->host, $service->port, $meta, $data, $servername, 'cassandane');
+
+    $self->assert_num_equals($version, $newversion);
+    $self->assert_num_gt($size, $newsize);
+
+    # XXX - add more tests to check the content of the backups
```
wolfsage, line 41: IMO these are the important tests to add, and to add real soon. They'll save us a lot of time down the road if we have to fix / change anything that affects backups because test failures will let us know an incompatibility is introduced

docsrc/imap/reference/manpages/systemcommands/backupcyrusd.rst
```diff
@@ -0,0 +1,64 @@
...
+
+.. parsed-literal::
+
+    **backupcyrusd** [ **-C** *config-file* ] [ **-U** *uses* ] [ **-T** *timeout* ] [ **-D** ]
+
+Description
+===========
+
+**backupcyrusd** is server which speaks the backup protocol used at Fastmail.
+It implements NO AUTHENTIATION so MUST only listen on safe internal networks
```
wolfsage: AUTHENTIATION -> AUTHENTICATION

cassandane/tiny-tests/FastMail/backupcyrusd
wolfsage: This is tracked in CYR-2423

brong: OK I'm going to merge this as-is and add both tests and a some restore tooling later.

PR #5675, opened 2025-11-10
tls.c: support TLS ALPN unconditionally

ksmurchison:
OpenSSL 3.0+ includes support for ALPN
and we no longer support anything earlier than 3.0

### DIFF
--- a/imap/cyr_buildinfo.c
+++ b/imap/cyr_buildinfo.c
@@ -173,15 +173,11 @@ static json_t *buildinfo()
 
     /* Build dependencies */
     json_object_set_new(dependency, "openssl", json_true());
+    json_object_set_new(dependency, "openssl_alpn", json_true());
 #ifdef HAVE_LDAP
     json_object_set_new(dependency, "ldap", json_true());
 #else
     json_object_set_new(dependency, "ldap", json_false());
-#endif
-#ifdef HAVE_TLS_ALPN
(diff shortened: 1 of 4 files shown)


imap/cyr_buildinfo.c
```diff
@@ -178,11 +178,6 @@ static json_t *buildinfo()
 #else
     json_object_set_new(dependency, "ldap", json_false());
 #endif
-#ifdef HAVE_TLS_ALPN
-    json_object_set_new(dependency, "openssl_alpn", json_true());
-#else
```
wolfsage, line 183: .oO(Should we just keep reporting we support openssl_alpn in case anyone is keying off of this?)

ksmurchison: Sure, maybe?  I'll leave that up to @elliefm to decide

elliefm: Great question.  Replication and murder can both use TLS, so to test those correctly between versions, the tests will need to know what behaviours to expect from each.  So I think we should do like we already did for "openssl" a few lines up, and hardcode it to `json_true()`, rather than remove it completely.

We don't currently do a lot of cross version testing, but it would be good to do more of it if it were easier.   Keeping this key will avoid making it harder.

ksmurchison: Done

PR #5679, opened 2025-11-12
assert on bad cached subject

elliefm:
We saw squatter crashes stemming from failures to parse cached message subjects, we believe due to corrupt cache.  This change asserts that the parse was successful, rather than potentially carrying on in a bad state only to crash elsewhere later.


brong (approved):
Thanks!  Won't stop the thing crashing but hey :)

PR #5682, opened 2025-11-12
Cyr 2308 notes updates

ksmurchison:
- Fix ensure_notes_collection() to re-fetch the mbentry after creation so that we have the proper ACL when trying to open the mailbox
-  Actually get the encoding and charset from the parsed bodystructure
- Convert CRLF and CR to bare LF in Note/get
- Allow /get{update} without specifying isHTML

### DIFF
--- a/imap/jmap_notes.c
+++ b/imap/jmap_notes.c
@@ -580,7 +580,8 @@ static int _notes_setargs_check(const char *id, json_t *args, json_t **err)
     return r;
 }
 
-static int _note_create(struct mailbox *mailbox, json_t *note, json_t **new_note)
+static int _note_create(struct mailbox *mailbox, json_t *note,
+                        modseq_t cmodseq, json_t **new_note)
 {
     struct stagemsg *stage = NULL;
     struct appendstate as;
@@ -705,7 +706,7 @@ static int _note_create(struct mailbox *mailbox, json_t *note, json_t **new_note
 
     /* Append the message to the mailbox */
     if (isFlagged) strarray_append(&flags, "\\Flagged");
-    r = append_fromstage(&as, &bodypart, stage, &now, 0,
+    r = append_fromstage(&as, &bodypart, stage, &now, cmodseq,
                          &flags, 0, /*annots*/NULL);
 
     if (r) {
@@ -776,7 +777,8 @@ static void _notes_update_cb(const char *id, message_t *msg,
           json_t *new_note = jmap_patchobject_apply(note, patch, NULL, 0);
 
             if (new_note) {
-                r = _note_create(msg_mailbox(msg), new_note, &updated_note);
+                r = _note_create(msg_mailbox(msg), new_note,
+                                 msg_createdmodseq(msg), &updated_note);
                 json_decref(new_note);
             }
             else {
@@ -913,7 +915,7 @@ static int jmap_note_set(jmap_req_t *req)
             err = json_pack("{s:s}", "type", "forbidden");
         }
         else if (!_notes_setargs_check(NULL/*id*/, val, &err)) {
-            r = _note_create(mbox, val, &new_note);
+            r = _note_create(mbox, val, 0/*cmodseq*/, &new_note);
             if (r) err = jmap_server_error(r);
         }
         if (err) {
--- a/cassandane/tiny-tests/JMAPNote/note_changes
+++ b/cassandane/tiny-tests/JMAPNote/note_changes
@@ -1,0 +1,137 @@
+#!perl
+use Cassandane::Tiny;
+use Test::Deep ':v1';
+
+sub test_note_changes
+    :JMAPExtensions
+{
+    my ($self) = @_;
+
+    my $jmap = $self->{jmap};
+
+    my $res = $jmap->CallMethods([['Note/get', { }, "R1"]]);
+    my $state = $res->[0][1]{state};
+
+    $res = $jmap->CallMethods([
+        ['Note/changes', {
+            sinceState => $state
+         },
+         "R1"]
+    ]);
+    $self->assert_cmp_deeply(
+        superhashof({
+            oldState       => $state,
+            newState       => $state,
+            hasMoreChanges => JSON::false,
+            created        => [],
+            updated        => [],
+            destroyed      => []
+        }),
+        $res->[0][1],
+    );
+
+    xlog "create note";
+    $res = $jmap->CallMethods([
+        ['Note/set', {
+            create => {
+                "1" => {
+                    title  => 'Hello World',
+                },
+            }
+         },
+         "R1"],
+        ['Note/changes', {
+            sinceState => $state
+         },
+         "R2"]
+    ]);
+    my $oldstate = $state;
+    $state = $res->[0][1]{newState};
+    $self->assert_str_not_equals($oldstate, $state);
+    my $id = $res->[0][1]{created}{"1"}{id};
+
+    $self->assert_cmp_deeply(
+        superhashof({
+            oldState       => $oldstate,
+            newState       => $state,
+            hasMoreChanges => JSON::false,
+            created        => [ $id ],
+            updated        => [],
+            destroyed      => []
+        }),
+        $res->[1][1],
+    );
+
+    xlog "update note";
+    $res = $jmap->CallMethods([
+        ['Note/set', {
+            update => {
+                $id => {
+                    title => 'Hello World 2',
+                },
+            }
+         },
+         "R1"],
+        ['Note/changes', {
+            sinceState => $state
+         },
+         "R2"]
+    ]);
+    $oldstate = $state;
+    $state = $res->[0][1]{newState};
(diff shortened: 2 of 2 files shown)


imap/jmap_notes.c
```diff
@@ -230,6 +231,12 @@ static int ensure_notes_collection(const char *accountid, mbentry_t **mbentryp)
                        annot, error_message(r));
                 goto done;
             }
+
+            /* Re-fetch the mbentry so it is fuly populated */
```
rsto: nit: typo

imap/jmap_notes.c
```diff
@@ -318,6 +318,9 @@ static int _note_get(message_t *msg, json_t *note, hash_table *props,
         }
         charset_free(&charset);
 
+        buf_replace_all(buf, "\r\n", "\n");  // replace CRLF with LF
+        buf_replace_all(buf, "\r", "\n");    // replace CR with LF
```
rsto, line 329: This is different from what we do for email body values in JMAP Email/get, see https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/jmap_mail.c#L8022 There, we remove any CR characters, rather than replacing them with LF. I would prefer to do the same thing for both object types, if only for consistency. But I don't know what led to the decision to replace CR with LF here? Should we adapt the JMAP Email code to do the same?

imap/message.c
```diff
@@ -5731,6 +5731,10 @@ EXPORTED int message_get_encoding(message_t *m, int *encp)
 {
     int r = message_need(m, M_CACHEBODY);
     if (r) return r;
+
+    if (!m->body->charset_enc && m->body->encoding)
```
rsto: This and the get_charset_id function changes suggest to me that we might want to get rid of `message_parse_cbodystructure` and replace it with calls to `message_read_bodystructure`. It's confusing to have two functions that load `struct body` from the cache but one of them only partially sets the message attributes.

I wonder if that is for optimization or only legacy from when the `message` API got implemented. Maybe @brong nows its backstory?

Depending on the urgency of this pull request, I could either approve this with a `TODO` added, or we give it a couple more time to try removing `message_parse_cbodystructure` if there is no good reason to keep it. I can help with that, if preferred.

cassandane/tiny-tests/JMAPNote/note_get
```diff
@@ -0,0 +1,37 @@
...
+    my $imap = $self->{store}->get_client();
+    $imap->create('INBOX.Notes') || die $@;
+    $imap->append('INBOX.Notes', $note) || die $@;
+
+    my $res = $jmap->CallMethods([['Note/get', { }, "R1"]]);
+    $self->assert_str_equals($id, $res->[0][1]{list}[0]{id});
+    $self->assert_str_equals('snowman', $res->[0][1]{list}[0]{title});
+    $self->assert_str_equals("☃\n\n", $res->[0][1]{list}[0]{body});
+    $self->assert_str_equals(JSON::false, $res->[0][1]{list}[0]{isHTML});
+}
```
wolfsage, line 92: I find this way of testing multiple values in a hash torturous to read (and write!).

Rik recently added assert_cmp_deeply which we should use more often I think.

Check this diff out and see if you think it's better?
```
-    $self->assert_str_equals($id, $res->[0][1]{list}[0]{id});
-    $self->assert_str_equals('snowman', $res->[0][1]{list}[0]{title});
-    $self->assert_str_equals("<E2><98><83>\n\n", $res->[0][1]{list}[0]{body});
-    $self->assert_str_equals(JSON::false, $res->[0][1]{list}[0]{isHTML});
+    $self->assert_cmp_deeply(
+      superhashof({
+        id     => $id,
+        title  => 'snowman',
+        isHTML => JSON::false,
+        body   => "<E2><98><83>\n\n"
+      }),
+      $res->[0][1]{list}[0],
+    );
```

(For superhashof() you need `use Test::Deep ':v1'` at the top of the test but we could fix that in tiny tests at least so those methods are always exported maybe....)

cassandane/tiny-tests/JMAPNote/note_set
```diff
@@ -0,0 +1,99 @@
...
+    xlog "create note with body but without isHTML";
+    my $res = $jmap->CallMethods([['Note/set',
+                                { create => {
+                                    "1" => {
+                                        title => "HTML",
+                                        body => "<html><head></head><body>Hello World!</body></html>",
+                                    },
+                                  }
+                                },
+                                "R1"]]);
```
wolfsage: Nit:

These probably follow our style guide in some way but they make it really hard to read the code.

IMO this is way more readable:
```
my $res = $jmap->CallMethods([[
    'Note/set' => {
        create => {
            "1" => {
                title => "HTML",
                body => "<html><head></head><body>Hello World!</body></html>",
            },
        },
    }, "R1",
]]);
```

imap/jmap_notes.c
wolfsage: nit: fuly -> fully

imap/jmap_notes.c
ksmurchison: @rjbs We discussed replacing bar CR, do we want to revisit that?

imap/jmap_notes.c
ksmurchison: Fixed

cassandane/tiny-tests/JMAPNote/note_get
ksmurchison: Done

cassandane/tiny-tests/JMAPNote/note_set
ksmurchison: Done

imap/message.c
ksmurchison: I reverted  the message.c changes and parsed the bodystructure myself in jmap_notes.c

Is this better?

rsto: Much better indeed. I'll give it a shot removing `message_parse_cbodystructure` in a different PR.

imap/jmap_notes.c
```diff
@@ -299,24 +299,29 @@ static int _note_get(message_t *msg, json_t *note, hash_table *props,
...
-        charset = charset_lookupname(charset_id);
-        if (encoding || strcasecmp(charset_canon_name(charset), "utf-8")) {
-            char *dec = charset_to_utf8cstr(buf_cstring(buf), buf_len(buf),
-                                            charset, encoding);
-            buf_setcstr(buf, dec);
-            free(dec);
+        if (!r) {
+            struct body *body;
+
+            message_read_bodystructure(msg_record(msg), &body);
```
wolfsage, line 307: I don't understand this commit. How is the new behaviour different from what we were doing before?

That is, why are message_get_encoding/message_get_charset_id not getting the charset from the "parsed body" 

Looking through those methods, it seems like they are doing something very similar / something "right"

ksmurchison: Those functions are broken in that the fields that we are fetching aren't actually populated.  There are derived from other values in the body, but the underlying message_need() isn't doing the right thing.

My  first iteration fixed those functions, but rsto didn't like it because he plans to remove the broken code altogether.

rsto: @wolfsage before, the code was calling `message_get_encoding` and `message_get_charset_id` which internally were calling `message_parse_cbodystructure`, a function that only partially reads the cached bodystructure. All other places in Cyrus call `message_read_bodystructure` which does things right.
I am working on a separate PR to remove the legacy `message_parse_cbodystructure` function.

rsto: FYI - this is PR https://github.com/cyrusimap/cyrus-imapd/pull/5683 , currently a draft

wolfsage: Thank you both, I see now

rsto (approved):
LGTM, but I suggest to align CR handling with JMAP Email, one way or the other.

imap/jmap_notes.c
```diff
@@ -292,24 +299,34 @@ static int _note_get(message_t *msg, json_t *note, hash_table *props,
...
+            if (body) {
+                int encoding = encoding_lookupname(body->encoding);
+                charset_t charset = charset_lookupname(body->charset_id);
+
+                if (encoding ||
+                    strcasecmp(charset_canon_name(charset), "utf-8")) {
+                    char *dec = charset_to_utf8cstr(buf_base(buf), buf_len(buf),
+                                                    charset, encoding);
+                    if (dec) {
```
wolfsage, line 316: We should add a test case for this branch

ksmurchison: JMAPNote.note_get tests the decode path

wolfsage: Does it test the crasher fix for a bad encoding?

ksmurchison: It does now.  Both bad charset and bad encoding, both of which get ignored

wolfsage: lgtm, thanks

PR #5685, opened 2025-11-14
sync_support: don't scan the whole mailbox every time we open it!

brong:
This is because sync_mailbox_version_check used a lot of CPU time and IO for a situation which never happens.

Instead, the sanity check is moved to the couple of places where it's really needed - before sending a file over the wire!


ksmurchison (approved):
Looks sane to me

elliefm (approved):

PR #5686, opened 2025-11-14
cassandane: Add `cyr-easycd` command to get at test data easily

wolfsage:
With this, you can run a test, which spits out:

    [169080] =====> Instance::stop[1875] stop main instance for test
    Cassandane::Cyrus::JMAPMailbox.mailbox_set: basedir /tmp/cass/2000160101

And run:

    ./cassandane/utils /tmp/cass/2000160101

Or:

    ./cassandane/utils /tmp/cass/2000160101 someuser

And you'll get dropped into a tmpfs directory that includes most or all of the data for that user and their mailboxes, etc... in a symlink tree which you can look through with standard filesystem tools (cd, ripgrep, etc...)

This will use, effectively, $(pwd)/imap/mbpath and other tools in the current cyrus-imapd build dir to determine mailbox names and other paths.

### DIFF
--- a/cassandane/utils/cyr-easycd
+++ b/cassandane/utils/cyr-easycd
@@ -1,0 +1,167 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+
+use lib 'perl/imap';
+
+use feature 'signatures';
+
+use Carp;
+use Cwd;
+use Path::Tiny;
+use Capture::Tiny ':all';
+use Process::Status;
+use Cyrus::Mbname;
+
+# Create a tmp directory then symlink all of a user's
+# mail/contacts/calendars etc into it, dropping you in a dir like:
+#
+# mbox/
+# mbox/INBOX/
+# mbox/INBOX/meta/
+# mbox/INBOX/spool/
+
+my $testdir = shift || die "Usage: $0 <testdir> [username - defaults to 'cassandane']\n";
+my $username = shift || 'cassandane';
+
+@ARGV && die "Unexpected extra arguments: @ARGV\n";
+
+# Allow quick copy/paste of output from test run and do the right thing
+$testdir =~ s{conf/.*\.conf$}{};
+
+# Ensure trailing / for linking
+$testdir =~ s{/$}{/};
+
+my $cfg = "$testdir/conf/imapd.conf";
+-e $cfg || die "$testdir doesn't contain conf/imapd.conf, not a cass test dir?\n";
+
+my ($cur, $tmpdir);
+
+END {
+    if ($cur && $tmpdir) {
+        chdir($cur);
+        $tmpdir->remove_tree;
+    }
+}
+
+# Try to cleanup on sensible exits
+$SIG{HUP} = $SIG{INT} = $SIG{KILL} = $SIG{TERM} = sub { exit; };
+
+$cur = getcwd;
+$tmpdir = create_tmpdir();
+
+chdir($tmpdir);
+mkdir("$tmpdir/mbox");
+
+# What kind of mailbox name format do we need?
+my $cnf = run_command("$cur/imap/cyr_info", "-C", $cfg, "conf-all");
+my $unixhiersep = 1;
+
+if ($cnf =~ /^unixhierarchysep: (.*?)$/m) {
+  $unixhiersep = $1 eq 'yes' ? 1 : 0;
+}
+
+my $user_inbox = $unixhiersep ? "user/$username" : "user.$username";
+my $mblist = run_command("$cur/imap/cyr_ls", "-C", $cfg, "-R", $user_inbox);
+my @mailboxes = map { /^([^\s]+):/ ? $1 : () } split("\n", $mblist);
+
+my %done;
+
+while (@mailboxes) {
+    my $mbname = shift @mailboxes;
+
+    if ($done{$mbname}++) {
+        warn "We've already processed $mbname?! Skipping...\n";
+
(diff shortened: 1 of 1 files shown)


cassandane/utils/cyr-easycd
```diff
@@ -0,0 +1,179 @@
...
+# Ensure trailing / for linking
+$testdir =~ s{/$}{/};
+
+my $cfg = "$testdir/conf/imapd.conf";
+-e $cfg || die "$testdir doesn't contain conf/imapd.conf, not a cass test dir?\n";
+
+my ($cur, $tmpfs);
+
+END {
+  if ($cur && $tmpfs) {
```
elliefm: nit: cyrus style is four-space indents, not two

cassandane/utils/cyr-easycd
```diff
@@ -0,0 +1,179 @@
...
+  if ($done{$mbname}++) {
+    warn "We've already processed $mbname?! Skipping...\n";
+
+    next;
+  }
+
+  if ($mbname =~ m{/}) {
+    # Should not happen right?
+    die "Unexpected '/' in foldername: $mbname\n";
+  }
```
elliefm: Whether this can happen will depend on what namespace `cyr_ls` uses for listing mailboxes, which I don't know off the top of my head.  Our command line tools ought to always use the admin namespace for shell I/O, but we're not very consistent about that yet, so it might actually be the internal namespace—in which case it'll work fine for now, but will start breaking when we fix `cyr_ls` to use the admin namespace, which uses `'/'` if unixhierarchysep is set.

wolfsage: Oh good call thanks. I've updated the code to work whether or not we're using unix hierarchy separators. Please re-review?

cassandane/utils/cyr-easycd
```diff
@@ -35,145 +36,134 @@ $testdir =~ s{/$}{/};
...
-my $user_inbox = Cyrus::Mbname->new_userfolder($username)->adminfolder();
-my $inbox_spool_path = mailbox_to_path($user_inbox);
-my ($mb_prefix, $mb_suffix) = $user_inbox =~ /^(.*?)(?:(@.*))?$/;
-$mb_suffix //= "";
+# What kind of mailbox name format do we need?
+my $cnf = run_command("$cur/imap/cyr_info", "-C", $cfg, "conf-all");
+my $unixhiersep = 1;
 
+if ($cnf =~ /^unixhierarchysep: (.*?)$/m) {
+  $unixhiersep = $1 eq 'yes' ? 1 : 0;
```
rjbs, line 62: The config value for a true bool can be `t` or `1` or `yes` or `true`.  We normalize this to "on" or "off" in libconfig.  But `cyr_info` prints it as "yes".  *FINE*.

cassandane/utils/cyr-easycd
```diff
@@ -35,145 +36,134 @@ $testdir =~ s{/$}{/};
...
-    next;
-  }
+    my $spool_path = mailbox_to_path($mbname);
+    my $meta_path = mailbox_to_path($mbname, 'meta');
 
-  my $spool_path = mailbox_to_path($mbname);
-  my $meta_path = mailbox_to_path($mbname, 'meta');
+    warn "IN: $mbname\n";
+    $mbname =~ s/^\Q$user_inbox\E(\.|\/)?//;
+    warn "OUT: $mbname\n";
```
rjbs: Are these `warn`s meant to stick around in merge?

cassandane/utils/cyr-easycd
wolfsage: Yeah I tested multiple ways to be sure :)

cassandane/utils/cyr-easycd
wolfsage: Sure aren't. Fixed

cassandane/utils/cyr-easycd
elliefm: There's still a smell here, namespace conversion is non-trivial...

I think you should be try to be clear on which namespace the `cyr_ls` output is giving you (my guess: probably "internal"), and which namespace you want to use, and then use `Cyrus::Mbname` to do the translation (if they're the same, there's nothing to do, but leave a comment to that effect).  You might, for example, choose to use something like "x namespace, but with `/` converted to `.`", in which case first do the namespace translation to x, then the regex substitutions.  (Bonus: once we're thinking in terms of namespaces, if `cyr_ls` later changes its output namespace, the updates needed to accommodate that change should be minimal.)

The way to check all this would be to try it on cass tests with/without `:UnixHierarchySep`, `:AltNamespace`, `:NoAltNamespace`, `:VirtDomains`, and `:CrossDomains` magic (maybe some others I'm forgetting...).  You've probably already done that and it works fine as is, but it would be less red-flaggy to read if it used the standard interfaces.

That said, I had a quick look at Cyrus::Mbname because I don't remember using it before myself, and it looks like it doesn't consider configuration, which is a bug because mailbox name spellings vary with configuration.  Sigh.

So I dunno.  I guess at this point I'm not asking you to make any further changes -- if it works it works -- just noting that here be dragons.  If you wanna work deeper into it I'll happily review again, if not no worries.

wolfsage: I think this is the problem I had - it turns out the perl libraries weren't enough.

I just ran a few test cases, and this seems to work. If we come across test runs where it doesn't, we can fix it more then.

Thanks for the review

PR #5687, opened 2025-11-14
Cassandane: introduce helpers for users and data

rjbs:
This produces a system for creating test data easily without thinking too much about the underlying protocols or all the required fields.  There are many types not yet implemented, or defaults not yet well-generated, but I think this is worth merging.  We can continue working on this later, by:

* writing more entity types
* converting existing tests to use this
* adding more useful features to entity types

### DIFF
--- a/cassandane/tiny-tests/JMAPContacts/card_set_create_basic
+++ b/cassandane/tiny-tests/JMAPContacts/card_set_create_basic
@@ -18,11 +18,11 @@ sub test_card_set_create_basic
     my $card = $abook->create_card({
         '@type' => 'Card',
         version => '1.0',
-        uid => $uid,
-        prodId => $prodid,
-        kind => 'individual',
+        uid     => $uid,
+        prodId  => $prodid,
+        kind    => 'individual',
         created => $created,
-        name => { full => $name }
+        name    => { full => $name }
     });
 
     $self->assert_not_null($card->id);


wolfsage:
Nit: rewrite "WIP: introduce a "user" object to tests" commit message

cassandane/Cassandane/TestUser.pm
```diff
@@ -0,0 +1,137 @@
...
+has instance => (is => 'ro', required => 1);
+
+has _common_service_args => (
+    is => 'ro',
+    lazy => 1,
+    default => sub {
+        my ($self) = @_;
+        my $service = $self->{instance}->get_service("https");
+        $service ||= $self->{instance}->get_service("http");
+        return if !$service; # XXX die
```
wolfsage: We should definitely die or croak or something. Let's do that now!

cassandane/Cassandane/TestUser.pm
```diff
@@ -0,0 +1,137 @@
...
+
+        return $caldav;
+    },
+);
+
+sub imap {
+    my ($self) = @_;
+
+    my $imap_svc = $self->instance->get_service('imap');
+    my $imap_store = $imap_svc->create_store(username => $self->username);
```
wolfsage, line 116: Ah! This is the bit I thought was weird. It seems like we're creating something here (A store has specific meaning but maybe only in Fastmail), and this confused me...

cassandane/Cassandane/TestEntity/Factory/Mailbox.pm
```diff
@@ -0,0 +1,52 @@
...
+
+    return;
+}
+
+sub inbox {
+    my ($self) = @_;
+    my $dt = $self->datatype;
+
+    my $jmap = $self->user->jmap;
+    delete $jmap->{CreatedIds};
```
wolfsage: This probably deserves a comment. Or something.

cassandane/Cassandane/TestEntity/Instance/Mailbox.pm
```diff
@@ -0,0 +1,139 @@
...
+    qw( id name )
+}
+
+__PACKAGE__->initialize_accessors;
+
+# This returns an Email::MIME object. It has some problems, but it makes it
+# easy to futz around with in testing where performance isn't really an issue.
+# (It's also what we use in the tests for all the other products, so you don't
+# need to remember what the interface of Mail::MemoryMessage or whatever is.)
+# -- michael, 2019-05-14
```
wolfsage, line 15: Woah, hi michael

cassandane/Cassandane/TestEntity/Instance/Mailbox.pm
```diff
@@ -0,0 +1,139 @@
...
+my sub fake_rfc822 {
+    my ($arg) = @_;
+    $arg //= {};
+    die "fake_rfc822 requires a 'to' argument" unless $arg->{to};
+
+    require Email::MessageID;
+    require Email::MIME;
```
wolfsage, line 22: Wonder if we'll want to update the developer docs to explicitly list these... (they seem to be pulled in as deps of something else anyway though...)

cassandane/tiny-tests/TestEntities/basic_entity_tests
```diff
@@ -0,0 +1,36 @@
...
+{
+    my ($self) = @_;
+
+    my $user  = $self->default_user;
+    my $abook = $user->addressbooks->create;
+
+    my $name = 'Mr. John Q. Public, Esq.';
+
+    my $card = $abook->create_card({
+        name => { full => $name }
```
wolfsage, line 15: Should we assert the name got used?

cassandane/tiny-tests/JMAPContacts/card_set_create_basic
```diff
@@ -9,67 +9,46 @@ sub test_card_set_create_basic
...
-        }, 'R1']
-    ]);
-
-    $self->assert_not_null($res->[0][1]{created}{1});
-    $self->assert_not_null($res->[0][1]{created}{1}{id});
-
-    my $href = $res->[0][1]{created}{1}{'cyrusimap.org:href'};
-    $res = $carddav->Request('GET', $href, '',
+    my $card = $abook->create_card({
+        '@type' => 'Card',
```
wolfsage, line 19: Good chance to align keys/values for readability...

cassandane/tiny-tests/JMAPContacts/card_set_create_basic
```diff
@@ -9,67 +9,46 @@ sub test_card_set_create_basic
...
+        version => '1.0',
+        uid => $id,
+        prodId => $prodid,
+        kind => 'individual',
+        created => $created,
+        name => { full => $name }
+    });
+
+    $self->assert_not_null($card);
```
wolfsage: IMO we should never have to assert these in a test.

cassandane/Cassandane/TestEntity/Instance/ContactCard.pm
```diff
@@ -9,6 +9,28 @@ sub datatype_properties {
...
 
+sub _as_vcard {
+    my ($self, $version) = @_;
+
+    my $href = $self->properties->{'cyrusimap.org:href'};
+    my $res = $self->user->carddav->Request(
+        'GET', $href, '',
+        Accept => "text/vcard; version=$version",
+    );
+
```
wolfsage, line 28: We should assert here that res is successful maybe

cassandane/Cassandane/TestEntity/AutoSetup.pm
```diff
@@ -0,0 +1,99 @@
...
+            as   => 'datatype',
+            code => sub { $datatype },
+        });
+    }
+
+    eval "require $instance_class; 1"
+        || die "Failed to load $instance_class: $@";
+
+    # Avert your eyes, this is nasty.
+    $class->can('with')->('Cassandane::TestEntity::Role::Factory');
```
wolfsage, line 63: Wow!

cassandane/Cassandane/TestUser.pm
```diff
@@ -0,0 +1,154 @@
...
+has _common_service_args => (
+    is => 'ro',
+    lazy => 1,
+    default => sub {
+        my ($self) = @_;
+        my $service = $self->{instance}->get_service("https");
+        $service ||= $self->{instance}->get_service("http");
+        return if !$service; # XXX die
+
+        my $ca_file = Cwd::abs_path("data/certs/cacert.pem");
```
wolfsage, line 21: You should import Cwd

cassandane/Cassandane/TestEntity/Factory/AddressBook.pm
```diff
@@ -1,6 +1,7 @@
 package Cassandane::TestEntity::Factory::AddressBook;
 use Moo;
 
+use lib '.';
```
wolfsage, line 4:
```
Cassandane: add "use lib '.'" to TestEntity classes

ugh
```

I laughed, but also, this commit could use an explainer. I'm surprised it's needed

cassandane/Cassandane/TestEntity/Instance/Email.pm
```diff
@@ -0,0 +1,8 @@
+package Cassandane::TestEntity::Instance::Email;
+use Moo;
+
+use lib '.';
+use Cassandane::TestEntity::AutoSetup properties => [ qw( id from ) ];
```
wolfsage: This is missing quite a lot, no?

cassandane/Cassandane/TestEntity/Instance/AddressBook.pm
```diff
@@ -0,0 +1,17 @@
+package Cassandane::TestEntity::Instance::AddressBook;
+use Moo;
+
+use lib '.';
+use Cassandane::TestEntity::AutoSetup properties => [ qw( id from ) ];
```
wolfsage: Is 'from' right here?

cassandane/Cassandane/TestUser.pm
rjbs: Done in new commit.

cassandane/Cassandane/TestUser.pm
rjbs: Gotcha.  This is really more of a handle on the store than what you might be thinking when you see the name "store".  For example there's a "mbox store" that just reads mail from an mbox.  This is like that, pointing at the "real" storage, the IMAP service.

cassandane/Cassandane/TestEntity/Factory/Mailbox.pm
rjbs: My feeling is that probably we should nerf this CreatedIds-tracking behavior, but I need to learn more.  For now, a comment.  (This will go away if we switch to JMAP::Tester.)

cassandane/Cassandane/TestEntity/Instance/Mailbox.pm
rjbs: These are a bit of a mess, but done.

cassandane/Cassandane/TestEntity/AutoSetup.pm
rjbs: I TOLD YOU NOT TO LOOK

cassandane/Cassandane/TestUser.pm
rjbs: OK

cassandane/Cassandane/TestEntity/Instance/AddressBook.pm
rjbs: No, I'll fix it.

cassandane/Cassandane/TestEntity/Instance/Email.pm
rjbs: Yes, I'll add "everything" sometime soon.

rjbs: Done in a new commit.

cassandane/Cassandane/TestEntity/Factory/Mailbox.pm
rjbs: Done.

cassandane/Cassandane/Cyrus/TestCase.pm
```diff
@@ -894,6 +894,125 @@ sub _need_http_tiny_env
...
 }
 
+package Cassandane::Cyrus::TestCase::User {
+  use Moo;
+
+  has username => (is => 'ro', required => 1);
+  has password => (is => 'ro', required => 1);
+  has instance => (is => 'ro', required => 1);
+
+  has _common_service_args => (
```
elliefm: Contextual detail that was lost in the move from `_setup_http_service_objects`: these args were common to _http_ services, not all services

cassandane/Cassandane/Cyrus/TestCase.pm
```diff
@@ -1079,6 +1170,16 @@ sub _start_instances
...
     }
+
+    my $user = Cassandane::Cyrus::TestCase::User->new({
+        username => 'cassandane',
+        password => 'cassandane',
+        instance => $self->{instance},
+    });
+
+    $self->{default_user} = $user;
+
```
elliefm: A test can have multiple instances running, and a user belongs to an instance...

Mapping a test's default_user to a user on the test's default instance makes some ergonomic sense, but it implies things are simpler than they are.  Perhaps default_user should be per-instance?  I dunno.  If you already know how to write multi-instance tests, you can still do so.  But if you don't know that it's even possible, discoverability is lacking with this API.

cassandane/tiny-tests/TestEntities/basic_entity_tests
```diff
@@ -0,0 +1,36 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_basic_entity_tests
+    :min_version_3_9 :needs_dependency_icalvcard
```
elliefm: `:min_version_3_9` copy and pasted?  Please remove it, so the next test copy and pasted from this doesn't inherit it again ;)

cassandane/Cassandane/TestUser.pm
```diff
@@ -114,41 +114,24 @@ sub imap {
...
-has mailboxes => (
-    is => 'ro',
-    lazy => 1,
-    default => sub {
-        require Cassandane::TestEntity::Factory::Mailbox;
-        return Cassandane::TestEntity::Factory::Mailbox->new({ user => $_[0] });
-    },
-);
+for my $pair (
+  [ addressbooks => 'AddressBook' ],
```
elliefm: nit: four-space indents

cassandane/doc/README.deps
```diff
@@ -160,11 +163,12 @@ Debian/Ubuntu copypasta:
...
        libtest-nowarnings-perl libtest-longstring-perl \
        libtest-warn-perl libnet-ldap-server-perl \
-       libdbd-sqlite3-perl libdigest-crc-perl libxml-simple-perl
+       libdbd-sqlite3-perl libdigest-crc-perl libxml-simple-perl \
+       libemail-mime-perl
 
   sudo cpan Math::Int64 Mail::JMAPTalk Mail::IMAPTalk \
        Net::CalDAVTalk Net::CardDAVTalk String::CRC32 \
-       IO::File::Lockable
+       IO::File::Lockable Email::MIME
```
elliefm, line 172: Add it to one or the other of these copypastas, not both.  The cpan step is for modules that aren't packaged in apt, or the apt version isn't (or wasn't) new enough for our purposes.  Someone following these instructions will install it twice!

docsrc/imap/developer/developer-testing.rst
```diff
@@ -51,6 +52,7 @@ Install and configure Cassandane
         sudo cpan -i XML::Fast
         sudo cpan -i Data::ICal::TimeZone
         sudo cpan -i Text::VCardFast
+        sudo cpan -i Email:MIME
```
elliefm: Ditto earlier comment -- this list of cpan commands is for stuff that wasn't available via the apt command above.

cassandane/Cassandane/TestEntity/Instance/AddressBook.pm
```diff
@@ -2,7 +2,10 @@ package Cassandane::TestEntity::Instance::AddressBook;
 use Moo;
 
 use lib '.';
-use Cassandane::TestEntity::AutoSetup properties => [ qw( id name ) ];
+use Cassandane::TestEntity::AutoSetup properties => [ qw(
+  id name description sortOrder isDefault isSubscribed
```
elliefm: nit: indent

cassandane/Cassandane/TestEntity/Instance/Email.pm
```diff
@@ -2,7 +2,11 @@ package Cassandane::TestEntity::Instance::Email;
 use Moo;
 
 use lib '.';
-use Cassandane::TestEntity::AutoSetup properties => [ qw( id from ) ];
+use Cassandane::TestEntity::AutoSetup properties => [ qw(
+  id blobId threadId mailboxIds keywords size receivedAt
```
elliefm: nit: indent

cassandane/Cassandane/Cyrus/TestCase.pm
rjbs: Renamed.

cassandane/Cassandane/Cyrus/TestCase.pm
rjbs: Right now, tests tended to use `$self->{caldav}` where `$self` is the test.   So, there is implicitly a default user on the test.  I'm not enough of an expert (yet) to say whether I think this is bad, but my train of thought goes:  If we moved default user onto an instance, we'd then have a default instance, and we'd want sugar on the test that was `->default_instance->default_user`, which I'd call `->default_user`.  

I think for now, this is probably okay and pretty close to the old way, but strictly better.  I think writing multi-instance tests should be pretty easy:  you can call ->create_user on each instance you want to.

If you think this is a blocker, LMK, but I don't think you do.  As for discoverability, what we need is docs!

cassandane/doc/README.deps
rjbs: Got it.  I've adjusted these.  I misunderstood how this worked.

cassandane/Cassandane/TestEntity/Instance/Mailbox.pm
mmcclimon: Hiiii!

cassandane/Cassandane/Cyrus/TestCase.pm
elliefm: Yeah, I agree, I don't think this is a blocker, unless you thought about it and decided that it was, which you did and didn't.

(And we can always change it later if our ergonomic needs change.)

elliefm (approved):
Looking good!

elliefm:
This is still marked "changes requested" by @wolfsage.  I believe the requested changes have been made, in which case you can clear the flag and make it mergeable by choosing "dismiss review" from the ellipsis menu next to the "wolfsage requested changes" entry.  You probably already know this, but just in case! :)

cassandane/Cassandane/TestEntity/Instance/ContactCard.pm
wolfsage: This still could use an assert

cassandane/Cassandane/TestEntity/Role/Factory.pm
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    unless ($res->[0][1]{created}{toCreate}) {
+        Carp::confess("failed to create $dt object")
+    }
+
+    $self->get($res->[0][1]{created}{toCreate}{id});
+}
+
+sub _update {
```
wolfsage, line 71: I'd make this public? It's nice that the accessors are read-write, but sometimes $mailbox->update({ a => 1, b => 2 }); is better than individual method calls (especially if you need to pass json path style updates...)

cassandane/tiny-tests/JMAPContacts/card_set_create_basic
wolfsage: I think we should still fix this before it becomes a copied pattern. The test setup methods should always succeed, so we never need to assert that they did

cassandane/tiny-tests/TestEntities/basic_entity_tests
wolfsage: I still think we should!

cassandane/Cassandane/TestEntity/Instance/ContactCard.pm
rjbs: I don't think this needs an assert, because Request should either return in success or die already.  I will double check.

rjbs: Yes, this is correct: 

```perl
  unless ($Response->{success}) {
    confess("ERROR WITH REQUEST\n" .
```

cassandane/tiny-tests/JMAPContacts/card_set_create_basic
rjbs: Fixed in new fixup commit to be squashed.

cassandane/Cassandane/TestEntity/Role/Factory.pm
rjbs: I left this private intentionally. 

First off, I don't know what I want `$instance->update(...)` to return yet.  I _think_ what I want is to switch this whole backend to use JMAP::Tester, then have this return a Response object.  Until then, "a one-property update works or dies" is simple enough.

Also, this update is to _Factory_ not _Instance_, so you don't get `$mailbox->update` from this, it'd need to be in Instance.  Work for next round, imho.

cassandane/tiny-tests/TestEntities/basic_entity_tests
rjbs: Oops, I missed this comment this morning somehow.

rjbs: Fixup commit pushed.

cassandane/Cassandane/TestEntity/Role/Factory.pm
wolfsage: Okay!

cassandane/Cassandane/TestEntity/Instance/ContactCard.pm
wolfsage: Great, my bad

PR #5691, opened 2025-11-19
Cyr 2397 dblookup uid2emails

ksmurchison:


imap/http_dblookup.c
```diff
@@ -304,6 +305,61 @@ static int get_uid2groups(struct transaction_t *txn,
...
+                         &emails, &member_uids);
+
+    /* Lookup group member email addresses (ignore subgroup members) */
+    jmap_get_card_emails(member_uids, CARDDAV_KIND_ANY, abook_sets,
+                         &emails, NULL);
+    strarray_free(member_uids);
+    strarray_fini(&card_uids);
+    jmap_free_abook_sets(abook_sets);
+
+    if (!strarray_size(&emails)) goto done;
```
rsto, line 339: This will leak memory in the `card_uids` string array? You might want to move the declaration of these arrays before the first jump to the `done` label.

imap/jmap_mail_query.h
```diff
@@ -159,6 +159,24 @@ extern int jmap_headermatch_match(struct jmap_headermatch *hm, message_t *msg);
...
+};
+
+extern ptrarray_t *jmap_get_accessible_addressbooks(const char *accountid,
+                                                    const struct auth_state *authstate,
+                                                    const struct namespace *namespace,
+                                                    struct carddav_db *carddavdb);
+
+extern void jmap_free_abook_sets(ptrarray_t *abook_sets);
+
+extern void jmap_get_card_emails(strarray_t *card_uids, unsigned card_kind,
```
rsto, line 184: Please document this and the other newly exported functions. For example, It's not clear to me from this declaration which of the pointer arguments are input and output arguments. Is `abook_sets` an output argument, e.g. is the ptrarray_t empty when calling this function and populated after?

imap/http_dblookup.c
ksmurchison: The card_uids strarray is cleaned right after its used and before we jump to done

rsto: OK

imap/jmap_mail_query.h
ksmurchison: Done

rsto (approved):

PR #5694, opened 2025-11-19
Test Entities: avoid issues with JMAPTalk's "using", add sharing helpers

rjbs:


cassandane/Cassandane/TestEntity/Instance/AddressBook.pm
```diff
@@ -7,11 +7,23 @@ use Cassandane::TestEntity::AutoSetup properties => [ qw(
...
+    $self->user->contacts->create({
+        %$prop,
+        addressBookIds => { $self->id => JSON::true() },
+    });
+}
+
+sub create_card_group {
+    my ($self, $prop) = @_;
 
     $self->user->contacts->create({
```
wolfsage, line 23: Could be contacts->create_group but it's fine

cassandane/Cassandane/TestEntity/Role/ShareableInstance.pm
```diff
@@ -0,0 +1,31 @@
+package Cassandane::TestEntity::Role::ShareableInstance;
+use Moo::Role;
+
+use Sub::Install ();
+
+# $mailbox->share_with($user => [ qw( mayFoo mayBar ) ], $user2 => ...)
+sub share_with {
+    my ($self, @input) = @_;
+
+    @input % 2 && Carp::confess("odd size input to ->share_with -- it must be pairs!");
```
wolfsage, line 15: Kinda surprised input isn't a hash then, keys shouldn't repeat

cassandane/Cassandane/TestEntity/Role/ShareableInstance.pm
```diff
@@ -0,0 +1,31 @@
...
+
+    my %to_update;
+    while (my ($target, $bits) = splice @input, 0, 2) {
+        my $ident = ref $target ? $target->username : $target;
+        my @bits  = ref $bits   ? @$bits : $bits; # the BITS! 🍟
+        $to_update{$ident} = { map {; $_ => JSON::true() } @bits };
+    }
+
+    $self->factory->_update($self => { shareWith => \%to_update });
```
wolfsage: Is this the interface we want? In newt, share_with() is additive, rather than replacing. "unshare_entirely" below implies that this method doesn't replace the entire list or it might be "share_with_entirely".

This feels like a readable way to say "share with a few folks"
```
$a->share_with($user1 => [...]);
$a->share_with($user2 => [...]);
```

Which way do you think this should go? (Could have ->share_with, ->replace_share_with, ... or something)

cassandane/Cassandane/TestEntity/Role/ShareableInstance.pm
rjbs: If it was a hash, it would stringify objects used as keys, which is no good.

cassandane/Cassandane/TestEntity/Role/ShareableInstance.pm
rjbs: This was a mistake on my part, I wanted to make it use patch syntax `shareWith/USER` but I forgot!  I'll fix that.

cassandane/Cassandane/TestEntity/Role/ShareableInstance.pm
wolfsage: Ah, right!

cassandane/Cassandane/TestEntity/Role/ShareableInstance.pm
rjbs: I fixed this.

rjbs:
I've updated instance properties to be lazy and to re-fetch on demand and have a clearer.

I updated `$factory->_update` to clear properties on update.

This should fix patching in the general case.

PR #5697, opened 2025-11-21
append: always fdatasync any staged file

brong:
This API is a horror show and really really bad - there should ALWAYS be a finalise step when the spool file is done, on an object which is more than just the filehandle, but this is what we've got for now, so at least make sure we always fsync.

This commit does that, tracking down all uses of `append_newstage` and ensuring that they always fdatasync.

It also fixes a couple of places where we leaked open FILE* on error.

### DIFF
--- a/imap/carddav_db.c
+++ b/imap/carddav_db.c
@@ -1264,7 +1264,7 @@ static int _carddav_store(struct mailbox *mailbox, struct buf *vcard,
     fprintf(f, "%s", buf_cstring(vcard));
 
     if (fflush(f) || ferror(f) || fdatasync(fileno(f))) {
-        syslog(LOG_ERR, "append_setup(%s) failed: %s",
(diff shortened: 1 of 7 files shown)


imap/carddav_db.c
```diff
@@ -1262,6 +1263,14 @@ static int _carddav_store(struct mailbox *mailbox, struct buf *vcard,
     /* Write the vCard data to the file */
     fprintf(f, "%s", buf_cstring(vcard));
 
+    if (fflush(f) || ferror(f) || fdatasync(fileno(f))) {
+        syslog(LOG_ERR, "append_setup(%s) failed: %s",
```
elliefm: Since we're logging with LOG_ERR and we're about to return IMAP_IOERROR, do we want the usual "IOERROR:" slug on the message?

brong: yeah, we should.  I'll fix all those

brong: Done

PR #5699, opened 2025-11-21
imapd.c: SORT/THREAD should lock cstate to prevent lock inversions

ksmurchison:
This mimics what is done for SEARCH and is primarily for FUZZY SEARCH


rsto (approved):
LGTM

ksmurchison:
@brong Does this look correct to you?  Should we only open conv.db early if we're doing a fuzzy search?

brong (approved):

PR #5700, opened 2025-11-21
Cassandane: abstract some duplicated code, eliminate much "use lib"

rjbs:
This is two unrelated changes that I performed together.

1. Some quota-related methods that were _mostly_ duplicated between a few tests are unified into a mixin.
2. the need to `use lib "."` in almost every Cassandane perl module is eliminated

### DIFF
--- a/cassandane/Cassandane/BuildInfo.pm
+++ b/cassandane/Cassandane/BuildInfo.pm
@@ -42,7 +42,6 @@ use strict;
 use warnings;
 use JSON;
 
-use lib '.';
 use Cassandane::Cassini;
 use Cassandane::Util::Log;
 
--- a/cassandane/Cassandane/Cyrus/Mixin/QuotaHelper.pm
+++ b/cassandane/Cassandane/Cyrus/Mixin/QuotaHelper.pm
@@ -1,0 +1,186 @@
+#!/usr/bin/perl
+#
+#  Copyright (c) 2011-2024 FastMail Pty Ltd. All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions
+#  are met:
+#
+#  1. Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+#
+#  2. Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in
+#     the documentation and/or other materials provided with the
+#     distribution.
+#
+#  3. The name "Fastmail Pty Ltd" must not be used to
+#     endorse or promote products derived from this software without
+#     prior written permission. For permission or any legal
+#     details, please contact
+#      FastMail Pty Ltd
+#      PO Box 234
+#      Collins St West 8007
+#      Victoria
+#      Australia
+#
+#  4. Redistributions of any form whatsoever must retain the following
+#     acknowledgment:
+#     "This product includes software developed by Fastmail Pty. Ltd."
+#
+#  FASTMAIL PTY LTD DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE,
+#  INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY  AND FITNESS, IN NO
+#  EVENT SHALL OPERA SOFTWARE AUSTRALIA BE LIABLE FOR ANY SPECIAL, INDIRECT
+#  OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF
+#  USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
+#  TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE
+#  OF THIS SOFTWARE.
+#
+
+package Cassandane::Cyrus::Mixin::QuotaHelper;
+use strict;
+use warnings;
+
+use File::Path ();
+
+# Utility function to check that quota's usages
+# and limits are where we expect it to be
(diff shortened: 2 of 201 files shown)


cassandane/Cassandane/BuildInfo.pm
```diff
@@ -42,7 +42,6 @@ use strict;
 use warnings;
 use JSON;
 
-use lib '.';
```
elliefm, line 45: The reason we were always doing this manually was to have a clear separation between which dependencies we expected to find in the source distribution, vs which dependencies we expected to be installed somewhere standard.  You'll notice the pattern was always:

```
use Various::Core::Or::CPAN::Modules;

use lib '.';
use Cassandane::Whatever;
```

I believe this is (or was) good general perl security practice because it prevents accidentally running malicious copies of standard modules that have been copied to the current directory (which probably has broader write permissions than the system directories).  For how we actually run Cassandane (i.e. on individual developer machines or in short-lived containers), I don't think this is an issue in practice, just noting the intent.  It might also have been made moot by the `use lib '.';` in testrunner.pl anyway.

cassandane/Cassandane/Cyrus/Mixin/QuotaHelper.pm
```diff
@@ -0,0 +1,186 @@
...
+#  FASTMAIL PTY LTD DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE,
+#  INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY  AND FITNESS, IN NO
+#  EVENT SHALL OPERA SOFTWARE AUSTRALIA BE LIABLE FOR ANY SPECIAL, INDIRECT
+#  OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF
+#  USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
+#  TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE
+#  OF THIS SOFTWARE.
+#
+
+package Cassandane::Cyrus::Mixin::QuotaHelper;
```
elliefm: Can you move this to say, `Cassandane::Mixin::QuotaHelper` please?  We expect everything under `Cassandane::Cyrus` to be a test suite (with the annoying exception of `Cassandane::Cyrus::TestCase`), and this breaks that expectation.

Though I suppose `Cassandane::Cyrus::TestEntities` also breaks that expectation now, and I failed to notice it in the last PR. D'oh.

(The pattern is: cassandane self-tests are in `Cassandane::Test` and `use base Cassandane::Unit::TestCase`. Cyrus tests are in `Cassandane::Cyrus` and `use base Cassandane::Cyrus::TestCase`.  I don't love this pattern but it's the pattern we have.  I don't mind changing the pattern, but not by accident.)

cassandane/Cassandane/Cyrus/Mixin/QuotaHelper.pm
```diff
@@ -0,0 +1,186 @@
...
+use strict;
+use warnings;
+
+use File::Path ();
+
+# Utility function to check that quota's usages
+# and limits are where we expect it to be
+sub _check_usages
```
elliefm: With these functions becoming part of a standard interface, maybe we don't need leading underscores on the names anymore?

elliefm (changes requested):

cassandane/Cassandane/Cyrus/Mixin/QuotaHelper.pm
rjbs: I will move the Mixin namespace up.  I'll move the TestEntity one up when a few other changes in flight land.

cassandane/Cassandane/Cyrus/Mixin/QuotaHelper.pm
rjbs: I'd say that we didn't nee them before, either, really.  A TestCase object isn't something I expect to be instantiated and used outside the context of the test runner.  That said, sure!  I'm adding a commit to un-underscore them.   That is, to *derscore* them.

cassandane/Cassandane/BuildInfo.pm
rjbs: I don't think that would make a coherent security policy.  Any library that we know *will* be loaded from `.`, like Cassandane::Whatever, could do:

```perl
delete $INC{'Some/Core/Library.pm`};
require Some::Core::Library; # reload from less-secure path
```

It makes sense to keep `.` out of the path for globally installed programs, especially those that might be run with escalated privs.  Here, I'm not sure I think it adds much at all.

Also, as you say, `testrunner.pl` just goes ahead and crams dot back into `@INC`. ;-)

elliefm (approved):
Thanks!

cassandane/Cassandane/BuildInfo.pm
elliefm: Looks like I added all those use lib lines in d03d49bef, to fix some problem that was occurring with "newer perls".  What was the problem?  Who knows!  It was already in testrunner.pl at that point, so maybe I was trying to fix the Makefile syntax checks?

rjbs: Almost certainly it was that back in the day, `@INC` by default contained `.`.  It was eventually removed, which was sort of a huge pain in the butt, but it *was* a security issue, for the cases I cited:  somebody runs a server that chdirs into a low-priv area and loads code.  Now your server is owned.

I think even now, `.` is added back during module configuration and installation (from the CPAN) because there was too much test code that relied on this in things that were largely abandoned, but used.  Oy!

elliefm:
**backporting:** will need the Makefile change, and maybe the "remove use lib ." one, but probably skip the rest.

PR #5701, opened 2025-11-22
convert much of Cassandane to JMAP::Tester

rjbs:
JMAP::Tester is a bit more full-featured that JMAPTalk, and provides more methods for easy testing of things.  Right now this branch adds a CallMethods emulation layer, but other methods will need to be emulated or converted if we want to replace the whole thing:

```
tove:~/fm/code/cyrus/imapd$ rg --pcre2 -- 'jmap->(?!Call)' cassandane/  | perl -nle '/jmap->(\w+)/ && print $1' | sort |uniq -c | sort -nr
  99 DefaultUsing
  82 Upload
  55 Download
  35 ua
  28 auth_header
  22 AddUsing
  10 Request
   3 downloaduri
   1 uri
```

### DIFF
--- a/cassandane/Cassandane/TestUser.pm
+++ b/cassandane/Cassandane/TestUser.pm
@@ -120,6 +120,8 @@ sub new_jmaptester ($self, $new_arg = undef) {
     my $scheme = $arg{scheme};
 
     require Cassandane::JMAPTester;
+
+    # This causes all requests and responses to be printed to STDERR.
     local $ENV{JMAP_TESTER_LOGGER} = 'HTTP:-2'
         unless exists $ENV{JMAP_TESTER_LOGGER};
 


cassandane/Cassandane/TestEntity/Instance/ContactCard.pm
```diff
@@ -21,7 +21,7 @@ sub _as_vcard {
 
     my $href = $self->properties->{'cyrusimap.org:href'};
     my $res = $self->user->carddav->Request(
-        'GET', $href, '',
+        'GET', "$href", '',
```
rjbs, line 24: maybe strip it off when building properties instead

cassandane/Cassandane/JMAPTester.pm
```diff
@@ -0,0 +1,20 @@
...
+
+  %headers
+    && Carp::confess("Cassandane::JMAPTester does not support custom HTTP headers");
+
+  my $res = $self->request({
+    methodCalls => $sentences,
+    ($using ? (using => $using) : ()),
+  });
+
+  return $res->as_stripped_triples;
```
wolfsage: If `$res` isn't a successful res this will crash with an unhelpful "Can't call method...". This should assert that the response was successful and if not output the response like the existing CallMethods does

cassandane/Cassandane/TestUser.pm
```diff
@@ -62,6 +62,54 @@ has jmap => (
...
+
+        unless ($self->instance->{config}->get_bit('httpmodules', 'jmap')) {
+            Carp::croak("User JMAP::Tester requested, but jmap httpmodule not enabled");
+        }
+
+        my %arg = $self->_common_http_service_args->%*;
+
+        my $host = $arg{host};
+        my $port = $arg{port};
+
```
wolfsage: I know you'd like to fix it all to be better, but at least for now to be compatible I think you should do this before creating a new JMAP::Tester object:
```
        # Send jmap traffic to STDERR like Mail::JMAPTalk above does
        local $ENV{JMAP_TESTER_LOGGER} = 'HTTP:-2'
            unless exists $ENV{JMAP_TESTER_LOGGER};
```

That way we get roughly the same debug output at the right point in the tests

rjbs: Done.

cassandane/Cassandane/JMAPTester.pm
rjbs: I added assert_success.

rjbs:
All tests now passing.

elliefm:
Might be worth making whatever `use` lines etc explicitly request 0.29+ too, if you haven't already done that, so that trying to run cassandane in a state known to lose failures will instead fail early on the dependency error.

cassandane/Cassandane/TestEntity/Factory/Mailbox.pm
```diff
@@ -19,22 +19,26 @@ sub inbox {
     my $jmap = $self->user->entity_jmap;
     local $jmap->{CreatedIds}; # do not pollute the client for later use
 
-    my $res = $jmap->CallMethods([
-        [ "$dt/query", { filter => { role => 'inbox' } }, 'a' ],
+    my $res = $jmap->request([
+        [ "$dt/query", { filter => { role => 'inbox' } } ],
```
elliefm: `resultOf` a few lines below still references `'a'`, is that broken now?

cassandane/Cassandane/JMAPTester.pm
```diff
@@ -7,6 +7,24 @@ has fallback_account_id => (
...
+
+sub AddUsing {
+    my ($self, @to_add) = @_;
+    my %already_using = map {; $_ => 1 } $self->default_using->@*;
+
+    for my $capa (@to_add) {
+      next if $already_using{$capa}++;
+
+      push $self->default_using->@*, $capa;
+    }
```
elliefm, line 30: nit: mixed indent sizes

cassandane/Cassandane/TestUser.pm
```diff
@@ -123,6 +123,14 @@ sub new_jmaptester ($self, $new_arg=undef) {
 
         api_uri => "http://$host:$port/jmap/",
         upload_uri => "http://$host:$port/jmap/upload/{accountId}/",
+
+        # The session actually providers a query string of "?accept={type}" but
```
elliefm: :thinking: should "providers" be "requires", or "provides"?

cassandane/tiny-tests/JMAPCore/blob_upload_bad_url
```diff
@@ -8,18 +8,24 @@ sub test_blob_upload_bad_url
     my $jmap = $self->{jmap};
 
     xlog "Assert Problem Details report";
-    my $httpReq = {
```
elliefm, line 11: (not related to this line)

The commit with message starting "Cassandane: rework JMAP upload calls for cutover to JMAP::Tester" has a bunch of diffs embedded in the commit message.  I assume that was an editor whoopsy and not intentional?

rjbs: Fixed locally!

cassandane/Cassandane/TestUser.pm
rjbs: Thanks, fixed locally.

cassandane/Cassandane/JMAPTester.pm
rjbs: Thanks, fixed locally.

cassandane/Cassandane/TestEntity/Factory/Mailbox.pm
rjbs: Great catch!  This would not have broken, because JMAP::Tester will start with `a` but that's terrible to rely on!  Fixed locally.

cassandane/Cassandane/JMAPTester.pm
```diff
@@ -0,0 +1,157 @@
...
+extends 'JMAP::Tester';
+
+use experimental 'signatures';
+
+use Encode ();
+
+has fallback_account_id => (
+  is       => 'ro',
+  required => 1,
+);
```
elliefm, line 13: couple more 2-spacers :sweat_smile:

elliefm (approved):
Great.  There's still a couple of 2 space indents that jump out at me, but I don't need to review it again if you only change those.

cassandane/Cassandane/JMAPTester.pm
rjbs: Fixed through amendment.

cassandane/Cassandane/TestUser.pm
```diff
@@ -97,6 +99,63 @@ has entity_jmap => (
...
+    urn:ietf:params:jmap:calendars
+    urn:ietf:params:jmap:contacts
+
+    https://cyrusimap.org/ns/jmap/mail
+    https://cyrusimap.org/ns/jmap/calendars
+    https://cyrusimap.org/ns/jmap/contacts
+
+    https://cyrusimap.org/ns/jmap/performance
+    https://cyrusimap.org/ns/jmap/backup
+    https://cyrusimap.org/ns/jmap/blob
```
wolfsage, line 99: I wonder if we should include debug too?

cassandane/Cassandane/TestUser.pm
```diff
@@ -97,6 +99,63 @@ has entity_jmap => (
...
+);
+
+# Either 0-arg to get a default-config one, or provide just [using...] for
+# custom using, or {k=>v,...} to override constructor args.
+sub new_jmaptester ($self, $new_arg=undef) {
```
wolfsage: nit: `$new_arg = undef` is more readable, do we have a style guidance here?

cassandane/Cassandane/TestUser.pm
```diff
@@ -97,6 +99,63 @@ has entity_jmap => (
...
+    my %overrides;
+    if ($new_arg) {
+        %overrides  = ref $new_arg eq 'HASH'  ? %$new_arg
+                    : ref $new_arg eq 'ARRAY' ? (using => $new_arg)
+                    : Carp::confess("expected hash or array reference to ->jmap_tester, got neither");
```
wolfsage: nit: `->jmap_tester` should be `->new_jmaptester`

wolfsage: We should import Carp in this file

cassandane/Cassandane/TestUser.pm
```diff
@@ -97,6 +99,63 @@ has entity_jmap => (
...
+    }
+
+    my %arg = $self->_common_http_service_args->%*;
+
+    my $host = $arg{host};
+    my $port = $arg{port};
+
+    require Cassandane::JMAPTester;
+    my $jtest = Cassandane::JMAPTester->new({
+        fallback_account_id => $self->username,
```
wolfsage, line 129: nit 1: We use `default_using` but `fallback_account_id`. Feels like this should be `default_account_id`?
nit 2: This commit doesn't introduce fallback_account_id in Cassandane::JMAPTester but uses it here

cassandane/Cassandane/TestUser.pm
```diff
@@ -97,6 +99,63 @@ has entity_jmap => (
...
+
+        api_uri => "http://$host:$port/jmap/",
+        upload_uri => "http://$host:$port/jmap/upload/{accountId}/",
+        default_using => [ @DEFAULT_USING ],
+        %overrides,
+    });
+
+    $jtest->ua->set_default_header(
+        'Authorization',
+        q{Basic } .  MIME::Base64::encode_base64(
```
wolfsage: qbasic! Wow!

wolfsage: We should import MIME::Base64

cassandane/Cassandane/TestEntity/Role/Factory.pm
```diff
@@ -86,21 +77,14 @@ sub _update {
...
     ]]);
 
-    unless ($res->[0][0] eq "$dt/set") {
-        Carp::confess("failed to complete $dt/set call")
-    }
-
-    if ($res->[0][1]{notUpdated}{$id}) {
-        Carp::confess("failed to update $dt object")
-    }
+    my $set = $res->single_sentence("$dt/set")->as_set->assert_no_errors;
```
wolfsage: nit: No need to assign to $set but doesn't hurt

cassandane/Cassandane/JMAPTester.pm
```diff
@@ -2,5 +2,52 @@ package Cassandane::JMAPTester;
...
+# We will start by forbidding custom headers or list context.
+sub Upload {
+    my $self    = shift;
+    my $headers = (ref $_[0] eq 'HASH') ? shift : {};
+    my $bytes   = shift;
+    my $type    = shift;
+    my $accountId = shift // $self->fallback_account_id;
+
+    wantarray
+        && Carp::confess("Cassandane::JMAPTester can't emulate JMAPTalk->Upload in list context");
```
wolfsage, line 77: I think you could using $upload->http_response, but if nothings even using it, no point I suppose

cassandane/Cassandane/JMAPTester.pm
```diff
@@ -7,6 +7,24 @@ has fallback_account_id => (
     required => 1,
 );
 
+# This emulates JMAPTalk's DefaultUsing
+sub DefaultUsing {
```
wolfsage, line 16: Kinda weird with this new code we aren't using subroutine signatures?

cassandane/Cassandane/JMAPTester.pm
```diff
@@ -2,6 +2,8 @@ package Cassandane::JMAPTester;
 use Moo;
 extends 'JMAP::Tester';
 
+use experimental 'signatures';
```
wolfsage, line 5: Here we pull in sigs but we aren't using them!

cassandane/Cassandane/JMAPTester.pm
```diff
@@ -67,5 +69,60 @@ sub Upload {
...
+      (defined $blob_id    ? (blobId    => $blob_id   ) : ()),
+      (defined $name       ? (name      => $name      ) : ()),
+    },
+    \%download_arg,
+  );
+
+  my $have_headers = $download->http_response->headers;
+  my %tiny_headers;
+
+  $have_headers->scan(sub ($k, $v) {
```
wolfsage: nit: A comment here could be nice to explain what we're doing/why, but it's fine

cassandane/tiny-tests/JMAPEmail/email_legacy_ids
```diff
@@ -165,14 +165,12 @@ EOF
...
-            'Authorization' => $jmap->auth_header(),
-        },
-        content => '',
-    };
-    my $RawResponse = $jmap->ua->get($jmap->uri(), $RawRequest);
-    my $session = eval { decode_json($RawResponse->{content}) };
+    # JMAPTalk doesn't have a method to get the session.  JMAPTester does, but
+    # (for now) only if you also reconfigure from it.  To avoid doing that,
+    # we'll make a new JMAPTester and get its session.  Silly, but it'll work.
+    my $auth = $jmap->get_client_session;
```
wolfsage, line 167: I don't understand this comment at all. Where are we making a new jmaptester and gettings its session?

cassandane/Cassandane/TestUser.pm
```diff
@@ -118,6 +118,9 @@ sub new_jmaptester ($self, $new_arg=undef) {
     my $port = $arg{port};
 
     require Cassandane::JMAPTester;
+    local $ENV{JMAP_TESTER_LOGGER} = 'HTTP:-2'
```
wolfsage, line 125: > Cassandane: when using JMAP::Tester, log telemetry to stderr 

This commit message could use a "why" -> "To mimic JMAPTalk's logging of all requests/responses in test output"

cassandane/tiny-tests/FastMail/rename_deepuser_standardfolders
```diff
@@ -67,6 +67,7 @@ sub test_rename_deepuser_standardfolders
     $self->{instance}->getsyslog();
     my $res = $admintalk->rename('user.cassandane', 'user.newuser');
     $self->assert(not $admintalk->get_last_error());
+    $self->{jmap}->set_username_and_password('newuser', 'pass');
```
wolfsage, line 70: Looks like we're still fiddling below on line (81|82):
```
$self->{jmap}->{user} = 'newuser';
```

cassandane/Cassandane/Cyrus/Conversations.pm
```diff
@@ -260,32 +260,40 @@ sub test_move_200
...
     xlog $self, "create bar mailbox";
-    $res = $jmap->CallMethods([
+    $res = $jmap->CallMethods(
+        [
             ['Mailbox/set', { create => { "1" => {
                             name => "bar",
              }}}, "R1"]
-    ]);
+        ],
+        [ qw( urn:ietf:params:jmap:core urn:ietf:params:jmap:mail ) ],
```
wolfsage, line 269: The issue is the test doesn't enable jmap_nonstandard_extensions, so 'https://cyrusimap.org/ns/jmap/mail' isn't allowed. The commit message isn't quite clear on that

cassandane/tiny-tests/JMAPBlob/blob_get
```diff
@@ -34,11 +34,15 @@ sub test_blob_get
     $self->assert_not_null($blobId);
 
     xlog "Test without capability";
```
wolfsage, line 36: The commit message for this change: "This commit just makes adds a clear" -> "just makes adds" feels like a typo

cassandane/tiny-tests/JMAPBlob/blob_get
```diff
@@ -34,11 +34,15 @@ sub test_blob_get
...
-    # XXX: this will be replaced with the upstream one
-    $jmap->AddUsing('https://cyrusimap.org/ns/jmap/blob');
+    my @using = grep {; $_ ne 'https://cyrusimap.org/ns/jmap/blob' }
+                $jmap->default_using->@*;
+
+    $res = $jmap->CallMethods(
+        [['Blob/get', { ids => [$blobId], properties => [ 'data:asText', 'size' ] }, 'R1']],
+        \@using,
+    );
+    $self->assert_str_equals($res->[0][0], 'error');
```
wolfsage, line 45: This should also assert that [0][1] is { type => "unknownMethod" } or it could be any kind of error at all

wolfsage: (Same for later tests)

I think this is important especially because these two API calls are not equiv - this one doesn't use data:asBase64, the one below does, etc etc

cassandane/tiny-tests/JMAPEmail/misc_download_shared
```diff
@@ -45,10 +45,6 @@ sub test_misc_download_shared
     xlog $self, "Unshare mailbox";
```
wolfsage, line 46: Not your bug, but test above downloads the file and never confirms the download succeeded. It should!

cassandane/tiny-tests/JMAPMailbox/mailbox_move_to_deleted_parent
```diff
@@ -8,37 +8,14 @@ sub test_mailbox_move_to_deleted_parent
...
-                }
-            }, "R1"]]
-        }',
-    };
-    my $RawResponse = $jmap->ua->post($jmap->uri(), $RawRequest);
-    if ($ENV{DEBUGJMAP}) {
-        warn "JMAP " . Dumper($RawRequest, $RawResponse);
-    }
-    $self->assert($RawResponse->{success});
+    my $res = $jmap->CallMethods([["Mailbox/set", {
```
wolfsage, line 11: I'm not sure I agree with this change and the similar one below. The comment above `# Assert mailboxes are created in the right order` suggests that it's testing if the JMAP parser sees "C" first, it still creates "A" first. I suppose this all depends on what JANSSON is doing too, so maybe in the end it doesn't matter, but it certainly seems like the explicit hand crafted JSON was intentional

cassandane/Cassandane/JMAPTester.pm
```diff
@@ -42,7 +42,7 @@ sub CallMethods {
         ($using ? (using => $using) : ()),
     });
 
-    return $res->as_stripped_triples;
+    return $res->assert_successful->as_stripped_triples;
 }
```
wolfsage, line 47: Nit on this commit message: "and will instead do provide a" "instead do provide a" seems like a typo

cassandane/Cassandane/TestUser.pm
rjbs: I'm going to leave it out for now and make a note to come back to this when I know more about its impact.

cassandane/Cassandane/TestUser.pm
rjbs: No, but the space is what we'd normally do, but Python style is to have no space, so sometimes I get confused.

rjbs: (So I've fixed it.)

cassandane/Cassandane/TestUser.pm
rjbs: Okay, done.  I was going to argue "not needed, because Moose actually imports `confess` for you, so you know it is loaded," but Moo doesn't do that, I see!

cassandane/Cassandane/TestUser.pm
rjbs: nit 1:  I called it fallback because there _isn't_ a default in JMAP, and I hate that we need this, and it should go away, not be enshrined.  I'm gonna push back on renaming, but honestly I don't care that much.  It just wan't a mistake.

nit 2:  Oops.  I _might_ fix that, but as you can imagine this was originally a huge change that I teased apart into _n_ commits, so there might be some other secret dependency here.  if I can't fix this fast, I may say "so it goes"

cassandane/tiny-tests/JMAPBlob/blob_get
rjbs: I am not enthusiastic about taking this point.  Right now, this branch is meant to fix the "how" and not the "what" we are testing.  You're saying that this test should *assert what kind of error we get*, which means that to refactor this code, I have to go look through the specs to consider whether the existing test is rigorous enough.  I will do it, but I get prickly when refactoring work gets pushed to become unrelated imrovements.

cassandane/Cassandane/JMAPTester.pm
rjbs: Fixed.

cassandane/tiny-tests/JMAPEmail/misc_download_shared
rjbs: Done, with similar, "I'm here to move the furniture, not paint the walls" comment. ;-)

cassandane/tiny-tests/JMAPMailbox/mailbox_move_to_deleted_parent
rjbs: I thought about that and decided I didn't think it was needed.  But because you and I both wondered about it, I have gone and asked others.  The consensus is that generating in Perl is okay.  I'll add this discussion to the commit message.

cassandane/Cassandane/TestUser.pm
rjbs: [🍌 🍌 🍌](https://en.wikipedia.org/wiki/Gorillas_(video_game))

rjbs: Fixed.

cassandane/Cassandane/TestEntity/Role/Factory.pm
rjbs: Fixed.

cassandane/Cassandane/JMAPTester.pm
rjbs: The problem is that some *can* and some can't, because of things like optional middle parameters.  Or I wasn't sure how many of the existing parameters were optional and didn't want to screw up arity.  I can tidy this up, but basically I left the "these should be destroyed" methods sort of as-is.

Unless you feel strongly, I'd like to just leave it as is and focusing on removing this code in phase 3 or 4…

cassandane/Cassandane/JMAPTester.pm
rjbs: They're using in set_username_and_password.

cassandane/tiny-tests/JMAPEmail/email_legacy_ids
rjbs: This comment is outdated.  I improved JMAP::Tester to make this work, but forgot to remove the comment!  Doing so now.

cassandane/tiny-tests/FastMail/rename_deepuser_standardfolders
rjbs: Thanks, this no longer did anything (because we already set the username).  I've removed the line.

cassandane/Cassandane/JMAPTester.pm
rjbs: This'll be better when emulation is removed.

cassandane/Cassandane/JMAPTester.pm
rjbs: Done.

cassandane/Cassandane/TestUser.pm
rjbs: It's not so much "To mimic", because *all* the clients (IMAP, CalDAV) do this stuff.  I've just put in a comment to say what this does, since it's a bit obscure, rather than to reference the soon-to-be-forgotten Mail::JMAPTalk.

PR #5706, opened 2025-11-25
jmap_contact.c: rewrite conversion of OnlineService object

rsto:
jmap_contact.c: rewrite conversion of OnlineService object

This implements converting JSContact OnlineService objects from
and to vCard.

All of the following vCard properties convert to OnlineService:
- IMPP, the most widely support property, only supports URIs.
- SOCIALPROFILE, the property we defined in RFC 9554.
- XM-FM-ONLINE-OTHER, a custom property we used prior to JMAP
  for Contacts. Supposed to only store a username, not an URI.
- X-SOCIAL-PROFILE, a custom property we used prior to JMAP
  for Contacts. This likely was a misnomer and should have been
  X-SOCIALPROFILE.
- X-CYRUS-ONLINESERVICE, a custom property that we'll use until
  we either see other implementations actually support the
  SOCIALPROFILE property or any other property that IETF will
  come up with.

We do not convert Apple's X-SOCIALPROFILE but keep it as unknown
property in the vCard data.

When creating an OnlineService object we:
- Use IMPP if the 'user' property is not set.
- Use X-CYRUS-ONLINESERVICE otherwise.

When updating an OnlineService object, we try to keep using
whatever property this OnlineService was converted from, but
there are exception. We:
- Convert any X-FM-ONLINE-OTHER and X-SOCIAL-PROFILE properties
  to X-CYRUS-ONLINESERVICE.
- Convert an IMPP property to X-CYRUS-ONLINESERVICE if its
  uri property got removed, or if the user property gets set
  and the IMPP property didn't already have the X-USER or
  USERNAME parameter set.

-----
**Reviewers**: This pull request requires a custom libicalvcard build to pass CI. It includes a couple of bug fixes and a change to the build script, outlined in this pull request: https://github.com/cyrusimap/cyruslibs/pull/8


cassandane/tiny-tests/JMAPContacts/card_set_onlineservices
```diff
@@ -0,0 +1,303 @@
+#!perl
+use Cassandane::Tiny;
+
+use Digest::MD5 qw(md5_hex);
+
+my sub install_test {
+    my ( $action, $name, $arg ) = @_;
```
rjbs: as of yesterday you can instead write:

```perl
my sub install_test ($action, $name, $arg) {
```

Note that this enforces strict arity:  Without a default, an argument is mandatory.

rsto: Thanks. Done.

rsto:
@ksmurchison CI currently fails for this because the `JMAPContacts/card_get_v3` contains this test data:

```
IMPP;PROP-ID=OS1;X-SERVICE-TYPE=Facebook:foo
```

The test expects that the `user` property of the OnlineService is set to `foo`. The implementation in this pull request instead sets it in the `uri` property, even though `foo` is not a valid URI.

Did you see  real-world vCard data where the `IMPP` property value is set to a user name instead of an URI?

ksmurchison (approved):
Looks like it does what it says on the tin.

cassandane/tiny-tests/JMAPContacts/card_set_onlineservices
```diff
@@ -0,0 +1,320 @@
+#!perl
+use Cassandane::Tiny;
+
+use Digest::MD5 qw(md5_hex);
+
+my sub install_test($name, $arg) {
+    my $action = (split /_/, $name)[0];
```
rjbs: usual idiom here is:

```perl
my ($action) = split /_/, $name;
```

It's _nearly_ the same, and just more like what our Perl programmers expect to see.  (The "nearly" is barely interesting, but I'll note it:  if you assign to a list of _N_ variables, the runtime will know that it can split into N+1 parts, with the last ignored, so it acts as if you supplied N+1 as the 3rd argument to split.  If you use the construction here, it will split into as many parts as possible.  In this case, it's sure not to make a performance difference.)

cassandane/tiny-tests/JMAPContacts/card_set_onlineservices
```diff
@@ -0,0 +1,320 @@
...
+    die "name must start with 'create_' or 'update_'"
```
rjbs: Replace `die X` with `Carp::croak(X)`.  Either way you get:

```
X at ${file}, line ${line}
```

With die, it's the file/line of the `die` statement.  With croak, it will be the caller — usually much more useful.

cassandane/tiny-tests/JMAPContacts/card_set_onlineservices
```diff
@@ -0,0 +1,320 @@
...
+      unless $action eq 'create' || $action eq 'update';
+
+    $arg->{desc} //= do {
+        my ( undef, $file, $line ) = caller;
+        "test declared in $file at line $line",;
```
rjbs: weird unneeded comma on this line

cassandane/tiny-tests/JMAPContacts/card_set_onlineservices
```diff
@@ -0,0 +1,320 @@
...
+    };
+
+    require Sub::Install;
+    Sub::Install::install_sub(
+        {
+            as   => "test_card_set_onlineservices_$name",
+            code => sub {
```
rjbs: You could move the `($self) = @_` up into the signature here: `code => sub ($self) {` — but it's not a barrier to merging!

cassandane/tiny-tests/JMAPContacts/card_set_onlineservices
```diff
@@ -0,0 +1,320 @@
...
+    wantVcard => ['X-CYRUS-ONLINESERVICE;X-SERVICE-TYPE=Github;PROP-ID=[\w]+:user;https://uri.local'],
+});
+
+install_test(update_xfmonlineother => {
+    fromVcard => ['X-FM-ONLINE-OTHER;LABEL=Foo:https://uri.local'],
+    update    => { nicknames => { 1 => { name => 'nickname' } } },
+    wantVcard => ['X-CYRUS-ONLINESERVICE;X-SERVICE-TYPE=Foo;PROP-ID=[\w]+:;https://uri.local'],
+});
+
+sub assert_card_set_create_onlineservices($self, $tc)
```
rjbs: Small nit:  I'm not sure what `$tc` is named for.  I _guessed_ "test case", but it isn't the TestCase object, because `$self` is the TestCase object.  Test::Unit calls the individual `test_FOO` subroutines "test methods", but this isn't really that, either.  The caller just calls it `$arg` which might be less confusing here.  It's low-information, but at least it's not confusing. :)

cassandane/tiny-tests/JMAPContacts/card_set_onlineservices
```diff
@@ -0,0 +1,320 @@
...
+    my $id = $res->[0][1]{created}{card1}{id};
+    $self->assert_not_null($id);
+
+    xlog $self, "Assert Card object";
+    $self->assert_deep_equals( $tc->{onlineServices},
+        $res->[1][1]{list}[0]{onlineServices} );
+
+    xlog $self, "Assert vCard";
+
+    # TODO: do not match vCard verbatim
```
rjbs: Once upon a time, there was a suggestion that all TODO should have associated tickets.  Also, we made a decision about standardizing TODO v. XXX, which I'd have to look up to find (and my time today is running short).  But at least double-check that we're using the right one? :)

(Note to self: the github workflow related to this was added 2025-03-25, q.v. `tools/find-fixme-markers`)

cassandane/tiny-tests/JMAPContacts/card_set_onlineservices
rsto: good to know, thanks

cassandane/tiny-tests/JMAPContacts/card_set_onlineservices
rsto: Right, I removed the TODO. I'm sure I'll remember to update this code when we have more sophisticated vCard testing in Cassandane.

cassandane/tiny-tests/JMAPContacts/card_set_onlineservices
rsto: Yep, when I started reading your comment I immediately thought of using `arg` as argument name. Then you suggested the same.

PR #5708, opened 2025-11-26
remove libcap support

elliefm:
This PR removes support for Linux capabilities(7) (libcap).

We recently tried this feature out at Fastmail, and discovered that the implementation caused problems with pipe buffer size limits.  The feature didn't provide enough benefit to justify the effort to fix it.

Given that we know it's broken, and don't intend to fix, use, or maintain it, we're removing the feature so it doesn't trip up anyone else.

### DIFF
--- a/changes/next/cyr-2428-remove-libcap-support
+++ b/changes/next/cyr-2428-remove-libcap-support
@@ -1,0 +1,23 @@
+Description:
+
+libcap support has been removed
+
+
+Documentation:
+
+None
+
+
+Config changes:
+
+None
+
+
+Upgrade instructions:
+
+None
+
+
+GitHub issue:
+
+None


changes/next/cyr-2428-remove-libcap-support
```diff
@@ -0,0 +1,23 @@
+Description:
+
+libcap support has been removed
```
wolfsage, line 3: When big (does this count?) changes like this are announced, should we provide some justification or a link to find out where/why this is happening?

As a consumer of such changelog entries, having to track down the Why for something I care about is a bummer

elliefm: In this specific case, I'd include a link to this PR in the release notes.  The justification is in the PR description, so the bullet under "Major changes since 3.12" in the release notes can be concise.

I don't think this is a big change though: given nobody noticed it was broken before us, that suggests it wasn't being used much, if at all.  Probably very few people knew it existed, since it wasn't documented anywhere except `./configure --help`.  I think this feature was originally a patch from Fastmail anyway (see 2e9a36cd49d8), so it's interesting that we also never used it until just recently.

If I did think it was a big change, it'd get its own section in the release notes, and I'd copy (or write) the story in there.  See ["storage changes" from the 3.6 release notes](https://www.cyrusimap.org/imap/download/release-notes/3.6/x/3.6.0.html#storage-changes) for an example of this sort of thing.

Issue #5647, opened 2025-10-15
SEGV in httpd when listing another users calendar


narc-Ontakac2:
```gdb
(gdb) bt
#0  0x00007fd85f775262 in __GI___libc_free (mem=0x840) at ./malloc/malloc.c:3453
#1  0x00007fd86074ca86 in _dlist_clean (dl=0x562e49a7bbf0) at imap/dlist.c:378
#2  0x00007fd86074ca05 in _dlist_clean (dl=<optimized out>) at imap/dlist.c:372
#3  dlist_free (dlp=dlp@entry=0x7ffc98ec8718) at imap/dlist.c:755
#4  0x0000562e2d13bca0 in caldav_get_validators (mailbox=0x562e499020c0, data=<optimized out>, userid=0x562xxxxxxxxx "someone", record=0x7ffc98ec8960, 
    etag=0x7ffc98ec87d0, lastmod=0x0) at imap/caldav_util.c:202
#5  0x0000562e2d1659ab in propfind_getetag (name=0x562e498a6500 "getetag", ns=0x562e498a64c0, fctx=0x7ffc98ec8ed0, prop=<optimized out>, 
    resp=<optimized out>, propstat=0x7ffc98ec8850, rock=0x0) at imap/http_dav.c:1973
#6  0x0000562e2d164943 in xml_add_response (fctx=fctx@entry=0x7ffc98ec8ed0, code=code@entry=0, precond=precond@entry=0, desc=desc@entry=0x0, 
    location=location@entry=0x0) at imap/http_dav.c:1594
#7  0x0000562e2d165229 in propfind_by_resource (rock=0x7ffc98ec8ed0, data=0x7ffc98ec8be0) at imap/http_dav.c:5942
#8  0x00007fd8606061ff in sqldb_exec (open=0x562e498a68f0, 
    cmd=cmd@entry=0x7fd860812a80 "SELECT  ical_objs.rowid, ical_objs.creationdate, ical_objs.mailbox, ical_objs.resource, ical_objs.imap_uid, ical_objs.lock_token, ical_objs.lock_owner, ical_objs.lock_ownerid, ical_objs.lock_expire, i"..., bvals=bvals@entry=0x7ffc98ec8c80, 
    cb=cb@entry=0x7fd8607ce0f0 <read_cb>, rock=rock@entry=0x7ffc98ec8bb0) at lib/sqldb.c:432
[...]

PR #5709, opened 2025-11-26
Caldav.pm: test calendar-query report on shared calendar

ksmurchison:
This is a test to replicate a crasher reported against Cyrus 3.10.2
(https://github.com/cyrusimap/cyrus-imapd/issues/5647)
which appears to have been fixed in a later version

### DIFF
--- a/imap/caldav_util.c
+++ b/imap/caldav_util.c
@@ -170,10 +170,12 @@ EXPORTED int caldav_get_validators(struct mailbox *mailbox, void *data,
     if ((namespace_calendar.allow & ALLOW_USERDATA) &&
         cdata->dav.imap_uid && cdata->comp_flags.shared &&
         caldav_is_personalized(mailbox, cdata, userid, &userdata)) {
-        struct dlist *dl;
+        struct dlist *dl = NULL;
 
         /* Parse the userdata and fetch the validators */
-        dlist_parsemap(&dl, 1, buf_base(&userdata), buf_len(&userdata));
+        r = dlist_parsemap(&dl, 1, buf_base(&userdata), buf_len(&userdata));
+        buf_free(&userdata);
+        if (r) return r;
 
         if (etag) {
             struct message_guid *userdata_guid;
@@ -201,7 +203,6 @@ EXPORTED int caldav_get_validators(struct mailbox *mailbox, void *data,
         }
 
         dlist_free(&dl);
-        buf_free(&userdata);
     }
     else if (cdata->comp_flags.defaultalerts) {
         if (etag) {


cassandane/tiny-tests/Caldav/calendar_query_shared
```diff
@@ -0,0 +1,78 @@
...
+EOF
+
+    xlog "Run calendar-query report on shared calendar";
+    my $res = $CalDAV->Request('REPORT', $sharedId, $xml,
+                               'Content-Type' => 'application/xml',
+                               'Depth' => 1
+    );
+    $self->assert_str_equals($href1,
+                             $res->{'{DAV:}response'}[0]{'{DAV:}href'}{content});
+    $self->assert_str_equals($href2,
```
wolfsage: Is the ordering guaranteed to always come back with href1 first?

wolfsage (commented):
Could you update the commit message to provide more details about what the bug is?

The message as is doesn't provide much detail and is maybe missing a word:
```
This is a test to replicate a bug reported against Cyrus 3.10.2
which appears to have been since in  later version
```

"have been since in  later version" <- What's this supposed to say? (Also extra space between in/later)

If there's a bug report in github, a link to that would be helpful

cassandane/tiny-tests/Caldav/calendar_query_shared
ksmurchison: Probably not.  Can I use bag() or something else to check for these two hrefs in this response:
```

<?xml version="1.0" encoding="utf-8"?>
<D:multistatus xmlns="urn:ietf:params:xml:ns:caldav" xmlns:D="DAV:">
  <D:response>
    <D:href>/dav/calendars/user/cassandane/test.Default/DA48EC0E-CAF2-11F0-B4D9-EC568C8A99EF.ics</D:href>
    <D:propstat>
      <D:prop>
        <D:getetag>"0954632be601eaddbfe85ac68cf7ab3cbf551b9c"</D:getetag>
      </D:prop>
      <D:status>HTTP/1.1 200 OK</D:status>
    </D:propstat>
  </D:response>
  <D:response>
    <D:href>/dav/calendars/user/cassandane/test.Default/DA513E22-CAF2-11F0-B4D9-EC568C8A99EF.ics</D:href>
    <D:propstat>
      <D:prop>
        <D:getetag>"0bea9a08d01d5c7afbb9f366869bf106e76e794f"</D:getetag>
      </D:prop>
      <D:status>HTTP/1.1 200 OK</D:status>
    </D:propstat>
  </D:response>
</D:multistatus>

```

wolfsage: Yeah something like..
```
use Test::Deep ':v1';

[...]

$self->assert_cmp_deeply(
    bag($href1, $href2),
    [
        $res->{'{DAV:}response'}[0]{'{DAV:}href'}{content},
        $res->{'{DAV:}response'}[1]{'{DAV:}href'}{content},
    ],
);

```

ksmurchison: Done

cassandane/tiny-tests/Caldav/calendar_query_shared
```diff
@@ -0,0 +1,82 @@
...
+  </D:prop>
+  <filter>
+    <comp-filter name="VCALENDAR">
+      <comp-filter name="VEVENT"/>
+    </comp-filter>
+  </filter>
+</calendar-query>
+EOF
+
+    xlog "Run calendar-query report on shared calendar";
```
rsto: This is missing `$self`

rsto (changes requested):
Please indicate in the title of the commit message that this is a Cassandane change. Looking back at the Git history, we do that by prefixing the title with the name of the test suite, so the prefix `Caldav:` or `Caldav.pm:` looks like a good choice to me.

The title could also more clearly explain what this test actually achieves, rather than just stating that a new test function got added. The xlog message at the end of the test function already describes this well in my opinion, so the commit title could simply be "test calendar-query report on a shared calendar"

cassandane/tiny-tests/Caldav/calendar_query_shared
ksmurchison: Fixed

PR #5710, opened 2025-11-28
Twom speedups

brong:
This MR came from some shower thoughts.

1) I always meant to add `yield` support to mupdate and then remove always-yield from our use of twom.  Doing that really reduces the number of locks that are needed for a non-transactional foreach, which is a big win for reducing syscalls in a bunch of our code.  So I did that.

2) I realised that we didn't need to repeat the HEADLOCK every time we get a file change, or re-lock within a foreach.  It's OK to favour currently running clients more, and that ALSO reduces the number of syscalls.

So I did both of those.

### DIFF
--- a/lib/cyrusdb_sql.c
+++ b/lib/cyrusdb_sql.c
@@ -921,6 +921,9 @@ HIDDEN struct cyrusdb_backend cyrusdb_sql =
     &cyrusdb_generic_noarchive,
     NULL,
 
+    NULL, /*yield*/
+    NULL, /*yieldall*/
+
     &myopen,
     &myclose,
 
--- a/lib/cyrusdb_sql.c
+++ b/lib/cyrusdb_sql.c
@@ -922,7 +922,6 @@ HIDDEN struct cyrusdb_backend cyrusdb_sql =
     NULL,
 
     NULL, /*yield*/
-    NULL, /*yieldall*/
 
     &myopen,
     &myclose,
--- a/lib/twom.c
+++ b/lib/twom.c
@@ -46,6 +46,13 @@
 /* release lock in foreach at least every N records */
 #define FOREACH_LOCK_RELEASE 1024
 
+/* When re-locking during a long foreach we normally use TWOM_ONELOCK to skip
+ * the header "gate" lock and save syscalls.  But the gate is what lets a
+ * waiting writer fend off new readers, so if we ALWAYS skipped it a busy set
+ * of overlapping foreaches would starve writers indefinitely.  So every N
+ * re-locks we fall back to the full double-lock, bounding writer starvation. */
+#define FOREACH_GATE_RELOCKS 64
+
 /* type aliases */
 #define LLU long long unsigned int
 #define LU long unsigned int
@@ -133,6 +140,7 @@ struct twom_txn {
     struct tm_file *file;
     size_t end;
     uint64_t counter;
+    uint64_t nlocks;
     unsigned readonly:1;
     unsigned nosync:1;
     unsigned noyield:1;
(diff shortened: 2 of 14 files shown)


lib/cyrusdb_sql.c
```diff
@@ -918,6 +918,9 @@ HIDDEN struct cyrusdb_backend cyrusdb_sql =
     &cyrusdb_generic_noarchive,
     NULL,
 
+    NULL, /*yield*/
+    NULL, /*yieldall*/
+
```
elliefm, line 925: Is "yieldall" cruft that got left behind from an earlier sketch?  All the other cyrusdb_foo.c are only adding a "yield" line here.  I think this is breaking the struct, `myopen` and `myclose` and the rest are gonna be a slot off from where the struct expects them to be.  Surprised this compiles?!

lib/twom.c
```diff
@@ -1549,13 +1549,14 @@ static int write_lock(struct twom_db *db, struct twom_txn **txnp,
     struct flock fl;
     int r = 0;
     int cmd = (flags & TWOM_NONBLOCKING) ? F_SETLK : F_SETLKW;
+    int take_headlock = (flags & TWOM_ONELOCK) ? 0 : 1;
```
rsto, line 1566: nit: this could be type `bool`, now that we agreed to use C11 or later.

lib/twom.h
```diff
@@ -46,6 +46,7 @@ enum twom_flagspec {
     TWOM_NOCSUM          = 1<<2,    /* Don't check checksums on read */
     TWOM_NOSYNC          = 1<<3,    /* Don't msync/fsync on write */
     TWOM_NONBLOCKING     = 1<<4,    /* When taking a lock, return immediately if the file is already locked */
+    TWOM_ONELOCK         = 1<<5,    /* When locking, skip the double-lock process */
```
rsto, line 49: Is there ever a chance that we want twom callers to decide if to double-lock? It seems to me as if that could be kept inside twom.c.

rsto (commented):
Other than my question about if to keep the ONELOCK flag internal to the implementation, I am not sure I get the point of adding `yield` support in this pull request. Does yield actually *do* something already, or is this PR just adding the groundwork to benefit from using yield vs. yieldall in some later change?

lib/cyrusdb_sql.c
brong: Yeah, must be.  Hmm.  I GUESS this thing isn't compiled in for some reason.

lib/twom.h
brong: @rsto they might know that they're running in an environment where they can't get lock starvation for some other reason, so this would reduce the number of sysctls required.

Also, it's useful when you're releasing and re-locking a few times in a single operation, for the same reason that we're using it internally - it makes the caller more greedy at grabbing the lock (though it's not a win guarantee!)

Basically - to get fairness, you want every process to go through the full double-lock "every so often"; so others get a chance - but not doing it every time is fine if you're releasing and re-locking a lot.

rsto: OK, thanks.

lib/cyrusdb_sql.c
brong: Fixed with a patch to cyrus-docker which builds cyrusdb_sql, and by stripping that line.  Good catch!

lib/twom.c
brong: I'm going to leave this one - twom.c is going to be extracted out into its own separate repo soon, and I'm not sure what to target for that.

rsto (approved):
LGTM

elliefm (approved):

brong:
Marking "DO NOT MERGE" for now, was implicated in some test failures at Fastmail so we're going to do some more investigation about whether there's a deadlock possibility that I missed here.

brong:
> Marking "DO NOT MERGE" for now, was implicated in some test failures at Fastmail so we're going to do some more investigation about whether there's a deadlock possibility that I missed here.

I've now fixed up the starvation issue, and also (thanks Claude) reviewed the mupdate calls

brong:
Just dumping this, it's not working out and probably all bad ideas.

PR #5711, opened 2025-11-30
get Cassandane Test::Core tests running in CI

elliefm:
This arranges for the docker container to run in privileged mode, which lets us set the kernel.core_pattern sysctl to a value that Cassandane knows how to work with.
    
The Test::Core tests aren't directly interesting, but if they can run and pass, that proves that Cassandane can detect when a core file was generated by a running test.  Tests that generate core files automatically fail.
    
This means PRs that introduce crash bugs will now fail in CI... as long as the crashing code is exercised by a test, anyway.


elliefm:
(rebased on latest master, and dropped the temporary commits)

PR #5714, opened 2025-12-02
jmap_contact_from_vcard(): use PREF param for flagging isDefault email

ksmurchison:
We might be reading a v4 vCard which uses PREF=<n> rather than TYPE=pref which is used by v3

### DIFF
--- a/cassandane/tiny-tests/JMAPContacts/contact_get_emails_v4
+++ b/cassandane/tiny-tests/JMAPContacts/contact_get_emails_v4
@@ -1,0 +1,57 @@
+#!perl
+use Cassandane::Tiny;
+use Test::Deep ':v1';
+
+sub test_contact_get_emails_v4
+    :JMAPExtensions
+    ($self)
+{
+    my $jmap = $self->{jmap};
+    my $carddav = $self->{carddav};
+
+    xlog $self, "create a v4 contact with multiple emails";
+    my $uid = '816ad14a-f9ef-43a8-9039-b57bf321de1f';
+    my $href = "Default/$uid.vcf";
+    my $card = <<EOF;
+BEGIN:VCARD
+VERSION:4.0
+UID;VALUE=TEXT:$uid
+PRODID:-//CyrusIMAP.org//Cyrus 3.13.0-alpha0-1560-g0834bc3ab-dirty//EN
+EMAIL;PROP-ID=e1;TYPE=WORK:jqpublic\@xyz.example.com
+EMAIL;PROP-ID=e3;PREF=2:jdoe\@example.com
+EMAIL;PROP-ID=e2;PREF=1:jane_doe\@example.com
+FN:Jane Doe
+END:VCARD
+EOF
+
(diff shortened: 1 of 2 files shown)


cassandane/tiny-tests/JMAPContacts/contact_get_emails_v4
```diff
@@ -0,0 +1,68 @@
...
+use Test::Deep ':v1';
+
+sub test_contact_get_emails_v4
+    :JMAPExtensions
+    ($self)
+{
+    my $jmap = $self->{jmap};
+
+    my $service = $self->{instance}->get_service("http");
+    $ENV{DEBUGDAV} = 1;
```
rsto: I'm pretty sure there is a `$self->{carddav}` on the test case instance

rsto (approved):
Found one nit in the cass test, otherwise LGTM

cassandane/tiny-tests/JMAPContacts/contact_get_emails_v4
ksmurchison: Yeah.  This test was copied from another and I didn't even pay attention to that.  Fixed.

imap/jmap_contact.c
```diff
@@ -1866,6 +1866,13 @@ static json_t *jmap_contact_from_vcard(const char *userid,
                     label = param->value;
                     if (label) label_len = strlen(label);
                 }
+                else if (!strcasecmp(param->name, "pref")) {
+                    int pref = atoi(param->value);
```
rsto: I think in this particular case we should deal more robustly with invalid PREF parameter values, as this could otherwise lead to people inadvertently sending emails using not the actual default email address? A valid PREF parameter value is in the range [1,100], but `atoi` returns 0 for invalid input. Also, `atoi` (which is `strtol` in disguise) accepts leading whitespace (probably OK) and an optional sign (not OK, or at least we should ignore negative PREF values).

rsto (changes requested):

imap/jmap_contact.c
ksmurchison: This would imply that we can't trust the vCard data that Cyrus writes via CardDAV PUT or ContactCard/set, correct?

rsto: Couldn't any CardDAV client do the same?

ksmurchison: I'm not sure what you're asking.  If a CardDAV client tries to create a vCard on Cyrus with PREF=foo or PREF=99999999999, that should be rejected by libicalvcard and/or carddav_put().

That said, I will switch out atoi() for some validation code.

rsto: That's my concern: I don't think that carddav_put does validate that and I'm also pretty sure that vparse accepts any string value that is a legit parameter value. I typically wouldn't be concerned about that, there's so much crap in vCard data already. But here we take a value which impacts business logic, so I think we should be more defensive.

ksmurchison: Done

rsto (approved):

PR #5716, opened 2025-12-04
Cyr 2446 - add api to cass for easy timsieved testing and add a test for #5713

wolfsage:
This provides a test case for https://github.com/cyrusimap/cyrus-imapd/pull/5713/files

### DIFF
--- a/cassandane/tiny-tests/Sieve/timsieved_utf8_scripts
+++ b/cassandane/tiny-tests/Sieve/timsieved_utf8_scripts
@@ -1,0 +1,26 @@
+#!perl
+use Cassandane::Tiny;
+use Encode qw(encode);
+
+sub test_timsieved_utf8_scripts
+    ($self)
+{
+    my $client = $self->timsieved_client;
+
+    timsieved_write($client, "LISTSCRIPTS\n");
+    $self->assert_str_equals("OK\r\n", timsieved_read($client));
+
+    my $snowman = encode("UTF-8", "\N{SNOWMAN}");
+    my $slen = length($snowman);
+
+    # Create a utf8 sieve script name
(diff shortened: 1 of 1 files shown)


ksmurchison (approved):
Looks great.  Thanks!

cassandane/Cassandane/Cyrus/Sieve.pm
```diff
@@ -398,6 +398,63 @@ EOF
...
+    my $sock = shift;
+
+    my $buf;
+
+    eval {
+        local $SIG{ALRM} = sub { die "alarm\n" };
+        alarm 5;
+        $sock->recv($buf, 4096);
+    };
+    alarm(0);
```
rjbs: I'm "sure" this is okay, but probably you should…

```perl
my $oldalarm = 0;
eval {
    ...
    $oldalarm = alarm 5;
    ....
};
alarm $oldalarm;
```

rjbs (approved):
I feel like it's rarely clear to me whether our various testing clients will return UTF-8 or text.  I imagine that in the future I will either say "oh it's 98% text" and we fix the remaining 2%, or I'll suggest a bigger change.

But everything in here is reasonable and "we can test" is a billion times better than "we can't".  Thanks for writing it, approved!

cassandane/Cassandane/Cyrus/Sieve.pm
wolfsage: Yeahhhhhhhhh I was lazy. Fine, will fix

wolfsage: Fixed

rjbs (approved):

PR #5718, opened 2025-12-05
cyrusdb sqlite fixes

elliefm:
This updates the unit tests for cyrusdb to also exercise the cyrusdb_sql backend when available (using the sqlite engine), and applies low-intrusion fixes to the sqlite engine wrapper to get those tests actually passing (or skipping).

There's more we could do here, but it wouldn't offer much real benefit, and would come at the risk of breaking the mysql or postgres wrappers, which aren't routinely or easily tested.

### DIFF
--- a/cunit/unit.c
+++ b/cunit/unit.c
@@ -79,6 +79,11 @@ int timeouts_flag = 1;
     else fprintf(stderr, "\nunit: "fmt"\n", (a1), (a2));                    \
 } while (0)
 #else
+/* no valgrind header? assume we're NOT running on valgrind. might
+ * lead to test failures if we are running on valgrind after all,
+ * but without the header there's no way to tell
+ */
+#define RUNNING_ON_VALGRIND (0)
 #define log1(fmt, a1) do {                                                  \
     fprintf(stderr, "\nunit: "fmt"\n", (a1));                               \
 } while (0)
@@ -478,6 +483,16 @@ int main(int argc, char **argv)
     CU_initialize_registry();
     register_cunit_suites();
     parse_args(argc, argv);
+
+    if (RUNNING_ON_VALGRIND) {
+        /* Tests that time out are killed without cleaning up properly, and
+         * this produces unhelpful valgrind noise.  This isn't helped by things
+         * generally running slower under valgrind to begin with.  So just
+         * don't enable timeouts when valgrind is detected.
+         */
+        timeouts_flag = 0;
+    }
+
     switch (mode) {
     case RUN:
         run_tests();


cunit/aaa-db.testc
```diff
@@ -19,7 +19,14 @@ struct binary_result
     size_t datalen;
 };
 
-static const char *backend = CUNIT_PARAM("skiplist,flat,twom,twoskip");
+#ifdef USE_CYRUSDB_SQL
+// #define BACKENDS "skiplist,flat,twom,twoskip,sql"
```
brong: does this work?  cunit.pl says:

```
   # Also scan for variable declarations of the form
   #
   # [static] [const] char * foo = CUNIT_PARAM("stringliteral");
   #
   # and make them a parameter for the containing test suite.
```

Or does cunit expect to parse those values after the preprocessor has run?

lib/cyrusdb_sql.c
```diff
@@ -649,8 +665,16 @@ static int fetch_cb(void *rock,
...
-        *(frock->data) = xrealloc(*(frock->data), datalen);
-        memcpy(*(frock->data), data, datalen);
+        if (datalen) {
+            *(frock->data) = xrealloc(*(frock->data), datalen);
+            memcpy(*(frock->data), data, datalen);
+        }
+        else {
+            /* datalen is 0: allocate a single byte and set it to "" */
+            *(frock->data) = xrealloc(*(frock->data), 1);
+            assert(data[0] == '\0');
```
brong: do we need this?  This whole block could just be `*frock->data = '\0';` or so.

cunit/aaa-db.testc
elliefm: It does!  cunit.pl scans the source file naively (no CPP), just doing regex matches, and keeping whatever it finds...

If I were to do the obvious thing and put `CUNIT_PARAM("foo")` and `CUNIT_PARAM("bar")` into the branches of the `#ifdef`, then cunit.pl would be very mad about CUNIT_PARAM being called twice.  So instead I use the branches to define a macro for the string literals I want, and invoke CUNIT_PARAM just once with that macro as its argument.  cunit.pl is happy, and outputs the argument as given wherever it needs it.  The preprocessor resolves it all later, when the generated source file is compiled.

lib/cyrusdb_sql.c
elliefm: Not sure, I haven't thought it through much, so far I've just been tinkering with a light touch until tests pass.

Here, we're trying to accept `("foo", 3)` (sensible value), `("", 0)` (empty string is allowed and distinct), or `(NULL, 0)` (no value), and handle each appropriately.  Not sure what we expect to do with `("foo", 0)` or `(NULL, 3)`.

brong: `(NULL, 3)` is an error.  `("foo", 0)` is totally legit, there's no requirement that you pass in a NULL terminated string, so the buffer could contain stuff from a previous use.

brong:
> I'm not convinced cyrusdb_sql is fit for purpose. Looks like we've extended our expectations of the cyrusdb API over the years, but not kept cyrusdb_sql up to date, except superficially. I guess that's what happens when something never gets tested! Right now it's not any better than "flat", at least with sqlite as it's underlying database. It might be different with mysql or postgress, I dunno, I haven't checked.
> 
> There's layers on layers here (cyrusdb_backend, sql_engine) so whatever we do to make sqlite work may have ramifications for mysql/postgres. We can only have _one_ database engine in any given cyrus build, so whichever one we choose for CI (i.e. sqlite, due to the cyrus-docker PR) will be the only one that's regularly tested. That's better than none being regularly tested, at least...

It's not supposed to be a GENERAL backend as such, it's mostly used as a read-only db, e.g. "user_deny".  At least that's my theory of why it was created.

lib/cyrusdb_sql.c
elliefm: I've cleaned up the details a bit, but we do need the realloc juggle.  Even though the previous allocation is definitely long enough for one byte, the first time through there won't be a previous allocation.

PR #5723, opened 2025-12-10
http_dblookup.c: fix uid2groups to work for ContactCard/set{group}

ksmurchison:
The uid2groups endpoint allows filtering group members by OtherUser. However, the ContactCard API doesn't distinguish between member cards owned by the user creating the group and members owned by other users. As result, the 'otheruser' column in the 'vcard_groups' DAV DB table is left empty.

This PR does the filtering by first fetching the member card and verifying that it appears in an addressbook accessible by the user.

### DIFF
--- a/imap/http_dblookup.c
+++ b/imap/http_dblookup.c
@@ -259,6 +259,10 @@ static int get_uid2groups(struct transaction_t *txn,
     const char **mailboxhdrs;
     const char *mailbox = "Default";
     mbentry_t *mbentry = NULL;
+    ptrarray_t *abook_sets = NULL;
+    struct auth_state *authstate = auth_newstate(userid);
+
+    if (!authstate) goto done;
 
     otheruserhdrs = spool_getheader(txn->req_hdrs, "OtherUser");
     if (otheruserhdrs) {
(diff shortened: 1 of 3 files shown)


imap/http_dblookup.c
```diff
@@ -276,12 +278,56 @@ static int get_uid2groups(struct transaction_t *txn,
...
+    /* Open DAV DB for the specified user.
+     * It will be assigned to an abook_set in jmap_get_accessible_addressbooks()
+     * and closed in jmap_free_abook_sets().
+     */
     db = carddav_open_userid(userid);
     if (!db) goto done;
 
-    array = carddav_getuid2groups(db, key, mbentry, otheruser);
-    if (!array) goto done;
+    if (!authstate) goto done;
```
wolfsage: If we somehow get here with no authstate, we've already opened the dav db, and will fail to close it in `done`. This check should be almost the first thing this method does, probably before checking otherUser header

ksmurchison: Good catch, I moved the authstate check during testing.  I'll move it back above opening DAV DB.

PR #5725, opened 2025-12-11
replace more of JMAPTalk in Cassandane

rjbs:
This completes the transition of Cassandane to JMAP::Tester, although many tests still use the API adapter layer.

### DIFF
--- a/cassandane/Cassandane/Unit/TestCase.pm
+++ b/cassandane/Cassandane/Unit/TestCase.pm
@@ -52,8 +52,6 @@ use DateTime::Format::ISO8601;
 use Cassandane::Util::Log;
 use Cassandane::Util::TestUrl;
 
-use Mail::JMAPTalk;
-
 my $enabled;
 my $buildinfo;
 
--- a/cassandane/doc/README.deps
+++ b/cassandane/doc/README.deps
@@ -16,7 +16,6 @@ Cassandane needs the following software to work:
 
     There is an Ubuntu package for this, libmail-imaptalk-perl.
 
- * Mail::JMAPTalk
  * Net::DAVTalk
  * Net::CalDAVTalk
  * Net::CardDAVTalk
@@ -171,7 +170,7 @@ Debian/Ubuntu copypasta:
        libdbd-sqlite3-perl libdigest-crc-perl libxml-simple-perl \
        libemail-mime-perl
 
-  sudo cpan Math::Int64 Mail::JMAPTalk Mail::IMAPTalk \
+  sudo cpan Math::Int64 Mail::IMAPTalk \
        Net::CalDAVTalk Net::CardDAVTalk String::CRC32 \
        IO::File::Lockable Archive::Tar::Stream Test::Unit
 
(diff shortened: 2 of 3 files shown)


cassandane/Cassandane/Instance.pm
```diff
@@ -3168,4 +3171,130 @@ sub make_folder_intermediate
     $self->start();
 }
 
+sub _common_http_service_args ($self) {
```
elliefm, line 3186: nit: our style has the opening `{`  on its own line for subs/functions/other top level thingies (then vim `[[` and `]]` can step through them), though our perl isn't as consistent about this as our C is

I realise this is code that was moved from elsewhere, and I clearly didn't notice this nit last time...

rjbs: I will create a to-discuss ticket about this and merge as is, for now.  (ellie and I discussed this approach.)

cassandane/Cassandane/Instance.pm
```diff
@@ -1164,6 +1167,18 @@ sub create_user
     return $user_obj;
 }
 
+sub create_trivial_user {
```
wolfsage, line 1170: This method is confusing to me. It says "Create" which to me implies "set up in cyrus" but I think it's only useful on already existing cyrus users, and is just getting us an object to play with?

rjbs: It's creating the TestUser object.  It's crappy, but I don't have a great alternative.  Basically, if you authenticate as "bob", you can make a bob mailbox, etc.  Users don't exist, once again.

So this is like "create user object but don't do underlying user setup".  We could call it `create_user_without_initialization` if you prefer.  This name should be longer than `create_user`, which we should want more things to use.

elliefm: The other way to think about it is that "users" exist in Cyrus as soon as they exist in the authentication database (external to Cyrus, queried via SASL).  If a user can authenticate, then they can usually create their own mailboxes (e.g. during IMAP client setup) due to the `implicit_owner_rights` imapd.conf setting, or with the "autocreate" feature Cyrus will create a configurable set of mailboxes on first login.  In this sense, setting up a "user" in Cyrus just means "adding them to the authentication database"; mailboxes, ACLs, etc may or may not come later.

Under Cassandane, the "authentication database" as such doesn't exist; `utils/fakesaslauthd` just accepts any username, with any password except "bad".   So creating a TestUser object is all the "setup in Cyrus" that's needed.

Thinking about it in these terms makes me think that method names like `create_user()` and `create_user_with_mailboxes_etc()` would make some sense, and not be misleading about the relationship between users and mailboxes.  But I think the latter is the one we'd mostly use, so it being the undifferentiated one makes ergonomic sense.

wolfsage: Okay...I suppose we can leave it rather than bike shed it. It does seem we use it 90% of the time where a user already exists or for the admin user, and if you try using it (at least for jmap) before a user exists, jmap requests all fail because the underlying user mailbox doesn't exist.

I'd almost call it "retrieve_user" or something, but we can keep it as is.

rjbs: I have already renamed it to `create_user_without_setup` in now-merged MR.  Foolishly, I did not ping you from that one.  (I considered and rejected "retrieve" because there is nothing on file to retrieve!  The metaphor is weird, because what we are creating is an access handle for data that will autovivify as needed, basically.)

I hope you like the new name okay!

wolfsage: Works for me, thanks!

PR #5727, opened 2025-12-15
consolidate cassandane 'other' configuration sections

elliefm:
This PR:

* consolidates the former `[cyrus replica]` and `[cyrus murder]` cassandane.ini sections into a single `[cyrus other]` section that's used for both.  This makes setting up cross-version testing a bit more ergonomic.
* fixes a few places where Cassandane's cross-version testing support had been broken by changes that didn't take it into account.

With these changes on master, and a subset backported to older branches, it is once again possible to use Cassandane to test replication and murder interactions between Cyrus versions.  (Not all tests pass like this, but the failures are not due to Cassandane bugs.)

### DIFF
--- a/cassandane/Cassandane/Config.pm
+++ b/cassandane/Cassandane/Config.pm
@@ -95,8 +95,6 @@ sub new
 
 sub default
 {
-    my $default_backend = $ENV{CASSANDANE_DEFAULT_DB} // 'twom';
-
     if (!defined($default)) {
         $default = Cassandane::Config->new(
             admins => 'admin mailproxy mupduser repluser',
@@ -120,24 +118,6 @@ sub default
             chatty => 'yes',
             debug => 'yes',
             httpprettytelemetry => 'yes',
-
-            # from cyr_info conf-default | grep _db:
-            annotation_db => $default_backend,
-            conversations_db => $default_backend,
-            duplicate_db => $default_backend,
-            mboxkey_db => $default_backend,
-            mboxlist_db => $default_backend,
-            ptscache_db => $default_backend,
-            quota_db => 'quotalegacy',
-            search_indexed_db => $default_backend,
-            seenstate_db => $default_backend,
-            subscription_db => 'flat',
-            statuscache_db => $default_backend,
-            sync_cache_db => $default_backend,
-            tlscache_db => 'twoskip', # deprecated, does not allow twom
-            tls_sessions_db => $default_backend,
-            userdeny_db => 'flat',
-            zoneinfo_db => $default_backend,
 
             # smtpclient_open should fail by default!
             #
--- a/cassandane/Cassandane/Config.pm
+++ b/cassandane/Cassandane/Config.pm
@@ -197,6 +197,24 @@ sub set
     }
 }
 
+sub set_if_undef
+{
+    my ($self, %nv) = @_;
+
+    while (my ($n, $v) = each %nv) {
+        if (exists $bitfields{$n}) {
+            # XXX bitfield behaviour?
+            die "can't set_if_undef with bitfield '$n'";
+        }
+        elsif (not defined $self->get($n)) {
+            $self->{params}->{$n} = $v;
+        }
+        else {
+            # nothing to do
+        }
+    }
+}
+
 sub set_bits
 {
     my ($self, $name, @bits) = @_;
--- a/cassandane/Cassandane/Cyrus/CassMeta.pm
+++ b/cassandane/Cassandane/Cyrus/CassMeta.pm
@@ -1,0 +1,135 @@
+#!/usr/bin/perl
+# Cassandane::Cyrus::CassMeta: Cassandane meta-tests that need Cyrus
+# (as distinct from Cassandane::Test::*, which don't need Cyrus)
+# SPDX-License-Identifier: BSD-3-Clause-CMU
+# See COPYING file at the root of the distribution for more details.
+
+package Cassandane::Cyrus::CassMeta;
+use strict;
+use warnings;
+use Data::Dumper;
+
+use base qw(Cassandane::Cyrus::TestCase);
+use Cassandane::Util::Log;
+
+# See Cassandane::Cyrus::TestCase::_create_instances()
+my @all_instance_names = qw(instance replica frontend backend2);
+
+sub new
+{
+    my $class = shift;
+
+    my $config = Cassandane::Config->default()->clone();
+
+    # override ONE cyrusdb backend to a value that is:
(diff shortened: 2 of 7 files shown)


cassandane/Cassandane/Instance.pm
```diff
@@ -160,14 +160,56 @@ sub new
...
+                $default_backend = $b;
+                last;
+            }
+        }
+
+        # every Cyrus version supports twoskip so this won't happen..
+        die "couldn't find a supported cyrusdb backend"
+            if not $default_backend;
+
+        $self->{config}->set(
```
wolfsage: Up on line 143:
```
    $self->{config} = $params{config}->clone()
        if defined $params{config};
```

Config comes in from Cassandane::Cyrus::TestCase::_create_instances().

If a test sets one of the parameters below in the `sub new` setup like so:
```
    my $config = Cassandane::Config->default()->clone();
    $config->set(annotation_db => 'twoskip');

    my $self = $class->SUPER::new({
        config => $config,
        jmap => 1,
        adminstore => 1,
        services => [ 'imap', 'http' ]
    }, @args);

```

the test setting for annotation_db gets overridden below!

wolfsage (dismissed):

cassandane/Cassandane/Instance.pm
wolfsage: (This implies we should have a sanity test that does the above and then confirms the resulting config is correct...)

elliefm: Ahh yeah, good catch!

elliefm: We have meta-tests for Cassandane::Config in Cassandane::Test::Config, but being in the Test namespace, it doesn't spin up any cyrus instances, so can't examine instance config.  Pity, cause that would be a useful sanity check to have.  Perhaps we could add a Cassandane::Cyrus::Cassandane suite as a place to stash cassandane meta-tests that need a running cyrus.

I've added two new commits that should stop instance construction from overriding database backends set by the test.  One adds a utility function needed by the fix, the other adds the fix.  I expect I'll squash the fix later, but I've left it as its own commit for now for ease of iterative review.

cassandane/Cassandane/Config.pm
```diff
@@ -197,6 +177,21 @@ sub set
...
+{
+    my ($self, %nv) = @_;
+
+    while (my ($n, $v) = each %nv) {
+        if (exists $bitfields{$n}) {
+            # XXX bitfield behaviour?
+            die "can't set_if_undef with bitfield '$n'";
+        }
+        else {
+            $self->{params}->{$n} //= $v;
```
wolfsage: This doesn't have the desired effect - apparently config is a tree of child/parent configs. See `sub get` --  if *we* don't have the necessary config, it looks up at our parent's config to find that value -- where as set_if_undef only checks the current level.

This bit here:

```
    $self->{config} = $params{config}->clone()
        if defined $params{config};

```

... turns out clone() does:

```
sub clone
{
    my ($self) = @_;

    my $child = Cassandane::Config->new();
    $child->{parent} = $self;
```

which sets that parent relationship.

Here's a test case you can use to make sure your fix works:
```
diff --git a/cassandane/Cassandane/Cyrus/JMAPContacts.pm b/cassandane/Cassandane/Cyrus/JMAPContacts.pm
index b3842ea6a7..393aa30fbc 100644
--- a/cassandane/Cassandane/Cyrus/JMAPContacts.pm
+++ b/cassandane/Cassandane/Cyrus/JMAPContacts.pm
@@ -71,6 +71,8 @@ sub new
                  vcard_max_size => 100000,
                  jmap_nonstandard_extensions => 'yes');
 
+    $config->set(sync_cache_db => 'flat');
+
     my $self = $class->SUPER::new({
         config => $config,
         jmap => 1,

```

And then test `JMAPContacts.card_set_update`

flat isn't valid for sync_cache_db, so the test should fail if the config is set properly.

wolfsage (dismissed):
It's still not working right. I've included how you can manually test it lightly, but at this point I think a real test would be valuable. Any tiny test would really do - it could set each config param to something different then the default then inspect the config after cyrus starts up or something....

cassandane/Cassandane/Config.pm
elliefm: Oh yeah, I knew that, and had planned to implement set_if_undef in terms of the getter and setter to avoid duplicating all the tree logic.  But then I got distracted thinking about what to do with bitfields, forgot I had Plans, and just used `//=` like a fool.  D'oh!

But this is pretty solid evidence supporting "this needs a proper test".  Will do that too.

cassandane/Cassandane/Cyrus/CassMeta.pm
```diff
@@ -0,0 +1,114 @@
...
+
+    my %irregular_databases = (
+        quota_db => 'quotalegacy',
+        subscription_db => 'skiplist', # usually 'flat', but we overrode it!
+        tlscache_db => 'twoskip',
+        userdeny_db => 'flat',
+    );
+
+    foreach my $instance_name (@all_instance_names) {
+        my $instance = $self->{$instance_name} || die;
```
wolfsage: die should always have a string so if it happens we know what the problem is.

wolfsage (approved):
Cool

wolfsage (dismissed):
Just fix up the lone `die;` perhaps?

cassandane/Cassandane/Cyrus/CassMeta.pm
elliefm: I don't disagree, but I don't know what would cause this to happen in practice... maybe if someone removes the "murderimap" or "replication" flags from the constructor, so that we're running without all the instances we expect initialised, or we add new kinds of instances and haven't updated the constructor to request them, or we add junk to `@all_instance_names`, or we rename the instances upstream and forget to rename them here?  But like, I don't think it's possible for it to happen except by someone breaking something, and it's self evident from the line of code that we expected to find an instance with a particular name and didn't.  Whoever hypothetically breaks it will know what they just did better than I could predict.

I could have written `$self->assert_non_null($instance);`, but that would imply that the test has found something wrong with the system under test.  In this case, that didn't happen; the programmer broke the test.

If you have a suggestion for a die message that would actually be useful, I'm all ears, but anything I write right now would just be noise for the sake of writing something.

elliefm: I slept on it and decided knowing the name of the instance we failed to find would be useful, so there's now a die message reporting that.

wolfsage (approved):

PR #5730, opened 2025-12-17
configure option to set zoneinfo_dir at compile time

elliefm:
Adds a configure option to set a default zoneinfo_dir at compile time.  The zoneinfo_dir imapd.conf value always takes precedence over this.

This will mostly be useful for cross-version Cassandane testing, where different Cyrus versions are probably linked against different library versions with different timezone data, and the easiest place to specify the right path for each build is during the builds.


rsto (approved):
Nice

ksmurchison (approved):

imap/ical_support.c
```diff
@@ -76,12 +76,12 @@ EXPORTED void ical_support_init(void)
...
-    const char *tzpath = config_getstring(IMAPOPT_ZONEINFO_DIR);
     icalarray *timezones;
 
-    if (tzpath) {
-        syslog(LOG_DEBUG, "using timezone data from zoneinfo_dir=%s", tzpath);
-        icaltimezone_set_zone_directory((char *) tzpath);
+    if (config_zoneinfo_dir) {
+        syslog(LOG_DEBUG, "using timezone data from zoneinfo_dir=%s",
+                          config_zoneinfo_dir);
+        icaltimezone_set_zone_directory(config_zoneinfo_dir);
```
wolfsage, line 84: Huh, why don't we need to cast this anymore?

wolfsage (approved):
lgtm - though I'm curious why having two different cassandane.inis isn't an option? Is this because a single cassandane might now start two different installs of cyrus? I suppose so, and having it adjust configs on the fly is maybe too much of a pain?

imap/ical_support.c
elliefm: That cast was incredibly sketchy, but it being there makes me think libical was previously not const-correct, and the sketchy cast was to shut up a warning.  Must have been fixed in libical at some point, since it now allows the const argument without us needing to lie that it's modifiable.

PR #5731, opened 2025-12-17
http_carddav.c: handle no-uid-conflict per RFC 6352, Section 6.3.2.1

ksmurchison:
- Prevent the UID of an existing resource to be changed
- Prevent two resources in an addressbook from having same UID

### DIFF
--- a/cassandane/tiny-tests/Carddav/uid_conflict
+++ b/cassandane/tiny-tests/Carddav/uid_conflict
@@ -1,0 +1,174 @@
+#!perl
+use Cassandane::Tiny;
+
+my sub put_vcard ($carddav, $args)
+{
+    my ( undef, undef, $line ) = caller;
+    my $dt = DateTime->now();
+    $dt->set_time_zone('Etc/UTC');
+    my $rev = $dt->strftime('%Y%m%dT%H%M%SZ');
+
+    my $vcard = <<EOF;
+BEGIN:VCARD
+PRODID:-//foo//bar//EN
+VERSION:3.0
+UID:$args->{uid}
+N:;;;;
+FN:test created in line $line
+REV:$rev
+END:VCARD
+EOF
+    $vcard =~ s/\r?\n/\r\n/gs;
+
+    my $method = 'PUT';
+    my $url = $carddav->request_url($args->{resource});
+    my $res = $carddav->ua->request($method, $url,
+        {
+            content => $vcard,
+            headers => {
+                'Content-Type' => "text/vcard; version=3.0",
+                Authorization => $carddav->auth_header(),
+            }
+        }
+    );
+
+    if ($ENV{DEBUGDAV}) {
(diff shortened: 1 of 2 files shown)


cassandane/tiny-tests/Carddav/uid_conflict
```diff
@@ -0,0 +1,140 @@
...
+EOF
+    $vcard =~ s/\r?\n/\r\n/gs;
+
+    my $res = $carddav->ua->request('PUT',
+        $carddav->request_url($args->{resource}),
+        {
+            content => $vcard,
+            headers => {
+              'Content-Type' => "text/vcard; version=3.0",
+              Authorization => $carddav->auth_header(),
```
wolfsage: nit: 4 space indents here

cassandane/tiny-tests/Carddav/uid_conflict
```diff
@@ -0,0 +1,140 @@
...
+    Carp::croak("method must be 'MOVE' or 'COPY'")
+        unless $method eq 'MOVE' || $method eq 'COPY';
+
+    my $res = $carddav->ua->request($method,
+        $carddav->request_url($args->{from}),
+        {
+            content => undef,
+            headers => {
+              Authorization => $carddav->auth_header(),
+              Destination => $carddav->request_url($args->{to}),
```
wolfsage: nit: 4 space indents here

cassandane/tiny-tests/Carddav/uid_conflict
ksmurchison: Fixed

cassandane/tiny-tests/Carddav/uid_conflict
ksmurchison: Fixed

rsto:
> I think the real, better fix, would be to adjust Net::*DavTalks APIs to allow for returning the results on failure rather than throwing an exception.

Very much this. I'm missing this feature since years.

cassandane/tiny-tests/Carddav/uid_conflict
```diff
@@ -0,0 +1,174 @@
...
+        {
+            content => $vcard,
+            headers => {
+                'Content-Type' => "text/vcard; version=3.0",
+                Authorization => $carddav->auth_header(),
+            }
+        }
+    );
+
+    if ($ENV{DEBUGDAV}) {
```
rsto, line 35: Is this necessary if you instead set `$ENV{DEBUGDAV} = 1;` at the start of test?

ksmurchison: No, because $carddav->ua->request doesn't log at all.  Only $carddav->Request logs, and I copied what it does into this test.  `$ENV{DEBUGDAV} = 1; `is already set in Carddav.pm, which is why my added logging works.

If someone wants to add/change/remove the logging here, feel free.

wolfsage (approved):

PR #5732, opened 2025-12-18
http_dav.c: set deletedmodseq if a URL gets recreated

ksmurchison:
If we create a resource with same URL as a destroyed one, we can't calculate changes before the original was destroyed because we only allow one DAV DB record per mailbox+URL tuple and the old record with the old modseq gets overwritten

### DIFF
--- a/cassandane/tiny-tests/JMAPContacts/card_changes_recreate_url
+++ b/cassandane/tiny-tests/JMAPContacts/card_changes_recreate_url
@@ -1,0 +1,70 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_card_changes_recreate_url
+    :needs_dependency_icalvcard
+    ($self)
+{
+    $ENV{DEBUGDAV} = 1;
+    my $carddav = $self->{carddav};
+    my $jmap = $self->{jmap};
+
+    my $href = "Default/test.vcf";
+
+    my $uid1 = 'ae2640cc-234a-4dd9-95cc-3106258445b9';
+    my $card1 = <<EOF;
+BEGIN:VCARD
+VERSION:3.0
+UID:$uid1
+N:Gump;Forrest;;Mr.
+FN:Forrest Gump
+END:VCARD
+EOF
+
+    xlog $self, "create card";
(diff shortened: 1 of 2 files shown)


cassandane/tiny-tests/JMAPContacts/card_changes_recreate_url
```diff
@@ -0,0 +1,70 @@
...
+EOF
+
+    xlog $self, "create card";
+    $card1 =~ s/\r?\n/\r\n/gs;
+    $carddav->Request('PUT', $href, $card1, 'Content-Type' => 'text/vcard');
+
+    xlog $self, "get card and current state";
+    my $res = $jmap->CallMethods([['ContactCard/get', { }, 'R1']]);
+    my $id = $res->[0][1]{list}[0]{id};
+    my $state = $res->[0][1]{'state'};
```
rsto: nit: this might be named `startState` for contrast to `newState`

imap/http_dav.c
```diff
@@ -7335,6 +7336,33 @@ int meth_put(struct transaction_t *txn, void *params)
...
+                break;
+            case MBTYPE_ADDRESSBOOK:
+                deletedmodseq = &counters.carddavdeletedmodseq;
+                break;
+            }
+
+            if (deletedmodseq) {
+                const char *mboxname = mailbox_name(mailbox);
+
+                assert(!mboxname_read_counters(mboxname, &counters));
```
rsto, line 7359: Rather than crashing, this could also more gracefully log an error and return `500`? The effect to the client might be the same but we give the process a change to clean up state.
That being said I haven't checked how we handle failure on `mboxname_read_counters` at other places, I agree it's a fatal error in any case.

cassandane/tiny-tests/JMAPContacts/card_changes_recreate_url
ksmurchison: Fixed

imap/http_dav.c
ksmurchison: As discussed internally, we are going to leave the assert()

PR #5733, opened 2025-12-18
cyrdbbench: use the provided options argument

elliefm:
The argument was previously marked as unused to shush a compiler warning, but the better fix is to use it.

It came up because 3.12 suffers the same warning, but the commit that shushed the warning on master isn't suitable for backporting.  This PR gets us a better fix, but mostly gets me something I can backport to 3.12.

### DIFF
--- a/bench/cyrdbbench.c
+++ b/bench/cyrdbbench.c
@@ -368,13 +368,14 @@ static size_t do_write(int txnmode, int insmode)
     return bytes;
 }
 
-static int parse_options(int argc, char **argv, const struct option *options __attribute__((unused)))
+static int parse_options(int argc, char **argv, const struct option *options)
 {
     int option;
     int option_index;
 
     while ((option = getopt_long(argc, argv, "d:b:n:t:h?",
-                                 long_options, &option_index)) != -1) {
+                                 options, &option_index)) != -1)
+    {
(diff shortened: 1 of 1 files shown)


rsto (approved):

elliefm:
See, the reason I don't think more detail is necessary in _this_ commit message is _because git blame exists_.

If you saw this commit and, for some reason, wanted to know what its deal was, you would git blame one commit earlier to find the previous commit(s) that changed these lines, and look at those commits.  You would quickly see that an `__attribute__((unused))` was added to shut up a compiler warning as part of other unrelated changes, and then my commit, which surgically addresses the same compiler warning but better, would make perfect sense (if it somehow didn't already).

This is basically the same process I used to find out what had changed when trying to figure out why master compiled this file without warnings but 3.12 didn't.  If you had been dealing with this issue yourself, you would've probably done all the same things without even thinking too hard about it, and would similarly feel like it doesn't warrant a longwinded explanation.

If this commit was doing something complicated or interesting, sure, a good commit message should contain details.  But part of good written communication is keeping the word count proportional to the significance.  This commit is just not significant.

This is not a case where I banged out a quick commit message without thinking about it.  Are you mistaking "few words" for "no thought"?

PR #5734, opened 2025-12-19
ContactCard/set{update}: allow patching addressbookIds

ksmurchison:
and enforce maxAddressBooksPerCard:1

### DIFF
--- a/cassandane/tiny-tests/JMAPContacts/card_set_addressbookids
+++ b/cassandane/tiny-tests/JMAPContacts/card_set_addressbookids
@@ -1,0 +1,244 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_card_set_addressbookids
+    :needs_dependency_icalvcard
+{
+    my ($self) = @_;
+    my $jmap = $self->{jmap};
+
+    xlog $self, "get default addressbook";
+    my $res = $jmap->CallMethods([
+            ['AddressBook/get', { }, "R1"]
+    ]);
+    my $abookid1 = $res->[0][1]{list}[0]{id};
+    $self->assert_not_null($abookid1);
+
+    xlog $self, "create second addressbook";
+    $res = $jmap->CallMethods([
+        ['AddressBook/set', {
+            create => {
+                "1" => {
+                    name => "foo"
+                }
+            }
+         }, "R1"]
+    ]);
+    my $abookid2 = $res->[0][1]{created}{"1"}{id};
+    $self->assert_not_null($abookid2);
+
+    xlog $self, "attempt to create a card in NULL addressbooks";
+    $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            create => {
+                "1" => {
+                    '@type' => 'Card',
+                    version => '1.0',
+                    addressBookIds => JSON::null
+                }
+            }
+        }, 'R1']
+    ]);
+    $self->assert_cmp_deeply(
+        superhashof({
+            notCreated => {
+                1 => {
+                    type => 'invalidProperties',
+                    properties => [ 'addressBookIds' ]
+                }
+            }
+        }),
+        $res->[0][1]
+    );
+
+    xlog $self, "attempt to create a card in zero addressbooks";
+    $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            create => {
+                "1" => {
+                    '@type' => 'Card',
+                    version => '1.0',
+                    addressBookIds => { }
+                }
+            }
+        }, 'R1']
+    ]);
+    $self->assert_cmp_deeply(
+        superhashof({
+            notCreated => {
+                1 => {
+                    type => 'invalidProperties',
+                    properties => [ 'addressBookIds' ]
+                }
+            }
+        }),
+        $res->[0][1]
+    );
+
+    xlog $self, "attempt to create a card in multiple addressbooks";
+    $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            create => {
+                "1" => {
+                    '@type' => 'Card',
+                    version => '1.0',
+                    addressBookIds => {
+                        $abookid1 => JSON::true,
+                        $abookid2 => JSON::true
+                    }
+                }
+            }
+        }, 'R1']
+    ]);
+    $self->assert_cmp_deeply(
+        superhashof({
+            notCreated => {
+                1 => {
+                    type => 'invalidProperties',
+                    properties => [ 'addressBookIds' ]
+                }
+            }
+        }),
+        $res->[0][1]
+    );
+
+    xlog $self, "create a card in a single (default) addressbook";
+    $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            create => {
+                "1" => {
+                    '@type' => 'Card',
+                    version => '1.0'
+                }
+            }
+        }, 'R1']
+    ]);
+    my $id = $res->[0][1]{created}{1}{id};
+    $self->assert_not_null($id);
+
+    xlog $self, "attempt to update a card to be in NULL addressbooks";
+    $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            update => {
+                $id => {
+                    addressBookIds => JSON::null
+                }
+            }
+         }, "R1"]
+    ]);
+    $self->assert_cmp_deeply(
+        superhashof({
+            notUpdated => {
+                $id => {
+                    type => 'invalidProperties',
+                    properties => [ 'addressBookIds' ]
+                }
+            }
+        }),
+        $res->[0][1]
+    );
+
+    xlog $self, "attempt to update a card to be in zero addressbooks";
+    $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            update => {
+                $id => {
+                    addressBookIds => { }
+                }
+            }
+         }, "R1"]
+    ]);
+    $self->assert_cmp_deeply(
+        superhashof({
+            notUpdated => {
+                $id => {
+                    type => 'invalidProperties',
+                    properties => [ 'addressBookIds' ]
+                }
+            }
+        }),
+        $res->[0][1]
+    );
+
+    xlog $self, "attempt to update a card to be in multiple addressbooks";
+    $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            update => {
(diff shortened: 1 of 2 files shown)


cassandane/tiny-tests/JMAPContacts/card_set_multiple_addressbookids
```diff
@@ -0,0 +1,122 @@
...
+    my $res = $jmap->CallMethods([
+            ['AddressBook/get', { }, "R1"]
+    ]);
+    my $abookid1 = $res->[0][1]{list}[0]{id};
+
+    xlog $self, "create addressbook";
+    $res = $jmap->CallMethods([
+            ['AddressBook/set', { create => { "1" => {
+                            name => "foo",
+                            description => "My foo",
```
rsto: The `description` property is not required for this test.

cassandane/tiny-tests/JMAPContacts/card_set_multiple_addressbookids
```diff
@@ -0,0 +1,122 @@
...
+             }}}, "R1"]
+    ]);
+    my $abookid2 = $res->[0][1]{created}{"1"}{id};
+
+    my $prodid = '-//Example Corp.//CardDAV Client//EN';
+    my $name = 'Mr. John Q. Public, Esq.';
+    my $uid = 'urn:uuid:ae2640cc-234a-4dd9-95cc-3106258445b9';
```
rsto: None of these property values later are required in asserts in this test, there is no need to define them here.

cassandane/tiny-tests/JMAPContacts/card_set_addressbookids
```diff
@@ -0,0 +1,122 @@
...
+    my $prodid = '-//Example Corp.//CardDAV Client//EN';
+    my $name = 'Mr. John Q. Public, Esq.';
+    my $uid = 'urn:uuid:ae2640cc-234a-4dd9-95cc-3106258445b9';
+
+    xlog $self, "attempt to create a card in multiple addressbooks";
+    $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            create => {
+                "1" => {
+                    '@type' => 'Card',
```
rsto, line 83: This card contains multiple properties such as `propId`, `uid`, etc. that are irrelevant to this test. Using a minimal Card would allow the reader to focus on the essence of this test.

cassandane/tiny-tests/JMAPContacts/card_set_addressbookids
```diff
@@ -0,0 +1,122 @@
...
+    $self->assert_str_equals('invalidProperties',
+                                $res->[0][1]{notCreated}{1}{type});
+    $self->assert_str_equals('addressBookIds',
+                             $res->[0][1]{notCreated}{1}{properties}[0]);
+
+    xlog $self, "create a card in a single addressbook";
+    $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            create => {
+                "1" => {
```
rsto, line 109: Again: the Card contains properties that are irrelevant for the test.

cassandane/tiny-tests/JMAPContacts/card_set_multiple_addressbookids
```diff
@@ -0,0 +1,122 @@
...
+                                $res->[0][1]{notUpdated}{$id}{type});
+    $self->assert_str_equals('addressBookIds',
+                             $res->[0][1]{notUpdated}{$id}{properties}[0]);
+
+    xlog $self, "patch a card, moving it between addressbooks";
+    $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            update => {
+                $id => {
+                    "addressBookIds/$abookid1" => JSON::false,
```
rsto: This is wrong, `false` is not a valid value in neither the patch object nor the actual data structure. The only allowed value in a JMAP set is `true`. To remove an addressbook id one must set `null` (=Perl `undef`) in the patch object.

imap/jmap_contact.c
```diff
@@ -12409,25 +12351,101 @@ static int _card_set_update(jmap_req_t *req, bool apply_empty_updates,
...
         json_decref(old_obj);
         if (!new_obj) {
             r = HTTP_BAD_REQUEST;
             goto done;
         }
 
+        /* Validate addressBookIds */
+        const char *addressbookId = NULL;
+        json_t *jval = json_object_get(new_obj, "addressBookIds");
+        if (jval) {
```
rsto, line 12395: This does not seem to handle the erroneous case of a client deleting the `addressBooksIds` property entirely, e.g.
```
{ "addressBookIds": null }
```

Now that `old_obj` gets the original addressbook id set, the absence of `addressBookIds` here can only be the case if the client had deleted the whole property.

imap/jmap_contact.c
```diff
@@ -12409,25 +12351,101 @@ static int _card_set_update(jmap_req_t *req, bool apply_empty_updates,
...
+            void *iter = json_object_iter(jval);
+            while (iter) {
```
rsto: I don't understand the need to iterate here in contrast to the former code: the only valid value for `addressBookIds` still is a JSON object that has a single key and a value `true`.

Any `null`ed original addressbook id already was handled by applying the patch object earlier and any patch that attempted to patch an addressbookid that was not present in the old (also single-valued) `addressbookIds` should have been reported in the `invalid` array in `jmap_patchobject_apply`?

cassandane/tiny-tests/JMAPContacts/card_set_multiple_addressbookids
ksmurchison: Removed

cassandane/tiny-tests/JMAPContacts/card_set_multiple_addressbookids
ksmurchison: Removed

cassandane/tiny-tests/JMAPContacts/card_set_addressbookids
ksmurchison: Removed

cassandane/tiny-tests/JMAPContacts/card_set_multiple_addressbookids
ksmurchison: Yeah.  Brain fart.  Fixed

imap/jmap_contact.c
ksmurchison: This was a result of the earlier brain fart.  Fixed.

cassandane/tiny-tests/JMAPContacts/card_set_addressbookids
ksmurchison: Removed

imap/jmap_contact.c
ksmurchison: Added a test for this

cassandane/tiny-tests/JMAPContacts/card_set_addressbookids
```diff
@@ -0,0 +1,160 @@
...
+    :needs_dependency_icalvcard :MailboxLegacyDirs
+{
+    my ($self) = @_;
+    my $jmap = $self->{jmap};
+
+    xlog $self, "get default addressbook";
+    my $res = $jmap->CallMethods([
+            ['AddressBook/get', { }, "R1"]
+    ]);
+    my $abookid1 = $res->[0][1]{list}[0]{id};
```
wolfsage, line 14: We should assert here (and just below) that we get a value in $abookid1 and $abookid2

ksmurchison: Done

imap/jmap_contact.c
rsto: I can't find it.

imap/jmap_contact.c
```diff
@@ -12278,70 +12280,12 @@ static int _card_set_update(jmap_req_t *req, bool apply_empty_updates,
     cdata->vcard_uid = uid = xstrdup(cdata->vcard_uid);
     cdata->dav.resource = resource = xstrdup(cdata->dav.resource);
 
-    json_t *jkind = json_object_get(jcard, "kind");
```
rsto, line 12281: I don't understand why this needs to get moved to a different code location. I guess because the logic for the `num_props` conditional is preserved, but why is that `num_props` conditional here in the first place? It looks as if it's meant to provide a fast-path for patches that only set the `importance` property? If that's the case, why not just test the patch object for containing that and only that property?

cassandane/tiny-tests/JMAPContacts/card_set_multiple_addressbookids
```diff
@@ -0,0 +1,198 @@
...
+        ['AddressBook/set', {
+            create => {
+                "1" => {
+                    name => "foo"
+                }
+            }
+         }, "R1"]
+    ]);
+    my $abookid2 = $res->[0][1]{created}{"1"}{id};
+    $self->assert_not_null($abookid1);
```
wolfsage: Copy pasta - should be `$abookid2` here

ksmurchison: Thanks.  Fixed

imap/jmap_contact.c
ksmurchison: This was moved as a part of WIP and I never moved it back.  Now fixed.

imap/jmap_contact.c
rsto: I do not think this is resolved. I currently see the following cases tested for `set{update}`:

```
addressBookIds => { }
```
```
addressBookIds => {
  $abookid1 => JSON::true,
  $abookid2 => JSON::true
}
```
```
"addressBookIds/$abookid1" => JSON::null
```
```
addressBookIds/$abookid2" => JSON::true
```

```
{
"addressBookIds/$abookid1" => JSON::null,
"addressBookIds/$abookid2" => JSON::true
}
```

But my comment is about

```
addressBookIds => JSON::null
```
My main concern isn't even about it not being tested, but instead it not being handled correctly as an erroneous request that should get rejected.

ksmurchison: Add a test and fixed the addressBookId validation to correctly fail.

wolfsage (approved):

PR #5737, opened 2025-12-23
Enforce ContactCard mandatory properties

ksmurchison:
- Commits 1-3 deal with mandatory properties
- Commit 4 is a minor fix/optimization for parsing property names/paths
- Commit 5 consolidates /get property validation rather than having duplicate code

The perfect hash code referenced in comments below has been moved to 
https://github.com/ksmurchison/cyrus-imapd/pull/12

### DIFF
--- a/imap/jmap_api.c
+++ b/imap/jmap_api.c
@@ -1524,6 +1524,39 @@ HIDDEN const jmap_property_t *jmap_property_find(const char *name,
 
 /* Foo/get */
 
+HIDDEN hash_table *jmap_get_validate_props(jmap_req_t *req,
+                                           struct jmap_parser *parser,
+                                           const jmap_property_t *valid_props,
+                                           const char *key,
+                                           json_t *jprops)
+{
+    hash_table *props = xzmalloc(sizeof(hash_table));
+    construct_hash_table(props, json_array_size(jprops) + 1, 0);
+
+    json_t *val;
+    size_t i;
+    json_array_foreach(jprops, i, val) {
+        const char *name = json_string_value(val);
+        const jmap_property_t *propdef = NULL;
+        if (name) {
+            propdef = jmap_property_find(name, valid_props);
+            if (propdef && propdef->capability &&
+                !jmap_is_using(req, propdef->capability)) {
+                propdef = NULL;
+            }
+        }
+        if (!propdef || (propdef->flags & JMAP_PROP_REJECT_GET)) {
+            jmap_parser_push_index(parser, key, i, name);
+            jmap_parser_invalid(parser, NULL);
+            jmap_parser_pop(parser);
+            continue;
+        }
+        hash_insert(name, (void*)1, props);
+    }
+
+    return props;
+}
+                           
 HIDDEN void jmap_get_parse(jmap_req_t *req,
                            struct jmap_parser *parser,
                            const jmap_property_t valid_props[],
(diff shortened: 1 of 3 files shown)


cassandane/tiny-tests/JMAPContacts/cardgroup_changes
```diff
@@ -7,14 +7,30 @@ sub test_cardgroup_changes
...
+    my $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            create => {
+                "a" => {
+                    '@type' => 'Card',
+                    version => '1.0',
+                    firstName => "a",
+                    lastName => "a"
+                },
+                "b" => {
```
wolfsage, line 20: These ones also need the `@type` and `version` keys, we should probably assert that they are all successfully created too, since they are later added as members below here where we grab $contacta / $contactb / $contactc / $contactd...

Later when we update groups, we're passing `"members": { "": true }` ... I'm surprised that works!

imap/jmap_api.c
```diff
@@ -1546,28 +1546,113 @@ HIDDEN int jmap_parse_strings(json_t *arg,
...
-                return prop;
+    const jmap_property_t *prop = prop_set->table->lookup(name, strlen(name));
+
+    if (!prop) {
+        for (int i = 0; i < strarray_size(&prop_set->wildcards); i++) {
+            const char *wild = strarray_nth(&prop_set->wildcards, i);
+            size_t len = strlen(wild);
+
+            if (!strncmp(name, wild, len-1)) {
+                prop = prop_set->table->lookup(wild, strlen(wild));
```
wolfsage: nit: Can use len here instead of strlen(wild)

wolfsage (commented):
(Some starting comments, more later)

cassandane/tiny-tests/JMAPContacts/cardgroup_changes
ksmurchison: Fixed

imap/jmap_api.c
ksmurchison: Yep. Fixed

wolfsage (approved):
Seems good, though I wonder about the cost of initializing all of those tables at startup. Do they all initialize every time any httpd starts up?

rsto (commented):
No objections to enforcing mandatory properties and consolidating the `Foo/get` property argument parsing code.

For switching the logic to perfect hashes, I do have a couple of questions/remarks though (not necessarily objections, I am asking for clarity):
- Did the linear search code show up as a performance issue?
- The code seems to combine hash lookups with linear search for special properties, namely wildcards, always_get, mandatory, external. If that reading is correct, the performance gain over the former all-linear search code depends on the number of "normal" properties being substantially higher than the ones that require linear search for each lookup, correct?
- As an alternative to switching to compile-time perfect hashes, could we instead of using linear search use binary search while keeping the former properties declarations? That would require either us developers to keep the property list sorted by name, or that sorting could be done in httpd during startup?
- If we use the gperf-generated code, wouldn't we want to keep the generated code source out of version control?

rsto (approved):
LGTM now that the perfect hash code got moved to another PR.

@ksmurchison I suggest you merge this PR, then rebase https://github.com/ksmurchison/cyrus-imapd/pull/12 on the main branch rather than this feature branch. Please add me as reviewer then.

PR #5738, opened 2025-12-24
configure.ac: add Cyrus::Backup modules

rjbs:
Without these, we don't package these modules, and "make distcheck" will fail (and an installed-from-source Cyrus will not include these).


elliefm:
Huh, I wonder why this was never a problem for me...

elliefm (approved):

PR #5739, opened 2025-12-25
replace most files' copy of license with a pointer to central license

rjbs:
This is probably okay to merge, but we need to discuss a few things, like the files that say ©️ Fastmail but then include CMU license text, or files with multiple copyright attributions.

### DIFF
--- a/bench/cyrdbbench.c
+++ b/bench/cyrdbbench.c
@@ -1,45 +1,6 @@
-/* cyrusdb_bench.c: cyrusdb benchmark tool.
- *                  [Based on the db_bench tool in LevelDB]
- *
- * Copyright (c) 1994-2019 Carnegie Mellon University.  All rights reserved.
- *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions
- * are met:
- *
- * 1. Redistributions of source code must retain the above copyright
- *    notice, this list of conditions and the following disclaimer.
- *
- * 2. Redistributions in binary form must reproduce the above copyright
- *    notice, this list of conditions and the following disclaimer in
- *    the documentation and/or other materials provided with the
- *    distribution.
- *
- * 3. The name "Carnegie Mellon University" must not be used to
- *    endorse or promote products derived from this software without
- *    prior written permission. For permission or any legal
- *    details, please contact
- *      Carnegie Mellon University
- *      Center for Technology Transfer and Enterprise Creation
- *      4615 Forbes Avenue
- *      Suite 302
- *      Pittsburgh, PA  15213
- *      (412) 268-7393, fax: (412) 268-7395
- *      innovation@andrew.cmu.edu
- *
- * 4. Redistributions of any form whatsoever must retain the following
- *    acknowledgment:
- *    "This product includes software developed by Computing Services
- *     at Carnegie Mellon University (http://www.cmu.edu/computing/)."
- *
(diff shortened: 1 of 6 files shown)


COPYING
```diff
@@ -1,43 +1,43 @@
...
- *     at Carnegie Mellon University (http://www.cmu.edu/computing/)."
- *
- * CARNEGIE MELLON UNIVERSITY DISCLAIMS ALL WARRANTIES WITH REGARD TO
- * THIS SOFTWARE, INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY
- * AND FITNESS, IN NO EVENT SHALL CARNEGIE MELLON UNIVERSITY BE LIABLE
- * FOR ANY SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
- * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
- * AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING
- * OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+Copyright (c) 1994-2016 Carnegie Mellon University.  All rights reserved.
```
elliefm: I bet the weird leading asterisks came from somebody copy and pasting:

```
/*
 * Copyright (c) 1994-2016 Carnegie Mellon [...]
 * [...]
 */
```

from a C file and then only removing the `/*` and `*/` outer lines...

cunit/times.testc
```diff
@@ -1,44 +1,6 @@
...
- *
- * CARNEGIE MELLON UNIVERSITY DISCLAIMS ALL WARRANTIES WITH REGARD TO
- * THIS SOFTWARE, INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY
- * AND FITNESS, IN NO EVENT SHALL CARNEGIE MELLON UNIVERSITY BE LIABLE
- * FOR ANY SPECIAL, INDIRECT OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
- * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN
- * AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING
- * OUT OF OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
- */
+/* test_time.c -- regression test for cyrus time routines */
```
elliefm, line 1: Here's a case where the comment names the file, but gets the name wrong.  The file was probably renamed at some point.  I wonder if there's an easy way to find and fix these too?

perl/sieve/scripts/installsieve.pl
```diff
@@ -1,4 +1,6 @@
 #! /bin/sh
+exec perl -x -S $0 ${1+"$@"} # -*-perl-*-
+#!perl -w
 # SPDX-License-Identifier: BSD-3-Clause-CMU
```
elliefm, line 4: This change looks like it's fixing an overzealous deletion from earlier, but was squashed with the wrong commit?

rjbs: Whoops, fixed, thanks!

cunit/times.testc
rjbs: I want to normalize all these soon, so if this is merged, expect a later commit that checks all the names and standardizes on `-` or `--` not both.  There are also some places with two spaces before the filename.

PR #5740, opened 2025-12-29
Cassandane: when testing Cassini code, ignore CASSINI_FILENAME

rjbs:
Otherwise, cyd distcheck, which sets this, will hose the tests.


rjbs:
Short version is that `local %ENV` like that won't do what you want.  It will take place at the runtime phase of loading Cassini.pm, and then its duration will end.  Later, when methods in Cassini are being called, %ENV will have been restored.

PR #5741, opened 2025-12-29
release prep for v3.13.1

rjbs:


rjbs:
@elliefm This isn't _really_ ready to go.  Is it?  I think that most of what I want to improve might be "let's talk about how to improve this for the February release" rather than "let's do now", but _probably_ I've stuffed some things up.

Give this a look sometime, and next time we talk let's figure out the path to shipping a tolerable-but-imperfect 3.13.1, so we can ship a better 3.13.2…

docsrc/imap/download/release-notes/3.13/x/3.13.1.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Security fixes
+==============
+
+* None yet!
+
+Significant bugfixes
+====================
+
+* ``lmtpd`` will no longer reset the session id after ``LHLO``.
```
elliefm: I usually style bug fixes something like:

```
* Fixed :issue:`1234`: :cyrusman:`lmtpd(8)` will no longer [...]
```

I find the PR number by getting the commit id that touched the changes file, searching for that in the output of `git log --graph`, and then looking for where that branch was merged (if you have a more efficient way to resolve a commit to its merge point I'd love to hear it!)

If the PR references an issue number I'll use the issue number, if not I'll use the PR number.  Using a PR number for an `:issue:` is a little weird, but the link works correctly in the end.

I don't always link to an issue, sometimes there isn't a useful thing to link to.  In that case I'll just write "* Fixed: thing that was broken" or whatever

docsrc/imap/download/release-notes/3.13/x/3.13.1.rst
```diff
@@ -0,0 +1,79 @@
...
+* The annotator is now called only on items added to email-type mailboxes.
+* There is now a ``--with-zoneinfo-dir`` option to configure.
+* ``ctl_conversationsdb`` now has ``-U`` for version upgrades.
+* ``lmtpd`` now provides a capability called ``TRACE`` and a new ``TRACE``
+  command for setting an interal trace-id, which will be logged.
+* The ``admins_get_internal_seen`` configuration option has been added.  If
+  set, the admin users will always get and set the same ``\Seen`` flags as the
+  mailbox owner.
+* ``master(8)`` will refuse to start if multiple partitions appear to be
+  identical or overlap.
```
elliefm: General style notes:

* when referring to a Cyrus tool, service daemon, config file, etc, use `:cyrusman:` so the web reader gets a hyperlink directly to the docs for it.  The syntax has backticks so it's hard to show here, look for examples in use.
* when referring to imapd.conf options, I would usually say something like "The ``admins_get_internal_seen`` [cyrusman imapd.conf] option has been added".  (I'd prefer to link directly to the option, but we can't.  I have plans, but they're part of a big imapoptions refactor...)

rjbs:
> I'm a little uncomfortable with the "don't use this" bit in the upgrade docs, but I am only ever writing upgrade docs for Big Actual Releases. I think it's fine for _this_ snapshot, as long as we do actually iterate on it -- I'm scared of setting a precedent that upgrade docs aren't necessary for dev snapshots, and then we get to the Big Actual Release and they still haven't been written.

To be clear, I'd like to wear this warning down to "hey this isn't for general use" and less "don't use this!" as we get the release process just a little more stable and reliable-feeling.  :)

docsrc/imap/download/release-notes/3.13/x/3.13.1.rst
dilyanpalauzov: As a matter of fact, when announcing major version the ticket numbers for the changes are on many places not mentioned, but when announcing minor release, the ticket number are very often included in the release notes.  Compare https://www.cyrusimap.org/dev/imap/download/release-notes/3.12/x/3.12.1.html and https://www.cyrusimap.org/dev/imap/download/release-notes/3.12/x/3.12.0.html .  Announcing a change sometimes with associated tracking number and sometimes without the tracking number is for me inconsistent.

dilyanpalauzov:
The changes/statements in rfs-support.rst about implemented JMAP RFCs/drafts, differ from the statemets in docsrc/imap/download/installation/http/jmap.rst “In development”, cf. https://github.com/cyrusimap/cyrus-imapd/issues/5745 . Such changes to the documentation should be done irrespective of the 3.13 series.

The purpose of the changes/next/ files is to prepare the release notes.  Once the information is included in the release notes, why are these files not deleted?

PR #5743, opened 2025-12-30
Cassandane: merge test entity Factory/Instance classes into single module

rjbs:


cassandane/Cassandane/TestEntity/DataType/AddressBook.pm
```diff
@@ -0,0 +1,53 @@
+use v5.28.0;
+package Cassandane::TestEntity::DataType::AddressBook;
+
+package Cassandane::TestEntity::Factory::AddressBook {
+  use Moo;
```
wolfsage: Should be 4 space indent I think?

rjbs: oops!

wolfsage (approved):

PR #5744, opened 2025-12-30
Cassandane: add Pod to TestUser

rjbs:


cassandane/Cassandane/TestEntity/DataType/Mailbox.pm
```diff
@@ -182,6 +203,28 @@ package Cassandane::TestEntity::Instance::Mailbox {
...
+may change) and returns an Email instance object.
+
+Valid arguments to include in the passed hashref are:
+
+    to      - the string to use as the To header
+    from    - the string to use as the From header
+    subject - the string to use as the Subject header
+    message_id - the string to use in the Message-Id header; include <...>!
+
+    headers - an reference to an array in the form [ name, value, n2, v2, ... ]
```
wolfsage: nit: an reference -> a reference

cassandane/Cassandane/TestUser.pm
```diff
@@ -1,6 +1,32 @@
 package Cassandane::TestUser;
 use Moo;
 
+=head1 NAME
+
+Cassandane::TestUser - a handle on a Cyrus user with test-related methods
```
wolfsage, line 6: nit: "handle" is weird isn't it?

rjbs: I don't think so.  I can reword it if you want, but I think this is totally normal jargon:

https://en.wikipedia.org/wiki/Handle_(computing)

PR #5747, opened 2026-01-02
carddav_db.c: Strip shared/ prefix from group member queries

wolfsage:
We're already stripping it from _getemails, but we failed to strip it when querying groups for a list of members.

This prevented Email/query fromContactGroupId from working when the legacy 'shared/' prefix was used before the group uid.


ksmurchison (approved):
If we need it and it works, LGTM

rsto:
> We're just going to leave it include-in-fm until we don't need it then we can close this

I understand this to mean that we do not want to merge this PR, so I'll tag is as such. Please remove the label if I'm mistaken.

rjbs:
@wolfsage I think this can be closed, what do you think?  `shared/` prefix has been fully removed from Fastmail..?

wolfsage:
I'd say... we can do it and hope nothing breaks :) (Might need to fix up some tests though). Or we could modify this MR to add logging if this condition is hit, run it for a week, and then remove it if we see no hits. Your call!

rjbs:
I've removed the label and will do a test run.  I think there's some code we can delete yet, in Fm, but only stuff to parse the shared/ form which we don't have anymore.

wolfsage:
I believe this is no longer needed.

PR #5750, opened 2026-01-07
TestCase: add :min_other_version and :max_other_version magic

elliefm:
This is the same as :min_version and :max_version, except that it asserts that the 'other' instance meets some version criteria for tests that spin up multiple instances.  This is the 'replica' instance in replication tests, and the 'frontend' and 'backend2' instances in murder tests.

This should be used sparingly for preventing false negatives during cross-version testing:

* tests that exercise some new protocol feature that is known to be not understood by older versions can be marked :min_older_version_x_y (where x.y is when the feature was added)
* tests that exercised some protocol feature that has been removed can be annotated :max_other_version_x_y (where x.y was the last version that contained the feature)

Also fixes a couple of cross-version problems of the low-hanging fruit variety.

### DIFF
--- a/cassandane/Cassandane/Cyrus/FastMail.pm
+++ b/cassandane/Cassandane/Cyrus/FastMail.pm
@@ -157,7 +157,7 @@ sub new
         jmap => 1,
         deliver => 1,
         adminstore => 1,
-        services => [ 'imap', 'http', 'sieve', 'backupcyrusd' ]
+        services => [ 'imap', 'http', 'sieve' ]
     }, @args);
 
     $self->needs('component', 'jmap');
--- a/cassandane/Cassandane/Cyrus/Reconstruct.pm
+++ b/cassandane/Cassandane/Cyrus/Reconstruct.pm
@@ -696,6 +696,7 @@ sub test_downgrade_upgrade
 sub test_upgrade_v19_to_v20
     :MailboxLegacyDirs :NoAltNameSpace :Conversations :Replication
     :NoStartInstances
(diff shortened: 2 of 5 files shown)


rsto (approved):

cassandane/Cassandane/Cyrus/FastMail.pm
```diff
@@ -157,7 +157,7 @@ sub new
         jmap => 1,
         deliver => 1,
         adminstore => 1,
-        services => [ 'imap', 'http', 'sieve', 'backupcyrusd' ]
+        services => [ 'imap', 'http', 'sieve' ]
```
wolfsage, line 160: This was added explicitly by @brong in f5f9c71233f5f333558ba2c07bf28bb1a305ecda (though it's unclear why); is it okay for this to go?

elliefm: The only test that actually used backupcyrusd is now annotated to get it.

Having it enabled for _every_ test (in this suite) broke every replication test (in this suite) when the replica installation didn't have such a service to start.  If the intention is to always have it running but doing nothing (for some reason), that's probably doable, but will require some different, more complicated conditionalisation in order for it to not break in cross-version tests.  In that case I'll drop the one commit and leave it in the too hard basket for now.  But making backupcyrusd specific to the one test that needs it looked like a quick win, so that's what I did.

wolfsage: Great, thanks

PR #5771, opened 2026-01-20
JMAP: require accountId unless specified otherwise

ksmurchison:
This is just a start.  A ton of Cass tests will have to be updated unless the underlying perl test code can be made to auto-include accountId.

AND, this should not be merged until we know that the FM UI and middleware do the right thing.


rjbs:
I've just done that by pushing two commits to this PR.  One adds accountId in most places, the other is sort of a demonstration fix for a common problem.

ksmurchison:
This needs a fix to JMAP-TestSuite to either auto-add accountId or to all of the tests to do so.

brong:
Superseded.

PR #5776, opened 2026-01-23
imapoptions refactor

elliefm:
Replaces the file `lib/imapoptions` with a directory of files in a format that's easy to edit and parse, and replaces the `tools/config2header`, `tools/config2rst`, and `tools/config2sample` scripts with a new one that's easy to extend and maintain.

New perl dependencies: Moo (libmoo-perl), App::Cmd (libapp-cmd-perl).  You probably already have these if you work on Fastmail perl stuff.

For this PR, one of my goals was to keep the output files as close as possible to what the previous scripts produced, to make it easy to verify that I hadn't broken anything.  To this end, there are some things in here that aren't great and/or which reproduce quirks of the old scripts; and we're not yet taking any advantage of the new architecture.  I expect to improve on this in future PRs.

I regard the "imapd.conf.sample" file as the least important of the generated files, and so it has the most differences for now.  These are side effects from making the other output files match closely.

**Reviewers:** I suggest reviewing commit by commit.  I don't suggest reading all 500ish new imapoptions files though!  Rather, I suggest looking at diffs of the generated files, like this:

1. **Starting on the master branch**, apply the sorted-imapoptions patch from #5814
2. Build as usual, and also run `make doc`
3. Copy the following generated files aside for later (perhaps to `/tmp`):
   * lib/imapopts.c
   * lib/imapopts.h
   * docsrc/reference/manpages/configs/imapd.conf.rst
   * doc/examples/imapd_conf/imapd.conf.sample
   * doc/html/reference/manpages/configs/imapd.conf.html
   * man/imapd.conf.5
4. Check out the branch for this PR
5. Build as usual, and also run `make doc`
6. Diff the outputs from this PR against the ones you saved earlier.  `vimdiff` is helpful here, if you already know vim.

Expect only the following differences:
* the "auto-generated..." comments at the top of each file
* imapopts.c: (no other differences)
* imapopts.h: there's now a comma after the final `IMAP_ENUM_VIRTDOMAINS_ON,`
* imapd.conf.rst: some added or removed empty lines; "Deprecated..." text is now consistent; some options that were missing this text now have it; "Allowed values: ..." is now indented correctly; junk in `replicaonly` fixed
* imapd.conf.html: as per imapd.conf.rst
* imapd.conf.5: as per imapd.conf.rst
* imapd.conf.sample: no longer includes the man page header or footer; some added or removed empty lines; documentation blobs now contains rst markup; "Deprecated..." text is now consistent


elliefm:
I've ploughed forward; what's there now might not make much sense (or even work) because I stopped in the middle of something to refactor an annoying detail (the `default_value` handling), but ran out of steam with the refactor barely started.  More next time...

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -0,0 +1,297 @@
...
+
+sub _allows_null
+{
+    my ($type) = @_;
+
+    return ($type eq 'STRING'
+            || $type eq 'STRINGLIST'
+            || $type eq 'BYTESIZE'
+            || $type eq 'DURATION');
+}
```
wolfsage, line 142: Written like a C programmer :). It's fine, but often we use hash lookups for stuff like this:
```
my %allows_null = map { $_ => 1 } qw(STRING STRINGLIST BYTESIZE DURATION);

sub _allows_null
{
    my ($type) = @_;
    return $allows_null{$type};
}
```

tools/lib/Cyrus/IMAPOptions/App/Command/csource.pm
```diff
@@ -0,0 +1,125 @@
...
+    return (
+        [ 'cc=s', 'specify the C compiler in use', { default => 'gcc' } ],
+    );
+}
+
+sub validate_args
+{
+    my ($self, $opt, $args) = @_;
+
+    $self->usage_error("need imapoptions path") if @{$args} < 1;
```
wolfsage: nit here and elsewhere: `@{$args}` is far noisier than `@$args`

tools/lib/Cyrus/IMAPOptions/App/Command/csource.pm
```diff
@@ -0,0 +1,125 @@
...
+    $self->usage_error("too many arguments") if @{$args} > 1;
+
+    $self->usage_error("imapoptions must be a directory")
+        if ! -d $args->[0];
+}
+
+sub execute
```
wolfsage, line 26: Nit here and elsewhere: `use experimental 'signatures'` and then this:

```
sub execute
{
    my ($self, $opt, $args) = @_;
```

becomes:

```
sub execute ($self, $opt, $args)
{
```

which is imo way better

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -0,0 +1,297 @@
...
+    is => 'ro',
+    default => undef,
+);
+has documentation => (
+    isa => ArrayRef[Str],
+    is => 'ro',
+    required => 1,
+);
+
+sub BUILDARGS
```
wolfsage: IMO overriding this entirely is kind of strange. Also we tend to treat arguments to Moo/Moose classes as hashrefs. Moo by default will allow either.

I think these should instead be:
```
around BUILDARGS => sub
{
    my ($orig, $class, @args) = @_;

    # Normalizes new(x => 1) and new({ x => 1}) into a hashref for us
    my $args = $class->$orig(@args);
    if (my $filename = delete $args->{from_file}) {
        ...
    }

    return $args;
}
```

tools/lib/Cyrus/IMAPOptions.pm
```diff
@@ -0,0 +1,57 @@
...
+# See COPYING file at the root of the distribution for more details.
+package Cyrus::IMAPOptions;
+use Moo;
+
+use Cyrus::IMAPOptions::Option;
+use Types::Standard qw(HashRef InstanceOf);
+
+# XXX this says "either C:I:O or M:O", but i need "both C:I:O and M:O".
+# XXX it's not enough to just say "C:I:O", it complains that it's also M:O.
+has options => (isa => HashRef[InstanceOf['Cyrus::IMAPOptions:Option',
```
wolfsage: nit: The indenting of this entire block is bizarre to me. I see how you got there though.
I'd do:
```
has options => (
    isa => HashRef[InstanceOf[qw(Cyrus::IMAPOptions::Option Moo::Object)]],
    is => 'ro',
);
```

Though I'm curious why you're allowing any Moo::Object derived class here?

wolfsage (commented):
Overall I think our Moo / App::Cmd use is mostly fine! Rik is the authority though so maybe wait for him.

That said, I think `sub BUILDARGS` is weird (I've detailed why inline).

I think you might also want your Moo classes to `use MooX::StrictConstructor` -- you'll need to install it, but without it you could pass typos to the classes and not have the data you expect.

tools/lib/Cyrus/IMAPOptions/Option.pm
elliefm: Yeah right, I do that all the time too.  So distracted with working out how to use Moo that I forgot how to write perl lmao.

tools/lib/Cyrus/IMAPOptions/App/Command/csource.pm
elliefm: House style at the place I learned perl was to _always_ use `x{}` when dereferencing, for reasons that made sense at the time but I no longer particularly recall.  Perhaps because in complex statements it makes it unambigous what exactly is being dereferenced?  And it's easier to enforce "always do it this way" than to split hairs about which situations are complex or simple.

I don't have an opinion of my own, that's just muscle memory at work.

tools/lib/Cyrus/IMAPOptions/App/Command/csource.pm
elliefm: Ah yeah, I've seen that in some of Rik's newer stuff, and was thinking of trying it out, but didn't (yet).

This tool will run at Cyrus compile time, so there's a line to be navigated between making sure it works with whatever perl (and modules) the person building it happens to have, and making our own lives easier by assuming that's reasonably modern.  I don't know where that line is exactly, though I'm already breakdancing all over it by making this tool depend on Moo and App::Cmd...

I'll probably have a chat with Rik sometime about the minimum perl version I ought to be targeting (since last time I paid specific attention to new perl versions was when 5.10 came out), and then I'll be able to think about using quality of life features that are available in that version.  Though `experimental` gives me pause; in my very limited experience those have a habit of being removed later.

tools/lib/Cyrus/IMAPOptions/Option.pm
elliefm: Interesting.  I saw examples of doing this both ways, and it wasn't clear why they were making those choices.  `around` is neat, but it's not a tool I've used before, so I went with what was simpler (familiar) and figured if it broke I'd change it, and then it worked so I didn't give it any further thought.

tools/lib/Cyrus/IMAPOptions.pm
elliefm: Comment right above explains why!

What I want here is just `HashRef[InstanceOf[qw(Cyrus::IMAPOptions::Option)]]`, but when I did that and then passed it objects of that class, it complained that the objects were of both types, which they are by inheritance.  I couldn't quickly find a way to express precisely what I was after; this is just what got me unblocked.  If you know how to do this correctly please lmk!

tools/lib/Cyrus/IMAPOptions/App/Command/csource.pm
wolfsage: It's fine, just noting that it could be easier to read IMO :)

tools/lib/Cyrus/IMAPOptions/App/Command/csource.pm
wolfsage: Sounds good. FWIW signatures came out in 2014 (12 years ago!) so hopefully it's on all major operating systems

tools/lib/Cyrus/IMAPOptions.pm
wolfsage: hah sorry, the weird indenting distracted me too much and I failed to read the comment!

That InstanceOf is strange, but okay!

tools/lib/Cyrus/IMAPOptions/App/Command/cheader.pm
```diff
@@ -0,0 +1,157 @@
+# SPDX-License-Identifier: BSD-3-Clause-CMU
+# See COPYING file at the root of the distribution for more details.
+package Cyrus::IMAPOptions::App::Command::cheader;
+
```
wolfsage, line 6: All of the Command::* classes need `use strict; use warnings;`.

The other stuff that does `use Moo;` gets that for free, but importing `Cyrus::IMAPOptions::App -command` isn't enough.

tools/lib/Cyrus/IMAPOptions/App/Command/csource.pm
```diff
@@ -0,0 +1,129 @@
...
+               $option->imapopt_last_modified,
+          ), ",\n";
+    print '    ';
+    print join(', ',
+               $option->imapopt_deprecated_since,
+               $option->imapopt_replaced_by, # XXX check that this option exists
+          ), ",\n";
+
+    my ($ctype, $default_value) = $option->imapopt_default_value;
+    foreach (qw(val, def)) {
```
wolfsage: Once `use warnings` is enabled, it'll complain about the comma here.

This is setting `$_` to 'val,' and then 'def', which I'm guessing isn't intended.

Even weirder, this loop isn't using `$_` at all, so I'm not sure what it is or isn't doing that it's supposed to be.

wolfsage (dismissed):

tools/lib/Cyrus/IMAPOptions/App/Command/cheader.pm
elliefm: Ahh right.  I always use those two, except the Moo docs told me I don't need to with Moo so I didn't.  But then when I added the App::Cmd stuff, which is also new to me, I forgot that it wasn't Moo.

tools/lib/Cyrus/IMAPOptions/App/Command/csource.pm
elliefm: Hah.  This was originally something like `for (1..2)`, but I added the unused strings as a bit of documentation of what the two fields we're writing are.  But hmm.  Syntax error aside, the fact they're unused is itself a confusing distraction, when I just looked at this code again myself I was like "???" for a moment.  I'll go back to the simple loop and put a comment instead.

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -0,0 +1,218 @@
+# SPDX-License-Identifier: BSD-3-Clause-CMU
+# See COPYING file at the root of the distribution for more details.
+package Cyrus::IMAPOptions::Option;
+use Moo;
+use feature 'state';
```
rjbs: I think we should target using a specific version bundle, for consistency across files.  We currently use v5.28.0 for that, so let's use that here.  That will enable `state`.  (Later, I will argue for requiring v5.36.0, but that can be tomorrow's problem.)

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -0,0 +1,218 @@
...
+
+use File::Basename;
+use Types::Standard qw(ArrayRef Bool Enum Int Maybe Split Str);
```
rjbs: Not positive we want to bring this in.  It's kind of a weird library… but now I see we already use it in Cyrus::Mbwhatever.  Okay, go for it.  But I'll be lookin' like this: 👀

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -0,0 +1,218 @@
...
+
+my @option_types = qw(BITFIELD BYTESIZE DURATION ENUM INT
+                      STRING STRINGLIST SWITCH);
+
+has name => (
+    isa => Str,
+    is => 'ro',
+    required => 1,
+);
+has type => (
```
rjbs, line 20: Please put a blank between `has` calls.  (Fastmail style: "code in paragraphs")

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -0,0 +1,218 @@
...
+    required => 1,
+);
+has allowed_values => (
+    isa => ArrayRef->of(Str)->plus_coercions(Split[qr/\s/]),
+    is => 'ro',
+    predicate => 1,
+    coerce => 1,
+);
+has default_value => (
+    isa => Maybe[Str] | ArrayRef[Str],
```
rjbs, line 33: I'll try to remember to look at how this is used, but:  imho, union types are a code smell, and `Maybe[Str]` is already a union of `undef | Str`.  If `[X]` and `X` mean the same thing, I'd always require `[X]` and you could coerce `X` into it.  (Where do we have a multivalue default, btw?)

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -0,0 +1,218 @@
...
+);
+has last_modified => (
+    isa => Str,
+    is => 'ro',
+    required => 1,
+);
+has deprecated_since => (
+    isa => Maybe[Str],
+    is => 'ro',
+    predicate => 1,
```
rjbs, line 47: I suspect this is a mistake.  Sorry if this is pedantic and unneeded:

A Moo[se] attribute has two states:  set and unset.  This is analogous to the `exists` test on a hash.  Then the value that exists might be defined or undefined.  The `isa` type constraint governs what values are permitted, if one is set.

Do you really want a distinction between "there is a deprecated_since, which is undef" and "there is no deprecated_since"?

I suspect you want the type to be `Str` and to have a predicate.  That way, it can't be set to undef.  It'll either be unset (`! $opt->has_deprecated_since`) or set and a string.

rjbs: This comment applies to a number of attributes in this class.

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -0,0 +1,218 @@
...
+    return exists $allows_null{$type};
+}
+
+sub _from_file
+{
+    my ($args, $filename) = @_;
+
+    open my $in, '<', $filename or die "$filename: $!";
+
+    do {
```
rjbs, line 96: I don't think you need that `do` — this is just a bare block, which is permitted.  You only need `do` to make it an expression rather than a statement.

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    return $args;
+};
+
+sub _type_allows_null
+{
+    my ($type) = @_;
+
+    state %allows_null = map { $_ => 1 } qw(BYTESIZE DURATION STRING STRINGLIST);
```
rjbs, line 85: (This construct (assignment to an aggregate type in `state`) specifically requires v5.28, another reason to put `use v5.28.0` atop this file.)

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -0,0 +1,218 @@
...
+
+    return exists $allows_null{$type};
```
rjbs, line 87: All your assigned values are true so you can drop `exists` if you want. :)

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -0,0 +1,218 @@
...
+{
+    my ($self, $args) = @_;
+
+    my $type = $self->type;
+    my $dv = $self->default_value;
+
+    die "$type can not have NULL default_value"
+        unless defined $dv || _type_allows_null($type);
+
+    if ($type eq 'STRINGLIST' || $type eq 'ENUM') {
```
rjbs, line 154: This is fine and will work, but I think you're undercommitting to OO here, and it's working against code clarify.  The general red flag here is:

```perl
sub do_something ($self) {
    if    ($self->type eq 'Foo') { … }
    elsif ($self->type eq 'Bar') { ... }
   ...
}
```

The question here is always:  is `type` property standing in for a subclass?  To me, it seems like it is.  If you had a very small subclasses, (Option::Bitfield, etc.) you could put this kind of checking on those.

This advice might actually be too onerous to take, I'm not trying the refactor in my head to say, but it also might eliminate a bunch of if/else and get you clearer code.

It's also maybe where/how you'd be able to un-union some types.

tools/lib/Cyrus/IMAPOptions.pm
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }
+    closedir $dh;
+}
+
+sub BUILD
+{
+    my ($self, $args) = @_;
+
+    while (my ($opt_name, $option) = each %{$self->options}) {
```
rjbs: Rule of thumb: never use `each`.  It _can_ be safe, but it's too easy to screw up.  Instead, the more annoying:

```perl
for my $opt_name (keys $self->options->%*) {
  my $option = $self->options->{$opt_name};
  ...
}
```

The issue is that `each` stores the iterator state on the hash itself, making it non-reentrant, so if you ever accidentally create an option set while creating an option set… boom.  You can often see "well, _here_ it's safe," but then everybody reading the code needs to be confident enough in the whole system to know that, too.

tools/lib/Cyrus/IMAPOptions/App/Command/csource.pm
rjbs: I will jump into this resolved thread and say:  circumfix dereferencing is *nearly* indefensible these days, and we should eliminate it.  More:

`@{$foo}` is always less clear than `@$foo` as wolfsage said, but you could say "well we always use that because then we *always* have the braces like we need in more complex sitautions."

Those more complex situations are things like:
* `@{ $obj->method }`
* `@{ $arr->[0]{key} }`
* `$#{ $arrayref_for{key} }`

But all of these forms have been superseded by postfix deref, which *always* reads *left to right*, with no scanning back and forth:

* `$obj->method->@*`
* `$arr->[0]{key}->@*`
* `$arrayref_for{key}->$#*`

(Slice syntax similarly can be postfixed now.)

So usual Fm Perl style is:  prefix a simple variable, postfix a complex expression, circumfix never.

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -237,4 +243,154 @@ sub docs_default_value
...
+        return ('uint64_t', $dv);
+    }
+    elsif ($type eq 'ENUM') {
+        return ('enum enum_value',
+                _c_enum_name($self->name, $self->default_value));
+    }
+    elsif ($type eq 'INT') {
+        return ('long', $self->default_value);
+    }
+    elsif ($type eq 'SWITCH') {
```
rjbs, line 346: This comb of types makes me want to reemphasize my "maybe subclasses" here, so I am doing so :)

tools/lib/Cyrus/IMAPOptions/App/Command/cheader.pm
```diff
@@ -34,7 +29,12 @@ sub execute
 {
     my ($self, $opt, $args) = @_;
 
-    my $imapoptions = Cyrus::IMAPOptions->new(from_path => $args->[0]);
+    my $imapoptions = Cyrus::IMAPOptions->new(
+        from_path => $args->[0],
+        forbid_unreleased => $opt->{forbid_unreleased},
```
rjbs: Always use `$opt->option_name` and never `$opt->{option_name}`.  The second one is mostly an implementation detail, and subject to change.

rjbs (commented):
I added a bunch of commentary on here, but did not really attempt a holistic review of everything.  Mostly I said what I thought should change, but this looks good, and while it's still complex, it seems like easier to manage complexity than we have now.

tools/lib/Cyrus/IMAPOptions/Option.pm
elliefm: Thanks, and that answers another question I was about to have on a different PR: "what minimum version of perl should configure insist upon".

tools/lib/Cyrus/IMAPOptions/Option.pm
elliefm: So, I spent ages trying different approaches to `default_value` (all rebased out), and this is the cleanest implementation I landed on.

BITFIELD types can and do have multi-value defaults (e.g. sieve_extensions), which is why this isn't just Maybe[Str].  Always coercing to an ArrayRef[Str] was one of the approaches I tried and threw out -- it made BITFIELDS easy to handle, but required special handling everywhere for everything else.  The union puts all the complexity in the initial parse (`_from_file`, called from `BUILDARGS`) and validation (`BUILD`), and then using the object is smooth sailing.

The three possibilities in this union are:

* undef: This will become `NULL` in C, which is needed for some of the STRING(ish) options.
* ArrayRef[Str]: Strictly BITFIELD options only.
* Str: Any type other than BITFIELD.  It may be an empty string, which will become `""` in C.

tools/lib/Cyrus/IMAPOptions/Option.pm
elliefm: Ahh, thanks.  You're right, it's a mistake.

`default_value` is the only attribute for which "exists but is undef" is valid.   I think I discovered `Maybe[]` while figuring out how to achieve that, but then misunderstood it and misapplied it to the optional attributes too.  Will tidy up.

tools/lib/Cyrus/IMAPOptions/Option.pm
elliefm: Bare blocks look like typos to me, whereas `do block;` looks weird but intentional.

tools/lib/Cyrus/IMAPOptions/Option.pm
elliefm: Ahh interesting, didn't know that!

tools/lib/Cyrus/IMAPOptions/Option.pm
elliefm: I usually make my lookup tables have undef values, and then use exists on the key.  I wonder why I made this one have true values??

Actually, I think I originally did this with a hash slice, something like `my %allows_null; @allows_null{qw(BYTESIZE etc)} = (undef);`.  I guess when I converted to using `state`, and initialising it in the declaration, I had to switch to map.  Still not sure why I used `{ $_ => 1 }` rather than `{ $_ => undef }`...

I'll change... something about this.  Any performance or style preference for true values vs exists lookups?

tools/lib/Cyrus/IMAPOptions/Option.pm
elliefm: Undercommitting to OO is a deliberate choice here.  I did think of doing something like that, and would have if there were many more types with many more variant behaviours.  This one's kind of an edge case -- there are variations, but only a couple.  Feels clearer to keep the variants together where you can see how they differ, rather than spread the differences out across a bunch of files.

`c_default_value()` is probably the worst example of this with the most variants, but I'm especially reluctant to spread the C formatting across multiple files.

tools/lib/Cyrus/IMAPOptions.pm
elliefm: Whoa, good tip.  Will change.

tools/lib/Cyrus/IMAPOptions/Option.pm
elliefm: Hah, this is the very spot I called out in my earlier comment as being a set of variants I specifically want to keep together!

tools/lib/Cyrus/IMAPOptions/AllowedValues.pm
```diff
@@ -0,0 +1,128 @@
...
+sub value_alias_strings
+{
+    my ($self) = @_;
+
+    my @strings;
+
+    foreach my $value ($self->_order->@*) {
+        # XXX postfix deref?
+        my @aliases = @{$self->_values->{$value} || []};
+
```
elliefm, line 103: @rjbs,

It won't let me reply to your comment about postfix derefs, but I just read about them and I like them a lot, thanks!  Have switched over to using them, except on this one line which I'm not sure how to handle.

The mechanical conversion seems like:

```
my @aliases = ($self->_values->{$value} || [])->@*;
```

But that's using logical `()` in list context, which feels misleading.  Is there a better way to do this?

elliefm:
~~Needs to be rebased and updated after #5831 lands.~~

rsto (commented):
I fail to even build the docs, which likely isn't related to this PR, though:

    $ make doc
    [...]
    error: missing documentation dependencies. cannot build html documentation
    make: *** [Makefile:10052: doc/html/.sphinx-build.stamp] Error 1

    $ sudo apt install python3-sphinx python3-git python3-sphinx-rtd-theme doxygen
    Reading package lists... Done
    Building dependency tree... Done
    Reading state information... Done
    python3-sphinx is already the newest version (5.3.0-4).
    python3-git is already the newest version (3.1.30-1+deb12u2).
    python3-sphinx-rtd-theme is already the newest version (1.2.0+dfsg-1).
    doxygen is already the newest version (1.9.4-4).

I can't tell from `Makefile.am` what's missing, but I haven't looked closely. I'll try sort this out, but I'm also fine if someone else reviews and approves this PR so please don't wait for my approval.

changes/next/cyr-1192-imapoptions-refactor
```diff
@@ -0,0 +1,24 @@
...
+
+Config changes:
+
+(all, or none, depending on how you think about it)
+
+
+Upgrade instructions:
+
+Need perl modules "Moo" and "App::Cmd".  In apt, these are libmoo-perl
+and libapp-cmd-perl.
```
dilyanpalauzov, line 19: These modules should be stated at https://www.cyrusimap.org/dev/developer/compiling.html.

Makefile.am
```diff
@@ -503,12 +513,11 @@ dist_noinst_SCRIPTS = \
     perl/sieve/scripts/sieveshell.pl \
     tools/arbitronsort.pl \
     tools/compile_st.pl \
-    tools/config2header \
-    tools/config2rst \
```
dilyanpalauzov, line 507: docsrc/Makefile uses tools/config2rst

wolfsage:
| make: *** [Makefile:10052: doc/html/.sphinx-build.stamp] Error 1

This is broken for me too. Seems like doc/html doesn't event exist?

tools/lib/Cyrus/IMAPOptions/Option.pm
rjbs: Looks like I never replied:  Nah.  I like the PBP (Perl Best Practices) advice to name vars like this as if they were predicate functions, and use true values.

```perl
my %is_cool = map {; $_ => 1 } qw( john george ringo );

say "They're cool" if $is_cool{ $user };
```

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -56,6 +56,12 @@ has documentation => (
     is => 'ro',
     predicate => 1,
 );
+has c_enum_size => (
+    isa => Maybe[Int],
+    is => 'rwp',
+    lazy => 1,
+    builder => sub { scalar shift->c_allowed_values; },
```
rjbs: That line gives me the willies!

I propose you make it:

```perl
builder => sub ($self) { scalar $self->c_allowed_values }
```

Eliminating the `shift` simplifies the precedence story for the skimmer.

tools/lib/Cyrus/IMAPOptions/App/Command/cheader.pm
```diff
@@ -0,0 +1,160 @@
+# SPDX-License-Identifier: BSD-3-Clause-CMU
+# See COPYING file at the root of the distribution for more details.
+package Cyrus::IMAPOptions::App::Command::cheader;
```
rjbs, line 3: cheader? no thanks, I prefer gouda

tools/lib/Cyrus/IMAPOptions.pm
```diff
@@ -63,6 +68,40 @@ sub BUILD
...
+            # "UNRELEASED" strings in lib/imapoptions with the version
+            # number that is about to be released.
+            # If you're not building a release, ignore it. :)
+            my $prefix;
+
+            if ($self->forbid_unreleased) {
+                $prefix = -t STDERR ? "\033[31;1merror:\033[0m" : 'error:';
+            }
+            else {
+                $prefix = -t STDERR ? "\033[33;1mwarning:\033[0m" : 'warning:',
```
rjbs, line 93: This is fine, but:  Term::ANSIColor is core, and can make this kind of thing easier to read. (You want `colored`).

tools/lib/Cyrus/IMAPOptions/AllowedValues.pm
```diff
@@ -0,0 +1,125 @@
...
+        die "'$v' defined twice"
+            if exists $values{$v} or exists $aliases{$v};
+
+        push @order, $v;
+
+        if (@a) {
+            $values{$v} = [ @a ];
+
+            foreach my $a (@a) {
```
rjbs: `$a` best avoided, as it is not subject to "use strict 'vars'" requirements.  I suggest … anything else, except `$b` which goes the same way.

lib/imapoptions/README.md
```diff
@@ -0,0 +1,137 @@
+Files in this directory define entries in the imapoptions database.  The
+imapoptions database is the canonical source of the imapd.conf options
+recognised by Cyrus.
+
+This database is used by tools/imapoptions.pl, which uses it to produce outputs
+in the various formats needed by parts of the build.
+
+FORMAT
+======
```
rjbs: Our other Markdown uses the `#` .. `##` format of headers, which I'd like to stick with, since it means more lines of content fit on the editor's screen at once.  Obviously, this is a nitpick.

rjbs (approved):
I made a few "this oughta change" suggestions, but nothing really critical.  More critical is that it seems doc building is broken (I have not tested, and trust the other commenters here).  Obviously, that needs fixed.

rjbs:
(I feel like all my previous commentary earlier in process got all my actual review done, this was just light follow-up!)

Makefile.am
elliefm: Ah, bugger.  Good catch, thanks.  Fixed (but I haven't pushed yet).

changes/next/cyr-1192-imapoptions-refactor
elliefm: Good call.  Also added `perl` which was missing.

tools/lib/Cyrus/IMAPOptions.pm
elliefm: I don't disagree, but here I'm copying the behaviour that already existed.  Behaviour changes will come later, when they can be tested and reviewed properly without ~600 other changed files cluttering things up.

elliefm:
@wolfsage,

> | make: *** [Makefile:10052: doc/html/.sphinx-build.stamp] Error 1
> 
> This is broken for me too. Seems like doc/html doesn't event exist?

I've fixed the problem Dilyan reported in the dodgy second Makefile, so `dar makedocs` works now.

Robert's problem turned out to be missing doc build dependencies, we talked about it in the meeting a few days ago, and it started working once he installed the stuff it was asking for.  I think he was missing `myst_parser` and `breathe`, which are part of Rik's new doxygen support.  If you've built the docs before but not in a while, you're probably missing them too.

Same advice applies here: look at the report from the end of the configure output.  Do you have all the "Documentation dependencies"?  If not, install them and try again.  If you do and it still doesn't work, please give me more details.  `config.log` (created by configure, send it by email) would be helpful, as would the full make command and its output.

wolfsage:
> Robert's problem turned out to be missing doc build dependencies, we talked about it in the meeting a few days ago, and it started working once he installed the stuff it was asking for. I think he was missing `myst_parser` and `breathe`, which are part of Rik's new doxygen support. If you've built the docs before but not in a while, you're probably missing them too.

Ah thanks;  `cyd build` without `--with-sphinx` does ` '--with-sphinx-build=no'`. Okay! It would be nice if the error message said *what* was missing or to check the `Documentation dependencies:` section in the config output

wolfsage (approved):

wolfsage:
Oh, this bit: "To install all dependencies from packages on Debian Jessie, use this:" ...

should that list include the new perl deps? (or is it so old who knows how out of date it is?)

elliefm:
As long as that blob of documentation is "for Debian Jessie", I can't meaningfully update it for this change, because I don't know if Jessie even has the required packages, or whether it would work with them if it does.

It sucks that it's so out of date, but fixing it properly isn't in scope for this PR.

PR #5783, opened 2026-01-27
calalarmd: cleanup conversations if email is not found

brong:
Without this, if a message didn't exist any more, the conversations database would be held open which then blocks that user until the server is restarted.

### DIFF
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -1871,10 +1871,10 @@ static int process_futurerelease(struct caldav_alarm_data *data,
         /* Locate email in \Scheduled mailbox */
         r = find_scheduled_email(emailid, &frock);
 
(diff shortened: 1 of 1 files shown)


imap/caldav_alarm.c
wolfsage: I *think* we bail out here (This is the only realistic place we wouldn't end up with a guid) so it'd be nice to see a log message is conversations_jmapid_guidrep_lookup returns an error code

wolfsage (changes requested):
Requested we get error reporting for conversations_jmapid_guidrep_lookup since it seems like that's where the problem lies...

imap/caldav_alarm.c
brong: Yep, fixed in the new version of this with the correct branch name!

PR #5784, opened 2026-01-27
 calalarmd: cleanup conversations if email is not found

brong:
This is a redo of !5783 with the correct branch name!  I've fixed it to return the error if the guid lookup fails, and also fixed the syslogged error message to show the emailid being looked up.


ksmurchison (approved):
LGTM

PR #5805, opened 2026-02-12
Make perl packages install

brong:
This moves the location of the Perl packages inside cyrus-imapd/perl/imap/ to a `lib` subdir so that `make install` will include them.


rjbs (approved):
Seems reasonable (although I look forward to ditching the "use lib" in cass… thought I had already!

Anyway, distcheck passes, too, so I think it'll be okay!

PR #5807, opened 2026-02-15
search_squat: Do not search in SEARCH_PART_NONE

### LINKED ISSUE #3562: 3.4.1 imapd segfaults
Hi,
I recently upgraded to 3.4.1 and ran into the issue of imapd segfaulting under very specific circumstances: 
Client: horde - it doesn't happen with thunderbird
search_engine:squat - it doesn't happen, if search_engine is undefined
Distros: debian 9 and debian 10 on x64 
Compiled with 
./configure --prefix=/usr/local/cyrus-imapd-3.4.1 --with-cyrus-prefix=/usr/local/cyrus-imapd-3.4.1 --with-syslogfacility=MAIL --enable-replication --with-perl=/usr/bin/perl --with-openssl --enable-idled --mandir=/usr/share/man --docdir=/usr/share/doc --with-zlib --with-bdb=no 

The core dump looks like this:
Program terminated with signal SIGSEGV, Segmentation fault.
#0  convert_cat (rock=rock@entry=0x5627e47c8c20, s=s@entry=0x1 <error: Cannot access memory at address 0x1>) at lib/charset.c:349
349         while (*s) {
(gdb) bt
#0  convert_cat (rock=rock@entry=0x5627e47c8c20, s=s@entry=0x1 <error: Cannot access memory at address 0x1>) at lib/charset.c:349
#1  0x00007f0e11d8b508 in charset_convert (s=s@entry=0x1 <error: Cannot access memory at address 0x1>, charset=<optimized out>, flags=5) at lib/charset.c:2086
#2  0x00007f0e122faaa3 in match (bx=0x5627e47c1d00, part=<optimized out>, str=0x1 <error: Cannot access memory at address 0x1>) at imap/search_squat.c:320
#3  0x00007f0e122f0c06 in subquery_run_indexed (key=<optimized out>, data=<optimized out>, rock=0x5627e47c0fd0) at imap/search_query.c:717
#4  0x00007f0e11aa79d1 in hash_enumerate (table=table@entry=0x5627e47c1018, func=func@entry=0x7f0e122f0a70 <subquery_run_indexed>, rock=rock@entry=0x5627e47c0fd0) at lib/hash.c:311
#5  0x00007f0e122f0e93 in search_query_run (query=query@entry=0x5627e47c0fd0) at imap/search_query.c:996
#6  0x00007f0e122b2844 in index_search (state=0x5627e477f6b0, searchargs=searchargs@entry=0x5627e477f540, usinguid=usinguid@entry=0) at imap/index.c:1931
#7  0x00005627e276dc60 in cmd_search (tag=0x5627e477b2c0 "6", usinguid=0) at imap/imapd.c:6080
#8  0x00005627e2785380 in cmdloop () at imap/imapd.c:1969
#9  0x00005627e278b8b8 in service_main (argc=<optimized out>, argv=<optimized out>, envp=<optimized out>) at imap/imapd.c:999
#10 0x00005627e2779c7d in main (argc=<optimized out>, argv=<optimized out>, envp=0x7fffd14456a8) at master/service.c:647 

This seems to go away when rerunning squatter

swegener:
When searching for flags, the can_match function from the search engine is called with SEARCH_PART_NONE and squat wrongfully indicates that it can handle it.

As a result, the match function from the search engine then gets called with the char pointer from the search_value union, while the search_value union contains an integer for the search_flags_match handler. The char pointer is invalid and causes a segmentation fault.

Closes: https://github.com/cyrusimap/cyrus-imapd/issues/3562

### DIFF
--- a/imap/search_squat.c
+++ b/imap/search_squat.c
@@ -1013,7 +1013,7 @@ static int end_update(search_text_receiver_t *rx)
 static int can_match(enum search_op matchop, enum search_part partnum)
 {
     return (matchop == SEOP_MATCH || matchop == SEOP_FUZZYMATCH) &&
-        doctypes_by_part[partnum];
+        partnum != SEARCH_PART_NONE && doctypes_by_part[partnum];
 }
 
 const struct search_engine squat_search_engine = {


elliefm (approved):
Looks good, thanks!

PR #5808, opened 2026-02-16
Remove legacy JMAP Contacts code and tests (old API using vparse.c)

ksmurchison:
This also does some code cleanup.  This is best reviewed commit-by-commit, although removing the legacy code from jmap_contact.c left a large and sometimes confusing diff.

**DO NOT MERGE until all FM use of the old API has been deprecated**


rsto:
@ksmurchison Is this already ready for review? It's failing CI and we don't want to merge this now, right?

ksmurchison:
Damn!  I Need to fix JMAPBackup too

imap/jmap_contact.c
```diff
@@ -956,6 +874,7 @@ static int _contacts_changes(struct jmap_req *req, int kind)
     return 0;
 }
 
+#if 0
```
rsto: Why keep this?

rsto (commented):
This is my initial review. This PR is huge and with the changes especially in jmap_contact.c I find it hard to look through the line noise. It's likely I missed something, but overall this looks good. I'm especially happy that this patch also removes `vparse`.

My concerns mostly are around what we lose by throwing away Cassandane tests. It's clear that we want to get rid of the tests that assert specifics of the legacy API. But some of these tests call the legacy API only to validate how Cyrus deals with data or scenarios that will show up one way or the other also in the new API.

In the JMAPContacts suite that's notably the `contact_get_invalid_utf8` test which surely was introduced because we have this data on disk somewhere, but now don't assert how Cyrus deals with it. Similary, I wonder if the `_shared` tests of the new API are as extensive as the legacy API.

More concerning to me looks to throw away anything contacts-related in the JMAPEmail test suite. Don't we keep using the same filter criteria? Why would these test change at all?

rjbs (approved):

rjbs:
There are failures in:
* JMAPContacts.card_set_media_from_deleted_card
* JMAPContacts.card_set_media_blob_shared

…in CI, but I can't reproduce this locally.  Maybe you can @ksmurchison ?

imap/jmap_contact.c
ksmurchison: Oops!  Fixed

PR #5815, opened 2026-02-20
Convert everything to urn:ietf:params:jmap:blob

brong:
RFC9404 has been out for years, we don't need to support our stale old API any more

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/blob_get
+++ b/cassandane/tiny-tests/JMAPEmail/blob_get
@@ -30,11 +30,11 @@ sub test_blob_get
     $jmap->DefaultUsing(\@using);
 
     $res = $jmap->CallMethods([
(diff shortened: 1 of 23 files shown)


ksmurchison (commented):
Looks like JMAPEmail.blob_get needs to use `typeNames` instead of `types`.  
Otherwise LGTM.

ksmurchison (commented):
Looks like JMAPEmail.blob_get is still trying to parse the old-style response for Blob/lookup

cassandane/tiny-tests/JMAPEmail/blob_get
ksmurchison: I think this should be $blob->{matchedIds}{Mailbox}

ksmurchison (approved):

PR #5831, opened 2026-02-27
perl is mandatory

elliefm:
**Reviewers:** I'd suggest reading this commit by commit so that each change can be seen in its own context, rather than all intermingled.  Please also pull this branch and test it in your environment.

This PR: 

* Makes perl mandatory, with a minimum version of 5.28.0.  `--without-perl` or `--with-perl=no` are no longer valid configure options.  If no suitable perl is found, configure fails.  The default perl is the perl that was in your `$PATH` at the time configure was run.
* You can choose a different perl in the usual ways, e.g. by one of `./configure PERL=/path/to/perl [...]` or `./configure --with-perl=PATH [...]`.  (The former instructs configure to use _exactly that perl_, the latter instructs configure to look for perl in the specified PATH instead of `$PATH`.)
* All parts of Cyrus should now use the perl that was chosen by configure.  If I've missed any, that's a mistake, but it's hard to detect.  (Note that I have not touched _Cassandane_ in this PR, it still does whatever it always did.)
* Fixes some old bugs in the perl handling that I tripped over while researching/testing these changes

The `JUNK fixsearchpath debug` commit is useful for testing my changes to the tools/fixsearchpaths.pl boilerplate.  If you run the installed `cyradm` or `sieveshell` scripts, they will tell you where they expect to be, where they actually are, and what DESTDIR they inferred from that.  **I plan to drop this commit before merging.**


elliefm:
(Rebased on master and updated to accommodate #5786's changes)

rsto (commented):
I'll leave approval to reviewers with stronger opinions on Perl dependency management. But this branch builds and runs without issue on my setup.

Makefile.am
```diff
@@ -686,7 +686,7 @@ CUNIT_INTERMEDIATES = $(patsubst %.testc,%.c,$(cunit_TESTS))
 # if something depending on them is being rebuilt.
 .SECONDARY: $(CUNIT_INTERMEDIATES)
 
-CUNIT_PL = $(top_srcdir)/cunit/cunit.pl --srcdir $(abs_srcdir)
+CUNIT_PL = $(PERL) $(top_srcdir)/cunit/cunit.pl --srcdir $(abs_srcdir)
```
rjbs, line 696: Might suggest adding a tiny bit of detail to the commit message:  "Without this, the shebang wins, because the code hasn't been rewritten/installed yet, so it could use the wrong perl.  Even with +x and a shebang, we need to invoke our perl explicitly".

perl/imap/cyradm.sh
```diff
@@ -1,9 +1,10 @@
-#! /bin/sh
-exec perl -x -S $0 ${1+"$@"} # -*-perl-*-
```
rjbs, line 2: seeing lines like this being deleted, my day has gotten 1% better

tools/fixsearchpath.pl
```diff
@@ -78,9 +78,10 @@ BEGIN
 # Filter stdin to stdout
 while (<STDIN>)
 {
-    if (defined $boilerplate && m/^use\s/) {
+    if (defined $boilerplate && m/^use\s+(?!strict|warnings)/) {
```
rjbs, line 108: Could you say — either in a comment here or by adding a comment to `fixsearchpath.pl` — just what this tool is *for*?

This change seems okay, but this tool makes me grimace a little.  I wonder whether we can do something less "yow!" here.

rjbs (approved):
This seems decent, as far as I can tell.  It works.  It worked before.  I know you tested some of the useful cases this might fix.  The level of autoconf/EUMM stuff here is *at best* right below the threshold of maximum understanding for me, and honestly a bit past it here and there.  I think it makes some things clearer, and doesn't seem to make anything worse.  This is a big "mostly trusting ellie here" review, but I do think it all seems sensible.

tools/fixsearchpath.pl
elliefm: This tool also makes me grimace.  I don't think I entirely understand its purpose, and it's possible I'm breaking something I don't know about by changing it, which is why I'm making small tweaks to what's there rather than substantial refactoring.  I'll add a comment documenting my understanding, and that'll be an improvement, even if it's partially wrong.

elliefm: Added a comment!

Makefile.am
elliefm: Updated the commit message

elliefm:
I think I'll split "make mkimap properly installable" off as a separate task, and not mess with it as part of this one.  There's already a lot of moving pieces here, and it'll require a synchronised cyrus-docker change (and probably FM build scripts too) to remove the dodgy manual `cp`.

wolfsage:
> Ahh:
> 
> https://github.com/cyrusimap/cyrus-docker/blob/67340ab1fe65165ef481d91c59a30d94e4c32e19/Debian/lib/Cyrus/Docker/Command/build.pm#L73
> 
> > `run(qw( sudo cp tools/mkimap /usr/cyrus/bin/mkimap ));`
> 
> Hmm. Yeah well that won't work.

Ah! My bad, great.

wolfsage (approved):
Seems good to me!

wolfsage (changes requested):
(Fix up the calls to a few tools using the wrong perl as noted and I think this is good)

rjbs (approved):
My approval here means "Once the stuff Matthew called out is sorted, 👍🏽 "

elliefm:
> Ha, okay I replaced /usr/bin/perl with this, and moved /usr/bin/perl to /usr/bin/realperl:
> [...]
> And captured this:

Oh, fantastic, thanks!  I'd been pondering a similar strategy, but without a second perl I trusted, felt like I would just wind up debugging the wrong thing.

wolfsage (approved):
lgtm!

dilyanpalauzov:
Is now `ExtUtils::MakeMaker` also mandatory, as the perl/ directories are handled unconditionally?

Why, when I run `./configure`,  I get?
```
configure: creating ./config.status
config.status: creating perl/sieve/managesieve/MANIFEST
config.status: creating perl/imap/MANIFEST
config.status: creating perl/annotator/MANIFEST
config.status: creating tools/mkimap
config.status: creating tools/rehash
config.status: creating libcyrus_imap.pc
config.status: creating libcyrus_min.pc
config.status: creating libcyrus.pc
config.status: creating libcyrus_sieve.pc
config.status: creating Makefile
config.status: creating perl/annotator/build.cfg
config.status: creating perl/imap/build.cfg
config.status: creating perl/sieve/managesieve/build.cfg
config.status: creating config.h
config.status: executing depfiles commands
config.status: executing libtool commands
config.status: executing perl/annotator/Makefile commands
Checking if your kit is complete...
Looks good
Only one of PREFIX or INSTALL_BASE can be given.  Not both.
config.status: error: failed to generate Makefile for perl/annotator
```
I have
```
$ cpan -l|grep -i makema
ExtUtils::MakeMaker     7.76
ExtUtils::MakeMaker::CPANfile   0.09
ExtUtils::MakeMaker::Config     7.76
vendor_perl::ExtUtils::MakeMaker        7.76
vendor_perl::ExtUtils::MakeMaker::CPANfile      0.09
vendor_perl::ExtUtils::MakeMaker::Config        7.76
```

PR #5846, opened 2026-03-11
jmap_calendar.c: support compact IDs for JMAP Calendars

ksmurchison:
Uses an encoded form of mailbox createdmodseq as the ID

### DIFF
--- a/cassandane/Cassandane/TestEntity/DataType/Calendar.pm
+++ b/cassandane/Cassandane/TestEntity/DataType/Calendar.pm
@@ -31,8 +31,26 @@ true C<isDefault> property.
 
     sub default {
         my ($self) = @_;
+        my $dt = $self->datatype;
 
-        return $self->get('Default');
+        my $jmap = $self->user->entity_jmap;
+
+        my $res = $jmap->request([[ "Calendar/get", {} ]]);
+
+        my $get = $res->single_sentence('Calendar/get');
+
+        my @objs = $get->arguments->{list}->@*;
+        @objs > 0 || Carp::confess("user has no Calendars");
+
+        my ($default, @extra) = grep {; $_->{isDefault} } @objs;
+        $default || Carp::confess("user has no default Calendar");
+        @extra && Carp::confess("user has more than one default Calendar");
+
+        $self->instance_class->new({
+            id  => "$objs[0]{id}",
+            factory    => $self,
+            properties => $objs[0],
+        })
     }
 
     use Cassandane::TestEntity::AutoSetup;
--- a/cassandane/tiny-tests/CaldavAlarm/recurring_allday_floating
+++ b/cassandane/tiny-tests/CaldavAlarm/recurring_allday_floating
@@ -23,8 +23,9 @@ END:VTIMEZONE
 END:VCALENDAR
 EOF
 
-    my $CalendarId = $CalDAV->NewCalendar({name => 'foo', timeZone => $UTC});
-    $self->assert_not_null($CalendarId);
+    my $default_calendar = $self->default_user->calendars->default;
+    my $CalendarId = $default_calendar->id;
+    my $CalendarHref = $default_calendar->properties->{'x-href'} . "";
 
     my $now = DateTime->today();
     $now->set_time_zone('Etc/UTC');
@@ -38,7 +39,7 @@ EOF
     my $trigger="-PT5H";
 
     my $uuid = "574E2CD0-2D2A-4554-8B63-C7504481D3A9";
-    my $href = "$CalendarId/$uuid.ics";
+    my $href = "$CalendarHref/$uuid.ics";
     my $card = <<EOF;
 BEGIN:VCALENDAR
 VERSION:2.0
@@ -103,7 +104,7 @@ EOF
 EOF
 
     my $res = $CalDAV->Request('PROPPATCH',
-                               "/dav/calendars/user/cassandane/". $CalendarId,
+                               $CalendarHref,
                                $xml, 'Content-Type' => 'text/xml');
 
     $now->add(days => 7);
@@ -148,7 +149,7 @@ EOF
 EOF
 
     $res = $CalDAV->Request('PROPPATCH',
-                            "/dav/calendars/user/cassandane/". $CalendarId,
+                            $CalendarHref,
                             $xml, 'Content-Type' => 'text/xml');
 
     $now->add(days => 7);
--- a/cassandane/tiny-tests/Sieve/imip_invite_calendarid
+++ b/cassandane/tiny-tests/Sieve/imip_invite_calendarid
@@ -17,7 +17,7 @@ sub test_imip_invite_calendarid
 
     my $uuid = "6de280c9-edff-4019-8ebd-cfebc73f8201";
 
-    xlog $self, "Install a sieve script to process iMIP";
+    xlog $self, "Install a sieve script to process iMIP using legacy calendar id";
     $self->{instance}->install_sieve_script(<<EOF
 require ["body", "variables", "imap4flags", "processcalendar"];
 if body :content "text/calendar" :contains "\nMETHOD:" {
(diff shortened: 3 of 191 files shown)


imap/lmtp_sieve.c
```diff
@@ -1679,6 +1680,27 @@ static int sieve_processcal(void *ac, void *ic, void *sc, void *mc,
         goto done;
     }
 
+    const char *calendarid = cal->calendarid;
+    if (calendarid && USER_COMPACT_EMAILIDS(ctx->cstate)) {
```
dilyanpalauzov: I suggest moving the declariton of `freeme` lower in the code, just before it is used, likewise moving `free(freeme)` upper in the code; and rewriting
```c
if (calendarid && USER_COMPACT_EMAILIDS(ctx->cstate)) {
   if (calendarid[0] == JMAP_CALENDARID_PREFIX) {
```
as
```c
if (calendarid && USER_COMPACT_EMAILIDS(ctx->cstate) && calendarid[0] == JMAP_CALENDARID_PREFIX) {
```

In fact after the assignment `calendarid = freeme = xstrdup(strarray_nth(mbname_boxes(mbname), -1));` I do not see that the value of `calendarid` is read.

imap/jmap_calendar.c
```diff
@@ -871,10 +878,51 @@ static int has_calendars(jmap_req_t *req)
...
+        idtype = "name";
+        r = mboxlist_lookup(mboxname, mbentry, NULL);
+        free(mboxname);
+    }
+
+    if (r) {
+        if (r != IMAP_NOTFOUND) {
+            syslog(LOG_ERR,
+                   "Error reading mbentry for addressbook via %s %s: %s",
+                   idtype, id, error_message(r));
```
rsto: The `idtype` variable only seems to be used in case of error reporting ? Could get rid of it and replace it here with `USER_COMPACT_EMAILIDS(req->cstate) ? "jmapid" : "name".

imap/jmap_calendar.c
```diff
@@ -894,7 +942,11 @@ static int jmap_calendar_get(struct jmap_req *req)
     }
 
     /* Build callback data */
-    struct getcalendars_rock rock = { req, &get, 1 /*skiphidden*/ };
+    default_calname = caldav_scheddefault(req->accountid, 1);
+    default_cal_mboxname = caldav_mboxname(req->accountid, default_calname);
+    free(default_calname);
```
rsto: suggest to use `xzfree` when freeing memory in the middle of the code.

rsto (approved):
I only reviewed the changes in the C code. These look good to me, apart from a few nits. I would prefer the commit message to change, though.

It's not a hard rule because every contributor seems to do this slightly differently, but I would prefer an action in the commit message to make clear what's happening, in this case "*support* compactids for *JMAP* calendars". On the other hand, the "( 'C' + b64j(cmodseq) )" is too much detail for the commit message, maybe not even relevant for the commit description.

imap/jmap_calendar.c

imap/jmap_calendar.c

imap/lmtp_sieve.c
ksmurchison: You're right.  I will fix if this code if it stays.  I need to figure out if we will even accept the JMAP Calendar IDs for Sieve `processcal :calendarid`

ksmurchison: Fixed

cassandane/Cassandane/TestEntity/DataType/Calendar.pm
```diff
@@ -31,8 +31,28 @@ true C<isDefault> property.
...
+        my $get = $res->single_sentence('Calendar/get');
+
+        my @objs = $get->arguments->{list}->@*;
+        @objs > 0 || Carp::confess("user has no Calendars");
+
+        my ($default) = grep {; $_->{isDefault} } @objs;
+        @objs > 0 || Carp::confess("user has no default Calendar");
+
+        my ($default) = grep {; $_->{isDefault} } @objs;
+        @objs < 2 || Carp::confess("user has multiple default Calendars");
```
rjbs: We can change this later, but I would write:

```perl
my ($default, @extra) = grep {; $_->{isDefault} } @objs;
$default || Carp::confess("user has no default Calendar");
@extra && Carp::confess("user has more than one default Calendar");
```

This isn't a blocker.

cassandane/tiny-tests/CaldavAlarm/recurring_allday_floating
```diff
@@ -23,8 +23,9 @@ END:VTIMEZONE
 END:VCALENDAR
 EOF
 
-    my $CalendarId = $CalDAV->NewCalendar({name => 'foo', timeZone => $UTC});
-    $self->assert_not_null($CalendarId);
+    my $CalendarId = $self->default_user->calendars->default->id;
+    my $CalendarHref =
+        $self->default_user->calendars->default->properties->{'x-href'} . "";
```
rjbs: In these cases, you'll want to get the default calendar once, then call two things on it, like:

```perl
my $default_calendar = $self->default_user->calendars->default;
my $calendar_id = $default_calendar->id;
my $href = $default_calendar->properties->{'x-href'};
```

Otherwise, each call to ->default does another JMAP request, it's not cached.

rjbs (approved):
Looks good.  Probably worth looking at the places where default calendar is fetched >1 time, for efficiency, but nothing seems wrong.

cassandane/Cassandane/TestEntity/DataType/Calendar.pm
ksmurchison: Ha!  The original code was yours, but I updated to your preference.

cassandane/tiny-tests/CaldavAlarm/recurring_allday_floating
ksmurchison: I fixed all of these instances as well as any test whiuch fetched the default calendar id but didn' t actually use it.  I assume the latter was an artifact of some automated tooling

imap/jmap_calendar.c
```diff
@@ -2227,44 +2245,43 @@ static int jmap_calendar_set(struct jmap_req *req)
...
-        char *calhome_name = caldav_mboxname(req->accountid, NULL);
-        annotate_state_t *calhome_astate = NULL;
-        struct mailbox *calhome_mbox = NULL;
+        /* make sure new default calendar exists */
+        mbentry_t *mbentry = NULL;
+        calid_to_mbentry(req, newid, &mbentry);
+        if (mbentry &&
+            jmap_hasrights_mbentry(req, mbentry, JACL_ADMIN_CALENDAR)) {
+            /* set CALDAV:schedule-default-calendar annotation */
+            const char *annot =
```
dilyanpalauzov: I suggest either changing the type to `static const char[] annot = DAV_ANNOT_NS "<" XML_NS_CALDAV ">schedule-default-calendar";`, so that the variable goes into read-only segment, or event better,  replace
```c
r = annotatemore_writemask(calhome_mboxname, annot, req->accountid, &buf);
```
with
```c
 r = annotatemore_writemask(calhome_mboxname, DAV_ANNOT_NS "<" XML_NS_CALDAV ">schedule-default-calendar", req->accountid, &buf);
```

`annot` is used once, so I do not see a need to create a variable, which is used only once.

dilyanpalauzov (commented):
In imap/jmap_calendar.c:getcalendars_cb():
```
   char id[JMAP_MAX_ADDRBOOKID_SIZE];
   jmap_set_calendarid(req->cstate, mbentry, id);
```

and jmap_util.c:jmap_set_calendarid(…char *calid) does:
```
strlcpy(calid, mbentry->jmapid, JMAP_MAX_CALENDARID_SIZE);
…
strlcpy(calid, id, JMAP_MAX_CALENDARID_SIZE);
```
By coincidence `JMAP_MAX_CALENDARID_SIZE` is the same as `JMAP_MAX_ADDRBOOKID_SIZE`.  I suggest adding `nonnull` attributes around the declaration of `jmap_set_calendarid()` and spelling in jmap_util.h with words, what the size of `calid` is expected to be.

I suggest unifying JMAP_MAX_ADDRBOOKID_SIZE and JMAP_MAX_CALENDARID_SIZE into a single macro.

cassandane/tiny-tests/Sieve/imip_invite_calendarid
```diff
@@ -70,4 +70,48 @@ EOF
     my $events = $CalDAV->GetEvents($CalendarId);
     $self->assert_equals(1, scalar @$events);
     $self->assert_str_equals($uuid, $events->[0]{uid});
+
+    my $jmap = $self->{jmap};
+    my $res = $jmap->CallMethods([['Calendar/get', {}, "R1"]]);
+
+    my %calendars = map { $_->{name} => $_ } @{$res->[0][1]{list}};
+    my $jmapid = $calendars{'foo'}{id};
```
wolfsage, line 78: We should assert we got a $jmapid here

ksmurchison: Done

rsto:

PR #5848, opened 2026-03-11
fix racl group key lookups

zisisvakras:
This is a fix for #5847 but as I said in my issue I am not quite sure if this won't break any other auth mechs

### DIFF
--- a/cassandane/Cassandane/Cyrus/UnixGroups.pm
+++ b/cassandane/Cassandane/Cyrus/UnixGroups.pm
@@ -40,18 +40,22 @@ sub set_up
     $self->SUPER::set_up();
     $self->{instance}->create_user("otheruser");
 
+    $self->_discover_unix_groups();
+}
+
+sub _discover_unix_groups
+{
+    my ($self) = @_;
+
     my $userid = 'cassandane';
     my @member_groups;
     my %seen;
 
     my @pw = getpwnam($userid);
+    my $primary_group;
     if (@pw) {
         my $primary_gid = $pw[3];
-        my $primary_group = getgrgid($primary_gid);
-        if (defined $primary_group) {
-            push @member_groups, $primary_group;
-            $seen{$primary_group} = 1;
-        }
+        $primary_group = getgrgid($primary_gid);
     }
 
     setgrent();
@@ -63,6 +67,13 @@ sub set_up
         $seen{$name} = 1;
     }
     endgrent();
+
+    # Prefer supplementary groups for group ACL tests: they exercise reverse
+    # ACL updates more reliably than the account's primary group name.
+    if (!@member_groups && defined $primary_group) {
+        push @member_groups, $primary_group;
+        $seen{$primary_group} = 1;
+    }
 
     my $nonmember_group;
     setgrent();
@@ -82,6 +93,11 @@ sub set_up
 sub skip_check
 {
     my ($self) = @_;
+
+    # skip_runtime_check is evaluated before set_up(), so ensure these are
+    # populated before applying prerequisite checks.
+    $self->_discover_unix_groups()
+        if not defined $self->{unix_member_groups};
 
     if ($self->{_name} =~ /list_order_racl/) {
         return "requires at least two unix groups containing user 'cassandane'"
--- a/cassandane/Cassandane/Cyrus/UnixGroups.pm
+++ b/cassandane/Cassandane/Cyrus/UnixGroups.pm
@@ -1,11 +1,17 @@
 #!/usr/bin/perl
 
+# Test environment prerequisites:
+# - A local user named "cassandane" must exist.
+# - At least two unix groups must include "cassandane" as a member.
+# - At least one unix group must NOT include "cassandane".
+#
+# These tests derive dynamic group names from the host system at runtime.
+
 package Cassandane::Cyrus::UnixGroups;
 use strict;
 use warnings;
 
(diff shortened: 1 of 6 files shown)


cassandane/Cassandane/Cyrus/UnixGroups.pm
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/perl
+
+package Cassandane::Cyrus::UnixGroups;
+use strict;
+use warnings;
+
+use base qw(Cassandane::Cyrus::TestCase);
+use base qw(Cassandane::Unit::TestCase);
+use Cassandane::Util::Log;
```
elliefm, line 15: You only need one base class here.

```suggestion

use base qw(Cassandane::Cyrus::TestCase);
use Cassandane::Util::Log;
```

cassandane/Cassandane/Cyrus/UnixGroups.pm
```diff
@@ -0,0 +1,178 @@
...
+        auth_mech => 'unix',
+        auth_unix_group_enable => 'yes',
+    );
+
+    my $self = $class->SUPER::new({
+        config => $config,
+        adminstore => 1,
+        services => [qw( imap )],
+        start_instances => 0,
+    }, @args);
```
elliefm, line 31: Looks like the first thing you do in `set_up()` is start the instances anyway, so I think there's no reason to request they're not started here.  I guess you based these tests on the LDAP ones.  They need this mechanism because they start a subprocess and need to put its PID into the config before Cyrus starts, but you're not doing any of that.

```suggestion
    my $self = $class->SUPER::new({
        config => $config,
        adminstore => 1,
        services => [qw( imap )],
    }, @args);
```

cassandane/Cassandane/Cyrus/UnixGroups.pm
```diff
@@ -0,0 +1,178 @@
...
+    return $self;
+}
+
+sub set_up
+{
+    my ($self) = @_;
+
+    $self->SUPER::set_up();
+
+    $self->_start_instances();
```
elliefm: As per earlier comment, you don't need the explicit start handling.

cassandane/Cassandane/Cyrus/UnixGroups.pm
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/perl
+
+package Cassandane::Cyrus::UnixGroups;
+use strict;
```
elliefm, line 11: It looks like this test suite depends on having particular users and groups set up on the system it's running on.  It's a bit strange to depend on the system, but I can see why it's necessary for this in particular.  Please add a comment after the `#!` line describing what users/groups/memberships are expected to exist for people who want to run these tests.

cassandane/Cassandane/Cyrus/UnixGroups.pm
```diff
@@ -0,0 +1,178 @@
...
+    my %adminfolders = (
+        'user.otheruser.order-user' => 'cassandane',
+        'user.otheruser.order-co' => $group_co,
+        'user.otheruser.order-c' => $group_c,
+        'user.otheruser.order-o' => $group_o,
+        'shared.order-co' => $group_co,
+        'shared.order-c' => $group_c,
+        'shared.order-o' => $group_o,
+    );
+
```
elliefm: It's not super obvious, but in the LDAP tests you're mimicking, the "c", "o", "co" in group and mailbox names are specifically describing the expected group memberships/mailbox permissions.  That is:

* "c" -> contains (is accessible by) the user "**c**assandane", but not the user "**o**theruser"
* "o" -> contains (is accessible by) the user "**o**theruser", but not the user "**c**assandane"
* "co" -> contains (is accessible by) both "**c**assandane" and "**o**theruser" users

The idea here is that if the test fails, someone looking at the output can see what was _supposed_ to happen from the names.  e.g. if the "cassandane" user does a `LIST` and sees a mailbox called "foo-o", something is wrong, because foo-o should only be accessible to "otheruser".

If I'm reading this implementation right, you're only differentiating between groups that do and don't include "cassandane".  So please name your groups/mailboxes in a way that makes sense in that context.

elliefm (changes requested):
Thanks.  The C changes look okay to me (if they work), but I have some comments on the tests.

Also, please drop the commit that merges master into your branch.  Backwards merges like this make a mess of the history.  To bring your branch up to date with current master, please use rebase instead of merge.  Other projects may have different preferences, but this is how we do it.

zisisvakras:
Hello, thanks a lot for the detailed review.

For the C part: After the tests failed I checked the code for the other auth mechs and noticed that they keep the prefix in their internal structure and UNIX was the outlier. I updated UNIX to keep group identifiers in group:-prefixed form internally, which aligns with the RACL lookup expectations and fixes the mismatch I was seeing.

For Cassandane: I noticed that all the other mechs did in fact have tests for this and I wanted to somehow copy them to UNIX. I guess its not as straightforward as I thought, so I kind of hacked it using the host's groups, I am not really sure if there is a better way to do it. I have applied your comments to the code, please verify that they are fine, the naming scheme also should be better. I am very new to cassandane so any more comments are very welcome.

zisisvakras:
I updated the patch to include support for raclmodseq. Now auth_newstate should also return the userids as required. I also fixed the skip_check on cassandane to populate the discovered groups before checking the prerequisites (I didn't understand the execution order before). This has run for about 2 weeks on our setup without any problems.

PR #5853, opened 2026-03-16
jmap: make MIME boundaries deterministic for Email/set create

brong:
Replace random UUID-based MIME boundaries with deterministic ones derived from a SHA1 hash of the canonicalized input JSON properties. This ensures that creating the same Email twice via Email/set produces byte-identical RFC 822 output with the same blobId.

This is important for idempotent retries: if a client retries an Email/set create after a network failure, and the first attempt actually succeeded, the second attempt will produce the same blob and return alreadyExists with the existing id, rather than creating a duplicate message with different MIME boundaries.

The boundary generation uses a struct boundary_gen containing a hash base string and a counter, threaded through all functions that create multipart MIME structures:
- _emailpart_new_multi (body structure building)
- _emailpart_write_headers (header generation)
- _emailpart_to_mime (MIME output)
- _emailpart_body_to_mime (leaf part output)

The hash base is computed once in _email_create from json_dumps(jemail, JSON_COMPACT|JSON_SORT_KEYS), ensuring canonical key ordering. Each boundary is then SHA1(base + counter) encoded as hex, which is safe against an attacker trying to craft content that matches the boundary (SHA1 second preimage resistance).

The base string is xstrdup'd to avoid the static buffer issue with message_guid_encode (which returns a pointer to a single static buffer that would be overwritten by subsequent calls).

The random boundary fallback is preserved for code paths that don't have a boundary_gen (e.g. if the struct is zero-initialized).

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_set_create_deterministic_boundaries
+++ b/cassandane/tiny-tests/JMAPEmail/email_set_create_deterministic_boundaries
@@ -1,0 +1,113 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_set_create_deterministic_boundaries
+    :min_version_3_1 :needs_component_sieve
+    ($self)
+{
+    my $jmap = $self->{jmap};
+
+    xlog $self, "Create a multipart email with text and html bodies";
+    my $res = $jmap->CallMethods([
+        ['Email/set', { create => {
+            "a" => {
+                mailboxIds => { '$inbox' => JSON::true },
+                from => [{ name => "Test", email => 'test@example.com' }],
+                to => [{ name => "Dest", email => 'dest@example.com' }],
+                subject => "Deterministic boundary test",
+                "header:Message-Id" => '<test1234@example.com>',
+                "header:Date" => 'Mon, 01 Jan 2024 00:00:00 +0000',
+                bodyStructure => {
+                    type => 'multipart/alternative',
+                    subParts => [{
+                        type => 'text/plain',
+                        partId => 'text',
+                    }, {
+                        type => 'text/html',
+                        partId => 'html',
+                    }],
+                },
+                bodyValues => {
+                    text => { value => "Hello world" },
+                    html => { value => "<p>Hello world</p>" },
+                },
+            },
+        }}, "R1"],
+    ]);
+    my $emailId1 = $res->[0][1]{created}{a}{id};
+    my $blobId1 = $res->[0][1]{created}{a}{blobId};
+    $self->assert_not_null($emailId1);
+    $self->assert_not_null($blobId1);
+
+    xlog $self, "Delete the email";
+    $res = $jmap->CallMethods([
+        ['Email/set', { destroy => [$emailId1] }, "R2"],
+    ]);
+    $self->assert_str_equals($emailId1, $res->[0][1]{destroyed}[0]);
+
+    xlog $self, "Create the exact same email again - must produce same blobId";
+    $res = $jmap->CallMethods([
+        ['Email/set', { create => {
+            "b" => {
+                mailboxIds => { '$inbox' => JSON::true },
+                from => [{ name => "Test", email => 'test@example.com' }],
+                to => [{ name => "Dest", email => 'dest@example.com' }],
+                subject => "Deterministic boundary test",
+                "header:Message-Id" => '<test1234@example.com>',
+                "header:Date" => 'Mon, 01 Jan 2024 00:00:00 +0000',
+                bodyStructure => {
+                    type => 'multipart/alternative',
+                    subParts => [{
+                        type => 'text/plain',
+                        partId => 'text',
+                    }, {
+                        type => 'text/html',
+                        partId => 'html',
+                    }],
+                },
+                bodyValues => {
+                    text => { value => "Hello world" },
+                    html => { value => "<p>Hello world</p>" },
+                },
+            },
+        }}, "R3"],
+    ]);
+    my $blobId2 = $res->[0][1]{created}{b}{blobId};
+    $self->assert_not_null($blobId2);
+
+    xlog $self, "Same input must produce same blobId (identical RFC822 output)";
+    $self->assert_str_equals($blobId1, $blobId2);
+
+    xlog $self, "Create email with different body content";
+    $res = $jmap->CallMethods([
+        ['Email/set', { create => {
+            "c" => {
+                mailboxIds => { '$inbox' => JSON::true },
+                from => [{ name => "Test", email => 'test@example.com' }],
+                to => [{ name => "Dest", email => 'dest@example.com' }],
+                subject => "Deterministic boundary test",
+                "header:Message-Id" => '<test1234@example.com>',
+                "header:Date" => 'Mon, 01 Jan 2024 00:00:00 +0000',
+                bodyStructure => {
+                    type => 'multipart/alternative',
+                    subParts => [{
+                        type => 'text/plain',
+                        partId => 'text',
+                    }, {
+                        type => 'text/html',
+                        partId => 'html',
+                    }],
+                },
+                bodyValues => {
+                    text => { value => "Hello different world" },
+                    html => { value => "<p>Hello different world</p>" },
+                },
+            },
+        }}, "R4"],
+    ]);
+    my $blobId3 = $res->[0][1]{created}{c}{blobId};
+    $self->assert_not_null($blobId3);
+
+    xlog $self, "Different content must produce different blobId";
+    $self->assert_str_not_equals($blobId1, $blobId3);
+}
--- a/cassandane/tiny-tests/JMAPEmail/email_set_create_deterministic_boundaries
+++ b/cassandane/tiny-tests/JMAPEmail/email_set_create_deterministic_boundaries
@@ -2,44 +2,45 @@
 use Cassandane::Tiny;
 
 sub test_email_set_create_deterministic_boundaries
-    :min_version_3_1 :needs_component_sieve
+    :needs_component_sieve
     ($self)
 {
     my $jmap = $self->{jmap};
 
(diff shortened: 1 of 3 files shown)


imap/jmap_mail.c
```diff
@@ -8735,16 +8735,30 @@ static int jmap_email_parse(jmap_req_t *req)
     return 0;
 }
 
-static char *_mime_make_boundary()
+struct boundary_gen {
+    const char *base;  /* hash of Email/set properties, or NULL for random */
```
ksmurchison: Why not make this char* and free it in _email_fini() rather than explicitly freeing the value created in _email_create()?

imap/jmap_mail.c
```diff
@@ -8735,16 +8735,30 @@ static int jmap_email_parse(jmap_req_t *req)
...
+    int counter;       /* incremented for each boundary */
```
ksmurchison: Any reason to not make this unsigned?

imap/jmap_mail.c
```diff
@@ -8735,16 +8735,30 @@ static int jmap_email_parse(jmap_req_t *req)
...
+};
+
+static char *_mime_make_boundary(struct boundary_gen *gen)
 {
-    char *boundary, *p, *q;
+    if (gen && gen->base) {
+        /* deterministic boundary from hash of Email/set properties */
+        struct message_guid guid;
+        struct buf buf = BUF_INITIALIZER;
+        buf_printf(&buf, "%s-%d", gen->base, gen->counter++);
```
ksmurchison: Why not add the struct buf to struct boundary_gen and only buf_free() it in _email_fini()?

ksmurchison (changes requested):

cassandane/tiny-tests/JMAPEmail/email_set_create_duplicate_retry
```diff
@@ -0,0 +1,51 @@
...
+        ['Email/set', { create => { "b" => {%emailProps} } }, "R2"],
+    ]);
+
+    xlog $self, "Must get alreadyExists with the existing id";
+    $self->assert_null($res->[0][1]{created}{b});
+    my $err = $res->[0][1]{notCreated}{b};
+    $self->assert_not_null($err);
+    $self->assert_str_equals("alreadyExists", $err->{type});
+    $self->assert_str_equals($emailId, $err->{existingId});
+}
```
ksmurchison, line 53: Just a nit.  I guess Claude isn't aware of  the new JMAP (simplified) response checking that we have started using. E.g.:

```
    $self->assert_cmp_deeply(
        superhashof({
            type       => 'alreadyExists',
            existingId => $emailId
        }),
        $res->[0][1]{notCreated}{b},
    );

```

ksmurchison (approved):
Approved with a comment, assuming CI passes

cassandane/tiny-tests/JMAPEmail/email_set_create_duplicate_retry
brong: I got Claude to read this as well, and it updated the test

cassandane/tiny-tests/JMAPEmail/email_set_create_duplicate_retry
```diff
@@ -0,0 +1,56 @@
...
+
+    xlog $self, "Retry the same create (simulating network retry)";
+    $res = $jmap->CallMethods([
+        ['Email/set', { create => { "b" => {%emailProps} } }, "R2"],
+    ]);
+
+    xlog $self, "Must get alreadyExists with the existing id";
+    my $notCreated = $res->[0][1]{notCreated};
+    $self->assert_not_null($notCreated);
+    $self->assert_not_null($notCreated->{b});
```
ksmurchison: I don't think the two checks above are necessary.  assert_cmp_deeply() should cover both of these

brong: fixed and rebased together

cassandane/tiny-tests/JMAPEmail/email_set_create_deterministic_boundaries
```diff
@@ -0,0 +1,113 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_set_create_deterministic_boundaries
+    :min_version_3_1 :needs_component_sieve
```
rsto: no need for these `min_version` tags anymore

cassandane/tiny-tests/JMAPEmail/email_set_create_deterministic_boundaries
```diff
@@ -0,0 +1,113 @@
...
+
+    xlog $self, "Create a multipart email with text and html bodies";
+    my $res = $jmap->CallMethods([
+        ['Email/set', { create => {
+            "a" => {
+                mailboxIds => { '$inbox' => JSON::true },
+                from => [{ name => "Test", email => 'test@example.com' }],
+                to => [{ name => "Dest", email => 'dest@example.com' }],
+                subject => "Deterministic boundary test",
+                "header:Message-Id" => '<test1234@example.com>',
```
rsto: why are these headers set explicitly, rather than with their `messageId` and `sentAt` convenience properties?

Also: are they actually required for the test? I figured they might be required to produce the same SHA1, but then I realized that the SHA1 only is determined by the Email/set arguments? In which case they seem to not be relevant?

cassandane/tiny-tests/JMAPEmail/email_set_create_duplicate_retry
```diff
@@ -0,0 +1,53 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_set_create_duplicate_retry
+    :min_version_3_1 :needs_component_sieve
```
rsto: `min_version` not required

imap/jmap_mail.c
```diff
@@ -11250,9 +11275,20 @@ static void _email_create(jmap_req_t *req,
...
+    char *json = json_dumps(jemail, JSON_COMPACT|JSON_SORT_KEYS);
+    struct message_guid boundary_guid;
+    message_guid_generate(&boundary_guid, json, strlen(json));
+    free(json);
+
     /* Parse Email object into internal representation */
     struct jmap_parser parser = JMAP_PARSER_INITIALIZER;
-    struct email email = { HEADERS_INITIALIZER, NULL, NULL, { 0 }, 0, NULL };
+    struct email email = { HEADERS_INITIALIZER, NULL, NULL, { 0 }, 0, NULL,
+                           { xstrdup(message_guid_encode(&boundary_guid)),
```
rsto: I'd prefer if this `boundary_gen` structure to had its own `_init` (or INITIALIZER macro) and `_fini` functions. I agree it might be overkill since it's only initialized and freed at one place in the code, but its a pattern that we are following mostly.

imap/jmap_mail.c
```diff
@@ -8735,16 +8735,30 @@ static int jmap_email_parse(jmap_req_t *req)
...
+struct boundary_gen {
+    char *base;            /* hash of Email/set properties, or NULL for random */
+    unsigned counter;      /* incremented for each boundary */
+    struct buf buf;        /* reusable buffer for boundary generation */
+};
+
+static char *_mime_make_boundary(struct boundary_gen *gen)
 {
-    char *boundary, *p, *q;
+    if (gen && gen->base) {
```
rsto: Is `gen` ever supposed to be NULL? It looks as if we only ever call this on the call tree of `_email_to_mime`, which is supposed to initialize the generator for deterministic boundaries?

rsto (commented):
Looks like it gets the job done, but a few nits and I think this code get sanitized a bit more.

cassandane/tiny-tests/JMAPEmail/email_set_create_duplicate_retry
rjbs: In a world where Cassandane is part of cyrus-imapd.git, is there any reason to ever use `min_version` again?  Should we purge it?

cassandane/tiny-tests/JMAPEmail/email_set_create_deterministic_boundaries
brong: Those two items are needed because the server creates them if not present, and will choose a new value each time

cassandane/tiny-tests/JMAPEmail/email_set_create_duplicate_retry
brong: @rjbs the only thing is running latest tests against older versions to confirm back ported fixes, but... maybe not.

cassandane/tiny-tests/JMAPEmail/email_set_create_duplicate_retry
```diff
@@ -0,0 +1,53 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_set_create_duplicate_retry
+    :needs_component_sieve
```
rsto, line 5: nit: this test does not require sieve

cassandane/tiny-tests/JMAPEmail/email_set_create_deterministic_boundaries
```diff
@@ -0,0 +1,73 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_set_create_deterministic_boundaries
+    :needs_component_sieve
```
rsto, line 5: nit: this test does not require sieve

cassandane/tiny-tests/JMAPEmail/email_set_create_deterministic_boundaries
rsto: I figured as much but wondered why those would be required to produce the exact same boundaries. Turns out they are not, the boundaries are the same for both emails also if they are not set in the Email. But of course the two properties must be set to produce the same blobId, which is the real goal of this patch.

rsto (approved):

PR #5873, opened 2026-03-25
tools/coverage-report: a tool for summarizing lcov results

rjbs:
Parses the .info file directly, handling both both the extended FNL/FNA format and traditional FN/FNDA format.

Auto-detects the path prefix to strip (the /srv/cyrus-imapd/ part, for a cyd build).

Prints a per-directory summary header, then per-file rows

* --threshold filters to files below a line coverage percentage
* --subsystem filters to a directory
* --sort orders by line, func, branch, or name
* --functions adds the list of uncovered function names under each file

Typical invocations for future use:

    # Quick overview of weak spots
    tools/coverage-report --threshold 50 coverage.info

    # Deep dive into one subsystem with function names
    tools/coverage-report --subsystem imap --threshold 30 --functions coverage.info

    # Everything, sorted by function coverage
    tools/coverage-report --sort func coverage.info

### DIFF
--- a/tools/coverage-report
+++ b/tools/coverage-report
@@ -1,0 +1,305 @@
+#!/usr/bin/env perl
+use v5.28.0;
+use warnings;
+
+use experimental 'signatures';
+
+# coverage-report -- summarise an lcov .info file for coverage review
+#
+# See --help for usage.
+
+use Getopt::Long::Descriptive;
+
+my ($opt, $usage) = describe_options(
+    '%c %o coverage.info',
+    [ 'subsystem=s'  => 'only show files under this directory (e.g. "imap")' ],
+    [ 'threshold=f'  => 'only show files with line coverage below N% (default: 100)',
+                        { default => 100 } ],
+    [ 'sort=s'       => 'sort by: line (default), func, branch, name',
+                        { default => 'line' } ],
+    [ 'functions'    => 'list uncovered functions for each reported file' ],
+    [ 'include-gperf' => 'include .gperf files (excluded by default)' ],
+    [ 'strip=s'      => 'strip this prefix from source paths (auto-detected if omitted)' ],
+    [ 'help'         => 'show this help message', { shortcircuit => 1 } ],
+);
+
+sub usage_text {
+    return $usage->text . <<'END';
+
+Examples:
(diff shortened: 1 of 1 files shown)


elliefm:
For what it's worth, here's how I currently handle coverage reports for Cyrus: https://github.com/elliefm/cyrus-build-tools/blob/master/cyrus-coverage  I can't remember if I've sent you this before, apologies if I have.  There might be something useful in there, or you might see something I could be doing better.

That said, I haven't looked much at this particular script in a while.  I'll send you a link privately to what I use to manage coverage on my extracurricular projects, which will be a better reflection of my "current" coverage practice, such as it is.

tools/coverage-report
```diff
@@ -0,0 +1,310 @@
...
+for my $path (sort keys %files) {
+    my $display = $path;
+    $display =~ s/^\Q$strip\E// if $strip;
+
+    # subsystem filter
+    if ($subsystem) {
+        next unless $display =~ m{^\Q$subsystem\E(/|$)};
+    }
+
+    next if !$include_gperf && $display =~ /\.gperf$/;
```
nwc10: This logic is repeated, and I don't think that it should be.

I guess it ought to split into a first loop that filters on subsystem, and remembers which files are going to be processed further.

Looking at the implementation, looks like the computed value `$display` can be cached in the hasref that is the value in `%files`

Also, I wonder if one asks Claude to refactor this, whether Claude will "instinctively" do the `grep` (O(n)) before doing the `sort` (O(n log n))

Currently the more expensive thing is first. But for the number of files we have, even reading this comment is more expensive than the optimisation I'm implying :-)

nwc10 (changes requested):
I think the duplicate code needs to de-duplicate.

Other than that, it looks useful. I tried it and it made output (and some of it was interesting already)

I'm not going to spot if it all conforms to our coding standards, etc.

tools/coverage-report
rjbs: I've pushed an amendment which greps once and only sorts when the rows are ready.

nwc10: Thanks

nwc10 (approved):

PR #5876, opened 2026-03-26
caldav_alarm.c:  Limit recurrence expansion to 90 days from today.

ksmurchison:
process_alarm_cb() will stop expanding recurrences if the trigger for a VALARM is in the future more than 60 days from today. However, if the master component does not include any VALARMS the trigger check is not done and recurrence expansion would otherwise continue until the libical-specified MAX year.

### DIFF
--- a/imap/caldav_alarm.c
+++ b/imap/caldav_alarm.c
@@ -981,11 +981,14 @@ static time_t process_alarms(const char *mboxname, uint32_t imap_uid,
     struct get_alarm_rock rock =
         { userid, mboxname, imap_uid, floatingtz, lastrun, runtime, 0, dryrun };
     struct icalperiodtype range = icalperiodtype_null_period();
-    /* Limit our recurrence expansion to a decade from today.
-       This will end up missing events that occur less frequent than
-       decennially, but what are the chances? */
+    /* Limit our recurrence expansion to 90 days from today.
+       process_alarm_cb() will stop expanding recurrences if the trigger
+       for a VALARM is in the future more than 60 days from today.  
+       However, if the master component does not include any VALARMS
+       the trigger check is not done and recurrence expansion would
+       otherwise continue until the libical-specified MAX year. */
     range.end = icaltime_today();
-    icaltime_adjust(&range.end, 10*365, 0, 0, 0);
+    icaltime_adjust(&range.end, 90, 0, 0, 0);
     icalcomponent_myforeach(ical, range, floatingtz, process_alarm_cb, &rock);
 
     if (myical) icalcomponent_free(myical);


rsto (approved):

imap/caldav_alarm.c
```diff
@@ -981,11 +981,14 @@ static time_t process_alarms(const char *mboxname, uint32_t imap_uid,
     struct get_alarm_rock rock =
         { userid, mboxname, imap_uid, floatingtz, lastrun, runtime, 0, dryrun };
     struct icalperiodtype range = icalperiodtype_null_period();
-    /* Limit our recurrence expansion to a decade from today.
-       This will end up missing events that occur less frequent than
-       decennially, but what are the chances? */
```
wolfsage: Was this old comment incorrect?

It suggests that if an event was greater than our range, that event would not get processed, but why would that be?

If our range is + 10 years from today, and the event happens every 11 years, but one of the occurrences falls within the next 10 years, surely it would be processed?

If it *wouldn't* be processed, then won't shrinking that window to 90 days miss anything that occurs less frequently than 90 days?

imap/caldav_alarm.c
```diff
@@ -981,11 +981,14 @@ static time_t process_alarms(const char *mboxname, uint32_t imap_uid,
...
+    /* Limit our recurrence expansion to 90 days from today.
+       process_alarm_cb() will stop expanding recurrences if the trigger
+       for a VALARM is in the future more than 60 days from today.  
+       However, if the master component does not include any VALARMS
+       the trigger check is not done and recurrence expansion would
+       otherwise continue until the libical-specified MAX year. */
     range.end = icaltime_today();
-    icaltime_adjust(&range.end, 10*365, 0, 0, 0);
+    icaltime_adjust(&range.end, 90, 0, 0, 0);
```
wolfsage, line 991: Does this mean that with this, if an event is 4 months out and says "Alert me 4 months before", we'll no longer get an alarm?

imap/caldav_alarm.c
ksmurchison: Hmm.  If the first occurrence is outside the 90 day window then the callback wouldn't get called and we wouldn't schedule the alarm.  I guess we need a pretty wide window, but having the window go all the eay to year 20000 is crazy.  Maybe we do 11 years so that if we have a 10 year recurrence we are guaranteed to catch it.

imap/caldav_alarm.c
ksmurchison: Yeah, I think you're right.  I guess we need to pick a sane range other than 2582 or 20000.

wolfsage: I think what we should do is decide "what is a reasonable lead-time on alarms that we need to support" and figure out how to do that.

I also think that without some extensive tests to prove out our plan (and/or understand how things currently work) we shouldn't change the behaviour because we don't know what we're going to break.

rsto: I approved with the understanding that "4 months in advance" alerts make no sense for the server. In any case, systematic tests always are great. Right now, we've a bunch of alert tests scattered across Cassandane suites (mostly JMAPCalendars and CaldavAlarm.

rjbs (approved):

PR #5893, opened 2026-04-05
backup: fix deleted folder offset and add Cassandane test suite

brong:
While working on converting Fastmail's backups over to the modules in Cyrus, I was building out the test suite and ran into an issue.  Getting Claude to copy over the tests and build them out into Cassandane surfaced another bug.  This fixes both and adds the related tests.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Backup.pm
+++ b/cassandane/Cassandane/Cyrus/Backup.pm
@@ -1,0 +1,71 @@
+# SPDX-License-Identifier: BSD-3-Clause-CMU
+# See COPYING file at the root of the distribution for more details.
+
+package Cassandane::Cyrus::Backup;
+use strict;
+use warnings;
+use experimental 'signatures';
+use DBI;
+use File::Path;
+
+use base qw(Cassandane::Cyrus::TestCase);
+use Cassandane::Util::Log;
+
+sub new ($class, @args)
+{
+    my $config = Cassandane::Config->default()->clone();
+    $config->set(servername => 'backuptest');
+
+    return $class->SUPER::new({
+        config => $config,
+    }, @args);
+}
+
+sub set_up ($self)
+{
+    $self->SUPER::set_up();
+}
+
+sub tear_down ($self)
+{
+    $self->SUPER::tear_down();
+}
+
+# Helper: set up backup directories and return ($meta, $data, $service, $servername)
+sub _backup_setup ($self)
+{
+    my $servername = $self->{instance}->get_servername;
+    my $service = $self->{instance}->get_service('backupcyrusd');
+
+    my $data = "$self->{instance}->{basedir}/backupdata";
+    my $meta = "$self->{instance}->{basedir}/backupmeta";
+    mkdir($data);
+    mkdir($meta);
+
+    return ($meta, $data, $service, $servername);
+}
+
+# Helper: run BackupUser and return ($version, $size)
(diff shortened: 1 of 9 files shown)


ksmurchison (approved):
LGTM, but we should have one of our Perl experts also take a look

cassandane/Cassandane/Cyrus/Backup.pm
```diff
@@ -0,0 +1,80 @@
...
+
+    my $servername = $self->{instance}->get_servername;
+    my $service = $self->{instance}->get_service('backupcyrusd');
+
+    my $data = "$self->{instance}->{basedir}/backupdata";
+    my $meta = "$self->{instance}->{basedir}/backupmeta";
+    mkdir($data);
+    mkdir($meta);
+
+    $ENV{DEBUGIO} = 1 if $self->{store}->{verbose};
```
rjbs: This is changing global state.  I _suspect_ it's safe, because generally we shunt individual test runs into their own worker process, but I don't love it.  Good news, I _think_, is that you can just put this as a `local` into `_do_backup`, since I think only the actual backing up uses this code.

cassandane/tiny-tests/Backup/compress_deleted_folder
```diff
@@ -0,0 +1,95 @@
...
+# Regression test: deleted folders must stay deleted after compression.
+#
+# Before the fix, the imap row's offset was not updated when a folder was
+# deleted, so CompressBackup's StreamCopy matched the creation entry (keeping
+# it with deleted=0) and dropped the deletion entry.  After compression the
+# folder silently reappeared.
+sub test_compress_deleted_folder
+    :min_version_3_13 :want_service_backupcyrusd
+{
+    my ($self) = @_;
```
rjbs: Here and in other tests:  we should be using signatures.  All tests were converted to signatured subs a few months ago.  (I should add a linter for this…)

I see that our recent merge of non-tiny tests into tiny-tests brought back a bunch of snails. I'll fix that in another PR.

cassandane/tiny-tests/Backup/restore_getfile
```diff
@@ -0,0 +1,80 @@
...
+    # Back up
+    my ($version, $size) = $self->_do_backup($meta, $data, $service, $servername);
+    $self->assert_num_gt(0, $size);
+
+    # Get the SHA1 guids of backed-up files from the state DB
+    my $dbh = $self->_open_backup_db($meta);
+    my $guids = $dbh->selectcol_arrayref("SELECT guid FROM files");
+    $dbh->disconnect;
+
+    $self->assert(scalar @$guids >= 2, 'at least 2 message files in backup');
```
rjbs: unnecessary use of scalar

rjbs (changes requested):
Seems legit to me.  I think you should fix the subs to use sigs, and the ENV change to be localized.  The rest seems good AFAICT.

cassandane/Cassandane/Cyrus/Backup.pm
brong: It got removed by the logger changes anyway

cassandane/tiny-tests/Backup/compress_deleted_folder
brong: done

rjbs (approved):
I'm boldly marking approved while the checks are still running, but like… wait for those too. :-D

PR #5895, opened 2026-04-05
Cassandane update davtalk modules

brong:
I've changed the upstream Net::CalDAVTalk and Net::CardDAVTalk to use separate modules (Text::JSContact and Text::JSCalendar) for conversion, which means a couple of changes to cassandane to use those instead.

This might break tests until this is merged if the docker environment gets upgraded to the newer CPAN modules.


cassandane/Cassandane/Cyrus/Caldav.pm
```diff
@@ -156,12 +156,24 @@ sub _all_keys_match
...
+                } values %$b;
+                unless ($found) {
+                    push @$errors, "no key $key";
+                    return 0;
+                }
+                $bval = $found;
             }
             my @err;
-            unless (_all_keys_match($a->{$key}, $b->{$key}, \@err)) {
+            unless (_all_keys_match($a->{$key}, $bval, \@err)) {
```
rjbs, line 176: Gut feeling:  We could/should rewrite this in terms of Test::Deep.  But that can happen some other time.

cassandane/doc/README.deps
```diff
@@ -184,7 +196,8 @@ Fedora copypasta:
         perl-NNTPClient perl-IO-Socket-INET6 perl-Net-Server \
         perl-Unix-Syslog perl-File-chdir perl-Math-Int64 \
         perl-Mail-IMAPTalk perl-String-CRC32 perl-File-chdir \
-        perl-Unix-Syslog perl-Net-Server perl-Net-CalDAVTalk \
-        perl-Net-CardDAVTalk perl-AnyEvent \
+        perl-Unix-Syslog perl-Net-Server \
+        perl-Text-JSCalendar perl-Text-JSContact \
```
rjbs: Does perl-Text-JSCalendar exist?  Or perl-Text-JSContact?  I don't know yum, but I don't _think_ they're in there.  This section is stale, but wrong is worse than stale.  I think? 🤔

rjbs (changes requested):
My requested changes are tiny.  If you want, LMK and I'll sort out merging this and the new C*DAV-talk-based image in sequence.

cassandane/doc/README.deps
brong: Yeah, probably not - but they will at some future point I hope!  The face that Net-CalDAVTalk and Net-CardDAVTalk are in there either means they pick up any requested CPAN module pretty quick or the user is forced back to cpamn anyway.

cassandane/Cassandane/Cyrus/Caldav.pm
brong: Yeah

rjbs (approved):

elliefm:
Might need to backport this to the older branches, if the older branches' tests won't run with the newer perl modules installed.  I don't currently know if that's the case, but if it turns out it is, the backport label will be the breadcrumb to find this PR with the changes needed.

elliefm:
Backporting this will break CI for the branches it's backported to until their pinned docker images are rebuilt with the newer modules....

elliefm:
If we backported #6011 as well, that would allow us to backport this without needing to rebuild the docker images, but I don't like how it smells.

elliefm:
Decided not to backport this, and just switch to the older davtalk packages when testing builds from these branches.

PR #5883, opened 2026-03-30
Calendars: interim changes during transition to compact calendar object ids

ksmurchison:
DO NOT MERGE

This is a temporary patch until FM middlware catches up

PR #5896, opened 2026-04-06
http_caldav.c: prevent creating short C-prefixed calendars

ksmurchison:
Require any calendar href that begins with 'C' to be at least 13 chars long so they can be distinguished from new-style compact calendar Ids, which begin with 'C' and are at most 12 chars long.

This is a temp fix until FM middlware is transitioned to new-style Ids only.

### DIFF
--- a/cassandane/tiny-tests/Caldav/calendar_create_badname
+++ b/cassandane/tiny-tests/Caldav/calendar_create_badname
@@ -1,0 +1,19 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_calendar_create_badname ($self)
+{
+    my $CalDAV = $self->{caldav};
+
+    # Starts with 'C' and <= 12 chars long
+    my $res = $CalDAV->SafeRequest('MKCALENDAR', 'C0123456789A');
+    $self->assert_num_equals(403, $res->{http_response}{status});
+
+    # Starts with 'c' and <= 12 chars long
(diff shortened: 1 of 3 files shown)


cassandane/tiny-tests/Caldav/calendar_create_badname
```diff
@@ -0,0 +1,29 @@
...
+sub test_calendar_create_badname ($self)
+{
+    my $CalDAV = $self->{caldav};
+
+    my %Headers = (
+      'Authorization' => $CalDAV->auth_header()
+    );
+
+    # Starts with 'C' and <= 12 chars long
+    my $res = $CalDAV->{ua}->request('MKCALENDAR',
```
rsto: You don't need to call `ua` anymore. Using latest CalDAVTalk you can just call `SafeRequest` on the CalDAV Talk instance. See https://metacpan.org/pod/Net::DAVTalk#%24Self-%3ESafeRequest(%24method%2C-%24path%2C-%24content%2C-%25headers)

rsto (approved):

cassandane/tiny-tests/Caldav/calendar_create_badname
ksmurchison: Fixed

rjbs (approved):

rjbs:
Merged into https://github.com/cyrusimap/cyrus-imapd/pull/5883 to reduce WIP include-in-fastmail PRs

PR #5900, opened 2026-04-08
annotate.c: hide some symbols

dilyanpalauzov:


rjbs:
@dilyanpalauzov Is this one ready for review?

dilyanpalauzov:
Are there any hints suggesting this is not ready?

rjbs:
> Are there any hints suggesting this is not ready?

Not particularly.  During a team review this morning, we went through a number of things, and the question "is this its final form?" was repeatedly asked, for various reasons, before we got to this one.  So, you get asked, because I figured you'd reply quickly.

I'll note that you didn't actually answer, but your tone implies that it is ready.

elliefm (approved):
Thanks

PR #5902, opened 2026-04-10
Refactor the hash code in preparation for dynamic resizing

nwc10:
* don't store the chained buckets in sorted key order
* store newly inserted hash entries at the head of the linked list
* merge the "insert a key into an existing list" and "insert a key into an empty list" code paths
* store the 32 bit hash with the key, and check it before calling `strcmp`
* store hash keys using a flexible array in the bucket struct

### DIFF
--- a/lib/hash.c
+++ b/lib/hash.c
@@ -94,7 +94,8 @@ EXPORTED hash_table *construct_hash_table(hash_table *table, size_t size, int us
 */
 EXPORTED void *hash_insert(const char *key, void *data, hash_table *table)
 {
-      unsigned val = strhash_seeded(table->seed, key) % table->size;
+      uint32_t hash = strhash_seeded(table->seed, key);
+      unsigned val = hash % table->size;
       bucket *ptr, *newptr;
 
       /*
@@ -104,8 +105,7 @@ EXPORTED void *hash_insert(const char *key, void *data, hash_table *table)
       for (ptr=(table->table)[val];
            ptr;
            ptr=ptr->next) {
-          int cmpresult = strcmp(key,ptr->key);
-          if (!cmpresult) {
+          if (hash == ptr->hash && !strcmp(key, ptr->key)) {
               /* Match! Replace this value and return the old */
               void *old_data;
 
@@ -125,6 +125,7 @@ EXPORTED void *hash_insert(const char *key, void *data, hash_table *table)
           newptr=(bucket *)xmalloc(sizeof(bucket));
           newptr->key = xstrdup(key);
       }
+      newptr->hash = hash;
       newptr->data = data;
       newptr->next = (table->table)[val];
       (table->table)[val] = newptr;
@@ -140,21 +141,20 @@ EXPORTED void *hash_insert(const char *key, void *data, hash_table *table)
 
 EXPORTED void *hash_lookup(const char *key, hash_table *table)
 {
-      unsigned val;
       bucket *ptr;
 
       if (!table->size)
           return NULL;
 
-      val = strhash_seeded(table->seed, key) % table->size;
+      uint32_t hash = strhash_seeded(table->seed, key);
+      unsigned val = hash % table->size;
 
       if (!(table->table)[val])
             return NULL;
 
       for ( ptr = (table->table)[val];NULL != ptr; ptr = ptr->next )
       {
-          int cmpresult = strcmp(key, ptr->key);
-          if (!cmpresult)
+          if (hash == ptr->hash && !strcmp(key, ptr->key))
               return ptr->data;
       }
       return NULL;
@@ -168,7 +168,8 @@ EXPORTED void *hash_lookup(const char *key, hash_table *table)
  * since it will leak memory until you get rid of the entire hash table */
 EXPORTED void *hash_del(const char *key, hash_table *table)
 {
-      unsigned val = strhash_seeded(table->seed, key) % table->size;
+      uint32_t hash = strhash_seeded(table->seed, key);
+      unsigned val = hash % table->size;
       bucket *ptr, *last = NULL;
 
       if (!(table->table)[val])
@@ -186,8 +187,7 @@ EXPORTED void *hash_del(const char *key, hash_table *table)
             NULL != ptr;
             last = ptr, ptr = ptr->next)
       {
-          int cmpresult = strcmp(key, ptr->key);
-          if (!cmpresult)
+          if (hash == ptr->hash && !strcmp(key, ptr->key))
           {
               void *data = ptr->data;
               if (last != NULL )
--- a/lib/hash.c
+++ b/lib/hash.c
@@ -135,28 +135,11 @@ EXPORTED void *hash_insert(const char *key, void *data, hash_table *table)
               old_data = ptr->data;
               ptr -> data = data;
               return old_data;
-          } else if (cmpresult < 0) {
-              /* The new key is smaller than the current key--
-               * insert a node and return this data */
-              if(table->pool) {
-                  newptr = (bucket *)mpool_malloc(table->pool, sizeof(bucket));
-                  newptr->key = mpool_strdup(table->pool, key);
-              } else {
-                  newptr = (bucket *)xmalloc(sizeof(bucket));
-                  newptr->key = xstrdup(key);
-              }
-              newptr->data = data;
-              newptr->next = ptr;
-              *prev = newptr;
-              table->count++;
-              check_load_factor(table);
-              return data;
           }
       }
 
       /*
-      ** This key is the largest one so far.  Add it to the end
-      ** of the list (*prev should be correct)
+      ** Add it to the end of the list (*prev should be correct)
       */
       if(table->pool) {
           newptr=(bucket *)mpool_malloc(table->pool,sizeof(bucket));
@@ -196,8 +179,6 @@ EXPORTED void *hash_lookup(const char *key, hash_table *table)
           int cmpresult = strcmp(key, ptr->key);
           if (!cmpresult)
               return ptr->data;
-          else if(cmpresult < 0) /* key < ptr->key -- we passed it */
-              return NULL;
       }
       return NULL;
 }
@@ -255,10 +236,6 @@ EXPORTED void *hash_del(const char *key, hash_table *table)
               }
               table->count--;
               return data;
-          }
-          if (cmpresult < 0) {
-              /* its not here! */
-              return NULL;
           }
       }
 
--- a/lib/hash.c
+++ b/lib/hash.c
@@ -96,7 +96,6 @@ EXPORTED void *hash_insert(const char *key, void *data, hash_table *table)
 {
       unsigned val = strhash_seeded(table->seed, key) % table->size;
       bucket *ptr, *newptr;
-      bucket **prev;
 
       /*
       ** NULL means this bucket hasn't been used yet.  We'll simply
@@ -124,9 +123,9 @@ EXPORTED void *hash_insert(const char *key, void *data, hash_table *table)
       ** This spot in the table is already in use.  See if the current string
       ** has already been inserted, and if so, replace its data
       */
-      for (prev = &((table->table)[val]), ptr=(table->table)[val];
+      for (ptr=(table->table)[val];
            ptr;
-           prev=&(ptr->next),ptr=ptr->next) {
+           ptr=ptr->next) {
           int cmpresult = strcmp(key,ptr->key);
           if (!cmpresult) {
               /* Match! Replace this value and return the old */
@@ -139,7 +138,7 @@ EXPORTED void *hash_insert(const char *key, void *data, hash_table *table)
       }
 
       /*
-      ** Add it to the end of the list (*prev should be correct)
+      ** Add new keys to the start of the list
       */
       if(table->pool) {
           newptr=(bucket *)mpool_malloc(table->pool,sizeof(bucket));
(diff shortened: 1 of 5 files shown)


nwc10:
I've been testing this with the hash table benchmarking code at https://jacksonallan.github.io/c_cpp_hash_tables_benchmark/ to try to detect if I introduce any performance regressions

That benchmark code makes various assumptions about how the table behaves. Specifically (so far)

* it expects a C++ style iterator API
* and that a "lookup" for a key returns an **iterator** that it can then advance
* that a hash that stores `char *` keys stores the actual pointer, and not a copy of the C string
* implicitly, that the hash will dynamically resize

I have a tweaked version here: https://github.com/nwc10/c_cpp_hash_tables_benchmark/commits/main/

The various versions tested are

|name|description|
|-|-|
|Cyrus classic|the current code, with a load factor of 1|
|Cyrus classic|... with a load factor of 2 (ie size is "max keys testeed"/2)|
|Cyrus 1|load factor of 2; 
[result_2026-04-10T10_19_28.html](https://github.com/user-attachments/files/26632305/result_2026-04-10T10_19_28.html)
HEAD^ from this branch - all the refactorings except...|
|Cyrus 2|this branch - so the commit that inlines the key into a `char[]` array|

[result_2026-04-10T10_19_28.html](https://github.com/user-attachments/files/26632348/result_2026-04-10T10_19_28.html)

nwc10:
"future direction" that might be "not this year" - we might want to replace our hash code with *Versatable* written by the author of that benchmarking code. It can

* support custom allocators (so the mpool stuff)
* deletion at the current iteration point (which `hash_enumerate` would need)

I have not looked that the actual code to see what it's really like. Just the numbers in benchmark and the API description. The effort needed for a change might outweigh any benefits.

lib/hash.c
```diff
@@ -94,80 +94,47 @@ EXPORTED hash_table *construct_hash_table(hash_table *table, size_t size, int us
...
-      ** of the list (*prev should be correct)
+      ** Add new keys to the start of the list (which might be empty)
+      */
+
+      /*
+      ** sizeof(bucket) is 24 on 64 bit systems, as it has to allow for padding.
+      ** whereas offsetof(...) is 20. So using it saves 4 bytes on average
       */
+      size_t key_len = strlen(key) + 1; /* including the trailing NUL byte */
+      size_t wanted = offsetof(bucket, key) + key_len;
```
wolfsage, line 127: With short keys you can malloc less memory than the sizeof the struct requires. Is that safe?

wolfsage: Okay, I suppose perhaps this field only technically requires 1 byte (for a char), and the rest is probably just padding, so this is fine?

elliefm: That's an interesting point.  I can't cite standards, but my gut feel is that having a pointer of `type` to a block that's smaller than `sizeof(type)` is almost certainly "undefined behaviour".

Something like `wanted = MIN(sizeof(bucket) + 1, offsetof(bucket, key) + key_len)` would keep it safe.  Our `MIN` macro is dangerous with side effects, but those arguments don't have side effects.

And now I'm second guessing myself on whether the sizeof term needs the +1 or not, I haven't worked it through on paper.  But I think it does: we can't assume there's always padding, even though we expect that to usually be the case.

elliefm: Of course that wants to be `MAX`, not `MIN`; I always make this mistake until I test it and fix it, but I can't test a comment :sweat_smile:

elliefm:
What's our cunit-only* test coverage of these modules like?  I have a feeling it's okay, but not amazing.  Though a lot of code has been removed here so it may have improved.

For this kind of change (rearranging the internals of a low level API), I would want to have a look at the coverage and make sure there aren't any gaping holes in our tests.  Maybe you already did that and there aren't any, I'm just a little surprised to see no cunit changes.

*For `lib/` code, I don't like to rely on coverage that includes a Cassandane run, because that coverage can disappear when higher level parts of the system change how they use the API.

lib/hash.c
```diff
@@ -94,7 +94,8 @@ EXPORTED hash_table *construct_hash_table(hash_table *table, size_t size, int us
 */
 EXPORTED void *hash_insert(const char *key, void *data, hash_table *table)
 {
-      unsigned val = strhash_seeded(table->seed, key) % table->size;
+      uint32_t hash = strhash_seeded(table->seed, key);
```
rsto, line 97: nit: `strhash_seeded` returns an `unsigned` and while that's going to be a 32bit unsigned in all practical cases, it's not guaranteed by the language. We already agreed to require C11 or higher for Cyrus builds, so a `_Static_assert` would guarantee the right byte size during the build.

lib/hash.c
rsto: I'm pretty sure that this is not undefined behavior. The C11 spec (actually, the latest publicly available draft [N1570](https://open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf)) has this example on page 134:

```
struct s { int n; double d[]; };
```
and states later on this page it is possible that
```
sizeof (struct s) >= offsetof(struct s, d) + sizeof (double)
```

So this `offsetof` looks safe to me and is likely to save memory.

rsto (approved):
This looks good to me! I like how the separate commits cleanly outline the changes, which made it great to review. I only found one nit in the code about the byte size of the hash function return value, but it's unlikely to be of any practical issue. I like how the changes make reasoning about the hash tables simpler.

I agree with ellie's comment about testing, I haven't checked how good coverage of the current unit tests is. It would be great to understand where we stand with coverage currently.

I'm going to approve this PR but we might want to keep it not merged and labelled as `include-in-fastmail` for a while.

lib/hash.c
nwc10: I rebased onto current `master` and amended that last commit to do this:

```
diff --git a/lib/hash.c b/lib/hash.c
index 84d8d6f83..6f79f71b1 100644
--- a/lib/hash.c
+++ b/lib/hash.c
@@ -125,6 +125,15 @@ EXPORTED void *hash_insert(const char *key, void *data, hash_table *table)
       */
       size_t key_len = strlen(key) + 1; /* including the trailing NUL byte */
       size_t wanted = offsetof(bucket, key) + key_len;
+
+      /* Code reviewers observed that for short keys the above calculation
+         might result in an allocation smaller than the (fully padded) struct.
+         We believe that this is all fine by the C standard, but compilers are
+         software too, and this sort of thing might trigger bugs (or false
+         positive warnings from UBSAN etc). So we play it safe: */
+      if(wanted < sizeof(bucket))
+          wanted = sizeof(bucket);
+
       if(table->pool) {
           newptr=(bucket *)mpool_malloc(table->pool,wanted);
       } else {
```

I suspect that the actual allocation won't change - I'm assuming that 100% of systems these days will round up the in underlying allocation to (at least) units of `sizeof(void *)`. But it feels safer to be clear and "burn" a couple more CPU instructions (which are probably pretty much free on superscalar CPUs)

lib/hash.c
nwc10: IIRC on K-series Crays, `sizeof(short)`, `sizeof(int)` and `sizeof(long)` were all `8`. But the last of these that I was aware of was decommissioned 22 years ago: https://perldoc.perl.org/perl587delta#Platform-Specific-Problems

(this was a machine at Boeing. We also know that Airbus asked questions about `Devel::Cover`.)

It's a good observation, and one I missed. But I assume it unlikely that we're going to get future hardware within our lifetimes where it would matter.

I intend (well, **need**) to change that line further in the near future. I'll address the inconsistency then, if that's OK.

rsto: Maybe the best is to change the return value of `strhash_seeded` to `uint32_t` if that's what it is producing. But I haven't dug in the code on what other parts of the code would be impacted.

nwc10: I'm going to end up reading a lot of this code again before this gets merged, so I'll look out for this

ksmurchison (approved):

lib/hash.h
```diff
@@ -17,9 +17,10 @@
 */
 
 typedef struct bucket {
-    char *key;
     void *data;
     struct bucket *next;
+    uint32_t hash;
+    char key[];
```
dilyanpalauzov, line 23: For `c.c`:
```c
#include <stdint.h>
typedef struct bucket {
    void *data;
    struct bucket *next;
    uint32_t hash;
    char key[];
} bucket;

int main() {
  struct bucket m;
  return 0;
}
```

```
$ clang -Wpadded c.c -o c
c.c:2:16: warning: padding size of 'struct bucket' with 4 bytes to alignment boundary
      [-Wpadded]
    2 | typedef struct bucket {
      |                ^
1 warning generated.
```
Even if on many other places this padding warning does appear, I think for modified or new C-structs, the warning should be avoided.  E.g. by adding a field with name `padding` or alike.

One point with this warning is that it sometimes correctly hints, that by just changing the order of the struct-members, the memory needed by an instance of the struct decreases.

nwc10: I agree with your intent, but I don't think that I agree with "add a field" as a solution, because there isn't going to be a single correct answer.

If I build Cyrus on a 32 bit platform, the correct size of padding to add is zero, because all the members will already align.

As a specific example, I doubt that anyone is running Cyrus in production on a 32 bit system, but it was the fastest example I could think of. The general case holds - clang is correct, but padding varies depending on the platform ABI. And there can be some very quirky rules. eg (from memory) this:

```
struct foo {
   char bar[1];
};
```

has size **4** on the (long obsolete) original ARM ABI, because arrays were padded to 4 bytes.

The rules can be nutty. (On that platform IEEE floats were 8 bytes, but mixed endian)

If you can suggest a clean patch that

* removes the warning
* on 32 and 64 bit platforms (and anything else [we can throw at it](https://portal.cfarm.net/machines/list/))
* doesn't actually increase the structure size

then we'll integrate it. But I have no inspiration on how to achieve the above 3 criteria for a general solution, so I personally I don't want to detour into this.

dilyanpalauzov: Here three suggestions:
```c
#include <stdint.h>
typedef struct bucket {
    void *data;
    struct bucket *next;
    uint32_t hash;
//#if UINTPTR_MAX == 0xFFFFFFFFFFFFFFFFu
//#if __SIZEOF_POINTER__ == 8
#if defined(__x86_64__)
    uint32_t padding;  
#endif
    char key[];
} bucket;

struct bucket m;

```

> clang -Wpadded -c -o c.o c.c

nwc10: That completely defeats the point of the layout, which was to exploit those (potential) 4 bytes as space to use for the key.

The warning is a warning. The space optimisation is an intended feature.

If anything, the solution might be the relevant Clang `#pragma` (if it exists) to silence the warning in a tight scope. Because here the warning is actually **wrong** - clang isn't figuring out that the variable sized array sits in the padding. If I run `pahole` we get the true picture:

```
struct bucket {
        void *                     data;                 /*     0     8 */
        struct bucket *            next;                 /*     8     8 */
        uint32_t                   hash;                 /*    16     4 */
        char                       key[];                /*    20     0 */

        /* size: 24, cachelines: 1, members: 4 */
        /* padding: 4 */
        /* last cacheline: 24 bytes */
};
```

There is no padding. There is a variable sized array.

We seem to be talking cross purposes. You're keen to avoid warnings (and I agree with that). You're keen for new code not to add warnings, and be of a higher standard than before (and I agree with that). But I'm not going to placate the compiler if it's making mistakes.

lib/hash.c
elliefm: I agree that changing strhash_seeded to return uint32_t specifically is the right approach.

lib/hash.c
elliefm: For what it's worth, our style is `if (wanted)` (i.e. with a space between `if` and `(`), but I can see you're mimicking the style from the `if(table->pool)` line below.  It just so happens that the nearest example is a bad one -- this particular file is rife with them.

Generally, we do not put a space in `function_call(foo)`, but we do in `keyword (foo)`.  Though it's not quite exactly that, because we usually pretend the `sizeof`, `offsetof`, etc operators are functions rather than language keywords.  So the delineation is more like, if it behaves kinda like a function call, no space; otherwise space.

@rsto:

> it is possible that `sizeof (struct s) >= offsetof(struct s, d) + sizeof (double)`, so this offsetof looks safe and is likely to save memory

In context, this is describing the behaviour of `t1.d[0]` after `t1` has been declared as `struct s t1 = { 0 };`, which is then described as "probably undefined behaviour" and "cannot appear in strictly conforming code".  In this case, `t1` _is allocated on the stack_, where flexible array members don't exist, and therefore there is no 0th element to the `d` array (except accidentally via padding), which is why it's bad.  But it's not relevant here.

In our case, we're talking about a structure _allocated on the heap_, which we have a pointer to.  It's not the `offsetof` that's dangerous.  The undefined behaviour (if any) will arise from the relationship between the expected size of the type being pointed to (which includes padding, and excludes flexible array members), and the actual size of the heap allocation.

Specifically, if the key is only 1-2 characters long, we'll only malloc space for 2-3 bytes (including '\0') after `hash`.  On systems where the struct is padded out to a multiple of 8 bytes, the expected size of the type will include 4 bytes of padding after `hash`.  That's greater than the 2-3 bytes that we allow for the key, putting us in a position where we've malloced less than the pointed-to type expects.

By ensuring we always allocate at least as many bytes as the expected size of the type, there's no more problem.  At worst, we "waste" two bytes for a 1 character key... though they probably would've been wasted anyway, because malloc usually returns 8-byte aligned addresses on platforms we care about.

PR #5903, opened 2026-04-10
more fixes to sieveshell

pieska:
some more minor fixes, see commit msg

### DIFF
--- a/perl/sieve/scripts/sieveshell.pl
+++ b/perl/sieve/scripts/sieveshell.pl
@@ -268,7 +268,9 @@ while(defined($_  = ($interactive ? $term->readline('> ') : <$filehandle>))){
     }
 }
 
-exit $exitcode
+close($filehandle) if !$interactive;
+
+exit $exitcode;
 
 __END__
 
--- a/perl/sieve/scripts/sieveshell.pl
+++ b/perl/sieve/scripts/sieveshell.pl
@@ -73,13 +73,12 @@ my $filehandle;
 my $interactive;
 
 if ($exfile ne "") {
-    open(FILEH,"<$exfile") || die "unable to open file: $?";
-    $filehandle = *FILEH;
+    open($filehandle, '<', $exfile) || die "unable to open file: $!";
     $interactive = 0;
 } elsif ($ex ne "") {
     $filehandle = tempfile();
 
-    if (!$filehandle) { die "unable to open tmp file: $?"; }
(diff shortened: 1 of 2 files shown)


perl/sieve/scripts/sieveshell.pl
```diff
@@ -52,13 +51,13 @@
 my $interactive;
 
 if ($exfile ne "") {
-    open(FILEH,"<$exfile") || die "unable to open file: $?";
-    $filehandle = *FILEH;
+    open(my $fh, '<', $exfile) || die "unable to open file: $!";
```
rjbs: Seems to me like in this case, `$fh` isn't needed.  Instead you could write:

```perl
open($filehandle, '<', $exfile) || die "unable to open file: $!";
```

…and avoid a needless variable.

perl/sieve/scripts/sieveshell.pl
```diff
@@ -107,8 +106,8 @@ sub prompt {
 
   print "$prompt: ";
 
-  $b = <STDIN>;
-  chop($b);
+  my $b = <STDIN>;
+  chomp($b);
```
rjbs: So, the reason this worked before is that `$a` and `$b` are always permitted, despite `use strict` being in place, because they're magic for `sort`.  For that reason, I advise one should never use these variables.  How about renaming this?  (Also, `$b` is a weird name to begin with!  Feels like it should be `$input`?)

perl/sieve/scripts/sieveshell.pl
```diff
@@ -237,9 +236,9 @@ sub show_help {
             $exitcode = 1;
         } else {
             if ($words[2]) {
-                open (OUTPUT,">$words[2]") || die "Unable to open $words[2]";
-                print OUTPUT $str;
-                close(OUTPUT);
+                open(my $output, '>', $words[2]) || die "Unable to open $words[2]";
```
rjbs: While you're in here, you should tack on `$!` to this error.

rjbs (changes requested):
Thanks, these look like improvements.  I have asked you to slightly improve the improvements. :)

pieska:
I tried to improve the improvements and added some more

rjbs:
Thanks!  I've rebased and squashed some commits to get them about normal size for the project.  Normally I'd ask you to do that, but you've already been waiting weeks for me to get back to you (I've been tied up!).  Merged!

PR #5904, opened 2026-04-10
jmap_calendar.c: support compact IDs for JMAP CalendarEvents

ksmurchison:
Uses an encoded form of index record createdmodseq as the ID


imap/jmap_calendar.c
```diff
@@ -3729,11 +3734,12 @@ static int jmap_calendarevent_get(struct jmap_req *req)
         /* Split into single-valued uids and event recurrence instance ids */
         json_array_foreach(get.ids, i, jval) {
             const char *id = json_string_value(jval);
-            struct jmap_caleventid *eid = jmap_caleventid_decode(id);
-            ptrarray_t *eventids = hash_lookup(eid->ical_uid, &eventids_by_uid);
+            struct jmap_caleventid *eid = jmap_caleventid_lookup(id, db, NULL);
+            const char *key = eid->ical_uid ? eid->ical_uid : eid->raw;
```
rsto, line 3738: Why fall back to `raw` here?

rsto (commented):
This is my first review of this PR. I might follow up with a more detailed review once the jscalendarbis code is ready to land. Right now, I have only one question.

imap/jmap_calendar.c
ksmurchison: Because we need some kind of unique key to store the `eid` in the hash table

rsto: But would we ever find a key with that id?

ksmurchison: No we wouldn't.  But I couldn't figure out the `nseen` / `want_eventids` logic and was afraid of incorrectly reporting `notFound`

rjbs:
I think this all seems in order.  We need to coordinate before deploying this.

One thing I was not 100% on here:  Are we confident that we can't end up with two occurrences of one event having two distinct CMS?  I think it should be something like "they must be in one resource if on one calendar, and we try hard not to allow one uid to be on two calendars", so we're good… right?

rjbs (approved):

imap/caldav_db.c
```diff
@@ -432,6 +432,30 @@ EXPORTED int caldav_lookup_uid(struct caldav_db *caldavdb, const char *ical_uid,
 }
 
 
+#define CMD_SELCMODSEQ CMD_READFIELDS \
+    " WHERE createdmodseq = :cmodseq AND mailbox != :inbox AND alive = 1;"
```
wolfsage, line 436: Why not inbox? Can a cmodseq be identical in the inbox and on an event? Does this not also apply to outbox or any other?

imap/jmap_calendar.c
```diff
@@ -6021,8 +6033,11 @@ static int jmap_calendarevent_set(struct jmap_req *req)
     json_object_foreach(set.update, id, arg) {
         jmap_caleventid_free(&eid);
 
-        eid = setcalendarevents_parse_id(req, id);
-        if (!eid) {
+        struct caldav_data *cdata = NULL;
+        eid = setcalendarevents_lookup_id(req, id, db, &cdata);
+        if (!eid || !cdata || !cdata->dav.alive ||
+            !cdata->dav.rowid || !cdata->dav.imap_uid ||
+            cdata->comp_type != CAL_COMP_VEVENT) {
```
wolfsage, line 6040: Why all the extra checks (especially on comp_type). What other comp_types might be found by this lookup?

(And why aren't these checks on _destroy?)

wolfsage (commented):
It seems torturous to determine if an id is valid or not.

Some things do:
```
        eid = setcalendarevents_lookup_id(req, id, db, &cdata);
        if (!eid || !eid->ical_uid) {
```
and others:
```
        eid = setcalendarevents_lookup_id(req, id, db, &cdata);
        if (!eid || !cdata || !cdata->dav.alive ||
            !cdata->dav.rowid || !cdata->dav.imap_uid ||
            cdata->comp_type != CAL_COMP_VEVENT) {
```

It might be safer/cleaner to have an api that returns undef if the thing isn't found?

Maybe relatedly, this crashes now (it didn't before):
```
#!perl
use Cassandane::Tiny;

sub test_calendarevent_set_destroy_itip
    :min_version_3_7
    ($self)
{
    my $jmap = $self->default_user->jmap;
    my $res = $jmap->CallMethods([
        ['CalendarEvent/set', {
            'destroy' => [ "a", "b",],
        }, 'a' ],
    ]);
}

```

imap/caldav_db.c
ksmurchison: The Inbox is a CalDAV-only calendar that holds incoming iTIP resources.  We hide it from JMAP clients.  And, even in the CalDAV context, when looking for events by UID, we ignore the Inbox because those resources will also have an event with matching UIDs in other calendars.

imap/jmap_calendar.c
ksmurchison: I've now consolidated the checks into setcalendarevents_lookup_id().

We check for VEVENT because we currently only support CalendarEvent.  Perhaps down the road we will support CalendarTask (VTODO)

imap/jmap_calendar.c
ksmurchison: Figured out the logic and fixed this.

ksmurchison:
> I think this all seems in order. We need to coordinate before deploying this.
> 
> One thing I was not 100% on here: Are we confident that we can't end up with two occurrences of one event having two distinct CMS? I think it should be something like "they must be in one resource if on one calendar, and we try hard not to allow one uid to be on two calendars", so we're good… right?

Yes.  Two standalone recurrences should end up in the same resource and have the same CMS

rsto:
@ksmurchison this was reviewed and approved already. What specifically would you like to get re-reviewed?

ksmurchison:
> @ksmurchison this was reviewed and approved already. What specifically would you like to get re-reviewed?

Mostly making sure that you are happy with the change where we immediately flag events where eid->ical_uid wasn't resolved as notFound (and doesn't use eid->raw as a hash key for an event that won't be found anyway)

rsto (approved):
LGTM

rjbs (approved):

rsto:

PR #5905, opened 2026-04-10
twom long key support

brong:
We discovered when somebody had a 100kb key that twom didn't correctly store FAT keys and values.  All the logic was there, but the line to switch it on got lost in a refactor somewhere!  Dammit.

This has appeared both in duplicate db and conversationsdb in Fastmail production with 100kb messageids (yes, really)

### DIFF
--- a/cunit/aaa-db.testc
+++ b/cunit/aaa-db.testc
@@ -2012,4 +2012,62 @@ static void test_foreach_replace(void)
 }
 
 
+static void test_huge_key(void)
+{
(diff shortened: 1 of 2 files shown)


ksmurchison (commented):
Should this have a cunit test or something?

brong:
> Should this have a cunit test or something?

it does in the twom repo.  I can copy it over

ksmurchison (approved):

PR #5906, opened 2026-04-14
Email/set: destroy before creating

rjbs:
With deterministic email creation, you can end up trying to create a new draft identical to the old one.  (Weird, but true.)  Create fails, but if you did a destroy of the old in the same request, destroy succeeds.  Oops!

Destroy first.

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_set_create_identical_draft_replace
+++ b/cassandane/tiny-tests/JMAPEmail/email_set_create_identical_draft_replace
@@ -1,0 +1,67 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_set_create_identical_draft_replace ($self)
+{
+    my $user = $self->default_user;
+    my $jmap = $user->jmap;
+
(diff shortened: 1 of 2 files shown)


elliefm (commented):
Destroy-before-create makes me nervous.  What if the create fails for some reason after the destroy has succeeded?

I don't have a better suggestion; and maybe it's safe anyway due to some part of how the bigger picture hangs together, which I'm not familiar with.  So this is more of a drive by peanut gallery comment.

rjbs:
I think it's a bit of the same thing:

1. create first; it fails because of dupe; destroy after; that works: no draft remains!
2. destroy first; it works!  create after; it fails because of sunspots: no draft remains!

The real answer, imho, is that _a client should be holding onto the data until the update confirms it was applied correctly_.  If that happens, _both_ are safe, but:

1. the first case would still keep failing, and the client would need an algorithm fix
2. the second case would work when the sunspots clear up

ksmurchison (approved):
LGTM

PR #5908, opened 2026-04-15
mailbox: auditlog rename of UUID-mailboxes

elliefm:
We usually auditlog after the mailbox disk changes have succeeded, but there aren't any mailbox disk changes when renaming a UUID mailbox.

But not auditlogging renames at all is confusing, so do it anyway.


elliefm:
@rjbs I think this is all we need to satisfy the immediate need for logging of renames.

rjbs (approved):
This will be helpful!

PR #5909, opened 2026-04-15
emacs: open `.testc`  files in c-mode

dilyanpalauzov:


elliefm (approved):
Ahh cool, thanks.  I'd been idly wondering if there was a way to do this, since we have the vim equivalent in the files directly, but nothing for emacs.

I don't use emacs myself, but I'm gonna trust that this works, since you're the only person who's contributed to this file anyway.

PR #5912, opened 2026-04-16
sieve_db.c: support a shared #sieve mailbox for global scripts

ksmurchison:
Sieve scripts created by admins are stored in a shared space rather than in user space.  Prior to this fix, we were trying use a #sieve mailbox under the admins (non-existent) INBOX.

### DIFF
--- a/imap/sieve_db.c
+++ b/imap/sieve_db.c
@@ -547,6 +547,7 @@ static int store_script(struct mailbox *mailbox, struct sieve_data *sdata,
 
     /* Create RFC 5322 header for script */
     char *userid = mboxname_to_userid(mailbox_name(mailbox));
+    if  (!userid) userid = xstrdup(strarray_nth(config_admins,0));
     if (strchr(userid, '@')) {
         buf_printf(&buf, "<%s>", userid);
     }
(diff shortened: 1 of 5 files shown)


imap/sieve_db.c
```diff
@@ -1042,7 +1041,10 @@ EXPORTED char *sieve_mboxname(const char *userid)
 
     buf_setcstr(&boxbuf, config_getstring(IMAPOPT_SIEVE_FOLDER));
 
-    res = mboxname_user_mbox(userid, buf_cstring(&boxbuf));
+    if (userid)
+        res = mboxname_user_mbox(userid, buf_cstring(&boxbuf));
+    else
+        res = buf_release(&boxbuf);
```
dilyanpalauzov, line 1047: This function can be short-circuited to:
```c
init_internal();
return userid
    ? mboxname_user_mbox(userid, config_getstring(IMAPOPT_SIEVE_FOLDER))
    : config_getstring(IMAPOPT_SIEVE_FOLDER);
```

dilyanpalauzov (commented):
I do not see how this is done here, but for me it is useful to have global Sieve scripts per domain and global Sieve scripts for all domains (⇔for no-domain).

That is, if the user, for whom the Sieve script is currently evaluated, is `a@b` — in domain `b`, and that current Sieve script does `include :global "c";` then first it is tried to import the Sieve script from the global scripts of the domain `b`.  If such Sieve script does not exist, then the global Sieve scripts for “no-domain” is imported.

This allows defining global policies, and for some domains to overwrite the global policy.

dilyanpalauzov:
In fact at https://mail.aegee.org/cgit/aegee-milter/tree/src/lists/list_timsieved_scripts.cpp#n106 I have written a function to obtain the a global Sieve script from the file system for a particular domain, and for the no-domain.  Somehow I have also patched Cyrus IMAP to store these distinctly.

PR #3742, opened 2021-10-31
Make some variables static

dilyanpalauzov:
as suggested by clang -Wmissing-variable-declarations

PR #5913, opened 2026-04-17
Implement jscalendarbis and jscalendar-icalendar

rsto:
This adds support for the [jscalendarbis](https://datatracker.ietf.org/doc/draft-ietf-calext-jscalendarbis/) and [jscalendar-icalendar](https://datatracker.ietf.org/doc/draft-ietf-calext-jscalendar-icalendar/) IETF internet drafts. It also updates some but not all of the JMAP Calendars implementation to support the new JSCalendar version and related definitions in the [jmap-calendars ](https://datatracker.ietf.org/doc/draft-ietf-jmap-calendars/)internet draft.

This pull request adds a new JMAP capability `https://cyrusimap.org/ns/jmap/jscalendarbis`. JMAP requests that use this capability will process JSCalendar data according the new specifications.

The JMAP Calendars tests in Cassandane have been duplicated to test both the old and the new implementation: the `JMAPCalendarsRFC8984`suite tests the old implementation, the `JMAPCalendars` suite the new one. The tests in the new suite have been updated with the new specifications.

## 👉 Reviewers
No one can expect you to review over 600 files and thousands of lines of code. To help you understand the changes on a high level, here is a breakdown of the change sets. Since the Git refs will get outdated with every rebase, I only include the commit titles (from older commit to most recent one):

### Preparatory Changes
These all are preparatory commits to handle new libical and for hooking jscalendarbis conversion into the calendaring code:

- `jmap_util.c: use new JSID property for Alert id`
- `caldav_util.c: fix iterator bug when comparing property updates`
- `caldav_util: compare property kinds numerically, not by string`
- `itip_support.c: set SENT-BY parameter on ORGANIZER`
- `caldav_db.c: determine privacy from both CLASS and X-JMAP-PRIVACY`
- `JMAPCalendars: fix assert for X-JMAP-PRIVACY property`
- `jcal.c: support unknown IANA token names`
- `ical_support.c: instruct libical to not drop unknown IANA tokens`
- `ical_support.h: declare icalcomponent_add_required_timezones function`
- `ical_support.h: add icaltimezone_is_builtin_timezone_tzid function`
- `jmap_util.h: add jmap_parser_path_at function`
- `jcal.c: set x-component name when converting from jCal`
- `util.c: add function to generate UUID version 5 using SHA1`
- `jmap_util.c: use parser-internal buffer in jmap_parser_path`
- `jmap_util: add function to apply PatchObject in-place`
- `ical_support.c: return 'Etc/UTC' in icaltimezone_get_location_tzid`
- `jmap_util.c: add KEEP_EXISTING flag to JMAP patchobject_apply`
- `jcal.c: export icalparameter_to_jcal_parameter`
- `xcal.c: handle new libical value types`
- `json_support: add json_object_steal helper function`
- `charset.c: add percent encoding function`
- `jcal.c: get component name of x-component`
- `jcal.h: refactor begin_jcal to http_caldav.c`
- `jcal.h: add include guards`
- `jcal.h: export property conversion functions`

### JSCalendar Conversion
This adds jscalendarbis/iCalendar conversion with the new `jscal.h` and `jscal.c` files:
- `jscalendar.c: implement jscalendar-icalendar conversion`

### New HTTP module
This adds an experimental HTTP module `convert`. This module is not enabled by default and only is used for testing the conversion rules of the jscalendar-icalendar internet draft. We will decide if to merge this module or not when this PR is finished:

- `http_convert.c: add HTTP module stub for jscalenar-icalendar tests`

### JMAP Calendars Changes
These hook jscalendarbis conversion into the JMAP Calendars API, supporting both old-style and new conversion. If you are looking to review the changes with the highest chance of impact, this is it. This also snapshots the existing JMAPCalendars tests into the JMAPCalendarsRFC8984 test suite and adapts the JMAPCalendars tests for jscalendarbis:

- `jmap_calendar.c: support jscalendarbis extension`
- `JMAPCalendars.pm: adapt tests for jscalendarbis`
- `JMAPCalendarsRFC8984.pm: snapshot current JMAP Calendars tests`

### Compact Ids

These merge the pull requests that enable compact ids for calendars and calendar events. Since these PRs we written in parallel to this PR, the tests had to be adapted in both the JMAPCalendars and JMAPCalendarsRFC8984 test suites. The PRs are:

- https://github.com/cyrusimap/cyrus-imapd/pull/5846
- https://github.com/cyrusimap/cyrus-imapd/pull/5904
- https://github.com/cyrusimap/cyrus-imapd/pull/5946

and the commits are:
- `JMAPCalendars: adapt tests for compact event ids`
- `jmap_calendar.c: support compact IDs for JMAP CalendarEvents`
- `JMAPCalendars: adapt tests for compact calendar ids`
- `jmap_calendar.c: support compact IDs for JMAP Calendars`
- `jmap_calendar.c: prefix Calendar[Event] state strings with 'J'`
- `jmap_mail[box].c: use jmap_state_matches() where appropriate`
- `jmap_api.c: add jmap_state_matches()`


lib/charset.c
```diff
@@ -4661,6 +4661,49 @@ EXPORTED int charset_decode_percent(struct buf *dst, const char *val)
 }
 
 // clang-format: off
+const unsigned char IS_PERCENT_ENCODE_SAFE_CHAR[256] = {
```
dilyanpalauzov: This should be `static`, so that it moves to read-only memory and all binaries loadng libcyrus.so use the same memory and allocate that memory once.  Without `static` is binary, loading libcyrus.so, has its own copy of `IS_PERCENT_ENCODE_SAFE_CHAR`.

lib/util.c
```diff
@@ -1275,6 +1276,79 @@ EXPORTED const char *makeuuid()
...
+        case 20:
+            uuidstr[j++] = '-';
+            // fallthrough
+        default:
+            uuidstr[j++] = hex[i];
+        }
+    }
+    uuidstr[sizeof(uuidstr) - 1] = '\0';
+
+    return uuidstr;
```
dilyanpalauzov, line 1348: The last line before `#else` and before `#endif` is identical, so it can be moved after `#endif`.

imap/http_convert.c
```diff
@@ -0,0 +1,256 @@
...
+        txn->error.desc = "Could not parse JSON data";
+        ret = HTTP_BAD_REQUEST;
+        goto done;
+    }
+
+    /* Convert to iCalendar */
+    ical = jscal_to_ical(&jscal_cfg, jobj, &parser);
+    if (ical) {
+        const char *resp_payload = icalcomponent_as_ical_string(ical);
+        write_body(HTTP_OK, txn, resp_payload, strlen(resp_payload));
```
dilyanpalauzov, line 180: Can `icalcomponent_free(ical);` from `done:` be moved here?

imap/http_convert.c
```diff
@@ -0,0 +1,256 @@
...
+        txn->flags.conn = CONN_CLOSE;
+        return ret;
+    }
+
+    /* Parse the request body */
+    const char **hdr = spool_getheader(txn->req_hdrs, "Content-Type");
+    if (hdr && is_mediatype("text/calendar", hdr[0])) {
+        return convert_to_jscal(txn);
+    }
+    else if (hdr && is_mediatype("application/jscalendar+json", hdr[0])) {
```
dilyanpalauzov: `else` is not necessary.

imap/ical_support.c
```diff
@@ -202,6 +205,12 @@ icalproperty_get_datetimeperiod(icalproperty *prop)
     return ret;
 }
 
+EXPORTED bool icaltimezone_is_builtin_timezone_tzid(const char *tzid)
```
dilyanpalauzov: The function `icaltimezone_is_builtin_timezone_tzid` is not called from anywhere, so it can be made `static`.

imap/jcal.c
```diff
@@ -618,7 +620,7 @@ static icalvalue *json_object_to_icalvalue(json_t *jvalue,
 /*
  * Construct an iCalendar property from a JSON array.
  */
-static icalproperty *json_array_to_icalproperty(json_t *jprop)
+EXPORTED icalproperty *jcal_array_as_icalproperty(json_t *jprop)
```
dilyanpalauzov, line 623: The function `jcal_array_as_icalproperty` is not called from outside this file, so it can be `static`.

imap/jmap_api.c
```diff
@@ -3636,3 +3645,14 @@ EXPORTED void jmap_report_isdefault(struct jmap_set *set, const char *name,
                             json_pack("{s:b}", "isDefault", isdef));
     }
 }
+
+EXPORTED bool jmap_state_matches(struct conversations_state *cstate,
```
dilyanpalauzov: This function is not called outside `httpd`, so it should be `HIDDEN`.

The brackets after the second `return` below do not help

dilyanpalauzov (commented):
There is no documentation for the `convert` http module.  Ideally it should be possible to disable it at compile time.

Implemented drafts should be mentioned in docsrc/rfc-support.rst.

lib/charset.c
rsto: Right, done.

lib/util.c
rsto: But their return value `uuidstr` is declared at different locations in the `#if` and `#else` block, so it seems clearer to me to not share any code between the two branches.

imap/http_convert.c
rsto: It could, but since it's declared at the top of the function it's equally fine to only free it at the end of the function.

imap/http_convert.c
rsto: done

imap/ical_support.c
rsto: thanks, that was a left-over from a previous refactor.

imap/jcal.c
rsto: For consistency with `jcal_string_as_icalcomponent` and `jcal_array_as_icalcomponent` I think this should be kept in the header.

imap/jmap_api.c
rsto: @ksmurchison this is code that got merged from your PRs. I think both suggestions are generally sound, but I don't have a strong opinion. `HIDDEN` sounds like the right call, but the use of `HIDDEN` vs `EXPORTED` is all over the place in `jmap_api.c` and it's likely all of it should become `HIDDEN` anyway. If the "brackets after the second return" mean the surrounding parentheses in `(atomodseq_t(if_in_state) == modseq)` then I'm fine with removing them.

rsto:
Thanks for your review, I updated the branch.

> There is no documentation for the `convert` http module. Ideally it should be possible to disable it at compile time.

Ideally, we do not even merge it, but we need to first port the examples defined in the RFC documents to proper Cassandane tests.  If we do merge, I agree we need to add documentation. We prefer not to add more build time config options, unless required. Here, I think not disabling the module is good enough.

> Implemented drafts should be mentioned in docsrc/rfc-support.rst.

Done.

imap/json_support.c
```diff
@@ -80,3 +80,10 @@ json_t *json_object_get_vanew(json_t *obj, const char *key,
 
     return val;
 }
+
+json_t *json_object_steal(json_t *object, const char *key)
```
dilyanpalauzov: This function is not called from anywhere.

imap/jscalendar.h
```diff
@@ -0,0 +1,90 @@
...
+     *  the default email address URI to set in the ATTENDEE property
+     *  of the VALARM, unless the "iCalComponent" property of the Alert
+     *  object already contains ATTENDEE properties.
+     *
+     *  If the value is NULL then no ATTENDEE property is set. Note this
+     *  violates the VALARM definition in RFC 5545, section 3.6.6.*
+     */
+    const char *emailalert_default_uri;
+
+    /**
```
dilyanpalauzov: This line ends in a space.

imap/jscalendar.h
```diff
@@ -0,0 +1,90 @@
...
+
+} jscal_cfg_t;
+
+/** @brief Convert a JSCalendar object to an iCalendar object.
+ *
+ *  @param cfg     Conversion configuration, or NULL for defaults.
+ *  @param jobj    JSON object representing a JSCalendar Group or Event.
+ *  @param parser  JMAP parser used to report property errors.
+ *  @return        A newly allocated iCalendar object, or NULL on error. */
+extern icalcomponent *jscal_to_ical(jscal_cfg_t *cfg,
```
dilyanpalauzov: `extern` for function declaration has no added value - without it, it is the same.

imap/jscalendar.c
```diff
@@ -0,0 +1,6711 @@
...
+    const char *prop_name = NULL;
+
+    if (kind == ICAL_X_PROPERTY)
+        prop_name = icalproperty_get_x_name(prop);
+    else if (kind == ICAL_IANA_PROPERTY)
+        prop_name = icalproperty_get_iana_name(prop);
+    else
+        prop_name = icalproperty_kind_to_string(kind);
+
+     return !strcasecmpsafe(name, prop_name);
```
dilyanpalauzov: `return` is indented one space too much; likewise in the next function.

imap/jmap_calendar.c
```diff
@@ -3229,10 +3375,216 @@ static void remove_jsicalprops(json_t *jsobj, struct jmap_parser *parser)
...
+
+        json_t *jgroup = jscal_from_ical(&cfg, ical);
+        if (!jgroup) return NULL;
+        json_t *jevent = json_incref(
+            json_array_get(json_object_get(jgroup, "entries"), 0));
+        json_object_set(jevent, "version", json_object_get(jgroup, "version"));
+        json_decref(jgroup);
+        return jevent;
+    }
+    else {
```
dilyanpalauzov, line 3591: `else` can be deleted.

imap/jmap_calendar.c
```diff
@@ -3229,10 +3375,216 @@ static void remove_jsicalprops(json_t *jsobj, struct jmap_parser *parser)
...
+        if (!jgroup) return NULL;
+        json_t *jentries = json_incref(json_object_get(jgroup, "entries"));
+        for (size_t i = 0; i < json_array_size(jentries); i++) {
+            json_object_set(json_array_get(jentries, i),
+                    "version", json_object_get(jgroup, "version"));
+        }
+        json_decref(jgroup);
+        return jentries;
+    }
+    else {
```
dilyanpalauzov, line 3616: `else` can be deleted.

imap/jscalendar.c
```diff
@@ -0,0 +1,6711 @@
+/* jscalendar.c -- Routines for converting JSCalendar and iCalendar
+ *
+ * Copyright (c) 2025 Fastmail Pty Ltd
```
ksmurchison: Do we want to use a different copyright/license for these files rather than the standard
```
/* SPDX-License-Identifier: BSD-3-Clause-CMU */
/* See COPYING file at the root of the distribution for more details. */

```

imap/jscalendar.c
```diff
@@ -0,0 +1,6711 @@
...
+
+static bool is_known_prop(icalcomponent *comp, icalproperty *prop)
+{
+    icalcomponent_kind comp_kind = icalcomponent_isa(comp);
+    icalproperty_kind prop_kind = icalproperty_isa(prop);
+
+    // Some properties are known for any components.
+    if (myicalproperty_has_name(prop, PROP_JSID))
+        return true;
+    else if (myicalproperty_has_name(prop, PROP_JSPROP))
```
dilyanpalauzov: `else` can be removed.

imap/jscalendar.c
```diff
@@ -0,0 +1,6711 @@
...
+         icalpropiter_next(&iter))
+
+#define myicalproperty_foreach_parameter(prop, param_kind, param, iter)        \
+    for (iter = icalproperty_begin_parameter(prop, param_kind);                \
+         (param = icalparamiter_deref(&iter));                                 \
+         icalparamiter_next(&iter))
+
+// Non-standard iCalendar elements.
+#define PARAM_JSID              "JSID"
+#define PARAM_JSPTR             "JSPTR"
```
ksmurchison, line 45: Any reason to not add these params and the correspondinng props to libical?

ksmurchison (approved):
LGTM.  I see no major issues

imap/jmap_api.c
ksmurchison: Fine by me on both points

imap/jscalendar.c
rsto: I do not want to add non-standard iCalendar elements to libical. Once the RFCs are published, we can add them. While it's unlikely that they'll change, they might.

Also: not having them in libical allowed me to verify that the support for arbitrary IANA elements that we recently added to libical is good enough.

imap/jscalendar.c
rsto: Good catch, thanks. This used the old-style copyright because I had started working on this branch before we switched to the new style.

imap/json_support.c
rsto: Thanks, again left-over from a refactor.

imap/jscalendar.h
rsto: Done

imap/jscalendar.c
rsto: done

imap/jmap_calendar.c
rsto: Right, but in the near future we will want to remove all `else` branches for conditionals like `jmap_is_using(req, JMAP_JSCALENDARBIS_EXTENSION)`, so we'll keep that for now.

imap/jscalendar.c
rsto: done

imap/jscalendar.h
rsto: I was tempted to say that using `extern` is a stylistic convention in Cyrus, but it turns out our headers are inconsistently using it. I removed `extern` here.

imap/jmap_calendar.c
rsto: Right, but in the near future we will want to remove all else branches for conditionals like jmap_is_using(req, JMAP_JSCALENDARBIS_EXTENSION), so we'll keep that for now.

imap/jmap_api.c
rsto: done

imap/jscalendar.h
```diff
@@ -0,0 +1,88 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "jmap_util.h"
+
+#include <jansson.h>
+#include <libical/ical.h>
+
```
dilyanpalauzov, line 14: I find it unusual, that `#include` with `"…"` (cyrus-local) is done before includes with `<…>` (global files).

All three functions in jscalendar.c can be changed from EXPORTED to HIDDEN, as they are used only within `httpd`.

lib/charset.c
```diff
@@ -4661,7 +4661,50 @@ EXPORTED int charset_decode_percent(struct buf *dst, const char *val)
...
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+};
+
+static const char NIBBLE_TO_HEX_CHAR[16] = {
+    '0', '1', '2', '3', '4', '5', '6', '7',
+    '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'
+};
+
+EXPORTED void charset_encode_percent(struct buf *dst, const char *val)
```
dilyanpalauzov: The function `charset_encode_percent` is not called from anywhere.

imap/jscalendar.h
rsto: yeah, right. done.

imap/jscalendar.c
```diff
@@ -0,0 +1,6709 @@
...
+static icalcomponent *jobj_get_icalcomp(jscal_ctx_t *ctx,
+                                        json_t *jobj,
+                                        icalcomponent_kind want_kind,
+                                        enum get_ical_flags flags)
+{
+    icalcomponent *comp = NULL;
+
+    if (!ctx->cfg.use_icalendar_convprops) {
+        if (want_kind != ICAL_ANY_COMPONENT && (flags & GET_ICAL_CREATE)) {
+            comp = icalcomponent_new(want_kind);
```
dilyanpalauzov: I find it simpler to write without assigning to `comp`:
```c
if (want_kind … ) {
  return icalcomponent_new(want_kind);
}
return NULL;
```

imap/jscalendar.c
```diff
@@ -0,0 +1,6709 @@
...
+                    icalparameter *param = icalparamiter_deref(&pi);
+                    while (param) {
+                        icalparameter *next = icalparamiter_next(&pi);
+                        if (is_known_param(prop, param)) {
+                            icalproperty_remove_parameter_by_ref(prop, param);
+                        }
+                        param = next;
+                    }
+                }
+                if (prop) {
```
dilyanpalauzov: Line 1523 already contains `if (prop) {`.  Can this `if (prop) {` be avoided?

imap/jscalendar.c
```diff
@@ -0,0 +1,6709 @@
...
+    }
+    if (jsid && !json_object_get(jobj, jsid)) {
+        buf_setcstr(buf, jsid);
+        return buf_cstring(buf);
+    }
+
+    // Generate UUIDv5 from property value.
+    myicalproperty_make_uuid5(prop, buf);
+    if (!json_object_get(jobj, buf_cstring(buf))) {
+        return buf_cstring(buf);
```
dilyanpalauzov, line 1667: The previous line ensures that buf contains `0` terminated string, so here it can be safety written `return buf->s;`.

imap/jscalendar.c
```diff
@@ -0,0 +1,6709 @@
...
+static void jsid_to_comp(icalcomponent *comp, const char *jsid)
+{
+    struct buf buf = BUF_INITIALIZER;
+
+    if (icalcomponent_isa(comp) == ICAL_PARTICIPANT_COMPONENT) {
+        // Omit JSID property if it is derived from CALENDAR-ADDRESS property.
+        icalproperty *prop =
+            myicalcomponent_get_property(comp, ICAL_CALENDARADDRESS_PROPERTY);
+        if (prop) {
+            buf_reset(&buf);
```
dilyanpalauzov: I do not see what role has `buf_reset(&buf)` here.

imap/jscalendar.c
```diff
@@ -0,0 +1,6709 @@
...
+    case ICAL_SKIP_UNDEFINED:
+        break;
+    }
+    if (buf_len(&skip)) {
+        json_object_set_new(jrrule, "skip", json_string(buf_lcase(&skip)));
+    }
+    buf_free(&skip);
+
+    // firstDayOfWeek
+    static const char *const weekdays[7] = { "su", "mo", "tu", "we",
```
dilyanpalauzov: I guess here changing the definition to `static const char const [7][3]` would move the variable to read-only segment (just by using `[]` instead of `*` does it).

imap/jscalendar.c
```diff
@@ -0,0 +1,6708 @@
...
+    json_t *jval;
+    json_object_foreach(jnday, key, jval)
+    {
+        if (!strcmp("@type", key)) {
+            if (strcmpsafe("NDay", json_string_value(jval)))
+                jmap_parser_invalid(parser, key);
+        }
+        else if (!strcmp("day", key)) {
+            static strarray_t enums = STRARRAY_INITIALIZER;
+            if (!strarray_size(&enums)) {
```
dilyanpalauzov, line 4019: Identical strarrays `enums` are utilized on two places: here and on line 4153.

Can instead of `strarray_find(&enums, s, 0)` be used the variable `weekdays` from line 1317?

lib/charset.c
rsto: Right, this was left from a previous work-in-progress implementation.

rsto: Removed.

imap/jscalendar.c
rsto: Agreed

imap/jscalendar.c
rsto: Yes, done.

imap/jscalendar.c
rsto: That's correct, but we do want to treat struct buf as opaque, at least in new code.

imap/jscalendar.c
rsto: Usually, we follow the idiom that a `buf` is correctly initialized and the caller is responsible to reset it after use. In this function, I chose to flip that idiom the other way round, I can't recall why. Fixed.

imap/jscalendar.c
rsto: agreed

imap/jscalendar.c
rsto: I prefer to keep it as-is. I plan to experiment rewriting the validation functions for efficiency, most likely using gperf. Keeping the function variables scoped within each function will make that simpler.

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+    buf_setcstr(buf, makeuuid());
+    return buf_cstring(buf);
+}
+
+static void jsid_to_comp(icalcomponent *comp, const char *jsid)
+{
+    struct buf buf = BUF_INITIALIZER;
+
+    if (icalcomponent_isa(comp) == ICAL_PARTICIPANT_COMPONENT) {
+        // Omit JSID property if it is derived from CALENDAR-ADDRESS property.
```
dilyanpalauzov, line 1680: I suggest in this snippet to create a variable `buf`, which is not exposed in the whole function.  Likewise for the next `if {}` block.

lib/charset.c
```diff
@@ -4661,7 +4661,7 @@ EXPORTED int charset_decode_percent(struct buf *dst, const char *val)
 }
 
 // clang-format: off
-const char QSTRINGCHAR[256] = {
+static const char QSTRINGCHAR[256] = {
```
dilyanpalauzov, line 4664: This change was supposed to happen by https://github.com/cyrusimap/cyrus-imapd/pull/3742 in 2021, but in 2024 eventually it was discarded.

imap/jscalendar.c
dilyanpalauzov: Treating something as opaque and trying to add a null-byte at the end, when it is know that the end is `\0`-terminated, are two different things.

dilyanpalauzov:
With 618 files changed, reviewing this PR is unrealistic.

lib/charset.c
rsto: I'm not sure about the full chain of events, but it looks like it got manually merged by elliefm, see https://github.com/cyrusimap/cyrus-imapd/pull/3742#issuecomment-2392579350

I did not check when and how it got dropped again.

rsto:
> With 618 files changed, reviewing this PR is unrealistic.

Yes, as the description of this PR states:

>  No one can expect you to review over 600 files and thousands of lines of code.

I agree it's not good form, but we couldn't figure out how to deliver this change in smaller packages.

dilyanpalauzov:
> I agree it's not good form, but we couldn't figure out how to deliver this change in smaller packages.

A smaller package could be just the switch from `X-JMAP-PRIVACY` to `CLASS`.

As far as I see in jmap_ical.c the property `JMAPICAL_XPROP_PRIVACY` is still set/obtained unconditionally instead of property `CLASS`.  It would be good to have an imapd.conf option, which allows toggling it on systems, which do not have `X-JMAP-PRIVACY` properties.

lib/charset.c
dilyanpalauzov: The reason, why it was dropped, is that after the change was proposed until it was handled, three years have passed.

Makefile.am
```diff
@@ -1390,7 +1392,9 @@ imap_httpd_SOURCES += \
     imap/jmap_push.h \
     imap/jmap_quota.c \
     imap/jmap_util.c \
-    imap/jmap_util.h
+    imap/jmap_util.h \
```
dilyanpalauzov, line 1398: The line with jmap_util.h can be removed.  The purpose of listing `.h` files as _SOURCES in Makefile.am is not to enforce dependencies.  The dependencies on .h files are generated in/read from the `.deps` directories.  The purpose of including `.h` files in _SOURCES in Makefile.am is to include the files in the `make dist` tarball.  For this it is sufficient to mention the files once, and jmap_util.h is in `imap_libcyrus_imap_la_SOURCES` and in `imap_httpd_SOURCES`.

imap/jmap_util.c
```diff
@@ -4,6 +4,7 @@
 
 #include <config.h>
 
+#include <assert.h>
```
dilyanpalauzov, line 7: This the second `#include <assert.h>` in this file, the next is 3 lines lower.

Besides, most but not all of Cyrus source code uses `#include "assert.h"` instead of `#include <assert.h>`, which is supposed to mean that not `assert()` from libc is used, but `assert()` from cyrus-imap/lib/assert.c .

imap/http_convert.c
```diff
@@ -0,0 +1,236 @@
...
+/* SPDX-License-Identifier: BSD-3-Clause-CMU */
+/* See COPYING file at the root of the distribution for more details. */
+
+#include <config.h>
+
+#ifdef HAVE_UNISTD_H
+#include <unistd.h>
+#endif
+#include <string.h>
+#include <syslog.h>
```
dilyanpalauzov: Why is `#include <syslog.h>` present?

imap/http_convert.c
```diff
@@ -0,0 +1,236 @@
+/* http_convert.c -- Routines for converting media types over HTTP */
+/* SPDX-License-Identifier: BSD-3-Clause-CMU */
+/* See COPYING file at the root of the distribution for more details. */
+
+#include <config.h>
```
dilyanpalauzov, line 5: I suggest using `#include "config.h"`.  By passing options to the compiler, different search paths can be utilized for `#include "quoted"` and `#include <less-greater>`. This allows under circumstances to enforce preference to header files in the cyrus-imap source tree over system header files under `/usr/include` - when the same files is in both paths.

imap/http_convert.c
```diff
@@ -0,0 +1,236 @@
...
+{
+    size_t len;
+    char *p;
+
+    if (*tgt->path)
+        return 0; /* Already parsed */
+
+    /* Make a working copy of target path */
+    strlcpy(tgt->path, path, sizeof(tgt->path));
+    p = tgt->path;
```
dilyanpalauzov: I consider it easier to understand the source code, if this form is used:
```c
char *p = tgt->path;
```
As a rule of thumb, the longer and more complex a function is, the easier its understand is, when variables appear in the source code, when they are needed.

An anti-patters would be to define global variable in a function, when the variable is used only in the second-half of a loop-body in that function, cf. `acl_ptr` in `cyrus_acl_myrights()`.

rsto:
I removed compact-id support from this branch.

wolfsage:
nit on commit message:
```
charset.c: declare decoder character tables as static
This allows the compilter to move them to read-only memory and
only allocate them once.
```

compilter -> compiler

imap/jmap_util.c
```diff
@@ -172,7 +172,12 @@ EXPORTED json_t* jmap_patchobject_apply(json_t *val,
                 r = 0;
             }
             else {
-                r = json_object_set(it, ref, newval);
```
wolfsage, line 175: I feel like this commit message could be expanded:

"jmap_util.c: add KEEP_EXISTING flag to JMAP patchobject_apply"

In what case would we want to ignore part of the patch and keep existing data?

imap/caldav_util.c
```diff
@@ -350,22 +350,23 @@ static int validate_propupdates(icalcomponent *ical, icalcomponent *oldical,
                    prop ? icalproperty_get_value_as_string(prop) : "M");
     }
 
+    // Compare properties.
```
wolfsage, line 353: (Sorry no good place to comment on a commit...)

In "caldav_util.c: fix iterator bug when comparing property updates", only change is to caldav_util.c Is there a test that shows the problem (and the fix?)

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+
+    // Return early if there is nothing to do.
+    if (!ptrarray_size(&custom_vtzs)) goto done;
+
+    // Determine the timespan of the calendar component.
+    unsigned is_recurring = 0;
+    icalcomponent *comp = icalcomponent_get_first_real_component(ical);
+    if (!comp) goto done;
+
+    struct icalperiodtype guess_span = icalrecurrenceset_get_utc_timespan(ical,
```
wolfsage, line 345: If HAVE_GUESSTZ isn't enabled, we don't need to bother filling in guess_span either. I don't know how expensive it is, so maybe it doesn't matter. It also seems like we might construct &ctx->iana_tzid_by_custom_tzid but not use it with no guesstz?

It almost seems like it might be worth refactoring this method to not use the custom_vtzs variable, but instead fill everything into ctx->unguessable_vtimezones, and then return early if we don't HAVE_GUESSTZ. Otherwise, process unguessable_vtimezones and ptrarray_remove() any entries we successfully guess and put those into ianta_tzid_by_custom_tzid

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+            if (myicalproperty_has_name(prop, PROP_XJMAP_MAY_INVITE_OTHERS))
+                return true;
+            if (myicalproperty_has_name(prop, PROP_XJMAP_MAY_INVITE_SELF))
+                return true;
+            if (myicalproperty_has_name(prop, PROP_XJMAP_PRIVACY))
+                return true;
+            if (myicalproperty_has_name(prop, PROP_XJMAP_SENT_BY))
+                return true;
+            // XXX quirk: our previous jscalendar draft implementation
+            // erroneously used the X-APPLE-DEFAULT-ALARM annotation ino
```
wolfsage: nit: "ino" a typo?

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+        }
+
+        // It's a UUID, use it.
+        if (i == len) return true;
+    }
+
+    // Check for a reasonable length.
+    if (len < 24 || len > 64) return false;
+
+    // Check contents and charachter count.
```
wolfsage: nit: charachter -> character

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+    size_t chars_count[128] = { 0 };
+    size_t max_count = 0;
+    for (size_t i = 0; i < len; i++) {
+        char c = str[i];
+
+        // Only allow alphanumerics and commonly seen punctuation.
+        if (!isalnum(c) && !strchr("-_@.:", c)) {
```
wolfsage: I guess this is safe since cyrus never calls setlocale(LC_ALL, "") or equivalent? If it did run in any locale that's not C, isalnum could return true for, say, 254 (in ru_RU), and then we'd write to random memory below...

Maybe as a precaution we fail if c > 127?

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+    geouri_fini(geouri);
+
+    // geo:
+    if (strncmpsafe("geo:", str, 4)) return false;
+    str += 4;
+
+    // coord-a "," coord-b [ "," coord-c ]
+    int i;
+    for (i = 0; i < 3; i++) {
+        if ((geouri->coords[0] || geouri->coords[1])) {
```
wolfsage: can you have a coords[1] without a coords[0] ? Would this be better written if (i)?

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+            if (frac == num) break;
+
+            num = frac;
+        }
+
+        geouri->coords[i] = xstrndup(str, num - str);
+        str = num;
+    }
+
+    if (!geouri->coords[0] || !geouri->coords[1]) return false;
```
wolfsage: Same here - if coords[1] is set, coords[0] must be, right? So would this better as `if (!geouri->coords[1]) return false;`

Or is listing both just explicit for reader comprehension?

wolfsage: This return false can leak geouri->coords[0] if we don't have geouri->coords[1], it needs to call geouri_fini like below's false return

wolfsage (dismissed):
(Stopping for now, more another time)

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+        return false;
+    }
+
+    return true;
+}
+
+static struct icalgeotype geouri_to_icalgeo(const char *uri, bool *is_lossy)
+{
+    struct icalgeotype icalgeo = { 0 };
+    icalgeo.lat[0] = '0';
```
wolfsage, line 1019: Huh, wouldn't `struct icalgeotype icalgeo = { 0 }` initialize the whole struct to 0s already?

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+    }
+
+    *is_lossy = geouri.coords[2] || geouri.p;
+
+    size_t n = strlen(geouri.coords[0]);
+    if (n >= ICAL_GEO_LEN) {
+        n = ICAL_GEO_LEN - 1;
+        *is_lossy = true;
+    }
+    strncpy(icalgeo.lat, geouri.coords[0], n);
```
wolfsage: Okay, but seems like strlcpy would be clearer/better?

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+        }
+    }
+
+    //clang-format off
+    static struct
+    {
+        const char *name;
+        const char *icalname;
+    } byx[] = {
+        {"byMonthDay",     "BYMONTHDAY"},
```
wolfsage: nit: weird spacing on byMonthDay

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+
+    switch (rrule->week_start) {
+    case ICAL_NO_WEEKDAY:
+    case ICAL_MONDAY_WEEKDAY:
+        break;
+    default: {
+        json_object_set_new(
+            jrrule,
+            "firstDayOfWeek",
+            json_string(weekdays[rrule->week_start - ICAL_SUNDAY_WEEKDAY]));
```
wolfsage: Why the custom table instead of `icalrecur_weekday_to_string(rrule->week_start)` ?

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+                              icalcomponent *comp)
+{
+    json_t *jlocs = json_object_get(jobj, "locations");
+    if (!json_object_size(jlocs)) return;
+
+    struct jmap_parser parser = JMAP_PARSER_INITIALIZER;
+    jmap_parser_push(&parser, "locations");
+
+    const char *mainloc_id =
+        json_string_value(json_object_get(jobj, "mainLocationId"));
```
wolfsage, line 2209: I can't seem to use this (mainLocationId) in a CalendarEvent/set on this branch. Should I be able to? Should we have tests for this?

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+                    icalproperty_add_parameter(organizer,
+                                               icalparameter_new_sentby(uri));
+                }
+            }
+
+            if (attendee) {
+                jval = json_object_get(jpart, "scheduleSequence");
+                if (json_is_integer(jval) && json_integer_value(jval) >= 0) {
+                    struct buf buf = BUF_INITIALIZER;
+                    buf_printf(&buf, "%lld", json_integer_value(jval));
```
wolfsage: Probably fine but all of these llds with json_integer_value should probably be using JSON_INTEGER_FORMAT?

imap/jscalendar.c
```diff
@@ -0,0 +1,6707 @@
...
+    json_object_foreach(jpatch, pkey, pval) {
+        for (int i = 0; skip_prefixes[i]; i++) {
+            if (!strncmp(pkey, skip_prefixes[i], strlen(skip_prefixes[i]))) {
+                strarray_append(&del_keys, pkey);
+                break;
+            }
+        }
+    }
+    for (int i = 0; i < strarray_size(&del_keys); i++) {
+        json_object_del(jpatch, strarray_nth(&del_keys, i));
```
wolfsage, line 3055: How would one have a string that's a prefix of one of these strings? (I also don't understand the intent here from 3.3.4.  recurrenceOverrides of https://datatracker.ietf.org/doc/draft-ietf-calext-jscalendarbis/

"
      A pointer in the PatchObject MUST be ignored if it starts with one
      of the following prefixes:
"

Why a prefix and not an exact match? CalendarEvent/set doesn't allow, for example:
```
recurrenceOverrides => {
  $date => {
    uidfoobar => "...",
  }
```

which seems right to me, but this code and the spec want to allow uidaslkfjsaldkfasd?

wolfsage (commented):
More later

imap/jmap_util.c
rsto: I updated the commit message, here copied verbatim:

    A patch object that patches an existing JSON key replaces the value
    of that key with whatever value is set in the patch. But a patch
    object coming from an untrusted source, such as the JSPROP property
    in iCalendar data, should not be able to replace existing values.

    This patch adds the KEEP_EXISTING flag to allow for this. It instructs
    the patching function to only apply patches for JSON keys which do
    not exist in the patched JSON object.

imap/caldav_util.c
rsto: This bug surfaced when testing the new JSCalendar conversion code with the existing JMAPCalendars tests. It was related to the CLASS or maybe the COLOR property tests. This branch since then has split and squashed commits, that's why the change in caldav_util.c is separate from the test that demonstrates the issue.

imap/jscalendar.c
rsto: We will pull guesstz into the Cyrus repo soon and it's very likely to become mandatory for the new JSCalendar code. We want to reduce the number of build options.

Meanwhile, I did slightly refactor this function to fill unguessable_vtimezones directly if the Cyrus build doesn't enable guesstz.

imap/jscalendar.c
rsto: Fixed

imap/jscalendar.c
rsto: Fixed

imap/jscalendar.c
rsto: Yep, I replaced this with `n_coords` (instead of `i`).

imap/jscalendar.c
rsto: Also refactored to use `n_coords`

imap/jscalendar.c
rsto: correct, but `'0'` is decimal 48

imap/jscalendar.c
rsto: Right, refactored to use `strlcpy`

imap/jscalendar.c
rsto: fixed

imap/jscalendar.c
rsto: Mostly to avoid the heap allocation when lowercasing the string from libical. But in the grand scheme of things, these heap allocations are a drop in the bucket and it allows us to keep mapping enum to string in libical, so I changed it to use `icalrecur_weekday_to_string`.

imap/jscalendar.c
rsto: This is part of the RFC tests: https://www.ietf.org/archive/id/draft-ietf-calext-jscalendar-icalendar-22.html#section-2.3.25

rsto: As to why you couldn't set in on `CalendarEvent/set`: turns out it wasn't registered in the gperf tables. Now fixed. This and similar case should all surface once we hook the RFC tests into Cassandane.

rsto: And I also hooked mainLocationId now into the existing JMAPCalendars location tests.

imap/jscalendar.c
rsto: done. The compiler typically catches mismatch between type and format literal, but let's use the proper format string.

imap/jscalendar.c
rsto: The spec is making that not clear enough, I will update its phrasing. When it currently states that the pointer starts with a prefix it means that the first token in the JSON pointer (RFC 6901) starts with this prefix. So it instructs to ignore both patches that patch the exact prefix, or the prefix followed by `/`. This should be phrased better and probably also split in exact prefix for non-complex types (such as `uid`) and nested types (such as `relatedTo`).

In any case, I've update the logic to check for the exact match or the `/` character.

As to why allow unknown properties: this is for compatibility with JSContact and likewise to follow what iCalendar and vCard do with `iana-token`. There's always the conflict between either being permissive with unknown elements (like JSCalendar and JSContact), or restrictive (like JMAP Email). I see pros and cons for both, but we decided that at least JSCalendar and JSContact should follow the same design either way.

rsto:
@wolfsage I updated the PR with your review comments.

imap/jscalendar.c
rsto: Keeping in scope of the whole function allows to reuse the buffer, potentially avoiding a heap allocation. It's a common idiom in a lot of places of the Cyrus codebase.

imap/jmap_util.c
rsto: Agreed, it's all over the place. I count 22 files including stdlib assert.h vs. 60 files including `"assert.h"`. We should clean that up, but this PR won't be it.

imap/http_convert.c
rsto: Left-over from a refactor. Removed.

imap/http_convert.c
rsto: Sounds good, but almost all Cyrus source files include `<config.h>` right now. So like for `assert.h`, this sounds reasonable but this PR isn't the place for it.

imap/http_convert.c
rsto: agreed. fixed for `http_convert.c`

imap/jscalendar.c
wolfsage: hah! Wow, missed that. THanks

imap/jscalendar.c
wolfsage: The leak is still here.

rsto: this also checks `n_coords` now

rsto: sorry, removed it now

imap/http_caldav.c
```diff
@@ -2679,6 +2692,48 @@ static int caldav_get(struct transaction_t *txn, struct mailbox *mailbox,
...
+    r = caldav_parse_path(txn->req_uri->path, &txn->req_tgt, &txn->error.desc);
+
+    /* Make sure method is allowed (only allowed on child of home-set) */
+    if (!txn->req_tgt.collection || txn->req_tgt.resource ||
+        (txn->req_tgt.collection[0] == 'C' &&
+         txn->req_tgt.collen <= CONV_JMAPID_SIZE + 1)) {
+        txn->error.precond = CALDAV_LOCATION_OK;
+        return HTTP_FORBIDDEN;
+    }
+    else if (r) {
```
dilyanpalauzov: `else` is not necessary.  After the next two `break;` the lines contain only spaces.

imap/http_convert.c
```diff
@@ -0,0 +1,232 @@
...
+        {NULL, NULL}, /* PATCH        */
+        {&meth_post, NULL}, /* POST   */
+        {NULL, NULL}, /* PROPFIND     */
+        {NULL, NULL}, /* PROPPATCH    */
+        {NULL, NULL}, /* PUT          */
+        {NULL, NULL}, /* REPORT       */
+        {NULL, NULL}, /* SEARCH       */
+        {NULL, NULL}, /* TRACE        */
+        {NULL, NULL}, /* UNBIND       */
+        {NULL, NULL}        /* UNLOCK       */
```
dilyanpalauzov, line 61: Here the first entry - for `/* ACL */` - has less indentation, than the remailing elements.

The last entry - for UNLOCK - has too many spaces before the start of the comment.

imap/http_convert.c
```diff
@@ -0,0 +1,232 @@
...
+    char *resp_payload = NULL;
+    jscal_cfg_t jscal_cfg = {
+        .emailalert_default_uri = httpd_userid,
+        .use_icalendar_convprops = true,
+        .no_quirk = true,
+    };
+    int ret = 0;
+
+    /* Parse the request body */
+    ical = icalparser_parse_string(buf_cstring(&txn->req_body.payload));
```
dilyanpalauzov, line 119: Instead of
```c
icalcomponent *ical = NULL;
ical = icalparser_parse_string(buf_cstring(&txn->req_body.payload));
```
it is simpler to use
```c
icalcomponent *ical = icalparser_parse_string(buf_cstring(&txn->req_body.payload));
```

dilyanpalauzov:
https://www.cyrusimap.org/imap/developer/compiling.html says LIBUUID is a required build dependency, but lib/util.[ch] behave differently, depending on whether `./configure` has found LIBUUID.

imap/jmap_util.c
wolfsage: Awesome, thanks

imap/jscalendar.c
wolfsage: Ah, great, thanks

imap/jscalendar.c
wolfsage: I see you fixed this. Great!

wolfsage (commented):
I'd still like to do more review but won't hold things up!

rjbs:
@rsto While we were discussing the merge of my changes to this branch, @ksmurchison was writing a PR against mine.  I have taken the liberty (after reviewing it) of bringing it into this branch in the same way: rebased and tucked under the top three commits — so I can mint a tag for testing.

imap/jscalendar.c
dilyanpalauzov: > That's correct, but we do want to treat struct buf as opaque, at least in new code.

How about using `inline buf_base(struct buf *buf)` ⇔ `return buf->s;`?

rjbs:
Rebased, removing the forbidding of deleting default calendars.

PR #5915, opened 2026-04-20
cassandane: no more manual sync of bitfield imapoptions

elliefm:
Cassandane::Config has a table of the imapd.conf bitfield options and the values they accept.  The table is manually maintained, which means it's usually out of date because we forget to update it.

This PR replaces the table with one that's generated during build, using the new imapoptions tools.

The generated table is structured identically to the old one.  This has some limitations, but we're not currently using it in ways where those limitations would be a problem.  For example, Cassandane::Config doesn't know the bits' _values_, nor can it differentiate between real bit names and aliases.  It would not be hard to refactor if we wanted to, but that is not part of this PR.


rsto (commented):
I'm not enough into Perl to have opinions on the actual bitfield implementation. I will remove myself from the reviewer list.

ksmurchison:
I feel the same as rsto.  This should probably be review by rjbs or wolfsage

tools/lib/Cyrus/IMAPOptions/App/Command/bitfields.pm
```diff
@@ -0,0 +1,90 @@
...
+sub execute
+{
+    my ($self, $opt, $args) = @_;
+
+    my $imapoptions = Cyrus::IMAPOptions->new(
+        from_path => $args->[0],
+        forbid_unreleased => $opt->forbid_unreleased,
+    );
+
+    $imapoptions->check_unreleased($self->app->arg0);
```
wolfsage, line 35: What's `$args->[0]` vs `$self->app->arg0` ?

wolfsage (approved):

tools/lib/Cyrus/IMAPOptions/App/Command/bitfields.pm
rjbs: It's "`$0` captured at time of App::Cmd creation, and basenamed".  App::Cmd-ism.

- @ARGV holds the elements of `argv` *starting with 1*
- Getopt::L::D modifies `@ARGV`, removing elements consumed into switches
- `$cmd->args` is the remaining items in `@ARGV`
- $0 holds `argv[0]`
- $0 is sometimes assigned to, so you lose what it was originally
- $0 might be /usr/bin/yourface and "the program name" is really yourface

check_unreleased wants to say "imapoptions: you ate dirt", where "imapoptions" is the program you ran.

tools/lib/Cyrus/IMAPOptions/App/Command/bitfields.pm
```diff
@@ -0,0 +1,90 @@
...
+package Cassandane::Config::Bitfields;
+use warnings;
+use strict;
+
+use Exporter qw(import);
+our @EXPORT_OK = qw();
+
+# XXX this doesn't differentiate real bit values from aliases
+our %bitfields = (
+END_HEADER
```
rjbs, line 57: Not a blocker to merge, but I will note that an indented heredoc can help with skimming here.

```perl
sub foo {

    my $pm = <<~'END_HEADER';
    # auto-generated
    use strict;

    sub {...}
    END_HEADER

}
```

(example)

rjbs (approved):

tools/lib/Cyrus/IMAPOptions/App/Command/bitfields.pm
elliefm: Thanks, not yet accustomed to those being indentable! :sweat_smile: 

I'll make a change along those lines to all the commands at once, in its own PR.

PR #5916, opened 2026-04-20
imapoptions: make individual options hyperlinkable in docs

elliefm:
Primarily, this PR adds an `:imapdconf:` sphinx role/directive pair that we can use for hyperlinking directly to individual imapd.conf options.  Role and directive are sphinx terms, see their docs if you care about the details (you probably don't).

It also changes all the places where we previously embedded chunks of imapd.conf(5) directly into pages -- we now link to those options instead (which aids discovery of other related options that may not have been embedded).  The awkward embedding mechanism has been removed.

Finally, it updates a small selection of the documentation to use the new role.  I deliberately kept these changes as minimal as possible to reduce PR/code review clutter.  Think of these changes as examples of the kinds of changes we can now make, not as a comprehensive documentation refresh.

**Reviewers:** I suggest stepping through this one commit by commit.  Otherwise the substantive changes will be lost amongst the prose changes.

---

Use the new role like this (for example):

```
:imapdconf:`reverseacls`
```

from anywhere in documentation to insert a link to the documentation for the `reverseacls` imapd.conf option.  "Anywhere in documentation" means any rst file under docsrc, and **it also includes changes files, lib/imapoptions files, release notes, etc**.

If you link to an option that doesn't exist, sphinx-build will complain, and the CI doc gen job will fail.


rsto (approved):
This builds the documentation cleanly on my setup and looks good to me..

rjbs (approved):
Yeah, nice!

There was a _lot_ here, so I may have missed some kind of silly mistake, but the bones look really good, and it's nice to see how much this simplifies things!

ksmurchison (approved):

PR #5918, opened 2026-04-21
reject unknown "kind" components, cope with bad JSPROP

rjbs:
We're doing two jobs here:

1. stop accepting components with an unknown "kind", so we stop writing JSPROP properties that patch nonexistent patches
2. if we see one of those JSPROP properties, patching something that's null, don't turn the whole card into null, just discard the patch and log it

### DIFF
--- a/cassandane/tiny-tests/JMAPContacts/card_set_create_addresses_unknown_kind
+++ b/cassandane/tiny-tests/JMAPContacts/card_set_create_addresses_unknown_kind
@@ -1,0 +1,53 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_card_set_create_addresses_unknown_kind
+    :min_version_3_9
+    ($self)
+{
+    my $jmap = $self->{jmap};
+
+    my $id = 'ae2640cc-234a-4dd9-95cc-3106258445b9';
+
+    # "streetAddress" is not a valid AddressComponent.kind per RFC 9553
+    # section 2.5.1.1; the street name kind is "name".  An unknown kind
+    # should be rejected with invalidProperties rather than silently
+    # stored as a JSPROP fallback that no CardDAV client can display.
+    my $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            create => {
+                "1" => {
+                    '@type' => 'Card',
+                    version => '1.0',
+                    uid => $id,
+                    name => { full => 'John Doe' },
+                    addresses => {
+                        a0 => {
+                            '@type' => 'Address',
+                            components => [
+                                {
+                                    kind => 'streetAddress',
+                                    value => '42 Rue de Rivoli'
+                                },
+                                {
+                                    kind => 'locality',
+                                    value => 'Paris'
+                                }
+                            ],
+                            countryCode => 'FR'
+                        }
+                    }
+                }
+            }
+        }, 'R1']
+    ]);
+
+    $self->assert_null($res->[0][1]{created}{1});
+    $self->assert_not_null($res->[0][1]{notCreated}{1});
+    $self->assert_str_equals("invalidProperties",
+                             $res->[0][1]{notCreated}{1}{type});
+    $self->assert_deep_equals(
+        ["addresses/a0/components[0]/kind"],
+        $res->[0][1]{notCreated}{1}{properties}
+    );
+}
--- a/imap/jmap_contact.c
+++ b/imap/jmap_contact.c
@@ -7977,8 +7977,22 @@ static json_t *jmap_card_from_vcard(const char *userid,
         json_t *patched = jmap_patchobject_apply(jcard, crock.patch, NULL, 0);
 
         json_decref(crock.patch);
-        json_decref(jcard);
-        jcard = patched;
+        if (patched) {
+            json_decref(jcard);
+            jcard = patched;
+        }
+        else {
+            /* The JSPTR of at least one of the JSPROP properties pointed into
+             * a JSContact subtree that the vCard-to-JSContact conversion did
+             * not reconstruct, so the patch could not be applied. Keep the
+             * unpatched card as defined in RFC 9555, Section 3.2.1.  The
+             * JSPROP contents are preserved in the stored vCard and a client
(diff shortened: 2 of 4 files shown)


ksmurchison (approved):
LGTM

cassandane/tiny-tests/JMAPContacts/card_set_create_addresses_unknown_kind
```diff
@@ -0,0 +1,53 @@
...
+            }
+        }, 'R1']
+    ]);
+
+    $self->assert_null($res->[0][1]{created}{1});
+    $self->assert_not_null($res->[0][1]{notCreated}{1});
+    $self->assert_str_equals("invalidProperties",
+                             $res->[0][1]{notCreated}{1}{type});
+    $self->assert_deep_equals(
+        ["addresses/a0/components[0]/kind"],
```
rsto, line 50: No change required but a nit: this is not valid JSON pointer format (RFC 6901), that would be `components/0/kind`. Then again JMAP Core doesn't specify the format in which nested properties should be reported, so the format in this test is as good as any.

imap/jmap_contact.c
```diff
@@ -7977,8 +7977,22 @@ static json_t *jmap_card_from_vcard(const char *userid,
...
 
         json_decref(crock.patch);
-        json_decref(jcard);
-        jcard = patched;
+        if (patched) {
+            json_decref(jcard);
+            jcard = patched;
+        }
+        else {
+            /* A JSPROP's JSPTR pointed into a JSContact subtree that
```
rsto, line 7985: The implementation looks correct to me but this comment makes me double-guess: According to RFC 9555 [Section 3.2.1.](https://datatracker.ietf.org/doc/html/rfc9555#name-jsprop) the implementation should ignore **all** JSPROP patches if at least one JSPROP patch is wrong. That's because a PatchObject is invalid if at least one patch is invalid. But this comment kind of suggests as if it only ignores one specific JSPROP. I don't think it does, though?

rsto (commented):
Generally looks good to me, but one question for clarification (for which most likely only a comment must be rephrased rather than code changed).

Additional remark: It's great for review to have separate commits that build on each other, but for merging I think we should squash and some commit messages are very verbose (specifically the one for "don't silently drop ContactCard/get results when a JSPROP fails").

cassandane/tiny-tests/JMAPContacts/card_set_create_addresses_unknown_kind
rjbs: not merely `components/0/kind` but `addresses/a0/components/0/kind` right?  I'll change it -- otherwise it confuses the wise reader

rjbs:
> Additional remark: It's great for review to have separate commits that build on each other, but for merging I think we should squash and some commit messages are very verbose (specifically the one for "don't silently drop ContactCard/get results when a JSPROP fails").

Yeah, I'm not particularly rigorous on this but my usual practices are either:

1. commit test
2. commit fix

OR

1. commit test and fix

The first is useful when developing, but if we prefer I could try to always convert it the second form.  It's an easy step before finalization!

imap/jmap_contact.c
rjbs: I am treading very carefully here, because I feel like I'm trying to tell you what you wrote! :D  But the good news is that we agree on what should happen.

9555 says that "**all** JSPROP properties in a vCard together form **a** PatchObject".  So, there is only one patch.  I don't see clear language that it must only be applied atomically, but I think we'd both agree that's correct.

This matches the code:  we generate one patch and we either apply it or don't.

If you propose alternate language for the comment, I can take it, but I think this language seems clear and correct *to me* such that I can't see how to improve it!

rjbs:
@rsto I've squashed and tried to reduce the verbosity of the commit.  This is a bit tricky for me, because I have a more shallow understanding of the code, so the detail helps *me* understand, but I'm not necessarily the target audience!  Anyway:  Give it another look?

cassandane/tiny-tests/JMAPContacts/card_set_create_addresses_unknown_kind
rjbs: Oh, right, this bit is being generated by the patch applicator.  Something to fix later and elsewhere, I think.

rsto: Yeah, sorry, should be `addresses/a0/components/0/kind`. But as you said, it's coming from the JMAP tooling code so let's keep it as-is for now.

imap/jmap_contact.c
rsto: I'm fine with the comment as-is, maybe it's just me who second-guessed it's meaning. I guess I would have written this slightly differently (changes higlighted) to make this more clear:

**The JSPTR of at least one of the JSPROP properties** pointed into a JSContact subtree that the vCard-to-JSContact conversion did not reconstruct, so the patch could not be applied.  Keep the unpatched card a**s defined in RFC 9555, Section 3.2.1**; the JSPROP contents are preserved in the stored vCard and a client that requests them directly can still retrieve them.

I think there is no need to highlight that the Card object isn't dropped - that was a bug anyway.

rsto (approved):

rjbs:
@rsto I put your comment text in, nearly verbatim.

elliefm:
> Yeah, I'm not particularly rigorous on this but my usual practices are either:
> 
>     1. commit test
> 
>     2. commit fix
> 
> 
> OR
> 
>     1. commit test and fix
> 
> 
> The first is useful when developing, but if we prefer I could try to always convert it the second form. It's an easy step before finalization!

If the new/updated test will fail without the fix, the former approach can throw false positives while you're bisecting something unrelated later.  So usually I find it cleaner to commit them together.  But sometimes that makes each a distraction from the other (like if it's a complicated new test that needs review in its own right, rather than a small correction to an existing one).  So there are also times when I will commit the fix first and then the tests after, so they each get their own narrative space, while not introducing a failing test.

PR #5921, opened 2026-04-21
xapian_wrap.cpp: fix phrase search queries for CJK text

rsto:
This fixes a bug where phrase queries for CJK text were queried as loose term search, e.g. the order of terms was irrelevant for a match.

This patch also disables a workaround for CJK text, unless the queried Xapian index has a legacy version. This is because the workaround is not compatible with phrase search queries. Rather than rewriting the workaround to deal with phrase search, the Xapian index should get reindexed in case of false positives.


ksmurchison (approved):
I don't know much, if anything, about how Xapian works, but if this passes CI and solves the reported customer issue, LGTM

PR #5922, opened 2026-04-21
index.c: prohibit APPEND/COPY/MOVE into calendar/addressbook-home-set

ksmurchison:
These folders should never contain messages

### DIFF
--- a/cassandane/tiny-tests/Append/append_insert_prohibited
+++ b/cassandane/tiny-tests/Append/append_insert_prohibited
@@ -77,6 +77,27 @@ EOF
 
     $imaptalk->copy(1, 'INBOX.Snoozed');
     $self->assert_str_equals('no', $imaptalk->get_last_completion_response());
+
+    xlog $self, "Reconnect with magic +DAV";
+    $imaptalk->logout();
+    $imaptalk = $self->{store}->get_client(NoLogin => 1);
+    $imaptalk->login("cassandane+dav", "secret");
+
+    # APPEND/COPY into calendar/addressbook-home-set is always prohibited
+    xlog $self, "Attempt to APPEND/COPY into calendar/addressbook-home-set";
+    $imaptalk->append('INBOX.#calendars', $mimeMessage);
+    $self->assert_str_equals('no', $imaptalk->get_last_completion_response());
+
+    $imaptalk->append('INBOX.#addressbooks', $mimeMessage);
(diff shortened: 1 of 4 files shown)


imap/index.c
rsto, line 7518: Should the comment from line 7494 (`/* prohibit inserting into \Snoozed mailbox */`) move here? It looks as if the code previously assumed only the "Snoozed" mailbox contains alarms, but now it does not.

rsto (approved):
LGTM, only two nits: a comment for a deleted line now seems to be at the wrong place. And the commit message (and the PR title/description) does not mention the changes for the "Snoozed" mailbox.

wolfsage:
I don't think this goes far enough. Without +DAV you shouldn't be able to append/copy into non-mail mailboxes at all, right?

If you adjust your test to create a #sieve folder, you can append or copy into that and cyrus gets very unhappy about it.

I haven't tried with #jmap, #jmapnotifications (or whatever it is..) etc, but all of those should be blocked, yeah? They don't come back from x list "" * so IMO append/copy/etc should not see them

ksmurchison:
> I don't think this goes far enough. Without +DAV you shouldn't be able to append/copy into non-mail mailboxes at all, right?

@wolfsage This should be fixed now

rjbs (approved):

cassandane/tiny-tests/Append/append_insert_prohibited
```diff
@@ -0,0 +1,101 @@
...
+    $admintalk->create("user.cassandane.#addressbooks",
+                       ['TYPE', 'ADDRESSBOOK']) || die $@;
+    $admintalk->create("user.cassandane.#addressbooks.Default",
+                       ['TYPE', 'ADDRESSBOOK']) || die $@;
+
+    my $imaptalk = $self->{store}->get_client();
+
+    $imaptalk->create('INBOX.Snoozed', "(USE (\\Snoozed))") || die $@;
+
+    xlog $self, "Append duplicate messages with different internaldates";
```
wolfsage: nit: This message feels like copypasta

wolfsage (approved):

cassandane/tiny-tests/Append/append_insert_prohibited
ksmurchison: Yep.  I will fix

PR #5923, opened 2026-04-21
cyr_expire.c: --prune-userflags does NOT take an argument

ksmurchison:
This PR previously had changes to reconstruct which have since been deemed unnecessary.  I have stripped this down to just the fix to cyr_expire

### DIFF
--- a/imap/cyr_expire.c
+++ b/imap/cyr_expire.c
@@ -857,7 +857,7 @@ static int parse_args(int argc, char *argv[], struct arguments *args)
         { "no-conversations", no_argument, NULL, 'c' },
         { "help", no_argument, NULL, 'h' },
         { "prefix", required_argument, NULL, 'p' },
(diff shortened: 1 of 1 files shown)


brong:
Looks good - but also, this seems very amenable to a test case (maybe one for cyr_expire too!)  You can read the permanentflags when selecting the mailbox to see if the flag has been removed.

rsto (commented):
LGTM - but agree with Bron that this should get a test.

ksmurchison:
> LGTM - but agree with Bron that this should get a test.

It turns out that the changes to reconstruct aren't necessary as `cyr_expire -t` works in all cases and is a better spot for this work.  We already have `Flags.expunge_removeflag` to test this.  All that remains in this PR is the fix for `--prune-userflags`

rsto (approved):

PR #5927, opened 2026-04-23
mboxname.c Don't re-encode already mUTF7-encoded detail part of localpart

### LINKED ISSUE #5925: Delivery doesn't work on non-ascii mailboxes
I've tried sending emails to our non-ascii shared folders but I came across the following problem: commit 369009f1d7289fa5761df0ec0055dcad7beb06ea always encodes the detail part of a recipient from utf8 to mutf7, but as far as I know lmtpd only supports mutf7 encoding, any 8bit characters in rcpt command will be rejected. Let's say for example we have the following shared folder `&A5EDkQORA5EDkQ-` to deliver to this the recipient must be `bb+&A5EDkQORA5EDkQ-` but when this gets encoded it will become `&-A5EDkQORA5EDkQ-` (notice the escaped &) and delivery will fail.

Am I missing something here and it's possible to do this in another way or is it broken? Currently we have this fixed by disabling the encode on lmtpd and smmapd. I'd like some insight into why the commit was made in the first place so I can do a proper fix. @ksmurchison perhaps you can comment on this?

ksmurchison:
If we received an email for foo+&JgM-@example.com, we were re-encoding the detail as &-JgM-

This fixes #5925

### DIFF
--- a/cassandane/tiny-tests/Delivery/plus_address_mutf7
+++ b/cassandane/tiny-tests/Delivery/plus_address_mutf7
@@ -1,0 +1,64 @@
+#!perl
+use Cassandane::Tiny;
+use utf8;
+
+sub test_plus_address_mutf7
+    ($self)
+{
+    xlog $self, "Testing behaviour of plus addressing with mUTF7 mailbox names";
+
+    # IMAPTalk will mUTF7-encode mailbox name in CREATE/SELECT
+    my $imaptalk = $self->{store}->get_client();
+    $self->{store}->set_fetch_attributes('uid');
+
+    xlog $self, "Create folders";
+    $self->setup_mailbox_structure($imaptalk, [
+        [ 'create' => [ "foo", "☃", "a&b" ] ]
+    ]);
+
+    my %msgs;
+    my @testCases = ({
+        uid => 1,
+        plusaddr => "foo",
+        mailbox => "foo",
+    },
+    {
+        uid => 1,
+        plusaddr => "&JgM-", # mUTF7-encoded
+        mailbox => "☃",
+    },
+    {
+        uid => 1,
+        plusaddr => "a&b",
+        mailbox => "a&b",
+    },
+# Can't test UTF-8 plus address until we support SMTPUTF8
+#    {
+#        uid => 2,
+#        plusaddr => "☃",
+#        mailbox => "☃",
+#    },
+    {
+        uid => 2,
+        plusaddr => "a&-b", # mUTF7-encoded
+        mailbox => "a&b",
+    });
+
+    foreach my $tc (@testCases) {
+        my $uid = $tc->{uid};
+
+        if (!$msgs{$uid}) {
+            xlog $self, "Create message $uid";
+            $msgs{$uid} = $self->{gen}->generate(subject => "Message $uid");
+            $msgs{$uid}->set_attribute(uid => $uid);
+        }
+
+        xlog $self, "Deliver message $uid";
+        $self->{instance}->deliver($msgs{$uid},
+                                   user => "cassandane+$tc->{plusaddr}");
+
+        xlog $self, "Check that the message made it";
+        $self->{store}->set_folder($tc->{mailbox});
+        $self->check_messages(\%msgs, check_guid => 0, keyed_on => 'uid');
+    }
+}
--- a/imap/mboxname.c
+++ b/imap/mboxname.c
@@ -386,6 +386,9 @@ EXPORTED mbname_t *mbname_from_userid(const char *userid)
     return mbname;
 }
 
+static const char b64mbox[] =
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+,";
+
 EXPORTED mbname_t *mbname_from_recipient(const char *recipient, const struct namespace *ns)
 {
     mbname_t *mbname = xzmalloc(sizeof(mbname_t));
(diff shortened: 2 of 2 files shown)


cassandane/tiny-tests/Delivery/plus_address_mutf7
```diff
@@ -0,0 +1,31 @@
...
+use utf8;
+
+sub test_plus_address_mutf7
+    :NoAltNameSpace
+    ($self)
+{
+    xlog $self, "Testing behaviour of plus addressing where case matches";
+
+    # IMAPTalk will mUTF7-encode the name
+    my $folder = "INBOX.☀";
```
rjbs: I was so confused for a bit.  I didn't realize this was BLACK SUN WITH RAYS, and thought it was plain old ASTERISK.  Then the test was is-ascii! :)

imap/mboxname.c
```diff
@@ -410,20 +410,34 @@ EXPORTED mbname_t *mbname_from_recipient(const char *recipient, const struct nam
...
 
-        /* Encode mailbox name in IMAP UTF-7 */
-        charset_t cs = charset_lookupname("utf-8");
-        char *detail =
-            charset_to_imaputf7(plus+1, strlen(plus+1), cs, ENCODING_NONE);
+        /* Does 'detail' contain non-ASCII (UTF-8) chars? */
+        for (const char *cp = plus+1; *cp; cp++) {
+            if (!isascii(*cp)) {
+                need_encode = 1;
+                break;
```
rjbs: Is this enough of a test?  I think it's close to right, but I *think* `detail` could contain `&`.

Local part is (roughly) `*atext` and `atext` contains `&`, and so you'd need to encode `&`… but I could be confused about exactly how this gets used in Cyrus land as opposed to interchange.

rsto: Similar to @rjbs comment, I do not think this test is enough (or even needed).

It is not enough because [RFC 3501, Section 5.1.3](https://datatracker.ietf.org/doc/html/rfc3501#section-5.1.3) states:

> In modified UTF-7, printable US-ASCII characters, except for "&", represent themselves; that is, characters with octet values 0x20-0x25 and 0x27-0x7e.  The character "&" (0x26) is represented by the two-octet sequence "&-".
> 
> All other characters (octet values 0x00-0x1f and 0x7f-0xff) are represented in modified BASE64, with a further modification from [[UTF-7](https://datatracker.ietf.org/doc/html/rfc3501#ref-UTF-7)] that "," is used instead of "/".  Modified BASE64 MUST NOT be used to represent any printing US-ASCII character which can represent itself.

But other than saving the CPU time, the test might not be needed at all. The IMAP UTF-7 encoded name of a name that does not need encoding should stay the same, right? So this codepath could always encode and keep the decision when to encode internal to the charset library?

cassandane/tiny-tests/Delivery/plus_address_mutf7
rsto: I agree, this character is confusing. It does not look as if the test has to use this character, maybe you can change it to use a character that's more clearly not an ASCII character?

rsto (commented):
Basically I have the same comments as Rik but added my 2 cents to it :)

imap/mboxname.c
ksmurchison: As stated in the commit message, the whole point of the change is that if the +detail is already mUTF7, that blindly encoding it (as we currently do) will encode any '&' that are part of the original encoding (&JgA-).

I guess I need to change the code to check if the given string is already mUTF7 encoded

rsto: Sorry, I was jumping on the existing comment, I now see what you do here. I agree that checking it its already correctly encoded it the right thing.

cassandane/tiny-tests/Delivery/plus_address_mutf7
ksmurchison: I changed to using the snowman unicode char.

imap/mboxname.c
ksmurchison: I changed the code to look for a mUTF7-encoded sequence

cassandane/tiny-tests/Delivery/plus_address_mutf7
```diff
@@ -0,0 +1,56 @@
+#!perl
+use Cassandane::Tiny;
+use utf8;
+
+sub test_plus_address_mutf7
+    ($self)
+{
+    xlog $self, "Testing behaviour of plus addressing with mUTF7 mailbox names";
+
+    # IMAPTalk will mUTF7-encode the names
```
rsto: That test deserves either becoming a table-driven test (like `JMAPEmail/email_set_groupaddr`, or any other test in the JMAPEmail and JMAPCalendars test suites that uses a variable named `@testCases`).

Or alternatively this could get split into multiple subtests, like in `JMAPContacts.contact_member_uid_matrix`.

rsto (approved):
Implementation LGTM. The test code could get rewritten to a table-driven test.

cassandane/tiny-tests/Delivery/plus_address_mutf7
ksmurchison: Done

cassandane/tiny-tests/Delivery/plus_address_mutf7
```diff
@@ -0,0 +1,71 @@
...
+
+    xlog $self, "Create folders";
+    $self->setup_mailbox_structure($imaptalk, [
+        [ 'create' => [ $ascii, $snowman, $ampersand ] ],
+    ]);
+
+    my %msgs;
+    my @testCases = ({
+        uid => 1,
+        plusaddr => $ascii,
```
rsto: why not just declare the verbatim string values here, rather than the indirection with a variable?

rsto (approved):

cassandane/tiny-tests/Delivery/plus_address_mutf7
ksmurchison: The only reason is because three of them are used in multiple places (include mailbox creation)

rsto: I would have declared them multiple times, they are not always the same in both places. But I'm fine with it as-is, too, that's why I already approved.

rjbs (approved):

PR #5928, opened 2026-04-24
tools/imapoptions.pl: use indented heredocs for readability

elliefm:
Just indents the heredocs

In hide whitespace changes mode, the diff should just show the heredoc headers changing from `<< ` to `<<~`.


rjbs:
Nice.

rjbs (approved):

PR #5932, opened 2026-04-24
caldav_util.c: Limit msg Subject to the first 1k chars of SUMMARY

ksmurchison:
Otherwise, if the SUMMARY is huge, we could end up with more than 1000 header lines and we will fail to find the Content-Disposition header, and therefore have no resource URI.

### DIFF
--- a/imap/caldav_util.c
+++ b/imap/caldav_util.c
@@ -1137,8 +1137,22 @@ EXPORTED int caldav_store_resource(struct transaction_t *txn, icalcomponent *ica
     const char *summary = icalcomponent_get_summary(comp);
     if (summary) {
         int force = !!strchr(summary, '\n'); // force encoding if embedded LF
+        size_t limit = 1024, len = strlen(summary);
+        struct buf buf = BUF_INITIALIZER;
 
-        mimehdr = charset_encode_mimeheader(summary, 0, force);
+        /* Limit the subject to the first 1k chars of the SUMMARY.
+         * Otherwise, if the SUMMARY is huge, we could end up with more
+         * than 1000 header lines and we will fail to find the
+         * Content-Disposition header, and therefore have no resource URI.
+         */
+        if (len > limit) {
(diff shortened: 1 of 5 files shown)


imap/caldav_util.c
```diff
@@ -1137,8 +1137,19 @@ EXPORTED int caldav_store_resource(struct transaction_t *txn, icalcomponent *ica
...
-
-        mimehdr = charset_encode_mimeheader(summary, 0, force);
+        size_t limit = 1024, len = strlen(summary);
+
+        /* Limit the subject to the first 1k chars of the SUMMARY.
+           Otherwise, if the SUMMARY is huge, we could end up with more
+           than 1000 header lines and we will fail to find the
+           Content-Disposition header, and therefore have no resource URI.
+        */
+        if (len > limit) {
```
rsto, line 1148: I don't understand how CRLF is relevant here? The `summary` variable points to a string returned by `icalcomponent_get_summary` which is supposed to return the unfolded TEXT value. So this loop actually looks for LF (which is also what the conditional does).

I guess this heuristic is chosen because some client put a long description in the SUMMARY? Would it alternatively make sense to look for whitespace? Or just cap it at 1000 characters without trying to find a shorter string?

In any case, truncating at exactly 1000 characters could produce an invalid value if the text value contains multi-byte UTF-8.

I suggest to refactor `_email_extract_preview` from `jmap_mail.c` into a helper function `buf_truncate_utf8` in `buf.h`. This function then can make sure to safely truncate UTF-8 text.

ksmurchison: Done

rjbs:
Seems okay.  Really we the utf8 truncation should go by grapheme cluster rather than character, but that's an improvement for another day, and _much_ less straightforward.  Anybody who sends in a >1000 character summary is going to have to cope with a very light amount of text manglement.

rjbs (approved):

PR #5935, opened 2026-04-27
Always generate lookup tables from gperf source

ksmurchison:
Also use gperf declarations rather than command line options.


Makefile.am
```diff
@@ -1534,9 +1534,16 @@ imap_unexpunge_LDADD = $(LD_UTILITY_ADD)
 %_err.h %_err.c: %_err.et $(COMPILE_ET_DEP)
 	$(AM_V_GEN)(cd $(@D) && $(COMPILE_ET) $(realpath $<))
 
+lib/htmlchar.h: lib/htmlchar.gperf
+	$(AM_V_GEN)gperf $< > $@.NEW && mv $@.NEW $@
+
```
dilyanpalauzov, line 1538: I guess these gperf rules can be written in a general way for both htmlchar.h and css3_color_array.h, similar to
```
%_err.h %_err.c: %_err.et $(COMPILE_ET_DEP)
        $(AM_V_GEN)(cd $(@D) && $(COMPILE_ET) $(realpath $<))
```
So that eventually less make recipes exist.

ksmurchison: If you have a way to combine the rules for building `htmlchar.h` and `css_color_array.h` without breaking the rule that builds all of `jmap_*_props.[ch]`, I'd be happy to use it.  So far, I have tried and failed.

ksmurchison: To be clear, every time I try to change the rule for building `jmap_*_props.[ch]`, end up with undefined references for every `jmap_*_props_map` symbol

dilyanpalauzov: I mean something like this:
```
lib/htmlchar.h imap/mailbox_header_cache.h imap/css3_color_array.h: …
  $(AM_V_GEN)
```

If this works, then `jmap_*_props will not be affected by the change.

ksmurchison: I found a solution which works.  Its now the first commit

rsto (approved):

rjbs:
I can't contribute usefully to review of this, so removed myself.  Conceptually seems good to me.

Makefile.am
```diff
@@ -1407,36 +1405,36 @@ imap_libcyrus_jmap_props_la_LIBADD =
...
+    imap/jmap_contact_props.c \
+    imap/jmap_core_usercounters_props.c \
+    imap/jmap_mail_props.c \
+    imap/jmap_mail_submission_identity_props.c \
+    imap/jmap_mail_submission_props.c \
+    imap/jmap_mail_thread_props.c \
+    imap/jmap_mailbox_props.c \
+    imap/jmap_notes_props.c \
+    imap/jmap_quota_legacy_props.c \
+    imap/jmap_quota_props.c
```
elliefm: This breaks the distribution in two ways:

1) the gperf files aren't included in the tarball, so it can't be rebuilt from its own sources
2) the generated c files _are_, when they shouldn't be

`foo_SOURCES` should generally be regarded as very strictly a list of _authoritative sources_ for some target `foo`.  It's used equally for building and for choosing what gets distributed.  Generated intermediate files should not be listed in `foo_SOURCES`.  The ordinary make rules (including the implicit ones for e.g. building an .o from a .c, etc) take care of converting authoritative sources into the final products.  Once you start deviating from this pattern, other things further down the line start breaking in weird ways, and you end up with a mess of workarounds on workarounds.

I see that you're trying to fix some problem that affects the jmap_props stuff specifically when the generic make rules are in place.  I'm not sure what the right fix for that is off the top of my head, but it will probably involve revisiting the make rules, rather than tinkering with the automake definitions.

From here, you could work around the broken distribution by adding the gperf files to `EXTRA_DIST`, but that means maintaining the list of files in two places, and that's a mess.  You could expand `$(imap_jmap_props_la_SOURCES)` in `EXTRA_DIST`, but that's dodgy too, and creates evaluation order complexities that are tricky to reason about.  Neither of these workarounds correct the problem of the generated files being bundled into the tarball, either.  I think it's better to put the gperf sources back in `foo_SOURCES` where they belong, and then figure out and fix the real problem.

elliefm (changes requested):
Most of the changes here seem fine to my eye, but the changes that work around the jmap props issues are breaking the distribution.  (The other changes might have problems too, but as long as distcheck chokes on the jmap props stuff, it can't tell me about anything else wrong.)

There's workarounds you could use to get the distribution working again, but I think those workarounds would be worse than the problem we were setting out to solve here.

dilyanpalauzov:
> the generated c files are, when they shouldn't be

It is an option to include the .gperf and the .c files in the tarball, so that the person, who builds Cyrus IMAP, does not have to insatll gperf.

Makefile.am
ksmurchison: I've abandoned the commit that caused this issue and just grouped the other 4 gperf files into a static pattern rule

ksmurchison: @elliefm How does this look to you?

nwc10 (approved):

elliefm (approved):
Looks good, thanks!  Ran distcheck locally and it completes successfully.

PR #5936, opened 2026-04-28
JMAP Contacts: gracefully handle conversion properties in ContactCard/set

rsto:
This fixes a bug in `ContactCard/get` and introduces a workaround how to deal with its fallout in `ContactCard/set`.

JSContact RFC 9555 defines a couple of properties to preserve arbitrary vCard data in JSContact Card objects. We omit these properties in ContactCard/get, but due to a bug the `vCardParams` property was returned to the client. This patch makes sure that no `vCardParams` property is set for unknown TYPE parameter values.

This PR also updates `ContactCard/set` to optionally ignore vCard conversion properties, rather than rejecting them. Otherwise, Cyrus rejects updates to these Card objects if the clients preserve the conversion properties that Cyrus originally had served.

If the JMAP request capabilities include the Cyrus extension capability

    https://cyrusimap.org/ns/jmap/contacts

then conversion properties are silently discarded from the patch object before creation or update. Without the extension capability, conversion properties are rejected as invalid.


ksmurchison (approved):
LGTM

cassandane/tiny-tests/JMAPContacts/card_set_create_convprops
```diff
@@ -0,0 +1,99 @@
+#!perl
+use Cassandane::Tiny;
+
+use Text::VCardFast;
+
+sub test_card_set_create_convprops
```
wolfsage: nit: conv to me in cyrus terms means "conversations" so this was confusing, maybe use the full name (conversion_props) or since it's really about vcardprops use that instead (set_create_vcardprops)

cassandane/tiny-tests/JMAPContacts/card_set_create_vcardprops
```diff
@@ -0,0 +1,99 @@
...
+    my $carddav = $self->{carddav};
+
+    xlog $self, "Create card with vCard elements without Cyrus extension";
+    my $using = [
+        'urn:ietf:params:jmap:core',
+        'urn:ietf:params:jmap:contacts',
+        'https://cyrusimap.org/ns/jmap/debug',
+    ];
+
+    my $res = $jmap->CallMethods([ [
```
wolfsage, line 39: These two creates are identical, so you could put the data in a hash and reuse it so the test looks more like:
```
my @create = (
    'ContactCard/set',
    {
        create => { ... },
    },
    'R1',
);

my $res = $jmap->CallMethods([ \@create ]);
 	
my @wantInvalidProps = sort (
    'vCardProps', 'emails/email1/vCardParams', 'onlineServices/os1/vCardName'
);
$self->assert_not_null($res->[0][1]{notCreated}{1});
$self->assert_str_equals('invalidProperties',
    $res->[0][1]{notCreated}{1}{type});
my @gotInvalidProps = sort @{$res->[0][1]{notCreated}{1}{properties}};
$self->assert_deep_equals(\@wantInvalidProps, \@gotInvalidProps);

xlog $self, "Create card with vCard elements using Cyrus extension";
push @$using, 'https://cyrusimap.org/ns/jmap/contacts';
$res = $jmap->CallMethods([ \@create ]);

...
```
and then it's clearer that the two calls are identical and the ohnly change is the using string

cassandane/tiny-tests/JMAPContacts/card_set_create_vcardprops
```diff
@@ -0,0 +1,99 @@
...
+    $self->assert_not_null($cardId);
+    my $xhref = $res->[0][1]{created}{2}{'cyrusimap.org:href'};
+    $self->assert_not_null($xhref);
+
+    xlog $self, "Assert that vCard properties were ignored";
+    $res = $carddav->Request('GET', $xhref);
+    my $vcard = Text::VCardFast::vcard2hash($res->{content});
+    $self->assert_null($vcard->{objects}[0]{properties}{'x-foo'});
+    $self->assert_null($vcard->{objects}[0]{properties}{email}[0]{params}{'x-foo'});
+    $self->assert_null($vcard->{objects}[0]{properties}{impp});
```
wolfsage, line 78: This should assert something true about the vcard, because otherwise this test could be passing but only because say there was a typo and it turned out `$vcard->{objects}[0]{properties}` was not the correct path, or something.

maybe
```
my $props = $vcard->{objects}[0]{properties};
$self->assert_cmp_deeply(
    superhashof({
        email => [{
            'name' => 'email',
            'params' => {
              'prop-id' => [
                'email1',
              ],
            },
            'value' => 'test2@example.com',
        }],
    }),
    $props
);

$self->assert(! exists $props->{'x-foo'});
$self->assert(! exists $props->{impp});
```

cassandane/tiny-tests/JMAPContacts/card_set_update_convprops
```diff
@@ -0,0 +1,99 @@
...
+    $res = $jmap->CallMethods([ [
+        'ContactCard/set',
+        {
+            update => {
+                $cardId => {
+                    'emails/email1/vCardParams' => { "x-foo" => "Bar" },
+                    'onlineServices/os1/vCardName' => "impp",
+                    vCardProps => [ [ "x-foo", {}, "unknown", "Hello" ] ],
+                }
+
```
wolfsage: nit: extra space

cassandane/tiny-tests/JMAPContacts/card_set_update_convprops
```diff
@@ -0,0 +1,99 @@
...
```
wolfsage: nit: extra space (but also: This update could be stored like I mentioned in the create test above, and reused to reduce the amount of code / context one has to read

cassandane/tiny-tests/JMAPContacts/card_set_update_vcardprops
```diff
@@ -0,0 +1,99 @@
...
+        'R1'
+    ]], $using);
+    $self->assert(exists $res->[0][1]{updated}{$cardId});
+
+    xlog $self, "Assert that vCard properties were ignored";
+    $res = $carddav->Request('GET', $xhref);
+    my $vcard = Text::VCardFast::vcard2hash($res->{content});
+    $self->assert_null($vcard->{objects}[0]{properties}{'x-foo'});
+    $self->assert_null($vcard->{objects}[0]{properties}{email}[0]{params}{'x-foo'});
+    $self->assert_null($vcard->{objects}[0]{properties}{impp});
```
wolfsage, line 90: We should also check something positive about the card here

imap/jmap_contact.c
```diff
@@ -930,39 +937,43 @@ static void reject_convprops_internal(struct jmap_parser *parser,
...
+            if (!found_convprop) {
+                reject_or_remove_convprops_internal(parser, buf, jval,
+                                                    invalid, remove);
             }
 
             jmap_parser_pop(parser);
         }
     }
-    else if (json_is_array(jpatch)) {
+    else if (json_is_array(jpatch) && !remove) {
```
wolfsage: I think this code is entirely untested?

Why would we not remove vCardParams / vCardProperties inside of array objects?

wolfsage: (Is this only here because Contact/set had Emails => [ ... ] instead of Emails => { } ?

(
Unrelated, in testing this I found you could put whatever keys you want inside of components of a name object. Huh!

```
                        components => [
                            {
                                '@type' => 'NameComponent',
                                kind => 'given',
                                value => 'Gabriel',
                                blah => "...",
                            },

...
JSPROP;JSPTR=name/components:[{"@type":"NameComponent"\,"blah":"..."\,"kind":"given"\,"value":"Gabriel"}\,{"@type":"NameComponent"\,"kind":"given2"\,"value":"García"}\,{"@type":"NameComponent"\,"kind":"surname"\,"value":"Márquez"}]


```

imap/jmap_contact.c
```diff
@@ -6431,7 +6454,7 @@ static void _add_vcard_params(json_t *obj, vcardproperty *prop,
 
                         json_object_set_new(jprop, val, json_true());
                     }
-                    else {
+                    else if (convert_unknown) {
```
wolfsage, line 6457: Huh. What does this change do?

wolfsage (changes requested):

cassandane/tiny-tests/JMAPContacts/card_set_create_convprops
rsto: Renamed to vcardprops

cassandane/tiny-tests/JMAPContacts/card_set_create_vcardprops
rsto: Refactored to single structure to create.

cassandane/tiny-tests/JMAPContacts/card_set_create_vcardprops
rsto: Asserts now for presence of FN.

cassandane/tiny-tests/JMAPContacts/card_set_update_convprops
rsto: obsolete due to refactor

cassandane/tiny-tests/JMAPContacts/card_set_update_vcardprops
rsto: Asserts now for presence of FN.

imap/jmap_contact.c
rsto: > I think this code is entirely untested?
> 
> Why would we not remove vCardParams / vCardProperties inside of array objects?

This was indeed a bug from previous draft version of this fix, thanks.

imap/jmap_contact.c
rsto: This instructs the vCard-to-JSContact converter to set the vCardParams property if preserving vCard properties in JSContact is requested (it is not for ContactCard/get).

imap/jmap_contact.c
rsto: > (Is this only here because Contact/set had Emails => [ ... ] instead of Emails => { } ?
> 
> ( Unrelated, in testing this I found you could put whatever keys you want inside of components of a name object. Huh!
> 
> ```
>                         components => [
>                             {
>                                 '@type' => 'NameComponent',
>                                 kind => 'given',
>                                 value => 'Gabriel',
>                                 blah => "...",
>                             },
> 
> ...
> JSPROP;JSPTR=name/components:[{"@type":"NameComponent"\,"blah":"..."\,"kind":"given"\,"value":"Gabriel"}\,{"@type":"NameComponent"\,"kind":"given2"\,"value":"García"}\,{"@type":"NameComponent"\,"kind":"surname"\,"value":"Márquez"}]
> ```

Yep, that's by design. If you do set such a property and it is not registered at IANA you violate the spec, though.

wolfsage (approved):
Great

Issue #5910, opened 2026-04-15
Unable to build 3.8.6 on AlmaLinux 8


nlindq-maei:
When trying to build 3.8.6 on AlmaLinux 8.10, I receive the following error:

```
gcc -DHAVE_CONFIG_H -I.   -I. -I./lib -I. -I./lib -DLIBEXEC_DIR=\"/usr/libexec/cyrus-imapd\" -DSBIN_DIR=\"/usr/sbin\" -DSYSCONF_DIR=\"/etc\" -DHAVE_CONFIG_H     -I/usr/include -I/usr/include/libxml2               -I/usr/include/mysql -I/usr/include/mysql/mysql  -fPIC  -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection  -c -o imap/httpd.o imap/httpd.c
imap/httpd.c: In function 'tls_init':
imap/httpd.c:1263:43: error: 'OPENSSL_FULL_VERSION_STR' undeclared (first use in this function); did you mean 'OPENSSL_VERSION'?
     buf_printf(serverinfo, " OpenSSL/%s", OPENSSL_FULL_VERSION_STR);
                                           ^~~~~~~~~~~~~~~~~~~~~~~~
                                           OPENSSL_VERSION
imap/httpd.c:1263:43: note: each undeclared identifier is reported only once for each function it appears in
make[2]: *** [Makefile:5538: imap/httpd.o] Error 1
```

OpenSSL version is 1.1.1k. Versions up to 3.8.5 build successfully.

dilyanpalauzov:
You have stated at https://github.com/cyrusimap/cyrus-imapd/pull/5484#issuecomment-3189957272 that you are using OpenSSL 1.1.1.  According to https://openssl-corporation.org/post/2023-09-11-eol-111/ OpenSSL 1.1.1 has reached its End of Life.

nlindq-maei:
[...]

PR #5937, opened 2026-04-28
Cyrus 3.13.4 prep

nwc10:


nwc10:
I'm not sure if #5905 and #5906 belong as entries in 3.13.4.rst

They were the only two (seemingly) significant changes that I spotted that weren't documented in `changes/next`

docsrc/download/release-notes/3.13/x/3.13.4.rst
```diff
@@ -61,6 +61,17 @@ Major changes since the 3.12 series
 * Cyrus now requires libxapian >= 2.0.0, or a recent 1.5.x bleeding edge
   snapshot.
 * JMAP methods now use perfect hashes (gperf) for increased performance.
+* The code in ``lib/imapoptions`` has been refactored for easier maintenance and
+  extensibility.
+* Building Cyrus now requires ``perl`` >= 5.28.0 and the modules ``Moo`` and
+  ``App::Cmd``. These are usually available as OS packages - e.g. on
+  Debian-based systems as ``libmoo-perl`` and ``libapp-cmd-perl``
+* The :cyrusman:`mkimap(8)` and :cyrusman:`rehash(8)` tools are now installed
+  to ``--sbinbindir`` by ``make install``.
```
wolfsage: `--sbinbindir` should be `--sbindir`

changes/3.13.4/cyr-2789-install-mkimap
```diff
@@ -1,7 +1,7 @@
 Description:
 
 The :cyrusman:`mkimap(8)` and :cyrusman:`rehash(8)` tools are now
-installed to $prefix/sbin by `make install`.
+installed to --sbinbindir by `make install`.
```
wolfsage: --sbinbindir should be --sbindir

wolfsage (changes requested):
1 repeated minor problem but otherwise seems okay

docsrc/download/release-notes/3.13/x/3.13.4.rst
```diff
@@ -61,6 +61,17 @@ Major changes since the 3.12 series
 * Cyrus now requires libxapian >= 2.0.0, or a recent 1.5.x bleeding edge
   snapshot.
 * JMAP methods now use perfect hashes (gperf) for increased performance.
+* The code in ``lib/imapoptions`` has been refactored for easier maintenance and
+  extensibility.
+* Building Cyrus now requires ``perl`` >= 5.28.0 and the modules ``Moo`` and
```
wolfsage, line 66: I think this note about building perl requiring 5.28.0 and those modules should be in docsrc/download/upgrade.rst

wolfsage (changes requested):

docsrc/download/release-notes/3.13/x/3.13.4.rst
nwc10: Agree, but where?

Also, logically shouldn't the lines added last month also go there?

```
* Cyrus now requires libical >= 4.0.0, or a current development snapshot.
* Cyrus now requires libxapian >= 2.0.0, or a recent 1.5.x bleeding edge
  snapshot.
```

(I can't see a section to simply amend to)

The text about `libmoo-perl` and `libapp-cmd-perl` clearly sits in the "The full list (including all optional packages) for Debian is" section, so I'll add it there.

(and I suspect that we need to verify that that full list is correct at some point around the time of alpha releases)

changes/3.13.4/cyr-2789-install-mkimap
nwc10: Aaargh. I tested this in the VM, but clearly I didn't copy paste what I ran into the docks. D'oh! Well spotted. Thanks. (fixed)

docsrc/download/release-notes/3.13/x/3.13.4.rst
wolfsage: After
```
Objectstore removal
###################
```

I would wager, it seems to be a bunch of headings of "here's a big thing that changed"

I guess I missed the lib* updates :| Still getting used to the process

nwc10: I added another commit to the branch to move the 3 bullet points from `3.13.4.rst` to `upgrade.rst`

The generated HTML looks sane - is this structure what we want? (at least, "good enough for a dev release"?)

wolfsage: Hey thanks.

As to the HTML, if it looks good, wfm heh

wolfsage (approved):

docsrc/download/release-notes/3.13/x/3.13.4.rst
elliefm: Actually, for new/updated dependencies on master, we usually just note them among the "major changes since..." bullets in the release notes.  See for example the 3.10.0 release, where we bumped the required libical version and added support for pcre2.  (We didn't have any changes like this in 3.12.0, so 3.10.0 is the most recent example.)

I don't think dependency changes are worth including in the upgrade document, it's cluttered enough as it is.  If their system doesn't have the right dependencies, configure will tell them what they need, and it's reasonable to assume someone upgrading a mail server knows how to install/upgrade other packages on their system.  The upgrade document is about _Cyrus_, not so much the rest of the system around it.

For comparison, in a stable series release, if we're bumping dependencies or changing build behaviour, then we call them out specifically in like a "Dependency changes" or "Build changes" section in the release notes.  The latter is reasonably common (3.12.1 and 3.12.2 both have this section); the former is very rare (hasn't happened since 3.0.2).

dilyanpalauzov: > For comparison, in a stable series release, if we're bumping dependencies or changing build behaviour, then we call them out specifically in like a "Dependency changes" or "Build changes" section in the release notes.…  the former is very rare (hasn't happened since 3.0.2).

Building Cyrus IMAP 3.8 with OpenSSL 1.1.1 was discontinued by change in Cyrus IMAP 3.8 — https://github.com/cyrusimap/cyrus-imapd/issues/5910.

docsrc/download/release-notes/3.13/x/3.13.4.rst
```diff
@@ -0,0 +1,110 @@
...
+  identical or overlap. (:issue:`5573`:)
+* Cyrus syslogs are in the process of being converted to ``logfmt``, a
+  structured logging format.  So far, this affects primarily the login logs and
+  auditlog.
+* The ``backupcyrusd`` protocol daemon has been added to facilitate rolling
+  backups.
+* The ``expire_by_savedate`` and ``expire_keep_flagged``
+  :cyrusman:`imapd.conf(5)` options have been added to control the behavior of
+  ``cyr_expire``.
+* Time zone data is no longer included in iTIP messages.
```
dilyanpalauzov, line 41: What is this exactly?  `icalcomponent_add_required_timezones()` does remove known timezones from iTIP messages, thus keeping custom made timezones (removes timezones with  Olson names, keeps timezones with names invented by Microsoft).  So are now custom (arbitrary) made timezones also removed from the iTIP message?

docsrc/download/release-notes/3.13/x/3.13.4.rst
```diff
@@ -0,0 +1,110 @@
...
+* The new ``sieve_mdn_private`` :cyrusman:`imapd.conf(5)` option helps prevent
+  leaking usernames.
+* A new database backend called ``twom`` has been added.
+* Adds support for the Sieve 'vnd.cyrus.redirect-multiple' extension
+  which allows the 'redirect' action to accept a stringlist of email
+  addresses, in addition to the standard single string.
+  Cyrus will use a single SMTP transaction, providing each address
+  as a RCPT TO and stop script processing when any response is an error.
+* The 'sieve_extensions' option now accepts the value
+  'vnd.cyrus.redirect-multiple' to enable the extension in timsieved.
```
dilyanpalauzov, line 59: The `vnd.cyrus.redirect-multiple`  extension is mentioned in two bullets: on line 53 and on line 58.  As it is one thing, it should use one bullet.

docsrc/download/upgrade.rst
```diff
@@ -99,6 +99,17 @@ configured.  Our view is that this is so unlikely that we are not preparing
...
 :ref:`mailing list <feedback-mailing-lists>`.
 
+.. _build_dependencies:
+
+Build dependencies
+##################
+
+* Cyrus now requires ``libical`` >= 4.0.0, or a current development snapshot.
+* Cyrus now requires ``libxapian`` >= 2.0.0, or a recent 1.5.x bleeding edge
+  snapshot.
```
dilyanpalauzov: Recent should include a date, because this text can be read in 2028 and it shill has to be accurate.  Does it mean “at least the not yet released version 1.4.32?”?

https://xapian.org mentions the 1.4 and 2.0 series, there is no 1.5.

docsrc/download/upgrade.rst
```diff
@@ -163,19 +174,19 @@ Download the release :ref:`3.12 package tarball <getcyrus>`.
...
-    libio-stringy-perl libjansson-dev libldap2-dev libmysqlclient-dev \
-    libnet-server-perl libnews-nntpclient-perl libpam0g-dev libpcre2-dev \
-    libsasl2-dev libsqlite3-dev libssl-dev libtest-unit-perl libtool \
-    libunix-syslog-perl liburi-perl libxapian-dev libxml-generator-perl \
-    libxml-xpath-perl libxml2-dev libwrap0-dev libzephyr-dev lsb-base \
-    net-tools perl php-cli php-curl pkg-config po-debconf tcl-dev transfig \
-    uuid-dev vim wamerican wget xutils-dev zlib1g-dev sasl2-bin rsyslog sudo \
-    acl telnet
+    sudo apt-get install -y acl autoconf automake autotools-dev \
+    bash-completion bison build-essential comerr-dev debhelper flex g++ git \
```
dilyanpalauzov, line 167: I suggest removing `bash-completion`, `telnet`, `vim`, `wget` these are not used anyhow by Cyrus IMAP.

dilyanpalauzov (commented):
I still think that the information from the `changes/3.x` directories must be moved to some other place (user-visible documentation as html) and the `changes/3.x` directories removed.  In other words, what purpose are the changes/3.x directories supposed to serve?

docsrc/download/release-notes/3.13/x/3.13.4.rst
```diff
@@ -0,0 +1,110 @@
...
+ * caringo_hostname
+ * caringo_port
+ * object_storage_dummy_spool
+ * object_storage_enabled
+ * openio_account
+ * openio_autocreate
+ * openio_namespace
+ * openio_proxy_timeout
+ * openio_rawx_timeout
+ * openio_verbosity
```
dilyanpalauzov, line 104: `tls_server_dhparam`` should be mentioned here, too, and removed from a separate bullet on line 82.

For me there is a difference between an option, which is deprecated and an option, which has no effect, but exists solely not to create parsing errors.  Deprecated means: options should not be used any more, there are better ways to achieve the same, and the option still has some effect, when changing its value.  Option does nothing means that it the value of the option is irrvelevant.

dilyanpalauzov (commented):
In docsrc/download/upgrade.rst 
```
Objectstore removal
###################

In the very unlikely event that you're using objectstore, you'll need to stop.
```
I suggest removing `very unlikely`, it has no added value.  → `In the event that you …`

wolfsage:
> I suggest removing `very unlikely`, it has no added value. → `In the event that you …`

It has the added value of informing most folks that they probably don't need to worry about this.

dilyanpalauzov:
`very unlikely` is unscientific expression.

PR #5938, opened 2026-04-28
jmap_mail.c: always set mandatory EmailBodyPart.type property

rsto:
This updates Email/get to always set the 'type' property of an EmailBodyPart object, even if the type field is NULL in the Cyrus-internal bodypart structure.

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_get_bodypart_default_type
+++ b/cassandane/tiny-tests/JMAPEmail/email_get_bodypart_default_type
@@ -1,0 +1,96 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_get_bodypart_default_type
+    ($self)
+{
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):

wolfsage (changes requested):
This test passes for me on master suggesting it isn't triggering the bug being fixed?

rsto:
> This test passes for me on master suggesting it isn't triggering the bug being fixed?

Triggering this specific bug requires us to mess up the data on disc. I don't think that's worth the effort. I still see value in adding this test for regression, even if testing it to trigger this codepath requires manual testing. If you disagree, I'll remove the test.

rsto:
To be clear: no action planned from my side, so asking for re-review.

wolfsage:
Yeah okay, fair

wolfsage (approved):

PR #5939, opened 2026-04-29
docsrc: document JMAP for Contacts extensions

rjbs:
This needs good expert review.  It was produced by me, Claude, reading code, and looking at live Cyrus.


rjbs:
So!  This is pretty small, which makes me suspicious, but I think it's right?

I thought maybe the extension governed setting `vendor:property` style properties, but I don't think so.

It also governs extensions to Email/query filters, but I will cover those in a JMAP for Email extensions document.

ksmurchison (approved):
This looks complete except for one omission:

We also have `mailboxUniqueId` on AddressBooks

PR #5943, opened 2026-05-01
sphinxlocal cleanup

elliefm:
Sphinx 7 was complaining that cyrusman.py doesn't indicate whether or not it supports parallel read/write, and was falling back to serial mode.  I've fixed cyrusman.py to indicate that it does, assuming that that is true because the script is very simple.  While in there, I also removed a heap of unneccesary clutter that made it look far more complicated than it actually is.

Once I fixed cyrusman.py, it started complaining about the next one, and so on.  So I have updated all of them, thus:

* cyrusman.py: supports parallel r/w
* imapdconf.py: ditto
* gitstamp:py: does not support parallel r/w (it might, but I don't know that at a glance)
* sitemap.py: ditto

Sphinx still complains, but now it complains that gitstamp.py or sitemap.py (whichever it examines first) _doesn't support_ parallel r/w, instead of complaining that it _doesn't know_.  So I haven't fixed the "sphinx complains" issue but I've at least made things a little better.

### DIFF
--- a/docsrc/exts/sphinxlocal/builders/gitstamp.py
+++ b/docsrc/exts/sphinxlocal/builders/gitstamp.py
@@ -90,3 +90,8 @@ def what_build_am_i(app):
 # know what the build output format is.
 def setup(app):
     app.connect('builder-inited', what_build_am_i)
+    return {
+        # XXX Are we parallel r/w safe? Dunno! Say we're not, just in case.
+        'parallel_read_safe': False,
+        'parallel_write_safe': False,
+    }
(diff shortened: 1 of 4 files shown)


elliefm:
Anecdotally, I also tried marking gitstamp.py and sitemap.py as supporting parallel r/w, and tried building the docs.  With all extensions now claiming to support parallel builds, the docs build happens in parallel instead of series.  It doesn't fail, but I don't know how to evaluate whether the gitstamp/sitemap stuff actually still works as expected.

The parallel build cuts the `cyrus-cfg && make doc-html` time on my system from about 38 seconds to about 25.  Nothing to sneeze at, but not enough to make me want to study up the Sphinx API and confirm that gitstamp/sitemap are parallel safe.

wolfsage (approved):

rjbs (approved):

PR #5945, opened 2026-05-01
imapoptions: automate duration/bytesize unit handling

elliefm:
More long overdue improvements that have been enabled by the imapoptions refactor...

This PR resolves a bunch of double-handling that was previously necessary for duration and bytesize options (due to their human-readable units support, and historical options expecting different default units).

* The `defunit` argument to `config_getduration()` and `config_getbytesize()` is now a `Default-Unit:` field in the imapoptions data.  I expect you will almost never need to use this field, because the weird historical options that need it have it set by this PR, and we shouldn't add new weird options.  It is no longer possible to call these functions with the wrong `defunit`.
* The "if no unit is specified, x is assumed" text in the documentation is now generated from the `Default-Unit:` field, or its absence.  It is no longer possible to accidentally document this detail incorrectly.
* Duration and bytesize values are now parsed into numeric values once during initialization, where previously they were re-parsed every time they were accessed (possibly incorrectly, if the caller got `defunit` wrong!).

**Reviewers:** I suggest stepping through this one commit-by-commit.

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -294,9 +294,9 @@ EXPORTED const char *sync_get_config(const char *channel, enum imapopt opt)
     return response;
 }
 
-EXPORTED int sync_get_durationconfig(const char *channel, enum imapopt opt)
+EXPORTED int32_t sync_get_durationconfig(const char *channel, enum imapopt opt)
 {
-    int response = -1;
+    int32_t response = -1;
 
     assert(opt > IMAPOPT_ZERO && opt < IMAPOPT_LAST);
     assert(imapopts[opt].type == OPT_DURATION);
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -294,7 +294,7 @@ EXPORTED const char *sync_get_config(const char *channel, enum imapopt opt)
     return response;
 }
 
-EXPORTED int sync_get_durationconfig(const char *channel, enum imapopt opt, int defunit)
+EXPORTED int sync_get_durationconfig(const char *channel, enum imapopt opt)
 {
     int response = -1;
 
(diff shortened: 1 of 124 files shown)


imap/sync_support.c
```diff
@@ -272,75 +272,71 @@ static struct mboxlock *sync_lock(struct sync_client_state *sync_cs,
 
 /* channel-based configuration */
 
-EXPORTED const char *sync_get_config(const char *channel, const char *val)
+EXPORTED const char *sync_get_config(const char *channel, enum imapopt opt)
 {
     const char *response = NULL;
 
+    assert(opt > IMAPOPT_ZERO && opt < IMAPOPT_LAST);
```
rjbs, line 279: It's not a big deal, but even having recently reviewed a lot of the imapoptions stuff, I had to pause here to remember what ZERO and LAST were.  "Stop before the last one" is confusing unless you remember it's a sentinel.

Maybe IMAPOPT_START_SENTINEL and IMAPOPT_END_SENTINEL maybe would be clearer?  But this isn't a blocker, just how it struck me right now.

tools/lib/Cyrus/IMAPOptions/Option.pm
```diff
@@ -230,6 +268,13 @@ sub is_unreleased
     return $is_unreleased;
 }
 
+sub can_have_units
+{
+    my ($self) = @_;
```
rjbs, line 273: Just noticed this code isn't using subroutine signatures.  I think we should use that in all our Perl.  That said, it's not a blocker, imho.

rjbs (approved):
This looks really good. 👍🏽

imap/sync_support.c
elliefm: I dunno, FOO_LAST is a pretty idiomatic C idenfitier for an enum end sentinel.  Having a sentinel at zero is a little unusual, but its name says exactly what it is.

I don't disagree that the longer names would be _more_ clear, but they'd also probably mean a few lines might need to be wrapped where they weren't before.

tools/lib/Cyrus/IMAPOptions/Option.pm
elliefm: Yeah, I wrote the imapoptions core like this before we cared, and am matching existing style as I extend it.

I don't mind converting it to use them, but I'd do that as its own PR which just does that, not bundle it in piecemeal with other changes.

PR #5946, opened 2026-05-04
jmap_calendar.c: prefix Calendar[Event] state strings with 'J'

ksmurchison:
When deployed (presumably simultaneously with compact Calendar[Event] ids) this will trigger cannotCalculateChanges and force client to resync

### DIFF
--- a/imap/jmap_api.c
+++ b/imap/jmap_api.c
@@ -1318,8 +1318,19 @@ static char *_state_string(int prefixed_state, modseq_t modseq)
 EXPORTED char *jmap_state_string(jmap_req_t *req, modseq_t modseq,
                                  int mbtype, int flags)
 {
-    int prefixed_state =
-        (mbtype == MBTYPE_EMAIL) && USER_COMPACT_EMAILIDS(req->cstate);
+    int prefixed_state = 0;
+
+    if (USER_COMPACT_EMAILIDS(req->cstate)) {
+        switch (mbtype) {
+        case MBTYPE_EMAIL:
+        case MBTYPE_CALENDAR:
+            prefixed_state = 1;
+            break;
+
+        default:
+            break;
+        }
+    }
 
     // if we were not given a modseq, look it up by mbtype
     if (!modseq) modseq = jmap_modseq(req, mbtype, flags);
--- a/imap/jmap_api.c
+++ b/imap/jmap_api.c
@@ -3636,3 +3636,14 @@ EXPORTED void jmap_report_isdefault(struct jmap_set *set, const char *name,
                             json_pack("{s:b}", "isDefault", isdef));
     }
 }
+
+EXPORTED bool jmap_state_matches(struct conversations_state *cstate,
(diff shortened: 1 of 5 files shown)


imap/jmap_calendar.c
```diff
@@ -2110,15 +2111,19 @@ static int jmap_calendar_set(struct jmap_req *req)
         goto done;
     }
 
+    modseq_t old_modseq = jmap_modseq(req, MBTYPE_CALENDAR, 0);
```
rsto, line 2114: This looks like it could be declared in the scope of the "set.if_in_state" block below? This comment also applies to all other declarations of `old_modseq` in this PR.

imap/jmap_calendar.c
```diff
@@ -2110,15 +2111,19 @@ static int jmap_calendar_set(struct jmap_req *req)
...
     if (set.if_in_state) {
-        if (atomodseq_t(set.if_in_state) != jmap_modseq(req, MBTYPE_CALENDAR, 0)) {
+        const char *if_in_state = set.if_in_state;
+        if ((USER_COMPACT_EMAILIDS(req->cstate) &&  // check for mandatory prefix
+             *if_in_state++ != JMAP_STATE_STRING_PREFIX) ||
+            atomodseq_t(if_in_state) != old_modseq) {
```
rsto: This conditional-with-side-effect gets repeated several times in this PR and I think it should get refactored to a helper function `match_state` or the like. This not only is because of code hygiene but also because it looks to me as if the state where Cyrus code supports both old-style and new-style state strings is temporary, so it will become simpler to remove support for the old state strings later. Doing so, you might also add an according comment to that helper function with `XXX` or the like, so we can keep track of it.

rsto (changes requested):
The logic looks sound (albeit being untested). I think this code should get refactored.

rjbs:
Removing myself to leave this owe for @wolfsage who will be the one directly benefitting from this.

imap/jmap_calendar.c
ksmurchison: Yeah, good call.  Refactored out and also used in jmap_mail.c and jmap_mailbox.c

imap/jmap_api.c
```diff
@@ -3636,3 +3647,14 @@ EXPORTED void jmap_report_isdefault(struct jmap_set *set, const char *name,
                             json_pack("{s:b}", "isDefault", isdef));
     }
 }
+
+EXPORTED bool jmap_state_is_valid(struct conversations_state *cstate,
```
rsto: one more nit: this function does more than checking if the state string is valid (e.g. it's correctly formatted). It checks if the state **matches**. `jmap_state_matches` would read clearer to me.

rsto (approved):
LGTM, but I think the function name could be named clearer.

imap/jmap_api.c
ksmurchison: Done

rsto:

PR #5947, opened 2026-05-04
fix(jmap): derive FN in display order for unordered name components

tristan957:
In Fastmail, I use the workflow of adding contacts from email conversations, but I have found that contacts that I have added in the flow have vCard data with invalid name ordering. I use DAVx5 to sync my Fastmail contacts into my Android device, and this invalid name ordering makes its way into Google Messages, but not Google Contacts for some reason.

---

When creating a vCard with unordered name components (no JSCOMPS parameter), the derived FN was built by iterating components in N field order (family, given, additional, prefix, suffix), producing incorrect results like "Drouvot Bertrand" instead of "Bertrand Drouvot".

Fix by building the derived FN in conventional display order (prefix, given, given2, family, suffix) when components are not explicitly ordered. The ordered (isOrdered/JSCOMPS) path remains unchanged.

### DIFF
--- a/cassandane/tiny-tests/JMAPContacts/card_set_create_name
+++ b/cassandane/tiny-tests/JMAPContacts/card_set_create_name
@@ -72,4 +72,53 @@ sub test_card_set_create_name
     $self->assert_matches(qr/N;JSCOMPS=";1;2;0;5";SORT-AS=Pau Shou Chang,Robert:Shou,Chang;Robert;Pau;;;Chang;/, $card);
     $self->assert_matches(qr/FN;DERIVED=TRUE:Robert Pau Shou Chang/, $card);
     $self->assert_does_not_match(qr|JSPROP|, $card);
+
+    # Test unordered name components: derived FN should use display order
+    # (prefix given given2 family suffix), not N field order (family first)
+    my $id2 = 'ae2640cc-234a-4dd9-95cc-3106258445ba';
+
+    $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            create => {
+                "2" => {
+                    '@type' => 'Card',
+                    version => '1.0',
+                    uid => $id2,
+                    name => {
+                        '@type' => 'Name',
+                        components => [
+                            {
+                                kind => 'title',
+                                value => 'Dr.'
+                            },
+                            {
+                                kind => 'given',
+                                value => 'Bertrand'
+                            },
+                            {
+                                kind => 'credential',
+                                value => 'Jr.'
+                            },
+                            {
+                                kind => 'surname',
+                                value => 'Drouvot'
+                            }
+                        ]
+                    }
+                }
+            }
+        }, 'R2']
+    ]);
+
+    $self->assert_not_null($res->[0][1]{created}{2});
+
+    $href = $res->[0][1]{created}{2}{'cyrusimap.org:href'};
+    $res = $carddav->Request('GET', $href, '',
+                             'Accept' => 'text/vcard; version=4.0');
+
+    $card = $res->{content};
+    $card =~ s/\r?\n[ \t]+//gs;  # unfold long properties
+
+    $self->assert_matches(qr/N:Drouvot;Bertrand;;Dr\.;Jr\./, $card);
+    $self->assert_matches(qr/FN;DERIVED=TRUE:Dr\. Bertrand Drouvot Jr\./, $card);
 }
--- a/imap/jmap_contact.c
+++ b/imap/jmap_contact.c
@@ -3601,7 +3601,7 @@ static int _blob_to_card(struct jmap_req *req,
         buf_free(&buf);
         return HTTP_BAD_REQUEST;
     }
-    
+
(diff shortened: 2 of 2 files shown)


tristan957:
This is my first time contributing to the project, so this was the best that I could come up with that seemed reasonable. I developed this PR on a machine where it isn't feasible to run tests. Hoping I can make sure everything works later this week on my regular machine.

imap/jmap_contact.c
```diff
@@ -9904,6 +9904,27 @@ static vcardproperty *_jscomps_to_vcard(struct jmap_parser *parser, json_t *obj,
...
+        static const unsigned fn_order[] = {
+            VCARD_N_PREFIX, VCARD_N_GIVEN, VCARD_N_ADDITIONAL,
+            VCARD_N_FAMILY, VCARD_N_SUFFIX
+        };
+
+        for (size_t f = 0; f < sizeof(fn_order)/sizeof(fn_order[0]); f++) {
+            vcardstrarray *sa = vcardstructured_field_at(vals, fn_order[f]);
+            for (size_t v = 0; sa && v < vcardstrarray_size(sa); v++) {
+                const char *val = vcardstrarray_element_at(sa, v);
+                if (*val) {
```
rsto: let's please be super-cautious here and check `val` for being NULL, e.g. `if (val && *val)`

imap/jmap_contact.c
```diff
@@ -9904,6 +9904,27 @@ static vcardproperty *_jscomps_to_vcard(struct jmap_parser *parser, json_t *obj,
...
+                    if (buf_len(args->derived))
+                        buf_appendcstr(args->derived, " ");
```
rsto: nit: this could also be `buf_putc(args->derived, ' ')` since you only add a single character.

cassandane/tiny-tests/JMAPContacts/card_set_create_name
```diff
@@ -72,4 +72,53 @@ sub test_card_set_create_name
...
+    $res = $jmap->CallMethods([
+        ['ContactCard/set', {
+            create => {
+                "2" => {
+                    '@type' => 'Card',
+                    version => '1.0',
+                    uid => $id2,
+                    name => {
+                        '@type' => 'Name',
+                        components => [
```
rsto, line 89: This test passes even with the current implementation. Please change the order of components such that this test only passes with the actual changes in `jmap_contact.c`. E.g. you could move the surname to the last position, behind the credential.

rsto (changes requested):
Thank for this! I much appreciate both the test and the code contribution. I only have a few remarks. In addition to my specific remarks in the code, could you please slightly adapt the title of the commit? Please replace the `fix(jmap):` with `jmap_contact.c`, which is how we typically format commit titles.

rsto:
Adding @ksmurchison as second reviewer.

rsto:
One more remark: the NameComponent types in JSContact are defined such that they conform to the Unicode Person Name fields defined in https://www.unicode.org/reports/tr35/tr35-personNames.html#fields. This would allow to derive full names for diverse cultures. But this change is reasonable for a lot of names, so let's start with that.

imap/jmap_contact.c
ksmurchison: Alternatively, you could declare

`const char *sep = "";`

outside of the outer-most loop, and then add the string to the buffer in one go:

```
buf_printf(&args->derived, "%s%s", sep, val);
sep = " ";
```

ksmurchison (changes requested):
I agree with the changes suggested by @rsto, but otherwise looks good.

I also want to thank you for both the code contribution and test.  

Once this PR is updated with the changes, we will merge this straight away and you should see this on our production servers in a couple of weeks.

imap/jmap_contact.c
rsto: Caveat that `buf_printf` is the most costly operation, where `buf_appendcstr` and `buf_putc` are definitely cheaper. Won't break the bank here, but I suggest to not use `printf` here.

tristan957:
Really appreciate the review from both of you. I'll look into getting updated code to you all some time this week!

tristan957:
In my latest push, it looks like I accidentally committed some whitespace removals. Let me know if that will be a problem.

cassandane/tiny-tests/JMAPContacts/card_set_create_name
tristan957: I changed this in my latest push.

rsto (approved):
Thanks! Looks good to me.

rsto:
@ksmurchison your review is marked as "requesting changes" but I guess you meant the requested changes in my review? Waiting for your approval just to be sure, then we'll merge this.

ksmurchison (approved):

rsto:
thanks @tristan957!

tristan957:
Happy to contribute to Fastmail :D

rsto:
@tristan957 This now got deployed at Fastmail. Thanks again for your contribution!

tristan957:
Thank @rsto.

PR #5950, opened 2026-05-06
Further cleanup of the hash code prior to dynamic resizing

nwc10:
A common idiom in the Cyrus code is to look directly at the `.size` member of the hash table struct. This is problematic because

1. it breaks encapsulation
2. it's not always clear **why** the code is doing this
3. that member will be removed soon

hence this MR addresses that, and some other problems found

* fix errors in documentation
* rename `hash_numrecords` to `hash_count` and make it `static inline`
* jump through hoops to keep `clang` and `clang++` happy with the above
* improve test coverage of C code that is refactored
* remove code which inspected `.size` before calling `free_hash_table`
* change code implementing lazy hash initialisation to check `.table` instead of `.size`
* change code to use `hash_count` instead of reading `.size` for avoiding expensive key computation logic where there were no keys
* remove code which inspected `.size` simply to avoid calling into `hash_lookup`
* use booleans or NULL pointers to signal "no hash lookup needed" instead of looking at `.size`
* document the initialisation of `contactgroups->table`
* other minor cleanups

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_query_fromcontactcarduid
+++ b/cassandane/tiny-tests/JMAPEmail/email_query_fromcontactcarduid
@@ -118,9 +118,7 @@ sub test_email_query_fromcontactcarduid
             ],
         }, 'R1']
     ], $using);
-    $self->assert_num_equals(2, scalar @{$res->[0][1]{ids}});
-    $self->assert_str_equals($emailId1, $res->[0][1]{ids}[0]);
-    $self->assert_str_equals($emailId2, $res->[0][1]{ids}[1]);
+    $self->assert_deep_equals([$emailId1, $emailId2], $res->[0][1]{ids});
 
     # Filter by fromAnyContact
     $res = $jmap->CallMethods([
@@ -133,9 +131,7 @@ sub test_email_query_fromcontactcarduid
             ],
         }, 'R1']
     ], $using);
-    $self->assert_num_equals(2, scalar @{$res->[0][1]{ids}});
-    $self->assert_str_equals($emailId1, $res->[0][1]{ids}[0]);
-    $self->assert_str_equals($emailId2, $res->[0][1]{ids}[1]);
+    $self->assert_deep_equals([$emailId1, $emailId2], $res->[0][1]{ids});
 
     # Negate filter by contact group.
     $res = $jmap->CallMethods([
@@ -151,8 +147,7 @@ sub test_email_query_fromcontactcarduid
             ],
         }, 'R1']
     ], $using);
-    $self->assert_num_equals(1, scalar @{$res->[0][1]{ids}});
-    $self->assert_str_equals($emailId3, $res->[0][1]{ids}[0]);
+    $self->assert_deep_equals([$emailId3], $res->[0][1]{ids});
 
     # Unknown contact group should return an empty set.
     $res = $jmap->CallMethods([
@@ -234,12 +229,9 @@ sub test_email_query_fromcontactcarduid
             },
         }, 'R3']
     ], $using);
-    $self->assert_num_equals(1, scalar @{$res->[0][1]{ids}});
-    $self->assert_str_equals($emailId4, $res->[0][1]{ids}[0]);
-    $self->assert_num_equals(1, scalar @{$res->[1][1]{ids}});
-    $self->assert_str_equals($emailId5, $res->[1][1]{ids}[0]);
-    $self->assert_num_equals(1, scalar @{$res->[2][1]{ids}});
-    $self->assert_str_equals($emailId6, $res->[2][1]{ids}[0]);
+    $self->assert_deep_equals([$emailId4], $res->[0][1]{ids});
+    $self->assert_deep_equals([$emailId5], $res->[1][1]{ids});
+    $self->assert_deep_equals([$emailId6], $res->[2][1]{ids});
 
     # Support individual cards
     $res = $jmap->CallMethods([
@@ -287,7 +279,5 @@ sub test_email_query_fromcontactcarduid
             sort => [{ property => "subject" }],
         }, 'R1'],
     ], $using);
-    $self->assert_num_equals(2, scalar @{$res->[0][1]{ids}});
-    $self->assert_str_equals($emailId7, $res->[0][1]{ids}[0]);
-    $self->assert_str_equals($emailId8, $res->[0][1]{ids}[1]);
+    $self->assert_deep_equals([$emailId7, $emailId8], $res->[0][1]{ids});
 }
--- a/cassandane/tiny-tests/JMAPNote/note_get
+++ b/cassandane/tiny-tests/JMAPNote/note_get
@@ -60,31 +60,24 @@ EOF
     $imap->append('INBOX.Notes', $note3) || die $@;
 
     my $res = $jmap->CallMethods([['Note/get', { }, "R1"]]);
-    $self->assert_cmp_deeply(
+    $self->assert_cmp_deeply([
         superhashof({
             id     => $id1,
             title  => 'snowman',
             body   => "☃\n\n",
             isHTML => JSON::false
         }),
-        $res->[0][1]{list}[0],
-    );
-    $self->assert_cmp_deeply(
         superhashof({
             id     => $id2,
             title  => 'bad charset',
             body   => "foo bar\n",
             isHTML => JSON::false
         }),
-        $res->[0][1]{list}[1],
-    );
-    $self->assert_cmp_deeply(
         superhashof({
             id     => $id3,
             title  => 'bad encoding',
             body   => "foo bar\n",
             isHTML => JSON::false
         }),
-        $res->[0][1]{list}[2],
-    );
+    ], $res->[0][1]{list});
 }
--- a/cassandane/tiny-tests/JMAPNote/note_get
+++ b/cassandane/tiny-tests/JMAPNote/note_get
@@ -80,4 +80,30 @@ EOF
             isHTML => JSON::false
         }),
     ], $res->[0][1]{list});
+
+    my $id4 = $id2 =~ tr/0-9/5-90-4/r;
+    my $id5 = $id3 =~ tr/0-9/5-90-4/r;
+    $res = $jmap->CallMethods([['Note/get',
+                                { ids => [$id4, $id1, $id5] },
+                                "R2"]]);
+    $self->assert_cmp_deeply(
+        bag($id4, $id5),
+        $res->[0][1]{notFound},
+    );
+    $self->assert_cmp_deeply([
+        superhashof({
+            id     => $id1,
+            title  => 'snowman',
+            body   => "☃\n\n",
+            isHTML => JSON::false
+        })],
+        $res->[0][1]{list},
+    );
+
+    $res = $jmap->CallMethods([['Note/get',
+                                { ids => [] },
+                                "R2"]]);
+    $self->assert_cmp_deeply([],
+        $res->[0][1]{list},
+    );
 }
--- a/cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db
+++ b/cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db
@@ -1,0 +1,125 @@
+#!perl
+use Cassandane::Tiny;
+use Path::Tiny;
+
+# This test is based on email_query_fromanycontact_ignore_localpartonly but
+# modified to trigger an error path.
+#
+# Creating the contact creates a dav.db on disk
+# The method install_sieve_script directly calls sievec to compile the sieve
+# script, and stores both the source script and the bytecode version on disk
+#
+# The test then finds and scribbles on dav.db
+#
+# After this, a call to deliver
+# * fails to open dav.db
+# * assumes that what's on disk is good enough
+# * runs with what it has
+# * that bytecode-on-disk needs to load contacts
+# * that fails (fails to open dav.db)
+# * sieve filtering aborted; fallback to default delivery (to INBOX)
+#
+# We might consider the above detailed behaviour a bug - if the dav.db can't be
+# opened to check that the script is current, instead of running with a possibly
+# stale script, should we instead immediately do the "fallback to default
+# delivery"?
+#
+# Or instead, should we aggressively uncache the script from disk when the
+# dav.db is updated - ie move the staleness check into the write path?
+#
+# Either way, the end behaviour of *this* test should be the same - if the
+# dav.db is corrupt then sieve should not run, and delivery should be to INBOX
+# The precise why of how it happens might be wrong. (And we need other tests to
+# validate behaviour for an on-disk cached sieve script that didn't need dav.db)
+
+sub test_deliver_corrupt_dav_db
+    :min_version_3_3 :JMAPExtensions
+    :needs_component_sieve
+    ($self)
+{
+    my $contactId1 = $self->default_user->contacts->create({
+        emails => {
+            personal => {
+                contexts => { private => JSON::true },
+                address => 'email'
+            }
+        },
+    });
+    $self->assert_not_null($contactId1);
+
+    xlog "Assert JMAP sieve ignores localpart-only contacts";
+    $self->{store}->get_client()->create("INBOX.matches") or die;
+    $self->{instance}->install_sieve_script(<<'EOF'
+require ["x-cyrus-jmapquery", "x-cyrus-log", "variables", "fileinto"];
+if
+  allof( not string :is "${stop}" "Y",
+    jmapquery text:
+  {
+    "operator" : "NOT",
+    "conditions" : [
+        {
+           "fromAnyContact" : true
+        }
+    ]
+  }
+.
+  )
+{
+  fileinto "INBOX.matches";
+}
+EOF
+    );
+
+    # This is what is supposed to happen, given the sieve filter above.
+    # We keep this code in to confirm that the correct behaviour of the sieve
+    # filter. This call to ->deliver() isn't needed to create a dav.db on disk.
+    my $msg1 = $self->{gen}->generate(from => Cassandane::Address->new(
+            localpart => 'email', domain => 'local'
+    ), subject => 'Message 1');
+
+    $self->{instance}->deliver($msg1);
+    $self->{store}->set_fetch_attributes('uid');
+    $self->{store}->set_folder('INBOX.matches');
+    $self->check_messages({ 1 => $msg1 }, check_guid => 0);
+
+    my @found;
+    path($self->{instance}->{basedir})->visit(
+        sub ($path, $) {
+          return if $path->is_dir();
+          # We have to "corrupt" the file on disk. If we simply remove it then a
+          # fresh (valid) SQLlite database is created on the next call to
+          # sqlite3_open_v2()
+          # Instead, with a corrupt file sqlite3_open_v2() succeeds but the call
+          # sqlite3_exec(open->db, "PRAGMA user_version;", ...) fais and logs.
+          if($path->basename() eq 'dav.db') {
+            push @found, $path;
+            $path->spew("this space for rent\n");
+          }
+        },
+        { recurse => 1 }
+    );
+    $self->assert_num_equals(1, scalar @found, 'exactly one dav.db found');
+
+    # We repeat the above delivery, and test that the outcome differs:
+    my $msg2 = $self->{gen}->generate(from => Cassandane::Address->new(
+            localpart => 'email', domain => 'local'
+    ), subject => 'Message 2');
+    $msg2->set_attribute(uid => 1);
+
+    $self->{instance}->getsyslog();
+    $self->{instance}->deliver($msg2);
+    my @lines = $self->{instance}->getsyslog();
+    # Exactly what we log might change if we change the implementation. But I
+    # think it better to write a potentially fragile test, because with it we
+    # see whether the expected code paths are still reached.
+    # Cassandane::Instance::_check_syslog() faults these if it sees them:
+    my @dberror = grep /: DBERROR: get user version failed: sessionid=<[^>]+> fname=<\Q$found[0]\E> error=<file is not a database>/, @lines;
+    $self->assert_num_equals(2, scalar @dberror, '2 expected fatal log lines found');
+    # The actual error we're trying to reach doesn't default to a test failure:
+    my @carddav_error = grep /: jmap: carddav_open_userid\(cassandane\) failed$/, @lines;
+    $self->assert_num_equals(1, scalar @carddav_error, 'carddav_open_userid logged failure');
+
+    $self->check_messages({ 1 => $msg1 }, check_guid => 0);
+    $self->{store}->set_folder('INBOX');
+    $self->check_messages({ 1 => $msg2 }, check_guid => 0);
+}
--- a/lib/hash.c
+++ b/lib/hash.c
@@ -1,4 +1,3 @@
-/* +++Date last modified: 05-Jul-1997 */
 #ifdef HAVE_CONFIG_H
 #include <config.h>
 #endif
@@ -43,8 +42,8 @@ struct bucket {
 
 /* Initialize the hash_table to the size asked for.  Allocates space
 ** for the correct number of pointers and sets them to NULL.  If it
-** can't allocate sufficient memory, signals error by setting the size
-** of the table to 0.
+** can't allocate sufficient memory it will terminate the program with the
+** diagnostic "Virtual memory exhausted"
 */
 
 EXPORTED hash_table *construct_hash_table(hash_table *table, size_t size, int use_mpool)
--- a/lib/hash.c
+++ b/lib/hash.c
@@ -13,6 +13,13 @@
 #include "util.h"
 #include "xmalloc.h"
 
+/* Ideally having this declaration with EXPORTED would be sufficient, and we
+ * could remove it from the definition in the header. gcc is fine with this,
+ * but clang treats it as an error.
+ * See the commit message for attempted approaches that failed. */
+
+EXPORTED extern inline size_t hash_count(const hash_table *table);
+
 struct bucket {
     void *data;
     struct bucket *next;
@@ -164,7 +171,7 @@ EXPORTED void *hash_lookup(const char *key, hash_table *table)
 {
       bucket *ptr;
 
-      if (!table->size)
+      if (!table->size || !table->count)
           return NULL;
 
       uint32_t hash = strhash_seeded(table->seed, key);
@@ -332,12 +339,6 @@ EXPORTED strarray_t *hash_keys(const hash_table *table)
     }
 
     return sa;
-}
-
-EXPORTED int hash_numrecords(hash_table *table)
-{
-    /* XXX macro or inline this if we keep the count field long term */
-    return table->count;
 }
 
 EXPORTED void hash_enumerate_sorted(hash_table *table, void (*func)(const char *, void *, void *),
--- a/lib/hash.c
+++ b/lib/hash.c
@@ -5,6 +5,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <syslog.h>
+#include <stdbool.h>
 
 #include "assert.h"
 #include "hash.h"
@@ -19,6 +20,7 @@
  * See the commit message for attempted approaches that failed. */
 
 EXPORTED extern inline size_t hash_count(const hash_table *table);
+EXPORTED extern inline bool hash_constructed(const hash_table *table);
 
 struct bucket {
     void *data;
--- a/lib/hash.c
+++ b/lib/hash.c
@@ -13,6 +13,13 @@
 #include "util.h"
 #include "xmalloc.h"
 
+/* Ideally having this declaration with EXPORTED would be sufficient, and we
+ * could remove it from the definition in the header. gcc is fine with this,
+ * but clang treats it as an error.
+ * See the commit message for attempted approaches that failed. */
+
+EXPORTED extern inline size_t hash_count(const hash_table *table);
+
 struct bucket {
     void *data;
     struct bucket *next;
@@ -177,7 +184,7 @@ EXPORTED void *hash_lookup(const char *key, hash_table *table)
 {
       bucket *ptr;
 
-      if (!table->size)
+      if (!table->size || !table->count)
           return NULL;
 
       uint32_t hash = strhash_seeded(table->seed, key);
@@ -345,12 +352,6 @@ EXPORTED strarray_t *hash_keys(const hash_table *table)
     }
 
     return sa;
-}
-
-EXPORTED int hash_numrecords(hash_table *table)
-{
-    /* XXX macro or inline this if we keep the count field long term */
-    return table->count;
 }
 
 EXPORTED void hash_enumerate_sorted(hash_table *table, void (*func)(const char *, void *, void *),
--- a/lib/hash.h
+++ b/lib/hash.h
@@ -1,4 +1,3 @@
-/* +++Date last modified: 05-Jul-1997 */
 #ifndef HASH__H
 #define HASH__H
 
@@ -20,7 +19,7 @@ typedef struct bucket bucket;
 
 /*
 ** This is what you actually declare an instance of to create a table.
-** You then call 'construct_table' with the address of this structure,
+** You then call 'construct_hash_table' with the address of this structure,
 ** and a guess at the size of the table.  Note that more nodes than this
 ** can be inserted in the table, but performance degrades as this
 ** happens.  Performance should still be quite adequate until 2 or 3
@@ -37,8 +36,8 @@ typedef struct hash_table {
 } hash_table;
 
 /*
-** This is used to construct the table.  If it doesn't succeed, it sets
-** the table's size to 0, and the pointer to the table to NULL.
+** This is used to construct the table.  If it can't allocate sufficient memory
+** it will terminate the program with the diagnostic "Virtual memory exhausted".
 */
 
 hash_table *construct_hash_table(hash_table *table, size_t size,
@@ -94,9 +93,9 @@ int hash_numrecords(hash_table *table);
 ** it calls the function whose address it was passed, with a pointer
 ** to the data that was in the table.  The function is expected to
 ** free the data.  Typical usage would be:
-** free_table(&table, free);
+** free_hash_table(&table, free);
 ** if the data placed in the table was dynamically allocated, or:
-** free_table(&table, NULL);
+** free_hash_table(&table, NULL);
 ** if not.  ( If the parameter passed is NULL, it knows not to call
 ** any function with the data. )
 */
--- a/lib/hash.h
+++ b/lib/hash.h
@@ -3,6 +3,7 @@
 
 #include <stddef.h>           /* For size_t     */
 #include <stdint.h>
+#include <stdbool.h>
 #include "mpool.h"
 #include "strarray.h"
 
@@ -103,6 +104,29 @@ strarray_t *hash_keys(const hash_table *table);
 EXPORTED inline size_t hash_count(const hash_table *table)
 {
     return table->count;
+}
+
+/* true if construct_hash_table() has been called on this hash table
+ * false if the memory has only been initialized with HASH_TABLE_INITIALIZER
+ * undefined if neither HASH_TABLE_INITIALIZER was used nor
+ * construct_hash_table() called.
+ *
+ * As free_hash_table() resets the memory to that same state, will be true after
+ * it has been called
+ * Encapsulates an idiom used by lots of Cyrus code to defer expensive
+ * initialization of hash tables until the first time they are actually needed
+ */
+
+EXPORTED inline bool hash_constructed(const hash_table *table)
+{
+    /* Currently this tests whether .table is non-NULL, but in future we might
+     * change the implementation further such that table isn't even allocated
+     * until the first hash_insert(), so we hide this lookup in a trivial named
+     * function, instead of exposing implementation details that might change
+     * everywhere in the main codebase.
+     */
+
+    return table->table;
 }
 
 /*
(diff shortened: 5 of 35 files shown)


nwc10:
This branch builds on #5902 (which should be merged to master first)

The first 6 commits are from that PR, which github doesn't make clear. (it's not showing any sort of "tag" at *hash: store hash keys using a flexible array in the bucket struct*)

rsto (approved):
Looks good to me!

ksmurchison (approved):

nwc10:
Difference from the previously approved branch

1. rebased onto #5991 (which is current master + 1 commit)
2. removes the `#pragma` dance from the headers (as the above change makes `static inline`) "work" on clang)
4. adds `tiny-tests/JMAPSieve/deliver_corrupt_dav_db`
5. re-orders and merges two commits
6. changes some comments in them:

```
diff --git a/imap/jmap_mail.c b/imap/jmap_mail.c
index 233726114..8c1cd399f 100644
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -1930,12 +1930,26 @@ static void _email_search_contactgroup(search_expr_t *parent,
      * in that case. That choice seemed conceptually buggy, because it's not
      * consistent with the "hard false" SEOP_FALSE return just below - if the
      * user is asking to filter on a groupid, and there are *no* group IDs, then
-     * the filter should return zero results, not return all results. And if
-     * initialisation failed in some way, the search should not return a result
-     * (at all) - instead it should return an error condition.
-     * As is, I believe that this is already handled that way - the only path
-     * that might cause contactgroups not be initialised is unreachable, because
-     * Cyrus will have already returned a 503 status response.
+     * the filter should return zero results, not return all results.
+     *
+     * It's actually already handled this way (failure) -
+     * _email_parse_filter_cb() correctly checks and propagates the error
+     * returned from jmap_email_contactfilter_from_filtercondition() and the
+     * search fails. In that function, for carddav_open_userid() to return NULL
+     * dav_open_userid() must fail - the user's SQLite DB on disk must be
+     * present AND corrupt. However, before we even reach that code path Cyrus
+     * has called my_dav_auth() in http_dav_sharing.c
+     * That calls webdav_open_userid(userid), which also calls
+     * dav_open_userid(userid). The failure path there is:
+     * if (!auth_webdavdb) {
+     *     syslog(LOG_ERR, "Unable to open WebDAV DB for userid: %s", userid);
+     *     return HTTP_UNAVAILABLE;
+     * }
+     * Hence that failure isn't even possible via the functions this source file
+     * - the only path that might cause contactgroups not be initialised is
+     * unreachable, because Cyrus will have already returned a 503 status
+     * response if the dav.db exists but can't be read.
+     * (Hence I can't write a test for it)
      */

     strarray_t *members = hash_lookup(groupid, contactgroups);
diff --git a/imap/jmap_mail_query.c b/imap/jmap_mail_query.c
index 47aa039ea..c0cb2ad46 100644
--- a/imap/jmap_mail_query.c
+++ b/imap/jmap_mail_query.c
@@ -253,6 +253,10 @@ HIDDEN int jmap_email_contactfilter_from_filtercondition(js
on_t *filter,
         havefield = 1;
         break;
     }
+
+    /* (Observe that the loop above has the same json_*() API calls as the
+     * loop below - the test above is "will the loop below find anything")
+     */
     if (!havefield) goto done;

     /* ensure we have preconditions for lookups */
@@ -264,24 +268,7 @@ HIDDEN int jmap_email_contactfilter_from_filtercondition(js
on_t *filter,
     /* Open CardDAV db for accountid */
     struct carddav_db *carddavdb = carddav_open_userid(cfilter->accountid);
     if (!carddavdb) {
-        /* I believe this code to be unreachable. For carddav_open_userid() to
-         * return NULL dav_open_userid() must fail - the user's SQLite DB on
-         * disk must be missing or corrupt. However, before we even reach this
-         * code path Cyrus has called my_dav_auth() in http_dav_sharing.c
-         * That calls webdav_open_userid(userid), which also calls
-         * dav_open_userid(userid). The failure path there is:
-         * if (!auth_webdavdb) {
-         *     syslog(LOG_ERR, "Unable to open WebDAV DB for userid: %s", useri
d);
-         *     return HTTP_UNAVAILABLE;
-         * }
-         *
-         * ie the request has already been abandoned with a 503 error.
-         * Hence we can't reach this error path, hence we can't have
-         * cfilter->contactgroups in the state were construct_hash_table() was
-         * called on it, but no entries were added.
-         * (Observe that the loop above has the same json_*() API calls as the
-         * loop below - the test above is "will the loop below find anything")
-         */
+        /* This code is reachable by lmtpd, tested by deliver_corrupt_dav_db */

         syslog(LOG_ERR, "jmap: carddav_open_userid(%s) failed",
                cfilter->accountid);
@@ -702,6 +689,12 @@ static xapian_query_t *_matchmime_query_new_contactgroup(co
nst char *groupid,
 {
     xapian_query_t *xq = NULL;

+    /* If this code is reached, then cfilter->contactgroups has always been
+     * initialised. If jmap_email_contactfilter_from_filtercondition() above
+     * fails to open dav.db then its error return is correctly handled by
+     * jmap_email_matchmime(), meaning that the latter does not call
+     * _matchmime_buildquery() and in turn that does not call this function.
+     */
     strarray_t *members = hash_lookup(groupid, &cfilter->contactgroups);
     if (members && strarray_size(members)) {
         ptrarray_t xsubqs = PTRARRAY_INITIALIZER;
```

as thanks to a suggestion from @alh I was able to write a regression test to exercise one of the code paths, at which point it became clear that I still hadn't grasped the maze of code paths and hidden implications.

**I** think that we're good to merge now - I don't think that this branch will need to get reworked further.

wolfsage:
The commit message for "hash: hash_numrecords() is now static inline hash_count() returning size_t"  can now be rewritten a bit. It still seems to mention this:
```
Frustratingly, clang warns on unused static inline with -Wunused-function,
and we have -Werror too, so (by default) this breaks the build on clang.
The clang developers feel that this behaviour is sane, but it breaks the
use of static inline functions as clean macro replacements.

We can disable this with a #pragma. But "#pragma clang ..." doesn't work
because clang++ doesn't recognise "clang" and throws its own error
(thanks to "-Werror=unknown-pragmas"). Fortunately both can agree that they
accept "#pragma GCC ..." as a synonym for self.
```

but those pragma's have been removed since the inclusion of https://github.com/cyrusimap/cyrus-imapd/pull/5950/commits/e93174a26bed9e5af6c1f2ac21a3fa7a284bbcc8

cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db
```diff
@@ -0,0 +1,125 @@
...
+            localpart => 'email', domain => 'local'
+    ));
+
+    $self->{instance}->getsyslog();
+    $self->{instance}->deliver($msg1);
+    my @lines = $self->{instance}->getsyslog();
+    # Exactly what we log might change if we change the implementation. But I
+    # think it better to write a potentially fragile test, because with it we
+    # see whether the expected code paths are still reached:
+    my @expected = grep /: DBERROR: get user version failed: sessionid=<[^>]+> fname=<\Q$found[0]\E> error=<file is not a database>/, @lines;
```
wolfsage: Is this test actually asserting what we want? (This is a real question, not rhetorical).

The code we expect to hit is:
```
    /* Open CardDAV db for accountid */
    struct carddav_db *carddavdb = carddav_open_userid(cfilter->accountid);
    if (!carddavdb) {
        /* This code is reachable by lmtpd, tested by deliver_corrupt_dav_db */

        syslog(LOG_ERR, "jmap: carddav_open_userid(%s) failed",
               cfilter->accountid);
        r = CYRUSDB_INTERNAL;
        goto done;
    }
```

...from lmtpd.

Would it be good to check that we got that particular message?

In all of the errors we check for, would it be good to make sure they come from the 'lmtp' process?

Might it also be good to try a delivery first to make sure the message ends up in the .matches subfolder to make sure our sieve is right and that it didn't wind up in the inbox because of a typo?

wolfsage (commented):
It feels weird to me that we replaced hash->size with hash->table. Both feel like they're leaking internals. It's fine, but I wonder if something like hash_initialized() (like hash_count()) would be better? /shrug

In any case, I think a commit message should be adjusted as noted just a little while ago, and maybe the corruption test could be improved a little bit, but it's also fine if you don't change that test.

lib/hash.h
```diff
@@ -86,7 +86,10 @@ strarray_t *hash_keys(const hash_table *table);
 
 /* counts the number of nodes in the hash table */
 
-int hash_numrecords(hash_table *table);
+static inline size_t hash_count(hash_table *table)
+{
+    return table->count;
+}
```
elliefm: This is the wrong way to set up a general inline function in C99+.  You use `static inline` for _inlining a static function_ (i.e. one that is visible only within a single compilation unit / C file), which this is not.  (I wonder if this is actually what clang was really complaining about?)

When it's part of an API like this, the right way to do it is to _define_ the function with the `inline` keyword in the header file, and then _declare_ the function as `extern inline` in exactly one C source file, so that there is exactly one non-inline implementation availble to the linker, in case one is still needed after optimisation (not all inlines will be inlined, it's a hint to the optimiser only, but it will make up its own mind).  Yes, this is exactly backwards from how a non-inline function is declared and defined; you get used to it.

The way you currently have it, every single compilation unit that includes this header will end up with its own, private, non-inlined copy of the function.  Since all those copies are static, they might not be removed during linking, even if all calls to them have been optimised out.  It'll work fine (in this case!) but it's a mess; it's not the optimisation you thought you were getting by marking it inline.

In practice, what this all means is you should:

1. remove `static` from this definition 
2. in lib/hash.c, add an `extern inline size_t hash_count(hash_table *table);` declaration, probably near the top somewhere

We haven't used this construct much (at all?) in Cyrus, so there probably aren't good examples to copy from, but I use it all the time in my own stuff so I'm very familiar with it.  Hit me up on slack if you'd like links to examples.

lib/hashu64.h
```diff
@@ -71,9 +72,12 @@ void *hashu64_del(uint64_t key,hashu64_table *table);
 void hashu64_enumerate(hashu64_table *table,void (*func)(uint64_t ,void *,void *),
                     void *rock);
 
+/* counts the number of nodes in the hash table */
 
-/* just count how many items are in the table */
-size_t hashu64_count(hashu64_table *table);
+static inline size_t hashu64_count(hashu64_table *table)
```
elliefm: (Same comment as hash.h as regards inline functions in headers.)

cassandane/tiny-tests/JMAPEmail/email_query_fromcontactcarduid
```diff
@@ -118,9 +118,7 @@ sub test_email_query_fromcontactcarduid
             ],
         }, 'R1']
     ], $using);
-    $self->assert_num_equals(2, scalar @{$res->[0][1]{ids}});
-    $self->assert_str_equals($emailId1, $res->[0][1]{ids}[0]);
-    $self->assert_str_equals($emailId2, $res->[0][1]{ids}[1]);
+    $self->assert_deep_equals([$emailId1, $emailId2], $res->[0][1]{ids});
```
elliefm, line 121: The commit message for this one misspells "assert_deep_equals" as "assert_cmp_deeply", which I assume is the equivalent function from some other testing API... 😅

cassandane/tiny-tests/JMAPNote/note_get
```diff
@@ -60,31 +60,24 @@ EOF
     $imap->append('INBOX.Notes', $note3) || die $@;
 
     my $res = $jmap->CallMethods([['Note/get', { }, "R1"]]);
-    $self->assert_cmp_deeply(
+    $self->assert_cmp_deeply([
```
elliefm, line 63: Well, this one _does_ actually use assert_cmp_deeply, huh.

cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db
```diff
@@ -0,0 +1,125 @@
...
+          # sqlite3_open_v2()
+          # Instead, with a corrupt file sqlite3_open_v2() succeeds but the call
+          # sqlite3_exec(open->db, "PRAGMA user_version;", ...) fais and logs.
+          if($path->basename() eq 'dav.db') {
+            push @found, $path;
+            $path->spew("this space for rent\n");
+          }
+        },
+        { recurse => 1 }
+    );
```
elliefm, line 100: There's some 2SP indents snuck in here.  Looks like there's some in the sieve snippet above too, but I'm less concerned about that because that's the contents of a string.  There's also a weird one a few lines below, the `localpart` argument to generate.

Though, now I see from the comment at the top of the file that this was a copy-and-tweak, and the messy indents came from the other file.

elliefm (changes requested):
Broadly, looks good.  A few style nits, but the only thing I'm actually concerned about is the misuses of `static inline`.  Please fix those, details below.

I also agree with Matt's comments about the change from accessing foo.size directly to accessing foo.table directly: we should add an inline function for this like hash_count, with a good name, that reflects its purpose at the sites its called from.  I don't have a specific suggestion, but I will note that "initialized" is not a great choice, because we already use `X_INITIALIZER` throughout Cyrus to mean "not random bits; probably, but not necessarily, all zeroes", so the internal table having been malloced is a step beyond merely "initialized".

nwc10:
> The commit message for "hash: hash_numrecords() is now static inline hash_count() returning size_t" can now be rewritten a bit. It still seems to mention this:

You rock! I completely missed that and it would have been confusing (and embarrassing) to have that in the final version.

What happened was that I tested the "can I remove the `#pragma`?" code as a commit at the head of the branch. Then when it worked, I moved it to the right place to amend the original.  But as the "original" was split across two commits I got "creative" and did `pick ...` for that commit twice in the interactive rebase, editing away the conflicts. Which distracted me from the fact that the commit message also needed editing.

nwc10:
> It feels weird to me that we replaced hash->size with hash->table. Both feel like they're leaking internals. It's fine, but I wonder if something like hash_initialized() (like hash_count()) would be better? /shrug
> 
> In any case, I think a commit message should be adjusted as noted just a little while ago, and maybe the corruption test could be improved a little bit, but it's also fine if you don't change that test.

On reflection, I think **yes**, but I'm not sure what name it should be. "hash_initialized" isn't quite correct, as a hash initialised by assigning `HASH_TABLE_INITIALIZER` to it **is** intialised. 

At least somewhat initialised. You can't actually call `hash_insert` on it.

So it's more "did someone call `contruct_hash_table` on it?". Which doesn't slip of the tongue.

The reason I think that this matters is that if we can find a clear name for this "I told you what size to start with" then we can both

1. keep the current lazy initialisation paradigm working
2. also add a further optimisation - make it optional to call `construct_hash_table` before the first call to `hash_insert`, and have the latter allocate the array `table` only if it's really needed.

Maybe related background here is that there seem to be two different paradigms for using the hash table code

1. cunit tests don't use `HASH_TABLE_INITIALIZER` and always call `construct_hash_table`
2. deployed code always uses either `HASH_TABLE_INITIALIZER` or `memset`, and might end up checking `->size` (or in this branch `->table`) or might end up calling `free_hash_table` on the struct without it ever being "constructed".

both work. But they aren't consistent.

cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db
nwc10: Agree. This is what comes from me writing the test one week, and only getting back writing the comment in the C code a week later. I didn't realise what I was actually testing.

Will be updated in a version I'm about to push

cassandane/tiny-tests/JMAPNote/note_get
nwc10: It's in `Cassandane::Unit::TestCase`

No, I didn't realise any of this until you commented on it. The confusion/inconsistency isn't ideal. IIRC I was just stealing code from whatever was nearby in the same test script.

But this branch has lived for too long, and parts of it are not fresh in my mind.

Thanks for your patient and careful reveiew, spotting things like this.

cassandane/tiny-tests/JMAPEmail/email_query_fromcontactcarduid
nwc10: I've fixed the commit message in the version I'm about to push

lib/hash.h
nwc10: Thanks. I'd stolen the approach used by the perl core, but I think it dated from back when people still used older gcc versions that didn't behave like C99

I've implemented what you said, but had quite a fight with clang and our `EXPORTED` macro to get it all to work.

**Now** we have something for future us to copy from...

cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db
nwc10: At this point I think I'd prefer to leave them in (as there's a comment from @alh somewhere in one of the PR threads) suggesting that we that the sieve script works **before** the test then goes and corrupts `dav.db` on disk, which basically means putting most of the test **back** that I'd cut out to try to get something minimal.

But I need to stop for today, so I'll get to that with a fresh head tomorrow

nwc10:
> I also agree with Matt's comments about the change from accessing foo.size directly to accessing foo.table directly: we should add an inline function for this like hash_count, with a good name, that reflects its purpose at the sites its called from. I don't have a specific suggestion, but I will note that "initialized" is not a great choice, because we already use `X_INITIALIZER` throughout Cyrus to mean "not random bits; probably, but not necessarily, all zeroes", so the internal table having been malloced is a step beyond merely "initialized".

Yes, we should do this with a function. I think we're all currently stuck on a good name. Thanks for the elegant description of the various states from "random bits" to "ready to rock"

I'll pause here and see if tomorrow brings inspiration.

cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db
nwc10: > Might it also be good to try a delivery first to make sure the message ends up in the .matches subfolder to make sure our sieve is right and that it didn't wind up in the inbox because of a typo?

Yes, that seems sensible, as that makes sure that the test is really testing what it thinks that it's testing.

In which case...

`cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db` is basically a copy of `cassandane/tiny-tests/JMAPEmail/email_query_fromanycontact_ignore_localpartonly` with that part removed!

```
$ diff -u  cassandane/tiny-tests/*/email_query_fromanycontact_ignore_localpartonly cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db | diffstat
 deliver_corrupt_dav_db |   80 +++++++++++++++++++++++++++++++++++++------------
 1 file changed, 62 insertions(+), 18 deletions(-)
```

so about 75% of the latter remains in the former:

```
$ wc -l cassandane/tiny-tests/*/email_query_fromanycontact_ignore_localpartonly cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db
      86 cassandane/tiny-tests/JMAPEmail/email_query_fromanycontact_ignore_localpartonly
     130 cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db
     216 total
```

I already had the "problem" that since I first forked the code, murch made some change in master to an API, and once rebased my fork failed until I copied the fix. DRY violation.

So should I just abolish the separate test and fold the "trash `dav.db` and test what changes" onto the end of `cassandane/tiny-tests/JMAPEmail/email_query_fromanycontact_ignore_localpartonly`?

cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db
nwc10: Apart from what's listed below, I believe I've addressed all the feedback

1. `inline` is how Ellie requested it be corrected. Plus a fight involving `EXPORT` that neither of us foresaw.
2. The lookup of `.table` is now a function called `hash_constructed(...)`. We might find a better name later, but it's now easy to find and change
3. In doing that I realised that as my new function took a `const` pointer, so should `hash_count()` so I changed that too

Ellie observed bad indents in (I think only) the file `cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db` which I added. alh is suggesting that we add to that file a test for "what happens before we trash the `dav.db` on disk?" which basically is almost all of `cassandane/tiny-tests/JMAPEmail/email_query_fromanycontact_ignore_localpartonly`. Which is where the sins against indenting came from.

So I'm wondering if I just merge the new "trash the file" tests onto the end of the existing `email_query_fromanycontact_ignore_localpartonly` and make sure that my new parts don't have silly indenting.

lib/hash.c
```diff
@@ -7,12 +7,43 @@
...
+ * header include here. However, to make that work, we also need a forward
+ * definition of hash_table.
+ *
+ * If there is a cleaner solution, patches welcome!
+ */
+
+typedef struct hash_table hash_table;
+EXPORTED extern inline size_t hash_count(const hash_table *table);
+#include "hash.h"
+
```
elliefm: Oh, fascinating.

If I understand correctly, the tricky inversion is needed so we can use the `EXPORTED` macro?  In that case, I don't think I would bother with the macro, just set the visibility attribute directly; every worthwhile compiler supports it these days, and it cannot hurt to set it to "default" no matter what else we're doing with `-fvisibility`.  Something like: 

```
[lib/hash.h]

inline size_t hash_count(const hash_table *table)
{
   // ...
} __attribute__(visibility("default"))
```

```
[lib/hash.c]

#include "lib/hash.h"

extern inline size_t hash_count(const hash_table *table);
```

My syntax might be off; I remember that attaching attributes to function definitions is funny, but I can never remember funny how off the top of my head.  Maybe put a brief comment nearby explaining that the attribute syntax is being used because `EXPORTED` isn't available to installed headers.

elliefm:
(For what it's worth)

> I think we're all currently stuck on a good name.

Since the thing we care about is "have we allocated the internal storage", maybe something like `hash_is_allocated()` or `hash_has_alloc()`.  Some of our other dynamic data structures (`struct buf` et al) have an `alloc` field for how many bytes have been allocated, so we're already accustomed to examining something called "alloc" in a boolean context.

The other way I'm thinking about it is that we're probably enquiring about the allocation because we want to know if we will need to call `free_hash_table()`, in which case `hash_needs_free()` would answer that specific question quite directly.

> `HASH_INITIALIZER`, memset, etc

`construct_hash_table()` doesn't check the current state of the object being constructed, it overwrites it unconditionally.  So it's fine to call it with uninitialised memory if you know that's what it is, which is how hash.testc uses it.  But usually you want to initialise it with zeroes so that later you can tell whether it's been constructed yet or not.

memset is not a great way to do this.  `HASH_INITIALIZER` (or whatever) is always the better choice when it works, but basically all of our initializer macros are defined in a way that's only syntactically valid for initializing; they can't be used for regular assignments.  So in cases that would count as an assignment, we have to fall back on memset.  We also use memset in a bunch of cases that could've used an initializer, I guess cause whoever wrote that code didn't know about them.

C99 has syntax ("compound literals") that we could use to make our initialiser macros also work for assignments, which would let us get rid of a lot of memsets.  I haven't gotten around to converting the existing stuff yet, but `LOGFMT_INITIALIZER` is an example of this in new code (though, for all that, we only use that one as an initializer).

nwc10:
> Since the thing we care about is "have we allocated the internal storage", maybe something like `hash_is_allocated()` or `hash_has_alloc()`. Some of our other dynamic data structures (`struct buf` et al) have an `alloc` field for how many bytes have been allocated, so we're already accustomed to examining something called "alloc" in a boolean context.
> 
> The other way I'm thinking about it is that we're probably enquiring about the allocation because we want to know if we will need to call `free_hash_table()`, in which case `hash_needs_free()` would answer that specific question quite directly.

I think both those names are rather too close to the implementation.

The current use case is piggybacking an out-of-band boolean flag onto the hash that's used for "did I do something else yet?" where that something

1. starts with `construct_hash_table` because that's setting the state of the (semantic) flag
2. does some other expensive work
3. stores the results in the hash

where the idiom is lazy initialisation/deferring work until it's known that it will be used

There's already code (since 2013) that assumes that it's always going to call `free_hash_table` without caring whether it was ever "construct"ed.

So semantically naming this function as if it's about conditionally-needing-to-free already seems distracting

But further, I can see a plausible optimisation where it really stops being about "do I need to free it"/"is something allocated"

If

1. the existing code I described above does "expensive work" but doesn't actually store any values in the hash (looks like this can happen sometimes), then currently `construct_hash_table` will do allocation of `.table` immediately, and there's definitely something that needs to be passed to `free`
2. but we could (and maybe could usefully) make allocation of `.table` only happen on the first actual *insert*
3. in which case, `construct_hash_table` has been called, but nothing allocated, so
    * if we call the function something "is allocated" then that's not true
    * nor is "needs free"
    * but *construct* was called. So `hash_constructed()` as a name remains true
    * even though it's really "did I set the OOB signalling flag?"

lib/hash.c
nwc10: The inversion is because the `EXPORTED` macro isn't even defined everywhere. As per the comment I added to `hash.c`, the example file `example_libcyrus_min.c` doesn't include `config.h`, and the test that tests it then fails.

I quite like the `EXPORTED` macro. I can type it and spell it.

An alternative would be

```
#ifndef EXPORTED
#define EXPORTED
#endif
```

and then use `EXPORTED` in the header.

But I was happy enough with the ordering dance. Without the explanation comment what's needed looks like this:

```
typedef struct hashu64_table hashu64_table;
EXPORTED extern inline size_t hashu64_count(const hashu64_table *table);
#include "hashu64.h"
```

which I think reads reasonably clearly, and remains consistent with all the rest of the C file.

And the header makes no mention of `EXPORTED`, which should make it easier to read.

elliefm: Yeah, I wrote those example C files, and the tests that use them, specifically to make sure the installed headers and libraries are actually (minimally) usable from external code.  I know all about the situation with config.h, though this PR is the first time I've really thought about EXPORTED in headers, since we didn't previously have inline functions that needed to be externally visible.

To someone who knows how to use C99 inlining, the inversion is weird and subtle and warrants explanatory comments.  The style I've suggested is standard; the only explanation it needs is why we're using attribute syntax directly instead of the usual Cyrus EXPORTED macro, and the explanation to that is simple: the macro is not available in this context.

The mess that is the general config.h situation might stick around for a long time (it's a nightmare to clean up, and not important to anybody but me), but for things we are actually cleaning up, we should make them standard when we can, even if that means differing from the usual Cyrus way, and especially so when the usual Cyrus way is known to have problems.  This PR will become the defacto style for future additions of inlined functions to Cyrus APIs, so I would like it to be a good example.

I would be fine with something like this in lib/hash.h if you really wanted:

```
#ifndef EXPORTED
#define EXPORTED __attribute__((visibility("default")))
#endif
```

and in fact we already have examples of this sort of thing, in cunit and cassandane.

But I think (not certain, haven't tested) that on a function _definition_ any attribute specification needs to come after the body.  If so, even with EXPORTED defined in hash.h, you wouldn't be using it in the familiar way at the start of the signature.  As a reader, to understand the difference from what you're used to seeing, you'd need to go chasing the macro definitions around, whereas to understand the explicit attribute syntax, you just look it up directly.

To be clear, I don't consider the EXPORTED macro itself to be a problem, just the caveats around its usage.  Someday those caveats might be resolved and we'll be able to use it everywhere; in the meantime, we can't use it everywhere, and we shouldn't contort ourselves into uncomfortable shapes trying to force it.

wolfsage:
Yeah, it seems to me that there's only a handful of cases where we lazily construct the hash table (based on the calls to hash_constructed() I see in this branch). And a lot of them don't do any other special work, so they *could* go away if hash tables construct themselves on the fly.

This makes it feel like the API should be:

```
hash_table foo = { 0 }; // Or HASH_TABLE_INITIALIZER..
hash_lookup("key", foo); // hash_lookup knows the table hasn't been constructed, short circuits and returns false
hash_insert("key", foo); // hash_insert knows the table hasn't been constructed, constructs it, inserts key, profit
...
```

We could still have hash_constucted() for the couple of places that want to know because they do other work like:
```
        if (!hash_constructed(args.bodyprops)) {
            _email_init_default_props(args.bodyprops);
        }
```

With the dynamic resizing work, not giving the hash a starting size maybe isn't terrible since it'll resize as needed, but if we still want to provide hints, then hash tables could have a starting size field perhaps:
```
hash_table foo = HASH_TABLE_INITIALIZER_STARTING_SIZE(32);
hash_lookup("key", foo); // Still not constructed until hash_insert is called
```

This means each hash_ method has an extra branch, but would that be so bad? Just thinking outloud...

elliefm:
The elephant in the room is that, with the current implementation of `construct_hash_table()` (which unconditionally overwrites the memory it was given), we cannot tell the difference between uninitialized memory that happens to have non-zero values in pertinent places, or a properly constructed object.  That is to say, (pseudo) code like this is currently valid:

```
hash_table ht; // uninitialized, possibly random gibberish

if (sometimes_false)
    construct_hash_table(&ht);

[...]

if (hash_constructed(&ht)) {
    // if we never constructed ht, its fields can still be randomly non-zero,
    // so queries of constructedness can yield false positives
    do_something_with_hash(&ht);
    // SIGSEGV or SIGBUS
}
```

I don't believe a code path like this exists, but with the current implementation of `construct_hash_table()`, anything that receives a pointer to a hash table cannot know with certainty whether it's real or bullshit; the only safe hash table is one you set up yourself.  There is no way to implement `hash_constructed()` without potential false positives.

`construct_hash_table()` ought to assert that the memory it is constructing into matches the pattern set by HASH_INITIALIZER (currently, but not necessarily, all zeroes).  With such an assertion in place, the code above would assert during the `construct_hash_table()` call, and we'd be forced to fix the missing initialisation.  With the initialization fixed, when the `hash_constructed()` call is reached, the table is reliably either initialized or constructed; no more mystery random third state.

Such an assertion would also prevent another anti-pattern that's currently possible:

```
hash_table foo = HASH_INITIALIZER; // even if we do this properly...

construct_hash_table(&foo); // fine
hash_insert(key, &foo);
[...]
construct_hash_table(&foo); // succeeds! new hash table constructed, old contents are leaked
```

nwc10:
I believe that I've addressed the two remaining reviewer requests, and this is good to merge

1. @elliefm - `hash.h` and `hashu64.h` now conditionally `#define EXPORTED` and the function declarations are in now in the human-logical order
2. @wolfsage - I updated `deliver_corrupt_dav_db` with your suggestions about the newer APIs to reduce boilerplate, and (re)added the test for "does it deliver to the correct place before we trash `dav.db`"

Branch rebased on current `master` - the only commits that meaningfully change are

* *deliver_corrupt_dav_db: test lmtpd delivery with a corrupt user dav.db*
* *hash: hash_numrecords() is now inline hash_count() returning size_t*

(two further commits get `EXPORTED` in different places)

wolfsage:
lgtm! I just have one nit in general but I won't press on it:
```
    $self->{instance}->getsyslog();
    $self->{instance}->deliver($msg2);
    my @lines = $self->{instance}->getsyslog();
    # Exactly what we log might change if we change the implementation. But I
    # think it better to write a potentially fragile test, because with it we
    # see whether the expected code paths are still reached.
    # Cassandane::Instance::_check_syslog() faults these if it sees them:
    my @dberror = grep /: DBERROR: get user version failed: sessionid=<[^>]+> fname=<\Q$found[0]\E> error=<file is not a database>/, @lines;
    $self->assert_num_equals(2, scalar @dberror, '2 expected fatal log lines found');
    # The actual error we're trying to reach doesn't default to a test failure:
    my @carddav_error = grep /: jmap: carddav_open_userid\(cassandane\) failed$/, @lines;
    $self->assert_num_equals(1, scalar @carddav_error, 'carddav_open_userid logged failure');
```

Vertical whitespace before comment blocks (and often separating unrelated actions which you've already done) make this all much easier to read and is good practice in general I think:
```
    $self->{instance}->getsyslog();
    $self->{instance}->deliver($msg2);
    my @lines = $self->{instance}->getsyslog();

    # Exactly what we log might change if we change the implementation. But I
    # think it better to write a potentially fragile test, because with it we
    # see whether the expected code paths are still reached.
    # Cassandane::Instance::_check_syslog() faults these if it sees them:
    my @dberror = grep /: DBERROR: get user version failed: sessionid=<[^>]+> fname=<\Q$found[0]\E> error=<file is not a database>/, @lines;
    $self->assert_num_equals(2, scalar @dberror, '2 expected fatal log lines found');

    # The actual error we're trying to reach doesn't default to a test failure:
    my @carddav_error = grep /: jmap: carddav_open_userid\(cassandane\) failed$/, @lines;
    $self->assert_num_equals(1, scalar @carddav_error, 'carddav_open_userid logged failure');
```

lib/hash.c
```diff
@@ -13,6 +13,38 @@
...
+ * definition (in the header) is seen first, that needs EXPORTED
+ * If we don't include config.h then EXPORTED isn't defined as anything, so it's
+ * left in place as-is, and it's a syntax error.
+ * sieve-lex.o transitively includes hash.h, but doesn't itself include config.h
+ * so we try to fix that by including config.h in hash.h
+ * We can make that work, but then example_libcyrus_min.c (and the test for it)
+ * break because that uses installed headers, and we don't install config.h
+ *
+ * One solution that clang and gcc both are happy with is to move the
+ * declaration ahead of the definition, by moving it before the header include
```
elliefm: I'm not going to harp on it further, but I don't think this comment is necessary.  (I'm referring to the whole comment; github only let me select part of it.)  With the changes you've made, the `extern inline...` line below is now perfectly standard.  The style is new to the _project_, but it's safe to assume a C programmer knows it, or will learn it.  More below...

lib/hash.h
```diff
@@ -8,6 +8,10 @@
 
 #define HASH_TABLE_INITIALIZER {0, 0, 0, 0, NULL, NULL}
 
+#ifndef EXPORTED
+#define EXPORTED __attribute__((visibility("default")))
+#endif
+
```
elliefm, line 18: If there should be an explanatory comment anywhere, it should be here, because **this is the place we're doing something strange**.  It doesn't need to be an essay though; something like  "EXPORTED isn't necessarily available in installed headers, so manually define it" would suffice.  This file being an installed header is the cause of the entire complication, so call out that distinction.  We can assume the reader is competent to figure out the ramifications of that, if they need to, and don't already just know.

We don't need to narrate the history, or apologise for using widely supported GCC extensions.  All of that is already in the commit message anyway.

lib/hashu64.h
```diff
@@ -3,7 +3,11 @@
 
 #include <stddef.h>           /* For size_t     */
 
-#define HASHU64_TABLE_INITIALIZER {0, NULL, NULL}
+#define HASHU64_TABLE_INITIALIZER {0, 0, NULL, NULL}
+
+#ifndef EXPORTED
+#define EXPORTED __attribute__((visibility("default")))
+#endif
```
elliefm, line 10: Same comment here as hash.h.  This is a place we're doing something strange, and that warrants an explanatory comment.  It can be brief, just needs to mention that this is an installed header, so it can't assume EXPORTED is already defined.

lib/hash.h
```diff
@@ -95,6 +96,26 @@ EXPORTED inline size_t hash_count(const hash_table *table)
...
+EXPORTED inline bool hash_constructed(const hash_table *table)
+{
+    /* Currently this tests whether .table is non-NULL, but in future we might
+     * change the implementation further such that table isn't even allocated
+     * until the first hash_insert(), so we hide this lookup in a trivial named
+     * function, instead of exposing implementation details that might change
+     * everywhere in the main codebase.
+     */
+
+    return table->table;
```
elliefm, line 129: Personally, I would make the conversion to `bool` explicit; at the moment, at a glance, it looks like it returns a pointer.  Something like `return !!table->table;`.  Though I do expect a C programmer in 2026 to know how `bool` works, so it's not necessary, just what I would do.

The internal comment basically just describes encapsulation.  I don't think we need that explained.

lib/hash.h
```diff
@@ -95,6 +96,26 @@ EXPORTED inline size_t hash_count(const hash_table *table)
     return table->count;
 }
 
+/* true if construct_hash_table() has been called on this hash table
+ * false if the memory has only been initialized with HASH_TABLE_INITIALIZER
+ * As free_hash_table() resets the memory to that same state, will be true after
+ * it has been called
+ * Encapsulates an idiom used by lots of Cyrus code to defer expensive
+ * initialization of hash tables until the first time they are actually needed
+ */
```
elliefm, line 118: As I explained in my previous review (the "elephant in the room" comment), this function isn't safe.  But what it replaces wasn't safe either, so it's not making anything _worse_.  But it's also not quite the improvement that it looks like it wants to be.

The provided pointer may be pointing to gibberish.  In that case, it is _not_ a constructed object, but if the gibberish happens to have a non-zero bit somewhere in the 8 bytes of the table field, this function will say it is.  I don't think the function is complicated or surprising enough to warrant a usage comment, but if there's going to be one, the one detail that's really critical is that calling it on uninitialized memory will have undefined results.

I'm stepping through this commit by commit; you may have already made the changes that would make this safe, and I just haven't seen them yet.

elliefm (approved):
(Accidentally hit submit too soon on previous review, d'oh.)

My comments at this point are mostly about the comments.  Some places do not call out details that they should call out; and some places have long comments that are disproportional to the complexity of what's being described.

I recognise that, in working on hash.c in particular (which is some ancient and half-assed code), you might expect some resistance to modernisation, and feel a need to justify your changes.  But we all want this modernised, we just hadn't done it ourselves yet.  So when writing comments, don't worry so much about explaining standard design choices or documenting language features; instead, think about what details a competent reader might miss or be surprised by, and document those.

There's nothing here really blocking anything, so I'm clicking approve now.  I hope you'll do something with my comment suggestions, but I don't need to re-review it if you do.

I see that you didn't do anything to make `hash_constucted()` safe.  That's fine for now, and a good choice even, because I suspect making it safe might be a breaking change.  We'll need it if we want to add lazy initialisation, but we can do it then.

nwc10:
Cleanup after !6229

* `imap/jmap_ical.c` is gone
* some changed code moved between files
* some changed code was extracted to a static function

PR #5952, opened 2026-05-06
dynamic hash tables

nwc10:
* round the hash table size up to the next power of two
* store the hash table size as log base 2
* calculate the bucket using bitshifts instead of a modulo operator
* dynamically resize the hash table as needed to maintain a maximum load factor
* mix a random 32 bit value into the bucket calculation to perturb the iterator order

Add `lib/hash_priv.h` for definitions needed in both `hash.c` and `hashu64.c`

### DIFF
--- a/cunit/hash.testc
+++ b/cunit/hash.testc
@@ -120,8 +120,8 @@ static void test_empty(void)
     hash_enumerate(&ht, count_cb, &count);
     CU_ASSERT_EQUAL(0, count);
 
-    /* check hash_numrecords */
-    CU_ASSERT_EQUAL(0, hash_numrecords(&ht));
+    /* check hash_count */
+    CU_ASSERT_EQUAL(0, hash_count(&ht));
 
     /* free the hash table */
     free_hash_table(&ht, NULL);
@@ -152,8 +152,8 @@ static void test_reinsert(void)
     hash_enumerate(&ht, count_cb, &count);
     CU_ASSERT_EQUAL(1, count);
 
-    /* check hash_numrecords */
-    CU_ASSERT_EQUAL(1, hash_numrecords(&ht));
+    /* check hash_count */
+    CU_ASSERT_EQUAL(1, hash_count(&ht));
 
     /* re-insert into the hash table */
     d = hash_insert(KEY0, VALUE1, &ht);
@@ -169,8 +169,8 @@ static void test_reinsert(void)
     hash_enumerate(&ht, count_cb, &count);
     CU_ASSERT_EQUAL(1, count);
 
-    /* check hash_numrecords */
-    CU_ASSERT_EQUAL(1, hash_numrecords(&ht));
+    /* check hash_count */
+    CU_ASSERT_EQUAL(1, hash_count(&ht));
 
     /* delete from the hash table */
     d = hash_del(KEY0, &ht);
@@ -185,8 +185,8 @@ static void test_reinsert(void)
     hash_enumerate(&ht, count_cb, &count);
     CU_ASSERT_EQUAL(0, count);
 
-    /* check hash_numrecords */
-    CU_ASSERT_EQUAL(0, hash_numrecords(&ht));
+    /* check hash_count */
+    CU_ASSERT_EQUAL(0, hash_count(&ht));
 
     /* free the hash table */
     free_hash_table(&ht, NULL);
@@ -254,8 +254,8 @@ static void test_many(void)
     hash_enumerate(&ht, count_cb, &count);
     CU_ASSERT_EQUAL(N, count);
 
-    /* check hash_numrecords */
-    CU_ASSERT_EQUAL(N, hash_numrecords(&ht));
+    /* check hash_count */
+    CU_ASSERT_EQUAL(N, hash_count(&ht));
 
     /* delete from the hash table */
     for (i = 0 ; i < N ; i++) {
@@ -274,8 +274,8 @@ static void test_many(void)
     hash_enumerate(&ht, count_cb, &count);
     CU_ASSERT_EQUAL(0, count);
 
-    /* check hash_numrecords */
-    CU_ASSERT_EQUAL(0, hash_numrecords(&ht));
+    /* check hash_count */
+    CU_ASSERT_EQUAL(0, hash_count(&ht));
 
     /* free the hash table */
     freed_count = 0;
@@ -307,8 +307,8 @@ static void test_freeing_nonempty(void)
     hash_enumerate(&ht, count_cb, &count);
     CU_ASSERT_EQUAL(N, count);
 
-    /* check hash_numrecords */
-    CU_ASSERT_EQUAL(N, hash_numrecords(&ht));
+    /* check hash_count */
+    CU_ASSERT_EQUAL(N, hash_count(&ht));
 
     /* free the hash table */
     freed_count = 0;
@@ -399,7 +399,7 @@ static void test_load_factor_warning(void)
     for (i = 0; i < n_words; i++) {
         hash_insert(words[i], NULL, &ht);
     }
-    CU_ASSERT_EQUAL(n_words, hash_numrecords(&ht));
+    CU_ASSERT_EQUAL(n_words, hash_count(&ht));
     CU_ASSERT_SYSLOG(syslog_index, 2);
 
     free_hash_table(&ht, NULL);
--- a/cunit/hash.testc
+++ b/cunit/hash.testc
@@ -394,13 +394,11 @@ static void test_load_factor_warning(void)
     CU_ASSERT_EQUAL(3, ht.size_log2);
 
     syslog_index = CU_SYSLOG_MATCH("hash table load factor exceeds 3.0");
-    /* 8 buckets will hit load factor 3.0 after 24 insertions, 4.0 after 32
-     * insertions, and it won't reach 5.0, so we should warn exactly twice */
     for (i = 0; i < n_words; i++) {
         hash_insert(words[i], NULL, &ht);
     }
     CU_ASSERT_EQUAL(n_words, hash_count(&ht));
-    CU_ASSERT_SYSLOG(syslog_index, 2);
+    CU_ASSERT_SYSLOG(syslog_index, 0);
 
     free_hash_table(&ht, NULL);
 }
@@ -408,6 +406,7 @@ static void test_load_factor_warning(void)
 static void test_hash_size_rounding(void)
 {
     size_t tests[][2] = {
+        { 0, 3 },
         { 1, 3 },
         { 2, 3 },
         { 3, 3 },
--- a/cunit/hash.testc
+++ b/cunit/hash.testc
@@ -391,7 +391,7 @@ static void test_load_factor_warning(void)
     CU_ASSERT_EQUAL_FATAL(33, n_words);
 
     construct_hash_table(&ht, 5, 0);
-    CU_ASSERT_EQUAL(8, ht.size);
+    CU_ASSERT_EQUAL(3, ht.size_log2);
 
     syslog_index = CU_SYSLOG_MATCH("hash table load factor exceeds 3.0");
     /* 8 buckets will hit load factor 3.0 after 24 insertions, 4.0 after 32
@@ -408,26 +408,26 @@ static void test_load_factor_warning(void)
 static void test_hash_size_rounding(void)
 {
     size_t tests[][2] = {
-        { 1, 8 },
-        { 2, 8 },
-        { 3, 8 },
-        { 4, 8 },
-        { 5, 8 },
-        { 6, 8 },
-        { 7, 8 },
-        { 8, 8 },
-        { 9, 16 },
-        { 15, 16 },
-        { 16, 16 },
-        { 17, 32 },
-        { 42, 64 },
-        { 54, 64 },
-        { 255, 256 },
-        { 256, 256 },
-        { 257, 512 },
-        { 65535, 65536 },
-        { 65536, 65536 },
-        { 65537, 131072 },
+        { 1, 3 },
+        { 2, 3 },
+        { 3, 3 },
+        { 4, 3 },
+        { 5, 3 },
+        { 6, 3 },
+        { 7, 3 },
+        { 8, 3 },
+        { 9, 4 },
+        { 15, 4 },
+        { 16, 4 },
+        { 17, 5 },
+        { 42, 6 },
+        { 54, 6 },
+        { 255, 8 },
+        { 256, 8 },
+        { 257, 9 },
+        { 65535, 16 },
+        { 65536, 16 },
+        { 65537, 17 },
     };
 
     unsigned i = sizeof(tests) / sizeof(size_t) / 2;
@@ -437,7 +437,7 @@ static void test_hash_size_rounding(void)
         hash_table ht;
 
         construct_hash_table(&ht, test[0], 0);
-        CU_ASSERT_EQUAL(test[1], ht.size);
+        CU_ASSERT_EQUAL(test[1], ht.size_log2);
         free_hash_table(&ht, NULL);
     }
 }
--- a/cunit/hash.testc
+++ b/cunit/hash.testc
@@ -377,24 +377,24 @@ static void test_load_factor_warning(void)
     const char *const words[] = {
         "id", "faucibus", "nisl", "tincidunt", "eget", "nullam", "non", "nisi",
         "est", "sit", "amet", "facilisis", "magna", "etiam", "tempor", "orci",
-        "eu", "lobortis", "elementum", "nibh", "tellus", "molestie"
+        "eu", "lobortis", "elementum", "nibh", "tellus", "molestie", "congue",
+        "consequat", "ac", "felis", "volutpat", "mauris", "nunc", "sed",
+        "augue", "lacus", "viverra"
     };
-    const size_t n_words = sizeof words / sizeof words[0]; /* 22 */
-    const size_t n_buckets = n_words / 4; /* 5 */
+    const size_t n_words = sizeof words / sizeof words[0]; /* 33 */
     unsigned i;
     unsigned int syslog_index;
 
     hash_table ht;
 
     /* make sure numeric assumptions hold */
-    CU_ASSERT_EQUAL_FATAL(22, n_words);
-    CU_ASSERT_EQUAL_FATAL(5, n_buckets);
+    CU_ASSERT_EQUAL_FATAL(33, n_words);
 
-    construct_hash_table(&ht, n_buckets, 0);
-    CU_ASSERT_EQUAL(n_buckets, ht.size);
+    construct_hash_table(&ht, 5, 0);
+    CU_ASSERT_EQUAL(8, ht.size);
 
     syslog_index = CU_SYSLOG_MATCH("hash table load factor exceeds 3.0");
-    /* 5 buckets will hit load factor 3.0 after 15 insertions, 4.0 after 20
+    /* 8 buckets will hit load factor 3.0 after 24 insertions, 4.0 after 32
      * insertions, and it won't reach 5.0, so we should warn exactly twice */
     for (i = 0; i < n_words; i++) {
         hash_insert(words[i], NULL, &ht);
@@ -403,6 +403,43 @@ static void test_load_factor_warning(void)
     CU_ASSERT_SYSLOG(syslog_index, 2);
 
     free_hash_table(&ht, NULL);
+}
+
+static void test_hash_size_rounding(void)
+{
+    size_t tests[][2] = {
+        { 1, 8 },
+        { 2, 8 },
+        { 3, 8 },
+        { 4, 8 },
+        { 5, 8 },
+        { 6, 8 },
+        { 7, 8 },
+        { 8, 8 },
+        { 9, 16 },
+        { 15, 16 },
+        { 16, 16 },
+        { 17, 32 },
+        { 42, 64 },
+        { 54, 64 },
+        { 255, 256 },
+        { 256, 256 },
+        { 257, 512 },
+        { 65535, 65536 },
+        { 65536, 65536 },
+        { 65537, 131072 },
+    };
+
+    unsigned i = sizeof(tests) / sizeof(size_t) / 2;
+
+    while(--i) {
+        size_t *test = tests[i];
+        hash_table ht;
+
+        construct_hash_table(&ht, test[0], 0);
+        CU_ASSERT_EQUAL(test[1], ht.size);
+        free_hash_table(&ht, NULL);
+    }
 }
 
 /* vim: set ft=c: */
--- a/cunit/hash.testc
+++ b/cunit/hash.testc
@@ -372,42 +372,10 @@ static void test_iter(void)
     free_hash_table(&ht, NULL);
 }
 
-static void test_load_factor_warning(void)
-{
-    const char *const words[] = {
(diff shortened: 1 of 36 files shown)


nwc10:
This PR is 4 further commits atop #5950

lib/hash.h
```diff
@@ -89,9 +89,29 @@ void hash_enumerate_sorted(hash_table *table,void (*func)(const char *,void *,vo
 /* gets all the keys from the hashtable */
 strarray_t *hash_keys(const hash_table *table);
 
+/* Clang has decided that unused `static inline` functions should trigger a
```
wolfsage: It's a bit more silly/weird than that.

When building with gcc:
```
checking for inline... inline
```

And config.h has:
```
/* Define to `__inline__' or `__inline' if that's what the C compiler
   calls it, or to nothing if 'inline' is not supported under any name.  */
#ifndef __cplusplus
/* #undef inline */
#endif

```

When building with clang:
```
checking for inline... no
```

and config.h has:
```
/* Define to `__inline__' or `__inline' if that's what the C compiler
   calls it, or to nothing if 'inline' is not supported under any name.  */
#ifndef __cplusplus
#define inline 
#endif

```

So with clang the function isn't declared as inline, hence the warning.

Wild.

(With light searching I haven't seen obvious explanations or solutions to tell automake to detect that clang can inline things).

Interestingly... https://clang.llvm.org/compatibility.html#inline

But I have yet to trick automake into detecting inline support in clang. Not sure what we want to do here. (Mostly, we tend to use clang as a diagnostic build, but if we expect people to use it we should get it building optimally)

wolfsage: Ah, because the test automake does is:
```
| #ifndef __cplusplus
| typedef int foo_t;
| static inline foo_t static_foo (void) {return 0; }
| inline foo_t foo (void) {return 0; }
| #endif
| 
configure:7069: clang -c  -g -fPIC -W -Wall -Wextra -Werror -Wwrite-strings -Wformat=2   conftest.c >&5
conftest.c:43:25: error: unused function 'static_foo' [-Werror,-Wunused-function]
static __inline__ foo_t static_foo (void) {return 0; }

```

and clang indeed does not like a static inline function in a .c file with -Wall. (Feels like a bug in automake to me...)

nwc10: Aaargh, it's the same "feature":

```
configure:7052: checking for inline
configure:7069: clang -c  -g -fPIC -W -Wall -Wextra -Werror -Wwrite-strings -Wformat=2   conftest.c >&5
conftest.c:43:21: error: unused function 'static_foo' [-Werror,-Wunused-function]
static inline foo_t static_foo (void) {return 0; }
                    ^
1 error generated.
configure:7069: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "cyrus-imapd"
| #define PACKAGE_TARNAME "cyrus-imapd"
| #define PACKAGE_VERSION "3.13.4-48-g60b588652"
| #define PACKAGE_STRING "cyrus-imapd 3.13.4-48-g60b588652"
| #define PACKAGE_BUGREPORT "https://github.com/cyrusimap/cyrus-imapd/issues"
| #define PACKAGE_URL "https://www.cyrusimap.org"
| #define PACKAGE "cyrus-imapd"
| #define VERSION "3.13.4-48-g60b588652"
| #define CYRUS_USER "cyrus"
| #define HAVE_STDIO_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_INTTYPES_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_SYS_STAT_H 1
| #define HAVE_SYS_TYPES_H 1
| #define HAVE_UNISTD_H 1
| #define HAVE_WCHAR_H 1
| #define HAVE_SYS_TIME_H 1
| #define STDC_HEADERS 1
| #define _ALL_SOURCE 1
| #define _DARWIN_C_SOURCE 1
| #define _GNU_SOURCE 1
| #define _HPUX_ALT_XOPEN_SOCKET_API 1
| #define _NETBSD_SOURCE 1
| #define _OPENBSD_SOURCE 1
| #define _POSIX_PTHREAD_SEMANTICS 1
| #define __STDC_WANT_IEC_60559_ATTRIBS_EXT__ 1
| #define __STDC_WANT_IEC_60559_BFP_EXT__ 1
| #define __STDC_WANT_IEC_60559_DFP_EXT__ 1
| #define __STDC_WANT_IEC_60559_FUNCS_EXT__ 1
| #define __STDC_WANT_IEC_60559_TYPES_EXT__ 1
| #define __STDC_WANT_LIB_EXT2__ 1
| #define __STDC_WANT_MATH_SPEC_FUNCS__ 1
| #define _TANDEM_SOURCE 1
| #define __EXTENSIONS__ 1
| #define HAVE_LONG_FILE_NAMES 1
| /* end confdefs.h.  */
| #ifndef __cplusplus
| typedef int foo_t;
| static inline foo_t static_foo (void) {return 0; }
| inline foo_t foo (void) {return 0; }
| #endif
|
```

and then the same for attempts at `__inline__` and `__inline`

Because `clang` "feels" that this is worthy of a warning, and `-Werror` promotes those, `autoconf` is (correctly) determining that something is screwed up about `inline`

I'm not sure what to do here either. When I am dictator of the universe I'd bang some heads together. But until then, it seems that different projects have different ideas about correctness, and we're stuck in the middle.

imap/jmap_mail_query.c
```diff
@@ -260,6 +264,25 @@ HIDDEN int jmap_email_contactfilter_from_filtercondition(json_t *filter,
     /* Open CardDAV db for accountid */
     struct carddav_db *carddavdb = carddav_open_userid(cfilter->accountid);
     if (!carddavdb) {
+        /* I believe this code to be unreachable. For carddav_open_userid() to
```
wolfsage: This can also be called from lmtpd through jmap_email_matchmime() and http_dblookup methods. 

The http_dblookup looks safe as it checks first for failures and bails out. I'm not sure lmtpd does though.

It's opening the dav db in sieve_find_script which does:
```
    if (script) {
        snprintf(fname, size, "%s/%s.bc", sievedir, script);
        sieve_script_rebuild(userid, sievedir, script);
        r = 0;
    }
```

sieve_script_rebuild opens the dav db, if it fails it returns an error, but sieve_script_rebuild ignores it

wolfsage (commented):
> hash: Round the hash table size up to a power of 2

This commit could use an explainer perhaps to say why or how this relates to using a mask with bit shifting instead of modulo. (And why is that better than modulo?)

imap/jmap_mail_query.c
nwc10: This commit is technically in #5950...

OK, interesting, thanks.

I'll see if I can write a regression test to cover this code path. Although, in turn, it's not clear whether the current behaviour is a bug

lib/hash.h
wolfsage: Ah okay so, this is self inflicted:

https://github.com/cyrusimap/cyrus-docker/blob/master/Debian/lib/Cyrus/Docker/Command/build.pm#L188

We set a bunch of cflags before calling configure (including -Wall, -Wextra, -Werror).

Maybe we don't want to do that; or at least for clang we want to include -Wno-unused-function.

This works for example:
```
cyd build --clang --cflags=-Wno-unused-function
```
and automake finds inline as usable. 

@elliefm Maybe you have thoughts about what flags we do or do not want to use with Configure

nwc10: Or maybe we don't want to add `-Werror` until after autoconf.

This might seem crazy (it's not "test as you fly") but if we want `-Werror` to actually find warnings that would trigger, then what we have now causes autoconf to drop features because it thinks that the compiler doesn't support them. Whereas what (I think) our intent is, is to have the features enabled and then use `-Werror` to find the places that they trigger by making the build break hard.

nwc10:
> > hash: Round the hash table size up to a power of 2
> 
> This commit could use an explainer perhaps to say why or how this relates to using a mask with bit shifting instead of modulo. (And why is that better than modulo?)

:-) You are a hard task master. But to be fair, I **knew** about this one. I just had to find it again:

https://probablydance.com/2017/02/26/i-wrote-the-fastest-hashtable/

> So then how do we solve the problem of the slow integer modulo? For this I’m using a trick that I copied from boost::multi_index: I make all integer modulos use a compile time constant. I don’t allow all possible prime numbers as sizes for the table. Instead I have a selection of pre-picked prime numbers and will always grow the table to the next largest one out of that list. Then I store the index of the number that your table has. When it later comes time to do the integer modulo to assign the hash value to a slot, you will see that my code does this:
> 
> switch(prime_index)
> {
> case 0:
>     return 0llu;
> case 1:
>     return hash % 2llu;
> case 2:
>     return hash % 3llu;
> 
> ...
> 

nutty. And then in trying to find an explainer at CPU clock cycle level, I found that someone had actually done **all** my homework for me:

https://johnnysswlab.com/make-your-programs-run-faster-avoid-expensive-instructions/

> To test the speed difference, we implemented an [open addressing hash map](https://github.com/ibogosavljevic/johnysswlab/blob/master/2020-09-expensiveinstr/hash_map.h) whose size is always the power of two. We used working sets of several sizes and tested it. The results are in the table below:

and shifting is between 12% and 40% faster. Which was insanely more than I thought.

I'll add these (roughly) as a comment somewhere findable on the rebased version.

(the last part is [citation needed] but people have favoured integer modulo with a size that is a prime number to avoid (believed?) patterns in weak hash functions. Which sounds like a technical solution to a "social" problem - get a better hash function)

imap/jmap_mail_query.c
nwc10: I've currently got a rats nest of global variables and `abort()`s, but I think I can see how to trigger this code path. But it's going to take a fresh head (ie Monday) to take an existing test and copy and adapt it to hit the error path. (I have all the bits locally. It's just putting them together)

wolfsage: .oO(A cassandane test that deletes the dav db on disk and then delivers a message...?)

wolfsage:
> :-) You are a hard task master.

Apologies if it seems too much! But I've been banging on the "commits should explain why" drum pretty hard lately :)

lib/hash.h
elliefm: This is complicated, and there's multiple moving parts that I don't have swapped in at the moment.  Also, I believe (but haven't confirmed) that clang's -Werror works differently from gcc's, which complicates the whole mess further, but of the two behaviours I'm thinking of, I don't remember which compiler does which...

IIRC, autoconf generally does not expect `-Werror` to be on, _but we do_.  `cyd` is doing the correct thing as per our expectations.  But since autoconf does not expect -Werror to be on, its feature detections macros often rely on minimal code chunks that compilers warn about, which then break with -Werror.

We usually wrap checks that break like that in a block that saves/modifies/restores CFLAGS.  Look for "saved_CFLAGS" in configure.ac for existing examples.  You probably just need to put one of these blocks around the `AC_C_INLINE` check.

nwc10: It's taken me a long time to get there, but I think that the right solution is just to remove the entire `AC_C_INLINE` check - #5991

imap/jmap_mail_query.c
nwc10: You actually have to corrupt the dav db on disk. If you "just" delete it, it respawns. (This isn't treated as an error)

nwc10:
> > :-) You are a hard task master.
> 
> Apologies if it seems too much! But I've been banging on the "commits should explain why" drum pretty hard lately :)

It's a policy I agree with. I'm looking back at previous commits and wondering "why?". It was probably obvious at the time to the folks involved, but it's not to me, and it's probably not to them either. (Personal experience is that if I ask myself "what were you thinking when you did this six months ago?" past me answers "I forgot" and present me cusses)

lib/hash_priv.h
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * 0) encode a binary De Bruijn sequence into an integer constant
+ * 1) do some bit manipulation tricks that convert the input into values that
+ *    are powers of 2 (so exactly 1 bit is set) - 2 ** $n
+ * 2) multiply the two (which is equivalent to a logical left shift)
+ * 3) take the top $n bits (shift down) and index those into an array
+ * 4) which gives the value of $n
+ *
+ * However, *that* code has been changed:
+ * ... shaved off a couple operations by requiring v be rounded up to one less         than the next power of 2 rather than the power of 2.
```
wolfsage: nit: weird line space break here

lib/hash_priv.h
```diff
@@ -0,0 +1,87 @@
...
+
+static uint64_t round_up_log_base2(uint64_t v) {
+  static const uint8_t MultiplyDeBruijnBitPosition[64] = {
+    0, 1, 2, 7, 3, 13, 8, 19, 4, 25, 14, 28, 9, 34, 20, 40, 5, 17, 26, 38, 15,
+    46, 29, 48, 10, 31, 35, 54, 21, 50, 41, 57, 63, 6, 12, 18, 24, 27, 33, 39,
+    16, 37, 45, 47, 30, 53, 49, 56, 62, 11, 23, 32, 36, 44, 52, 55, 61, 22, 43,
+    51, 60, 42, 59, 58
+  };
+
+  /* this rounds (v - 1) down to one less than a power of 2 */
```
wolfsage: This comment is unclear to me.

Is it saying "if v is exactly a power of two this drops us below it"?

Or is it saying "This block of bit manipulations rounds down v to 1 less than a power of 2" and then the comment below is what puts us at the power of 2?

If it's the latter, that seems wrong - it rounds v *up* to 1 less than a power of 2.

That is 10 becomes 15, 13 becomes 15, etc etc.

lib/hash.c
```diff
@@ -48,6 +48,25 @@ struct bucket {
...
 */
 
+static size_t table_size(const hash_table *table) {
+    return table && table->table ? (1ULL << table->size_log2) : 0;
+}
+
+/* The arithmetic in this code is 64 bit so that
+ * 1) it's future proof if we move to a 64 bit hash
+ * 2) it's consistent with the hashu64 code
+ * 3) the known working sizeof(size_t) doesn't need changing and deubgging
```
wolfsage: nit: deubgging -> debugging

wolfsage: Why do we bother with 8*sizeof(size_t) instead of using 64? On anything where sizeof(size_t) != 8 we'll get wildly wrong values won't we?

lib/hash.c
```diff
@@ -48,6 +48,25 @@ struct bucket {
...
+ *
+ * Multiplying by 9e3779b97f4a7c15ULL and shifting is Fibonacci Hashing
+ * It's not great in itself, but it's computationally cheap. We use this as a
+ * final mixing stage as djbx33x lacks one, hence leaves all the recent entropy
+ * in the lowest bits of the hash.
+ */
+
+static size_t table_index(const hash_table *table, uint32_t hash) {
```
wolfsage: Can/Should we inline this and table_size above?

wolfsage (commented):
Some questions/thoughts...

Will review the dynamic resizing tomorrow

lib/hash.c
```diff
@@ -52,6 +52,111 @@ static size_t table_size(const hash_table *table) {
...
+** Cyrus) which is behind a layer protecting it from brute force flooding.
+**
+** The attacker knows
+** * the hash function (converting string keys to integers)
+** * the hash table structure (here an array of linked lists)
+** * the threshold where the hash resizes
+**
+** The attacker controls
+** * the strings for the keys
+** * the order order that those keys are inserted
```
wolfsage: nit: "order order"

lib/hash.c
```diff
@@ -52,6 +52,111 @@ static size_t table_size(const hash_table *table) {
...
+** and with a small number of keys one can rapidly (attempt to) exhaust all
+** memory on the victim machine, likely instead causing the process to die with
+** an allocation attempt for an array with 2**29 (or 2**30) entries
+**
+** Hence
+**
+** 1) the only safe trigger for splitting appears to be "load factor"
+** 2) it's important to hide the iteration order
+**
+** Perl 5 achieves the latter by randomising iteration order for each iteration
```
wolfsage: nit: This line probably needs a period at the end given it's a part of a bigger paragraph.

lib/hash.c
```diff
@@ -52,6 +52,111 @@ static size_t table_size(const hash_table *table) {
...
+** I believe that whilst correct, given that the attacker already knows the
+** timing of the array split, there's no need to conceal when the split happened
```
wolfsage: Same here... (Lots of lines are missing periods when others have them. Might be good to be consistent

lib/hash.c
```diff
@@ -52,6 +52,111 @@ static size_t table_size(const hash_table *table) {
...
+** Hence it's equivalently secure simply to ensure that for any level of "split"
+** the iteration order gives no clue about the low (or high) $n bits of the
+** full hash code used to pick the bucket.
+**
+** The hash code needs to stay the same as the hash splits (to avoid the
+** performance hit of recomputing every key's hash code)
+** Hence this code mixes a random 32 bit integer into the hash code before
+** computing the bucks, and changes that integer each time the array splits.
```
wolfsage: nit: bucks?

lib/hash.c
nwc10: You are totally correct.

The first version of the code (before it was pushed) was doing 32 bit calculations and IIRC I'd written it as sizeof(something) because I seem to have got into a (not always great) habit of doing that. Something to do with starting to program C when 32 bit platforms were ubiquitous and DEC Alphas were on a floor which I didn't have access to. So I later got to enjoy the transition to "maybe you can't just assume 32 bits"

So I was getting confused and thinking that this calculation depends on the size of pointers. Which (you are right) it doesn't.

I'm going to split it as two expressions, where the first assigns to a variable of type `uint64_t` and the second can confidently just be `64`. Because some day somewhere the code might reach a platform where `ULL` constants are 128 bits, and the intermediate assignment will explicitly ensure the modulo 64 arithmetic the shifting assumed.

lib/hash.c
nwc10: what do you mean "inline"?

I don't think you mean "manually inline them" as `table_index` is used in 5 places in each file, and `table_size` is 5 in `hash.c` and 2 in `hashu64.c`

Did you mean declare them as `static inline`?

If so, I was sort of getting lazy and assuming that the compiler would do the inlining for me once the optimiser is turned on.

imap/jmap_mail_query.c
nwc10: I think that this conversation is resolved. Now that we have a test, there's a different conversation about where the best file for it is. (a different and also useful conversation)

lib/hash_priv.h
nwc10: oh my yes. fixed locally, pushing soon

lib/hash_priv.h
nwc10: Oh my yes, re-reading it, it isn't clear. Better comment locally, pushing soon.

lib/hash.c
nwc10: bucket!

D'oh!

lib/hash.c
nwc10: D'oh!

Please keep this up. At Geizhals my colleague Matthias would say "this is going to warn if ..." and he was inevitably right. After a while I'd think

* this is ready to push
* oh wait, what is Matthias going to say?
* and re-read it pretending to be Matthias, and amazingly this would actually cause me to find problems that I'd missed

nwc10:
I believe I've resolved all the feedback on the 4 commits exclusive to this branch.

We've still got an open question on #5950 about `cassandane/tiny-tests/JMAPSieve/deliver_corrupt_dav_db` and `cassandane/tiny-tests/JMAPEmail/email_query_fromanycontact_ignore_localpartonly`

nwc10:
Rebased onto #5950 with no changes

lib/hash.c
wolfsage: I meant static inline like hash_count and hash_constructed are. These seem small enough, so didn't know if it was worth it or where/how we decided whether to inline them or not.

-Og is a common build style so I'm not sure we can rely on the optimizer to do the work for us

cunit/hash.testc
```diff
@@ -394,20 +394,19 @@ static void test_load_factor_warning(void)
...
 
     syslog_index = CU_SYSLOG_MATCH("hash table load factor exceeds 3.0");
-    /* 8 buckets will hit load factor 3.0 after 24 insertions, 4.0 after 32
-     * insertions, and it won't reach 5.0, so we should warn exactly twice */
     for (i = 0; i < n_words; i++) {
         hash_insert(words[i], NULL, &ht);
     }
     CU_ASSERT_EQUAL(n_words, hash_count(&ht));
-    CU_ASSERT_SYSLOG(syslog_index, 2);
+    CU_ASSERT_SYSLOG(syslog_index, 0);
```
wolfsage: Since you've deleted the warning entirely, I think this whole "test_load_factor_warning" test can go away

wolfsage (changes requested):

lib/hash.c
nwc10: oops, that `3) ...` comment was still there. Have just deleted it.

cunit/hash.testc
nwc10: gone!

lib/hash.c
nwc10: Good hunch. For `gcc (Debian 14.2.0-19) 14.2.0`

With `static size_t table_index`

* `-g` does not inline (no surprise)
* `-Og` does not
* `-O2` does

with `static inline size_t table_index`, `-O2` is byte for byte identical assembler, `-g` changes some labels (but identical instructions), but `-Og` now inlines.

So I think that this is worth it, and will update the relevant commit(s) that introduce the two copies of this function.

nwc10: Done

nwc10:
I believe that I've addressed all requested changes, and this is ready for a (final) reviewer to tag as "include in fastmail"

The branch (here) is atop https://github.com/cyrusimap/cyrus-imapd/pull/6272 but

1. that will be merged to master soon
2. as I understand it "include in fastmail" is an octopus merge to a private branch, so this commit will arrive for either state of master, current or future

wolfsage (approved):
lgtm!

nwc10:
> (the last part is [citation needed] but people have favoured integer modulo with a size that is a prime number to avoid (believed?) patterns in weak hash functions. Which sounds like a technical solution to a "social" problem - get a better hash function)

Er, yes, oops, as we found ourselves. Large performance gain from:

```
diff --git a/imap/search_expr.c b/imap/search_expr.c
index ead49c68a0..ac3d5c8f68 100644
--- a/imap/search_expr.c
+++ b/imap/search_expr.c
@@ -827,7 +827,7 @@ static void sort_children(search_expr_t *e)
 {
     search_expr_t *child;
     hashu64_table maxcostcache = HASHU64_TABLE_INITIALIZER;
-    construct_hashu64_table(&maxcostcache, 512, 0);
+    construct_hashu64_table(&maxcostcache, 509, 0);
     hashu64_table *costcache = &maxcostcache;

     if (sizeof(uint64_t) < sizeof(search_expr_t*)) {
```

fixed instead by the multiplication/shift in

```
static size_t table_index(const hashu64_table *table, uint64_t key) {
    uint64_t mixed = (key ^ table->chaff) * 0x9e3779b97f4a7c15ULL;
    return mixed >> (64 - table->size_log2);
}
```

PR #5953, opened 2026-05-08
imapopts: more improvements to generated C source 

elliefm:
Another housekeeping change, but this time the changes have substance.

(This is a prerequisite for real changes I have in the pipeline.  Getting the housekeeping out of the way first will make the real changes that follow neat and easy to reason about.)

**Reviewers:** I suggest reading the diffs commit by commit.

### DIFF
--- a/cunit/libconfig.testc
+++ b/cunit/libconfig.testc
@@ -33,8 +33,8 @@ static int tear_down(void)
 
 static void test_int(void)
 {
-    int boundary_limit = -1;
-    int conversations_max_thread = -1;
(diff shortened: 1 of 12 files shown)


rjbs:
I have no objection to the perl changes I saw, and it seems fine to me, but I have removed myself rather than approved, as the meat of the change seems to be around the generated C.  To me it is occasionally clearly better, and other times I have no opinion.  Better for a C-head to really review.

ksmurchison (approved):

rsto (approved):

PR #5954, opened 2026-05-08
Add a test showing isDefault JMAP Calendar doesn't persist

wolfsage:
If the config option `caldav_create_default` is disabled, cyrus will pick different calendar ids in different calls to Calendar/get.

It should pick one and persist it...


ksmurchison:
Superseded by #5964

PR #5955, opened 2026-05-10
jmap_push.c: use correct state strings for StateChange object

ksmurchison:
- Mailbox, Calendar, and AddressBook states need to use the email, caldav, and carddav modseqs respectively rather than *foldersmodseq to match what a client receives in JMAP responses
- Email/Mailbox states need to have a "J" prefix
- Uses gperf array of JMAP Data Types

Should be reviewed commit-by-commit

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -111,6 +111,7 @@ CLEANFILES = \
     lib/imapopts.c \
     lib/imapopts.h \
     imap/css3_color_array.h \
+    imap/jmap_data_types.h \
     imap/rfc822_header.h \
     imap/mailbox_header_cache.h
 
@@ -360,6 +361,7 @@ EXTRA_DIST = \
     imap/http_err.et \
     imap/imap_err.et \
     imap/jmap_err.et \
+    imap/jmap_data_types.gperf \
     imap/lmtp_err.et \
     imap/mailbox_header_cache.gperf \
     imap/mupdate_err.et \
@@ -1308,6 +1310,7 @@ imap_httpd_LDADD = $(LD_SIEVE_ADD) $(LD_SERVER_ADD)
 
 if JMAP
 BUILT_SOURCES += \
+    imap/jmap_data_types.h \
     imap/jmap_err.c \
     imap/jmap_err.h
 
@@ -1430,6 +1433,7 @@ imap_libcyrus_jmap_props_la_SOURCES += \
 endif
 
 nodist_imap_httpd_SOURCES += \
+    imap/jmap_data_types.h \
     imap/jmap_err.c \
     imap/jmap_err.h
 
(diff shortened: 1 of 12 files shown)


Makefile.am
```diff
@@ -1378,6 +1381,7 @@ imap_httpd_SOURCES += \
     imap/jmap_calendar.h \
     imap/jmap_contact.c \
     imap/jmap_core.c \
+    imap/jmap_data_types.h \
```
elliefm: This imap_httpd_SOURCES section should list the gperf file (the authoritative source), not the generated header file

Makefile.am
```diff
@@ -360,6 +362,7 @@ EXTRA_DIST = \
     imap/http_err.et \
     imap/imap_err.et \
     imap/jmap_err.et \
+    imap/jmap_data_types.gperf \
```
elliefm, line 364: Once this is listed properly as a source (below), it won't need to be in EXTRA_DIST

elliefm (commented):
I haven't looked closely at the implementation details because I'm not familiar enough with jmap to notice anything being off.

rsto (commented):
Can you please describe in the commits and the pull request why these changes are required? I find it hard to review to the pull request without knowing what it aims to accomplish. What was wrong before that's now fixed?

As a side remark, it feels like we are cramming more and more logic into gperf files but they are barely documented.

Makefile.am
ksmurchison: Hmm.  I followed the same pattern as the other gperf files.

If I make the change as suggested, I get a linker error:

```
  CCLD     imap/httpd
/usr/bin/ld: imap/jmap_data_types.o: in function `jmap_data_types_lookup':
/home/ken/cyrus-imapd/imap/jmap_data_types.gperf:53: multiple definition of `jmap_data_types_lookup'; imap/jmap_api.o:/home/ken/cyrus-imapd/imap/jmap_data_types.gperf:53: first defined here
/usr/bin/ld: imap/jmap_data_types.o: in function `jmap_data_types_foreach':
/home/ken/cyrus-imapd/imap/jmap_data_types.gperf:62: multiple definition of `jmap_data_types_foreach'; imap/jmap_api.o:/home/ken/cyrus-imapd/imap/jmap_data_types.gperf:62: first defined here
collect2: error: ld returned 1 exit status
make: *** [Makefile:4954: imap/httpd] Error 1

```

rsto (approved):
Ken and I had an offline review where I had asked to not create a fake cstate, as I expect this to be hard to maintain in the future. The new implementation now uses a real cstate. Should we come to decide that opening and closing a cstate is too costly, we should rather refactor the state string generation functions to not require a `jmap_req_t`. 

I'm going to approve this now, on the expectation that CI passes and that ellie's review is met.

Makefile.am
ksmurchison: This happens because a `jmap_data_types.c` gets created by the `.gperf.c:` rule.  If you have a fix that properly builds the "generic" *.gperf files and the jmap_*_props.gperf files I'd love to use it.

ksmurchison: @elliefm Thoughts on the Makefile.am stuff or should I just merge as-is?

rjbs (approved):

Makefile.am
ksmurchison: @elliefm Thoughts on the Makefile.am stuff or should I just merge as-is?

PR #5958, opened 2026-05-11
sessionid: allow '.' in trace ids

Davis-A:
Some clients that talk to Cyrus (e.g. FastMTA) use '.' as a value inside their identifiers, and the existing base64url-without-padding alphabet rejected those ids outright. Add '.' to TRACE_ID_GOODCHARS, update the LMTP extension docs, and extend the cunit and cassandane tests to exercise a dotted id.

### DIFF
--- a/cassandane/tiny-tests/Delivery/traceid
+++ b/cassandane/tiny-tests/Delivery/traceid
@@ -3,7 +3,7 @@ use Cassandane::Tiny;
 
 sub test_traceid ($self)
 {
-    my $good_traceid = "01234567890abcdefghijklmnopqrstuvwxyz";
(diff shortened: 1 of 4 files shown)


ksmurchison (commented):
LGTM.  @elliefm Was there a reason why '.' was excluded?

elliefm:
The original set was the base64jmap set, and dot is not part of that set.  I don't remember if we chose that set for a specific reason, or if was just an arbitrary starting point that we expected to refine later.  I have no objection to adding it, but @rjbs should weigh in.

elliefm (approved):
Looks good to me, but I don't recall if there was a special reason we originally excluded dot.

elliefm:
Spoke to Rik, he has no objection to adding dot.

PR #5959, opened 2026-05-11
jmap_contact.c: add Card `vCardVersion` property for debugging

rsto:
This adds the `cyrusimap.org:vCardVersion` debugging property to ContactCard/get. Its value is the version of the vCard from which the ContactCard got converted from. This property is server-set and immutable.

Requesting this property requires the JMAP request to use the extension `https://cyrusimap.org/ns/jmap/debug`, otherwise the /get method is rejected as requesting an invalid property. A client must explicitly for this property in the `properties` argument.

### DIFF
--- a/cassandane/tiny-tests/JMAPContacts/card_get_vcardversion
+++ b/cassandane/tiny-tests/JMAPContacts/card_get_vcardversion
@@ -1,0 +1,80 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_card_get_vcardversion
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):
You will probably want to rebase this as I've just recently merged the PR which removes the legacy contacts code.

rsto:
> You will probably want to rebase this as I've just recently merged the PR which removes the legacy contacts code.

Done, thanks. Now rerunning CI.

PR #5961, opened 2026-05-11
Sync no whack users

brong:
Right now, `sync_client -O -u` will wipe out new users with the same name as the source folders if a user was renamed, which is BAD.  This fixes that bug.

### DIFF
--- a/cassandane/tiny-tests/Replication/rename_create_resync
+++ b/cassandane/tiny-tests/Replication/rename_create_resync
@@ -1,0 +1,85 @@
+#!perl
+use Cassandane::Tiny;
+
+#
+# Test: user "victim" renamed to "mover" on a second server (simulating a
+# user move), while a new user "victim" is created on the first server and
+# synced to the replica.  Syncing "mover" from the second server with -O
+# (no_copyback) must NOT wipe "victim"'s inbox on the replica.
+#
+# Bug: sync_client -u mover walks the name_history of mover's mailbox
+# (which includes "victim"'s old mailbox name), queries the replica for
+# that name, finds the new victim's inbox there, and since that UUID is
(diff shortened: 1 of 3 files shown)


rjbs:
> Tests the scenario where user A is renamed to user B on a second server (simulating a user move), while a new user A is created on the first server and synced to the replica.  Syncing B from the second server with -O (no_copyback) must NOT wipe user A's inbox on the replica.

"on a second server" is not very clarifying.  I think this is:

* on server 1: user named A exists
* on server 1: user named A is renamed to B
* on server 2: user named A is created
* user B is synced from server 1 to server 2

This **must not delete server 2's user named A**.

Maybe the first commit message could be clearer, maybe it just didn't hit the sweet spot for me.  But: this is how I'm reading the problem report.

rjbs (approved):
I think the commit/comment could be a bit clearer, but it's not a blocker.

wolfsage (approved):

brong:
> > Tests the scenario where user A is renamed to user B on a second server (simulating a user move), while a new user A is created on the first server and synced to the replica.  Syncing B from the second server with -O (no_copyback) must NOT wipe user A's inbox on the replica.
> 
> "on a second server" is not very clarifying. I think this is:
> 
> * on server 1: user named A exists
> * on server 1: user named A is renamed to B
> * on server 2: user named A is created
> * user B is synced from server 1 to server 2
> 
> This **must not delete server 2's user named A**.
> 
> Maybe the first commit message could be clearer, maybe it just didn't hit the sweet spot for me. But: this is how I'm reading the problem report.

I guess that's the shortest version!  My setup was server A replicates to a target, server B replicates the same target.  Server B has the rename on it.  But you can simplify it to what you described.

PR #5962, opened 2026-05-11
JMAPCalendars: test Calendar/set with create + onSuccessSetIsDefault …

rjbs:
…when no calendar exists

Reproduces a crash where Calendar/set crashes when creating the very first calendar in an account and using onSuccessSetIsDefault in the same call.


ksmurchison:
Superseded by #5963

PR #5964, opened 2026-05-12
itip_support.c: select calendar with lowest cmodseq as default

ksmurchison:
When a user had no calendar marked as CALDAV:schedule-default-calendar,
    we used to simply select the first calendar writable by the user as isDefault.
    However, if a new calendar was created with a name that appeared earlier
    in the ASCII sort order, isDefault would migrate to the new calendar.
    
    This PR searches all writable calendars for the one with the lowest
    createdmodseq, which will always yield consistent results.

### DIFF
--- a/imap/itip_support.c
+++ b/imap/itip_support.c
@@ -110,6 +110,7 @@ HIDDEN int organizer_changed(icalcomponent *oldcomp, icalcomponent *newcomp)
 
 struct pick_scheddefault_rock {
     strarray_t ignore;
+    modseq_t createdmodseq;
     char *collname;
 };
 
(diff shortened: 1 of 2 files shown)


imap/itip_support.c
```diff
@@ -132,9 +133,11 @@ static int pick_scheddefault_cb(const mbentry_t *mbentry, void *vrock)
                     if (!strcmp(userid, ownerid) || !strcmp(userid, "anyone")) {
                         int rights = 0;
                         cyrus_acl_strtomask(strarray_nth(acl, i+1), &rights);
-                        if (rights & ACL_INSERT) {
+                        if ((rights & ACL_INSERT) &&
+                            mbentry->createdmodseq < rock->createdmodseq) {
+                            rock->createdmodseq = mbentry->createdmodseq;
+                            free(rock->collname);
```
rjbs, line 139: I saw the ACL check and I thought, "Wait, is this iterating over shared calendars?  If your access to the calendar changes will that _again_ create instability?"  Looking into it, I think that actually this callback is only called from caldav_scheddefault, where it's used in a mboxlist_mboxtree iterator looking under the user's home set, where "surely" they always have insert, so I'm not gonna worry about it.  (Also, I expect this whole thing becomes less interesting when you can't delete default anymore.)

Let me know if I've misunderstood!

rjbs (approved):

PR #5965, opened 2026-05-12
http_dav.h: Use 'i' right rather than 'p' for adding a DAV resource

ksmurchison:
This syncs CalDAV behavior with with JMAP.
JMAP only uses the 'i' for adding/updating a resource, but CalDAV used 'p' for add and 'i' for update.
When sharing a calendar as read-write via JMAP, only the 'i' right was granted to the sharee, so a CalDAV MOVE of an event into the shared calendar would fail with
DAV:need-privileges (CY:add-resource).
With this PR, now DAV only requires 'i' for add/update.

### DIFF
--- a/docsrc/download/installation/http/caldav.rst
+++ b/docsrc/download/installation/http/caldav.rst
@@ -84,129 +84,159 @@ The tables below show how the access controls are used by the CalDAV module.
       <caption>Mapping of IMAP Rights to WebDAV Privileges & HTTP Methods</caption>
       <tr>
         <th>IMAP rights</th>
-        <th colspan=2>WebDAV privileges</th>
+        <th colspan=3>WebDAV privileges</th>
         <th>HTTP methods</th>
       </tr>
       <tr>
-        <td>l - lookup
+        <td style="vertical-align:middle">l - lookup
           <br>r - read</td>
-        <td>DAV:read</td>
-        <td>DAV:read-current-user-privilege-set
-          <br>CALDAV:read-free-busy</td>
+        <td colspan=2 rowspan=2 style="vertical-align:middle">
+          <a href="https://datatracker.ietf.org/doc/html/rfc3744#section-3.1">DAV:read</a></td>
+        <td style="vertical-align:middle">
+          <a href="https://datatracker.ietf.org/doc/html/rfc3744#section-3.7">DAV:read-current-user-privilege-set</a></td>
         <td>GET/HEAD
-          <br>COPY/MOVE <small>(on source)</small>
-            <br>PROPFIND
-              <br>REPORT</td>
+          <br>COPY/MOVE
+            <br><small>(on source)</small>
+              <br>PROPFIND
+                <br>REPORT</td>
       </tr>
       <tr>
-        <td><s>s - seen</s></td>
-        <td colspan=2/>
+        <td style="vertical-align:middle">9 - freebusy
+          <br><small>(regular calendar collections ONLY)</small></td>
+        <td style="vertical-align:middle">
+          <a href="https://datatracker.ietf.org/doc/html/rfc4791#section-6.1.1">CALDAV:read-free-busy</a></td>
+        <td>GET/HEAD
+          <br><small>(<a href="#freebusy-url">Freebusy URLs</a> ONLY)</small>
+            <br>REPORT
+              <br><small>(<a href="https://datatracker.ietf.org/doc/html/rfc4791#section-7.10">CALDAV:free-busy-query</a> ONLY)</small></td>
+      </tr>
+      <tr>
+        <td><s>s - seen</s>
+          <br><s>p - post</s></td>
+        <td colspan=4/>
         <td/>
       </tr>
       <tr>
-        <td>w - write
-          <br>n - write shared annotation</td>
-        <td colspan=2>DAV:write-properties</td>
-        <td>PROPPATCH
-          <br>COPY/MOVE <small>(on destination)</small></td>
+        <td style="vertical-align:middle">w - write</td>
+        <td rowspan=6 style="vertical-align:middle">
+          <a href="https://datatracker.ietf.org/doc/html/rfc3744#section-3.2">DAV:write</a></td>
+        <td rowspan=2 style="vertical-align:middle">
+          <a href="https://datatracker.ietf.org/doc/html/rfc3744#section-3.3">DAV:write-properties</a></td>
+        <td>CY:write-properties-collection</td>
+        <td rowspan=2>PROPPATCH
+          <br>COPY/MOVE
+            <br><small>(on destination)</small></td>
       </tr>
       <tr>
-        <td>i - insert</td>
-        <td colspan=2>DAV:write-content</td>
-        <td>PUT
-          <br>PATCH
-            <br>COPY/MOVE <small>(on destination resource)</small>
-              <br>LOCK
-                <br>UNLOCK <small>(lock owner ONLY)</small></td>
+        <td>n - write shared annotation</td>
+        <td>CY:write-properties-resource</td>
       </tr>
       <tr>
-        <td>p - post</td>
-        <td rowspan=2>DAV:bind</td>
-        <td>CYRUS:add-resource</td>
-        <td>POST</td>
+        <td style="vertical-align:middle">i - insert</td>
+        <td rowspan=2 style="vertical-align:middle">
+          <a href="https://datatracker.ietf.org/doc/html/rfc3744#section-3.9">DAV:bind</a></td>
+        <td style="vertical-align:middle">
+          <a href="https://datatracker.ietf.org/doc/html/rfc3744#section-3.4">DAV:write-content</a></td>
+        <td>POST
+          <br><small>(<a href="https://datatracker.ietf.org/doc/html/rfc5995#section-3.1">Add Member URI</a>)</small>
+            <br>PUT
+              <br>PATCH
+                <br>COPY/MOVE
+                  <br><small>(on destination resource)</small>
+                    <br>LOCK
+                      <br>UNLOCK
+                        <br><small>(lock owner ONLY)</small></td>
       </tr>
       <tr>
-        <td>k - create mailbox</td>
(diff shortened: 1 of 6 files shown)


dilyanpalauzov:
Does the documentation for the privileges for CalDAV under https://www.cyrusimap.org/dev/download/installation/http/caldav.html#calendar-access-controls and under https://www.cyrusimap.org/dev/reference/extensions/dav-extensions.html#privileges need to be updated with the current change?

dilyanpalauzov:
Are JMAP privileges documented, as are CalDAV privileges described on the two link above?

docsrc/download/installation/http/caldav.rst
```diff
@@ -123,7 +123,7 @@ The tables below show how the access controls are used by the CalDAV module.
         <td>p - post</td>
         <td rowspan=2>DAV:bind</td>
         <td>CYRUS:add-resource</td>
-        <td>POST</td>
+        <td>INSERT</td>
```
dilyanpalauzov: This would change on the line `p - post` the last column “HTTP method” from `POST` to `INSERT`.  However the HTTP verb `INSERT` is … not common.

dilyanpalauzov:
In the table “Calendar access controls” at https://www.cyrusimap.org/dev/download/installation/http/caldav.html#calendar-access-controls the first column is "UNIQUE” - each IMAP right appears once.  With the current change, `i` will appear twice in the first column.  For `s - seen` it is suggested, that it does nothing, as it is struck out.  For `p` privilege there is nothing said anymore.  Nevertheless in the next table, labelled “Default WebDAV Privileges by Collection”, the `p` right is granted by default.  This bit now does nothing, like the `s` ACL bit, while by default the `s` bit is not set, but the `p` bit is set.

Moreover my understanding is that the `DAV:bind` privilege, before this change, is aggregated - it is granted when both `p`⇔ CYRUS:add-resource and `k` ⇔CYRUS:make-collection privileges are granted.  After the current change, the need for the distinct `CYRUS:add-resource` privilege disappears, as it is identical to the `DAV:write-content` privilege, bound to the `i` ACL bit.  So the `CYRUS:add-resource` privilege should be deleted (substituted with `DAV:write-content`). This keeps the `DAV:bind` privilege aggregate, but it is now composition of the `DAV:write-content` plus the `CYRUS:make-collection` privilege.

Isn’t it simpler to adapt the JMAP ACL model (`p`/`i` bit), to follow the one for WebDAV?

In any case, to understand these nuances, and simplify the comparison, I think it is reasonable, if somebody creates similar table for ACLs in JMAP, as it exists for CalDAV.

rjbs (approved):

docsrc/download/installation/http/caldav.rst
ksmurchison: I removed CY:add-resource entirely and reworked the table so that DAV:write-content is aggregated under DAV:bind

imap/http_dav.c
```diff
@@ -7137,10 +7124,9 @@ int meth_put(struct transaction_t *txn, void *params)
     struct buf msg_buf = BUF_INITIALIZER;
     user_nslock_t *user_nslock = NULL;
 
-    if (txn->meth == METH_POST) {
-        reqd_rights = DACL_ADDRSRC;
-    }
-    else {
+    reqd_rights = DACL_WRITECONT;
```
dilyanpalauzov: This assignment could be done 12 lines higher at `int ret, r, precond, rights, reqd_rights;`.

docsrc/download/installation/http/caldav.rst
```diff
@@ -84,128 +84,141 @@ The tables below show how the access controls are used by the CalDAV module.
       <caption>Mapping of IMAP Rights to WebDAV Privileges & HTTP Methods</caption>
       <tr>
         <th>IMAP rights</th>
-        <th colspan=2>WebDAV privileges</th>
+        <th colspan=3>WebDAV privileges</th>
         <th>HTTP methods</th>
       </tr>
       <tr>
-        <td>l - lookup
+        <td style="vertical-align:middle;">l - lookup
```
dilyanpalauzov: The semi-colon does not harm, but is also not necessary.

dilyanpalauzov (commented):
I realize now, that before this change the `i - insert` ⇔ `DAV:write-content` privilege mentioned method `PUT`, without clarifying that PUT could be used here only for changing existing file.  Likewise the `p - post` ⇔ `CYRUS:add-resource` privilege mentioned `POST`, implying file creation over RFC 5995, but it has not stated, that the privilege applies also for the `PUT` verb, when used to upload a new file.  This is shown here more correctly: `PUT` means both file modification and file creation.


caldav_util.c:caldav_create_defaultcalendars() invokes:
```c
// acl.h
#define ACL_ALL         (ACL_LOOKUP|ACL_READ|ACL_SETSEEN|ACL_WRITE\
                        |ACL_INSERT|ACL_POST|ACL_CREATE|ACL_DELETEMBOX\
                        |ACL_DELETEMSG|ACL_EXPUNGE|ACL_ADMIN\
                        |ACL_ANNOTATEMSG) 



r = _create_mailbox(userid, mailboxname, MBTYPE_CALENDAR, comp_types,
                    ACL_ALL | DACL_READFB, DACL_READFB,
                    config_getstring(IMAPOPT_CALENDAR_DEFAULT_DISPLAYNAME), namespace, authstate, 1, &user_nslock);

r = _create_mailbox(userid, mailboxname, MBTYPE_CALENDAR, 0,
                    ACL_ALL | DACL_SCHED, DACL_SCHED, NULL,
                    namespace, authstate, 0, &user_nslock);

r = _create_mailbox(userid, mailboxname, MBTYPE_CALENDAR, 0,
                    ACL_ALL | DACL_SCHED, 0, NULL,
                    namespace, authstate, 0, &user_nslock);

r = _create_mailbox(userid, mailboxname, MBTYPE_COLLECTION, 0,
                    ACL_ALL, ACL_READ, NULL,
                    namespace, authstate, 0, &user_nslock);
```
When I look on standard calendars, created long time ago, I see they have the `s` ACL bit.  So, just by reading the source code (I have not tried it), I assume that `caldav_create_defaultcalendars()` does set `ACL_SETSEEN` ⇔ `s` and `ACL_POST` ⇔ `p` ACL bits on newly created Default, Inbox, Outbox, Attachments mailboxes, even if these ACL bits have no added value.  Currently, the behaviour seems different from what is documented.

docsrc/reference/extensions/dav-extensions.md
```diff
@@ -70,7 +70,7 @@ in different IMAP constructs with different ACL bits.
 | Privilege            | Description                        | IMAP ACL bit |
 |----------------------|------------------------------------|--------------|
 | `CY:make-collection` | Create a new sub-collection        | k            |
-| `CY:add-resource`    | Add a new resource to a collection | p            |
+| `DAV:write-content`  | Add a new resource to a collection | i            |
```
dilyanpalauzov: `DAV:write-content` is not Cyrus-specific extension, so it can be removed from this file.

rsto (approved):

docsrc/reference/extensions/dav-extensions.md
rjbs: I agree.  This line should go.  Not a blocker, we know the DAV docs need to be streamlined and unified.  But easy pickins.

rjbs (approved):

wolfsage:
Looks like we have no tests for some of these changes?
```
 <D:propfind xmlns:D="DAV:" xmlns:A="http://apple.com/ns/ical/">
  <D:allprop/>
  <D:prop>
    <D:supported-privilege-set/>
  </D:prop>
 </D:propfind>

```

What I'm not sure of is if these changes are right (or if it even matters). 

Does https://www.rfc-editor.org/rfc/rfc3744.html define this stuff? Is it current?

It says things like:
```
   | PUT (target exists)             | <D:write-content> on target resource
   | PUT (no target exists)          | <D:bind> on parent collection of target |
```
So D:write-content to update something, D:bind to create something new

Then this section is unclear to me too, about aggrevation: https://www.rfc-editor.org/info/rfc3744/#section-3.12

```
  DAV:write MUST contain DAV:bind, DAV:unbind, DAV:write-properties and
   DAV:write-content.
```

however we've moved write-content under bind?
```
<D:supported-privilege>
                <D:privilege>
                  <D:bind/>
                </D:privilege>
                <D:description xml:lang="en">Add new member to collection</D:description>
                <D:supported-privilege>
                  <D:privilege>
                    <CY:make-collection/>
                  </D:privilege>
                  <D:description xml:lang="en">Make new collection</D:description>
                </D:supported-privilege>
                <D:supported-privilege>
                  <D:privilege>
                    <D:write-content/>
                  </D:privilege>
                  <D:description xml:lang="en">Write resource content</D:description>
                </D:supported-privilege>
              </D:supported-privilege>

```

however before this patch it was under write:
```
            <D:supported-privilege>
              <D:privilege>
                <D:write/>
              </D:privilege>
              <D:description xml:lang="en">Write any object</D:description>
              <D:supported-privilege>
                <D:privilege>
                  <D:write-content/>
                </D:privilege>
                <D:description xml:lang="en">Write resource content</D:description>
              </D:supported-privilege>

```

dilyanpalauzov:
RFC 3744 defines this stuff and it is current.

> Then this section is unclear to me too, about aggregation: https://www.rfc-editor.org/info/rfc3744/#section-3.12
>
>>  DAV:write MUST contain DAV:bind, DAV:unbind, DAV:write-properties and DAV:write-content.
>
> however we've moved write-content under bind?

Section 3.12 starts with

> Server implementations are free to aggregate the predefined privileges (defined above in Sections [3.1](https://www.rfc-editor.org/info/rfc3744/#section-3.1)-[3.10](https://www.rfc-editor.org/info/rfc3744/#section-3.10)) subject to the following limitations: …

As can be seen below, `DAV:write` contains directly `DAV:bind`, `DAV:unbind` and `DAV:write-properties` and indirectly `DAV:write-content`.  I see no contradiction: If the `DAV:write-content` privilege is not granted, then users have no `DAV:write` privilege.

image

PR #5969, opened 2026-05-13
imapoptions: use the {configdirectory} mechanism where feasible

elliefm:
There's a few imapd.conf options that specify a default value of `NULL` in the imapoptions data, but then compute a non-NULL default value based on `configdirectory` at runtime.  The automated documentation based on the data files doesn't match reality, so these options then need extra verbiage explaining the real behaviour.

We already have a mechanism for deriving config values from the `configdirectory` value, so use it for these options' defaults.  This makes the automatic documentation correct.

I've also fixed some inconsistencies in how cunit handled configdirectory, which made fixing the real problem easier.

### DIFF
--- a/cunit/aaa-db.testc
+++ b/cunit/aaa-db.testc
@@ -1910,8 +1910,8 @@ static int set_up(void)
     if (!basedir)
         return -1;
 
-    snprintf(buf, sizeof(buf), "configdirectory: %s/conf\n", basedir);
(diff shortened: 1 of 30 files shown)


ksmurchison (approved):
LGTM

rsto (approved):

rjbs (approved):

PR #5988, opened 2026-05-15
Sync no whack users

brong:
Here's some more username protections - since one user got hit AGAIN after the first fix.  I wasn't able to recreate the exact cause with tests, however I'm happy that these changes are safe and better than before, particularly the "avoid renaming into the other user" part of the intermediate discovery logic.

### DIFF
--- a/cassandane/tiny-tests/Replication/rename_move_resync
+++ b/cassandane/tiny-tests/Replication/rename_move_resync
@@ -1,0 +1,85 @@
+#!perl
+use Cassandane::Tiny;
+
+#
+# Invariant: syncing one user must never delete another user's mailboxes,
+# even when that user's name_history references the other user's namespace
+# (e.g. mover was once renamed from/to victim) and the sync runs with -O
+# (no_copyback).
+#
+sub test_rename_move_resync
+    :NoAltNameSpace :AllowMoves :Replication :DelayedDelete :Conversations
+    ($self)
+{
+    my $admintalk = $self->{adminstore}->get_client();
+    my $mastersvc = $self->{instance}->get_service('imap');
+    my $replicasvc = $self->{replica}->get_service('imap');
+
+    my $mover_store = $mastersvc->create_store(username => 'mover');
+    $mover_store->set_fetch_attributes('uid');
+
+    xlog $self, "OldStore: create user.mover and give it some data";
+    $admintalk->create("user.mover");
+    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());
+
+    $admintalk->setacl("user.mover", "admin", "lrswipkxteacdn");
+    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());
+
+    $admintalk->create("user.mover.SeparateFolder");
+    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());
+
+    $admintalk->create("user.mover.Trash");
+    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());
+
+    $self->make_message("Message A", store => $mover_store);
+    $self->make_message("Message B", store => $mover_store);
+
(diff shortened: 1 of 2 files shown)


cassandane/tiny-tests/Replication/rename_move_resync
```diff
@@ -0,0 +1,87 @@
...
+# containing user.victim).  When mover is subsequently synced to the
+# replica with -O (no_copyback), mover's name_history causes the replica's
+# user.victim mailboxes to appear as unknown remote folders, and the
+# NO_COPYBACK path deletes them -- wiping the already-moved victim's email.
+#
+# Bug: do_mailbox_info adds name_history entries to mboxname_list without
+# filtering by userid.  "user.victim" from mover's name_history is added,
+# the replica's user.victim is found there but has no local counterpart
+# (victim was deleted from master in the move), and the NO_COPYBACK path
+# sends UNMAILBOX for user.victim -- wiping victim's inbox on the replica.
```
rjbs: We should standardize a bit what the start-of-test comment is.  Claude likes to write a very detailed example of how a bug happened, but over time that's less useful than "this is the invariant we are testing for" (and then in the commit message "we added this because of a bug").  But this isn't a blocker, just a musing that we should come together on one practice soon!

rjbs (approved):

imap/sync_support.c
wolfsage, line 6968: What about this branch? How would we get here, and does this also need username protection?

Ah. I guess this is covered below when we process rename_folders which covers both here and earlier on 6935? Okay.

cassandane/tiny-tests/Replication/rename_move_resync
```diff
@@ -0,0 +1,87 @@
...
+    $admintalk->create("user.mover");
+    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());
+    $admintalk->setacl("user.mover", "admin", "lrswipkxteacdn");
+    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());
+    $admintalk->create("user.mover.SeparateFolder");
+    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());
+    $admintalk->create("user.mover.Trash");
+    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());
+    $self->make_message("Message A", store => $mover_store);
+    $self->make_message("Message B", store => $mover_store);
```
wolfsage, line 35: This is all fine but newlines between blocks of related code would make this way more readable:
```
    xlog $self, "OldStore: create user.mover and give it some data";
    $admintalk->create("user.mover");
    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());

    $admintalk->setacl("user.mover", "admin", "lrswipkxteacdn");
    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());

    $admintalk->create("user.mover.SeparateFolder");
    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());

    $admintalk->create("user.mover.Trash");
    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());

    $self->make_message("Message A", store => $mover_store);
    $self->make_message("Message B", store => $mover_store);
```

wolfsage: (Here and the rest of the file)

wolfsage (approved):

cassandane/tiny-tests/Replication/rename_move_resync
brong: Yeah I asked Claude told Claude that over time this turns a codebase into an unreadable history lesson and to tidy it up and remember that!  It's lots shorter now.

cassandane/tiny-tests/Replication/rename_move_resync
brong: Done.

PR #5989, opened 2026-05-16
Check on tmpfs

brong:
dar build on my Mac fails in the check stage because the MacOS virtual filesystem isn't good for running DB files on.

There's also one tiny bug fix which I picked out into its own commit.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -739,11 +739,15 @@ check-local:
 		test -z "$$VG" || vg="$(VALGRIND)" ; \
 		f="-v" ; \
 		test "x$$CUFORMAT" = xjunit && f="-x" ; \
-		cd cunit ; \
-		$$vg ./unit $$f ; \
+		tmpdir=$$(mktemp -d "$${TMPDIR:-/tmp}/cyrus-cunit-XXXXXX") || exit 1 ; \
+		trap "rm -rf $$tmpdir" EXIT ; \
+		cd $$tmpdir ; \
+		cp $(abs_srcdir)/cunit/*.pem ./ ; \
+		$$vg $(abs_builddir)/cunit/unit $$f ; \
 		retval=$$? ; \
 		if [ "x$$CUFORMAT" = xjunit ] ; then \
-			$(RM) -rf reports ; mkdir reports ; $(PERL) ./cunit-to-junit.pl ; \
+			$(RM) -rf reports ; mkdir reports ; $(PERL) $(abs_srcdir)/cunit/cunit-to-junit.pl ; \
+			cp -r reports $(abs_builddir)/cunit/ ; \
 		fi ; \
 		exit $$retval
 
@@ -759,18 +763,19 @@ check-discrete: cunit/unit
 	@echo "Running unit tests, one at a time"
 	@vg= ; \
 		test -z "$$VG" || vg="$(VALGRIND) --quiet --log-fd=3" ; \
-		cd cunit ; \
 		fails=0 ; \
 		tests=0 ; \
 		while read t ; do \
 			tests=$$(($$tests + 1)) ; \
-			if $$vg ./unit $$t >/dev/null 3>&2 2>/dev/null ; then \
+			tmpdir=$$(mktemp -d "$${TMPDIR:-/tmp}/cyrus-cunit-XXXXXX") ; \
+			if (cd $$tmpdir && cp $(abs_srcdir)/cunit/*.pem ./ && $$vg $(abs_builddir)/cunit/unit $$t >/dev/null 3>&2 2>/dev/null) ; then \
 				echo "[  OK  ] $$t" ; \
 			else \
 				fails=$$(($$fails + 1)) ; \
 				echo "[FAILED] $$t" ; \
 			fi ; \
-		done < <( ./unit -l ) ; \
+			rm -rf $$tmpdir ; \
+		done < <( $(abs_builddir)/cunit/unit -l ) ; \
 		echo "$$fails/$$tests tests failed" ; \
 		exit $$fails
 
@@ -783,10 +788,12 @@ check-%: cunit/unit
 	@echo "Running unit tests for $*"
 	@vg= ; \
 		test -z "$$VG" || vg="$(VALGRIND)" ; \
-		cd cunit ; \
-		$$vg ./unit -v $* ; \
-		retval=$$? ; \
-		exit $$retval
+		tmpdir=$$(mktemp -d "$${TMPDIR:-/tmp}/cyrus-cunit-XXXXXX") || exit 1 ; \
+		trap "rm -rf $$tmpdir" EXIT ; \
+		cd $$tmpdir ; \
+		cp $(abs_srcdir)/cunit/*.pem ./ ; \
+		$$vg $(abs_builddir)/cunit/unit -v $* ; \
+		exit $$?
 
 endif
 
--- a/cunit/backend.testc
+++ b/cunit/backend.testc
@@ -1821,7 +1821,7 @@ static int set_up(void)
     buf_printf(&myconfig, "configdirectory: %s/conf\n", DBDIR);
 
     /* certs are in current directory */
-    buf_printf(&myconfig, "tls_server_ca_file: %s/cacert.pm\n", cwd);
+    buf_printf(&myconfig, "tls_server_ca_file: %s/cacert.pem\n", cwd);
     buf_printf(&myconfig, "tls_server_cert: %s/cert.pem\n", cwd);
     buf_printf(&myconfig, "tls_server_key: %s/key.pem\n", cwd);
 
--- a/cunit/pcunit
+++ b/cunit/pcunit
@@ -15,7 +15,9 @@ use warnings;
 
 use Time::HiRes qw(gettimeofday);
 use POSIX ":sys_wait_h";
-use File::Temp qw(tempfile);
+use File::Temp qw(tempfile tempdir);
+use File::Basename qw(dirname);
+use Cwd qw(abs_path);
 
 my $jobs = shift;
 if (defined $jobs) {
@@ -27,7 +29,8 @@ if (defined $jobs) {
     $jobs = 8;
 }
 
-chdir "cunit";
+my $srcdir = abs_path(dirname($0));
+my $unit = "$srcdir/unit";
 
 # Convert a list of tests into a list of just the suites that contain them
 # so we can run each suite in its own process.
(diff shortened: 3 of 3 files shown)


Makefile.am
```diff
@@ -737,11 +737,14 @@ check-local:
...
-		$$vg ./unit $$f ; \
+		tmpdir=$$(mktemp -d /tmp/cyrus-cunit-XXXXXX) || exit 1 ; \
+		trap "rm -rf $$tmpdir" EXIT ; \
+		cd $$tmpdir ; \
+		$$vg $(abs_builddir)/cunit/unit $$f ; \
 		retval=$$? ; \
 		if [ "x$$CUFORMAT" = xjunit ] ; then \
-			$(RM) -rf reports ; mkdir reports ; $(PERL) ./cunit-to-junit.pl ; \
+			$(RM) -rf reports ; mkdir reports ; $(PERL) $(abs_builddir)/cunit/cunit-to-junit.pl ; \
+			cp -r reports $(abs_builddir)/cunit/ ; \
```
elliefm, line 750: It looks like this now runs from `/tmp/cyrus-cunit-whatever`, but unlike `pcunit`, it hasn't copied the cunit/*.pem files across.   I think this is what's making the backend TLS tests fail -- the test suite is configuring cyrus to look for certs in "./foo.pem", but they aren't there.  Same for the `check-%` and `check-discrete` targets.

cunit/backend.testc
```diff
@@ -1803,27 +1803,24 @@ static int sasl_initialised;
...
         fprintf(stderr, "Starting server!\n");
 
-    sr = getcwd(cwd, sizeof(cwd));
-    if (!sr) {
-        fprintf(stderr, "getcwd() failed: %s\n", strerror(errno));
-        return -1;
-    }
+    /* certs live in the cunit source dir; prefer explicit env over cwd */
+    certdir = getenv("CUNIT_SRC_DIR");
+    if (!certdir) certdir = ".";
```
elliefm: I think it'd better to keep using `getcwd()` for the fallback case, rather than hardcoding `"."`.  We're going to use this string to set an imapd.conf option, and relative paths in imapd.conf smell really nasty, even from within a test case.

Something like this ought to work (and be neater than the old mess) if you also bring back `char cwd[PATH_MAX];` at the top of the function (but we don't need `sr`).

```suggestion
    if (!certdir) {
        if (!getcwd(cwd, sizeof(cwd))) {
            perror("getcwd");
            return -1;
        }
        certdir = cwd;
    }
```

elliefm (commented):
The backend TLS tests are failing when run from `make check` et al, looks like the pem files aren't being copied to the tmpdir?

Makefile.am
elliefm: Why on earth did it attach this comment to these lines?  This comment is supposed to apply to the change a few lines above, where we set up tmpdir and then run `cunit/unit`.

elliefm:
Oh I see.  CI uses `cyd build`, which uses `make check-fast` (i.e. `pcunit`) by default.  The only CI job that actually runs the regular `check` rules is the valgrind one.  That's why it looks like it fails under valgrind specifically, when actually it always fails except for `check-fast`.

Makefile.am
```diff
@@ -737,11 +737,14 @@ check-local:
...
-		cd cunit ; \
-		$$vg ./unit $$f ; \
+		tmpdir=$$(mktemp -d /tmp/cyrus-cunit-XXXXXX) || exit 1 ; \
+		trap "rm -rf $$tmpdir" EXIT ; \
+		cd $$tmpdir ; \
+		$$vg $(abs_builddir)/cunit/unit $$f ; \
 		retval=$$? ; \
 		if [ "x$$CUFORMAT" = xjunit ] ; then \
-			$(RM) -rf reports ; mkdir reports ; $(PERL) ./cunit-to-junit.pl ; \
+			$(RM) -rf reports ; mkdir reports ; $(PERL) $(abs_builddir)/cunit/cunit-to-junit.pl ; \
```
elliefm: This isn't breaking anything that wasn't already broken, but cunit-to-junit.pl isn't "built", so the only way it's in the build dir is if the build dir is the same as the source dir.  Better to refer to it by where it actually is.

```suggestion
			$(RM) -rf reports ; mkdir reports ; $(PERL) $(abs_srcdir)/cunit/cunit-to-junit.pl ; \
```

Makefile.am
```diff
@@ -737,11 +737,14 @@ check-local:
 		test -z "$$VG" || vg="$(VALGRIND)" ; \
 		f="-v" ; \
 		test "x$$CUFORMAT" = xjunit && f="-x" ; \
-		cd cunit ; \
-		$$vg ./unit $$f ; \
+		tmpdir=$$(mktemp -d /tmp/cyrus-cunit-XXXXXX) || exit 1 ; \
+		trap "rm -rf $$tmpdir" EXIT ; \
+		cd $$tmpdir ; \
+		$$vg $(abs_builddir)/cunit/unit $$f ; \
```
elliefm: Something like this seems to fix the issues with `make check`, `make check-discrete`, and `make check-backend`:

```
diff --git a/Makefile.am b/Makefile.am
index 149b7ebaf..7085eba0f 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -740,6 +740,7 @@ check-local:
 		tmpdir=$$(mktemp -d /tmp/cyrus-cunit-XXXXXX) || exit 1 ; \
 		trap "rm -rf $$tmpdir" EXIT ; \
 		cd $$tmpdir ; \
+		cp $(abs_srcdir)/cunit/*.pem ./ ; \
 		$$vg $(abs_builddir)/cunit/unit $$f ; \
 		retval=$$? ; \
 		if [ "x$$CUFORMAT" = xjunit ] ; then \
@@ -765,7 +766,7 @@ check-discrete: cunit/unit
 		while read t ; do \
 			tests=$$(($$tests + 1)) ; \
 			tmpdir=$$(mktemp -d /tmp/cyrus-cunit-XXXXXX) ; \
-			if (cd $$tmpdir && $$vg $(abs_builddir)/cunit/unit $$t >/dev/null 3>&2 2>/dev/null) ; then \
+			if (cd $$tmpdir && cp $(abs_srcdir)/cunit/*.pem ./ && $$vg $(abs_builddir)/cunit/unit $$t >/dev/null 3>&2 2>/dev/null) ; then \
 				echo "[  OK  ] $$t" ; \
 			else \
 				fails=$$(($$fails + 1)) ; \
@@ -788,6 +789,7 @@ check-%: cunit/unit
 		tmpdir=$$(mktemp -d /tmp/cyrus-cunit-XXXXXX) || exit 1 ; \
 		trap "rm -rf $$tmpdir" EXIT ; \
 		cd $$tmpdir ; \
+		cp $(abs_srcdir)/cunit/*.pem ./ ; \
 		$$vg $(abs_builddir)/cunit/unit -v $* ; \
 		exit $$?
 
```

elliefm (changes requested):

cunit/backend.testc
```diff
@@ -1803,27 +1803,24 @@ static int sasl_initialised;
...
     if (verbose > 1)
         fprintf(stderr, "Starting server!\n");
 
-    sr = getcwd(cwd, sizeof(cwd));
-    if (!sr) {
-        fprintf(stderr, "getcwd() failed: %s\n", strerror(errno));
-        return -1;
-    }
+    /* certs live in the cunit source dir; prefer explicit env over cwd */
+    certdir = getenv("CUNIT_SRC_DIR");
```
elliefm: On further thought, I don't think this environment variable really gets us anything.  Like, the only thing that sets it is `pcunit`, but it immediately chdirs to the same directory anyway, so it would still work if the environment variable hadn't been set.

And then, why is it called "CUNIT_SRC_DIR" when it's actually some temporary directory with pem files copied into it?  "CUNIT_PEM_DIR" maybe, but I'm just not convinced we need it.  And if we do need it for some reason I'm missing, I don't like test infrastructure choices being buried in individual test suites.  The better implementation would be having `unit` accept some `--data-dir=/path` option, and a way for test suites to query that value so they can find their data files.

cunit/backend.testc
brong: Just removed the fallback case entirely, along with the ENV var

brong:
Thanks @elliefm and sorry for taking so long to get back to this, I simplified this back to the smallest set of changes and no configuration, it just always creates a directory undef /tmp/ now.

This still fixes my issues on MacOS; otherwise the tests fails reliably on annotate.

cunit/pcunit
```diff
@@ -43,7 +46,7 @@ chdir "cunit";
 #
 #     ('xsha1', 'xsyslog_ev')
 
-my $tests = `./unit -l`;
+my $tests = `cd /tmp && $unit -l`;
```
elliefm: Why are we changing to `/tmp` for the `unit -l` invocations?  This invocation just lists the available tests (which are compiled into the binary), there's no filesystem access, so CWD shouldn't matter at all.

Makefile.am
```diff
@@ -759,18 +763,19 @@ check-discrete: cunit/unit
...
+			tmpdir=$$(mktemp -d /tmp/cyrus-cunit-XXXXXX) ; \
+			if (cd $$tmpdir && cp $(abs_srcdir)/cunit/*.pem ./ && $$vg $(abs_builddir)/cunit/unit $$t >/dev/null 3>&2 2>/dev/null) ; then \
 				echo "[  OK  ] $$t" ; \
 			else \
 				fails=$$(($$fails + 1)) ; \
 				echo "[FAILED] $$t" ; \
 			fi ; \
-		done < <( ./unit -l ) ; \
+			rm -rf $$tmpdir ; \
+		done < <( cd /tmp && $(abs_builddir)/cunit/unit -l ) ; \
```
elliefm: Here's another case where we chdir to `/tmp` to run `unit -l`, unnecessarily I think

elliefm (approved):
This seems to work correctly now, thanks.  The "distcheck" CI job failed, but it looked unrelated, and succeeded on a re-run.

There's a couple of places where it chdirs to `/tmp` (n.b. not `$tmpdir`), seemingly unnecessarily -- perhaps leftovers from an earlier version?  If they can be tidied up they should be.  If the chdirs are load-bearing somehow, it'd be nice to see a comment explaining why, at least in pcunit.  If that's all you change, I don't need to review it again after, so I'm hitting approve now.

Makefile.am
brong: Yep, removed

brong:
@elliefm I removed the unnecessary chdirs, and ALSO added TMPDIR environment support so we don't use /tmp at all if it's set.

elliefm (approved):
Oh, nice touch, thanks!

PR #5995, opened 2026-05-18
thread trace id from http and lmtp into mbox events

rjbs:


ksmurchison (approved):
Approved provided CI passes

PR #5996, opened 2026-05-19
Fix typos in the `doc` and `docsrc` dirs and a few text files

kurtmckee:
The typos were identified using the `typos` CLI tool and were reviewed individually.


dilyanpalauzov:
The `existant` can be changed in all directories - cassandane/, cunit/quota.testc, imap/userdeny_db.c, lib/.

kurtmckee:
@dilyanpalauzov I think that I've addressed your feedback. I used a find-and-replace tool but confirmed each change.

Please let me know if there's anything else that needs to be addressed.

ksmurchison (approved):

PR #5999, opened 2026-05-19
docsrc/conf.py: make the left sidebar expand less

rjbs:
I often feel lost in the doc site, in part because it has *so much* navigation "help".  Rather than give me a deep, deep drilldown to every little subheading -- progressively disclosed so that I can't build up a place memory -- just stick to two levels.  A rebuild made the site feel much, much friendlier.

Lots of information architecture work left to do after this, though.

### DIFF
--- a/docsrc/conf.py
+++ b/docsrc/conf.py
@@ -193,8 +193,14 @@ html_context = {
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
-html_theme_options = {}
-
+html_theme_options = {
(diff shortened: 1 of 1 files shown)


elliefm (approved):
Oh yeah that's way better, the sidebar might actually be useful like this.

In the short term I think this might make navigation a little harder, but that's because our information architecture is lacking, not because the sidebar needs to list every fork and spoon.

With the sidebar thus constrained, I think it'll be easier to start to identify where we need separate or combined documents, by asking ourselves, "does this heading need to be in the sidebar, or not?" and arranging the documents accordingly.

rjbs:
@elliefm I'm glad we're so much on the same page about this!

PR #6000, opened 2026-05-19
jmap: report shareWith invalidProperty with the actual sharee id

rjbs:
Calendar/set and AddressBook/set were pushing the literal string "sharee" onto the invalidProperty path when rejecting an unknown right, producing e.g. "shareWith/sharee/mayBogus" regardless of which sharee the client named.  Use the sharee variable, matching what Mailbox/set already does.

Amusingly the tests didn't catch it (if we agree it was a mistake) because the sharing user was named "sharee" -- same as the literal string we were using!


ksmurchison (approved):

rjbs:
I'll wait until jscalendar-bis is merged.

PR #6010, opened 2026-05-26
Cyrus 3.13.5 prep

nwc10:


imap/jmap_api.c
```diff
@@ -17,7 +17,7 @@
 #include "http_dav.h"
 #include "http_dav_sharing.h"
 #include "http_jmap.h"
-#include "jmap_data_types.h"
+#include "imap/jmap_data_types.h"
```
elliefm: This is fine, but we usually put these ones in a separate block of include lines, with a comment very much like what you've used as the short commit message.  Usually when there's one such there's eventually another, and when the comment is there, the next one tends to be added correctly.  Whereas the first one, like this, is usually done wrong, and then needs to be fixed during release.

docsrc/download/release-notes/3.13/x/3.13.5.rst
```diff
@@ -99,6 +105,9 @@ Removed features
  * openio_proxy_timeout
  * openio_rawx_timeout
  * openio_verbosity
+ * restore_authname
+ * restore_password
+ * restore_realm
```
elliefm: I think leave the restore_foo out of this list.  They've been dead a long time; they do nothing in the current stable version, and probably a couple of previous stable versions before that.  We just forgot to mark them as deprecated at the time.

To put it another way, no behaviour has changed with respect to these options, we've just fixed some incorrect documentation.

docsrc/download/release-notes/3.13/x/3.13.5.rst
```diff
@@ -99,6 +105,9 @@ Removed features
...
```
elliefm, line 108: I think leave the `restore_foo` ones out of this list.  We removed the code that did anything with these options in _2020_ (3169b22801), we just forgot to mark them deprecated at the time.  They already don't do anything in 3.12, so there's no behaviour change in 3.13 for the release notes to report on.

elliefm (approved):
Looks good; a couple of nits but nothing critical.  If you make any changes based on my comments, I don't need to review them again for this to go ahead.

rjbs (approved):
LGTM.

imap/jmap_api.c
nwc10: Fixed.

git's defaults didn't actually show enough context to reveal that there **was** already such a block, so I moved the include there. I had missed the block because, I guess, I was confused and rushing - I didn't expect this step of the release process to break, and hence wasn't sure how many more new surprise blockers I was going to crash into.

I created *CYR-3063: Run `make distcheck` as part of CI* so that we find these problems earlier.

docsrc/download/release-notes/3.13/x/3.13.5.rst
nwc10: Done as requested.

docsrc/download/release-notes/3.13/x/3.13.5.rst
nwc10: I think that github duplicated your comment here. So I believe that this was done as described in the previous reply

PR #6011, opened 2026-05-26
cassandane: temporarily bundle our own Net::C---DAV clients

rjbs:
The newer Net::CalDAVTalk and Net::CardDAVTalk change their API substantially, and we're not in a good position to change that right now, for a variety of reasons.

This commit brings in specific versions: the ones that we know work. That's the (current) latest Net::CalDAVTalk, Net::CardDAVTalk, Text::JSContact, and a recent but not latest Text::JSCalendar.

I'd originally planned to make these internally-named copies, but that would have to touch many more files than I'd like to.  When we've carefully resolved the relevant issues, we can revert this commit (or just delete the files it added).

I do *not* plan to do anything about this in cyrus-docker (other than unbreak its current build).  It will continue to install latest from CPAN, which will get us all the prereqs.


wolfsage:
Thanks I hate it

wolfsage (approved):

PR #6012, opened 2026-05-27
TestPlan: report all tiny-tests errors, not just first seen

elliefm:
So you can just fix them all in one go instead of playing whack-a-mole with them


rjbs (approved):
Yeah nice.

wolfsage (approved):

PR #6018, opened 2026-05-28
jmap_calendar.c:setcalendars_update() avoid reading uninitialized value

dilyanpalauzov:


dilyanpalauzov:
In the path:
```c
static void setcalendars_update(jmap_req_t *req, const char *uid, json_t *arg, bool has_ext_props, json_t **record, json_t **err)
{
     char *mboxname = caldav_mboxname(req->accountid, uid);
     mbname_t *mbname = mbname_from_intname(mboxname);
 
     /* Make sure we don't mess up special calendars */
     if (jmap_calendar_isspecial(mbname)) {
         *err = json_pack("{s:s}", "type", "notFound");
         goto done; // ⇐ THIS IS EXECUTED
     }
     // THIS IS NOT EXECUTED
     /* Parse and validate properties. */
     struct setcalendar_props props;
     setcalendar_parseprops(req, &parser, &props, arg, mboxname);

…
done:
     setcalendar_props_fini(&props);
```
`setcalendar_props_fini` reads `props->defaultalarms_with_time` but props is not initialized.

rsto (approved):

Issue #4045, opened 2022-04-18
Reading uninitialized vanished_params->uidvalidity_is_max


dilyanpalauzov:
Clang static analzyer 14 sees at https://github.com/cyrusimap/cyrus-imapd/blob/master/imap/index.c#L1199 this flow:
```c
        struct vanished_params v;
        v.sequence = sequence;
        v.uidvalidity = state->mailbox->i.uidvalidity;
        v.modseq = fetchargs->changedsince;
        v.match_seq = fetchargs->match_seq;
        v.match_uid = fetchargs->match_uid;
        /* XXX - return error unless usinguid? */
        vanishedlist = index_vanished(state, &v); // here v->uidvalidity_is_max is not set

//continuing on line 932
seqset_t *index_vanished(struct index_state *state,
                              struct vanished_params *params)
{
    struct mailbox *mailbox = state->mailbox;
    seqset_t *outlist;
    seqset_t *seq;

    /* check uidvalidity match */
    if (params->uidvalidity_is_max) {
        if (params->uidvalidity < mailbox->i.uidvalidity) return NULL;
    }
```
`params->uidvalidity_is_max` is read, before being set in advance.  In fact the member `uidvalidity_is_max` is present only twice in the whole codebase: on the above line, and in the definition of struct vanished_params in index.h

ksmurchison:
@brong Do you remember the logic here?  This was added for xconvfetch per the commit history but not set anywhere

dilyanpalauzov:
Addressed by 50fafe3768bfb12977.

dilyanpalauzov:
>  In fact the member `uidvalidity_is_max` is present only twice in the whole codebase: on the above line, and in the definition of struct vanished_params in index.h

The above statement is still correct. `uidvalidity_is_max` is aways zero. Before https://github.com/cyrusimap/cyrus-imapd/commit/50fafe3768bfb129778801350e454a105dc2a858 it is was sometimes zero and sometimes it was read, before being initialized.

PR #6022, opened 2026-05-29
various: fix issues found by running cassandane under valgrind

elliefm:
Fixes or workarounds for issues found by running cassandane under valgrind (which CI doesn't do).

For whatever it's worth, a full `./testrunner.pl --valgrind` of the master branch takes around 1 hour 40 mins on my current laptop.  Which is still too long to do routinely, but is now at least feasible to run occasionally.

My priority here is to get the tests passing; this is not a comprehensive audit.

### DIFF
--- a/imap/index.c
+++ b/imap/index.c
@@ -922,6 +922,9 @@ seqset_t *index_vanished(struct index_state *state,
     seqset_t *seq;
 
     /* check uidvalidity match */
+    /* XXX Seems like uidvalidity_is_max never gets set to anything but zero.
+     * XXX What was its purpose?  Can we get rid of it?
+     */
     if (params->uidvalidity_is_max) {
         if (params->uidvalidity < mailbox->i.uidvalidity) return NULL;
     }
@@ -1231,7 +1234,7 @@ EXPORTED int index_fetch(struct index_state *state,
     }
 
     if (fetchargs->vanished) {
-        struct vanished_params v;
+        struct vanished_params v = {0};
         v.sequence = sequence;
         v.uidvalidity = state->mailbox->i.uidvalidity;
         v.modseq = fetchargs->changedsince;
--- a/cassandane/vg.supp
+++ b/cassandane/vg.supp
@@ -104,3 +104,19 @@
   ...
 }
 
+{
+   # shush leaked SASL internal state
+   # XXX this might be a real leak, but if so a fix is not immediately clear
+   sasl_client_add_plugin
+   Memcheck:Leak
+   match-leak-kinds: definite
+   fun:malloc
+   fun:xmalloc
+   obj:*
+   obj:*
+   fun:sasl_client_add_plugin
+   obj:/usr/lib/x86_64-linux-gnu/libsasl2.so.*
+   fun:sasl_client_init
+   fun:global_sasl_init
+   ...
+}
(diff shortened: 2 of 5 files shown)


nwc10:
I believe this commit to be correct, so I've approved the commit.

However, as I wasn't familiar with it, I went looking. I think that there's the same leak pattern in `static void cmd_getacl`:

```
    char *intname = mboxname_from_external(name, &imapd_namespace, imapd_userid);

    r = mlookup(tag, name, intname, &mbentry);
    if (r == IMAP_MAILBOX_MOVED) return;
```

(but curiously the `prot_printf(imapd_out, "%s NO %s\r\n", tag, error_message(r));` failure case is covered with a `free`)

...

and both in `static void cmd_myrights` and there I stopped looking.

Finding and fixing these should be a CYR-* ticket, right?

nwc10 (dismissed):

ksmurchison (dismissed):
This looks good, but @nwc10 Is correct about cmd_getacl().  It also looks like we leak mbname in cmd_setacl() in the same manner.  I think we should probably fix all three in one PR.

rjbs:
Recusing myself from this.  Nobody is less likely to find a problem than I am, of that list of reviewers. :)

elliefm:
I've extended this to fix the `cmd_getacl` leak that @nwc10 observed.  I've also extended it to fix everything else that valgrind complained about after a full run, and have consequently updated the PR title and description.

@ksmurchison I looked hard at `cmd_setacl` but couldn't see any leak there.  All the error paths in that function use `goto done` (no early returns), and the `done:` label frees `mbname`.

There is at least one more leak that I spotted in `ldap.c` while I was in there, but that module plays fast and loose with object lifetimes and ownership; the obvious quick fix would just introduce a "use after free" crash instead.  That is, the leak is load-bearing.  We don't test that code anyway, so the leak isn't holding me up, so I've left it alone rather than try to fix it properly.

With these patches, I still get around half a dozen test failures from `./testrunner.pl --valgrind`, but the remaining failures are all tests with timing quirks, which fail due to valgrind making things slow, not due to valgrind detecting problems.  So I'm leaving those alone for now too.

elliefm:
Found another leak which already exists, but is only exposed by some tests that aren't ready to land yet.  Back to draft while I address that one too!

rsto:
@elliefm Is this PR a draft or ready for review?

imap/index.c
```diff
@@ -1231,7 +1234,7 @@ EXPORTED int index_fetch(struct index_state *state,
     }
 
     if (fetchargs->vanished) {
-        struct vanished_params v;
+        struct vanished_params v = {0};
```
dilyanpalauzov, line 1237: This was reported at https://github.com/cyrusimap/cyrus-imapd/issues/4045.  Other findings by this static analyzer are not reported.

dilyanpalauzov: I think it is sufficient just to set the `uidvalidity_is_max` field, instead of `= { 0 }`.

elliefm:
> Found another leak which already exists, but is only exposed by some tests that aren't ready to land yet. Back to draft while I address that one too!

That is now fixed, and so is one more that turned up while re-testing.  This PR is now ready for review.

If anything else turns up I'll start a new PR, so that this one at least can move forward.

rsto (approved):

elliefm:
**Backporting:** also grab 633bf2824396cdf18b8460ff871b88e792f7e560 from #4672 if the target branch doesn't have it

PR #6028, opened 2026-05-29
jmap_contact.c:jsprop_from_vcard() - compare to `geo:`

dilyanpalauzov:
Clang says “comparison length is too long and might lead to a buffer overflow”.


ksmurchison:
I think the proper fix here is to compare against “geo:” rather than “geo” since it’s a URI. 
@rsto do you agree.

rsto:
> I think the proper fix here is to compare against “geo:” rather than “geo” since it’s a URI. @rsto do you agree.

Right, that should be `geo:`

ksmurchison (approved):

PR #6029, opened 2026-05-29
GitHub Actions: add merge-on-demand workflow

rjbs:
When someone with write access to the target repository commands "!!merge-this" on a blank line, the PR will be rebased and merged.  We create a semi-linear merge (--no-ff) so we can see what happened in what branch, but don't have to worry about cross-crossed hitory.  The process is:

1. fetch the feature branch
2. rebase it on the target
3. force-push back to the feature branch
4. merge, --no-ff, into the target
5. push the target

Processing multiple `!!merge-this` commands is serialized to prevent races.

Step 3 can be impossible, or can fail.  Maybe "Allow edits from maintainers" isn't ticked.  Maybe the push would send workflow changes into the feature branch -- which GitHub strictly forbids.  In those cases, we'll skip step 3 and just close the PR with a comment saying "rebased as XXX and merged as YYY".

In other error cases, we'll comment that something went wrong and leave it to the would-be merger to figure it out.

### DIFF
--- a/.github/workflows/merge-on-demand.yaml
+++ b/.github/workflows/merge-on-demand.yaml
@@ -1,0 +1,284 @@
+name: Merge on demand
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  merge-on-demand:
+    # This is a pre-filter.  This would match "Do not !!merge-this", or
+    # ">"-prefixed quoted form, both of which would be foolish to trigger on.
+    # But it lets us not run an action at all when it can't possibly match.
+    # The more careful checking is down in the steps below.
+    if: >-
+      github.event.issue.pull_request.url != '' &&
+      contains(github.event.comment.body, '!!merge-this')
+    runs-on: ubuntu-latest
+
+    # Serialize merges repo-wide: two commands racing to push the base
+    # branch would leave the loser with a non-fast-forward rejection. The
+    # base branch isn't in the issue_comment payload, so we can't key on
+    # it; merges are rare and fast, so serializing the whole workflow is
+    # cheap. Never cancel an in-flight merge mid-push.
+    concurrency:
+      group: merge-on-demand-${{ github.repository }}
+      cancel-in-progress: false
+
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+      - name: Check permission and get PR details
+        id: info
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
+        with:
+          script: |
+            const body = context.payload.comment.body || '';
+            const isCommand = body
+              .split(/\r?\n/)
+              .some(line => line.trim() === '!!merge-this');
+
+            if (!isCommand) {
+              core.info('no !!merge-this command line found; skipping');
+              return;
+            }
+
+            const login = context.payload.comment.user.login;
+
+            const { data: collab } = await github.rest.repos.getCollaboratorPermissionLevel({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              username: login,
+            });
+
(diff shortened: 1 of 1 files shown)


rjbs:
I'm shooting for making it easier to keep a semi-linear history without creating a bunch of tedious work.

wolfsage:
Have you read https://securitylab.github.com/resources/github-actions-new-patterns-and-mitigations/ specifically the IssueOops: Security Pitfalls with issue_comment Trigger section?

From when you add a comment to when the workflow triggers, a malicious user can change the repo. (Sometimes there's a 1-5 second delay, enough time if someone has set up a script to watch / pay attention). There's other issues too.

It sounds like the recommended approach is to trigger a worfklow on a label addition -
```
Since the Labeled activity type for a pull_request trigger contains details about the latest commit SHA of the Pull Request, there is no need for workflow to resolve a Pull Request number into the latest commit SHA, as it is the casse with issue_comment, and therefore there is no window for an attacker to modify the Pull Request. 
```
Anyway, if we stick with this one... it seems like:
* You can say !!merge-this on an already merged merge request and the bot will try to do something. It should detect that an MR isn't open anymore
* I'd argue that for !!merge-this it should be the *only* content of a comment which avoids all sorts of potential "attacks"

rjbs:
I talked about this some more with @wolfsage and discussed ways to avoid moving to label, which has its own issues, but eventually agreed we'd do that.

@nwc10 I suggest you just review the total diff.  The first commit is only kept separate for the sake of history.

.github/workflows/merge-on-demand.yaml
```diff
@@ -0,0 +1,284 @@
...
+          # the PR touches no workflow files: rebasing onto base pulls in
+          # base's own changes under .github/workflows/, and GITHUB_TOKEN may
+          # not push those to the fork. The base push is unaffected (base
+          # already has them), so for that one expected rejection we fall
+          # through to merge-and-close. Any other push failure is a genuine
+          # error and aborts before base is touched, rather than silently
+          # downgrading to a non-merged PR closure.
+          PUSHED_HEAD=false
+          if [ "${CAN_PUSH_HEAD}" = "true" ]; then
+            if push_err=$(git push --force "${FORK_URL}" "pr-branch:${HEAD_REF}" 2>&1); then
```
nwc10: I **think** that this would be better as `--force-with-lease`

(I looked at `--force` and thought "when is that ever a great idea?")

I think as-is, there's a small race condition that hurts the contributor (but not us, directly)

1. contributor pushes code
2. we review, say yes, add this label
3. contributor updates code
4. this workflow runs
5. this workflow erases their last changes

It will be clear that this has happened, the SHA of their last push I believe is in the PR history, and we will have merged what we reviewed. But unless we're looking out for this, it will get missed. Whereas a `--force-with-lease` would stop the workflow (as written) and hit the error case, and everyone gets a chance to re-review the new version.

Which then makes me wonder if there's a similar race condition in hitting "merge" on the github website. I hope not. I assume that someone would have tripped over it by now.

nwc10 (changes requested):
I think `--force-with-lease` would be better, but I don't know if I've missed something.

I've read all the rest, and it seems to make sense, and it appears not to fall into any of the security traps that the linked post mentions. But I haven't tested it, and I'm not really familiar with either JS or the GH API. (Or enough YAML to know when YAML is doing something to what I think is a literal string. The bash bits are OK, and everything seems to be in `""`)

nwc10:
You can test the rebase and fail-to-push with #6075

It changes a workflow (to remove a warning that GitHub themselves caused). I deliberated branched from out-of-date `master` to ensure that it needs a rebase before merging.

nwc10 (approved):

PR #6030, opened 2026-05-30
util.h: fix compilation with clang++ for alias("wrap_clock_gettime")

dilyanpalauzov:
Clang++ 22.1.6 fails compiling:
```
$ make imap/xapian_wrap.o CXX=clang++
  CXX      imap/xapian_wrap.o
In file included from imap/xapian_wrap.cpp:15:
./lib/util.h:468:12: warning: unused function 'wrap_clock_gettime' [-Wunused-function]
  468 | static int wrap_clock_gettime(clockid_t id, struct timespec *ts)
      |            ^~~~~~~~~~~~~~~~~~
./lib/util.h:472:22: error: alias must point to a defined variable or function
  472 | __attribute__((weak, alias("wrap_clock_gettime"), visibility("default")))
      |                      ^
./lib/util.h:472:22: note: the function or variable specified in an alias must refer to its mangled name
./lib/util.h:472:22: note: function by that name is mangled as "_ZL18wrap_clock_gettimeiP8timespec"
  472 | __attribute__((weak, alias("wrap_clock_gettime"), visibility("default")))
      |                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                      alias("_ZL18wrap_clock_gettimeiP8timespec")
1 warning and 1 error generated.
make: *** [Makefile:7776: imap/xapian_wrap.o] Error 1
```
This chaneg included, with g++ 16.1:
```
$ readelf --all -W imap/.libs/libcyrus_imap_la-xapian_wrap.o
Symbol table '.symtab' contains 668 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
    79: 00000000000048c0     5 FUNC    LOCAL  DEFAULT   64 wrap_clock_gettime
   409: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND clock_gettime
   410: 00000000000048c0     5 FUNC    WEAK   DEFAULT   64 cyrus_gettime
```
With clang++:
```
$ readelf --all W imap/.libs/libcyrus_imap_la-xapian_wrap.o
Symbol table '.symtab' contains 787 entries:
   Num:    Value          Size Type    Bind   Vis      Ndx Name
     3: 0000000000000000     5 FUNC    LOCAL  DEFAULT    2 _ZL18wrap_clock_gettimeiP8timespec
   388: 0000000000000000     0 NOTYPE  GLOBAL DEFAULT  UND clock_gettime
   784: 0000000000000000     5 FUNC    WEAK   DEFAULT    2 cyrus_gettime
```

### DIFF
--- a/lib/util.h
+++ b/lib/util.h
@@ -466,11 +466,13 @@ enum xsyslog_ev_arg_type {
  * We need this shim because, unlike gettimeofday, clock_gettime itself is
  * not a weak alias, so it can't be overridden directly.
  */
+#ifndef __cplusplus
 static int wrap_clock_gettime(clockid_t id, struct timespec *ts)
 {
     return clock_gettime(id, ts);
 }
 __attribute__((weak, alias("wrap_clock_gettime"), visibility("default")))
 extern int cyrus_gettime(clockid_t, struct timespec *);
+#endif
 
 #endif /* INCLUDED_UTIL_H */


elliefm (commented):
Oh, interesting.  Can we rely on the mangled name to be consistent across builds/platforms/compilers?

xapian_wrap.cpp doesn't actually use `cyrus_gettime`, and icu_wrap.cpp doesn't include util.h.  Those are our only C++ source files.  So the other approach we might take here is to just wrap the whole `cyrus_gettime` section of util.h in `#ifndef __cplusplus`, and leave it unavailable from C++ code.  What do you think?

dilyanpalauzov:
> Can we rely on the mangled name to be consistent across compilers?

Here one mangling has to be used for GCC (the so called no-mangling/C-binding) and another for CLANG.  For this reason the mangled name is not consistent across compilers.  This might be a bug in clang.

I do not know how `wrap_clock_gettime` is used.  Any approach, which permits compiling the source code with clang, is good.

Also the ld.bfd and gold linkers have `--wrap`  option, which seems to be intended for similar uses cases. (But again this is a guess, as I have not looked in the details here.)

elliefm:
> I do not know how `wrap_clock_gettime` is used.

When cyrus is running normally, calls to `cyrus_gettime()` are calls to `wrap_clock_gettime()`, which just calls the platform `clock_gettime()`.  I expect tail call optimisation probably eliminates the extra function call if optimisations are turned on, but it doesn't really matter if it doesn't.

When a unit test is running, a mock of `cyrus_gettime()` is provided by the unit testing infrastructure, which lets the test control how time passes, for testing time-based behaviours.  There's similar mocks for `gettimeofday()`, `time()`, `sleep()`, and `nanosleep()`; they're all in `cunit/unit-timeofday.c`.  You can see examples of how this stuff is used in the testc files that `#include "cunit/unit-timeofday.h"`.

In libc, `gettimeofday()` etc are weak aliases, which is how we can replace them with mocked versions under testing.  But `clock_gettime()` is not a weak alias, so we can't mock it.  Instead, libcyrus provides `cyrus_gettime()`, which does the same thing, but _is_ a weak alias that we can mock.  Cyrus code should always call `cyrus_gettime()`, and never call `clock_gettime()` directly.

> For this reason the mangled name is not consistent across compilers. [...]
> Any approach, which permits compiling the source code with clang, is good.

In that case, I think the cleanest thing to do is just put `#ifndef __cplusplus` around this section, so that C++ code doesn't even see it, no matter what compiler is in use.  That might be a problem in the future if for some reason we need to write time dependent code in C++, but that doesn't seem likely, so I don't think we need to worry about it now.

elliefm (approved):
Thanks!

Sorry for the delay, I didn't notice that you'd updated it.

PR #6032, opened 2026-05-30
configure.ac: require libical 4.0 instead of 3.99.99

dilyanpalauzov:
Version 3.99.99 is hard to install and hard to define what exactly it means.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1461,7 +1461,7 @@ dnl
                                 [Do we have support for xmlBufferDetach()?]))
                 ])
 
-        PKG_CHECK_MODULES([ICAL], [libical >= 3.99.99], [
+        PKG_CHECK_MODULES([ICAL], [libical >= 4.0], [
                 AC_DEFINE(HAVE_ICAL,[],[Build CalDAV support into httpd?])
                 LIBS="$LIBS ${ICAL_LIBS} -licalvcard"
                 with_ical=yes


elliefm:
My understanding (which may be out of date) is that the cyrus-imapd master branch currently requires a bunch of custom patches on top of the libical master branch, and that when built in this way, the libical version is reported as 3.99.99.  So we can't update the configure check, yet.

Longer term, we expect most or all of our patches to be integrated into libical 4.

The next major cyrus-imapd release will require libical 4, and will be expected to work with a vanilla libical 4.  By then, the configure checks will have been updated accordingly.  But we're not there yet.

In the meantime, the patched version of libical in [cyruslibs](https://github.com/cyrusimap/cyruslibs) is known to work with the cyrus-imapd master branch.

rsto:
> My understanding (which may be out of date) is that the cyrus-imapd master branch currently requires a bunch of custom patches on top of the libical master branch, and that when built in this way, the libical version is reported as 3.99.99. So we can't update the configure check, yet.

That is not the case anymore. Current main branch builds on libical 4, there are no custom libical patches anymore. The libical submodule in cyruslibs points to a pinned libical 4 release, which we expect to update regularly until some point where we switch to a (Debian-packaged) libical 4 release.

dilyanpalauzov:
It would be good to spell explicitly, that cyrus imap 3.10 and 3.12 work only with the released libical 3.0 version or with the patched version of libical in cyruslibs, while current development branch / 3.13 requires libical 4.0,

Currently https://www.cyrusimap.org/imap/download/release-notes/3.10/x/3.10.0.html states:

> Adds experimental support for JMAP Contacts per upcoming IETF standards. Requires the not-yet-released [libicalvcard](https://github.com/libical/libical/pull/584)

But compiling with the realeased libicacvcard in 3.10 or 3.12 fails.  So the above statement is not correct, and it applies (is inherited, is not cancelled)  for cyrus imap 3.12.

PR #6034, opened 2026-06-01
Add a test showing CalDAV PUT mbox ids are wrong

wolfsage:
Cyrus sends an event with EM-- as the event id when using compact calendar event ids. -- is '0', which does not match up with the modseq of the actually created event (something like EM6F).


ksmurchison:
Superseded by [](https://github.com/rjbs/cyrus-imapd/pull/1)

PR #6037, opened 2026-06-02
jmap_blob.c: look up ContactCard ids, not the removed Contact type

rjbs:
Blob/lookup still implemented the (otherwise totally deleted) Contact datatype, and never got ContactCard.  Rip out Contact id resolution, bring in ContactCard resolution.


ksmurchison (approved):
Once this lands, I will rewrite jmap_blob.c to use jmap_data_types.h.  I don't know how I missed doing that when I introduced jmap_data_types.gerf and used it for jmap_push.c and jmap_quota.c

PR #6038, opened 2026-06-03
mboxevent.c: add vnd.fastmail.compactCalendarIds to each event

ksmurchison:


rjbs:
have applied to rsto/cyr-2761-jscalendarbis with false switched to true

PR #6040, opened 2026-06-04
acl_afs.c:cyrus_acl_myrights() - reduce scope of acl_ptr

dilyanpalauzov:
For me reducing the scope of `acl_ptr` makes understanding the function easier.

### DIFF
--- a/lib/acl_afs.c
+++ b/lib/acl_afs.c
@@ -29,10 +29,8 @@ EXPORTED int cyrus_acl_myrights(const struct auth_state *auth_state, const char 
     char *acl = xstrdupsafe(origacl);
     char *thisid, *rights, *nextid;
     long acl_positive = 0, acl_negative = 0;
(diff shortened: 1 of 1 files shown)


lib/acl_afs.c
```diff
@@ -47,10 +45,11 @@ EXPORTED int cyrus_acl_myrights(const struct auth_state *auth_state, const char
         }
         *nextid++ = '\0';
 
+        long *acl_ptr;
         if (*thisid == '-') {
             acl_ptr = &acl_negative;
             thisid++;
-        }
+        } else acl_ptr = &acl_positive;
```
ksmurchison: I'd like to see the else clause on its own line (possibly in { } ) and with a blank line following, so it is easier to see

ksmurchison (changes requested):

ksmurchison (approved):

PR #6045, opened 2026-06-05
jmap_mail.c: fix race condition between queryChanges and squatter

rsto:
This fixes a race condition between `Email/query[Changes]` and squatter for queries that require full text search. To do so, it adds a new entry in cyrus.indexed.db in the Xapian backend which tracks the highest createdmodseq of all messages in the search index. This allows to limit messages in `Email/query[Changes]` results to those messages that also already are indexed.

Without this change, a message that arrives before `Email/query` finishes but before squatter indexes it can produce a query state that encodes that message's createdmodseq. Following `queryChanges` request then skip past that message, producing invalid queryChanges results.

This change is backwards compatible: as long as the index didn't get written with a new message, `Email/query[Changes]` will execute as before (including the race condition bug). As soon as a new message got indexed, the `Email/query` responses will include the new query state and queryChanges requests with a stale query state will get rejected with `cannotCalculateChanges`.

This change is done in multiple steps, each step is a commit:

1. `search_engines.h: require a read-locked session for queries` refactors the existing query code so that multiple queries can share the same read-lock on the Xapian backend. This is required so that no squatter can run in between two subqueries of a query and invalidate the createdmodseq that's kept now in cyrus.indexed.db.
2. `search_xapian.c: store highest createdmodseq in indexed.db` updates the Xapian backend code to store the highest createdmodseq (and an index generation marker) in cyrus.indexed.db. This covers all usage scenarios of squatter, including compaction with and without reindexing.
3. `jmap_mail.c: fix race condition between queryChanges and squatter` then plumbs that newly introduced createdmodseq of the search index in the JMAP `Email/query[Changes]` code.

An additional tiny commit updates the documentation for squatter for the attachextract feature. This is not related to this change, but it naturally fit in when I updated the squatter docs.

Looking at the commit messages, you might be tempted to think that they are unnecessarily verbose LLM output. They are not, I wrote them as succinctly as possible to still cover the design. If you miss something there or in the source documentation, please let me know.

Thanks to @wolfsage for providing the initial Cassandane test that reproduced the race condition.

### DIFF
--- a/imap/search_engines.c
+++ b/imap/search_engines.c
@@ -48,7 +48,9 @@ static const struct search_engine default_search_engine = {
     NULL,
     NULL,
     NULL,
-    NULL
+    NULL,
+    NULL,
+    NULL,
 };
 
 EXPORTED const struct search_engine *search_engine(void)
@@ -144,19 +146,35 @@ EXPORTED int search_part_is_header(enum search_part part)
     return 0;
 }
 
-EXPORTED search_builder_t *search_begin_search(struct mailbox *mailbox, int opts)
+EXPORTED search_builder_t *search_begin_search(search_session_t *session)
 {
-    if (!mailbox) return NULL;
+    if (!session) return NULL;
 
     const struct search_engine *se = search_engine();
-    return (se->begin_search ?
-            se->begin_search(mailbox, opts) : NULL);
+    return (se->begin_search ? se->begin_search(session) : NULL);
 }
 
 EXPORTED void search_end_search(search_builder_t *bx)
 {
     const struct search_engine *se = search_engine();
     if (se->end_search) se->end_search(bx);
+}
+
+EXPORTED search_session_t *search_begin_session(struct mailbox *mailbox,
+                                                int opts)
+{
+    if (!mailbox) return NULL;
+
+    const struct search_engine *se = search_engine();
+    return (se->begin_session ?
+            se->begin_session(mailbox, opts) : NULL);
+}
+
+EXPORTED void search_end_session(search_session_t *session)
+{
+    if (!session) return;
+    const struct search_engine *se = search_engine();
+    if (se->end_session) se->end_session(session);
 }
 
 EXPORTED search_text_receiver_t *search_begin_update(int verbose)
--- a/imap/search_engines.c
+++ b/imap/search_engines.c
@@ -533,8 +533,9 @@ EXPORTED int search_update_mailbox(search_text_receiver_t *rx,
     if (!dynarray_size(&attachparts))
         goto done;
 
-    // Release mailbox lock
-    r = rx->end_mailbox(rx, mailbox);
+    // Release mailbox lock. Indexing this mailbox is not complete yet,
+    // it resumes after extracting attachments.
+    r = rx->end_mailbox(rx, mailbox, /*has_more*/true);
     if (r) goto done;
     mailbox_close(&mailbox);
 
@@ -592,7 +593,7 @@ EXPORTED int search_update_mailbox(search_text_receiver_t *rx,
     free(mycachedir);
     free(mboxname);
     {
-        int r2 = rx->end_mailbox(rx, mailbox);
+        int r2 = rx->end_mailbox(rx, mailbox, /*has_more*/false);
         if (r) return r;
         if (r2) return r2;
     }
--- a/imap/search_engines.c
+++ b/imap/search_engines.c
@@ -51,6 +51,7 @@ static const struct search_engine default_search_engine = {
     NULL,
     NULL,
     NULL,
+    NULL,
 };
 
 EXPORTED const struct search_engine *search_engine(void)
@@ -175,6 +176,16 @@ EXPORTED void search_end_session(search_session_t *session)
     if (!session) return;
     const struct search_engine *se = search_engine();
     if (se->end_session) se->end_session(session);
+}
+
+EXPORTED modseq_t search_session_get_highest_createdmodseq(search_session_t *session,
+                                                           uint64_t *index_generation)
+{
+    if (index_generation) *index_generation = 0;
+    if (!session) return 0;
+    const struct search_engine *se = search_engine();
+    return se->session_get_highest_createdmodseq ?
+            se->session_get_highest_createdmodseq(session, index_generation) : 0;
 }
 
 EXPORTED search_text_receiver_t *search_begin_update(int verbose)
--- a/imap/search_engines.c
+++ b/imap/search_engines.c
@@ -19,6 +19,7 @@
 #include "global.h"
 #include "ptrarray.h"
 #include "search_engines.h"
+#include "seqset.h"
 #include "attachextract.h"
 
 /* generated headers are not necessarily in current directory */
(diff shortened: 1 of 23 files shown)


wolfsage:
One thing that's unclear to me. If there's lots of subqueries/lots of folders now, does search take a lock and release the lock every time it switches between folders or subqueries?

It sounds like in this new system, it'll hold the lock for the *entire* search.

If these two behave differently as I'm guessing, I think this means queries that take awhile can block squatter from doing its work?

rsto:
> One thing that's unclear to me. If there's lots of subqueries/lots of folders now, does search take a lock and release the lock every time it switches between folders or subqueries?

Yes, currently for every subquery (a DNF clause in the normalized original query) it takes a read-lock on the activefile, processes that subquery, then releases the lock.

> It sounds like in this new system, it'll hold the lock for the _entire_ search.

Correct, it takes it for the entire query, not individual subqueries. But the query on the Xapian backend only is part of the entire search, all the post-processing for mailboxes and messages is done after the session lock is released.
 
> If these two behave differently as I'm guessing, I think this means queries that take awhile can block squatter from doing its work?

First of all, it means that a query gets evaluated on a consistent search index. Before, different subqueries could get run on different states of the search database, because a squatter could have indexed in between two subqueries.

This indeed means that squatter has to wait for the entire query, not just a subquery. But on the other side, a query now can't get blocked by an interleaved squatter, so that potentially makes queries faster on average. A typical query consists of one, maybe a few subqueries, so we only widen the worst-case window for the wait time for squatter a bit. In exchange for that we get consistent search results and can fix Email/queryChanges.

Should squatter wait time become too long then we should rewrite squatter to not block on a shared-lock user and just continue indexing some other user that isn't locked.

imap/search_query.c
```diff
@@ -828,12 +828,17 @@ static int add_found_uid(const char *mboxname, uint32_t uidvalidity,
...
 
+struct subquery_run_indexed_ctx {
+    search_query_t *query;
+    search_session_t *session;
+};
+
 static void subquery_run_indexed(const char *key __attribute__((unused)),
                                  void *data, void *rock)
 {
-//     const char *mboxname = key;
```
brong, line 834: I know this is a stale comment, but it did give a little value to knowing what the calling params are!

imap/search_xapian.c
```diff
@@ -1361,13 +1361,21 @@ struct opnode
...
-struct xapian_builder {
-    search_builder_t super;
+/*
+ * A search session takes a shared lock on the activefile. Within
+ * a session, all Xapian and indexeddb databases for tiers of that
+ * activefile are guaranteed to not change.
+ */
+struct search_session {
     struct xapiandb_lock lock;
-    seqset_t *indexed;
```
brong, line 1368: what happened to "indexed"?

brong: Hah, my parallel claude review agent said: 

`bb->indexed` was loaded but never consumed in master — removing it is safe.

brong (approved):
OK - I think I understand the process, and it's pretty clever!  My worry is that it will cause a lot of "cannotCalculateChanges" on busy accounts, where deliveries are frequently happening to multiple mailboxes, because squatter processes each mailbox in order and doesn't do a whole user in a transaction, so it'll bump the CREATEDMODSEQ too high on the first mailbox, then hit an inversion on the second, and vice versa.

I do think this is the right approach though, so I'm going to approve this with a caveat that we may need to make squatter process entire users in a batch as well if we see it causing too many problems.

imap/search_query.c
rsto: I renamed the `key` parameter to `serialised_sub`. It's not the mboxname, that only is correct for the `subquery_run_folder` function.

imap/search_xapian.c
rsto: yep, that was just dead code

rsto:
> My worry is that it will cause a lot of "cannotCalculateChanges" on busy accounts, where deliveries are frequently happening to multiple mailboxes, because squatter processes each mailbox in order and doesn't do a whole user in a transaction,

Right, that could become a real issue. I have added logging so that we can better keep track of this: we now log (LOG_INFO) whenever the index generation changes during indexing, or if a queryChanges gets rejected with `cannotCalculateChanges` because of the server and the client query states getting out of sync.

I also fixed another real issue that's similar to the high-churn user case: the two-phased attachment indexing for a mailbox releases the activefile lock between indexing messages without attachments and indexing messages with attachments (to avoid long-locking because of waiting for the attachment extractor). But that almost always invalidated the index generation! The new code now updates the index as before in transactional batches, but it only writes createdmodseq when squatter indicates the index backend that all new messages in the message got indexed.

rsto:
This needs more work.

rsto:
I added two commits, one of which contains a bunch of tests that assert failures modes for squatter and the other that actually changes squatter and the search backend to correctly handle these failures. The main issues were squatter not dealing correctly with mailboxes getting deleted in between the two indexing phases (first for plain text only message, the second for messages with attachments that called out to the extractor service), and messages getting unlinked in between these passes.

The main functional change now is that we do not index in two passes more at all. Instead, we first determine which messages of a mailbox need the extractor service, call the extractor for those, only then take the mailbox lock and index both the messages that needed the extractor and those that did not. This solves an issue where a message indexed in the first pass could extend the sequence set to a range of uids that includes a message of the second pass. If the message in that second pass failed it got stored as indexed in indexed.db, even though we did in fact not index it at all.

rsto:
To be clear, these failures modes already were present before introducing createdmodseq. But now they became more prominent, because we now carried over more state in between the two passes and also because we really want the highest_createdmodseq to be correct, as otherwise we again produce bogus responses in Email/queryChanges.

imap/search_engines.c
```diff
@@ -211,34 +212,76 @@ static int search_batch_size(void)
...
 
         const char *fname;
-        r = message_get_fname(msg, &fname);
-        if (r) return r;
-        r = warmup_file(fname, 0, 0);
-        if (r) return r; /* means we failed to open a file,
-                            so we'll fail later anyway */
+        if (!r) r = message_get_fname(msg, &fname);
+        if (!r) r = warmup_file(fname, 0, 0);
+        /* an error here means we failed to open a file,
```
wolfsage: Why not have an `if (r) break;` at the start or end of the loop?

Continuing to loop *seems* to serve no purpose, so it's mildy confusing that we do so. Makes me think "Is there some side effect I'm not seeing?"

imap/search_xapian.c
```diff
@@ -2445,34 +2424,58 @@ static int check_directory(const char *dir, int verbose, int create)
...
+                lf_s("mailbox", mailbox_name(tr->super.mailbox)),
+                lf_u("uncommitted", uncommitted),
+                lf_s("error", error_message(r)));
+        return r;
+    }
+    gettimeofday(&end, NULL);
+
+    xsyslog_ev(LOG_INFO, "committed Xapian transaction",
+                lf_u("committed", uncommitted),
+                lf_f("seconds", timesub(&start, &end)));
```
wolfsage, line 2456: Would "seconds" also be useful info in the event of "failed to commit transaction" above?

imap/search_engines.c
rsto: Yeah, that code originally wasn't great with its multiple return sites, and now it hasn't become better. I rewrote it to have exactly one return site at the end of the function.

imap/search_xapian.c
rsto: done, thanks

rsto:
The CI failure is a fluke, it's the `search_sort_thread_maxtime` test that is indeterministic.

PR #6047, opened 2026-06-07
mailbox: handle > 4Gb indexes safely

brong:
We hit an issue where a mailbox with tens of millions of messages hit the 4Gb cyrus.index size limit.

Now - we shouldn't have one of those, but Cyrus also shouldn't randomly corrupt things!

### DIFF
--- a/imap/mailbox.c
+++ b/imap/mailbox.c
@@ -1848,6 +1848,11 @@ static int mailbox_buf_to_index_header(const unsigned char *buf, size_t len,
     return 0;
 }
 
+static size_t mailbox_record_offset(const struct mailbox *m, uint32_t recno)
+{
+    return (size_t)m->i.start_offset + (size_t)(recno - 1) * m->i.record_size;
+}
+
 static int mailbox_refresh_index_map(struct mailbox *mailbox)
 {
     size_t need_size;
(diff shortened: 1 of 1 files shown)


rjbs:
The fix looks like an improvement, but I was surprised it was only these places.  I'm not an expert, but given the experts are mostly enjoying their weekends, I thought I'd pass it off to Claude for a look.  Claude points to fixes also needed to:
* `mailbox_wipe_index_record`
* `_commit_changes`
* `mailbox_refresh_index_map`

Claude also suggests adding `static off_t mailbox_record_offset(const struct mailbox *m, uint32_t recno)` which seems reasonable to me.

My read is that the `mailbox_refresh_index_map` probably means this isn't even a short-term fix to ship, but I think it's probably much more effective for me to say "Give this another look?" than to try to work through it myself.

If you like, I'll supply the whole analysis.

brong:
> The fix looks like an improvement, but I was surprised it was only these places. I'm not an expert, but given the experts are mostly enjoying their weekends, I thought I'd pass it off to Claude for a look. Claude points to fixes also needed to:
> 
> * `mailbox_wipe_index_record`
> * `_commit_changes`
> * `mailbox_refresh_index_map`
> 
> Claude also suggests adding `static off_t mailbox_record_offset(const struct mailbox *m, uint32_t recno)` which seems reasonable to me.
> 
> My read is that the `mailbox_refresh_index_map` probably means this isn't even a short-term fix to ship, but I think it's probably much more effective for me to say "Give this another look?" than to try to work through it myself.
> 
> If you like, I'll supply the whole analysis.

I took your comments there and went with the same idea but using size_t since we use that pretty consistently and it can't be negative.

The extra fix I had already identified got us most of the way there, this also fixes the ->index_size bit.

rjbs (approved):

PR #6048, opened 2026-06-08
cassandane: finer granularity cleanup option

elliefm:
The actual cleanup code is unchanged; this just makes the pre-run and post-run cleanup phases individually configurable, instead of only both or neither.

See the documentation in cassandane.ini.example for usage.

### DIFF
--- a/cassandane/testrunner.pl
+++ b/cassandane/testrunner.pl
@@ -174,9 +174,11 @@ while (defined(my $a = shift))
     {
         $cassini_filename = shift;
     }
-    elsif ($a eq '-c' || $a eq '--cleanup')
+    elsif ($a eq '-c' || $a =~ m/^--cleanup(?:=(\w*))?$/)
     {
-        push(@cassini_overrides, ['cassandane', 'cleanup', 'yes']);
+        my $v = $1 // 'yes';
+        usage if not grep { $v eq $_ } qw(yes no pre post);
+        push(@cassini_overrides, ['cassandane', 'cleanup', $v]);
     }
     elsif ($a eq '--no-cleanup')
     {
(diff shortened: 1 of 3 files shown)


cassandane/testrunner.pl
```diff
@@ -174,9 +174,11 @@ sub usage
     {
         $cassini_filename = shift;
     }
-    elsif ($a eq '-c' || $a eq '--cleanup')
+    elsif ($a eq '-c' || $a =~ m/^--cleanup(?:=(\w*))?$/)
```
wolfsage, line 177: Not sure about this. It wants `--cleanup=yes` but other args like --config or --jobs just want the param after a space. I'd drop the '=' necessity IMO

wolfsage: I see; it's because you're trying to handle back compat. Bleh. I'd almost prefer a breaking change to introducing this inconsistency...

I suppose another option would be have: `--cleanup`, `--cleanup-pre`, `--cleanup-post`, `--cleanup-no`

elliefm: The entire option parsing situation is already super inconsistent.  There isn't really any sensible pattern to follow, each has its own behaviours.  We should convert it all to GetOpt::Whatever at some point, but my objective in this PR is a small enhancement, not a big refactor.

It's not obvious from the diff (because it didn't change), but there is already a `--no-cleanup`, which made more sense when cleanup was strictly boolean.  I think for most cases cleanup is still mostly boolean, so keeping that consistent is worthwhile. `pre`/`post` seem like very niche use cases.

I also think `no` is kind of an unergonomic default, but the pre-run cleanup deletes entire directories based on filename patterns, and if you had stuff you cared about there for some reason, it would be a rude shock to have it eradicated.  So you only get cleanup if you actually ask for it.

wolfsage (approved):

PR #6055, opened 2026-06-08
rework developer section of docsrc

rjbs:


docsrc/developer/install-xapian.rst
```diff
@@ -1,79 +0,0 @@
...
-    DAEMON {
-      # run a rolling squatter
-      squatter cmd="squatter -R"
-    }
-
-* Enable sync logging: Set ``sync_log: on`` in :cyrusman:`imapd.conf(5)`.
-* Add a squatter sync log channel: ``sync_log_channels: squatter`` in :cyrusman:`imapd.conf(5)`.
-
-You also need (at least one) search tier. Add this to :cyrusman:`imapd.conf(5)`.
-
```
elliefm, line 66: I don't object to removing the "Compiling Xapian" docs, cyruslibs and docker already have that covered.  But I do think we should keep this "Configuring Xapian" section, probably move it out the developer tree though.

On the mailing list, the most common thing anyone says about xapian is like "I couldn't work out how to set it up, so I just used squat instead".  Maybe people would find the documentation more easily if it weren't buried down here.  JMAP requires xapian indexes, so if they can't get xapian working, they can't use JMAP...

Makefile.am
```diff
@@ -1900,9 +1900,8 @@ endif
 # This is a pain because most of us Cyrus devs build with -Werror.
 #
 # Instead of continuing to maintain an ever-growing, increasingly
-# precarious set of post-hoc fixes (see: the former lex-fix rule),
-# let's just ignore sign-comparison warnings from flex-generated
-# sources only.
+# precarious set of post-hoc fixes, let's just ignore sign-comparison
+# warnings from flex-generated sources only.
```
elliefm: If we're no longer mentioning the lex-fix rule, then there is also no longer an ever-growing, increasingly precarious set of post-hoc fixes to contrast this choice against.  So the whole "Instead of..." paragraph could be boiled down to just like, "so we ignore sign-comparison warnings from flex-generated sources" tacked onto the end of the "This is a pain..." sentence.

elliefm (approved):
Cool.  I have a couple of suggestions, but if you agree with and apply them, I don't need to review it again.

docsrc/developer/install-xapian.rst
rjbs: I've bought it back in installing.rst

Makefile.am
rjbs: Done.

PR #6057, opened 2026-06-08
timsieved: make the global variable sieve_dir local to actions_setuser()

dilyanpalauzov:
This makes the file-wide static variable `sieve_dir` local to `actions_setuser()` and the function-wide variable `struct buf buf` available only in a sub-snippet.

### DIFF
--- a/timsieved/actions.c
+++ b/timsieved/actions.c
@@ -39,8 +39,6 @@ extern sieve_interp_t *interp;
 static char *sieve_dir_config = NULL;
 static char *sieved_userid = NULL;
 
-static char *sieve_dir = NULL;
-
 static struct mailbox *sieve_mailbox = NULL;
 static struct sieve_db *sievedb = NULL;
 
@@ -69,12 +67,12 @@ int actions_init(void)
 
 int actions_setuser(const char *userid)
 {
-    struct buf buf = BUF_INITIALIZER;
-    int result;
+    char *sieve_dir;
 
     sieved_userid = xstrdup(userid);
 
     if (sieved_userisadmin) {
+        struct buf buf = BUF_INITIALIZER;
         char *domain = NULL;
 
         buf_setcstr(&buf, sieve_dir_config);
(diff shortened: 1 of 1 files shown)


wolfsage:
Please update the commit message - it should explain what is changing and why it is changing at minimum.

dilyanpalauzov:
The commit message should not describe obvious things and it is subjective.

Irrespective of this, in whatever way the commit message is written, does not mean, that after plenty of time the pending change will be not be anymore applicable, because the integrated source code was changed, and then, if nobody is willing to create the change anew, it will be discarded.

rjbs:
Matthew is a committer who is reviewing your commit and said that the commit message needs to change.  You can ignore his feedback as much as you want, but it's not going to get your commit merged.

You wrote a description of more of the change in the PR, but the commit message is roughly "simplify", which communicates nothing about "simplify how".  If there's a bug found in a year, somebody will have only "simplify" to go on.  Presumably you thought the text was worth writing in the PR description.  Why not worth preserving in the commit?

But more to the point, if the reviewer, who has a commit bit, says they need to see a change before it's merged, complaining that it will just be ignored later anyway is not productive.  Worse, it's likely a self-fulfilling prophesy.  If you are hard to work with, people will avoid working with you.

ksmurchison (approved):

ksmurchison (changes requested):
See comments from wolfsage and rjbs

ksmurchison (approved):

dilyanpalauzov:
https://gitlab.gnome.org/GNOME/evolution/-/merge_requests/67 , https://gitlab.gnome.org/GNOME/evolution-data-server/-/merge_requests/106 , https://gitlab.gnome.org/GNOME/evolution-data-server/-/merge_requests/65 are examples, where I have changed commit messages, after being asked for it.

https://github.com/cyrusimap/cyrus-imapd/pull/4062 is an example, where reviewer with commit bit asked for changes, I did the changes, and then nothing happened.

dilyanpalauzov:
https://github.com/cyrusimap/cyrus-imapd/pull/4059 is another example, where reviewer with commit bit asked for changes…

rjbs:
If you want to say, "These PRs have stalled", that's something we can look into.  Probably the best option is to email the [devel](https://cyrus.topicbox.com/groups/devel) list.

Let's try to clear those up.  But also, please work under the assumption that the Cyrus team is working on its own goals, and with an interest in making the system better over time… but with no interest in interpersonal interactions that make their work unpleasant.  Last year we sat down and cleared out hundreds of open issues, and we're trying to keep on top of what exists, but it's a slog, it's not the primary focus of our work, and we're all really busy.  Your PRs are _very often_ going to rank second in priority compared to the team's internal goals.  That's not malice or disrespect, it's just the structure of our work.  Sometimes you will need to remind us, and sometimes twice.  I would rather this was not true, but changing that is *also* not high on my list of priorities.

The best way to get your open PRs merged is not to refuse to amend other ones, but to bring up the open ones.

PR #6059, opened 2026-06-10
cunit: Use C99 struct initialisers to eliminate HAVE_CU_SETUPFUNC

nwc10:
HAVE_CU_SETUPFUNC was added in Sept 2014 by commit 6f8652b9445a1ef1:
    [cunit] cunit.pl: fix init of CU_SuiteInfo type for CUnit 2.1-3

    The struct CU_SuiteInfo in CUnit has two new members.  This means we
    need two extra NULLs in the initilization.

    This should allow the proper code to be generated, based on what
    autoconf detects

That commit added a probe to configure.ac, which added a macro to config.h. cunit/cunit.pl autogenerates wrapper C stubs for C unit, so it was changed to add code to each to `#include <config.h>`, and to conditionally compile the structure initialiser with 3 or 5 elements in the brace list, depending on the discovered CUnit version.

Arguably we could just ditch all support for CUnit 2.1.1 and earlier, but it's as easy and much clearer to change to C99 designators for the two elements that need to be non-zero. The variable being initialised has static scope, hence the other unmentioned elements will be initialised to zero, which is what we want, and what happened previously (just with more code).


nwc10:
I thought I needed this as part of the `EXPORTED` games in #5950 but ultimately I found I could `git revert` it when the branch was finished, and everything still passed, because I'd misunderstood the cause of my problems.

I think it's still worth doing, but it's actually independent of the hash stuff, so here's a simple PR

elliefm (approved):
Nice find!

PR #6060, opened 2026-06-10
cassandane.ini.dockerfile: Clean up previous test runs on each run

wolfsage:
With this setting change, running tests will leave all passing/failed dirs in /dev/shm/cass so you can inspect data if you like, but the next test run will remove those, when previously they'd never get cleaned up.

This gives a better default setup for most use cases:

 - For ci, nothing really changes.
 - For users with short lived containers, nothing really changes
 - For users with long lived containers, multiple test runs won't fill up /dev/shm/cass since previous runs' data will be cleaned out before each new run.

### DIFF
--- a/cassandane/cassandane.ini.dockertests
+++ b/cassandane/cassandane.ini.dockertests
@@ -42,7 +42,7 @@
 
 [cassandane]
 rootdir = /dev/shm/cass
-cleanup = no
+cleanup = pre
 
 # If we need do, we can put a list of (space-delimited) tests in this setting,
 # and those tests will be skipped.  In the past, we've used it for tests that


wolfsage:
I've added you all as reviewers to make sure you're happy with/aware of these changes.

This will mean that we need to backport the cassandane support of `cleanup = pre` to any old branches we keep alive and run CI on.

ksmurchison (approved):
WFM

elliefm:
> This will mean that we need to backport the cassandane support of `cleanup = pre` to any old branches we keep alive and run CI on.

I think that's only the case if we backport this PR to those branches.  If we don't, then CI will continue using `cleanup = no` based on the cassandane.ini.dockertests in the branch being tested, and cassandane won't need to know about `pre`.  That is to say: if we backport this we have to backport both, but we don't have to backport this.  Have I missed something?

**Backporting:** if you backport this, you'll also need #6048.

I don't think we'd backport either of these.  The only meaningful change is for long-lived containers, which are only used during active development, which doesn't really happen on the stable branches.

elliefm (approved):

wolfsage:
> > This will mean that we need to backport the cassandane support of `cleanup = pre` to any old branches we keep alive and run CI on.
> 
> I think that's only the case if we backport this PR to those branches. If we don't, then CI will continue using `cleanup = no` based on the cassandane.ini.dockertests in the branch being tested, and cassandane won't need to know about `pre`. That is to say: if we backport this we have to backport both, but we don't have to backport this. Have I missed something?
> 

Ah, right I confused myself once again about where/how ci runs for those branches. Thanks

rjbs:
> I've added you all as reviewers to make sure you're happy with/aware of these changes.

You added us all as *assignees*, I think.   Oops?  Anyway, LGTM!

PR #6062, opened 2026-06-10
jmap_api.h: do not require JACL_SETSEEN for JACL_WRITEALL

rjbs:
JACL_WRITEALL is only use in DAV-like datatypes, none of which use JACL_SETSEEN for anything.  In a recent commit, we stopped adding ACL_SETSEEN to the calendar homeset of new users, so new calendars didn't inherit it, so they weren't WRITEALL, so we reported that they could not write to the calendar.

We'd *let* them, because they had admin, but clients that took the myRights advice to heart would not offer the option to write.

So:  stop making JACL_SETSEEN part of JACL_WRITEALL, and assert that a new calendar is mayWriteAll (etc) for its owner.  No existing test caught this!

### DIFF
--- a/cassandane/tiny-tests/JMAPCalendars/calendar_set
+++ b/cassandane/tiny-tests/JMAPCalendars/calendar_set
@@ -24,6 +24,16 @@ sub test_calendar_set
     $self->assert_not_null($res->[0][1]{created});
     $self->assert_not_null($res->[0][1]{created}{1}{mailboxUniqueId});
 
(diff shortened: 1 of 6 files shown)


ksmurchison (approved):
LGTM provided CI passes this time

rjbs:
I had to go read through some code paths to understand why these things changed, but now I am comfortable with it.

ksmurchison (approved):

PR #6064, opened 2026-06-11
http_jmap.c: escape JMAP logHeaders values before logging

rjbs:
Cyrus's JMAP-WS Request object can have an extra, non-standard property, logHeaders, which is headers to consider syslogging.  The old code passed the JSON-string value straight through buf_printf with %s, so a client can inject CRLF and forge what looks like a separate log entry. The httpd_log_headers whitelist constrains the header *name*, but the *value* is fully client-controlled.

Escape ASCII control characters as \xNN and backslash-escape " and \ before logging. (This code should probably all be converted to logfmt.)

Reported by Ahmed Said.

### DIFF
--- a/cassandane/tiny-tests/JMAPCore/loginject_via_logheaders
+++ b/cassandane/tiny-tests/JMAPCore/loginject_via_logheaders
@@ -1,0 +1,51 @@
+#!perl
+use Cassandane::Tiny;
+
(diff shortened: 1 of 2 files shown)


rjbs:
@ksmurchison This is one you approved in another form, just FYI.

ksmurchison (approved):

PR #6066, opened 2026-06-12
jmap_ical.c: end _vanew function with NULL

dilyanpalauzov:
All other …_vanew() function invocations end with `NULL` after https://github.com/cyrusimap/cyrus-imapd/pull/4278.  Using `(void *)0)` instead of `NULL` is correct, but inconsistent.


wolfsage:
> All other …_vanew() function invocations end with `NULL` after #4278. Using `(void *)0)` instead of `NULL` is correct, but inconsistent.

This would be more useful in the commit message than as a comment on the MR.

ksmurchison (approved):

ksmurchison:
repair_broken_ical() has been removed

PR #6068, opened 2026-06-12
index.c:index_fetch() Set value of `struct vanished_params v`  only once

dilyanpalauzov:
Before this change most fields in `struct vanished_params v` were first set to zero and immediately afterwards to some other value.


ksmurchison (approved):

rjbs:
Sorry for the weird noise.  Our "automatically rebase and merge" automation, which worked fine in some other PRs, is not working here. I don't have a diagnosis yet.

ksmurchison (approved):

PR #6073, opened 2026-06-15
Enforce geteuid()≠0 in some user-space tools

dilyanpalauzov:
The changed files are supposed to be run after version upgrade. Staying as the cyrus-user is error-prone, so these additions enforce changing the user.

### DIFF
--- a/imap/ctl_conversationsdb.c
+++ b/imap/ctl_conversationsdb.c
@@ -1212,6 +1212,8 @@ int main(int argc, char **argv)
         }
     }
 
+    if (geteuid() == 0) fatal("must run as the Cyrus user", EX_USAGE);
+
     if (mode == UNKNOWN)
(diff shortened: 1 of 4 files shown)


elliefm (approved):
Thanks, this looks good

elliefm:
@rjbs @wolfsage Can I get a second opinion?  This looks correct to me, and it should not have any particular ramifications, because you're supposed to run cyrus tools as the cyrus user anyway.  This just enforces that.  But I don't know what FM's tooling does, and if we're doing something silly like running this stuff as root, the change might break us.  What do you think?

(IMO: if this change breaks FM stuff, we need to fix the FM stuff.)

rjbs:
I would expect this to be safe.

rjbs (approved):

PR #6080, opened 2026-06-18
imapd.c: verify that UIDBATCHES batch size and range limits are non-zero

ksmurchison:
```
message-batches     = "UIDBATCHES" SP nz-number
                       [SP nz-number ":" nz-number]
```


rsto (approved):
I would have preferred the commit message to be a proper sentence, but I figured it out from the commit context. I notice that both the new comment and the existing commit for the batch range refer to Section, but do not indicate what document that section is in. I had to google for the UIDBATCHES IETF I-D to figure out which document is meant.

rjbs (approved):

PR #6081, opened 2026-06-19
imap: fix UID MOVE cross-backend in murder proxy

### LINKED ISSUE #51: [2.5.9] murder MOVE does COPY if backends differ
If I delete an E-Mail in Thunderbird and the source mailbox is a shared one on a different backend telemetry log shows eg:

```
<1475840848<22 uid move 22150 "INBOX.Trash"^M
>1475840848>22 OK [COPYUID 1123762451 766568] Completed^M
```

... and in fact the message was copied and reappears in the source mailbox.

### LINKED ISSUE #6079: MURDER: UID move to shared mailbox on different backend makes a copy
I'm not sure how much of this is related to  #4027.

In our cyrus imapd 3.12.1, issuing an IMAP move for a message in a folder on one backend to a shared folder (e.g. `Other Users.other`) makes the message appear in both mailboxes.

Following the logs, the message is `APPEND`ed to the destination mailbox but never expunged from the source.

Setting `suppress_capabilities: MOVE XMOVE` makes the client issue separate `COPY` and `EXPUNGE` operations.

jcdelepine:
Closes #51
Closes #6079

---

In a Cyrus murder cluster, a `UID MOVE` where source and destination live on different backends was silently degraded into a `COPY`: the message was appended to the destination but never expunged from the source.

Root cause: `cmd_copy()` calls `proxy_copy()` without the `ismove` flag, so no post-APPEND expunge was ever issued.

---

Implements the approach outlined by @elliefm in #51:

> 1. Check `ACL_EXPUNGE` on the source before proceeding.
> 2. `proxy_copy()` should return a status, not void.
> 3. On success, expunge the source messages.

See inline code comments for implementation details.

---

Two new tiny-tests under `cassandane/tiny-tests/MurderIMAP/`:

- `move_shared_across_backends`: cross-backend MOVE succeeds, source is verified empty.
- `move_shared_across_backends_no_expunge_acl`: MOVE is rejected with `NO` when the user lacks the `e` ACL right; source is untouched.

Full suite: `OK (2633 tests)`

---

- **No atomicity**: if `APPEND` succeeds but `UID EXPUNGE` fails, the message exists on both backends. The `int` return type of `proxy_copy()` is plumbing for a future compensation path.
- **Stale frontend state**: `pipe_until_tag_discard` suppresses `EXPUNGE` notifications, leaving the frontend's backend connection with a stale message count until the next `SELECT`. Real IMAP clients are unaffected (they track moved messages locally). See also the comment in the Cassandane tests.

---

- Applies cleanly on `master` @ f4e7cc15e with `patch -p1`.
- Backports to cyrus-imapd-3.10 and 3.12 can be provided as follow-up once this is accepted on master. A backport to cyrus-imapd-3.8 is planned; manual adaptation will be needed.
- No configuration change required.

### DIFF
--- a/cassandane/tiny-tests/MurderIMAP/move_shared_across_backends_no_expunge_acl
+++ b/cassandane/tiny-tests/MurderIMAP/move_shared_across_backends_no_expunge_acl
@@ -1,0 +1,36 @@
+#!perl
+use Cassandane::Tiny;
+sub test_move_shared_across_backends_no_expunge_acl
+    :NoAltNamespace :IMAPMurder :needs_component_murder
+    ($self)
+{
+    my $shared = 'shared';
+
+    # create a shared folder on backend2
+    my $admintalk = $self->{backend2_adminstore}->get_client();
+    $admintalk->create($shared);
+    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());
+    # grant rights without expunge ('e' is missing)
+    $admintalk->setacl($shared, 'anyone', 'lrswipkxtcd');
+    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());
+
+    # inject a message into the shared folder
+    my $frontend = $self->{frontend_store}->get_client();
+    $self->{frontend_store}->set_folder($shared);
+    $self->{frontend_store}->_select();
+    $self->{gen}->set_next_uid(1);
+    $self->make_message("Message A", store => $self->{frontend_store});
+
+    # verify the message is there
+    $frontend->select($shared);
+    $self->assert_str_equals('ok', $frontend->get_last_completion_response());
+    $self->assert_num_equals(1, $frontend->get_response_code('exists'));
+
+    # MOVE to INBOX (cross-backend) should fail without expunge right
+    $frontend->move('1', 'INBOX');
+    $self->assert_matches(qr{Permission denied}, $frontend->get_last_error());
+
+    # verify the message is still in the source
+    $frontend->select($shared);
+    $self->assert_num_equals(1, $frontend->get_response_code('exists'));
+}
--- a/imap/imap_proxy.c
+++ b/imap/imap_proxy.c
@@ -133,6 +133,7 @@ struct backend *proxy_findinboxserver(const char *userid)
    force_notfatal says to not fatal() if we lose connection to backend_current
    even though it is in 95% of the cases, a good idea...
 */
+
 static int pipe_response(struct backend *s, const char *tag, int include_last,
                          unsigned flags)
 {
@@ -970,13 +971,14 @@ static void copy_cb(uint32_t seqno, unsigned item, void *datap, void *rock)
     }
 }
 
-void proxy_copy(const char *tag, char *sequence, char *name, int myrights,
-                int usinguid, struct backend *s)
+int proxy_copy(const char *tag, char *sequence, char *name, int myrights,
+                int usinguid, struct backend *s, int ismove)
 {
     struct copy_rock crock = { HASHU64_TABLE_INITIALIZER, { 0 }, s, NULL };
     unsigned items = FETCH_FLAGS | FETCH_INTERNALDATE;
     char *freeme = NULL;
     int res;
+    int ret = PROXY_NO;
 
     construct_hashu64_table(&crock.msg_byuid, 1024, 0);
 
@@ -1014,6 +1016,38 @@ void proxy_copy(const char *tag, char *sequence, char *name, int myrights,
         res = pipe_until_tag(s, tag, 0);
 
         if (res == PROXY_OK) {
+            /* expseq is freeme when set (UID set fetched from the source via FETCH
+             * callbacks), otherwise the original sequence argument.  Either way it
+             * refers to messages on backend_current, the source backend. */
+            if (ismove) {
+                char movetag[128];
+                const char *expseq = freeme ? freeme : sequence;
+
+                /* mark \Deleted on source backend */
+                proxy_gentag(movetag, sizeof(movetag));
+                prot_printf(backend_current->out,
+                            "%s UID Store %s +FLAGS.SILENT (\\Deleted)\r\n",
+                            movetag, expseq);
+                if (pipe_until_tag(backend_current, movetag, 0) != PROXY_OK) {
+                    syslog(LOG_ERR, "MOVE proxy: UID STORE \\Deleted failed on source "
+                                    "after successful APPEND, expunge will be skipped "
+                                    "(user: %s)",
+                                    imapd_userid);
+                }
+
+                /* UID EXPUNGE (RFC 4315, always available on Cyrus backends) */
+                proxy_gentag(movetag, sizeof(movetag));
+                prot_printf(backend_current->out,
+                            "%s UID Expunge %s\r\n",
+                            movetag, expseq);
+                if (pipe_until_tag(backend_current, movetag, 0) != PROXY_OK) {
+                     syslog(LOG_ERR, "MOVE proxy: UID EXPUNGE failed on source after "
+                                     "successful APPEND, message may be duplicated "
+                                     "(user: %s)",
+                                     imapd_userid);
+                }
+            }
+
             if (myrights & ACL_READ) {
                 appenduid = strchr(s->last_result.s, '[');
                 /* skip over APPENDUID */
@@ -1031,6 +1065,7 @@ void proxy_copy(const char *tag, char *sequence, char *name, int myrights,
                 prot_printf(imapd_out, "%s OK %s\r\n", tag,
                             error_message(IMAP_OK_COMPLETED));
             }
+            ret = PROXY_OK;
         } else {
             prot_printf(imapd_out, "%s %s", tag, s->last_result.s);
         }
@@ -1047,6 +1082,7 @@ void proxy_copy(const char *tag, char *sequence, char *name, int myrights,
     free_hashu64_table(&crock.msg_byuid, NULL);
     seqset_free(&crock.uidset);
     free(freeme);
+    return ret;
 }
 
 int proxy_fetch(char *sequence, int usinguid, unsigned items,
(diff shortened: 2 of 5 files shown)


cassandane/tiny-tests/MurderIMAP/move_shared_across_backends_no_expunge_acl
```diff
@@ -0,0 +1,36 @@
...
+    $self->make_message("Message A", store => $self->{frontend_store});
+
+    # verify the message is there
+    $frontend->select($shared);
+    $self->assert_str_equals('ok', $frontend->get_last_completion_response());
+    $self->assert_num_equals(1, $frontend->get_response_code('exists'));
+
+    # MOVE to INBOX (cross-backend) should fail without expunge right
+    $frontend->move('1', 'INBOX');
+    $self->assert_str_equals('no', $frontend->get_last_completion_response());
```
elliefm: You should probably also check to make sure this `no` was due to "Permission denied" and not some other error.  Something like:

```
$self->assert_matches(qr{Permission denied}, $frontend->get_last_error());
```

imap/imap_proxy.c
```diff
@@ -992,6 +1016,34 @@ void proxy_copy(const char *tag, char *sequence, char *name, int myrights,
...
+
+                /* mark \Deleted on source backend */
+                proxy_gentag(movetag, sizeof(movetag));
+                prot_printf(backend_current->out,
+                            "%s UID Store %s +FLAGS (\\Deleted)\r\n",
+                            movetag, expseq);
+                if (pipe_until_tag_discard(backend_current, movetag, 0) != PROXY_OK)
+                    syslog(LOG_ERR, "MOVE proxy: UID STORE \\Deleted failed on source "
+                    "after successful APPEND, expunge will be skipped "
+                    "(user: %s)", imapd_userid);
```
elliefm: Please use `{}` around multi-line blocks, even if they're logically only a single statement.

Please also indent the wrapped syslog arguments to match our style.

When concatenating string literals like this, I would suggest putting the `" "` between the lines at the start of the new line, rather than the end of the previous line.  This is visually a little odd, which aids readability by distinguishing it from a comma-separated list of strings.  I would also put the last argument on its own line, so that it's visually distinct from the wrapped concatenated literals.  Something like this:

```suggestion
                if (pipe_until_tag_discard(backend_current, movetag, 0) != PROXY_OK) {
                    syslog(LOG_ERR, "MOVE proxy: UID STORE \\Deleted failed on source"
                                    " after successful APPEND, expunge will be skipped"
                                    " (user: %s)",
                                    imapd_userid);
                }
```

I see you've used a similar pattern elsewhere.  This same advice applies to those places too.

A different way you could approach this would be to concatenate the literals into a temporary.  That would clearly distinguish them from the normal arguments.  Might produce warnings about the format argument not being a literal though.  Something like this:

```suggestion
                if (pipe_until_tag_discard(backend_current, movetag, 0) != PROXY_OK) {
                    const char *const fmt = "MOVE proxy: UID STORE \\Deleted failed on source"
                                            " after successful APPEND, expunge will be skipped"
                                            " (user: %s)";
                    syslog(LOG_ERR, fmt, imapd_userid);
                }
```

imap/imap_proxy.c
```diff
@@ -269,6 +269,29 @@ int pipe_until_tag(struct backend *s, const char *tag, int force_notfatal)
...
+ * discarding them without relaying to the client.
+ *
+ * Used after a cross-backend MOVE: the tagged OK for the destination APPEND
+ * has already been sent to the client, so any subsequent FETCH or EXPUNGE
+ * notifications from the source backend would arrive out-of-context.
+ *
+ * WARNING: non-reentrant -- temporarily swaps the process-global imapd_out.
+ * Must not be called from a context where fatal() could fire mid-call, as
+ * that would leave imapd_out pointing at the discard buffer. */
+int pipe_until_tag_discard(struct backend *s, const char *tag, int force_notfatal)
```
elliefm: I might call this function `discard_until_tag` rather than `pipe_until_tag_discard`.  If it's being discarded, it's not being piped, right?  So the verb "pipe" is misleading.

imap/imap_proxy.c
```diff
@@ -269,6 +269,29 @@ int pipe_until_tag(struct backend *s, const char *tag, int force_notfatal)
...
+{
+    struct buf discard = BUF_INITIALIZER;
+    struct protstream *discard_out = prot_writebuf(&discard);
+    struct protstream *saved_out = imapd_out;
+    imapd_out = discard_out;
+    int r = pipe_response(s, tag, 0, force_notfatal);
+    imapd_out = saved_out;
+    prot_free(discard_out);
+    buf_free(&discard);
+    return r;
```
elliefm: I don't love this implementation.  Since we just want to discard these lines, surely there's a way to do that without messing with the process state?  The edge cases and failure modes are really hard to reason about.

Ah, I see that all the real logic is in `pipe_response()`, and you don't want to duplicate that...

What if you added a `discard` flag to the `pipe_response()` arguments, and have it skip the `prot_write()` calls when that flag is set?  The other callers would need to be updated to set the new flag to false, and this caller would set it to true.  Then we wouldn't need to mess with `imapd_out`.  But maybe you tried something like that already, and it doesn't work for some reason I'm not seeing?

elliefm (commented):
Thanks, this looks promising.  I have some notes and suggestions.

elliefm:
@ksmurchison Can you take a look at this one too please?  It looks okay to me, but I don't know the dark corners of the proxy code, maybe there's gotchas I haven't noticed because I don't know to look for them.  Cheers

imap/imap_proxy.c
ksmurchison: I agree with this suggestion.

imap/imap_proxy.c
ksmurchison: What I might do is change the `int force_notfatal` argument to `unsigned flags`, then

```
#define PIPE_FLAG_FORCE_NOTFATAL (1<<0)
#define PIPE_FLAG_DISCARD                 (1<<1)
```
then:

`unsigned force_notfatal = flags & PIPE_FLAG_FORCE_NOTFATAL;`

Existing calls of pipe_until_flag() and pipe_including_tag() should work without change.

imap/imap_proxy.c
ksmurchison: Is there a reason why we do not want the unsolicited EXPUNGE responses?

ksmurchison (changes requested):
By and large this looks good pending the suggested code changes.
Thank you very much for this work!

cassandane/tiny-tests/MurderIMAP/move_shared_across_backends_no_expunge_acl
jcdelepine: done.

imap/imap_proxy.c
jcdelepine: done

imap/imap_proxy.c
jcdelepine: The current commit implements @elliefm's suggestion. I'll update it to use @ksmurchison's `flags` approach instead.

jcdelepine:
All review comments addressed, force-pushed. Full Cassandane suite still passes (2633 tests).

imap/imap_proxy.c
jcdelepine: To answer this question, I simply changed those `discard_until_tag` calls back to `pipe_until_tag`. Surprisingly, the tests still pass, even though the discard was previously needed to make them pass. I will try to understand why.

jcdelepine: After adding some logging, I can see that the `* 1 EXPUNGE` notification is indeed forwarded to the client after the MOVE. However, `Mail::IMAPTalk` does not update its internal `EXISTS` counter in response to an unsolicited EXPUNGE received during a MOVE, which is why the frontend-side assertion was failing. The server behaviour is correct.

The `discard_until_tag` calls were added during development in an attempt to fix that failing assertion, but without success. The real fix was to verify the expunge directly on backend2 rather than through the frontend — at which point the discard was no longer needed but was not removed.

I will push a new patch removing the discard logic entirely; `pipe_until_tag` works fine.

imap/imap_proxy.c
ksmurchison: I might still discard the FETCH responses. MOVE on a single server setup will return EXPUNGE but not FETCH.

jcdelepine:
Quoting Ken Murchison ***@***.***>:

> @ksmurchison commented on this pull request.
>
>
>
>> +    struct buf discard = BUF_INITIALIZER;
> +    struct protstream *discard_out = prot_writebuf(&discard);
> +    struct protstream *saved_out = imapd_out;
> +    imapd_out = discard_out;
> +    int r = pipe_response(s, tag, 0, force_notfatal);
> +    imapd_out = saved_out;
> +    prot_free(discard_out);
> +    buf_free(&discard);
> +    return r;
>
> I might still discard the FETCH responses. MOVE on a single server  
> setup will return EXPUNGE but not FETCH.

RFC 3501 requires clients to handle unsolicited responses at any time,  
so the `FETCH` should not be a problem for conformant clients. That  
said, it can be avoided by using `+FLAGS.SILENT` instead of `+FLAGS`  
in the `UID STORE` command.

With `+FLAGS.SILENT`, the `FETCH` disappears and the `EXPUNGE` is  
correctly forwarded to the client, as required by RFC 6851:

```
S: * 1 EXPUNGE
S: 8 OK [COPYUID 1782214524 1] Completed
```

All tests pass. Verification on backend2 directly is still needed due  
to `Mail::IMAPTalk` not updating its internal `EXISTS` counter in  
response to an unsolicited `EXPUNGE` received during a MOVE.
Force-pushed.




-- 
Service systèmes et réseaux - DISI

Université de Picardie Jules Verne
5, rue du moulin neuf - 80000 Amiens

ksmurchison (approved):

jcdelepine:
Thank you both for the thorough and constructive review — it has significantly improved the patch.

PR #6082, opened 2026-06-19
Sieve addannotation

brong:
This adds an "addannotation" command like "addheader" to sieve.  Maybe we should write it up as a spec.

I'm thinking this might be a useful tool for adding delivery data to messages without having to adulterate them with headers during the sieve phase.

### DIFF
--- a/sieve/sieve.y
+++ b/sieve/sieve.y
@@ -77,6 +77,8 @@ static commandlist_t *build_addheader(sieve_script_t*, commandlist_t *c,
                                       char *name, char *value);
 static commandlist_t *build_deleteheader(sieve_script_t*, commandlist_t *c,
                                          char *name, strarray_t *values);
+static commandlist_t *build_addannotation(sieve_script_t*, commandlist_t *c,
+                                          char *entry, char *value);
 static commandlist_t *build_log(sieve_script_t*, char *text);
 static commandlist_t *build_snooze(sieve_script_t *sscript,
                                    commandlist_t *c, arrayu64_t *times);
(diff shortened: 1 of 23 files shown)


rjbs:
Haven't really fully reviewed (and you have test failures), but I've got my hackles up about ACL lately so looked into it and there are some access control bugs here.  This allows adding annotations to shared folders lacking +n.  Also, allows the user to write to system annotations.  I had Claude produce tests for this which I believe are correct.  Attached.

[addannotation_acl_bypass.txt](https://github.com/user-attachments/files/29154494/addannotation_acl_bypass.txt)

[addannotation_internal_namespace.txt](https://github.com/user-attachments/files/29154495/addannotation_internal_namespace.txt)

sieve/sieve.y
```diff
@@ -1032,6 +1041,13 @@ ahtags: /* empty */              { $$ = new_command(B_ADDHEADER, sscript); }
         ;
 
 
+/* ADDANNOTATION tagged arguments */
+aantags: /* empty */             { $$ = new_command(B_ADDANNOTATION, sscript); }
+        | aantags PRIV           { $$ = $1; $$->u.aan.scope = 1; }
+        | aantags SHARED         { $$ = $1; $$->u.aan.scope = 0; }
```
ksmurchison, line 1047: If shared is the default, rather than two separate tagged args, why not just have :priv?  

And then the aan struct in tree.h can have `bool priv`

rjbs:
We already have a plan for header adulteration, for rules-application tracking, and I think it'll be fine without introducing this.  I didn't see a response to my concern about access control, and we spoke out loud about the "annotation on which instance" problem for history tracking.  So all things being equal, I'd rather not try to carefully review something I don't quite grasp yet, when we don't have a clear use case, and we have lots of other stuff in flight!

dilyanpalauzov:
Cyrus specific Sieve extensions are supposed to be documented at https://www.cyrusimap.org/dev/reference/admin/sieve.html#cyrus-imap-specific-extensions and at https://www.cyrusimap.org/dev/reference/extensions/sieve.html .  If the latter files are adjusted, it is not necessary to describe the functionality in detail in a changes/next/ file.

PR #6087, opened 2026-06-23
index.c: FETCH 1:* (VANISHED) has to look at ALL messages

ksmurchison:
Expunged messages may have UIDs greater than the last visible message. Without this fix, if the last UID were expunged, it would not be reported as VANISHED.


rsto (approved):
LGTM - instead of adding the issue number to the commit title, you can also write "Fixes #XXX" in the commit description, which will make Github close the issue when merged.

ksmurchison:
> LGTM - instead of adding the issue number to the commit title, you can also write "Fixes #XXX" in the commit description, which will make Github close the issue when merged.

TIL

PR #6088, opened 2026-06-23
xcal.c: avoid NULL dereference in xml_element_to_icalcomponent()

ksmurchison:


rsto (approved):
LGTM - also could move the issue number from the title to the commit description

PR #6091, opened 2026-06-25
jmap_contact.c: derive FN from Card properties instead of "No Name"

rsto:
When a Card has no explicit Name.full, the FN property of the stored vCard was hard-coded to "No Name". Instead, derive it from Card properties in decreasing order of precedence:

  1. Name.full
  2. the concatenated Name component values
  3. first non-empty organizations.name
  4. first non-empty nicknames.name
  5. first non-empty emails.address
  6. first non-empty phones.number
  7. the uid
  8. "No Name"


ksmurchison (approved):
LGTM

PR #6096, opened 2026-06-25
gitignore: update for the new location of cyradm

rjbs:
bb37546d1de moved cyradm from user to system commands, but the gitignore wasn't updated for it.


elliefm (commented):
The referenced commit moved cyradm.rst _from_ systemcommands _to_ usercommands, and updated the gitignore file accordingly.  So I'm confused about what this commit is trying to do.  Wire crossed somewhere?

Ahh, I see -- what really happened was that I failed to update the stupid bogus `docsrc/Makefile`, so it still builds cyradm in "systemcommands", when it should be "usercommands".

PR #6097, opened 2026-06-26
make http, caldav, carddav, etc. mandatory

rjbs:
As long discussed, this is another step along the "fewer build configurations" path.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -192,7 +192,7 @@ libexec_PROGRAMS += \
     imap/lmtpd \
     imap/pop3d \
     imap/backupcyrusd \
-		imap/calalarmd \
+    imap/calalarmd \
     imap/promstatsd \
     imap/smmapd \
     ptclient/ptloader
--- a/Makefile.am
+++ b/Makefile.am
@@ -1008,6 +1008,14 @@ imap_libcyrus_imap_la_SOURCES = \
     imap/auditlog.h \
     imap/backend.c \
     imap/backend.h \
+    imap/caldav_alarm.c \
+    imap/caldav_alarm.h \
+    imap/caldav_db.c \
+    imap/caldav_db.h \
+    imap/calsched_support.c \
+    imap/calsched_support.h \
+    imap/carddav_db.c \
+    imap/carddav_db.h \
     imap/conversations.c \
     imap/conversations.h \
     imap/convert_code.c \
@@ -1016,6 +1024,10 @@ imap_libcyrus_imap_la_SOURCES = \
     imap/css3_color.h \
     imap/dav_db.c \
     imap/dav_db.h \
+    imap/dav_util.c \
+    imap/dav_util.h \
+    imap/defaultalarms.c \
+    imap/defaultalarms.h \
     imap/dlist.c \
     imap/dlist.h \
     imap/duplicate.c \
@@ -1026,16 +1038,20 @@ imap_libcyrus_imap_la_SOURCES = \
     imap/haproxy.h \
     imap/http_client.c \
     imap/http_client.h \
+    imap/ical_support.c \
+    imap/ical_support.h \
+    imap/icu_wrap.cpp \
+    imap/icu_wrap.h \
     imap/idle.c \
     imap/idle.h \
     imap/idlemsg.c \
     imap/idlemsg.h \
     imap/imapparse.c \
     imap/imapparse.h \
-    imap/index.c \
-    imap/index.h \
     imap/index_file.c \
     imap/index_file.h \
+    imap/index.c \
+    imap/index.h \
     imap/jmap_util.c \
     imap/jmap_util.h \
     imap/json_support.c \
@@ -1046,18 +1062,18 @@ imap_libcyrus_imap_la_SOURCES = \
     imap/mailbox.h \
     imap/mbdump.c \
     imap/mbdump.h \
+    imap/mboxevent.c \
+    imap/mboxevent.h \
     imap/mboxkey.c \
     imap/mboxkey.h \
     imap/mboxlist.c \
     imap/mboxlist.h \
-    imap/mboxevent.c \
-    imap/mboxevent.h \
     imap/mboxname.c \
     imap/mboxname.h \
     imap/message_guid.c \
+    imap/message_priv.h \
     imap/message.c \
     imap/message.h \
-    imap/message_priv.h \
     imap/msgrecord.c \
     imap/msgrecord.h \
     imap/mupdate-client.c \
@@ -1079,20 +1095,20 @@ imap_libcyrus_imap_la_SOURCES = \
     imap/search_engines.h \
     imap/search_expr.c \
     imap/search_expr.h \
+    imap/search_part.h \
     imap/search_query.c \
     imap/search_query.h \
-    imap/search_part.h \
     imap/search_sort.h \
-    imap/seen.h \
     imap/seen_db.c \
+    imap/seen.h \
     imap/sievedir.c \
     imap/sievedir.h \
     imap/smtpclient.c \
     imap/smtpclient.h \
     imap/spool.c \
     imap/spool.h \
-    imap/statuscache.h \
     imap/statuscache_db.c \
+    imap/statuscache.h \
     imap/sync_log.c \
     imap/sync_log.h \
     imap/telemetry.c \
@@ -1103,15 +1119,18 @@ imap_libcyrus_imap_la_SOURCES = \
     imap/user.h \
     imap/userdeny_db.c \
     imap/userdeny.h \
+    imap/vcard_support.c \
+    imap/vcard_support.h \
     imap/version.c \
-    imap/version.h
+    imap/version.h \
+    imap/webdav_db.c \
+    imap/webdav_db.h
 
 if SIEVE
 imap_libcyrus_imap_la_SOURCES += \
     imap/sieve_db.c \
     imap/sieve_db.h
 endif # SIEVE
-
 
 if USE_SQUAT
 imap_libcyrus_imap_la_SOURCES += \
@@ -1122,28 +1141,6 @@ imap_libcyrus_imap_la_SOURCES += \
     imap/squat_internal.c \
     imap/squat_internal.h
 endif
-
-imap_libcyrus_imap_la_SOURCES += \
-    imap/caldav_alarm.c \
-    imap/caldav_alarm.h \
-    imap/caldav_db.c \
-    imap/caldav_db.h \
-    imap/carddav_db.c \
-    imap/carddav_db.h \
-    imap/dav_util.c \
-    imap/dav_util.h \
-    imap/defaultalarms.c \
-    imap/defaultalarms.h \
-    imap/ical_support.c \
-    imap/ical_support.h \
-    imap/icu_wrap.h \
-    imap/icu_wrap.cpp \
-    imap/calsched_support.c \
-    imap/calsched_support.h \
-    imap/vcard_support.c \
-    imap/vcard_support.h \
-    imap/webdav_db.c \
-    imap/webdav_db.h
 
 if CUNIT
 
--- a/Makefile.am
+++ b/Makefile.am
@@ -192,6 +192,7 @@ libexec_PROGRAMS += \
     imap/lmtpd \
     imap/pop3d \
     imap/backupcyrusd \
+		imap/calalarmd \
     imap/promstatsd \
     imap/smmapd \
     ptclient/ptloader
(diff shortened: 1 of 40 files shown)


changes/next/cyr-1715-mandatory-dav
```diff
@@ -0,0 +1,39 @@
...
+
+
+Config changes:
+
+None.
+
+
+Upgrade instructions:
+
+If you build Cyrus IMAP from source, make sure libical (>= 3.99.99),
```
elliefm: Maybe we should resolve the situation with libical 4.0 before landing this, so that the upgrade instructions can reference a version that actually exists.  Or just remove the version reference here, and let the requirement for a new enough version be implied like it is for the other dependencies.

n.b. #6032 proposes bumping the minimum libical version to 4.0, and Robert confirmed in a comment that we do in fact support libical 4.0 properly these days.  It languishes because I have not yet had a chance to think carefully about what else (if anything) needs to change along with the configure.ac change.

configure.ac
```diff
@@ -1601,18 +1580,8 @@ AS_IF([test "x$with_zoneinfo_dir" != "x"],
...
-AC_ARG_ENABLE(calalarmd,
-        [AS_HELP_STRING([--enable-calalarmd], [enable CalDAV alarm support])],,[enable_calalarmd="no";])
-if test "x$enable_calalarmd" = "xyes"; then
-    if test "x$enable_http" = "xyes"; then
-    AC_DEFINE(USE_CALALARMD,[],[Build with calendar alarm functionality])
-    else
-    AC_MSG_ERROR([--enable-calalarmd requires --enable-http])
-    fi
-fi
-AM_CONDITIONAL([CALALARMD], [test "$enable_calalarmd" != "no"])
```
elliefm: If the whole block's gone we don't need the comment anymore... especially not with the wrong version in it 😅

Makefile.am
```diff
@@ -192,6 +192,7 @@ libexec_PROGRAMS += \
     imap/lmtpd \
     imap/pop3d \
     imap/backupcyrusd \
+		imap/calalarmd \
     imap/promstatsd \
```
elliefm, line 196: Wrong indentation

cassandane/tiny-tests/ALPN/https_bad
```diff
@@ -2,7 +2,7 @@
 use Cassandane::Tiny;
 
 sub test_https_bad
-    :want_service_https :needs_component_httpd
+    :want_service_https
     ($self)
 {
```
elliefm, line 7: I don't think I would rush to remove these annotations.  It doesn't hurt anything for them to stay there, and for cross version testing, we'll still need to know whether the feature is present in the other build.  

(This specific test doesn't use the other build, so it doesn't matter here; but can you say the same for all the other tests this annotation was removed from?)

configure.ac
```diff
@@ -2403,13 +2369,13 @@ Cyrus Server configured components
    gssapi:             $gssapi
    autocreate:         $enable_autocreate
    idled:              $enable_idled
-   httpd:              $enable_http
+   httpd:              yes
```
dilyanpalauzov: The final section of configure.ac is supposed to be discovery/configuration results, not for unconditional output.  So this line can be removed.  Likewise for libical, sqlite3 and libxml2 below - if they are required, there is no need to spell out, if they were discovered.

doc/examples/cyrus_conf/normal.conf
```diff
@@ -60,4 +61,7 @@ EVENTS {
 DAEMON {
   # this is only necessary if using idled for IMAP IDLE
 #  idled         cmd="idled"
+
+  # required for CalDAV / JMAP-Calendars alarms to fire
```
dilyanpalauzov, line 65: This comment is undoubtedly correct, but at the same time there is no documentation for `calalarmd` - what it does, why does it exist.  https://www.cyrusimap.org/imap/reference/manpages/systemcommands/calalarmd.html is stub - where and how are the alarms sent.

docsrc/download/installation/manage-dav.rst
```diff
@@ -11,6 +11,16 @@ This assumes you already have the relevant modules enabled in your Cyrus
...
 manual installation.
 
+.. note::
+
+    As of Cyrus IMAP 3.13.5, HTTP support, CalDAV, CardDAV, WebDAV, and the
+    calendar alarm daemon (``calalarmd``) are mandatory parts of every
+    build; libical, libxml2, and SQLite3 are required build dependencies.
+    The former ``--enable-http`` and ``--enable-calalarmd`` configure
+    options have been removed. Selecting which HTTP modules a given
+    ``httpd`` instance serves continues to be controlled at runtime by the
```
dilyanpalauzov, line 21: I think the last sentence is not necessary: it is obvious, nothing has changed.

Makefile.am
```diff
@@ -1127,7 +1117,6 @@ imap_libcyrus_imap_la_SOURCES += \
     imap/squat_internal.h
 endif
 
-if HTTPD
 imap_libcyrus_imap_la_SOURCES += \
     imap/caldav_alarm.c \
```
dilyanpalauzov: Can all these files be moved to line 1010:
```
imap_libcyrus_imap_la_SOURCES = \
    imap/annotate.c \
    imap/annotate.h \
```
so that there is no unconditional `+=` applied to `imap_libcyrus_imap_la_SOURCES`?

docsrc/compiling.rst
```diff
@@ -297,10 +297,18 @@ via configure.
...
+JMAP
+####
 
-    ``./configure --enable-http --enable-calalarmd --enable-jmap``
+    ``./configure --enable-jmap``
+
+.. note::
+
+    HTTP, CalDAV, CardDAV, WebDAV and the calendar alarm daemon are
+    always built and no longer require configure options.  libical,
```
dilyanpalauzov, line 259: I suggest including a date or a version 3.13 in this `note:`, so after 10 years or so the note can be removed.

dilyanpalauzov (commented):
At https://www.cyrusimap.org/dev/developer/compiling.html are listed at the top Required Dependencies and in a different section dependencies for “CalDAV, CardDAV, or JMAP (httpd subsystem)”.  The latter lists libical, libxml2 and others.

I suggest moving libical, xxd, libxml2 from the “CalDAV, CardDAV, or JMAP (httpd subsystem)” section to the “Required Build Dependencies” section.

changes/next/cyr-1715-mandatory-dav
rjbs: I can just change the text, for now, to 4.0.0, because that is more or less enough.  We (Fastmail) now require 4.0.3 for a small line ending bug, but we'll _build_.

configure.ac
rjbs: Haha, yeahhhhh, thanks.

Makefile.am
rjbs: "It looked right with ts=2!" (fixed)

cassandane/tiny-tests/ALPN/https_bad
rjbs: I will remove this for now, but I am wary of "and then we have these forever".  It seems to me like we need these much more for cross-version-specific tests than for all tests.  But it's an easy commit to produce again later, so it can wait for discussion.

configure.ac
rjbs: Agreed, removed.

doc/examples/cyrus_conf/normal.conf
rjbs: Agreed.  Documenting this stuff is out of scope of this PR, but we should do it.

docsrc/download/installation/manage-dav.rst
rjbs: I'm going to leave it.  It anticipates the complaint, "But I don't want to run CalDAV!"

Makefile.am
rjbs: Good idea, done.

docsrc/compiling.rst
rjbs: I don't expect this file to survive more than a week or three, I have plans, so I'm not going to do this right now.

rjbs:
@elliefm _now_ I think this is maybe ready to go :)

Makefile.am
dilyanpalauzov, line 269: I suggest moving this `BUILT_SOURCES += ` to line 171 right after
```
if SERVER
BUILT_SOURCES += \
    imap/http_err.c \
    imap/http_err.h \
```
Likewise for line 190 and:
```
libexec_PROGRAMS += imap/fud
libexec_PROGRAMS += imap/httpd
```

configure.ac
dilyanpalauzov, line 446: Will trying to build without sqlite work?  The line above contains:

> dnl sqlite not specifically requested: not found isn't a failure (yet)

configure.ac
```diff
@@ -1434,166 +1434,145 @@ AC_PATH_PROG([XXD],[xxd])
...
 
-        PKG_CHECK_MODULES([ICAL], [libical >= 3.99.99], [
-                AC_DEFINE(HAVE_ICAL,[],[Build CalDAV support into httpd?])
-                LIBS="$LIBS ${ICAL_LIBS} -licalvcard"
-                with_ical=yes
-                ],
-                AC_MSG_ERROR([Need libical for http]))
+PKG_CHECK_MODULES([XML2], [libxml-2.0], [
+        LIBS="$LIBS ${XML2_LIBS}"
+        with_xml2=yes
```
dilyanpalauzov, line 1454: The value of the variable `with_xml2` is not read.

changes/next/cyr-1715-mandatory-dav
```diff
@@ -0,0 +1,39 @@
+Description:
+
+HTTP support, *DAV (CalDAV, CardDAV, WebDAV) and the calendar alarm daemon
+are now mandatory parts of every Cyrus IMAP build.  The ``--enable-http``
+and ``--enable-calalarmd`` configure options have been removed; the httpd
+binary and calalarmd are always built.  libical, libxml2 and SQLite3 are
```
dilyanpalauzov, line 6: > the httpd binary and calalarmd are always built

Correct is to say, that these are compiled by default.   After `./configure --disable-server` `calalarmd` and `httpd` are not built.

elliefm: Uggh, I'd forgotten about `--disable-server`.  What a nuisance.

configure.ac
elliefm: This is part of the mechanism for deciding whether to provide the cyrusdb "sql" backend, with sqlite as its underlying sql engine.  We don't enable that automatically, even when sqlite is available.

With this PR, sqlite will always be available, since httpd needs it.  But cyrusdb should still work as before: no "sql" backend unless you actually ask for it.

That said, the way you ask for the cyrusdb sql backend is a mess.  I expect we'll revisit it soon, but separately from this.

doc/examples/cyrus_conf/normal.conf
```diff
@@ -18,7 +18,8 @@ SERVICES {
 #  nntp          cmd="nntpd" listen="nntp" prefork=0
 #  nntps         cmd="nntpd -s" listen="nntps" prefork=0
 
-  # these are only necessary if using HTTP for CalDAV, CardDAV, or RSS
+  # http serves CalDAV, CardDAV, JMAP, RSS, etc.; see the httpmodules
+  # option in imapd.conf(5) to choose which modules each httpd advertises
   http          cmd="httpd" listen="http" prefork=0
   https         cmd="httpd -s" listen="https" prefork=0
```
elliefm: I'd suggest `# httpd serves [...]` rather than `# http serves [...]`.  `http` is a service name and corresponding port name, `httpd` is the daemon that serves it.

elliefm (approved):
One nit, but this looks fine.

I have a hunch we might be able to clean up configure.ac further, it looks like there might be a few details that were previously necessary but are now not.  But the objective here is to "make httpd mandatory", and that objective has been met, and it's easy to reason about because the changes are simple.

I expect I'll want to do some refactoring later to take advantage of the new simplicity, but probably not until after we've ripped out everything else we plan to rip out.

configure.ac
dilyanpalauzov: I do not get it.  Either sqlite3 is required by `./configure` and mandatory prerequisite for building (`make`), or it is optional both for `./configure` and `make`.  This change adds to `configure.ac` line 1438:

> dnl libical, libxml2, sqlite3 and xxd are required build dependencies.

If it is required, then the message `AC_MSG_NOTICE([sqlite was not found, will try to build without it])` is not correct.

PR #6100, opened 2026-06-26
http_dav_sharing: guard against empty <share-access/> child

rjbs:
dav_post_share() called xmlFirstElementChild() on the <share-access>
element and then immediately read share->name without checking whether
share was NULL.  A DAV admin POSTing a share-resource body with an empty
<share-access/> element would crash the httpd worker.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/imap/http_dav_sharing.c
+++ b/imap/http_dav_sharing.c
@@ -846,6 +846,7 @@ static int dav_store_notification(struct transaction_t *txn,
     }
 
 done:
+    dlist_free(&dl);
     buf_destroy(xmlbuf);
     mboxlist_entry_free(&mbentry);
     return r;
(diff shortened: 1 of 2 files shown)


imap/http_dav_sharing.c
```diff
@@ -846,6 +846,7 @@ static int dav_store_notification(struct transaction_t *txn,
     }
 
 done:
+    dlist_free(&dl);
```
ksmurchison: This addition seems wrong.  `goto done` appears in only two places.  In the first, `dl` isn't declared yet.  In the second, `dl` is `NULL` and doesn't need to be freed.  I believe this is what is causing ASAN to fail

ksmurchison (changes requested):

ksmurchison (approved):

PR #6101, opened 2026-06-26
http_caldav_sched.c: defensively NULL-check iSchedule response status

rjbs:
In sched_deliver_remote, when processing the <schedule-response> from a remote iSchedule peer, the per-response loop initialises `status` to NULL and only assigns it if a <request-status> child element is found. xmlNodeGetContent may also return NULL for an empty element.

Abort before dereferencing null.

### DIFF
--- a/imap/http_caldav_sched.c
+++ b/imap/http_caldav_sched.c
@@ -1154,6 +1154,16 @@ static void sched_deliver_remote(const char *cal_ownerid, const char *sched_user
                         status = xmlNodeGetContent(node);
                 }
 
+                /* A bogus response may omit <request-status>, or provide it as
+                 * an empty element (in which case xmlNodeGetContent can return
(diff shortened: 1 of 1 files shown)


imap/http_caldav_sched.c
```diff
@@ -1154,6 +1154,16 @@ static void sched_deliver_remote(const char *cal_ownerid, const char *sched_user
                         status = xmlNodeGetContent(node);
                 }
 
+                /* A bogsu response may omit <request-status>, or provide it as
```
ksmurchison: Typo

ksmurchison (approved):

imap/http_caldav_sched.c
rjbs: Oops, I had fixed it right after making it, but not before pushing it…

PR #6103, opened 2026-06-26
sync: use conversations DB for RESERVE GUID lookups

brong:
sync_apply_reserve previously located each requested GUID by opening every listed mailbox and walking all of its records - O(messages in all folders) work to find what is typically just a handful of messages.

Use the conversations DB instead, which indexes GUID -> (folder, uid). For each distinct user owning the listed folders we now:

  - collect candidate locations for every needed GUID via conversations_guid_foreach, opening no mailboxes (Phase A), looking GUIDs up in sorted order so reads into the G<guid>-keyed DB stay forward-only;
  - then open each containing mailbox at most once, most-loaded first, reading its UIDs in ascending order, and link/stage the files, stopping as soon as everything is found (Phase B).

A GUID may exist only as an embedded sub-part (e.g. a message/rfc822 attachment whose decoded content hashes to that GUID); when no whole-message copy exists anywhere we extract and stage those bytes instead, mirroring jmap_findblob. Whole-message hits are guarded against a stale conversations DB by verifying the record GUID matches.

The per-message link logic is factored into reserve_one(); the legacy brute-force folder scan (reserve_folder) is retained as a fallback only for folders whose owner has no conversations DB.

...

There's also a second commit on here renaming all the "part" usage in the replication code to "partition" because it's ambiguous with "part" from conversations lookup, which means "mime part id".

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -524,18 +524,18 @@ struct sync_reserve_list *sync_reserve_list_create(int hash_size)
 }
 
 struct sync_msgid_list *sync_reserve_partlist(struct sync_reserve_list *l,
-                                              const char *part)
+                                              const char *partition)
 {
     struct sync_reserve *item;
 
     for (item = l->head; item; item = item->next) {
-        if (!strcmpsafe(item->part, part))
+        if (!strcmpsafe(item->partition, partition))
             return item->list;
     }
 
     /* not found, create it */
     item = xmalloc(sizeof(struct sync_reserve));
-    item->part = xstrdup(part);
+    item->partition = xstrdup(partition);
     item->next = NULL;
     item->list = sync_msgid_list_create(l->hash_size);
 
@@ -556,7 +556,7 @@ void sync_reserve_list_free(struct sync_reserve_list **lp)
     while (current) {
         next = current->next;
         sync_msgid_list_free(&current->list);
-        free(current->part);
+        free(current->partition);
         free(current);
         current = next;
     }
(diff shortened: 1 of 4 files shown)


imap/sync_support.c
```diff
@@ -2480,6 +2501,354 @@ static void reserve_folder(const char *part, const char *mboxname,
     mailbox_close(&mailbox);
 }
 
+/* Extract a message that exists only as an embedded sub-part (e.g. a
+ * message/rfc822 attachment whose decoded content hashes to the GUID we
+ * want) and stage those bytes.  Only used when no whole-message copy of the
+ * GUID exists anywhere in the user's account. */
+static int reserve_one_part(const char *part, struct mailbox *mailbox,
```
brong: the name "part" is used interchangeably for "partition" and "partid" in this codebase, which is very confusing.  We should rename every use of 'part' to 'partition' or 'partid' depending which one it is in this file and the surrounding sync_client.c, sync_server.c and sync_support.h

imap/sync_support.c
```diff
@@ -2480,6 +2501,354 @@ static void reserve_folder(const char *part, const char *mboxname,
...
+
+        r = mailbox_open_irl(refs[i].mboxname, &mailbox);
+        if (!r) r = sync_mailbox_version_check(&mailbox);
+        if (r) continue;
+
+        for (j = 0; j < ptrarray_size(cands); j++) {
+            struct reserve_candidate *cand = ptrarray_nth(cands, j);
+            struct index_record record;
+
+            /* already found from an earlier (more loaded) mailbox? */
```
brong, line 2824: This will still open every mailbox, even if everything has been fetched that could be from that mailbox, or indeed if everything has been fetched that could be total -- so long as there was something requested which doesn't exist locally.  We should optimize this too.

ksmurchison (approved):
The logic looks sound and the Replication tests all pass!

brong:
Tagging include-in-fastmail so we can see it work on future first :)

PR #6107, opened 2026-06-29
imapoptions/specialuser_nochildren: fix default value

ksmurchison:
The correct specual-use attribute used throughout Cyrus is \\Snoozed

PR #6108, opened 2026-06-29
special use options: tweak defaults

rjbs:
* \Scheduled should be in lmtp_exclude_specialuse default
* \Snoozed, not \Snooze, should be in specialuse_protect default

### DIFF
--- a/cassandane/tiny-tests/Specialuse/default_snoozed
+++ b/cassandane/tiny-tests/Specialuse/default_snoozed
@@ -1,0 +1,29 @@
+#!perl
+use Cassandane::Tiny;
+
+# The default specialuse_protect and specialuse_nochildren lists must name
+# the \Snoozed special-use (with the trailing "d"), as that is the only
+# spelling the server ever applies.  This runs with the stock defaults, so it
+# would fail if the option defaults ever drifted back to "\Snooze".
+sub test_default_snoozed ($self)
+{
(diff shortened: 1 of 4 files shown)


rjbs:
This covers https://github.com/cyrusimap/cyrus-imapd/pull/6107/changes and a bit more.

cassandane/tiny-tests/Specialuse/default_snoozed
```diff
@@ -0,0 +1,31 @@
+#!perl
+use Cassandane::Tiny;
+
+# The default specialuse_protect and specialuse_nochildren lists must name
+# the \Snoozed special-use (with the trailing "d"), as that is the only
+# spelling the server ever applies.  This runs with the stock defaults, so it
+# would fail if the option defaults ever drifted back to "\Snooze".
+sub test_default_snoozed
+    :NoAltNamespace
```
rsto: I'm not sure the test needs the `NoAltNamespace` annotation?

rsto (approved):
LGTM - although I'm fuzzy about if "NoAltNamespace" is

ksmurchison (approved):

cassandane/tiny-tests/Specialuse/default_snoozed
rjbs: I just find it a bit simpler, but I've rewritten it to avoid the attribute.

PR #6109, opened 2026-06-29
Fix crash if a mailbox is deleted while append is happening

brong:
We get this crash very rarely, but it can happen!  Found one user who had 3 of them; wild - they deleted an entire tree while appending to it from a client.


ksmurchison (commented):
Why is CI running for 60+ minutes and failing?

brong:
> Why is CI running for 60+ minutes and failing?

yeah, there was an infinite loop in the test case!  Dammit.  Fixed.

ksmurchison (approved):

PR #6115, opened 2026-06-30
twom: check mmap() result against MAP_FAILED, not NULL

rjbs:
mmap(2) signals failure by returning MAP_FAILED, not NULL.

Three sites in lib/twom.c did `if (!file->base)` after the mmap, so the failure path was never taken on real mmap failure; the next read_header() call then dereferenced (void *)-1 and crashed.

Check MAP_FAILED at each site and reset file->base to NULL on failure so the cleanup paths (which do test for NULL before munmap) behave correctly afterwards.


ksmurchison (commented):
Seems sane to me but I will defer to @brong

brong (approved):

brong:
Yes, it's good - we should merge this.  AND THEN I need to port it to libtwom and also convert Cyrus to use libtwom.  Bah.

PR #6116, opened 2026-07-01
cunit: configurable tmpdir

elliefm:
Context: Our cunit tests use a lot of temporary files and directories.  Some honour `$TMPDIR`, but most are hardcoded to use `/tmp`.  Most linux systems default to putting `/tmp` on persistent storage, so that temporary files survive reboots and power failures.  Our temporary files are cleaned up at process exit anyway, so we don't need this.  It would be nice to have the option to configure cunit to use a path on tmpfs, to reduce wasted SSD write cycles.  Also, there's some tests in libconfig.testc that deliberately end up in `fatal()`, and don't/can't clean up their own temporary config files.

Primarily, this PR:

* Adds `cunit_tmpdir()` and `cunit_tmpfile()` APIs to our testing infrastructure so that tests can request temporary storage in a consistent way, and updates everything to use it.  These APIs honour `$CYRUS_CUNIT_TMPDIR` if set, or fall back to `$TMPDIR` or `/tmp` if not.
* Plumbs `$CYRUS_CUNIT_TMPDIR` all the way through the build system, so that it can be set at configure time.
* Arranges so that all temporary files and directories created by tests are created below the  working directory that was created for the `cunit/unit` processs.  This reduces clutter in the main tmpdir, and means any temporary files leaked by tests get cleaned up in the end when the working directory is removed.

To achieve those objectives, it also:

* Establishes a `data` subdirectory for test data files, replacing the hardcoded `*.pem` detail.
* Replaces the nasty multiline shell scripts embedded in the `check-local`, `check-discrete`, and `check-%` make targets with calls to a maintainable perl script.
* Adds a `check-discrete-%` make target, now that it's trivial to do so.

**If you don't use this feature**, everything will mostly work in the same way.  The main difference is that there will be less pollution in `ls /tmp` output (one directory per `cunit/unit` process, instead of possibly several per test).

**To use this feature**, pass `CYRUS_CUNIT_TMPDIR=/some/path` as an argument to configure, or arrange for it to be in the environment when configure runs.  The path must already exist, configure will error if it doesn't.

Setting `$TMPDIR` in your shell environment will affect most parts of your system.  If that's what you want, or already have, our tests will now respect it consistently.  But if you want to affect just our tests, `$CYRUS_CUNIT_TMPDIR` is the variable to set.


elliefm:
(One other interesting consequence that comes from this is that if your system somehow does not have `/tmp`, and you try to configure without choosing an alternative, you'll get a configure error, instead of tests breaking later.)

ksmurchison (approved):
LGTM

nwc10 (approved):
I tried `configure` with a tmpdir and the directory was empty at the end, so I wasn't sure if it was really being used...

I can get past the `configure` test for "does the 'file' exist" like this:

```
# touch /dev/shm/cunit
```

and we get to

```
Running unit tests
/usr/bin/perl cunit/run --verbose --tmpdir=/dev/shm/cunit
Error in tempdir() using /dev/shm/cunit/cyrus-cunit-XXXXXX: Parent directory (/dev/shm/cunit) is not a directory at cunit/run line 153.
make[3]: *** [Makefile:9767: check-local] Error 255
make[3]: Leaving directory '/Volumes/Code/Cyrus/cyrus-review'
make[2]: *** [Makefile:8412: check-am] Error 2
make[2]: Leaving directory '/Volumes/Code/Cyrus/cyrus-review'
make[1]: *** [Makefile:8112: check-recursive] Error 1
make[1]: Leaving directory '/Volumes/Code/Cyrus/cyrus-review'
make: *** [Makefile:8414: check] Error 2
```

Boom! Yes, it's using it. And cleaning up on completion.

Nice.

wolfsage (commented):
I like the idea. I wonder if this is a good time to standardize the naming a bit? Just a thought.

We have:
```
cunit-db-test.XXXXXX
cyrus-conv.datXXXXXX
cyrus-cunit-XXXXXX
cyrus-cunit-proctestc-XXXXXX
cyrus-protXXXXXXXX
```

I think the ideal situation is they identify both cyrus and cunit, and *also* the test file they come from.

If cunit_tempfile / cunit_tmpdir both automatically included 'cyrus-cunit') then each test could do:
```
fd = cunit_tmpfile(tempfile, sizeof(tempfile), "sieve");
```

for example

I think cunit_tmpfile(..., "aaa-db") would be better than cunit_tmpfile(..., "db") too.

Anyway, we don't *have* to, I just think it'd be nice/consistent

cunit/pcunit
```diff
@@ -63,16 +64,18 @@ while (@suites) {
     }
 
     my $next = shift @suites;
-    my (undef, $outfile) = tempfile(UNLINK => 0);
+    my (undef, $outfile) = tempfile(DIR => $tmpdir, UNLINK => 0);
```
wolfsage, line 67: While we're here, if this also had a template:
```
tempfile("cunit-fast-results-XXXXXXXX", DIR => $tmpdir, UNLINK => 0);
```

then these would be named nicely too instead of random strings in the tmpdir. (But again, not requesting a change, but we could change it to tidy it up more)

wolfsage (approved):
Looks good

elliefm:
@wolfsage Yeah.  I intend to clean up the mess of tmpdir naming soon, but I intentionally left it alone for now to keep the substantive changes easy to reason about.  Don't paint until the renovations are finished, right?

cunit/pcunit
elliefm: Yeah, I intend to.  But this PR is the renovations, repainting comes later.

elliefm:
@nwc10 If you want to watch this change _working_,  you can run something like `watch -n1 'ls -d /tmp/*'` in another terminal while cunit without this branch runs; and then do the same while this branch's cunit runs; and then do the same again after having reconfigured with CYRUS_CUNIT_TMPDIR set to somewhere else.  Without this branch, cunit creates and deletes tonnes of these files and/or directories (and leaks some at the end); this branch makes just one directory per cunit/unit process with the mess hidden inside it; and when configured with some other path, it doesn't touch `/tmp` at all.

brong (approved):
Works for me on Mac - build and full test suite

elliefm:
> I intend to clean up the mess of tmpdir naming soon

That's #6139

PR #6118, opened 2026-07-01
jmap_calendar.c: don't crash CalendarEvent/get on a startless event

rjbs:
getcalendarevents_get_utctimes reads the "start" property to derive utcStart/utcEnd, but a VEVENT with no DTSTART yields no "start", so it passed NULL to jmapical_localdatetime_from_string, which dereferenced it in strptime and crashed httpd.


ksmurchison (approved):
If we want to convert these events to JSCalendar and not ignore them, then this looks sane.

Approving, assuming CI passes

rjbs:
rsto wrote a better fix

PR #6119, opened 2026-07-02
backup: keep a deleted folder through compaction until it ages out

brong:
This came up when testing backup compaction for Fastmail, we found that a file would compact and immediately have more changes to compact!  Only the second pass would lock in entirely.

This showed that we were losing deleted folders immediately after the second repack; even though they're supposed to persist for some time for data safety.

This change persists the deletion marker, so the compaction continues to hold the remaining data until it times out.


rjbs (approved):
👍🏽

PR #6124, opened 2026-07-03
jmap_vacation.c: escape user-controlled strings spliced into Sieve script

rjbs:
VacationResponse/set built the generated Sieve script by interpolating the user-supplied `subject`, `textBody`, and `htmlBody` with bare `%s`:

```c
  buf_printf(&data, " :subject \"%s\"", subject);
  ...
  buf_printf(&data, "%s\r\n\r\n--%s\r\n", textBody, boundary);
  ...
  buf_appendcstr(&data, "\r\n.\r\n;\r\n}\r\n");
```

A `subject` containing `"` closed the Sieve quoted-string early and let the client splice in further tokens.  In theory, this is a security issue because it escalates VacationResponse/set into a SieveScript/set (more or less), but in reality this isn't an escalation: the two rights are the same.

Validate subject for control characters at parse time, then escape dquote and backslash when emitting.  Dot-stuff the multi-line body per RFC 5228 sec. 2.4.2.2 and normalize bare CR/LF to CRLF so a body line ending in just LF + `.` + LF still gets stuffed.

### DIFF
--- a/cassandane/tiny-tests/JMAPVacationResponse/vacation_set_sieve_injection
+++ b/cassandane/tiny-tests/JMAPVacationResponse/vacation_set_sieve_injection
@@ -1,0 +1,135 @@
+#!perl
+use Cassandane::Tiny;
+
+# VacationResponse/set builds a Sieve script by splicing the client-supplied
+# `subject`, `textBody`, and `htmlBody` into it.  Confirm those values are
+# neutralized, so that Sieve metacharacters in them stay inert data and are
+# never parsed as Sieve syntax:
+#
+#   1. `subject` becomes the argument to `:subject`, a Sieve quoted-string.
+#      A `"` in the value must be backslash-escaped so it cannot close the
+#      string early and let the rest be parsed as further tokens
+#      (`; redirect "..."`).
+#
+#   2. `textBody`/`htmlBody` become a `text:` multi-line section, terminated
+#      by a line containing only `.`.  A `.`-only line in the value must be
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):
LGTM, but do we want to combine sieve_qstring_valid() and sieve_emit_qstring().  I don't see any reason not to, and it avoids a second iteration through the subject string.

rjbs:
Does that make sense?  We're using `sieve_qstring_safe` to validate the `subject` argument to the create call, so we can reject something that we can't qstring-ify.

Later, we use `sieve_emit_qstring` to append the newly-qstring-ified string to the Sieve buf we're building.

The only ways I can see to combine them seem worse, but you might have a better idea than I'm able to form right now!

ksmurchison:
> Does that make sense? We're using `sieve_qstring_safe` to validate the `subject` argument to the create call, so we can reject something that we can't qstring-ify.
> 
> Later, we use `sieve_emit_qstring` to append the newly-qstring-ified string to the Sieve buf we're building.
> 
> The only ways I can see to combine them seem worse, but you might have a better idea than I'm able to form right now!

I'm thinking that while verifying that the subject is valid, we also create the quoted string.  Maybe I'm totally off my rocker.  I'll give this another look on Monday

rjbs:
I guess my concern would be:  if we do that, we need to reply to the /set with the server-changed subject property, because otherwise the /set they passed in wouldn't match the /get they pulled back.  Not a big deal, probably not much work, but I don't know how to do it! :D  Let me know what you think is best, I'm easy going!

PR #6125, opened 2026-07-03
lmtpengine.c: avoid one-byte underflow in clean_retpath on lone "<"

rjbs:
When the Return-Path header was exactly "<", clean_retpath() would strip the leading angle bracket, leaving a zero-length string, then test rpath[sl-1] (= rpath[-1]) for a closing ">".  That read (and the conditional write that followed) underflowed the heap buffer by one byte.

Reachable only through the Return-Path header path in savemsg(), as parseaddr() rejects "<" coming in via MAIL FROM, so most MTAs will have rewritten the header before it reaches lmtpd.

### DIFF
--- a/imap/lmtpengine.h
+++ b/imap/lmtpengine.h
@@ -98,6 +98,11 @@ void lmtpmode(struct lmtp_func *func,
               struct protstream *pout,
               int fd);
 
+/* strip the surrounding <> (and any trailing whitespace) from a return
+ * path; exposed here primarily so the cunit suite can exercise it
+ */
+void clean_retpath(char *rpath);
+
 /************** client-side LMTP ****************/
 
 #include "backend.h"
--- a/Makefile.am
+++ b/Makefile.am
@@ -648,6 +648,7 @@ cunit_TESTS = \
     cunit/imparse.testc \
     cunit/index_file.testc \
     cunit/libconfig.testc \
+    cunit/lmtpengine.testc \
     cunit/loginlog.testc \
     cunit/logfmt.testc \
     cunit/mailbox.testc \
@@ -696,7 +697,8 @@ cunit_TESTS += cunit/ical_support.testc
 endif
 
 cunit_unit_SOURCES = $(cunit_FRAMEWORK) $(cunit_TESTS) \
-    imap/css3_color.c imap/mutex_fake.c imap/spool.c imap/search_expr.c \
+    imap/css3_color.c imap/lmtpengine.c imap/mutex_fake.c imap/spool.c \
+    imap/search_expr.c \
     master/event.c master/cronevent.c
 nodist_cunit_unit_SOURCES = cunit/unit-registry.c
 cunit_unit_LDADD = $(LD_SIEVE_ADD) $(LD_UTILITY_ADD) -lcunit
--- a/cunit/lmtpengine.testc
+++ b/cunit/lmtpengine.testc
@@ -1,0 +1,68 @@
+#include <config.h>
+
+#include <stdlib.h>
+#include <string.h>
+
+#include "cunit/unit.h"
+
+#include "imap/lmtpengine.h"
+
+/* clean_retpath() strips a single layer of surrounding <> from the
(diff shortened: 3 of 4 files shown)


imap/lmtpengine.h
```diff
@@ -98,6 +98,11 @@ void lmtpmode(struct lmtp_func *func,
               struct protstream *pout,
               int fd);
 
+/* strip the surrounding <> (and any trailing whitespace) from a return
+ * path; exposed here primarily so the cunit suite can exercise it
+ */
+void clean_retpath(char *rpath);
```
dilyanpalauzov: This can get `__attribute__ ((nonnull))`.

nwc10 (commented):
The change looks correct.

One thing that did pop into my head - given the general reliability of `git blame`, why add annotations such as `-- rjbs, 2026-07-03` into the comment, when the commit history metadata can reveal that to anyone needing it? It's not wrong, but it's adding a line.

The linking that causes CI to fail **would** be fixed by the following change, but I'm not sure if it's the right plan. We don't seem to have written any unit tests for lmtp... before, but `deliver.c` seems to have had to bodge around the same issue.

```
diff --git a/cunit/lmtpengine.testc b/cunit/lmtpengine.testc
index c8f48d3b50..f6f0cc94a8 100644
--- a/cunit/lmtpengine.testc
+++ b/cunit/lmtpengine.testc
@@ -7,6 +7,14 @@

 #include "imap/lmtpengine.h"

+/* The (formerly) static function that we're testing causes all of lmtpengine.o
+ * to be linked. Hence its symbol dependencies need to be satisfied.
+ * The "real" deliver_logfd is in lmtpd.c, but we're not linking with lmtpd.o
+ * So fake it. Much like the comment in deliver.c:
+ *    unused for deliver.c, but needed to make lmtpengine.c happy
+ */
+int deliver_logfd;
+
 /* clean_retpath() strips a single layer of surrounding <> from the
  * return path.  These helpers verify that on every interesting input
  * it produces the documented result and is memory-safe.
```

rjbs:
The name/date on comments survives relocation, which confounds blame.

nwc10 (approved):
Oh boy.

So, in theory it's reachable via this

```
    if (!m->return_path &&
        (body = msg_getheader(m, "return-path"))) {
        /* let's grab return_path */
        m->return_path = xstrdup(body[0]);
        clean822space(m->return_path);
        clean_retpath(m->return_path);
    }
```

but that has to have been dead code since Feb 2000, as https://github.com/cyrusimap/cyrus-imapd/commit/689b67f6455629f1322985d8a36ca0eaf20841b0 removed the **other** call to `savemsg`

```
-    /* Copy message to temp file */
-    savemsg(msgdata, 0);
```

meaning the only route into `savemsg` is here, in the `DATA` command:

```
      case 'd':
      case 'D':
            if (!strcasecmp(buf, "data")) {
                int delivered = 0;
                int j;

                if (!msg->rcpt_num) {
                    prot_printf(pout, "503 5.5.1 No recipients\r\n");
                    continue;
                }
                /* copy message from input to msg structure */
                r = savemsg(&cd, func, msg);
```

and you can only get there if there are non-zero recipients, which means you got past here (`RCPT TO`):

```
      case 'r':
      case 'R':
            if (!strncasecmp(buf, "rcpt ", 5)) {
                char *rcpt = NULL;
                int ignorequota = 0;
                char *tmp;

                if (!msg->return_path) {
                    prot_printf(pout, "503 5.5.1 Need MAIL command\r\n");
                    continue;
                }
```

at which point, you have a return path.

So, should we remove the dead code? Or replace it with an assertion. Or a "this is impossible, right?" logging? (in a followup PR)

nwc10:
> So, should we remove the dead code? Or replace it with an assertion. Or a "this is impossible, right?" logging? (in a followup PR)

Oh, which makes `clean822space` dead code, if I am correct.

rjbs:
@nwc10 Short version: I'm all for removing truly dead code!

PR #6128, opened 2026-07-06
Makefile.am: remove broken junit output format

elliefm:
`make CUFORMAT=junit check` fails like this:

```
Running unit tests
Can't call method "getEncoding" on an undefined value at /usr/share/perl5/XML/DOM.pm line 1221.
```

Looks like something in cunit-to-junit.pl has bitrotted.  It's unused, untested, unmaintained; let's just get rid of it.


rsto (approved):

ksmurchison (approved):

rjbs (approved):

nwc10 (commented):
Not sure if we want to do this as part of this MR. I'm fine either way (so using "comment" rather than "approve" or "request changes")

I had no idea what this file was. I looked at its history. "Not much". But other code it dragged in that we can now remove is:

```
diff --git a/.gitignore b/.gitignore
index 32bea84819..57ab4813e3 100644
--- a/.gitignore
+++ b/.gitignore
@@ -49,8 +49,6 @@ conftest*
 com_err/et/compile_et
 coverage.xml
 cscope.out
-cunit/CUnitAutomated-Results.xml
-cunit/reports
 cunit/test-dbdir
 cunit/unit
 ### ignore intermediate c files from foo.testc
diff --git a/cunit/unit.c b/cunit/unit.c
index 676e23d048..694de7cb91 100644
--- a/cunit/unit.c
+++ b/cunit/unit.c
@@ -9,7 +9,6 @@
 #include <unistd.h>
 #include <CUnit/CUnit.h>
 #include <CUnit/Basic.h>
-#include <CUnit/Automated.h>
 #if HAVE_VALGRIND_VALGRIND_H
 #include <valgrind/valgrind.h>
 #endif
@@ -29,7 +28,6 @@ int verbose = 0;
 int num_testspecs = 0;
 const char **testspecs;
 enum { RUN, LIST } mode = RUN;
-int xml_flag = 0;
 int timeouts_flag = 1;

 #if HAVE_VALGRIND_VALGRIND_H
@@ -304,19 +302,6 @@ static void run_tests(void)
     if (timeouts_flag && timeout_init(handle_timeout) < 0)
         exit(1);

-    if (xml_flag) {
-        if (num_testspecs == 0) {
-            /* not specified: run all tests in order listed */
-            CU_automated_run_tests();
-            if (timeouts_flag)
-                timeout_fini();
-            return;
-        }
-        fprintf(stderr, "unit: test specifications not "
-                        "supported in XML mode, sorry\n");
-        exit(1);
-    }
-
     if (verbose)
         CU_basic_set_mode(CU_BRM_VERBOSE);

@@ -413,7 +398,7 @@ static void parse_args(int argc, char **argv)
 {
     int c;

-    while ((c = getopt(argc, argv, "hltvx")) > 0)
+    while ((c = getopt(argc, argv, "hltv")) > 0)
     {
         switch (c)
         {
@@ -429,9 +414,6 @@ static void parse_args(int argc, char **argv)
         case 'v':
             verbose++;
             break;
-        case 'x':
-            xml_flag = 1;
-            break;
         case '?':
             usage(1);
             exit(1);```
```

The `.gitignore` change starts as part of commit f12e4fa5f9c010bb175c4c609ae8767cf4e6a11c but must have got moved at some point.

elliefm:
@nwc10 That XML generation stuff is cunit's own XML output format.  I intentionally left it alone, because it appears to work fine, and isn't in my way.  If nothing else, its existence is useful as documentation of the fact that other ways are possible.

elliefm:
Though, good spot on the `cunit/reports` line in .gitignore, I'll clean that up!

nwc10 (approved):

PR #6129, opened 2026-07-06
http_caldav.c: remove JMAP X- properties from client-supplied data

ksmurchison:
Certain X- props are set via JMAP and we don't want CalDAV clients to create/update them via PUT or iTIP

Its probaby best to review commit-by-commit


imap/http_caldav.c
```diff
@@ -4050,6 +4066,19 @@ static int caldav_put(struct transaction_t *txn, void *obj,
...
         }
+
+        /* Remove JMAP X- properties */
+        for (prop = icalcomponent_get_first_property(comp, ICAL_X_PROPERTY);
+             prop;
+             prop = next) {
+            next = icalcomponent_get_next_property(comp, ICAL_X_PROPERTY);
+
+            const char *xname = icalproperty_get_property_name(prop);
+            if (strarray_find(&jmap_x_props, xname, 0) >= 0) {
```
rsto: Does libical normalize property names to upper case? If not, let's better switch to case-insensitive comparison while we are at it.

imap/caldav_util.c
```diff
@@ -967,25 +967,7 @@ static void preserve_jmap_permissions(icalcomponent *ical,
...
-            nextprop = icalcomponent_get_next_property(comp, ICAL_X_PROPERTY);
-
-            if (!strcmp(icalproperty_get_x_name(prop), JMAPICAL_XPROP_MAYINVITESELF) ||
-                !strcmp(icalproperty_get_x_name(prop), JMAPICAL_XPROP_MAYINVITEOTHERS)) {
-                icalcomponent_remove_property(comp, prop);
-                icalproperty_free(prop);
-            }
-        }
-    }
+    /* NOTE: JMAP permission X- properties have been stripped in caldav_put() */
```
rsto: This looks like a real regression. caldav_store_resource also is called for COPY/MOVE, iTIP and even GET if the VTIMEZONE is getting stripped.

imap/http_caldav.c
```diff
@@ -4050,6 +4066,19 @@ static int caldav_put(struct transaction_t *txn, void *obj,
...
                            comp, &overrides);
         }
+
+        /* Remove JMAP X- properties */
+        for (prop = icalcomponent_get_first_property(comp, ICAL_X_PROPERTY);
+             prop;
+             prop = next) {
+            next = icalcomponent_get_next_property(comp, ICAL_X_PROPERTY);
+
+            const char *xname = icalproperty_get_property_name(prop);
```
rsto: nit: because we already know it's an x-property we can call `icalproperty_get_x_name` which prevents libical from allocating a temporary buffer for the name.

rsto (changes requested):
This seems to introduce a regression by only stripping these special x-properties in PUT now, where before they also were stripped for iTIP and other CalDAV methods.

I'll address the question how to handle X-JMAP-PRIVACY separately.

imap/http_caldav.c
ksmurchison: I will fix

imap/caldav_util.c
ksmurchison: COPY, MOVE, GET all operate on resources already on disk, so there is no way for a client to insert X-JMAP-*.

Would we actually expect these properties to show up in an iTIP message?

If we move this outside of PUT, then SENTBY would be handled separately, since its inserted as part of PUT.

Looking for further guidance

imap/http_caldav.c
ksmurchison: Good point, I will fix

imap/caldav_util.c
rsto: > COPY, MOVE, GET all operate on resources already on disk, so there is no way for a client to insert X-JMAP-*.

Yeah, right. I looked at the call sites for caldav_store_resource only.

> Would we actually expect these properties to show up in an iTIP message?

We don't expect it, but we don't want it to be possible sending iTIP messages that alter the ACL for that event.

> If we move this outside of PUT, then SENTBY would be handled separately, since its inserted as part of PUT.

If that's required then SENTBY should indeed be handled separately? the iTIP case looks like a real regression to me, doesn't it?

imap/caldav_util.c
```diff
@@ -967,26 +967,8 @@ static void preserve_jmap_permissions(icalcomponent *ical,
...
-            if (!strcmp(icalproperty_get_x_name(prop), JMAPICAL_XPROP_MAYINVITESELF) ||
-                !strcmp(icalproperty_get_x_name(prop), JMAPICAL_XPROP_MAYINVITEOTHERS) ||
-                !strcmp(icalproperty_get_x_name(prop), JMAPICAL_XPROP_HIDEATTENDEES)) {
-                icalcomponent_remove_property(comp, prop);
-                icalproperty_free(prop);
-            }
-        }
-    }
+    /* NOTE: JMAP permission X- properties have been stripped in
+       either caldav_put() or sched_deliver_local() */
```
rsto, line 948: That still looks to have one regression that x-properties are getting not stripped for VTIMEZONE stripping in CalDAV GET:
https://github.com/cyrusimap/cyrus-imapd/blob/5520d37d63a8cd86ecaa63f9d4cb3604081e998f/imap/http_caldav.c#L2631-L2633

We load the iCalendar from disk, including its existing `X-JMAP` properties, then store it, which adds the same `X-JMAP` properties again in addition to the existing ones.

I don't think we'll hit that code-path as we don't strip timezones on GET due to our config, but it could happen for installations that do.

rsto (commented):
I think there is still one regression with timezone stripping on GET. I don't think that's a practical issue right now, as I doubt installations are using these JMAP properties and do timezone stripping, so I leave it to you to decide on the severity. I still don't understand why the code needed to get removed from `preserve_jmap_permissions`, though?

One other thing: please add a `changes` file that outlines that the X-JMAP-PRIVACY property got deprecated and installations that used must migrate their iCalendar data to CLASS themselves. There is no backwards-compatible handling in Cyrus.

ksmurchison:
> I think there is still one regression with timezone stripping on GET. I don't think that's a practical issue right now, as I doubt installations are using these JMAP properties and do timezone stripping, so I leave it to you to decide on the severity. 

I will look into this.

>I still don't understand why the code needed to get removed from `preserve_jmap_permissions`, though?

I removed it there because we are removing the privacy properties elsewhere and it seemed silly to attempt to remove them again.

> One other thing: please add a `changes` file that outlines that the X-JMAP-PRIVACY property got deprecated and installations that used must migrate their iCalendar data to CLASS themselves. There is no backwards-compatible handling in Cyrus.

Will do.

rsto (approved):
👍

PR #6130, opened 2026-07-06
provide JMAP state string versions of counter data in mbox event and UserState/get

rjbs:


ksmurchison (changes requested):
I think Claude made the first commit a little more complicated than it had to be.  `imap/jmap_data_types.c` is unnecessary and we no longer need to include `imap/jmap_data_types.h` in `jmap_api.c` since all of the hash functions are now called from `jmap_util.c`.  The attached patch takes care of these things:

[jmap_data_types.patch](https://github.com/user-attachments/files/29743771/jmap_data_types.patch)

rjbs:
Big ol' rebase and I think your comment has been addressed.

ksmurchison (approved):

PR #6131, opened 2026-07-07
twom read only consistency

brong:
Right now when you run a consistency check on a twom database, it needs to be able to write iff the last transaction aborted.  This updates the check to allow for a consistency check without actually doing the undo, while still making sure the rest of the links are good.

This PR also duplicates the twom.c file exactly from the standalone repo, including some apple code which we don't need for our Cyrus but makes the files sync up, so future conversion to the standalone library is safe.

### DIFF
--- a/lib/twom.c
+++ b/lib/twom.c
@@ -2432,8 +2432,15 @@ static int consistent1(struct twom_txn *txn, struct tm_loc *loc)
 
         uint8_t level = LEVEL(nextptr);
         for (i = 1; i < level; i++) {
-            /* check the old pointer was to here */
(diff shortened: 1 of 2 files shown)


ksmurchison (commented):
I don't really know the twom format well enough to evaluate this.  I trust that you know that this is correct.  Is there a way to test this?

brong:
yes!  @ksmurchison the test is in the twom repo though, rather than here in Cyrus.

https://github.com/brong/twom/commit/c962246881be4626b8c5d227526fffaa1807dc27

ksmurchison (approved):

PR #6132, opened 2026-07-07
quota: keep -J output valid JSON when repairs happen

brong:
Repair diagnostics from `quota -f` were printed as plain text to stderr, so capturing combined output (e.g. `quota -f -J 2>&1 | jq`) interleaved those lines with the JSON report on stdout and it no longer parsed.

In JSON mode, collect repair messages into a top-level "_repairs" array (leading underscore so consumers can skip metadata keys) instead of printing them; the non-JSON path still writes to stderr as before.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Quota.pm
+++ b/cassandane/Cassandane/Cyrus/Quota.pm
@@ -7,6 +7,7 @@ use warnings;
 use Cwd qw(abs_path);
 use DateTime;
 use Data::Dumper;
+use JSON::XS;
 
 use base qw(Cassandane::Cyrus::TestCase Cassandane::Mixin::QuotaHelper);
 use Cassandane::Util::Log;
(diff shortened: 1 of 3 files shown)


cassandane/Cassandane/Cyrus/Quota.pm
```diff
@@ -7,6 +7,7 @@ use warnings;
 use Cwd qw(abs_path);
 use DateTime;
 use Data::Dumper;
+use JSON;
```
rjbs: We should standardize on one JSON library.  It'll be JSON::XS.  I'll just make a new commit for this, though.  But feel free to amend this one so we have no race. ;-)

rjbs (commented):
LGTM.  Had to go review how variadic functions work!

cassandane/Cassandane/Cyrus/Quota.pm
brong: Fixed

rjbs (approved):
weird, surprised I didn't approve this on previous look

PR #6133, opened 2026-07-07
perl: use JSON::XS, not JSON

rjbs:
The speed is probably not a notable change (but I did not test), but consistency is nice.


rjbs:
I gotta fix this harder before it just works.

rjbs:
… in fact, going to take a totally different approach.  There are some weird incompatibilities!

PR #6134, opened 2026-07-08
jscalendar.c: don't crash reading iCal FEATURE param with unknown value(s)

ksmurchison:
icalparameter_enum_to_string() returns NULL on an unknown feature which when used in buf_setcstr() causes a crash


cassandane/tiny-tests/JMAPCalendars/calendarevent_get_virtuallocations_conference
```diff
@@ -9,13 +9,61 @@ sub test_calendarevent_get_virtuallocations_conference
...
+    $self->assert_num_equals(5, scalar (values %{$virtualLocations}));
 
     my $loc1 = $virtualLocations->{loc1};
-    $self->assert_str_equals('Moderator dial-in', $loc1->{name});
-    $self->assert_str_equals('tel:+123451', $loc1->{uri});
+    $self->assert_cmp_deeply(
+        superhashof({
+            name => 'Moderator dial-in',
+            uri => 'tel:+123451',
+            features => superhashof({
```
wolfsage: why superhashof for these? Shouldn't we know exactly which keys to accept?

wolfsage (approved):

cassandane/tiny-tests/JMAPCalendars/calendarevent_get_virtuallocations_conference
ksmurchison: @wolfsage I never remember which to use.  Should this just be hashof()?

ksmurchison: Actually, I left out @type so the test will fail unless I add it

ksmurchison: I removed superhashof() from features, but left it from the rest in case some other weird property shows up in the future

wolfsage: If you know exactly what's in the hash:
```
 $self->assert_cmp_deeply(
    superhashof({
        name => 'Moderator dial-in',
        uri => 'tel:+123451',
        features => {
            key => '...',
            key2 => '...',
        },
    }),

````

rsto (approved):

PR #6136, opened 2026-07-14
charset.testc: don't pass NULL pointers to memcmp

nwc10:


nwc10:
Currently we're only building with gcc's UBSAN in CI.

This is the only tweak needed to get the CUnit tests (also) passing with clang's UBSAN.

wolfsage (approved):
I do think it'd be good to have both gcc/clang asan/ubsan builds and also with varying options of -O3, etc... (it finds more problems by far)

cunit/charset.testc
```diff
@@ -1408,18 +1408,24 @@ static void test_utf8_to_searchform(void)
 static void test_charset_decode(void)
 {
 
+    /* clang's ASAN inserts a runtime check that __nonnull are indeed not NULL.
+     * glibc's <string.h> marks both pointers passed to memcmp() as __nonnull
+     * and this traps, even if the length is 0. gcc's ASAN is more forgiving.
+     * Ultimately we want the observed runtime behaviour of memcmp() - if length
+     * is 0 it returns 0, and ignores the two pointers, so we have to hard code
+     * a check on the length to get this to pass. */
```
elliefm: memcmp's behaviour is undefined when either of its arguments are NULL, so what we had was a bug, which clang's ASAN has detected for us, and you have fixed.  But the details of clang or gcc's ASAN behaviour aren't interesting here, what's interesting is that we corrected some undefined behaviour.

If it were me, I would discard this whole comment, and also rewrite the commit message to something like:

> charset.testc: don't pass NULL pointers to memcmp
> 
> encoded and decoded can both be NULL in this test, but memcmp's behaviour is undefined when either of its pointers are NULL

elliefm (changes requested):
The commit message (and PR title) are misleading.  I expected this to be making changes to `cunit/backend.testc`!

Please fix the commit message to name the module you're changing.  Something like "charset.testc: fix blah blah" or "charset: fix blah blah in tests".  If you wanted to be fancy you could say "cunit/charset.testc: fix blah blah", but it's not necessary (we know where testc files live), and we don't usually.

I have further comments about the way you've described the change (details follow), but those comments are _suggestions_, not _requested changes_.

cunit/charset.testc
nwc10: Thanks. I wasn't aware that `memcmp` was (strictly) undefined in that case. (I read some more, and I find threads grumbling about the C standard)

In trying to follow the thrust of the work "Commit Message vs Code Comments Style" document:

> Provide context to reaffirm why the code is built the way it is, especially when one suspects a future editor will try to simplify it into a non-working form

I'd prefer to keep something in a comment rather than a commit message, to stop any future me trying to "optimise" the code again. But I tried to tighten the wording, including why I had to grovel in other C source to work out why there was a problem at all.

nwc10:
> The commit message (and PR title) are misleading. I expected this to be making changes to `cunit/backend.testc`!

This is a complete brainfart on my part. Thank you for spotting it, and sorry for making you do work.

I can't remember for sure what happened, but I think I was trying to work out **what** the commit message shape should be, and trying to find similar previous cunit only commits to steal from. And clearly I found one, but didn't edit the most important part.

I have fixed both the commit and the PR title (which came from the commit) with your suggestion.

I hope this is now correct, but please re-check, as I'm not trusting myself this week.

cunit/charset.testc
elliefm: > I wasn't aware

In case you don't already know about it, https://en.cppreference.com/c is a fantastic resource for this sort of thing.  Tells you all the behaviours of everything, with the variations between C standard versions clearly marked.  Note that there's also an analogous C++ reference on the same site, so be sure you're reading the C version and not the C++ version (unless you're working on C++).  Here's what it says about C memcmp: https://en.cppreference.com/c/string/byte/memcmp

I wouldn't expect someone to try to re-simplify the conditions in this case.  It's clearly preventing calling `memcmp()` when we don't expect to have something to compare, and even if one believes that's safe to do, it's not the kind of optimisation one does without double checking (I assume).  If anything, I could see someone making the condition _stricter_ (such as actually checking for non-NULL, and not just non-zero length), but not sloppier.

In any case, if someone re-breaks it in the future, presumably we'll have cunit-ubsan in CI by then, and CI will complain about it. ;)

I think the advice re providing context is more about when there's multiple ways to do something, and why one way was chosen over another.  In this case, the way we were doing it was just wrong, and now it's not, I don't think it needs an explanatory comment.  I suppose you could comment about why you're checking the length only and not the pointer itself, but by the time you've written that comment you could've just written the extra check.

elliefm (approved):

PR #6137, opened 2026-07-14
http_caldav_sched.c: improve logging for ignored scheduling objects

rsto:
This improves logging for the case where Cyrus decides to not send an iTIP message although the calendar object is scheduled. This does not change existing functionality but is meant to support debugging.

The following cases now log with LOG_WARNING level:

- The CY:scheduling-enabled annotation is false for the user.
- The calendar user is neither the ORGANIZER nor an ATTENDEE.
- The Scheduling Outbox lacks DACL_INVITE or DACL_REPLY, for invites and replies respectively.

Tested by:
- Caldav.schedule_organizer_mismatch
- Caldav.scheduling_enabled
- Caldav.scheduling_privileges
- JMAPCalendars.calendar_scheduling_enabled
- JMAPCalendars.calendarevent_set_schedule_organizer_mismatch

### DIFF
--- a/imap/http_caldav_sched.c
+++ b/imap/http_caldav_sched.c
@@ -2507,7 +2507,7 @@ void sched_request(const char *cal_ownerid, const char *sched_userid,
 
     if (!(rights & DACL_INVITE)) {
         /* DAV:need-privileges */
-        syslog(LOG_DEBUG,
+        syslog(LOG_WARNING,
                "No schedule-send-invite privilege for user %s on Outbox %s",
                httpd_userid, organizer);
 
@@ -2600,6 +2600,7 @@ struct reply_data {
     int master_send;
     int do_send;
     icalparameter_scheduleforcesend force_send;
+    bool found_attendee;  /* one of our schedule addresses is an attendee */
 };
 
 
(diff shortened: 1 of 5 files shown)


imap/http_caldav_sched.c
```diff
@@ -3028,6 +3040,20 @@ void sched_reply(const char *cal_ownerid, const char *sched_userid,
         if (reply.itip) icalcomponent_free(reply.itip);
     }
 
+    /* If the event has attendees but none of our schedule addresses is one of
+       them then we are neither the organizer nor an attendee and nothing gets
+       scheduled, despite this being a scheduling object. */
+    icalcomponent *comp =
+        icalcomponent_get_first_real_component(newical ? newical : oldical);
+    if (!found_attendee && comp && icalcomponent_get_first_invitee(comp)) {
```
wolfsage: IMO I'd only call icalcomponent_get_first_real_component if found_attendee is false and avoid it otherwise (which should be the more common case) but it probably doesn't matter

wolfsage (approved):

imap/http_caldav_sched.c
rsto: Right, I did that and also rewrote the logger to use `xsyslog_ev` and made it log more debugging output

PR #6138, opened 2026-07-14
AccountSync: use external altnamespace names in dump/undump

brong:
dump_user/undump_user previously used Cyrus::Mbname's userfolder / new_userfolder pair, which represent mailboxes in internal form with a leading "INBOX" component and "." hierarchy separator (e.g. "INBOX.Archive"). This does not match the external altnamespace + unixhierarchysep naming that Apple (and other JMAP implementations) use for account dumps, so dumps could not interoperate and hand-authored external-format data failed to restore ("Unknown top-level Archive").

Switch to the extuserfolder / new_extuserfolder pair so mailbox names in the dumped data use the external representation: "INBOX" for the inbox, bare names for children ("Archive", "Sent"), and "/" for nested hierarchy ("Parent/Child"). The pair round-trips symmetrically, so dump -> undump -> dump is unchanged.

Extend the SyncProto dump/restore test to assert the external name format in addition to the existing round-trip check.

### DIFF
--- a/cassandane/tiny-tests/SyncProto/syncproto_dump_restore
+++ b/cassandane/tiny-tests/SyncProto/syncproto_dump_restore
@@ -25,6 +25,11 @@ sub test_syncproto_dump_restore ($self)
     my $sp = Cyrus::SyncProto->new($admintalk);
     my $as = Cyrus::AccountSync->new($sp);
     my $data = $as->dump_user(username => 'cassandane');
+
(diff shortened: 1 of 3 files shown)


rjbs:
At the very least, tests need updating for this change.  Guessing that's the problem, but didn't look deep.  May come back to this a bit later today and patch myself, but you may see this first and go "oh yeah" and push fix?  🤷🏽

rjbs (approved):

PR #6140, opened 2026-07-16
master-cronevent.testc: Avoid a fatalized format warning with gcc's ubsan

nwc10:
With the optimizer enabled, gcc -fsanitize=undefined issues a format warning. This is fatal with -Werror, stopping the build dead:

    error: '%d' directive output may be truncated writing between 1 and 11 bytes into a region of size 10 [-Werror=format-truncation=]
    note: directive argument in the range [-2147483647, 2]

This inference is actually incorrect - the range is [0, 2] hence "%d" can't overflow. The "obvious" fix is to increase the buffer length. A better fix is to change the variable to unsigned int and the format to "%u", as it's used as an array index. Using (default) signed variables for array indexes is an anti-pattern that can cause CVE-level bugs.

### DIFF
--- a/cunit/master-cronevent.testc
+++ b/cunit/master-cronevent.testc
@@ -199,17 +199,16 @@ static void test_cronevent_add_broken_command(void)
     };
     const size_t n_tests = sizeof(tests) / sizeof(tests[0]);
     dynarray_t *schedule, *details;
-    int i;
 
     cronevent_get_schedule(&schedule, &details);
 
-    for (i = 0; i < (int) n_tests; i++) {
+    for (size_t i = 0; i < n_tests; i++) {
         char name[16];
         const char *spec = "* * * * *";
         struct cronevent_details *event_details;
         unsigned j;
 
-        snprintf(name, sizeof(name), "event %d", i);
+        snprintf(name, sizeof(name), "event %zu", i);
 
         cronevent_add(name, spec, tests[i].cmd, false);
         assert_schedule_invariants();


dilyanpalauzov:
Can the type of `i` be changed to `size_t`, the cast on the `for`-line skipped, for printing `%zu` used, and the variable `i` declared on the `for`-line?

wolfsage (approved):

cunit/master-cronevent.testc
```diff
@@ -199,17 +199,17 @@ static void test_cronevent_add_broken_command(void)
     };
     const size_t n_tests = sizeof(tests) / sizeof(tests[0]);
     dynarray_t *schedule, *details;
-    int i;
+    unsigned int i;
 
     cronevent_get_schedule(&schedule, &details);
 
-    for (i = 0; i < (int) n_tests; i++) {
+    for (i = 0; i < (unsigned int) n_tests; i++) {
```
elliefm: The cast was only necessary because `i` was signed and `n_tests` is unsigned, and comparing signed and unsigned is a warning.  Now they're both unsigned, so the cast is unnecessary.  The fact the types are different sizes doesn't matter, automatic integer promotion take care of that already.  In fact, the cast is kinda misleading now, because it kinda looks like some kind of size coercion, but _it's not_.  I would just remove the cast.  In fact, if you `ack 'i < n_tests'` you can see this pattern all over the place, castless, with `unsigned i`.

I have no idea why I made this one `int` instead of `unsigned`, that's a big deviation from my usual style.  Perhaps an earlier design needed the signedness and I forgot to clean it up when it changed?  Wait, I see it! It's because this test is using the `dynarray` API, which uses `int` lengths.  If you look at the whole file, the tests that use `dynarray` have `int i`, and the tests that don't have `unsigned i`.

So actually, the bug here is that the buffer is arguably too small, like it's complaining about. (Except for the fact that `i` is `0..2`, and even in the most ambitious testing scenario will probably never exceed maybe a dozen).  I would just make the buffer bigger, leave everything else as it was.

cunit/master-cronevent.testc
```diff
@@ -199,17 +199,17 @@ static void test_cronevent_add_broken_command(void)
...
         char name[16];
```
elliefm, line 206:
```suggestion
        char name[20];
```

Is the entire fix I would've made here.

PR #6144, opened 2026-07-18
docsrc: more restructuring, this time Cassandane docs

rjbs:
This is more "rework the docs to be more navigable".  Also, to create better incentive structures for writing docs in the right place.  This PR comprises:

* converting testrunner.pl to GLD so it gets a usage message; adding Pod to it
* remove that content from the docsrc; you don't need it to just run/write quick tests
* update the cassandane.rst with more focus on "writing and running tests, the common day-to-day"
* ❗️publishing any present Cassandane Pod documentation to the cassandane page (later, probably want to expand this)

The changes I made to the Makefile.am feel a bit monstrous, but I wasn't sure whether (a) I could make them much better or (b) it was worth doing anyway.

There's a lot left to do, but I think this is a good increment to merge.  Likely future work:

* document more of Cassandane's perl
* (especially if we follow through on merging in Test::Unit)

Maybe significantly later:
* render non-Cassandane perl library docs, like Cyrus::IndexFile

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1991,27 +1991,30 @@ docsrc/reference/manpages/usercommands/sieveshell.rst: $(top_srcdir)/perl/sieve/
 # carries Pod, plus a cross-reference file (so L<...> between them renders as an
 # internal :doc: link rather than a metacpan link) and a listing that
 # developer/cassandane.rst includes.  It's all a build artifact, gitignored
-# like the other perl2rst output.  The module list is shared with docsrc/Makefile:
+# like the other perl2rst output.  The module list is discovered at make time
+# (see docsrc/cassandane-pod.mk), shared with docsrc/Makefile.
+cassandane_srcdir = $(top_srcdir)/cassandane
 include docsrc/cassandane-pod.mk
 
 cassandane_apidir  = docsrc/developer/cassandane-api
 cassandane_pod_rst = $(CASSANDANE_POD:%=$(cassandane_apidir)/%.rst)
+cassandane_pod_pm  = $(CASSANDANE_POD:%=$(cassandane_srcdir)/%.pm)
 
 # module -> doc-page map, read by perl2rst to resolve internal L<...> links
-$(cassandane_apidir)/.xref: $(top_srcdir)/tools/pod-index Makefile
+$(cassandane_apidir)/.xref: $(top_srcdir)/tools/pod-index $(cassandane_pod_pm)
 	@$(MKDIR_P) $(@D)
 	$(AM_V_GEN)$(PERL) $(top_srcdir)/tools/pod-index --xref $(CASSANDANE_POD) >$@.NEW && mv $@.NEW $@
 
 # one page per module, from its .pm and the shared xref
-$(cassandane_apidir)/%.rst: $(top_srcdir)/cassandane/%.pm $(cassandane_apidir)/.xref $(top_srcdir)/tools/perl2rst
+$(cassandane_apidir)/%.rst: $(cassandane_srcdir)/%.pm $(cassandane_apidir)/.xref $(top_srcdir)/tools/perl2rst
 	@$(MKDIR_P) $(@D)
 	$(AM_V_GEN)PERL2RST_NAMESPACE=Cassandane PERL2RST_XREF_FILE=$(cassandane_apidir)/.xref \
 	    $(PERL) $(top_srcdir)/tools/perl2rst $(subst /,::,$*) <$< >$@.NEW && mv $@.NEW $@
 
 # the reST bullet list + hidden toctree that developer/cassandane.rst includes
-$(cassandane_apidir)/listing.inc: $(top_srcdir)/tools/pod-index Makefile $(CASSANDANE_POD:%=$(top_srcdir)/cassandane/%.pm)
+$(cassandane_apidir)/listing.inc: $(top_srcdir)/tools/pod-index $(cassandane_pod_pm)
 	@$(MKDIR_P) $(@D)
-	$(AM_V_GEN)$(PERL) $(top_srcdir)/tools/pod-index --listing --srcdir $(top_srcdir)/cassandane $(CASSANDANE_POD) >$@.NEW && mv $@.NEW $@
+	$(AM_V_GEN)$(PERL) $(top_srcdir)/tools/pod-index --listing --srcdir $(cassandane_srcdir) $(CASSANDANE_POD) >$@.NEW && mv $@.NEW $@
 
 $(cassandane_apidir)/.stamp: $(cassandane_pod_rst) $(cassandane_apidir)/listing.inc
 	@touch $@
--- a/Makefile.am
+++ b/Makefile.am
@@ -1986,6 +1986,31 @@ docsrc/reference/manpages/usercommands/sieveshell.rst: $(top_srcdir)/perl/sieve/
 	@$(MKDIR_P) $(@D)
 	$(AM_V_GEN)$(PERL) $(top_srcdir)/tools/perl2rst sieveshell < $< > $@.NEW && mv $@.NEW $@
 
+# Render the Pod of every Cassandane module that carries any into reST, one
+# page per module, so it joins the doc site.  The pages are build artifacts,
+# gitignored like the other perl2rst output.  We build the ridiculous xref file
+# so that when we render the Pod into reST, we can know which L<...> sequences
+# should link internally, and which should go to metacpan.
+docsrc/developer/cassandane-api/.stamp: $(top_srcdir)/tools/perl2rst
+	@$(RM) -r $(@D)
+	@$(MKDIR_P) $(@D)
+	$(AM_V_GEN)pms=`cd $(top_srcdir) && grep -rl '^=head' cassandane/Cassandane --include='*.pm' | sort`; \
+	: > $(@D)/.xref; \
+	for pm in $$pms; do \
+	    mod=`echo "$$pm" | sed -e 's|^cassandane/||' -e 's|\.pm$$||' -e 's|/|::|g'`; \
+	    out=`echo "$$mod" | sed -e 's|::|-|g'`; \
+	    printf '%s\t/developer/cassandane-api/%s\n' "$$mod" "$$out" >> $(@D)/.xref; \
+	done; \
+	for pm in $$pms; do \
+	    mod=`echo "$$pm" | sed -e 's|^cassandane/||' -e 's|\.pm$$||' -e 's|/|::|g'`; \
+	    out=`echo "$$mod" | sed -e 's|::|-|g'`; \
+	    PERL2RST_NAMESPACE=Cassandane PERL2RST_XREF_FILE=$(@D)/.xref \
+	        $(PERL) $(top_srcdir)/tools/perl2rst "$$mod" \
+	        < $(top_srcdir)/$$pm > $(@D)/$$out.rst.NEW \
+	        && mv $(@D)/$$out.rst.NEW $(@D)/$$out.rst; \
+	done
+	@touch $@
+
 %.1: man/.sphinx-build.stamp
 	@$(MKDIR_P) $(@D)
 	$(AM_V_GEN)touch $@
@@ -2019,7 +2044,7 @@ SPHINX_OPTS = -d $(SPHINX_CACHE) -n -q $(SPHINX_FAIL_ON_WARNINGS) -j auto
 ## detect when source directory is not build directory (i.e. VPATH
 ## build), and clone the docsrc tree into the build directory, so
 ## that we have a single "source directory" for sphinx-build to use.
-docsrc/.sphinx-build.stamp: docsrc/reference/manpages/configs/imapd.conf.rst docsrc/reference/manpages/usercommands/cyradm.rst docsrc/reference/manpages/usercommands/sieveshell.rst
+docsrc/.sphinx-build.stamp: docsrc/reference/manpages/configs/imapd.conf.rst docsrc/reference/manpages/usercommands/cyradm.rst docsrc/reference/manpages/usercommands/sieveshell.rst docsrc/developer/cassandane-api/.stamp
 	$(AM_V_GEN)test x"$(top_srcdir)" = x"$(top_builddir)" || \
         (cd $(top_srcdir) && tar cf - --mode=gu+w docsrc doc/examples) | tar xf -
 	@touch $@
--- a/Makefile.am
+++ b/Makefile.am
@@ -539,6 +539,7 @@ dist_noinst_SCRIPTS = \
     tools/masssievec \
     tools/mknewsgroups \
     tools/perl2rst \
+    tools/pod-index \
     tools/translatesieve \
     tools/update-tzdata.sh
 noinst_MAN = \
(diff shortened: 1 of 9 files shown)


cassandane/testrunner.pl
```diff
@@ -158,130 +152,98 @@
...
+
+    [ 'format|f=s@',   "test report format, repeatable: xml, tap, pretty, or"
+                     . " prettier (default: prettier)" ],
+    [ 'list|l+',       "list matching tests instead of running them; repeat"
+                     . " (-ll) to list individual tests, not just suites" ],
+    [],
+    [ 'jobs|j=i',      "run this many test workers in parallel" ],
+    # The default really should be --no-keep-going like make
+    [ 'keep-going|k!', "keep going after a test fails (default: yes)",
+                       { default => 1 } ],
```
elliefm: Come to think of it, I disagree with the old comment that "The default should really be --no-keep going like make".  This is not a builder, it's a tester.  If one thing failed, you can't address it correctly without also seeing _what else failed_.  Stop-at-first-failure is only useful as a fast go/no-go switch, and we don't use such a thing anywhere.  I think the behaviour we have is the behaviour we want, and we can throw out that old, incorrect comment.

cassandane/testrunner.pl
```diff
@@ -2,6 +2,163 @@
...
+
+Cassandane runs out of its own directory; nothing gets installed.  Run this
+script with the F<cassandane/> directory as your working directory.
+
+=item *
+
+The Cyrus code under test must run as the superuser or as the C<cyrus> user.
+You invoke C<testrunner.pl> as yourself; if you're root it steps down to
+C<cyrus>, and otherwise it re-invokes itself under C<sudo>.
+
```
elliefm, line 45: Maybe we want to mention something about setting up your sudoers file appropriately, so that it can do this.  I think that's documented elsewhere, but this would be a good spot to mention it, if not thoroughly describe it.

These days you don't "usually" need to think about it, since cyd/dar/docker take care of that detail, but if you're in hard-hard mode (no docker container, or a broken docker container), that's when you'll need to know it.

cassandane/testrunner.pl
```diff
@@ -2,6 +2,163 @@
...
+=item B<-c>, B<--cleanup>[=I<yes|no|pre|post>]
+
+Clean up C<$rootdir> leftovers before the run and after each passing test.
+Useful when disk (e.g. a tmpfs) is tight.  Bare C<--cleanup> means C<yes>.  If
+you want this most of the time, set C<cassandane.cleanup> in F<cassandane.ini>
+and use C<--no-cleanup> to override it.
+
+=item B<-j>, B<--jobs> I<N>
+
+Run I<N> test workers in parallel.
```
elliefm: I would probably mention the `cassandane.maxworkers` setting here.  You almost always want to just set a default that works for your number of CPU cores, not have to type `--jobs=12` or whatever every time.  And then you use this switch to override that if you want something different for one run.

Lots of the command line options work like this (that is, they override some cassini setting), so maybe the usage output should name which ones they override.  That would aid casual discoverability of the cassini settings; otherwise you have to read cassandane.ini.example to know what exists.

docsrc/developer/cassandane.rst
```diff
@@ -534,3 +199,137 @@ ContactCard/query response.
...
 The rest of the test is more of the same.
 
+Assertions
+^^^^^^^^^^
+
+A test passes unless it dies or an assertion fails, so assertions are how you
+state what "correct" means.  The ones you'll reach for most often come from
+Test::Unit:
+
+``assert($bool)``, ``assert_str_equals($expect, $got)``,
```
elliefm: We should remove `assert($bool)` from this list, because it produces rubbish output on failure unless you took the time to specify a custom failure message, and nobody ever remembers to do that.  It's utility is as a low level function for building higher level assertions on top of.

docsrc/developer/cassandane.rst
```diff
@@ -534,3 +199,137 @@ ContactCard/query response.
...
+``assert_num_equals($expect, $got)``, ``assert_null`` / ``assert_not_null``,
+and ``assert_matches($regex, $string)`` for scalars; and
```
elliefm: You should should shout out the `assert_matches()` complement `assert_does_not_match()` here, because its name is weird, and so it's not readily discoverable -- searching the code for the obvious "assert_not_matches" finds nothing.

docsrc/developer/cassandane.rst
```diff
@@ -534,3 +199,137 @@ ContactCard/query response.
...
+of them); suites inheriting directly from ``Cassandane::Unit::TestCase`` ignore
+them, but you shouldn't be inheriting from that unless you're testing
+Cassandane itself.
+
+The ones you'll meet first are the ``:needs_*`` family, which skip a test
+unless Cyrus was built with some capability:
+
+* ``:needs_component_NAME`` — a Cyrus component (e.g. ``httpd``) is enabled.
+* ``:needs_dependency_NAME`` — a compiled-in library (e.g. ``icalvcard``) is
+  present.
```
elliefm: I would spell these as `:needs_CATEGORY_KEY`, and then shout out `cyr_buildinfo(8)`, which is what actually defines the categories and keys.

docsrc/developer/cassandane.rst
```diff
@@ -534,3 +199,137 @@ ContactCard/query response.
...
+
+If you need to branch on the Cyrus version *inside* a test (or inside
+infrastructure) rather than skip the whole thing, ``Cassandane::Instance``
+offers ``get_version()``, which can detect versions as far back as 2.5.0:
+
+.. code:: perl
+
+    my ($major, $minor, $revision, $extra) = Cassandane::Instance->get_version()
+
+The skip handling itself lives in ``Cassandane::Test::Skip``.
```
elliefm: `Cassandane::Test::Skip()` just tests that the version guards are working as expected (naively!), it doesn't have any interesting logic.

The skip handling proper lives in `Cassandane::Unit::TestCase::filter()`, which defines and implements all the available filters.  `testrunner.pl` chooses which filters to actually use (look for its `@filters` array).  Most are always used, some are conditional.

Makefile.am
```diff
@@ -1955,6 +1955,31 @@ docsrc/reference/manpages/usercommands/sieveshell.rst: $(top_srcdir)/perl/sieve/
...
 
+# Render the Pod of every Cassandane module that carries any into reST, one
+# page per module, so it joins the doc site.  The pages are build artifacts,
+# gitignored like the other perl2rst output.  We build the ridiculous xref file
+# so that when we render the Pod into reST, we can know which L<...> sequences
+# should link internally, and which should go to metacpan.
+docsrc/developer/cassandane-api/.stamp: $(top_srcdir)/tools/perl2rst
+	@$(RM) -r $(@D)
+	@$(MKDIR_P) $(@D)
+	$(AM_V_GEN)pms=`cd $(top_srcdir) && grep -rl '^=head' cassandane/Cassandane --include='*.pm' | sort`; \
```
elliefm: Have a look at #6116, in particular de006f956a and 935b057501, for an example of how I recently removed a lot of nasty multiline bash like this from the check recipes.

That said, even if you moved all this logic into a perl script, this looks like bad abuse of the make system.  There might be subtleties that I haven't appreciated yet, but at first blush it looks like you should do something like:

1. Set (or compute) the list of files this should apply to.  Maybe have a script generate the list, and have configure run the script and AC_SUBST the result, if you don't want to list them explicitly in Makefile.am.  But the simplest thing to do is just list them explicitly in a variable, which you can generate later once the rest is working.
2. Set up a make rule to build an xref file from this list
3. Set up a make pattern rule that produces an rst from a pm file plus the xref file
4. Set up a rule that depends on the rst output files (this might be a .stamp file)
5. Make that rule a prerequite of one of the doc targets

Makefile.am
```diff
@@ -1965,6 +1965,7 @@ docsrc/developer/cassandane-api/.stamp: $(top_srcdir)/tools/perl2rst
 	@$(MKDIR_P) $(@D)
 	$(AM_V_GEN)pms=`cd $(top_srcdir) && grep -rl '^=head' cassandane/Cassandane --include='*.pm' | sort`; \
 	: > $(@D)/.xref; \
+	: > $(@D)/summary.inc; \
```
elliefm: Under the scheme from my previous comment, this would become another make rule that just builds a summary file from the list, and the pattern rule that builds rst from pm would also depend on the summary file.

elliefm (changes requested):
The Makefile.am stuff is _so bad_.  If you merge this as is, you're creating housekeeping work for me later.  And having to clean up after Claude will piss me off.

I have a few notes on the documentation itself, but it's largely fine.

Makefile.am
rjbs: Thank, I'll see what I can do!  I also hated this, but in some ways, this level of the code is where once I hit "it works", I am not well equipped for "what would make it less crap" — so this advice is really appreciated!  I had considered the "shove into perl", but came to the same conclusion you make above.

Looking at the process you describe is on my todo now, but might wait until next week, as I'm keeping busy at IETF this week .

docsrc/developer/cassandane.rst
elliefm: Ah, and `Cassandane::Cyrus::TestCase::filter()` extends that.  Anyway yeah, all the logic for determining whether a requested test should actually run or not happens via the Test::Unit filter mechanism, our TestCase subclasses provide the filters, and testrunner chooses which filters to activate.

docsrc/developer/cassandane.rst
elliefm: If it's interesting, `:needs_...` is implemented by the `skip_missing_features` filter in `Cassandane::Unit::TestCase::filter()`.

docsrc/developer/cassandane.rst
elliefm: (Oh -- the `x` filter is an infrastructure hack; the rest are actually filters.)

cassandane/testrunner.pl
rjbs: Removed!

docsrc/developer/cassandane.rst
rjbs: 🌌 🧠 :  What if we make ->assert *require* a second argument, and fix up all its callers to provide one?  Then it's not so bad.

Makefile.am
rjbs: Okay, I finally came back to this.  I knew it was bad, but I didn't know how to make it better, and with your guidance, some docs, and some robo-consultation, I think I've got something that is _actually not bad_!  Like, it's bad, because make is bad, but it seems like pretty good use of make, and make is good in some ways, too!

So!  I think I won't have time before your next shift to address the rest of the easy, small comments, so I'm making this comment now so you can (I hope) review just the three commits that fix this up.  If this needs more work, cool cool.  But if not, I'll feel great knowing the rest of the stuff seems like small potatoes.

Thanks for the pushback!

docsrc/developer/cassandane.rst
elliefm: If we override `assert`, doesn't that become an inheritance chain nuisance for all the specific assertions that are built on top of it?

If we're going to change all the callers to provide an extra argument anyway, we could just as well change them to use a more specific assertion instead.  And then people would stop copy-pasting `assert` into new tests.

Mandating a second argument would at least prevent us writing `$self->assert($foo);` without also being forced to compose a description.  But bugs, like `$self->assert(@foo);` (or any other expr that has multiple values in list context) would continue to slip past with the second argument requirement satisfied by a junk message, just as they do already.  I don't know how many of these bugs exist, so I expect we'll find at least some when we change all the callers.

🌌🌌🧠: What if we make attempts to use `assert` from one of our TestCases a _compilation failure_?  I imagine we'd do this by overriding it in Cassandane::Unit::TestCase with something that refuses, but I'm not sure about the perl specifics to make that happen; and the weird upstream Test::Unit object model mess might make this unworkable for some other reason anyway.  But if we _could_ do this, it would prevent us ever lazily using `assert` directly in a test.  We would need to rewrite our wrappers to use something like `$self->fail($msg) unless $expr;` internally, instead of `$self->assert($expr, $msg)`, but that doesn't seem like a downside.

elliefm:
(The CI failure was due to the new test that sometimes runs too fast for itself.  I've restarted the failed job.  #6205 is on master now, so that shouldn't happen anymore.)

elliefm (commented):
Oh yeah, this is looking much better!  One more suggestion though:

Expanding `$(CASSANDANE_POD)` into arguments to the perl tools runs the risk of the command line becoming too long for the shell if we someday have a lot of cassandane API documentation.  I think using a variable here was my suggestion originally, sorry! 😅

Given that CASSANDANE_POD comes from the make fragment, which we must maintain or generate, we might as well do so as a data file rather than a make fragment.  Have pod-index take an input file argument to get its work list from rather than ARGV, and have make pass it the data file.  No more shell limits.

That does make computing the `cassandane_pod_rst` and `cassandane_pod_pm` dependency lists a little more complex, they'll need to transform records from the data file instead of transforming tokens in the variable, but I think that's doable.

rjbs:
> Expanding $(CASSANDANE_POD) into arguments to the perl tools runs the risk of the command line becoming too long for the shell if we someday have a lot of cassandane API documentation. I think using a variable here was my suggestion originally, sorry! 😅

I think I'd like to respectfully decline this one.  Linux ARG_MAX is like 2 megs.  FreeBSD, about the same, I think.

I'm going to look into whether automake/make make it effectively less, though.  But if it's going to work, I'd rather leave it simple.  I'll get back to you…

docsrc/developer/cassandane.rst
rjbs: Short version:  We can't really make it a compilation failure.  Method dispatch is highly dynamic and will "always" survive until runtime.

For now, I'll remove bare assert and we can come back for another round on this.

docsrc/developer/cassandane.rst
rjbs: ✅

docsrc/developer/cassandane.rst
rjbs: Thanks, that was useful and I have redocumented this.

rjbs:
I went and did the math on the "the risk of the command line becoming too long for the shell if we someday have a lot of cassandane API documentation".  FreeBSD is much more conservative than Linux as far as ARG_MAX goes -- around 256KiB, at the small end.  So we have room for about 8,000 module names, and are using about 3% of the space.  I'm going to skip the added complexity.  I had to learn more make just to get where it is now!

cassandane/testrunner.pl
rjbs: Added.

docsrc/developer/cassandane.rst
rjbs: I realized that the guts here are not something we should be pointing people to.  I think they cross the "you should just go RTFS" barrier and pointers nudge the wrong way.  Removed that line.

rjbs:
@elliefm I now believe this is ready to merge.

docsrc/developer/cassandane.rst
```diff
@@ -206,10 +206,13 @@ A test passes unless it dies or an assertion fails, so assertions are how you
...
-``assert($bool)``, ``assert_str_equals($expect, $got)``,
-``assert_num_equals($expect, $got)``, ``assert_null`` / ``assert_not_null``,
-and ``assert_matches($regex, $string)`` for scalars; and
-``assert_deep_equals($expect, $got)`` for nested data structures.
+* ``assert_str_equals($expect, $got)``
+* ``assert_num_equals($expect, $got)``
+* ``assert_null($value)``
+* ``assert_not_null($value)``
+* ``assert_matches($regex, $string)``
+* ``assert_not_matches($regex, $string)``
```
elliefm: You've made the same mistake everyone else does, and then chokes on:  the pair for `assert_matches()` is **not** `assert_not_matches()`, it is `assert_does_not_match()`.  `assert_not_matches()` doesn't exist.

It's documented in the pod for `Test::Unit::Assert`, but you'd be forgiven for not knowing to look there since the pod for `Test::Unit` doesn't so much as mention that it exists...

docsrc/developer/cassandane.rst
```diff
@@ -635,3 +199,158 @@ ContactCard/query response.
 
 The rest of the test is more of the same.
 
+Assertions
+^^^^^^^^^^
+
+A test passes unless it dies or an assertion fails, so assertions are how you
+state what "correct" means.  The ones you'll reach for most often come from
+Test::Unit:
```
elliefm: On further reflection, I think I'd shout out `Test::Unit::Assert` specifically here, since its pod actually documents all the assertions, whereas `Test::Unit`'s pod is practically empty.

```suggestion
A test passes unless it dies or an assertion fails, so assertions are how you
state what "correct" means.  The ones you'll reach for most often come from
``Test::Unit::Assert``:
```

I suppose we could even link to https://metacpan.org/pod/Test::Unit::Assert here, I dunno.

docsrc/developer/cassandane.rst
```diff
@@ -635,3 +199,158 @@ ContactCard/query response.
...
+
+* ``assert_str_equals($expect, $got)``
+* ``assert_num_equals($expect, $got)``
+* ``assert_null($value)``
+* ``assert_not_null($value)``
+* ``assert_matches($regex, $string)``
+* ``assert_not_matches($regex, $string)``
```
elliefm: You've made the very mistake that I requested we document correctly in order to prevent! 😅   The complement for `assert_matches()` is **not** `assert_not_matches()`, it is **`assert_does_not_match()`**.  `assert_not_matches()` doesn't exist.

Yeah, this is a ridiculous naming choice; `Test::Unit` strikes again.  The pair are documented in the pod for `Test::Unit::Assert`, but no-one looks there, since the pod for `Test::Unit` doesn't so much as mention that that subpackage exists...

(In practice, we actually override both `assert_matches()` and `assert_does_not_match()` in `Cassandane::Unit::TestCase` to provide better failure messages.  We also provide the analogously named `assert_syslog_matches()` and `assert_syslog_does_not_match()` in `Cassandane::Cyrus::TestCase` for encapsulating the complexities of examining syslog.)

```suggestion
* ``assert_matches($regex, $string)``
* ``assert_does_not_match($regex, $string)``
```

elliefm (approved):
One mistake that needs to be fixed; and one detail that could be improved with a slight tweak.  Otherwise, this looks good.

I'm approving this now because I don't think it needs to go through another review cycle once the mistake is fixed, but please do actually fix it!  Thanks.

docsrc/developer/cassandane.rst
rjbs: How embarrassing.  I have fixed this.

PR #6147, opened 2026-07-18
cassandane: make per-TestCase "default using" more consistent

rjbs:


cassandane/Cassandane/Instance.pm
```diff
@@ -3253,22 +3255,33 @@ sub _common_http_service_args ($self) {
...
+        https://cyrusimap.org/ns/jmap/performance
+        https://cyrusimap.org/ns/jmap/backup
+        urn:ietf:params:jmap:blob
+    ) ];
+}
+
+sub _jmap_default_using ($self) {
+    unless ($self->{test_case}) {
+      return $self->jmap_default_default_using;
+    }
```
wolfsage, line 3280: I think this "unless..." block is redundant with the ternary below

wolfsage (approved):
One thing to cleanup but looks good

cassandane/Cassandane/Instance.pm
rjbs: Oh yeah I was going to check ->can and then decided to make a stock implementation.  Fixed.

PR #6148, opened 2026-07-19
Cass TestEntity: Add a '->destroy' method

wolfsage:
This makes it easy to destroy objects without having to construct the JMAP calls by hand.

Because the entity objects use non standard jmap extensions we must enable non standard exceptions in the suite(s) using them. For now I've fixed up Cassandane::Cyrus::JMAPMailbox only so I could convert a test to ensure this new method works.

### DIFF
--- a/cassandane/Cassandane/TestEntity/Role/Factory.pm
+++ b/cassandane/Cassandane/TestEntity/Role/Factory.pm
@@ -91,5 +91,25 @@ sub _update {
     return;
 }
 
+sub _destroy {
+    my ($self, $instance, @rest) = @_;
+    @rest && Carp::confess("too many arguments to destroy");
+
+    my $dt = $self->datatype;
+    my $id = $instance->id;
+
+    my $jmap = $self->user->entity_jmap;
+
+    my $res = $jmap->request([[
+        "$dt/set",
+        { destroy => [ $id ] },
+        'FactorySetDestroy',
+    ]]);
+
+    $res->single_sentence("$dt/set")->as_set->assert_no_errors;
+
+    return;
+}
+
 no Moo::Role;
 1;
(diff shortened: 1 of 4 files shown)


cassandane/Cassandane/TestEntity/Role/Factory.pm
```diff
@@ -91,5 +91,26 @@ sub _update {
...
+        { destroy => [ $id ] },
+        'FactorySetDestroy',
+    ]]);
+
+    $res->single_sentence("$dt/set")->as_set->assert_no_errors;
+
+    return;
+}
+
+
```
rjbs: michaelspace

cassandane/Cassandane/TestEntity/Role/Factory.pm
```diff
@@ -91,5 +91,26 @@ sub _update {
     return;
 }
 
+sub _destroy {
+    my ($self, $instance, @rest) = @_;
+    @rest && Carp::confess("too many arguments to destroy");
```
rjbs, line 96: why not use a subroutine signature, which would enforce arity for you?

rjbs (approved):

cassandane/Cassandane/TestEntity/Role/Factory.pm
wolfsage: (consistency with the rest of this file which should be updated at some point). Maybe the whole test entity system could be updated. Unsure but will not do that here

cassandane/Cassandane/TestEntity/Role/Factory.pm
wolfsage: Bah. Fixed

PR #6149, opened 2026-07-19
sync_apply_unuser: don't clobber a renamed user's subscriptions

brong:
Subscriptions (and all other per-user files) are keyed by the INBOX uniqueid, which is unchanged across a user rename, so the old and new userids share one physical subs file.  A user rename logs USER <new> followed by UNUSER <old>, and rolling replication runs USER before UNUSER, so on the replica sync_apply_unuser(<old>) runs after the new user's subscriptions have already been synced in.

Its "nuke subscriptions" loop ran unconditionally, and because the defunct old userid resolves (via the name_history fallback in mboxname_conf_getpath) onto the shared uniqueid-keyed subs file, it deleted the new user's freshly-synced subscriptions.  The subs then only reappeared on the next full "sync_client -A", which never issues UNUSER for the vanished old user.

user_deletedata(), which wipes every other per-user file by uniqueid, was already gated on the tombstone predicate and so was correctly skipped for a renamed-away modern-dirs inbox; only the separate subs nuke sat outside that guard.  Hoist the inbox mbentry lookup above the nuke and gate the nuke on the same condition.

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -4654,18 +4654,30 @@ static int sync_apply_unuser(struct dlist *kin, struct sync_state *sstate)
 
     user_nslock_t *user_nslock = user_nslock_lock_w(userid);
 
-    /* Nuke subscriptions */
-    /* ignore failures here - the subs file gets deleted soon anyway */
-    strarray_t *list = mboxlist_sublist(userid);
-    for (i = 0; i < list->count; i++) {
-        const char *name = strarray_nth(list, i);
-        mboxlist_changesub(name, userid, sstate->authstate, 0, 0, 0, /*silent*/1);
-    }
-
     mbentry_t *mbentry = NULL;
     char *inbox = mboxname_user_mbox(userid, 0);
     mboxlist_lookup_allow_all(inbox, &mbentry, NULL);
     free(inbox);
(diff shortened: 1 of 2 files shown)


imap/sync_support.c
```diff
@@ -4654,19 +4654,31 @@ static int sync_apply_unuser(struct dlist *kin, struct sync_state *sstate)
...
-    }
-
     mbentry_t *mbentry = NULL;
     char *inbox = mboxname_user_mbox(userid, 0);
     mboxlist_lookup_allow_all(inbox, &mbentry, NULL);
     free(inbox);
 
+    /* Only destroy this user's per-user data (subscriptions, seen, ...) when
+     * this is a genuine deletion.  If the INBOX has been renamed away (a
+     * modern-dirs tombstone), the per-user files are keyed by the INBOX
```
rjbs, line 4664: What is "modern-dirs"?

rjbs (approved):

ksmurchison (approved):
Approved if CI passes

imap/sync_support.c
brong: We can rename this as non-legacy-dirs!

PR #6150, opened 2026-07-19
Cassandane: Add a separate jmap httpd for TestEntity use

wolfsage:
This way, the specific configurations required for the TestEntity system don't need to be enabled on every test suite that uses it, so it Just Works.

This also means the daemon(s) under test can have specific configurations that may behave differently if the jmap configuration changes were enabled. The only thing this might not be true for is conversations must be enabled for httpd to support jmap, so conversations dbs may come into existence when normally they wouldn't if these tests use the jmap entities at all.

### DIFF
--- a/cassandane/Cassandane/Cyrus/TestCase.pm
+++ b/cassandane/Cassandane/Cyrus/TestCase.pm
@@ -628,6 +628,14 @@ sub _create_instances
             die 'Use of config_<testname> subs is not supported anymore';
         }
 
+        # TestEntity stuff
+        my $teport = Cassandane::PortManager::alloc("localhost");
+        $conf->set(
+            tejmap_jmap_nonstandard_extensions => 'yes',
+            tejmap_httpmodules => 'jmap caldav carddav',
+            tejmap_conversations => 'yes',
+        );
+
         $instance_params{config} = $conf;
         $instance_params{install_certificates} = $want->{install_certificates};
         $instance_params{smtpdaemon} = $want->{smtpdaemon};
@@ -646,6 +654,8 @@ sub _create_instances
         $self->{instance}->add_services(@{$want->{services}});
         $self->{instance}->_setup_for_deliver()
             if ($want->{deliver});
+
+        $self->{instance}->add_service(name => 'tejmap', port => $teport, argv => ['httpd']);
 
         if ($want->{squatter}) {
             $self->{instance}->add_daemon(
--- a/cassandane/Cassandane/Instance.pm
+++ b/cassandane/Cassandane/Instance.pm
@@ -3229,10 +3229,16 @@ sub make_folder_intermediate
     $self->start();
 }
 
-sub _common_http_service_args ($self) {
-  $self->{_common_http_service_args} //= do {
-      my $service = $self->get_service("https")
-                 || $self->get_service("http");
+sub _common_http_service_args ($self, $want_service = "") {
+  $self->{"_common_http_service_args_$want_service"} //= do {
+      my $service;
+
+      if ($want_service) {
+          $service = $self->get_service($want_service);
+      } else {
+          $service = $self->get_service("https")
+                  || $self->get_service("http");
+      }
 
       $service
           || Carp::confess("can't build _common_http_service_args without an http service configured");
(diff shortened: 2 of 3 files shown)


wolfsage:
Other possible added benefits:
* logs show tejmap as the process for any requests done by the entity system which can help separate out test setup vs test calls
* any caching in the tejmap process shouldn't interfere with tests themselves

cassandane/Cassandane/Cyrus/TestCase.pm
```diff
@@ -628,6 +628,14 @@ sub _create_instances
             die 'Use of config_<testname> subs is not supported anymore';
         }
 
+        # TestEntity stuff
+        my $teport = Cassandane::PortManager::alloc("localhost");
+        $conf->set(
+            tejmap_jmap_nonstandard_extensions => 'yes',
+            tejmap_httpmodules => 'jmap caldav carddav',
+            tejmap_conversations => 'yes',
```
rjbs, line 636: "tejmap" is quite opaque.  "testentityjmap"?

cassandane/Cassandane/Instance.pm
```diff
@@ -3279,23 +3285,29 @@ sub new_jmaptester_ws_for_user ($self, $user, $new_arg = undef) {
   );
 }
 
+sub new_entity_jmaptester_for_user ($self, $user, $new_arg = undef) {
+  return $self->_new_jmaptester_for_user('Cassandane::JMAPTester', {}, $user, $new_arg, 'entity');
+}
```
rjbs, line 3290: Surprised by this.  Why do we need this instead of just a default one?

cassandane/Cassandane/TestUser.pm
```diff
@@ -80,6 +80,10 @@ JMAPTester.
 
 =cut
 
+sub new_entity_jmaptester($self, $new_arg = undef) {
```
rjbs, line 83: space between sub name and sig!

rjbs (approved):
LGTM, one musing and one style problem I think you should fix.

cassandane/Cassandane/Instance.pm
```diff
@@ -3229,10 +3229,16 @@ sub make_folder_intermediate
     $self->start();
 }
 
-sub _common_http_service_args ($self) {
-  $self->{_common_http_service_args} //= do {
-      my $service = $self->get_service("https")
-                 || $self->get_service("http");
+sub _common_http_service_args ($self, $want_service = "") {
+  $self->{"_common_http_service_args_$want_service"} //= do {
```
elliefm, line 3233: Doesn't this key now have a trailing `_` when $want_service is defaulted?  Perhaps nothing else expects to examine it so it doesn't matter...

Indentation is all wrong here, but it already was...

cassandane/Cassandane/Cyrus/TestCase.pm
```diff
@@ -647,6 +655,8 @@ sub _create_instances
         $self->{instance}->_setup_for_deliver()
             if ($want->{deliver});
 
+        $self->{instance}->add_service(name => 'tejmap', port => $teport, argv => ['httpd']);
+
```
elliefm, line 659: You should register "tejmap" (or whatever you end up calling it) in Cassandane::ServiceFactory, and then just add it to `@{$want->{services}}` when it's needed, and let the existing "start the things we'll need" mechanisms do their job.

If it's always needed, you could add it to want in `new()`, like we do for "imap".  But that doesn't smell right here (we don't do that for http, for example).

The way you've set it up, it'll never use TLS, which means anything we test using this mechanism will be tested for plaintext http only.  That's already true for most everything else anyway, but it's a bad pattern we shouldn't repeat in new code, so I think this service should use TLS rather than not.  The `JMAPCore.echo_tls` exercises a TLS connection to JMAP, so we know it's possible when it's set up.

cassandane/Cassandane/Cyrus/TestCase.pm
```diff
@@ -628,6 +628,14 @@ sub _create_instances
...
         }
 
+        # TestEntity stuff
+        my $teport = Cassandane::PortManager::alloc("localhost");
+        $conf->set(
+            tejmap_jmap_nonstandard_extensions => 'yes',
+            tejmap_httpmodules => 'jmap caldav carddav',
+            tejmap_conversations => 'yes',
+        );
+
```
elliefm, line 638: Since these config settings are added unconditionally, you should probably just set them in `Cassandane::Config::default()`.  You might need to massage some of the Info tests after doing that.

PR #6151, opened 2026-07-19
Cass TestEntity: Add an '->update' method for full updates

wolfsage:
This lets you update all fields at once rather than single fields at a time.


rjbs (approved):
Cool.

PR #6153, opened 2026-07-19
Cassandane::Tiny::Loader: Detect more bad situations

wolfsage:
This does two things:

 * Fail if a tiny-test file is found that doesn't create any test_* subroutines, as surely that's a mistake
 * Fail if a tiny test file has test_ as a prefix of its basename. If you try to run that test directly with testrunner 0 tests will run (though it would be run when you run all)


wolfsage:
@rjbs Erg, I realize now I fail two different conditions in two different ways (@bad collects all badly named test files, while test files that don't provide any tests die immediately. I'll make these two behave similar and then ask for review)

wolfsage:
@rjbs I believe this is ready

cassandane/Cassandane/Tiny/Loader.pm
```diff
@@ -27,12 +28,24 @@ sub import {
 
   chomp @tests;
 
+  my (@bad_name, @no_tests);
+
+  my $stash = Package::Stash->new($into);
```
rjbs, line 33: Is this a new prereq, or already here from some other library?  Easy to avoid it, if we don't want to add it.

rjbs (approved):

cassandane/Cassandane/Tiny/Loader.pm
wolfsage: Cyrus needs App::Cmd (who knew) which uses Class::Load which uses Package::Stash so it's already there

PR #6159, opened 2026-07-20
imapd: honor FUZZY for HEADER searches of indexed fields

### LINKED ISSUE #6157: imapd: HEADER search key ignores FUZZY, forcing a full scan for indexed fields
## Summary

With FUZZY active (RFC 6203, or `search_fuzzy_always`), the dedicated search keys SUBJECT, FROM, TO, CC and BCC are served from the Xapian index. The generic `HEADER <field-name>` key is not: it always builds a substring-match node, even when `<field-name>` is one of those same indexed fields. `HEADER SUBJECT x` therefore scans the whole mailbox, while `SUBJECT x` — which RFC 3501 §6.4.4 defines against the same header content — uses the index. Clients that phrase these as HEADER searches (e.g. Roundcube) get no index benefit.

## Observed (3.12.2, 588k-message mailbox, engine-level timing)

Identical 377-message result set:

```
fuzzy or SUBJECT x BODY y            316 ms   (index)
fuzzy or HEADER SUBJECT x BODY y    7263 ms   (full scan)
```

One un-indexed OR arm forces the whole query to scan, so a Roundcube search combining HEADER Subject/From/To with BODY scans the mailbox (telemetry: `INPROGRESS 412963/588752`, `OK Completed (26 msgs in 14.097 secs)`).

## Mechanism

`string_match()` (keyed SUBJECT/FROM/TO/...) builds a `SEOP_FUZZYMATCH` node when `fuzzy_depth > 0` and the attribute is fuzzable. The `HEADER` key path (imapparse.c) hardcodes `SEOP_MATCH` regardless of `fuzzy_depth`, even though `search_attr_find_field()` reduces subject/from/to/cc/bcc to the same fuzzable attributes.

## Fix (PR to follow)

In the HEADER path, build `SEOP_FUZZYMATCH` when `fuzzy_depth > 0` and the reduced attribute is fuzzable, mirroring `string_match()`. Non-fuzzable and arbitrary header fields keep the existing `SEOP_MATCH` behavior.

paulnivin:
Fixes #6157.

The generic `HEADER <field-name>` search key always built a SEOP_MATCH (substring) node, even under FUZZY and even when `<field-name>` reduces to an indexed, fuzzable attribute. The dedicated keys (SUBJECT, FROM, TO, CC, BCC) build a SEOP_FUZZYMATCH node via string_match() and are served from the search engine, so `HEADER SUBJECT x` fell back to a full-mailbox scan while `SUBJECT x` — the same header content per RFC 3501 §6.4.4 — used the index. Clients such as Roundcube phrase Subject/From/To searches as HEADER searches, so they could not use the index: on a 588k-message mailbox, an `OR HEADER SUBJECT / HEADER FROM / HEADER TO / BODY` search scanned for ~14 seconds where the keyed equivalent was index-served in ~0.3.

Factor the match-node construction out of string_match() into attr_match() and build the HEADER node with it too, so a field that reduces to a dedicated attribute matches exactly like its dedicated key. Arbitrary header fields keep substring semantics in every configuration: their generated attributes no longer claim SEA_FUZZABLE, a previously unconsumed flag that should not have been set; search_attr_find_field reduces indexed fields to their dedicated, fuzzable attributes.

Two tests: header_fuzzy asserts FUZZY HEADER Subject matches a stemmed variant the same way the dedicated SUBJECT key does — a stem match cannot be served by the substring fallback — and that a header field with no search attribute keeps substring semantics both ways (a mid-word substring matches, a stemmed variant does not). header_fuzzy_xh repeats the assertions under search_index_headers, additionally pinning that a HEADER search is never served from the field-less indexed header text. Both fail on current master and pass with the change; the full Cyrus::SearchFuzzy suite passes on current master with this branch.

### DIFF
--- a/changes/next/header-search-honor-fuzzy
+++ b/changes/next/header-search-honor-fuzzy
@@ -1,0 +1,27 @@
+Description:
+
+Under FUZZY search, a HEADER search of a field with a designated search
+part (Subject, From, To, Cc, Bcc) is now evaluated like the dedicated
+search key: it is served from the search engine and gains stemming, and
+no longer matches substrings or the empty string. HEADER searches of
+other fields are unchanged.
+
+
+Documentation:
+
+None
+
+
+Config changes:
+
+None
+
+
+Upgrade instructions:
+
+None
+
+
+GitHub issue:
+
+#6157
--- a/imap/imapparse.c
+++ b/imap/imapparse.c
@@ -1142,9 +1142,20 @@ static int get_search_criterion(struct protstream *pin,
             c = getastring(pin, pout, &arg2);
             if (c <= EOF) goto missingarg;
 
-            e = search_expr_new(parent, SEOP_MATCH);
-            e->attr = search_attr_find_field(arg.s);
-            e->value.s = charset_convert(arg2.s, base->charset, charset_flags|CHARSET_KEEPCASE);
+            const search_attr_t *attr = search_attr_find_field(arg.s);
+            if (base->fuzzy_depth > 0 && search_attr_is_fuzzable(attr)) {
+                /* RFC 6203: a fuzzable header field (Subject, From, To, ...)
+                 * uses the same index-backed match as its dedicated search
+                 * key rather than a full-text scan. */
+                e = search_expr_new(parent, SEOP_FUZZYMATCH);
+                e->attr = attr;
+                e->value.s = xstrdup(arg2.s);
+            }
+            else {
+                e = search_expr_new(parent, SEOP_MATCH);
+                e->attr = attr;
+                e->value.s = charset_convert(arg2.s, base->charset, charset_flags|CHARSET_KEEPCASE);
+            }
             if (!e->value.s) {
                 e->op = SEOP_FALSE;
                 e->attr = NULL;
--- a/imap/imapparse.c
+++ b/imap/imapparse.c
@@ -796,18 +796,18 @@ bad:
 }
 
 
-static void string_match(search_expr_t *parent, const char *val,
-                         const char *aname, struct searchargs *base)
+static void attr_match(search_expr_t *parent, const char *val,
+                       const search_attr_t *attr, struct searchargs *base)
 {
     search_expr_t *e;
-    const search_attr_t *attr = search_attr_find(aname);
     enum search_op op = SEOP_MATCH;
     char *searchval;
 
     if (base->fuzzy_depth > 0 &&
         search_attr_is_fuzzable(attr)) {
         op = SEOP_FUZZYMATCH;
-        searchval = xstrdup(val); // keep search value as-is
+        // transcode to UTF-8, but leave normalization to the search engine
+        searchval = charset_convert(val, base->charset, CHARSET_KEEPCASE);
     }
     else searchval = charset_convert(val, base->charset, charset_flags|CHARSET_KEEPCASE);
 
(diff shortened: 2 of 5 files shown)


paulnivin:
@rsto This fix makes the HEADER search key honor FUZZY the way SUBJECT, FROM, TO, CC and BCC already do, sharing string_match()'s node construction. Do you have time to review? This is the last needed change from my set of PRs enabling Roundcube to efficiently search large mailboxes with Cyrus. Thanks.

imap/imapparse.c
rsto: You dropped the charset conversion in `bytestring_match`, so this could now dump non-UTF8 charsets into the query. Convert it to UTF-8 without converting it to search form as

```
          // transcode to UTF-8, but leave normalization to the search engine
          searchval = charset_convert(val, base->charset, CHARSET_KEEPCASE);
```

imap/search_expr.c
```diff
@@ -3353,7 +3353,7 @@ EXPORTED const search_attr_t *search_attr_find_field(const char *field)
     char *key = NULL;
     static const search_attr_t proto = {
         "name",
-        SEA_FUZZABLE,
+        /*flags*/0, /* not SEA_FUZZABLE: fuzzable fields reduce to their own attrs */
```
rsto: The "not SEA_FUZZABLE" comment looks to only be relevant in context of the diff but afterwards serve no purpose in the code? If you agree, please remove it.

changes/next/header-search-honor-fuzzy
```diff
@@ -0,0 +1,25 @@
+Description:
+
+Under FUZZY search, a HEADER search of a field with a dedicated search
+attribute (Subject, From, To, Cc, Bcc) is served from the search engine
+like the dedicated key, instead of scanning the whole mailbox.
```
rsto: This doesn't describe the full effect of this change. Please also succinctly describes what changed for HEADER search as I outlined in my test commit message.

rsto (changes requested):
I think this is a good idea. It took me a while to grok the impact of this change and I think both the test code and the commit message could make this more clear. I rewrote the tests in https://github.com/rsto/cyrus-imapd/commit/dd0c30b40cf269f6d5e5c63b4713439d49d5a363 to a single test function and its commit message now clearly describes the impact. I would prefer you to use this test and commit message, at least the table that outlines the before/after. 

I also would prefer the test and the logic change to go into the same commit, because that will make "git blame" more ergonomic. If you do that, then there's also no need to refer to the pull request in the commit message. But even within the Cyrus team there a two schools of thoughts if to separate tests from logic changes, so if you prefer to keep them in two commits, I'm fine with that, too.

In any case, the dropped charset conversion must be addressed.

Lastly, while I do want to get this merged and make Roundcube better interoperate with Cyrus, I first have to check if the changed HEADER search would impact any widely used IMAP client. If so, we'll need a way to make this work for them, too.

imap/imapparse.c
paulnivin: Done. Applied in `attr_match`. Thanks for catching this.

imap/search_expr.c
paulnivin: Agreed and removed.

changes/next/header-search-honor-fuzzy
paulnivin: Done.

paulnivin:
Thanks for the review and the rewritten test -- adopted as-is with your table and squashed with the logic change.

Understood on possible client compatibility concerns. If there are clients relying on `HEADER <field> ""` as an exists-check under FUZZY, we could force non-FUZZY semantics as a special case. Let me know if you'd like me to add that.

rsto (approved):
Thanks!

PR #6160, opened 2026-07-20
search: resume a batched update where its last batch stopped

### LINKED ISSUE #6156: squatter: non-incremental Xapian indexing loops forever on large mailboxes
With the Xapian engine in the default conversations-indexed mode, running squatter's index mode without `-i` (e.g. plain `squatter`, `squatter -u user` — the manpage's documented default) can never complete a mailbox with more messages than `search_batchsize`: every batch pass selects the same first batch and the process loops forever with no error. Mailboxes that fit in one batch do complete — but every run re-adds all their messages as duplicate Xapian documents.

## Affected versions

`index_one()` has set SEARCH_UPDATE_BATCH unconditionally since 3.4.0 (earlier releases batched only in rolling mode).

## Mechanism

1. `begin_mailbox_update()` loads the already-indexed UID ranges (`read_indexed()`) only when `SEARCH_UPDATE_INCREMENTAL` or `SEARCH_UPDATE_AUDIT` is set; without `-i`, `tr->oldindexed` stays empty.
2. For messages in the mailbox currently being indexed, `is_indexed()` / `is_indexed_cb()` consult only the in-memory seqsets (`tr->indexed`, `tr->oldindexed`) — in CONVINDEXED mode there is no per-message check against the Xapian db itself (that exists only in XAPINDEXED mode, `-Z`).
3. So without `-i`, every message reads as unindexed; `build_batch()` stops at `search_batchsize`, `search_update_mailbox()` returns IMAP_AGAIN, and `index_one()` does `if (blocking && r == IMAP_AGAIN) goto again;` — selecting the identical batch again. Infinite loop, no error logged.
4. For mailboxes that fit in one batch the run terminates, but since `xapian_wrap.cpp` only ever calls `add_document()` (never `replace_document()`), each non-incremental run appends a full duplicate set of documents for every message.

Observed log signature (3.12.2, `search_batchsize: 8192`, 588k-message mailbox — identical cycles repeating every ~2 minutes indefinitely):

```
squatter[5601]: search_update_mailbox batching 8192 messages to user.foobar
squatter[5601]: Xapian committed 16142 updates in 10.186381 sec
squatter[5601]: write_indexed db=... uids=1:11057
squatter[5601]: mailbox: longlock user.foobar for 105.3 seconds
squatter[5601]: indexing mailbox user/foobar...
squatter[5601]: search_update_mailbox batching 8192 messages to user.foobar
```

Note this issue is invisible in many deployments: `search_engine` defaults to `none`; Xapian deployments predominantly run rolling mode (`-R`), which already forces `incremental_mode = 1`; and test mailboxes rarely exceed the default `search_batchsize`.

## Proposed fix (PR to follow)

Imply `-i` for the Xapian engine in the indexer and synclog modes unless `-Z` is given, mirroring what rolling mode already does, and log a NOTICE:

```c
    if ((mode == INDEXER || mode == SYNCLOG) &&
        !incremental_mode && !xapindexed_mode &&
        config_getenum(IMAPOPT_SEARCH_ENGINE) == IMAP_ENUM_SEARCH_ENGINE_XAPIAN) {
        syslog(LOG_NOTICE, "xapian: forcing incremental mode (-i implied)");
        incremental_mode = 1;
    }
```

paulnivin:
Fixes #6156.

`search_update_mailbox()` indexes at most `search_batchsize` messages per call and returns `IMAP_AGAIN`; squatter repeats the call until the mailbox is complete. Every repeat selected messages from the start of the mailbox again. An incremental update was unaffected, because it starts at the first UID its indexed ranges do not cover and those ranges advance with every batch. A non-incremental update ignores the ranges by design: with Xapian it re-selected the same first batch on every repeat, so bare `squatter` never completed a mailbox larger than `search_batchsize`, and re-added that batch's documents on every pass until it was killed.

This carries a resume UID across the repeats — the UID of the message that tripped the batch limit — so each repeat continues where the last one stopped. It replaces the `nindexedptr` parameter added in #6045: squatter now tells an update that made progress from one that did not by whether the resume UID advanced.

Reworked after #6045 landed. The earlier version of this PR made bare `squatter` imply `-i` for Xapian. #6045 added `squatter_createdmodseq_reindex`, which runs bare `squatter` on an already-indexed mailbox specifically so that its documents get re-added and the index generation bumps — implying `-i` breaks that test. So this version leaves the meaning of a bare run alone and only fixes the batching, which resolves #6156.

The test appends more messages than three batches hold and asserts a bare run completes, that every message is searchable, that one run indexes each message once, that `squatter -A` reports nothing, and that a following `-i` run adds nothing. It fails on current master (the run loops) and passes with the fix. The full `Cyrus::SearchFuzzy` suite passes on current master with this branch.

### DIFF
--- a/cassandane/tiny-tests/SearchFuzzy/squatter_index_default
+++ b/cassandane/tiny-tests/SearchFuzzy/squatter_index_default
@@ -1,0 +1,63 @@
+#!perl
+use Cassandane::Tiny;
+use POSIX qw(WNOHANG);
+
+sub test_squatter_index_default
+    :SearchBatchsize20
+    ($self)
+{
+    my $talk = $self->{store}->get_client();
+    my $basedir = $self->{instance}->{basedir};
+    my $outfile = "$basedir/audit.tmp";
+
+    xlog $self, "append enough messages for three full batches and a partial one";
+    my $nmsgs = 65;
+    foreach my $i (1..$nmsgs) {
+        $self->make_message("filler $i") || die;
+    }
+
+    xlog $self, "a bare squatter run must complete: when broken it re-selects the same batch forever";
+    my $pid = $self->{instance}->run_command(
+        {cyrus => 1, background => 1}, 'squatter');
+    my $done = 0;
+    my $batchlines = 0;
+    eval {
+        timed_wait(sub {
+            if (waitpid($pid, WNOHANG) == $pid) {
+                $done = 1;
+                return 1;
+            }
+            # a correct run logs one batching line per full batch, so a
+            # fourth one means it is looping
+            $batchlines += scalar $self->{instance}->getsyslog(qr/batching/);
+            return $batchlines > 3;
+        }, description => "squatter to finish", maxwait => 60);
+    };
+    # kill a looping squatter rather than letting it outlive the test
+    eval { $self->{instance}->stop_command($pid) } if not $done;
+    $self->assert($done);
+
+    xlog $self, "every message must be searchable";
+    $talk->select('INBOX');
+    my $uids = $talk->search('fuzzy', 'subject', 'filler');
+    $self->assert_num_equals($nmsgs, scalar @$uids);
+
+    xlog $self, "one run indexes each message once: no batch may re-add an earlier one";
+    my $xapdirs = ($self->{instance}->run_mbpath(-u => 'cassandane'))->{xapian};
+    my ($gdocs, $parts) = $self->delve_docs($xapdirs->{t1} . "/xapian");
+    $self->assert_num_equals($nmsgs, scalar @$gdocs);
+
+    xlog $self, "audit reports no unindexed messages";
+    $self->{instance}->run_command(
+        { cyrus => 1, redirects => { stdout => $outfile } },
+        'squatter', '-A');
+    open my $fh, '<', $outfile or die "open $outfile: $!";
+    my @audits = readline($fh);
+    close $fh;
+    $self->assert_num_equals(0, scalar @audits);
+
+    xlog $self, "an incremental run must find nothing left to index";
+    $self->{instance}->run_command({cyrus => 1}, 'squatter', '-i');
+    ($gdocs, $parts) = $self->delve_docs($xapdirs->{t1} . "/xapian");
+    $self->assert_num_equals($nmsgs, scalar @$gdocs);
+}
--- a/cassandane/tiny-tests/SearchFuzzy/squatter_index_default
+++ b/cassandane/tiny-tests/SearchFuzzy/squatter_index_default
@@ -1,0 +1,57 @@
+#!perl
+use Cassandane::Tiny;
+use POSIX qw(WNOHANG);
+
+sub test_squatter_index_default
+    :SearchBatchsize20
+    ($self)
+{
+    my $talk = $self->{store}->get_client();
+    my $basedir = $self->{instance}->{basedir};
+    my $outfile = "$basedir/audit.tmp";
+
+    xlog "append more messages than search_batchsize (default 20)";
+    foreach my $i (1..25) {
+        $self->make_message("filler $i") || die;
+    }
+
+    xlog "a bare squatter run must complete: when broken it re-selects";
+    xlog "the same batch forever, logging the batching line each pass";
+    my $pid = $self->{instance}->run_command(
+        {cyrus => 1, background => 1}, 'squatter');
+    my $done = 0;
+    my $batchlines = 0;
+    foreach (1..60) {
+        if (waitpid($pid, WNOHANG) == $pid) {
+            $done = 1;
+            last;
+        }
+        # a correct run logs "batching" once; twice means the loop
+        $batchlines += scalar $self->{instance}->getsyslog(qr/batching/);
+        last if $batchlines >= 2;
+        sleep 1;
+    }
+    # kill a looping squatter rather than letting it outlive the test
+    eval { $self->{instance}->stop_command($pid) } if not $done;
+    $self->assert($done);
+
+    $talk->select('INBOX');
+    my $uids = $talk->search('fuzzy', 'subject', 'filler');
+    $self->assert_num_equals(25, scalar @$uids);
+
+    xlog "a second bare run must not add duplicate documents";
+    $self->{instance}->run_command({cyrus => 1}, 'squatter');
+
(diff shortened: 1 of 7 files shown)


rsto (commented):
This looks like a genuine bug independent of https://github.com/cyrusimap/cyrus-imapd/pull/6045 
But again, I don't want to touch the Xapian code until https://github.com/cyrusimap/cyrus-imapd/pull/6045 lands

paulnivin:
@rsto Reworked after #6045: your `squatter_createdmodseq_reindex` relies on a bare run re-adding documents, which the old "imply -i" approach prevented. Now the batching layer just resumes where the last batch stopped, so squatter semantics are untouched and your test passes. This PR now just fixes the batching loop.

cassandane/tiny-tests/SearchFuzzy/squatter_index_default
```diff
@@ -0,0 +1,56 @@
...
+    my $talk = $self->{store}->get_client();
+    my $basedir = $self->{instance}->{basedir};
+    my $outfile = "$basedir/audit.tmp";
+
+    xlog "append more messages than search_batchsize (default 20)";
+    foreach my $i (1..25) {
+        $self->make_message("filler $i") || die;
+    }
+
+    xlog "a bare squatter run must complete: when broken it re-selects";
```
rsto: This `xlog` and the other ones in this test file are missing `$self` -> this should become `xlog $self,`. Also: please do not split log lines.

cassandane/tiny-tests/SearchFuzzy/squatter_index_default
```diff
@@ -0,0 +1,56 @@
...
+    xlog "the same batch forever, logging the batching line each pass";
+    my $pid = $self->{instance}->run_command(
+        {cyrus => 1, background => 1}, 'squatter');
+    my $done = 0;
+    my $batchlines = 0;
+    foreach (1..60) {
```
rsto: Use the `timed_wait` function that we have in cassandane, rather than handrolling your own wait loop. But keep killing the looping squatter.

cassandane/tiny-tests/SearchFuzzy/squatter_index_default
```diff
@@ -0,0 +1,56 @@
...
+    }
+    # kill a looping squatter rather than letting it outlive the test
+    eval { $self->{instance}->stop_command($pid) } if not $done;
+    $self->assert($done);
+
+    $talk->select('INBOX');
+    my $uids = $talk->search('fuzzy', 'subject', 'filler');
+    $self->assert_num_equals(25, scalar @$uids);
+
+    xlog "one run indexes each message once: the second batch must not";
```
rsto: please do not split log lines

cassandane/tiny-tests/SearchFuzzy/squatter_index_default
```diff
@@ -0,0 +1,56 @@
...
+    xlog "have re-added the first";
+    my $xapdirs = ($self->{instance}->run_mbpath(-u => 'cassandane'))->{xapian};
+    my ($gdocs, $parts) = $self->delve_docs($xapdirs->{t1} . "/xapian");
+    $self->assert_num_equals(25, scalar @$gdocs);
+
+    xlog "audit reports no unindexed messages";
```
rsto: In addition to audit, please also assert at the end of this test that an incremental `-i` squatter run doesn't add more messages. Use 3 or more batch runs, currently the 25 messages only test one full batch.

imap/search_engines.h
```diff
@@ -229,10 +229,19 @@ search_text_receiver_t *search_begin_update(int verbose);
...
+ * uid of at least *resumeuidptr (0 means all messages), and on IMAP_AGAIN
+ * it is set to the uid a repeated update should resume from. A caller that
+ * repeats a batched update must pass the same variable to every repeat,
+ * initialized to 0: without it, an update that does not trust the indexed
+ * uid ranges (a non-incremental one) reselects the same first batch on
+ * every repeat. */
 int search_update_mailbox(search_text_receiver_t *rx,
                           struct mailbox **mailboxptr,
                           int min_indexlevel, int flags,
+                          uint32_t *resumeuidptr,
```
rsto: Let's make `resumeuidptr` a mandatory property, e.g. it can't be NULL. Please also rephrase the comment above accordingly.

imap/search_engines.h
```diff
@@ -229,10 +229,19 @@ search_text_receiver_t *search_begin_update(int verbose);
...
                           size_t *nindexedptr);
```
rsto: Using `resumeuidptr` looks to be a good replacement for `nindexedptr`. Rather than checking for unproductive runs in squatter.c `index_one()` using `nindexed`, we could now check that `resumeuid` advanced. If you want to draft this in your PR feel free to do so, if you prefer to not do that, I'll take it after we merged your PR.

rsto (changes requested):
Thanks for highlighting this and creating this PR. The test code deserves some updates and I'd prefer to to make `resumeuid` a mandatory argument. That might also allow to remove `nindexed`. In addition to the specific comments, please also add a file in the "changes/next" directory that succinctly describes the change, one or two sentences is enough.

cassandane/tiny-tests/SearchFuzzy/squatter_index_default
paulnivin: Done

cassandane/tiny-tests/SearchFuzzy/squatter_index_default
paulnivin: Done

cassandane/tiny-tests/SearchFuzzy/squatter_index_default
paulnivin: Done

cassandane/tiny-tests/SearchFuzzy/squatter_index_default
paulnivin: Done. 65 messages (three full batches plus a partial); a trailing `-i` run must add no documents.

imap/search_engines.h
paulnivin: Done

imap/search_engines.h
paulnivin: Done.`nindexed` removed and `index_one` now treats "resume uid advanced" as progress.

rsto (approved):
Thanks for your contribution and patience.

PR #6161, opened 2026-07-20
search_xapian: confirm CONVINDEXED dedup against Xapian

### LINKED ISSUE #6158: squatter: CONVINDEXED indexing silently leaves messages unindexed
## Summary

With the Xapian engine in the default conversations-indexed mode (CONVINDEXED, i.e. without `-Z`), an incremental `squatter -i` build can silently leave live messages out of the index. `squatter -A` reports them as unindexed; a full-scan `squatter -Z` (XAPINDEXED) is currently the only way to recover them. The affected messages are not returned by index-served (fuzzy) searches until then.

The root cause is that CONVINDEXED decides "is this message already indexed?" from the recorded indexed UID ranges, treating range membership as a proxy for "this GUID's document exists in Xapian". That proxy is unreliable: a range can contain UIDs whose documents are not in Xapian, and when such a UID backs the dedup decision, a live message is skipped.

## Affected versions

The range-as-proxy dedup (`is_indexed()` / `is_indexed_part()` → `is_indexed_cb()`) and the indexed-range seed in `end_message_update()` are long-standing and present on `master`. Only CONVINDEXED is affected; XAPINDEXED (`-Z`) checks Xapian per message and does not have the problem.

## Mechanism

A range can claim a UID whose document is not in Xapian for several reasons; two combine to produce the observed gaps.

1. **Unreliable dedup proxy.** In CONVINDEXED mode `is_indexed()` runs `conversations_guid_foreach(... is_indexed_cb ...)`, which returns "indexed" when any occurrence of the GUID has a UID in that folder's recorded range (`seqset_ismember`). It never confirms that the GUID's document is actually present in Xapian. XAPINDEXED instead calls `xapian_dbw_is_indexed()`.

2. **Boundary-deferred live message marked indexed.** When a mailbox exceeds `search_batchsize`, `build_batch()` stops at the limit, having stepped to but not processed the next (live) message. `search_update_mailbox()` then reopens the mailbox for the attachment pass; `begin_mailbox_update()` re-reads the just-written range, and the seed in `end_message_update()` adds `seqset_firstnonmember(oldindexed)` — which is exactly that deferred live UID — to the indexed range. `first_unindexed_uid()` now steps past it, so it is never iterated again.

The result is a range that claims live UIDs with no Xapian document. Any live message that duplicates such a UID's GUID (cross-folder duplicates are common) is then skipped by (1); the boundary message from (2) is skipped outright. Both are absent from the index with no error logged.

### How to observe

On a mailbox larger than `search_batchsize` with cross-folder duplicates, after a normal build:

```
squatter -i -u foobar
squatter -A -u foobar
  Unindexed message(s) in user.foobar.Sent: 8294
  Unindexed message(s) in user.foobar.Work: 593 594 595 ...
```

`mbexamine` confirms the reported UIDs are live, current messages. A subsequent `squatter -Z -u foobar` indexes them and `-A` then reports clean — confirming the messages are fine and only the CONVINDEXED dedup missed them.

Note this is easy to miss: the default `search_batchsize` is 20, but larger deployments raise it, so boundaries are infrequent; the gaps are silent; and `-Z` masks them on the next repair pass. `squatter -A` also only verifies messages claimed by the recorded ranges, so a skipped message that was never claimed by a range evades the audit entirely (a separate PR extends the audit to cover this).

## Proposed fix (PR to follow)

1. In `is_indexed()` and `is_indexed_part()`, confirm a positive proxy result against Xapian (`xapian_dbw_is_indexed()`) before treating the message or part as indexed. A negative needs no such verification and is indexed as before; the db is opened lazily, so a no-op incremental run doesn't touch Xapian.

2. Guard the `end_message_update()` seed to claim only UIDs below the message actually being indexed, so it no longer claims a boundary-deferred live UID.

With (1) the range over-claim is harmless for dedup, and with (2) the boundary message is iterated; together a full CONVINDEXED build is gap-free without `-Z`.

paulnivin:
Fixes #6158.

In CONVINDEXED mode `is_indexed()` / `is_indexed_part()` decide whether a message is already indexed from the recorded indexed UID ranges, via `is_indexed_cb()` — range membership is used as a proxy for "this GUID's document exists in Xapian". The proxy is not reliable: a range can claim UIDs whose documents are not in Xapian (gap-filled ranges, an interrupted run, an expunged duplicate). When such a UID backs the dedup decision, a live message is skipped and never indexed. `squatter -A` reports it; only `-Z` (XAPINDEXED, which checks Xapian per message) recovers it.

This confirms a positive proxy result (reported as already indexed) against Xapian, which is authoritative, before skipping the message or part. A negative (reported as not indexed) is indexed as before. The db is opened lazily, with the other active tiers read-only (same as XAPINDEXED) so the check sees every tier; a no-op incremental run doesn't touch Xapian.

Updated after #6045: that PR fixes the batch-boundary case this PR originally also guarded against, so the seed-guard commit and its test are dropped. What remains is the range-without-document case, which #6045 does not address: `convindexed_expunged_duplicate` constructs such a range directly in `cyrus.indexed.db` and asserts an identical message delivered to another folder is still indexed, searchable, and not reported by `squatter -A`. It fails on current master and passes with this change. The full `Cyrus::SearchFuzzy` suite passes on current master with this branch.

### DIFF
--- a/cassandane/tiny-tests/SearchFuzzy/convindexed_expunged_duplicate
+++ b/cassandane/tiny-tests/SearchFuzzy/convindexed_expunged_duplicate
@@ -1,0 +1,84 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_convindexed_expunged_duplicate
+    ($self)
+{
+    # Like audit_unindexed, this test edits cyrus.indexed.db directly to
+    # construct a recorded range that claims a UID with no Xapian document,
+    # the state left behind by an indexing interruption or by a message
+    # expunged and expired before it was ever indexed.
+
+    my $talk = $self->{store}->get_client();
+    my $basedir = $self->{instance}->{basedir};
+    my $outfile = "$basedir/audit.tmp";
+
+    my $raw = ""
+        . "From: from\@local\r\n"
+        . "To: to\@local\r\n"
+        . "Subject: sharedguid\r\n"
+        . "Date: Mon, 13 Apr 2020 15:34:03 +0200\r\n"
+        . "MIME-Version: 1.0\r\n"
+        . "Content-Type: text/plain\r\n"
+        . "\r\n"
+        . "uniquebodyterm\r\n";
+
+    xlog "create and index message UID 1";
+    $self->make_message("filler one") || die;
+    $self->{instance}->run_command({cyrus => 1}, 'squatter', '-i');
+
+    xlog "append message UID 2 but do not index it";
+    $talk->append('INBOX', { Literal => $raw }) || die "append: $@";
+
+    my $data = $self->{instance}->run_mbpath(-u => 'cassandane');
+    my $xapdir = $data->{xapian}{t1};
+    my $format = $self->{instance}->{config}->get('search_indexed_db');
+
+    xlog "read current cyrus.indexed.db";
+    my ($key, $val);
+    my $result = $self->{instance}->run_dbcommand_cb(sub {
+      my ($k, $v) = @_;
+      return if $k =~ m/\*V\*/;
+      return if $k =~ m/\*HIGHEST_CREATEDMODSEQ\*/;
+      $self->assert_null($key);
+      ($key, $val) = ($k, $v);
+    }, "$xapdir/xapian/cyrus.indexed.db", $format, ['SHOW']);
+    $self->assert_str_equals('ok', $result);
+    $self->assert_not_null($key);
+    $self->assert_not_null($val);
+
+    xlog "add UID 2 to the recorded range without a document, as an";
+    xlog "interrupted indexing run or an expunged-then-expired message";
+    xlog "leaves it";
+    $self->{instance}->run_dbcommand(
+        "$xapdir/xapian/cyrus.indexed.db", $format,
+        ['SET', $key, $val . ':2']);
+
+    xlog "append an identical copy of the unindexed message to another";
+    xlog "folder and index it";
+    $talk->create('INBOX.other') || die "create: $@";
+    $talk->append('INBOX.other', { Literal => $raw }) || die "append: $@";
+    $self->{instance}->run_command({cyrus => 1}, 'squatter', '-i');
+
+    xlog "append and index another message there, so that folder's range";
+    xlog "claims UID 1";
+    my $raw2 = $raw;
+    $raw2 =~ s/sharedguid/othersubject/;
+    $raw2 =~ s/uniquebodyterm/otherbodyterm/;
+    $talk->append('INBOX.other', { Literal => $raw2 }) || die "append: $@";
+    $self->{instance}->run_command({cyrus => 1}, 'squatter', '-i');
+
+    xlog "the copy must be searchable";
+    $talk->select('INBOX.other');
+    my $uids = $talk->search('fuzzy', 'body', 'uniquebodyterm');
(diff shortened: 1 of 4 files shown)


rsto (commented):
There's a different PR that addresses this issue: https://github.com/cyrusimap/cyrus-imapd/pull/6045

I suggest we wait for that other PR to land and then decide what about this one.

paulnivin:
Happy to continue work once #6045 lands, but I think the two bugs are orthogonal -- this bug doesn't involve a race condition.

paulnivin:
@rsto #6045 fixes the batch-boundary case. This PR now covers the remaining portion to fix #6158 -- a range can claim a UID with no Xapian document, and the dedup then skips a live duplicate forever. The added test fails on current master and passes with this change.

cassandane/tiny-tests/SearchFuzzy/convindexed_expunged_duplicate
```diff
@@ -0,0 +1,84 @@
...
+      return if $k =~ m/\*V\*/;
+      return if $k =~ m/\*HIGHEST_CREATEDMODSEQ\*/;
+      $self->assert_null($key);
+      ($key, $val) = ($k, $v);
+    }, "$xapdir/xapian/cyrus.indexed.db", $format, ['SHOW']);
+    $self->assert_str_equals('ok', $result);
+    $self->assert_not_null($key);
+    $self->assert_not_null($val);
+
+    xlog "add UID 2 to the recorded range without a document, as an";
```
rsto, line 50: The test here manually corrupts `cyrus.indexed.db`, so it hardly demonstrates a bug that's reachable by normal mode of operations.

rsto (commented):
As it currently stands, this PR does not seem to demonstrate a real bug. Unless it can demonstrate the issue without manually editing `cyrus.indexed.db` into a corrupt state, I'm inclined to reject this PR.

Also see https://github.com/cyrusimap/cyrus-imapd/blob/81886f0be74bbf7c7f438cbac406a20adf91604a/imap/search_engines.c#L473-L478 where we take care to index expunged message to prevent what this PR seems to suggest needs fixing.

cassandane/tiny-tests/SearchFuzzy/convindexed_expunged_duplicate
paulnivin: Fine with closing this PR as it should not be possible to hit this bug now that #6045 has landed; however, if installations are in a broken state (mine was), without this PR, admins need to run `squatter -Z` after they're on a version that contains #6045.

rsto: if installations are in a broken state (mine was), without this PR, admins need to run squatter -Z after they're on a version that contains https://github.com/cyrusimap/cyrus-imapd/pull/6045.

Right. I note that in the changes file for https://github.com/cyrusimap/cyrus-imapd/pull/6045 I wrote:

> Installations that once fully built their index and then updated it with rolling squatters do not need to take any action. Other installations may consider rebuilding their search index if they want to avoid inconsistent Email/queryChanges results.

I know that "fully built their index" is doing a lot of work, but the changes file (and hence the upcoming release notes if this gets backported) does nudge people to reindex.

That being said, this PR takes the wrong assumptions anyway, so I wouldn't even use it just for a backport.

PR #6162, opened 2026-07-20
squatter: audit messages outside the recorded ranges

paulnivin:
Audit mode only verified messages claimed by the recorded indexed ranges and ignored the rest, presuming an unclaimed message is backlog that the next incremental run will index. That presumption trusts the machinery being audited: a message repeatedly skipped by the CONVINDEXED dedup (see #6158) is unclaimed and never indexed, so the audit missed exactly the messages most in need of reporting. squatter.8 also already promises that `-A` "reports any unindexed messages".

Check every live message against Xapian. Expunged messages are now skipped: their documents may legitimately have been expired or compacted away, and reporting them would be noise.

On a production 594k-message mailbox whose index predated the CONVINDEXED dedup fix, the extended audit found 70 unindexed live messages in under ten seconds -- spaced one per search_batchsize (8192) interval, the batch boundary loss pattern of #6158 -- in an index the unextended audit had repeatedly reported clean. A single squatter -Z pass then repaired them and the audit reported clean.

The audit_unrecorded test asserts a delivered-but-never-indexed message is reported. It fails against unpatched cyrus and passes with the change (validated on a 3.12.2 backport; the master port of the test is syntax-converted from the validated version). The full SearchFuzzy suite, including the existing audit_unindexed and index_tier_audit_bug tests, passes with the change applied.

### DIFF
--- a/cassandane/tiny-tests/SearchFuzzy/audit_unrecorded
+++ b/cassandane/tiny-tests/SearchFuzzy/audit_unrecorded
@@ -1,0 +1,39 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_audit_unrecorded
+    ($self)
+{
+    my $talk = $self->{store}->get_client();
+    my $basedir = $self->{instance}->{basedir};
+    my $outfile = "$basedir/audit.tmp";
+
+    xlog "create and index two messages";
+    $self->make_message("needle one") || die;
+    $self->make_message("needle two") || die;
+    $self->{instance}->run_command({cyrus => 1}, 'squatter', '-i');
+
+    xlog "audit reports nothing";
+    $self->{instance}->run_command(
+        { cyrus => 1, redirects => { stdout => $outfile } },
+        'squatter', '-A');
+    open my $fh, '<', $outfile or die "open $outfile: $!";
+    my @audits = readline($fh);
+    close $fh;
+    $self->assert_num_equals(0, scalar @audits);
+
+    xlog "append a third message without indexing it";
+    $self->make_message("needle three") || die;
+
+    xlog "audit reports the unindexed message even though no recorded";
+    xlog "range claims it";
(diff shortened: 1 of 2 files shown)


rsto (commented):
I _think_ this also is related to https://github.com/cyrusimap/cyrus-imapd/pull/6045 ? I suggest to wait for the outcome of https://github.com/cyrusimap/cyrus-imapd/pull/6045 before assessing what do about this one.

paulnivin:
Same as #6161, happy to wait for #6045 to land, but I think this bug is also orthogonal -- `squatter -A` never inspects messages outside recorded ranges.

paulnivin:
Rebased on current master. #6045 doesn't change what `-A` inspects -- test fails on master and passes with this change.

rsto (commented):
This whole PR seems to be based on a misunderstanding of the `-A` argument of squatter. Admittedly the man page is poorly worded. What it means when it states it reports "any unindexed message" is that it reports any message that cyrus.indexed.db records as indexed, but for which Xapian holds no document for that message's GUID. Messages that are not in cyrus.indexed.db are either pending work for the next indexer run, or already are overed by a document for the same GUID, but neither is index loss.

I'll take note to update the man page.

paulnivin:
Fine with closing this PR as well. Note that `squatter -A` can't detect a broken pre-#6045 state (my installation reported clean with 70 live messages unclaimed and undocumented), so perhaps the upgrade advice post #6045 shipping in a release is for installations to run `squatter -Z`?

rsto:
> perhaps the upgrade advice post #6045 shipping in a release is for installations to run `squatter -Z`?

I took note of this, I'll check if and how we can make this (and https://github.com/cyrusimap/cyrus-imapd/pull/6161) more clear in the upcoming release notes.

PR #6163, opened 2026-07-20
imapd: report slow/cancelled SORT and THREAD like SEARCH does

paulnivin:
Commit 78a791361 made SEARCH return `NO` with an error message when the search timer expires or the client cancels, but SORT and THREAD kept printing unconditional `OK` with whatever results had accumulated before the cutoff. index_sort()/index_thread() return only a count, so a search_maxtime expiry mid-scan silently truncates the result set — observed in production as `SORT (DATE) ...` returning `OK Completed (0 msgs)` after examining ~55k of 588k messages, indistinguishable from a genuine empty result.

This moves 78a791361's completion reporting into a shared search_reply_completed() helper used by cmd_search(), cmd_sort() and cmd_thread(), so all three report timer expiry and cancellation the same way — the same consolidation approach as b5fa6d319 took for the fuzzy check. Untagged data may already have been emitted before the NO; cmd_search has had the same property with ESEARCH since 2020 and the tagged status governs.

The added test uses criteria whose DNF normalisation reliably exceeds search_maxtime, independent of mailbox size, and asserts SEARCH, SORT and THREAD all report the timeout. It fails against unpatched cyrus and passes with the change (validated on a 3.12.2 backport; the master port of the test is syntax-converted from the validated version).

### DIFF
--- a/cassandane/tiny-tests/SearchFuzzy/search_sort_thread_maxtime
+++ b/cassandane/tiny-tests/SearchFuzzy/search_sort_thread_maxtime
@@ -1,0 +1,32 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_search_sort_thread_maxtime
+    :min_version_3_13 :needs_search_xapian :SearchMaxtime1Sec :SearchNormalizationMax20000
+    ($self)
+{
+    my $talk = $self->{store}->get_client();
+
+    $self->make_message("needle") || die;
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):

elliefm:
On my laptop, the SORT step of the new test runs fast enough to succeed instead of time out, and then the test fails. #6205 slows it down further.

PR #6165, opened 2026-07-21
jmap_mail_submission.c: return invalidRecipients rather than invalidProperties

ksmurchison:
The former is the more appropriate SetError for a bad rcpt email address

### DIFF
--- a/imap/jmap_mail_submission.c
+++ b/imap/jmap_mail_submission.c
@@ -488,6 +488,13 @@ static void _emailsubmission_create(jmap_req_t *req,
 {
     struct jmap_parser parser = JMAP_PARSER_INITIALIZER;
     struct buf buf = BUF_INITIALIZER;
+    char *mboxname = NULL;
+    struct mailbox *mbox = NULL;
+    json_t *myenvelope = NULL;
+    msgrecord_t *mr = NULL;
+    json_t *msg = NULL;
+    int fd_msg = -1;
+    int r = 0;
 
     /* messageId */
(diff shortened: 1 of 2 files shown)


imap/jmap_mail_submission.c
```diff
@@ -528,11 +528,21 @@ static void _emailsubmission_create(jmap_req_t *req,
...
+                    if (!invalid) invalid = json_array();
+                    json_array_append(invalid, json_object_get(addr, "email"));
+                }
                 jmap_parser_pop(&parser);
             }
+            if (invalid) {
+                *set_err = json_pack("{s:s s:o}", "type", "invalidRecipients",
+                                     "invalidRecipients", invalid);
+                jmap_parser_fini(&parser);
+                return;
```
rsto: I would prefer this to `goto done` so that future changes won't require to maintain cleaning up memory at two different places. But if there are strong reasons to return early here, I'm fine with it.

rsto (approved):

imap/jmap_mail_submission.c
ksmurchison: I returned here rather than `goto` because we define variables after this point which then get freed at `done`.  I can moved those variables to the top.

rsto: > I can moved those variables to the top

I would prefer that, it's one footgun less.

ksmurchison: Done

rsto (approved):

PR #6167, opened 2026-07-23
jmap_mail.c: gracefully handle duplicate modseqs with maxChanges

rsto:
This makes the Email/changes and Mailbox/changes commands handle duplicate modseqs in combination with the maxChanges argument. In a sane account, the modseqs never are duplicate. But to protect against the client state go corrupt in the erroneous case, we now overshoot the allowed maxChanges number of reported changes so that entries with duplicate modseqs do get reported as changed, rather than not silently omitted. Also, this patch logs the duplicate modseqs as an error.


ksmurchison (approved):
LGTM

PR #6171, opened 2026-07-24
jmap_admin.c: remove the Cyrus-custom JMAP admin capability

rsto:
This removes the custom 'https://cyrusimap.org/ns/jmap/admin' capability from the codebase. It only was relevant for a few data migrations at Fastmail and there is no point keeping on dragging it around.


ksmurchison (approved):

rsto:
@elliefm will ripping this out interfere with the 3.x releases?

elliefm:
I wouldn't backport it, so it shouldn't make any immediate difference.  The capability would continue to exist in the stable and older releases, but not on master.   Looks like it's existed since at least 3.8, I didn't check any further back.

If a security bug turns up in this feature, it will need to be fixed for the stable branches, even if the code no longer exists on master.  That means whoever is making and testing such a fix will need to do so against 3.12.  As far as I know, I'm the only dev who is routinely able to build and test 3.12, but I don't know the JMAP codebase.  (Presumably, once such fix is ready against 3.12, I would take care of backporting it to whichever older branches need it.)

None of this is to say that we shouldn't remove the capability, though, it's just flagging a hypothetical complexity if security bugs turn up in code that was removed from master.  Whether we remove this code or not, or whether we remove it now or later, that hypothetical complexity doesn't change.  Do we expect there might be security bugs in this code?  I don't know the feature, its context, or its implementation, so I don't know.

rsto:
> If a security bug turns up in this feature, it will need to be fixed for the stable branches, even if the code no longer exists on master. [...] Do we expect there might be security bugs in this code?

We do not expect security bugs in this code, there are just two JMAP methods implemented that only were relevant for rewriting data for experimental features. Both are guarded to reject any invocation other than the admin user, e.g. 
https://github.com/cyrusimap/cyrus-imapd/blob/8529841b16d16f07053f9827d798ad4599a3045e/imap/jmap_admin.c#L287-L290

I am going to remove them from master and in the unexpected case of having to backport changes in current stable will deal with the fallout.

PR #6173, opened 2026-07-24
gitignore: ignore gperf-generated code in the new jmap_props directory

rsto:
https://github.com/cyrusimap/cyrus-imapd/pull/6168 refactored the gperf files for JMAP into a separate directory, but the gitignore file needs to get updated to ignore them, too.


ksmurchison (approved):
Good catch. Should've done this myself

PR #6180, opened 2026-07-27
promstatsd timing fixes

elliefm:
This PR improves the regularity at which promstatsd produces prometheus reports.  Specifically, it fixes several sources of timing drift, and limits the impact an expensive report can have on other reports.

**Reviewers:** I suggest reading this commit by commit.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Prometheus.pm
+++ b/cassandane/Cassandane/Cyrus/Prometheus.pm
@@ -92,8 +92,14 @@ sub parse_report
     return $report;
 }
 
-sub slowtest_50000_users
-    :min_version_3_1
+# This is not a test, in that it doesn't really verify any particular result,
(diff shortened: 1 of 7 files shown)


rsto (commented):
I have neither developer nor user experience with prometheus. I remove myself from the reviewer list.

ksmurchison (approved):
Looks sane to me

wolfsage (commented):
Seems good.

In commit message:

```
promstatsd: clean up redundant defaults

config_getduration() already correctly handles defaults!

Anything hardcoded as a fallback from that never gets used, but
misleaders the reader.
```

"misleaders the reader" should probably be "misleads the reader"

wolfsage (approved):

rjbs (approved):
I large defer to my learned colleagues who got here first.  But looks good.  I will especially note that it's nice to see this building on some other recent successes. :)

PR #6182, opened 2026-07-27
sync_support.c: replicate mailbox deletedmodseq after sync_reset

rsto:
This fixes a bug where a JMAP client switching their Email/changes requests from master to the replica could silently get out of sync. This only occurs if the message got expunged from master before the replica could get notified about its deletion, and only if the replica recreated the mailbox from scratch. This happens both when the replica fully recreated the mailbox tree after a sync_reset, and when it replaced a mailbox whose uniqueid had changed. In these cases, the replica must receive the deletedmodseq from master, rather than keeping track of its own deletedmodseq as is done during regular replication.

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_changes_sync_reset_cannot_calculate_changes
+++ b/cassandane/tiny-tests/JMAPEmail/email_changes_sync_reset_cannot_calculate_changes
@@ -1,0 +1,92 @@
+#!perl
+use Cassandane::Tiny;
+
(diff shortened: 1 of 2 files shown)


brong (approved):
This looks good.  It wouldn't actually hurt to send a higher deletedmodseq to a replica, it just means it might return a cannotCalculateChanges to something which it actually had the data to do correctly -- but which the old master wouldn't have been able to.

Regardless, this is good and safe.

PR #6183, opened 2026-07-27
index.c: fix a lock-while-writing bug

rjbs:
Unlike other instances of the "tellchanges and also unlock" pattern, index.c would tell, then unlock.  In cases where tellchanges blocked, the lock could be held too long, blocking further updates.

The fix is simple: follow the usual "unlock, then tell" pattern.

The test is complex, setting up the conditions required for the tellchanges to block.


rjbs:
Original approval recorded in private repo as MR 94.

PR #6185, opened 2026-07-28
managesieve: link against libcap when enabled

edmundlod:
`perl/sieve/managesieve/Makefile.PL.in`'s `LIBS` list does not have `@LIBCAP_LIBS@` (unlike `perl/imap/Makefile.PL.in`).

This breaks `sieveshell` on builds that are built with `--with-libcap` e.g. Debian).

Since `libcyrus_min.a` (linked in as `MYEXTLIB`, so its own link requirements don't propagate) calls `cap_free()` when built `--with-libcap`, `managesieve.so` ends up with an undefined symbol at dlopen time, breaking `sieveshell`.

Fix by adding `@LIBCAP_LIBS@` to `LIBS`, matching perl/imap.


edmundlod:
Just for clarity's sake: This is again cyrus 3.12.

elliefm:
Thanks, this looks good, sorry about the delay.

Seems like our CI doesn't run on PRs submitted against the maint branches.  I don't expect CI to find any problems in this case, but that's not a reason to bypass it.

Could you please either a) submit a new PR against the master branch, or b) rebase/retarget this PR to the master branch, so that CI can run?  If you don't, I'll do it myself eventually, but I imagine you want your name on the work.  Please ping me @elliefm when it's ready and I'll shepherd it through.

PR #6187, opened 2026-07-28
jscalendar.c: VEVENT COLOR property has COLOR value

ksmurchison:
We were incorrectly setting as TEXT value resulting in VALUE=TEXT


rsto (approved):
I would prefer the tests to use the JMAPTester client, but it's not a deal-breaker to me.

PR #6190, opened 2026-07-28
prep release of 3.13.7 snapshot

rjbs:


rjbs:
@elliefm This was gonna be a Ken M. production, but we had to call an audible, so I did the minimum here.  Talk to you about it soon.

docsrc/download/release-notes/3.13/x/3.13.7.rst
```diff
@@ -0,0 +1,143 @@
...
+ * openio_autocreate
+ * openio_namespace
+ * openio_proxy_timeout
+ * openio_rawx_timeout
+ * openio_verbosity
+
+Security fixes
+==============
+
+* None yet!
```
elliefm: This snapshot has the same security fixes as 3.12.3, so you should list them here.  Easiest thing will be to copy them in from the 3.12.3 release notes.

elliefm (changes requested):
Need to list the security fixes!  Just copy paste from the 3.12.3 release notes, it's the same set.

Marking as "request changes" so that this can't accidentally land without that happening, but I don't personally need to review it again after you copy them in, so feel free to "dismiss stale review" at that point, and then anyone else can approve if I'm not around to.

elliefm (approved):

PR #6191, opened 2026-07-28
JMAP for Calendars: always use JSCalendar 2.0

rjbs:


rjbs:
@rsto There are a few big questions remaining, but I *think* they can be resolved later than merging this.  Will tell you more soon.

imap/jmap_calendar.c
```diff
@@ -3571,53 +3571,80 @@ jscal_cfg_t jmapical_ctx_to_jscalendar_cfg(struct jmapical_ctx *jmapctx)
...
 static json_t *ical_to_jsevents(jmap_req_t *req, icalcomponent *ical,
                                 hash_table *props,
                                 struct jmapical_ctx *jmapctx)
 {
     if (jmap_is_using(req, JMAP_JSCALENDARBIS_EXTENSION)) {
-        jscal_cfg_t cfg = jmapical_ctx_to_jscalendar_cfg(jmapctx);
-        if (jmap_is_using(req, JMAP_DEBUG_EXTENSION)) {
-            cfg.debug = true;
+        if (icalcomponent_isa(ical) != ICAL_XROOT_COMPONENT) {
+            return icalobj_to_jsevents_jscal(req, ical, props, jmapctx);
```
rsto: I would prefer `icalobj_to_jsevents_jscal` not getting introduced and rather `ical_to_jsevents` taking care of iCalendar object streams by itself. The `ical_to_jsevents` function already is the exception, rather than the rule, and introducing `icalobj_to_jsevents_jscal` only adds on that pile.

It seems to me that  only reason `icalobj_to_jsevents_jscal` got introduced is because the `icalcomponent *ical` argument can either be a single VCALENDAR component, or a XROOT with multiple VCALENDAR components embedded. With `icalobj_to_jsevents_jscal` already being a special-case handler, it might rather *embed* the sole VCALENDAR component in a XROOT, process that, then pop off the embedded VCALENDAR from the XROOT after.

Does this look reasonable of have I missed something?

cassandane/tiny-tests/JMAPCalendars/calendareventnotification_jscalendar2
```diff
@@ -0,0 +1,91 @@
+#!perl
```
rsto, line 1: nit: the commit message contains

> The new test test the shape of the notification content

This maybe means "The new test**s** test..."

cassandane/tiny-tests/JMAPCalendars/calendarevent_participantreply_patch
```diff
@@ -3,11 +3,9 @@ use Cassandane::Tiny;
 
 # Create a CalendarEvent, then reply to it via CalendarEvent/participantReply.
 #
-# CalendarEventNotifications and iMIP mboxevent payloads are still built as
-# JSCalendar 1.0, whatever the request asked for, so the patch in the
-# mboxevent uses the old-style RFC8984 participant identifiers. Verify that
-# those identifiers are the ones the client chose, and not something the
-# 1.0 converter invented.
+# The jsevent and patch in an iMIP mboxevent are JSCalendar 2.0, whatever the
```
rsto, line 6: nit: the commit message misses a verb:

> so they always JSCalendar 1.0.

I guess it means "they always **use** JSCalendar 1.0" or the like

(I can't comment on commit messages in the review tool so I just commented on some random line in the commit)

imap/jscalendar.h
```diff
@@ -77,6 +77,19 @@ icalcomponent *jscal_to_ical(jscal_cfg_t *cfg,
...
+ *  The Event is the first entry of the Group that jscal_from_ical would have
+ *  returned, with "version" copied down from the Group, since an Event without
+ *  an enclosing Group must set it. Use this where a caller wants one Event and
+ *  has nowhere to put the rest of the Group.
+ *
+ *  @param cfg   Conversion configuration, or NULL for defaults.
+ *  @param vcal  An iCalendar VCALENDAR component.
+ *  @return      A newly allocated JSON object representing a JSCalendar Event,
+ *               or NULL if the iCalendar object holds no entries. */
+json_t *jscal_event_from_ical(jscal_cfg_t *cfg, icalcomponent *vcal);
```
rsto, line 86: Non-actionable comment: I wish we would not declare this function in the jscalendar API, but it's also silly replicating that "take first Event from Group" code all over the place. So let's keep it here but let's also make sure we do not have any leftovers in other files that do the same thing (that being said, the function that converts all VEVENTs across all VCALENDAR components of an iCalendar object stream into the same array should stay in jmap_calendar.c.)

rsto (approved):
> Good bye to 105 functions, and nearly 7,500 LOC.  Take a moment to enjoy
> all that deletion.

🥳

This looks good. I only found one change in the jmap_calendar.c code where I would prefer not introducing a new function, but it's a minor edit and I won't insist on it. Other than that this is great and I look forward to see the 1.0 code gone for good.

imap/jmap_calendar.c
rjbs: It is very likely that what you want is better, but I need a better grasp on it.  I am going to carry on with testing the build as it is, and we can talk about this some time soon, since it should just be a refactoring, and not a major change.

rsto: It's just code hygiene, I'm fine if you merge and I might fix it up after.

rjbs:
@rsto I'll also note that rebasing this on master just now was not trivial, but I feel it's okay.  But FYI:  maybe I changed something you don't like. ;-)

Changes in b40331c491 and 783b0bafcf were the big complications, and I was sure to bump JMAPCACHE_CALVERSION again, although that wasn't technically a conflict.  Also, a test from the recently misc-fixes pulled out of GitLab was 1.0-centric.

I think nothing for you to do here unless you want to carefully re-review.  Just giving you a heads up.  Next step, build and put through its paces.

rsto:
> @rsto I'll also note that rebasing this on master just now was not trivial, but I feel it's okay. But FYI: maybe I changed something you don't like. ;-)
> 
> Changes in [b40331c](https://github.com/cyrusimap/cyrus-imapd/commit/b40331c4912efe7440dea1792da9ad42eaed4366) and [783b0ba](https://github.com/cyrusimap/cyrus-imapd/commit/783b0bafcfaf322fba6afeffc7b5714f5c399e2f) were the big complications, and I was sure to bump JMAPCACHE_CALVERSION again, although that wasn't technically a conflict. Also, a test from the recently misc-fixes pulled out of GitLab was 1.0-centric.
> 
> I think nothing for you to do here unless you want to carefully re-review. Just giving you a heads up. Next step, build and put through its paces.

I reviewed the resolved merge conflicts of the above commits and they look good to me.

rjbs:
Closed in favor of #6229

PR #6192, opened 2026-07-28
cassandane: rename assert_caldav_notified method

rjbs:
The name "assert_caldav_notified" makes it sound like "assert that a caldav user was notified".  We're actually asserting that the notifications queue looks like the given args.  This is most misleading when the *are* no given arts.  "assert notified" sounds like "that it was" when it actually means "that it was not"!


rsto (approved):
Agreed, this is better.

PR #6194, opened 2026-07-30
caldav: return ETag when uploading scheduling objects on PUT

### LINKED ISSUE #2456: Return ETag when uploading Scheduling Objects on PUT
Doing PUT in a cyrus 3.0 calendar, that creates a new scheduling object, returns Schedule-Tag header, but no ETag header.

As of https://tools.ietf.org/html/rfc6638#page-32 schedule-tag data is used to avoid conflicts, when the object is updated, while ETag is used in all other cases, e.g. synchronization.

Considering that CalDAV clients can do WebDAV synchronizaion and are possibly unaware of RFC6638, they have to make a GET after the PUT in order to the the ETag.  Doing experiments I cannot see, that the object was modified during uploading it and subsequent downloading it, hence the server should return Etag on PUT.

one-kash:
## Summary

Fixes #2456.

When a CalDAV client `PUT`s a scheduling object (an event/todo with an `ORGANIZER`), Cyrus returned the `Schedule-Tag` header but **no `ETag`**. `caldav_store_resource()` explicitly cleared the ETag/Last-Modified validators for scheduling objects, on the grounds that the iCalendar data had been rewritten during scheduling processing.

However, `dav_store_resource()` (and `caldav_get_validators()` for shared calendars) computes the validators from the **stored** resource, so they already reflect the rewritten data and are correct to return. Omitting the ETag breaks plain WebDAV-sync clients (e.g. Evolution) that are unaware of RFC 6638 and rely on the ETag for synchronisation. RFC 6638 treats `Schedule-Tag` and `ETag` as complementary (§3.2.10) and never forbids returning an ETag on PUT.

## Change

In `caldav_store_resource()`:
- Recompute the validators from the stored resource for scheduling objects too (removes the `!cdata->organizer` guard), so the returned ETag matches what a subsequent GET reports, including per-user data on shared calendars.
- Stop clearing `etag`/`lastmod` for scheduling objects.

The `Schedule-Tag` (`NEW_STAG`) assignment is unchanged, so both headers are now returned together. The separate ETag-clearing paths in `http_caldav.c` (stripped overrides, default alerts) are unaffected — they run after this function returns.

## Testing

- New Cassandane test `Caldav.put_scheduling_object_returns_etag`: PUT (create) and PUT (update) of a scheduling object both return `Schedule-Tag` **and** `ETag`; the ETag matches a subsequent GET and changes when the content changes.
- Full `Caldav` (109 tests) and `CaldavAlarm` (33 tests) suites pass; targeted regression checks pass, including `put_usedefaultalerts_no_etag` (still correctly returns no ETag), `put_changes_etag`, `shared_etag`, and all `rfc6638_*` scheduling tests.

### DIFF
--- a/cassandane/tiny-tests/Caldav/put_scheduling_object_returns_etag
+++ b/cassandane/tiny-tests/Caldav/put_scheduling_object_returns_etag
@@ -1,0 +1,88 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_put_scheduling_object_returns_etag
+    :min_version_3_13
+    ($self)
+{
+    my $CalDAV = $self->{caldav};
+
+    my $CalendarId = $CalDAV->NewCalendar({name => 'foo'});
+    $self->assert_not_null($CalendarId);
+
+    my $uuid = "6de280c9-edff-4019-8ebd-cfebc73f8201";
+    my $href = "$CalendarId/$uuid.ics";
+
+    # cassandane is the ORGANIZER of this scheduling object
+    my $card = <<EOF;
+BEGIN:VCALENDAR
+VERSION:2.0
+PRODID:-//Apple Inc.//Mac OS X 10.10.4//EN
+CALSCALE:GREGORIAN
+BEGIN:VEVENT
+CREATED:20150806T234327Z
+UID:$uuid
+DTEND;TZID=Australia/Melbourne:20160831T183000
+TRANSP:OPAQUE
+SUMMARY:An Event
+DTSTART;TZID=Australia/Melbourne:20160831T153000
+DTSTAMP:20150806T234327Z
+SEQUENCE:0
+ATTENDEE;PARTSTAT=ACCEPTED:MAILTO:cassandane\@example.com
+ATTENDEE;PARTSTAT=NEEDS-ACTION;RSVP=TRUE:MAILTO:test1\@example.com
+ORGANIZER;CN=Test User:MAILTO:cassandane\@example.com
+END:VEVENT
+END:VCALENDAR
+EOF
+
+    # annoyingly there's no "Request" that tells you the headers, so:
+    my %Headers = (
+        'Content-Type' => 'text/calendar',
+        'Authorization' => $CalDAV->auth_header(),
+    );
+
+    xlog $self, "PUT a new scheduling object as the ORGANIZER";
+    my $Response = $CalDAV->{ua}->request('PUT', $CalDAV->request_url($href), {
+        content => $card,
+        headers => \%Headers,
+    });
+
+    $self->assert_num_equals(201, $Response->{status});
+
+    xlog $self, "Server returns Schedule-Tag and ETag on create";
+    $self->assert_not_null($Response->{headers}{'schedule-tag'});
+    my $etag = $Response->{headers}{etag};
+    $self->assert_not_null($etag);
+
+    xlog $self, "Returned ETag matches the ETag of a subsequent GET";
+    $Response = $CalDAV->{ua}->request('GET', $CalDAV->request_url($href), {
+        headers => \%Headers,
+    });
+    $self->assert_num_equals(200, $Response->{status});
+    $self->assert_str_equals($etag, $Response->{headers}{etag});
+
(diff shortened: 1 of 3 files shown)


rsto:
@ksmurchison this looks reasonable to me, but in 2017 you had added the check that disables returning ETag for events with ORGANIZER: https://github.com/cyrusimap/cyrus-imapd/commit/50a6328a5fd Was this for interoperability?

dilyanpalauzov:
> Omitting the ETag breaks plain WebDAV-sync clients (e.g. Evolution) that are unaware of RFC 6638 and rely on the ETag for synchronisation.

Looking at https://gitlab.gnome.org/GNOME/evolution-data-server/-/work_items/45, my reading is that Evolution calls `GET` to obtain the `ETag`.  Thus Evolution is not broken.

Related to this I have locally this change since 2020:

caldav_put: do not send Schedule-Tag, if no scheduling was done
---------------------------------------------------------------
```diff
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -4154,10 +4154,10 @@ static int caldav_put(struct transaction_t *txn, void *obj,
                 if (ret) {
                     txn->error.precond = CALDAV_ALLOWED_ORG_CHANGE;
                 }
-                else {
-                    if (_scheduling_enabled(txn, mailbox) && !is_draft)
-                        sched_request(cal_ownerid, sched_userid, &schedule_addresses,
-                                      organizer, oldical, ical, SCHED_MECH_CALDAV);
+                else if (_scheduling_enabled(txn, mailbox) && !is_draft) {
+                    sched_request(cal_ownerid, sched_userid, &schedule_addresses,
+                                  organizer, oldical, ical, SCHED_MECH_CALDAV);
+                    flags |= NEW_STAG;
                 }
             }
             else {
@@ -4173,16 +4173,15 @@ static int caldav_put(struct transaction_t *txn, void *obj,
                     ret = HTTP_FORBIDDEN;
                 }
 #endif
-                else {
-                    if (_scheduling_enabled(txn, mailbox) && strarray_size(&schedule_addresses) && !is_draft)
-                        sched_reply(cal_ownerid, sched_userid, &schedule_addresses,
-                                    oldical, ical, SCHED_MECH_CALDAV);
+                else if (_scheduling_enabled(txn, mailbox) && strarray_size(&schedule_addresses) && !is_draft) {
+                    sched_reply(cal_ownerid, sched_userid, &schedule_addresses,
+                                oldical, ical, SCHED_MECH_CALDAV);
+                    flags |= NEW_STAG;
                 }
             }
 
             if (ret) goto done;
 
-            flags |= NEW_STAG;
         }
         break;

```

ksmurchison:
The problem with returning ETag when an organizer PUTs a scheduled event is that the server is adding SCHEDULE-STATUS parameters to the iCalendar data.  Returning an ETag to a PUT means that its the ETag of the representation sent by the client, NOT the content stored by the server.  If a client truly wants the stored content and its associated ETag, it should either include a `Prefer:return=representation` header field in the PUT request, or fetch the content with a subsequent GET request

one-kash:
> or fetch the content with a subsequent GET request

yeah, this is not a bad backup plan.

dilyanpalauzov:
Can the client predict, if the server will return `ETag` on `PUT` and based on the predicton send `Prefer:return=minimal` or `Prefer:return=representation` in order to avoid need for subsequent `GET` to get `ETag`?

Is it correct to say, that the server returns either `ETag` or `Schedule-Tag`?  When the server does not change `SCHEDULE-STATUS` value, the server should return `ETag`?

ETag can be sent when:
- the event does not have at the same time attendees and organizer, or
- neither organizer nor any attendee are in the calendar-user-address of the principal or of the calendar (the latter is Cyrus specific), or
- the calendar does no scheduling, because it has the `http://cyrusimap.org/ns/:scheduling-enabled` property with value `F`, or
- the event is uploaded with the header `Scheduling-Enabled: F` or `…: no` (this is Cyrus specific)?

Another case, where the Cyrus server modifies the event on upload, is when the server strips the `VTIMEZONE` component.

PR #6195, opened 2026-07-30
caldav: validate VTODO DUE against DTSTART on PUT

### LINKED ISSUE #4591: No validation that DUE and DTSTART have same value type
[RFC 5545 Section 3.8.2.3.  Date-Time Due](https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.2.3) says:

> For cases where this property is specified in a "VTODO" calendar component that also specifies a "DTSTART" property, the value type of this property MUST be the same as the "DTSTART" property, and the value of this property MUST be later in time than the value of the "DTSTART" property. Furthermore, this property MUST be specified as a date with local time if and only if the "DTSTART" property is also specified as a date with local time.

At https://github.com/cyrusimap/cyrus-imapd/pull/3795 I wanted to integrate validation in Cyrus IMAP on PUT, which checks for the above requirement.  This change request was superseded by https://github.com/cyrusimap/cyrus-imapd/pull/3802.

https://github.com/cyrusimap/cyrus-imapd/pull/3802 is closed with the remark “This PR has been superseded by later commits”.

Currently I find place in the code, which validates that DUE and DTSTART must have same value type.

one-kash:
## Summary

Fixes #4591.

RFC 5545 Section 3.8.2.3 requires that, when a VTODO specifies both `DTSTART` and `DUE`, the `DUE` property MUST have the same value type as `DTSTART` and MUST be later in time than `DTSTART`.

`validate_dtend_duration()` only enforced this for VEVENT `DTEND`/`DTSTART`, so an invalid VTODO — e.g. a `DATE-TIME` `DTSTART` with a `DATE` `DUE` — was accepted on PUT.

## Change

In `validate_dtend_duration()`:
- Fall back to `DUE` when `DTEND` is absent (the VTODO case) and apply the same value-type and ordering checks that already exist for `DTEND`.
- `DTSTART` is optional in a VTODO but required in a VEVENT, so the "DTSTART absent → skip comparison" short-circuit is scoped to `VTODO`: a VTODO with `DUE` and no `DTSTART` is accepted, while a VEVENT missing `DTSTART` continues to fail validation exactly as before.
- The function was also flattened with early returns to avoid the extra nesting and a duplicate `DTSTART` lookup.

## Testing

New Cassandane test `Caldav.put_vtodo_due_dtstart_mismatch` covers:
- VTODO with `DATE-TIME` `DTSTART` + `DATE` `DUE` → rejected (`valid-calendar-data`).
- VTODO with `DUE` and no `DTSTART` → accepted (DTSTART optional in VTODO).
- VEVENT with `DTEND` and no `DTSTART` → still rejected (relaxation must not leak to VEVENT).

Full `Caldav` (109 tests) and `JMAPCalendars` (306 tests) suites pass.

### DIFF
--- a/cassandane/tiny-tests/Caldav/put_vtodo_due_dtstart_mismatch
+++ b/cassandane/tiny-tests/Caldav/put_vtodo_due_dtstart_mismatch
@@ -1,0 +1,82 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_put_vtodo_due_dtstart_mismatch
+    :min_version_3_13
+    ($self)
+{
+    my $CalDAV = $self->{caldav};
+
+    my $CalendarId = $CalDAV->NewCalendar({name => 'foo'});
+    $self->assert_not_null($CalendarId);
+
+    # RFC 5545 3.8.2.3: in a VTODO, DUE MUST have the same value type as
+    # DTSTART.  Here DTSTART is a DATE-TIME but DUE is a DATE, which is invalid.
+    my $uuid = "5de280c9-edff-4019-8ebd-cfebc73f8202";
+    my $todo = <<EOF;
+BEGIN:VCALENDAR
+VERSION:2.0
+PRODID:-//Apple Inc.//Mac OS X 10.10.4//EN
+CALSCALE:GREGORIAN
+BEGIN:VTODO
+CREATED:20150806T234327Z
+UID:$uuid
+DTSTAMP:20150806T234327Z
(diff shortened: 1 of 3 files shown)


ksmurchison:
Thanks for this work.  We decided that all of this validation code should probably be in libical rather than in Cyrus.
That work, based partially on your is here: [https://github.com/libical/libical/pull/1353]

one-kash:
good deal.

PR #6196, opened 2026-07-30
caldav: return 404 for PROPFIND on a non-existing resource

### LINKED ISSUE #5667: WebDAV PROPFIND should return 404 on a non-existing resource
I'm using KDE's Dolphin with Cyrus WebDAV server. Every time I try to create a new file or folder Dolphin tells me the file already exists. 

From what I see Dolphin does PROPFIND request to `https://domain.tld/dav/drive/user/alec@domain.tld/non-existing-name`. This returns `HTTP/1.1 207 Multi-Status` with:
```
<?xml version="1.0" encoding="utf-8"?>
<d:multistatus xmlns:d="DAV:" xmlns:CS="http://calendarserver.org/ns/">
  <d:response>
    <d:href>/dav/drive/user/alec@domain.tld/non-existing-name</d:href>
    <d:status>HTTP/1.1 404 Not Found</d:status>
  </d:response>
</d:multistatus>
```
I think it should respond with `HTTP/1.1 404 Not Found` instead.

Cyrus 3.6.

one-kash:
## Problem

A WebDAV/CalDAV/CardDAV `PROPFIND` on a Request-URI that is **not mapped to a resource** returns a `207 Multi-Status` wrapping a nested `404`, instead of a top-level `404 Not Found`:

```
PROPFIND /dav/.../coll/does-not-exist.ics
=> HTTP/1.1 207 Multi-Status
   <d:multistatus>
     <d:response>
       <d:href>/dav/.../coll/does-not-exist.ics</d:href>
       <d:status>HTTP/1.1 404 Not Found</d:status>
     </d:response>
   </d:multistatus>
```

RFC 4918 §9.6 requires `GET/HEAD/PROPFIND` on an unmapped URL to return `404`:

> Thus, after a successful DELETE operation (and in the absence of other actions), a subsequent GET/HEAD/PROPFIND request to the target Request-URI MUST return 404 (Not Found).

Some plain WebDAV clients (e.g. KDE Dolphin) interpret the `207` as "the resource already exists" and refuse to create new files/folders.

Fixes #5667.

## Fix

`meth_propfind()` now checks, **before** emitting the multistatus response (the `207` status line is committed early, in the chunked path), that a resource-targeted request refers to a resource that exists in the DAV DB, and returns a top-level `404` if it does not. This mirrors the existing pre-checks in `meth_get_head()` and `meth_proppatch()`.

- Collection-targeted PROPFINDs are unaffected and still return `207`.
- The principal namespace (and root params) have no `davdb` ops, so the guard `fparams->davdb.open_db` short-circuits there — no behavior change and no NULL deref.
- `IMAP_MAILBOX_NONEXISTENT` from `mailbox_open_irl` is treated as `404` (not `500`), matching the adjacent home-set handling, to cover a collection deleted between lookup and open.

## Testing

Added `Caldav.propfind_nonexistent_resource` covering:
- missing resource in an existing collection → `404`
- missing resource in a missing collection → `404`
- existing collection → `207` (regression guard)

Verified at the HTTP surface (missing/existing resource, Depth 0/1, empty-body allprop, principal namespace) and ran the full `Caldav` (109) and `Carddav` (86) suites with no regressions.

### DIFF
--- a/cassandane/tiny-tests/Caldav/propfind_nonexistent_resource
+++ b/cassandane/tiny-tests/Caldav/propfind_nonexistent_resource
@@ -1,0 +1,46 @@
+#!perl
+use Cassandane::Tiny;
+
+# RFC 4918 Section 9.6: a PROPFIND on a Request-URI that is not mapped to a
+# resource MUST return 404 (Not Found), not a 207 Multi-Status wrapping a
+# nested 404.  (GitHub issue #5667)
+sub test_propfind_nonexistent_resource
+    :min_version_3_13
+    ($self)
+{
+    my $CalDAV = $self->{caldav};
+
+    my $CalendarId = $CalDAV->NewCalendar({name => 'foo'});
+    $self->assert_not_null($CalendarId);
+
+    my $propfind = <<'EOF';
+<?xml version="1.0" encoding="utf-8"?>
+<D:propfind xmlns:D="DAV:"><D:prop><D:getetag/></D:prop></D:propfind>
+EOF
+
+    my %headers = (
+        'Content-Type' => 'application/xml',
+        'Depth' => '0',
+        'Authorization' => $CalDAV->auth_header(),
+    );
+
+    # A PROPFIND on a non-existing resource in an existing collection must be
+    # a top-level 404, not a 207.
+    my $R = $CalDAV->{ua}->request('PROPFIND',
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):
This looks great.  Thank you!

PR #6198, opened 2026-07-30
Strip libical-internal X-LIC-ERROR from iCalendar data

rsto:
We already stripped X-LIC-ERROR in CalDAV GET, but multiget REPORT still contained the bogus properties. Also if the original iCalendar data contained X-LIC-ERROR properties we reject that data as invalid in PUT and incoming iTIP, even if it was Cyrus that had left these errors in the data in the first place.

### DIFF
--- a/cassandane/Cassandane/Cyrus/Sieve.pm
+++ b/cassandane/Cassandane/Cyrus/Sieve.pm
@@ -4,6 +4,7 @@
 package Cassandane::Cyrus::Sieve;
 use Net::CalDAVTalk 0.14;
 use Text::JSContact 0.01 qw(vcard_to_jscontact);
+use Text::VCardFast;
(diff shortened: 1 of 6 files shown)


wolfsage (approved):
Seems good. I wonder at the performance implications of doing it this way, but if that becomes a problem I suppose we can optimize then.

Looks like we only do this if we detect the string at all which should prevent any slowdown in the optimal cases, so probably fine

rsto:
> Seems good. I wonder at the performance implications of doing it this way, but if that becomes a problem I suppose we can optimize then.

Yeah, thought so, too. But hopefully it's rare, it only seems to have snuck in with the JSCalendar 2.0 code.

ksmurchison (approved):

PR #6199, opened 2026-07-31
GitHub Actions: be stricter about requiring changes files

rjbs:


rjbs:
@elliefm We'll want to talk about this team-wise before merging, but this is as discussed earlier this week.

elliefm (approved):
Is that language Go?  It looks kinda pythonish, but it has braces....

This seems fine, with the caveat that I don't know the language, so I'm reading it as pseudocode.  If you're holding it wrong, I won't pick that up.

I'm very pleased to see merge-on-demand learning to respect Do Not Merge.  I can't remember... do we already have a check that blocks normal merges when that label is present?  I've been wanting one forever, but I didn't just know how to set it up, and never tried to learn.  But maybe you or Matt added on at some point?  I dunno.

PR #6201, opened 2026-08-03
misc. changes from other branches

rjbs:
These are changes from external branches that are better served by being plucked out into an "miscellaneous little things" branch than staying tied up elsewhere.


rjbs:
These commits were all reviewed and approved elsewhere, so I did not wait for repeat approval here.

PR #6202, opened 2026-08-04
Sync stale cache selfheal

brong:
We had an issue where users would be wiped on a replica by a propogated sync_reset on the master, then synced in again and the cache on the master would make it think that the replica existed when in fact it didn't!

I looked a few different paths and settled on "every sync USER bypasses the cache entirely".  This was most robust, and full user syncs are rare in production.

### DIFF
--- a/imap/sync_support.c
+++ b/imap/sync_support.c
@@ -7637,10 +7637,17 @@ redo:
     struct dlist *kl = NULL;
     struct dlist *cachel = NULL;
 
+    /* only use the cache in normal rolling replication, that's where the bulk
+     * of the benefit comes from, and if we escalate to a full USER sync or
+     * run a sync manually, correctness outweighs efficiency, e.g. if the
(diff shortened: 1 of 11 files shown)


imap/sync_support.c
```diff
@@ -7631,10 +7641,17 @@ int sync_do_mailboxes(struct sync_client_state *sync_cs,
     struct dlist *kl = NULL;
     struct dlist *cachel = NULL;
 
+    /* the cache records what the replica looked like last time this channel's
+     * rolling process talked to it, so only it can trust a cached copy.  A
```
ksmurchison: I'm having a hard time parsing this comment

brong: Yeah that was a shit comment.  I rewrote it entirely to describe what's really going on.

ksmurchison (approved):

PR #6203, opened 2026-08-04
http_caldav.c: remove experimental secretary mode

rsto:
This removes the leftovers from the experimental calendar secretary mode feature.    

This was an experimental feature introduced in 3.8.0 and it was broken since. We removed JMAP support for it in 2022 but kept leftovers in the CalDAV implementation. Now it gets removed (almost) entirely:

The "shareesActAs" property of the JMAP calendar account is gone. The DAV property {urn:ietf:params:jmap:calendars}sharees-act-as on the calendar home still can be read using PROPFIND. It only can be removed using PROPPATCH, setting it to any value, including the formerly allowed values "self" and "secretary", now is forbidden.


ksmurchison (approved):
LGTM.  I didn't verify that any other artifacts might be left behind

PR #6204, opened 2026-08-04
jmap_email.c: prohibit storing messages in non-email mailboxes

ksmurchison:
Prior to this fix, we would attempt to store messages in calendars, addressbook, etc and would assert() due to missing/improper Content-Disposition headers

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -11160,6 +11160,16 @@ static int _email_to_mime(jmap_req_t *req, FILE *fp, void *rock, json_t **err)
     return 0;
 }
 
+static bool is_appendmailbox(jmap_req_t *req,
+                             const mbentry_t *mbentry, int need_rights)
+{
+    return (mbentry &&
+            !(mbentry->mbtype & MBTYPE_DELETED) &&
+            !mboxname_isdeletedmailbox(mbentry->name, NULL) &&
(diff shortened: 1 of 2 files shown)


imap/jmap_mail.c
```diff
@@ -12796,14 +12797,20 @@ static int _email_bulkupdate_open(jmap_req_t *req, struct email_bulkupdate *bulk
                 json_object_del(update->mailboxids, mbox_id);
                 json_object_set(update->mailboxids, mbentry->uniqueid, jval);
                 mbox_id = mbentry->uniqueid;
-                is_valid = 1;
+
+                is_valid = !((mbentry->mbtype & MBTYPE_DELETED) ||
```
rsto: This could get refactored with the conditional in lines [11226-L11229](https://github.com/ksmurchison/cyrus-imapd/blob/a790ef8b7cad8005856a5bea2e86a14b77eb6eda/imap/jmap_mail.c#L11226-L11229) into `is_appendmailbox` or the like

rsto (approved):
LGTM - maybe refactor the duplicate conditionals

imap/jmap_mail.c
ksmurchison: Done

rsto (approved):

PR #6205, opened 2026-08-05
SearchFuzzy: slow down search_sort_thread_maxtime

elliefm:
On my laptop the SORT was running fast enough to succeed rather than time out, and then the test would fail.

Adding one more step to the exponential criteria construction slows it down enough to pass, without making the whole test unbearably slow.  My laptop runtime for this test is now ~11 seconds, and it seems to pass reliably.

I tried adding another extra step, but that stretched it to over a minute, so I backed it out.  If it turns out we do need another step for full reliability, we might add the `_slow` suffix too.

### DIFF
--- a/cassandane/tiny-tests/SearchFuzzy/search_sort_thread_maxtime
+++ b/cassandane/tiny-tests/SearchFuzzy/search_sort_thread_maxtime
@@ -2,7 +2,7 @@
 use Cassandane::Tiny;
 
 sub test_search_sort_thread_maxtime
-    :min_version_3_13 :needs_search_xapian :SearchMaxtime1Sec :SearchNormalizationMax20000
+    :needs_search_xapian :SearchMaxtime1Sec :SearchNormalizationMax20000
     ($self)
 {
     my $talk = $self->{store}->get_client();
@@ -13,7 +13,7 @@ sub test_search_sort_thread_maxtime
     # criteria whose DNF normalisation reliably exceeds search_maxtime,
     # independent of mailbox size
     my $crit = 'SUBJECT a';
-    $crit = "OR NOT $crit NOT $crit" for (1..14);
+    $crit = "OR NOT $crit NOT $crit" for (1..15);
 
     xlog "SEARCH must report the timeout";
     my $uids = $talk->search(\$crit);


elliefm:
@rjbs I think your laptop is even faster than mine, so I'm particularly interested in whether the original test fails for you, and whether this change is enough to make it pass.

cassandane/tiny-tests/SearchFuzzy/search_sort_thread_maxtime
rsto: While we are it: I thought we don't use these annotations anymore in new tests?

rsto, line 4: I suggest to gate this test with the magic `_slow` suffix.

rsto (commented):
The original test fails on my M4 Pro, with this change it passes reliably (n=5).
In the past years, these tests that expect execution times turned out be flaky and their failure modes tend to turn up at the worst possible time (e.g. we want to test and ship a build fast). I wonder if we should just get rid of the test.

The test for this error is particularly difficult to get right: we can't crank up the complexity too high because otherwise the search runner will reject the query outright. It has to find the sweet spot of being in the window of allowed complexity, but effectively too high for anything other than very beefy machines.

cassandane/tiny-tests/SearchFuzzy/search_sort_thread_maxtime
ksmurchison: Agreed

ksmurchison: Agreed

paulnivin: I've updated my related search PRs (#6159, #6160, #6161, #6162, #6164) to conform to the new test syntax. Let me know if you have any questions on them.

rjbs:
As far as reviewers go, I think I'm at the "too many cooks" limit, and so am removing myself.  Having this not run by default does seems like a good idea, given it's been giving us heisenfails.  And if it's going to do that when we run it by hand, I'm not sure it's *very* helpful.

Anyway you asked about my fast laptop.  Robert's is faster and he answered. :)

cassandane/tiny-tests/SearchFuzzy/search_sort_thread_maxtime
elliefm: 11 seconds runtime isn't worth gating behind `--slow`.  That's not long enough to notice unless you're only running that one test, and if you're only running one test, 11 seconds isn't a long time to wait.  It won't meaningfully delay other tests during a normal run.

In my mind, `--slow` is for gating tests that take _minutes_, and don't amortise well on parallel runs.  Out of ~3000 tests, we only have eight of these, but running `./testrunner --slow-only` on my system takes 5 minutes.  11 seconds is nothing next to those tests.

If we added _another_ exponential step, then that would push this test's runtime on my system to ~70 seconds.  That would be worth gating.  But based on Robert's laptop results, we don't need to do that; the one extra step is sufficient.

elliefm:
I don't think we need to worry much about this test being flaky, once it's set adequately.  Remember, the +1 step increase here is exponential, not linear.  The flaky timing-based tests we've dealt with in the past have been due to linear time differences, which tend to jitter.

If it _does_ start producing false negatives again in the future, we might choose to deal with it differently then.  But for now, I think the test is useful and worth keeping, and this PR fixes the current problem.

cassandane/tiny-tests/SearchFuzzy/search_sort_thread_maxtime
elliefm: For comparison, the whole SearchFuzzy suite (including this test) takes 15 seconds on my laptop.  The SearchFuzzy suite minus this test takes 7 seconds.  So, even though it takes 11 seconds in isolation, as part of the suite it only adds 8 seconds.  If you run more suites, it matters even less.

elliefm: I've removed the min version attribute.

elliefm:
The original test is sometimes running too fast in CI too, not just on fast laptops.  So can we please get this fix approved and merged without too much bikeshedding?  Thanks.

ksmurchison (approved):

PR #6206, opened 2026-08-05
cassandane: make verbosity 2 actually useful

elliefm:
... by moving the IMAP protocol chatter from verbosity 1 to 3, and the JMAP protocol chatter from verbosity 0 to 3.

Some tests produce huge amounts of protocol chatter output, which makes the other output hard to see.  With the protocol chatter moved to verbosity 3, verbosity 2 becomes useful for seeing the other verbose output without it.

### DIFF
--- a/cassandane/Cassandane/IMAPMessageStore.pm
+++ b/cassandane/Cassandane/IMAPMessageStore.pm
@@ -106,7 +106,7 @@ sub connect
     }
 
     $client->set_tracing(1)
-        if $self->{verbose};
+        if $self->{verbose} >= 3;
 
     my $banner = $client->get_response_code('remainder');
     if (not $params{NoLogin}) {
--- a/cassandane/Cassandane/Instance.pm
+++ b/cassandane/Cassandane/Instance.pm
@@ -3317,7 +3317,7 @@ sub _new_jmaptester_for_user($self, $tester_class, $tester_arg, $user, $new_arg 
 
     # This causes all requests and responses to be printed to STDERR.
     local $ENV{JMAP_TESTER_LOGGER} = 'HTTP:-2'
-        unless exists $ENV{JMAP_TESTER_LOGGER};
+        if get_verbose() >= 3 && not exists $ENV{JMAP_TESTER_LOGGER};
 
     my $jtest = $tester_class->new({
         fallback_account_id => $user->username,


rsto (approved):
Works for me - I always run failing Cassandane tests with verbose level 3.

ksmurchison (approved):

wolfsage:
Will this make the ci runs less useful?

That is:

```
    - name: run cassandane quietly
      if: ${{ ! matrix.skip-cass }}
      id: cass1
      continue-on-error: true
      run: cyd test --no-ok --slow --format prettier "${{ matrix.skip }}"

```

Will we no longer have protocol output in the failures? IMO that protocol output is useful

rjbs:
@wolfsage If I'm reading ellie correctly, we were only sending JMAP, not IMAP, telemetry in these cases.  So, this is more consistent and if you want the logging, I don't think "not changing" is better.

But I do agree that if the test fails in CI, I want to know what happened in "as much details as possible".  I would want something like "verbosity zero but on failure show everything".  I think we can make that happen.  I am approving on the basis of "this is more consistent" but we should decide what we want holistically.

rjbs (approved):

elliefm:
The only verbosities that have been distinctly useful until now were 0 and 3.  If you wanted everything useful, you asked for 3; if you wanted nothing, you asked for 0 (or didn't ask).  Verbosities 1 and 2 were approximately the same as 3, with a few random details missing.   This change makes verbosity 2 useful.  Verbosity 1 is still not very useful.

I consider "JMAP protocol chatter at verbosity 0" to have been a bug that is now fixed.

I haven't found cassandane output _from CI_ to be particularly useful, other than the pass/fail status and the backtrace of the failure.  The web interface is terrible for reading a lot of log output, so if I need that detail I reproduce it locally.  CI deliberately runs at verbosity 0 to make it easy-ish to find the backtrace details.  If JMAP chatter was appearing in this output, that was a result of the bug, not an intentional design choice.

>  I would want something like "verbosity zero but on failure show everything". 

We have that, it's the "run cassandane quietly" and "rerun cassandane failures noisily" steps in the workflow.  If you want _verbose_ output on failures, make the noisy step ask for verbose output.  Though I expect that will just make it harder to find the backtrace in the output, and therefore less useful.

PR #6207, opened 2026-08-05
deduplicate: new utility to hardlink message files sharing a GUID

zisisvakras:
This introduces a new two-phase utility that recovers the hardlinks `singleinstancestore` normally provides at delivery time. It's the offline complement to that option: it can collapse duplicate spool files across the whole mailstore, even messages delivered before `singleinstancestore` was enabled, or duplicates that arose despite it.

My reason for making this is that we recently migrated to 3.12 by xfer-ing a user at a time, which was unavoidable for our process. This inflated our storage by about 20% and we wanted to reclaim the lost space.

Full details and examples are in the man page and the admin SOP guide included.
I want to be brief, so I will refer you to the documentation provided.

I have also included what I believe to be a somewhat complete Cassandane suite.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -218,6 +218,7 @@ sbin_PROGRAMS += \
     imap/cyr_synclog \
     imap/cyr_userseen \
     imap/cyr_virusscan \
+    imap/deduplicate \
     imap/deliver \
     imap/ipurge \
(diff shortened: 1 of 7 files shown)


zisisvakras:
This test failure seems to be related to #6205

PR #6213, opened 2026-08-07
add an option for "always use compact ids"

rjbs:
The real future is "only compact ids exist", but we need a release that supports both for the sake of upgrading.  This lets a fully upgraded install (and I know of a pretty big one) set this option and avoid any sort of fluke with conversation db downgrading a user.

In the next next release, we'll strip this option along with fat ids entirely.


ksmurchison (approved):
LGTM

PR #6215, opened 2026-08-07
cassandane: allow globbing of test case names

rjbs:
My real goal was to make `dar test JMAP*` work.  It took more than the five lines I was hoping for.

First, make the thing testable.  It was critical and untested.

Then, fix the things that testing exposed that seemed goofy: zero-test runs, how slow test selection affected things.

Then, refactoring to allow test case globbing to work at all.

Then the thing itself!

One thing I noticed while in here:  We say "suite" a lot for a test case, but the Test::Unit jargon is "test case".  I'd like us to reduce duplicative jargon.  But that can happen later, this already ate a lot more time than I meant it to!

### DIFF
--- a/cassandane/Cassandane/Fixture/TestPlan/Alpha/GlobOne.pm
+++ b/cassandane/Cassandane/Fixture/TestPlan/Alpha/GlobOne.pm
@@ -1,0 +1,16 @@
+# SPDX-License-Identifier: BSD-3-Clause-CMU
+# See COPYING file at the root of the distribution for more details.
+
+# A fixture suite for Cassandane::Test::TestPlan.  Not a test!  See the
+# README in the parent directory.
+package Cassandane::Fixture::TestPlan::Alpha::GlobOne;
+use strict;
+use warnings;
+
+use base qw(Cassandane::Unit::TestCase);
+
+sub test_alpha { }
+sub test_beta { }
+sub test_gamma_slow { }
+
+1;
--- a/cassandane/Cassandane/Unit/TestPlan.pm
+++ b/cassandane/Cassandane/Unit/TestPlan.pm
@@ -544,7 +544,8 @@ sub _schedule
     }
 }
 
-# Returns ($neg, $ostype, $ospath, $testname)
+# Returns a list of [ $neg, $ostype, $ospath, $testname ] tuples.  It's a list
+# because one specification can name more than one suite.
 sub _parse_test_spec
 {
     my ($self, $name) = @_;
@@ -572,16 +573,16 @@ sub _parse_test_spec
     foreach my $candidate (@paths) {
         foreach my $root ($self->{test_roots}->@*)
         {
-            return ($neg, 'd', $candidate, undef)
+            return [ $neg, q{d}, $candidate, undef ]
                 if ($root eq $candidate);
 
             my $fpath = $candidate;
             $fpath = "$root/$candidate"
                 if ("$root/" ne substr($candidate, 0, length($root)+1));
 
-            return ($neg, 'd', $fpath, undef)
+            return [ $neg, q{d}, $fpath, undef ]
                 if ( -d $fpath );
-            return ($neg, 'f', "$fpath.pm", undef)
+            return [ $neg, q{f}, "$fpath.pm", undef ]
                 if ( -f "$fpath.pm" );
 
             my $test;
@@ -592,9 +593,9 @@ sub _parse_test_spec
                 if ($test =~ /\*/) {
                     my @hunks = split /\*+/, $test, -1;
                     my $regex = join q{.*}, map {; quotemeta } @hunks;
-                    return ($neg, 'f', "$fpath.pm", qr{\A$regex\z});
+                    return [ $neg, q{f}, "$fpath.pm", qr{\A$regex\z} ];
                 } else {
-                    return ($neg, 'f', "$fpath.pm", $test)
+                    return [ $neg, q{f}, "$fpath.pm", $test ]
                 }
             }
         }
@@ -645,22 +646,25 @@ sub schedule
 
     foreach my $name (@names)
     {
-        my ($neg, $type, $path, $test) = $self->_parse_test_spec($name);
+        foreach my $spec ($self->_parse_test_spec($name))
+        {
+            my ($neg, $type, $path, $test) = @$spec;
 
-        if ($type eq 'd')
-        {
-            opendir DIR, $path
-                or die "Cannot open directory $path for reading: $!";
-            while ($_ = readdir DIR)
+            if ($type eq 'd')
             {
-                next unless m/\.pm$/;
-                $self->_schedule($neg, "$path/$_", undef, $name);
+                opendir DIR, $path
+                    or die "Cannot open directory $path for reading: $!";
+                while ($_ = readdir DIR)
+                {
+                    next unless m/\.pm$/;
+                    $self->_schedule($neg, "$path/$_", undef, $name);
+                }
+                closedir DIR;
             }
-            closedir DIR;
-        }
-        else
-        {
-            $self->_schedule($neg, $path, $test, $name);
+            else
+            {
+                $self->_schedule($neg, $path, $test, $name);
+            }
         }
     }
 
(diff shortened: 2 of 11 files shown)


elliefm:
> We say "suite" a lot for a test case, but the Test::Unit jargon is "test case".

(That's not quite right, but we'll get into the minutiae in a minute...)

So, this mismatch is a consequence of the way cassandane abuses Test::Unit.

We (Cyrus developers) use _suite_ to mean a collection of tests (like "JMAPCore" or "SearchFuzzy"), and _test_ or _test case_ to mean an individual test within such a collection (like "JMAPCore.echo_tls" or "ACL.delete").  We also use this terminology when talking about cunit tests, as does CUnit itself, so it is pretty consistent across the project as a whole.

Test::Unit has `TestSuite` and `TestCase`, so you'd think cassandane would just use those directly, but it doesn't.  Instead, it ignores `TestSuite`, and uses a mess of sketchy subclassing to more or less entirely redefine `TestCase`.  The only explanation I've ever found is this comment:

https://github.com/cyrusimap/cyrus-imapd/blob/229cdb2b444d68157a736122c571778cb6bbf429/cassandane/Cassandane/Unit/TestPlan.pm#L980-L985

It's a deeply unsatisfying explanation.  It's unclear what problem all this crap is supposed to solve.

Cassandane calls the mess of sketchy subclassing a `TestPlan`.  That appears to be cassandane-specific terminology, not something from Test::Unit.  It might be lingo borrowed from some other testing architecture that Cassandane was trying to mimic?  It doesn't directly map to _suite_ either; the _plan_ is approximately "all the tests we're going to run", which may be one _test_, or many _tests_ from many _suites_.  Cassandane tells Test::Unit that the _entire plan_ is a single TestCase.  Uggh.

cassandane/Cassandane/Fixture/TestPlan/Alpha/GlobOne.pm
```diff
@@ -0,0 +1,16 @@
+# SPDX-License-Identifier: BSD-3-Clause-CMU
+# See COPYING file at the root of the distribution for more details.
+
+# A fixture suite for Cassandane::Test::TestPlan.  Not a test!  See the
+# README in the parent directory.
```
elliefm, line 5: I don't see this README anywhere?

From commit message:
> This needs a fixed set of data [...]

For whatever it's worth, we already have `cassandane/data/` for various data files needed by tests.  That might be a good place to stash these ones?

cassandane/Cassandane/Unit/TestPlan.pm
```diff
@@ -75,6 +81,32 @@ sub _allow
...
 }
 
+# Returns, for each specification that selected individual tests from this
+# suite, a hash of the specification and the test names it actually matched.
+# Loading the suite is the only way to know those names, so this can't be
+# answered until scheduling is finished.
+#
+# Missing suppressions aren't a problem.  We want to keep working against
+# versions of Cyrus where the named test doesn't exist.
+sub _spec_matches ($self)
```
elliefm: The comment about missing suppressions feels misplaced, we're not doing any suppression logic here, just building a set of matches (possibly empty).

The function name is awkward since "matches" could be a noun or a verb.  And we already have `assert_matches` in the verb sense from Test::Unit, so we're primed to read it wrong.  Maybe `_spec_get_matches`?

cassandane/Cassandane/Unit/TestPlan.pm
```diff
@@ -695,13 +673,34 @@ sub schedule
...
     die "No tests matched: " . join(q{, }, @unmatched) . "\n" if @unmatched;
+
+    foreach my $match (@matches)
+    {
+        my @tests = $match->{tests}->@*;
+
+        if ($self->{skip_slow} and not grep {; $_ !~ m/_slow$/ } @tests)
+        {
+            xlog "$match->{spec} was explicitly requested. Enabling slow tests!";
+            $self->{skip_slow} = 0;
```
elliefm: That double negative is really hard to read.   Could we not just do a positive grep for tests that _do_ contain `/_slow$/`?

Right now it _looks like_ there's a bug where we'll mistake an empty set for "some slow test was explicitly requested".  There isn't such a bug, because if there were an empty set we'd have bailed out before the loop.  But it was a goose chase to figure that out.

cassandane/Cassandane/Unit/TestPlan.pm
```diff
@@ -612,44 +697,78 @@ sub schedule
...
 
-        if ($type eq 'd')
+    my @unmatched = grep {; ! $matched{$_}->@* } @specs;
+    die "No tests matched: " . join(q{, }, @unmatched) . "\n" if @unmatched;
+
+    foreach my $spec (@specs)
+    {
+        my @tests = $matched{$spec}->@*;
+
+        if ($self->{skip_slow} and not grep {; $_ !~ m/_slow$/ } @tests)
```
elliefm: Could we not just do a positive grep for tests that do match `/_slow$/`?  The double negative makes it look like there's a bug by which we'll mistake an empty set for a set containing only slow tests.  I don't think there is such a bug, because if the set were empty we would've rejected it before the loop, but it's a goose chase to get to that conclusion.

elliefm (commented):
This seems fine, a few detail nits.

Github seems to have gotten very confused about my review comments and which lines they apply to.  Apologies if they're a mess or if there's duplicates.  I'm not sure they won't all just disappear when I click submit...

elliefm:
Hmm...

> If a test spec selects nothing, die

Absolutely agreed.  An edge case I just thought of though: if the test spec selects some non-zero number of tests, but which all end up skipped by a filter, do we detect that case and die?

I think maybe you handled this case too, but I don't remember exactly and haven't dug back in to look.

cassandane/Cassandane/Fixture/TestPlan/Alpha/GlobOne.pm
rjbs: I know I wrote it, so I guess I never `git add`-ed it and it got lost at some point.

We spoke about this and I said, "Maybe I can fiddle with `@INC`."  Short version:  I don't think I will, today.  It can be done, but we're not really using `@INC` the way that would make that simplest, and it would mean more complexifying of TestPlan.  I think for now, I will rewrite the README, make this clear, and then maybe in some future pass clean this up more.  We already have evidence that getting the "0 test plan is fatal" logic done is a good idea.

rjbs: README, without leading `/`, was in .gitignore!

cassandane/Cassandane/Unit/TestPlan.pm
rjbs: How odd.  I remember writing this comment, but I thought it was somewhere else.  I'm just going to delete the comment.  Point taken on the sub name, I'll rename it.  On a future go-round, I foresee a lot of renaming things into coherence…

rjbs: I renamed it `_get_candidates` for now.

cassandane/Cassandane/Unit/TestPlan.pm
rjbs: Yes, copypasta, gross.  Fixed.

rjbs:
> Absolutely agreed. An edge case I just thought of though: if the test spec selects some non-zero number of tests, but which all end up skipped by a filter, do we detect that case and die?

```
tove:~/fm/code/cyrus/imapd$ dar test Sieve.enotify_text_variable_bad '!Sieve.enotify_text_variable_bad'
⏳ Executing command in container 2920d3597739...
make[1]: Entering directory '/srv/cyrus-imapd/cassandane/utils'
make[1]: Nothing to be done for 'all'.
make[1]: Leaving directory '/srv/cyrus-imapd/cassandane/utils'

Time:  0 wallclock secs ( 0.00 usr  0.01 sys +  0.09 cusr  0.16 csys =  0.26 CPU)
Finished: 21:28:49

OK (0 tests)
```

Looks like no!

elliefm (approved):

PR #6216, opened 2026-08-09
jmap_mailbox: don't return more changes than maxChanges

rjbs:
RFC 8620 section 5.2 says the server MUST ensure the number of ids returned across created, updated and destroyed does not exceed maxChanges, and that where it can't calculate an intermediate state it MUST return cannotCalculateChanges instead.  Mailbox/changes was deliberately overshooting.

The old code seemed to think this couldn't come up, anyway: "duplicate modseqs must not ever happen in a sane account", but that's not right. Two Mailboxes can end up with the same modseq based on the conversations in them.  In fact, I found this happening in our tests, in JMAPMailbox/mailbox_changes_on_thread_counts.

### DIFF
--- a/cassandane/Cassandane/Util/CRLF.pm
+++ b/cassandane/Cassandane/Util/CRLF.pm
@@ -1,0 +1,23 @@
+# SPDX-License-Identifier: BSD-3-Clause-CMU
+# See COPYING file at the root of the distribution for more details.
+
+package Cassandane::Util::CRLF;
+use strict;
+use warnings;
+use experimental 'signatures';
+use base qw(Exporter);
+
+our @EXPORT = qw(&to_crlf);
+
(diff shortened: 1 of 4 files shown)


rsto:
We deliberately decided to overshoot maxChanges, but I agree it's better to return `cannotCalculateChanges` if the maxChanges window is too small to ever fit all mailboxes with that modseq.

More importantly, this refutes the "all modseqs are unique" assumption and it not only does for mailboxes but also for emails. This needs more work and I've started building on top of your branch a change that also includes JMAP emails. I also would have some feedback on the cassandane test, but we can revisit that when I have a counter-proposal PR.

rsto (commented):
This comment is just here to make Github review tooling happy.

rjbs:
Robert is replacing this with an alternate take.

PR #6217, opened 2026-08-10
contrib: add twom-trace, an eBPF census and benchmark for cyrusdb

brong:
There was no way to see what shapes of database calls cyrus actually makes on a live server short of patching in counters and restarting, which rules out looking at production. These uprobes attach to the running libcyrus instead: no cyrus change, no restart, and they cover every process on the box including ones that start mid-capture.

twom-capture reports the operation mix, flags, key and value sizes and scan prefix lengths per database; twom-hitmiss splits lookups into hit, tombstone and never-existed, which the return code alone cannot tell you. Everything aggregates in kernel maps, so cost is one trap plus one string copy per call and nothing is emitted per event.

dbbench then replays the measured shape against any cyrusdb backend and reports CPU and disk alongside throughput, so "twom is faster" can be a number rather than a belief. It shares its workload with twom-bench so the single-engine and cross-engine runs cannot describe different things.

The struct offsets the tracer walks are checked against DWARF before each run and the flag names are read from the library being traced, because a wrong offset does not fail - it quietly reports nonsense.

### DIFF
--- a/contrib/twom-trace/.gitignore
+++ b/contrib/twom-trace/.gitignore
@@ -1,0 +1,3 @@
+build/
+twom-bench
+dbbench
--- a/contrib/twom-trace/README.md
+++ b/contrib/twom-trace/README.md
@@ -1,0 +1,157 @@
+# twom-trace
+
+Census of twom access patterns inside Cyrus, via eBPF uprobes on
+`libcyrus.so.0`, plus a benchmark that replays the measured shape against any
+cyrusdb backend.
+
+The tracer answers "what shapes of database calls does this server actually
+make", per database type: operation mix, flags, key/value sizes, scan prefix
+lengths, hit/tombstone/miss split, and shared-vs-exclusive transaction rates.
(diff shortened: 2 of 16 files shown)


rjbs (approved):
Let me be really clear:  I did not really substantially review this.  I am not competent to give a judgement on the C changes.  The Perl reporting tools look, like… fine.  They're not useful for me to review without knowing more about the domain.

The idea of having this is (of course) terrific.  I assume you have shown this builds and works.  The big question is, "would this cause trouble when in use".  We had some process — I forget the details — hard crash the kernel when we were ebpf-tracing a few years ago.  We should hit test environments with this (and given it was a kernel crasher, starting not on real imaprepls).

I will be trying to understand more of how this code works because I'm interested, but mostly I think "looks like something we could use".  If you're hoping for a more substantial crawl through the C, though, you should ask somebody else!

brong:
I have run these on a prod host to get some initial values.  The benching tool I expect to use on an otherwise empty host (say one of the STL imaps) to hammer the different database engines.

PR #6218, opened 2026-08-10
Zeroskip

brong:
This is a cleanroom, Claude implementation of Zeroskip.  It has its own repo which I've pushed as a replacement over https://github.com/brong/zeroskip

I spent a good couple of days refining the data structure and spec for this, so hopefully it's good :). Will definitely be testing it first.

This MR starts with a bunch of fixes to the library code to handle a database which is a directory rather than a single file.

I'm not planning to run this LIVE yet, but the supporting code is needed for any database format which is a directory, so that's the key bit to be separately reviewed.

### DIFF
--- a/lib/cyrusdb.c
+++ b/lib/cyrusdb.c
@@ -40,6 +40,7 @@ extern struct cyrusdb_backend cyrusdb_quotalegacy;
 extern struct cyrusdb_backend cyrusdb_sql;
 extern struct cyrusdb_backend cyrusdb_twom;
 extern struct cyrusdb_backend cyrusdb_twoskip;
+extern struct cyrusdb_backend cyrusdb_zeroskip;
 
 static struct cyrusdb_backend *_backends[] = {
     &cyrusdb_flat,
@@ -50,6 +51,7 @@ static struct cyrusdb_backend *_backends[] = {
 #endif
     &cyrusdb_twom,
     &cyrusdb_twoskip,
+    &cyrusdb_zeroskip,
     NULL };
 
 #define DEFAULT_BACKEND "twoskip"
--- a/lib/cyrusdb.c
+++ b/lib/cyrusdb.c
@@ -18,6 +18,7 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <dirent.h>
+#include <errno.h>
 
 #include "assert.h"
 #include "bsearch.h"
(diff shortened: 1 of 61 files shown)


lib/cyrusdb.c
```diff
@@ -618,8 +619,32 @@ EXPORTED const char *cyrusdb_detect(const char *fname)
 {
     FILE *f;
     char buf[32];
+    struct stat sbuf;
     int n;
 
+    if (stat(fname, &sbuf)) return NULL;
+
+    /* Every other backend is a single file identified by magic, but a zeroskip
+     * database is a directory.  Both its data files and its lock file are named
```
rjbs: not only zeroskip, according to previous commit, also quotalegacy, I think? (nitpick)

lib/cyrusdb.c
```diff
@@ -170,6 +170,21 @@ static int _myopen(const char *backend, const char *fname,
 
     /* check if it opens normally.  Horray */
     r = db->backend->open(fname, flags, &db->engine, tid);
+
+    /* A zeroskip database is a directory and every other backend is a file, so
+     * a shape mismatch fails deep in the backend as an I/O error - EISDIR from
+     * a file backend, ENOTDIR from zeroskip.  Neither reaches the conversion
+     * path on its own, so recognise it here, where the backend's expectation is
+     * known, and report it as the format mismatch it is. */
```
rjbs: I won't harp on it, so second and last time:  "every other backend" — is it?  or more to the point, maybe we should just say "directory backends do X" even though we think there's only one (with an asterisk) directory backend so far

rjbs (approved):
As discussed, this is entirely about the non-zeroskip-itself as a review.  Also, the automake change at the end was a bit beyond an easy read for me, so I have not read it carefully, but will do so on request.

lib/cyrusdb.c
brong: Quotalegacy is a deep dark hole of horror and suffering.  It doesn't take a configurable path I don't think, it's hard coded into the tree.  But yes, it's a directory structure.

brong: yeah, it'll never match but I updated the commits to be more explicit.

lib/cyrusdb.c
brong: Yeah, text changed even though this is the only one so far which can do this.

PR #6220, opened 2026-08-10
Correctly handle duplicate modseqs in Email/changes and Mailbox/changes

rsto:
In https://github.com/cyrusimap/cyrus-imapd/pull/6167 we made Cyrus gracefully handle duplicate modseqs on the assumption that these only occur for corrupted mailbox state. We also made it log errors in case it encounters duplicate modseqs. As it turns out, duplicate modseqs are a regular thing in Cyrus, both because foldermodseqs bump when a message in a thread gets a new modseq, and because the Email/set bulk update code can result in two messages in separate mailboxes getting the same modseq.

Whether we do want these duplicate modseqs to occur is a secondary question. As of now, they are a given and we need to better handle them than overshooting the `maxChanges` window in a `/changes` method based on the assumption that this only will happen in exceptional cases.

This PR now makes Cyrus correctly handle `maxChanges`: it caps the changes response and returns `cannotCalculateChanges` if it can't report a single change due to the limited `maxChanges` window.

This PR is based on the findings of https://github.com/cyrusimap/cyrus-imapd/pull/6216 for which this PR is a replacement.

### DIFF
--- a/cassandane/tiny-tests/JMAPMailbox/mailbox_changes_maxchanges_duplicate_modseqs
+++ b/cassandane/tiny-tests/JMAPMailbox/mailbox_changes_maxchanges_duplicate_modseqs
@@ -1,0 +1,77 @@
+#!perl
+use Cassandane::Tiny;
+
+# This test asserts that Mailbox/changes either reports all Mailboxes having
+# the same modseq in a response, or none of them.  It must not exceed
+# maxChanges to do so -- RFC 8620 §5.2 says "MUST NOT".
+sub test_mailbox_changes_maxchanges_duplicate_modseqs
+    ($self)
+{
+    my $user = $self->default_user;
+    my $jmap = $user->jmap;
+
+    xlog $self, "Set up three mailboxes with one email each";
+    my $mbox_a = $user->mailboxes->create({ name => 'A' });
+    my $mbox_b = $user->mailboxes->create({ name => 'B' });
+    my $mbox_c = $user->mailboxes->create({ name => 'C' });
+
+    my $email_a = $user->emails->create({
+        mailboxIds => { $mbox_a->id => JSON::true },
+        subject    => 'msgA',
+    });
+    my $email_b = $user->emails->create({
+        mailboxIds => { $mbox_b->id => JSON::true },
+        subject    => 'msgB',
+    });
+    my $email_c = $user->emails->create({
+        mailboxIds => { $mbox_c->id => JSON::true },
+        subject    => 'msgC',
+    });
+
+    my $res = $jmap->request([[ 'Mailbox/get', {} ]]);
+    my $state = $res->single_sentence('Mailbox/get')->arguments->{state};
+    $self->assert_not_null($state);
+
+    xlog $self, "Set flag on email in mailbox C";
+    $email_c->update({ 'keywords/$seen' => JSON::true });
+
+    xlog $self, "Set flag on emails in mailboxes A and B simultaneously";
+    $res = $jmap->request([
+        [ 'Email/set', {
+            update => {
+                $email_a->id => { 'keywords/$seen' => JSON::true },
+                $email_b->id => { 'keywords/$seen' => JSON::true },
+            },
+        } ],
+    ]);
+    $res->assert_single_successful_set('Email/set');
+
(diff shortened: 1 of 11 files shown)


cassandane/tiny-tests/JMAPMailbox/mailbox_changes_maxchanges_duplicate_modseqs
```diff
@@ -0,0 +1,82 @@
...
+    $self->assert_not_null($state);
+
+    xlog $self, "Set flag on email in mailbox C";
+    $jmap->request([
+        [ 'Email/set', {
+            update => {
+                $email_c->id => { 'keywords/$seen' => JSON::true },
+            },
+        } ],
+    ]);
```
wolfsage: Here, below, and in other tests we aren't asserting that our Email/set update was good - we should be, otherwise failures here could appear as failures later in the test, confusing debugging.

This can be solved with:
```
$res = $jmap->request([ ... ]);
$res->assert_single_successful_set;
```
or more explicitly:
```
$res->assert_single_successful_set("Email/set");
```

cassandane/tiny-tests/JMAPMailbox/mailbox_changes_maxchanges_duplicate_thread_modseqs
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+    $self->assert_str_equals($email1->threadId, $email2->threadId);
+
+    my $res = $jmap->request([[ 'Mailbox/get', {} ]]);
+    my $state = $res->single_sentence('Mailbox/get')->arguments->{state};
+    $self->assert_not_null($state);
+
+    xlog $self, "Set flag on the reply, which is in mailbox A only";
+    $jmap->request([
```
wolfsage: In cassandane/tiny-tests/JMAPMailbox/mailbox_changes_maxchanges_duplicate_modseqs, you make two different Email/set calls:
 * mark 1 message read
 * mark 2 messages read

So it makes sense to use `$jmap->request()` as it keeps similar things in that test that are somewhat related looking the same.

Here, we really only care that you're marking a message read. That means you could replace this:
```
$jmap->request([
        [ 'Email/set', {
            update => {
                $email2->id => { 'keywords/$seen' => JSON::true },
            },
        } ],
    ]);
```
With:
```
$email2->update({ 'keywords/$seen' => JSON::true });
```

and automatically get detection if the Email/set update fails.

imap/jmap_mail.c
```diff
@@ -5506,6 +5506,34 @@ static int jmap_email_querychanges(jmap_req_t *req)
...
+        = { changes->created, changes->updated, changes->destroyed };
+
+    for (size_t i = 0; i < 3; i++) {
+        while (json_array_size(arrays[i]) > mark[i]) {
+            json_array_remove(arrays[i], json_array_size(arrays[i]) - 1);
+        }
+    }
+}
+
+static void _changes_set_mark(const struct jmap_changes *changes,
```
wolfsage: These 3 methods are now duplicated in two files (jmap_mailbox.c and jmap_mail.c), whereas the jmap_changes struct is defined in jmap_api.h 

Should we move these three methods there so they are only defined once?

wolfsage (changes requested):
Seems sensible, though I think the tests could use some sanity checks and maybe we could deduplicate the code changes

rjbs (commented):
I agree with the things Matthew said: looks good, could reduce some duplication, but it'll be great to ship!

cassandane/tiny-tests/JMAPMailbox/mailbox_changes_maxchanges_duplicate_modseqs
rsto: TIL

cassandane/tiny-tests/JMAPMailbox/mailbox_changes_maxchanges_duplicate_thread_modseqs
rsto: TIL

rsto:
Rewrote the tests. I initially had decided not to deduplicate the code, but I was torn about that. Now deduplicated by popular demand.

wolfsage (approved):

PR #6222, opened 2026-08-11
Twom hardening port

brong:
This has a bunch of Claude reviews of twom, both an original hardening pass, then a rather embarrassing comparison against the zeroskip v2 it had just written, where my code had a much messier and inconsistent naming, a bug with case insignificant key lookups, some edge case inconsistencies I don't think we can hit in practice but good to fix.

So the churn looks really big, but the actual change in behavior is very small.  There's a changes file which explains the key changes.


ksmurchison (approved):
This all looks sane.

I really wish that Claude would've left the `//` comments alone instead of converting them to `/* */` comments.

That said, this probably needs at least one more set of eyes

brong:
> This all looks sane.
> 
> I really wish that Claude would've left the `//` comments alone instead of converting them to `/* */` comments.
> 
> That said, this probably needs at least one more set of eyes

Yeah I asked it to do that for consistency, probably unnecessary but with so much else also being changed I figured why not.  Easy to flip back.

brong:
@rjbs adding you as well.  This is a bunch of stuff from the twom repo, now the vendored module is 100% vendored, including generating uuids itself rather than using libuuid.  So this is pure open source twom[tm], copied in.

PR #6223, opened 2026-08-11
imapd.c: handle ID command with NIL "os" value

ksmurchison:
Previously, this would segfault in strcmp() with a NULL string

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -3490,12 +3490,12 @@ static void cmd_id(char *tag)
                 return;
             }
 
-            if (!strcmp(field.s, "os") && !strcmp(arg.s, "iOS")) {
+            if (!strcmp(field.s, "os") && !strcmpnull(arg.s, "iOS")) {
                 imapd_id.quirks |= QUIRK_SEARCHFUZZY;
             }
 
             /* ok, we're happy enough */
-            hash_insert(field.s, xstrdup(buf_cstring(&arg)), &imapd_id.params);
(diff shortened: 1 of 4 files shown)


imap/imapd.c
```diff
@@ -3490,12 +3490,14 @@ static void cmd_id(char *tag)
...
 
-            if (!strcmp(field.s, "os") && !strcmp(arg.s, "iOS")) {
+            if (!strcmp(field.s, "os") && !strcmpnull(arg.s, "iOS")) {
                 imapd_id.quirks |= QUIRK_SEARCHFUZZY;
             }
 
             /* ok, we're happy enough */
-            hash_insert(field.s, xstrdup(buf_cstring(&arg)), &imapd_id.params);
+            hash_insert(field.s,
+                        buf_base(&arg) ? xstrdup(buf_cstring(&arg)) : NULL,
```
rsto: this line can be shortened as `buf_releasenull(&arg)`

I take it that we can guarantee at this point that not entry in the hash table had that same key? Otherwise you'd need to `free` the result of `hash_insert`, too.

rsto (approved):
LGTM - just found one code line that could be expressed more succinctly.

imap/imapd.c
ksmurchison: Yeah, I previously has xstrdupnull(buf_cstringnull(&arg));

I returned to doing that, and also handle the case of the same key appearing more than once.

rsto (approved):

PR #6224, opened 2026-08-11
cassandane: report JMAP::Tester aborts as failures, not errors

rjbs:
A JMAP::Tester::Abort is a Throwable::Error.  Test::Unit only knows about Error.pm exceptions, and of those only a Test::Unit::Failure counts as a failing test, so a JMAP::Tester abort looked like ERROR when it should've looked like FAIL.  Do we care?  Well, maybe not all *that* much, but accuracy matters.

More importantly, JMAP::Tester aborts carry really nice diagnostics!  I want those.

The method here is gross as heck: we just replace the JMAP::Tester::Abort constructor, because otherwise there's sort of an air gap between some (but not all) of the JMAP::Tester classes and any central hub to change this code.

Also: I looked at making a multiply inheriting class (boo, hiss), but the big problem was going to be that the error has to be Storable, too. Just nevermind.

As for the .dataprinter file: it's mine!  We use it elsewhere, though, so it should be familiar to some.

### DIFF
--- a/cassandane/Cassandane/Role/JMAPTester.pm
+++ b/cassandane/Cassandane/Role/JMAPTester.pm
@@ -13,7 +13,7 @@ use JMAP::Tester::Abort ();
 # We want to throw an Error instead -- or, well, we don't want to, but we sort
 # of have to.  This solution is absolutely gross: just replace the
 # JMAP::Tester::Abort constructor.
-unless (JMAP::Tester->can('abort_class')) {
+{
     no warnings 'redefine';
     *JMAP::Tester::Abort::new = sub ($, @rest) {
         return Cassandane::JMAPAbort->new(@rest);
--- a/cassandane/Cassandane/Role/JMAPTester.pm
+++ b/cassandane/Cassandane/Role/JMAPTester.pm
@@ -3,12 +3,37 @@ use Moo::Role;
 
 use experimental 'signatures';
 
+use Cassandane::JMAPAbort ();
 use Encode ();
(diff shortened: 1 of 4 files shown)


cassandane/Cassandane/Role/JMAPTester.pm
```diff
@@ -3,13 +3,38 @@ use Moo::Role;
...
 use Encode ();
+use JMAP::Tester ();
+use JMAP::Tester::Abort ();
+
+# A JMAP::Tester::Abort is a Throwable::Error, which Test::Unit reports as an
+# error -- as though the suite were broken -- instead of as a failing test.
+# We want to throw an Error instead -- or, well, we don't want to, but we sort
+# of have to.  This solution is absolutely gross: just replace the
+# JMAP::Tester::Abort constructor.
+unless (JMAP::Tester->can('abort_class')) {
```
wolfsage: I don't understand this can() check - where is abort_class coming from and why wouldn't we want to override if it exists?

wolfsage (approved):
Approved but with a question

cassandane/Cassandane/Role/JMAPTester.pm
rjbs: Ha, sorry, that's from my first pass at doing this, which I didn't clear up when I decided to be gross.  Must've survived as copypasta.  The story of this branch was "inspiration, desperation, cut-n-pastin', decoration."

Fixed.

PR #6225, opened 2026-08-12
clarify detail about slow tests under CI

elliefm:
This PR has two commits, but only one of them is real.  CI failing means everything is okay.

The first commit introduces a slow test, `ACL.bogus_slow`, which intentionally fails.  The fact that it fails in CI means that CI can detect and report failures in slow tests.  Whew!  **I will discard this commit, and wait for a clean CI run, before merging.**

The second commit is the real commit.  It's a trivial, zero-impact change to the workflow file, removing a token that was misleading.  Details are in the commit message.


rjbs (approved):
Approved, but… y'know, remove that failing test. ;-)

PR #6226, opened 2026-08-12
limit recursion in imap search

rjbs:


rjbs:
I accidentally jumped the gun on the merge, but murch looked afterward and okayed this.

PR #6227, opened 2026-08-13
Mboxevent specialuse & JMAPId params and other fixups

robmueller:
This ended up with a few more commits that make an mboxevent more consistent and give it imap and jmap versions of all ids:
* for a `vnd.cmu.MessageCopy` event, add the `vnd.fastmail.specialUse` and `vnd.fastmail.oldSpecialUse` fields which is a list of the special use flags for the mailboxes the email is being copied to/from. Gated by `event_extra_params`
* add `vnd.fastmail.emailIds`, a cousin of `vnd.fastmail.emailid`, a json array of JMAP email ids of the messages referenced in the event. There is no "old" variant because JMAP email ids are stable. Gated by `event_extra_params`
* add `vnd.fastmail.mailboxJMAPId` and `vnd.fastmail.oldMailboxJMAPId`, JMAP id cousins of `mailboxID` and `oldMailboxID`. Gated by `event_extra_params`
* make `mailboxID` a URI per RFC 5423. At some point this set to the mailbox's JMAP id or uniqueid. Make this spec compliant (will require a fixup on the Fastmail side)
* honor the mbtype/mailboxACL/visibleUsers extra params. These were being added unconditionally, even though there were flags in `event_extra_params` for them (will require a fixup on the Fastmail side)
* add oldMailboxUniqueId event param, cousin to `mailboxUniqueId`


rjbs:
@robmueller If you point your agent at docsrc/developer/overview.rst it will be able to run the tests and solve these crashes/failures before someone here looks into this.  Give that a go, especially if you still have the session?

robmueller:
Ok, pushed an updated version.

rjbs:
Have assigned this to myself for review, hopefully Monday!

imap/mboxevent.c
```diff
@@ -258,14 +265,42 @@ EXPORTED void mboxevent_setnamespace(struct namespace *n)
...
+
+    return buf_release(&attrib);
+}
+
+/* Return the JMAP id of the mailbox: the short JMAP id when the user's
+ * conversations db uses compact ids, otherwise the mailbox's uniqueid */
+static const char *mboxevent_mailbox_jmapid(const struct mailbox *mailbox)
+{
+    struct conversations_state *cstate =
+        conversations_get_mbox(mailbox_name(mailbox));
```
rjbs: I think is wrong.  (I find these APIs a bit confusing and they aren't documented.)

I think conversations_get_mbox ends up delegating to `conversations_get_path` which can only return an already-open cstate.  Is it possible to hit this code with conversationdb not already open?  I'm not sure, but *also* I suspect we don't need to rely on that, and could use `mailbox_get_cstate` and it'd open if needed and reuse if open.   (If it can be closed, we could end up with the jmapid of the mailbox flipping between compact and not-compact on different events.)

I'll give it a quick look, but I _think_ just changing the API used here solves this.

rjbs:
We should add a changes file given that this changes mailboxID and makes mbtype, mailboxACL, visibleUsers all opt in.

I'm going to look into the cstate open-or-not thing, and ask murch for another pair of eyes, as he knows this code better, but I think this looks good otherwise and we can probably get this in the next build.

rjbs:
@ksmurchison I think this is looking good, but another pair of eyes would be good, especially because I added my own commit to the end, because the original branch was missing some cases.

rjbs:
Also worth noting:  MailboxSubscribe and MailboxUnSubscribe events don't get the mailboxJMAPId.  That's because they don't have a mailbox (sometimes they can't have one), so they'd need a name lock to get a cstate, which is more than I wanted to do.  Shouldn't affect our needs.  To be fixed in events 2.0 ;)

rjbs:
My changes are a problem, too.  We can't get the convdb without a name lock.  Even doing that we don't really end up consistent.  This is a bunch of change, most of which I think we don't need for the feature we discussed.  I've emailed @robmueller to ask if we can just focus on the bare necessity.  I'd love to rip out mboxevents and replace them with something better that can be sensible from day one.

PR #6228, opened 2026-08-13
Acl based subscription

brong:
This is the missing part to allow us to switch from individual ACLs on the Shared address book to a single customer-wide ACL without having to still do the per-user accounting for subscriptions.

### DIFF
--- a/cassandane/tiny-tests/JMAPContacts/addressbook_get_autosub
+++ b/cassandane/tiny-tests/JMAPContacts/addressbook_get_autosub
@@ -1,0 +1,52 @@
+#!perl
+use Cassandane::Tiny;
+
+# The '1' (ACL_AUTOSUB) right on a shared addressbook makes AddressBook/get
+# report isSubscribed even though the sharee never subscribed.
+sub assert_addressbook_get_autosub ($self, $grantee)
+{
+    my $jmap  = $self->default_user->jmap;
+    my $admin = $self->{adminstore}->get_client();
+
+    my $other = $self->{instance}->create_user('other');
+    my $abook = $other->addressbooks->default;
+
+    for my $case ([ lrs => jfalse() ], [ lrs1 => jtrue() ]) {
+        my ($acl, $want) = @$case;
+
+        xlog $self, "grant '$acl' to $grantee";
+        $admin->setacl('user.other.#addressbooks.Default', $grantee => $acl)
+            or die "setacl: $@";
+
+        my $res = $jmap->request([
+            [ 'AddressBook/get', {
+                accountId  => 'other',
+                ids        => [ $abook->id ],
+                properties => [ 'isSubscribed' ],
+            } ],
+        ]);
+
+        my $books = $res->single_sentence('AddressBook/get')->arguments->{list};
+
+        $self->assert_cmp_deeply(
+            [ superhashof({ id => $abook->id, isSubscribed => $want }) ],
+            $books,
+            "isSubscribed for ACL '$acl'",
+        );
+    }
+}
+
+sub test_addressbook_get_autosub_user ($self)
+{
+    $self->assert_addressbook_get_autosub('cassandane');
+}
+
+sub test_addressbook_get_autosub_group
+    :Mboxgroups
+    ($self)
+{
+    $self->{adminstore}->get_client
+         ->_imap_cmd('SETUSERGROUP', 0, '', 'cassandane', 'group:c1-all');
+
+    $self->assert_addressbook_get_autosub('group:c1-all');
+}
--- a/cassandane/tiny-tests/JMAPContacts/addressbook_set_autosub
+++ b/cassandane/tiny-tests/JMAPContacts/addressbook_set_autosub
@@ -1,0 +1,81 @@
+#!perl
+use Cassandane::Tiny;
+
+# Subscribing and unsubscribing an autosub addressbook both succeed, but the
+# '1' right keeps isSubscribed true until the right itself goes away.
+sub test_addressbook_set_autosub ($self)
+{
+    my $jmap  = $self->default_user->jmap;
+    my $admin = $self->{adminstore}->get_client();
+
+    my $other = $self->{instance}->create_user('other');
+    my $abook = $other->addressbooks->default;
+
+    $admin->setacl('user.other.#addressbooks.Default', cassandane => 'lrs1')
+        or die "setacl: $@";
+
+    xlog $self, "subscribe for real, then unsubscribe: both must succeed";
+    my $res = $jmap->request([
+        [ 'AddressBook/set', {
+            accountId => 'other',
+            update => { $abook->id => { isSubscribed => JSON::true } },
+        }, 'subscribe' ],
+        [ 'AddressBook/set', {
+            accountId => 'other',
+            update => { $abook->id => { isSubscribed => JSON::false } },
+        }, 'unsubscribe' ],
+        [ 'AddressBook/get', {
+            accountId  => 'other',
+            ids        => [ $abook->id ],
+            properties => [ 'isSubscribed' ],
+        }, 'get' ],
+    ]);
+
+    my $subscribe = $res->paragraph_by_client_id('subscribe')
+                        ->single('AddressBook/set')->arguments;
+    my $unsubscribe = $res->paragraph_by_client_id('unsubscribe')
+                          ->single('AddressBook/set')->arguments;
+    my $books = $res->paragraph_by_client_id('get')
+                    ->single('AddressBook/get')->arguments->{list};
+
+    $self->assert_cmp_deeply(
+        superhashof({ updated => { $abook->id => ignore() } }),
+        $subscribe,
+        "subscribing an autosub addressbook succeeds",
+    );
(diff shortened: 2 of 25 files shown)


imap/imapd.c
```diff
@@ -13647,6 +13656,46 @@ static int subscribed_cb(struct findall_data *data, void *rockp)
...
+ * already emitted from the real subscription list */
+static int autosub_cb(struct findall_data *data, void *rockp)
+{
+    struct autosub_rock *arock = (struct autosub_rock *) rockp;
+
+    if (!data) {
+        perform_output(NULL, NULL, arock->lrock);
+        return 0;
+    }
+    if (!data->is_exactmatch) return 0;
```
rsto: This early return, and the same one in recursivematch_autosub_cb (line 13874), make
ACL-based subscriptions differ from real ones: a real subscription marks the parent, an
ACL-based one doesn't. Affects LSUB as well as RECURSIVEMATCH — list_response (13257) uses
CHILDINFO_SUBSCRIBED to decide whether to mention a non-subscribed parent at all, and
subscribed_cb sets it (13653) where autosub_cb never does. RETURN (SUBSCRIBED) is
unaffected; 13281 strips the bit outside RECURSIVEMATCH.

note: dropping the early return won't fix it. recursivematch_cb runs off findsubmulti,
where a non-exactmatch entry means "ancestor of something subscribed"; these two run off
findallmulti, where it's just hierarchy traversal. Ancestors have to be derived by walking
up from the mailboxes that passed the ACL check.

imap/caldav_alarm.c
```diff
@@ -729,7 +729,7 @@ static int has_peruser_alarms_cb(const char *mailbox,
 
     if (!mboxname_userownsmailbox(userid, mailbox) &&
         ((hrock->mbox_options & OPT_IMAP_SHAREDSEEN) ||
-         mboxlist_checksub(mailbox, userid) != 0)) {
+         !mboxlist_issubscribed(mailbox, userid, NULL))) {
```
rsto: Passing NULL makes mboxlist_issubscribed build an auth state per per-user annotation
record, though for a given sharee it can't change. Inverts the old cost: mboxlist_checksub
made the unsubscribed-sharee case the cheapest, now it's the one that falls through to
auth_newstate.

In a test setup I measured ~2 constructions per event at store time and ~3 per event per
calalarmd run, single sharee.

Most of that is store time, not alarm time — has_peruser_alarms_cb runs from has_alarms()
on every record store, so every write to a shared calendar pays it in httpd.

Can't just pass an auth state in: annotatemore_findall_mailbox (967, 1373) iterates one
record's per-user annotations, so userid differs per invocation and isn't known until the
callback. Either cache by userid (hrock/prock are per-record, so that needs lifting to the
enclosing loop to help across records), or skip the check entirely when the mailbox ACL has
no ACE granting 1.

rsto (commented):
I wasn't on the original reviewer list, but when I looked at the Cassandane tests I got triggered on the `min_version_3_1` annotation and went down the rabbit hole.

I created https://github.com/brong/cyrus-imapd/pull/1 which contains rewrites of the Cassandane tests to use our new tooling, including JMAPTester, `Test::Deep` and all the other new goodness. This is not meant only to make the tests more readable but also to not increase the amount of old-style tests which the agents just seem to love copying over and over. That being said, rjbs and wolfsage definitely know best how these test could get reshaped.

Other than that, I think this PR contains one functional defect and one performance regression and I let Claude validate it by instrumenting the code. This looks real to me. I normally wouldn't copy-paste the agent output in my review feedback, but I know you are open to that.

**Edit**: I forgot to mention that the PR description only talks about addressbooks, but I noticed it also changes the calendars code. My PR with the suggested test changes also includes a test for JMAP Calendars.

cassandane/tiny-tests/JMAPContacts/addressbook_get_autosub
```diff
@@ -0,0 +1,52 @@
...
+    xlog $self, "create default addressbook";
+    my $othertalk = Net::CardDAVTalk->new(
+        user => "other",
+        password => 'pass',
+        host => $service->host(),
+        port => $service->port(),
+        scheme => 'http',
+        url => '/',
+        expandurl => 1,
+    );
```
rjbs: For the above, you can:

```perl
my $other = $self->{instance}->create_user('other');
my $othertalk = $other->carddav;
```

This adds up in noise reduction.

cassandane/tiny-tests/JMAPContacts/addressbook_set_autosub
```diff
@@ -0,0 +1,69 @@
...
+            properties => ['isSubscribed'],
+        }, 'R3'],
+    ]);
+    $self->assert(exists $res->[0][1]{updated}{$id});
+    $self->assert(exists $res->[1][1]{updated}{$id});
+
+    xlog $self, "server reports the still-true effective state";
+    $self->assert_equals(JSON::true,
+        $res->[1][1]{updated}{$id}{isSubscribed});
+    $self->assert_equals(JSON::true, $res->[2][1]{list}[0]{isSubscribed});
```
rjbs: I get it, but this is pretty weird.  JMAP for Calendars wants "you can't unsubscribe" to get forbidden:

> Users can subscribe or unsubscribe to a calendar by setting the "isSubscribed" property. The server MAY forbid users from subscribing to certain calendars even though they have permission to see them, rejecting the update with a forbidden SetError.

...but that's a slightly different situation.  For simplicity, I might stick with what it says.  ACL1 is "autosubscribed without the option to drop out".  I haven't carefully thought through both options FWIW.

cassandane/tiny-tests/Mboxgroups/list_autosub
```diff
@@ -0,0 +1,51 @@
...
+sub test_list_autosub
+    :NoAltNamespace
+    ($self)
+{
+    my $admintalk = $self->{adminstore}->get_client();
+    my $imaptalk = $self->{store}->get_client();
+
+    $admintalk->create("user.otheruser.autoshared");
+    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());
+    $admintalk->setacl("user.otheruser.autoshared", "group:group co", "lrs1");
```
rjbs: woah, we allow spaces in group identifiers?  weirded out.

cassandane/tiny-tests/Mboxgroups/list_autosub
```diff
@@ -72,4 +72,22 @@ sub test_list_autosub
     %reccount = ();
     $reccount{$_}++ for @recnames;
     $self->assert_num_equals(1, $reccount{'user.otheruser.autoshared'});
+
+    xlog $self, "granting '1' on a DAV mailbox doesn't surface it as subscribed";
```
rjbs: Heh!

docsrc/reference/admin/access-control/rights-reference.rst
```diff
@@ -407,3 +407,13 @@ Individual Rights Reference
...
+
+    Granting this right to the ``anyone`` identifier auto-subscribes
+    every user with access to the mailbox. That's coherent, but almost
+    never what's wanted; grant it to a ``group:`` identifier instead.
+
+    IMAP ``NOTIFY (SUBSCRIBED)`` and the subscribed-scope ``ESEARCH``
+    multisearch consult only real subscription entries, not this
+    right. This only matters when the right is granted on a regular
+    mail folder: neither feature applies to calendars, address books
+    or the other collection types this right can also be granted on.
```
rjbs: I'm surprised, how come?  I mean, using NOTIFY is rare, and managing SUBSCRIBED is pretty rare, too, but I'm surprised these don't treat a autosub as a real sub.  Feels like if anybody did want to use them, it would be confusing.  "Didn't get results!  Am I subscribed?  Yes."

rjbs (approved):
I think you should definitely use the more concise form of user creation and DAV client construction, if possible.  I had a question or two about "Is this the best choice?" but generally I think this will DTRT and saw no major problem.

ksmurchison (approved):

cassandane/tiny-tests/JMAPContacts/addressbook_set_autosub
brong: Yeah, if you don't want that behavior, don't use it!  ACL1 is the "you can't remove this" option.

cassandane/tiny-tests/Mboxgroups/list_autosub
brong: Yeah, pretty sure it just works[tm]

docsrc/reference/admin/access-control/rights-reference.rst
brong: I got all that fixed up.

imap/imapd.c
brong: Thanks!  I changed it all around so that '1' behaves the same in all cases (so long as you have RACLs on of course)

imap/caldav_alarm.c
brong: This was all fixed up in the later commits, just replying to close out the thread before merging.

PR #6229, opened 2026-08-14
JMAP for Calendars: always use JSCalendar 2.0

rsto:
This is my copy of https://github.com/cyrusimap/cyrus-imapd/pull/6191 with additional commits that fully remove the `jmap_ical.c` file.

It removes the old-style timezone cache for timezone guessing which only the jscalendar 1.0 code used. It also removes the code that converts CalDAV managed attachments to blobIds in the JSCalendar event, which never was live and most likely had its warts anyway.


rsto:
Rebased to master, rerunning CI now.

PR #6231, opened 2026-08-14
lmtpengine.c and more: add SMTPUTF8 support

arnt:
This also adds tests and makes them pass. Most notably, a test that a message can arrive via LMTP and then matches a UTF8 UID SEARCH, which required making conversations.db store UTF8 always.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -663,6 +663,7 @@ cunit_TESTS = \
     cunit/seqset.testc \
     cunit/sessionid.testc \
     cunit/slowio.testc \
+    cunit/smtpclient.testc \
     cunit/smallarrayu64.testc \
     cunit/xsyslog_ev.testc
 
--- a/cassandane/Cassandane/Net/SMTPServer.pm
+++ b/cassandane/Cassandane/Net/SMTPServer.pm
@@ -72,7 +72,8 @@ sub helo {
     $Self->mylog("SMTP: HELO $Host");
     return if $Self->override('helo');
     $Self->send_client_resp(250, "localhost",
-                            "AUTH", "DSN", "SIZE 10000", "ENHANCEDSTATUSCODES");
+                            "AUTH", "DSN", "SIZE 10000", "ENHANCEDSTATUSCODES",
+                            "SMTPUTF8");
 }
 
 sub mail_from {
--- a/cassandane/tiny-tests/Search/from_alabel
+++ b/cassandane/tiny-tests/Search/from_alabel
@@ -1,0 +1,62 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_from_alabel
+    :want_deliver :min_version_3_13
+    ($self)
+{
+    xlog $self, "test that the cache holds u-labels, not punycode";
+
+    # Punycode is for the DNS, so a message from xn--gr-zia.org is
+    # cached, reported and searched as grå.org. Note that both
+    # messages are pure ASCII on the wire: the a-label is what the
+    # sender wrote, and the u-label is what we make of it.
+
+    my %msgs;
+    $msgs{1} = $self->{gen}->generate(
+        subject => "IDN sender",
+        from => Cassandane::Address->new(
+            name => "Arnt",
+            localpart => "arnt",
+            domain => "xn--gr-zia.org"));
+    $msgs{2} = $self->{gen}->generate(
+        subject => "Sender with a broken a-label",
+        from => Cassandane::Address->new(
+            name => "Info",
+            localpart => "info",
+            domain => "xn--zz.example.com"));
+
+    xlog $self, "deliver the messages over LMTP";
+    $self->{instance}->deliver($msgs{1});
+    $self->{instance}->deliver($msgs{2});
+
+    my $talk = $self->{store}->get_client();
+    $talk->select('INBOX');
+
+    xlog $self, "FETCH ENVELOPE reports the u-label";
+    my $res = $talk->fetch('1:2', 'envelope')
+        or die "Cannot fetch: $@";
+    $self->assert_str_equals('grå.org',
+                             $res->{1}->{envelope}->{'From-Raw'}->[0]->[3]);
+
+    xlog $self, "an a-label that isn't valid punycode is kept as it came";
+    $self->assert_str_equals('xn--zz.example.com',
+                             $res->{2}->{envelope}->{'From-Raw'}->[0]->[3]);
+
+    xlog $self, "SEARCH FROM finds the u-label";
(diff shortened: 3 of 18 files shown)


arnt:
This is mostly the code I sent to @ksmurchison by email about two years ago. I looked it again and the thing I couldn't find a way to do back then was… simple.

arnt:
FWIW, Cassandane is broken for me on the tip. Not sure what's the matter, but I made the PR anyway.

rsto:
@arnt Definitely interested in this patch, but currently the build fails with an error
```
imap/mboxname.c:1912:25: error: comparison is always false due to limited range of data type [-Werror=type-limits]
 1912 |                   *name >= 128 || *name < 0 ||
```
I suggest to first fix that, we can then see which cassandane tests fail.

arnt:
Oddly, it does not fail when I build it?

Anyway, the part that fails is checking for a byte being outside the ASDCII range, basically `x > 127 || x < 0`. If char is signed, the first is necessarily false, if it's unsigned, the second is. I didn't see anything in the configuration that forces either unsigned or signed, so I wrote the code to work on both varieties of char.

What did I miss? I don't see -funsigned-char or -fsigned-char anywhere... Do you have an explicit policy on the signedness of char?

rsto:
> Do you have an explicit policy on the signedness of char?

We do not, it's most likely just because you built this on ARM with clang and/or with different compiler flags. You could replace the check with 
```
(unsigned char)*name >= 128
```
and that should be it. The same issue is also is in imap/lmtpengine.c around line 457.

arnt:
I did it like that, and also in a third place.

But then I sidetracked into fixing smtpclient.c, which can be called from things like a sieve redirect… done now, must reread the code, may push tomorrow.

rsto:
> But then I sidetracked into fixing smtpclient.c, which can be called from things like a sieve redirect… done now, must reread the code, may push tomorrow.

I'm happy if you update it, but I might at some point before merging decide to tweak this code some more. So feel free to hand this off to us whenever you want, unless you prefer us going back and forth on this in possibly multiple rounds.

What I do not fully understand is the goal of this pull request. Your description suggests that it aims to fix UTF8 UID SEARCH. But is that all or is there something else where Cyrus currently is not valid according to the specification? (_edit: added a not_).

arnt:
I amended the commit to make it build even when the compiler is smart about the range of char. Then I added more code in a separate commit.

The implementation of 6855 was good, but at the same time incomplete. `UID SEARCH FROM "grå"` worked and did not work. Notably, if a message arrived with `From: info@xn--gr-zia.org`, the search would not find it. The first commit extends message injection and the ENVELOPE caching enough that `UID SEARCH` works independently of the syntax used. I think there was another tiny bug in the 6855 fix, too.

The second commit was when I noticed that single-server injection works, but if you have several backends it can still break. So I fixed that, and because of code reuse that also gave sieve the ability to redirect to UTF8 domains.

Both cassandane files are 100% hope. The cunit tests pass when I run them.

The purpose of this… ah, well… Ken asked me whether I could review the 6855 implementation two years ago, and I did so, and fixed some tiny bugs but didn't manage to fix the UID SEARCH thing, then I ran out of time. Last week I started going through my various unfinished patches, and this one looked close enough to finish.

You might say that I'd like Cyrus to have a well-rounded functionality. If you disable the two variables so that UTF8=ACCEPT is enabled, then the UTF8=ACCEPT-related functionality should work, without known bugs. I don't particularly care whether you enable it by default, but I do care that when it's enabled, it works without known bugs. Does that make sense?

arnt:
There's a bug still; there should be an if around line 1160 of lmtpengine.c, so it only advertises SMTPUTF8 in the same cases as IMAP advertises UTF8=ACCEPT.

arnt:
I pushed a third commit with the fix to make LMTP and IMAP advertise support in tandem. What's there is now:
- Make UID SEARCH FROM and other address searches work equally whether punycode is used or not
- Advertise SMTPUTF8 via LMTP if and only if UTF8=ACCEPT is advertised via IMAP
- ... and accidentally, support these addresses in sieve redirections

arnt:
I see the Cassandane tests failed… I was nervous about those, actually. Didn't get Cassandane to run at all on my main development computer, and even if many of the lines are copied or modified from existing files, the total number of lines made me nervous. Could you (@rsto or any other team member) have a look at the Cassandane files and tell me what's wrong?

rsto:
> Last week I started going through my various unfinished patches, and this one looked close enough to finish.

I really appreciate the contribution, but I would prefer to not merge this. The blast radius of this change is higher than I want it to, the `lib/parseaddr.c` changes and the ones `auth_mboxgroups.c` are especially risky. I also don't like the idea of changing our on-disk cache format, and if we do we have to bump the cache version.

I agree that we should change Cyrus to be more compliant with RFC 6531 and RFC 9755, but I prefer we come up with a change ourselves. I want the correctness changes to happen on the edge, not in our crore data structures and code, because I can't and won't expect you to know every dark corner of the Cyrus codebase. I don't want to tie you and me up with a long back and forth until we got this PR into shape.

So, I'll take note that we should more strict with UTF8 support in LMTP and our SMTP client. We might also update IMAP SEARCH, but we already support A-label and U-label forms in FUZZY SEARCH. We'll also need to change our JMAP EmailSubmission implementation which hardcodes SMTP capabilities without SMTPUTF8 support, have to figure out what to do about the `lmtp_strict_rfc2821` config option and likewise make sure to either get rid of the `reject8bit/munge8bit` options or make sure the code does the right thing in all configurations.

arnt:
Fine, your call. If you want I'll separate out the LMTP, so that UTF8=ACCEPT and SMTPUTF8 are advertised together, and submit that as a PR with minimal blast radius. In that case I'll leave out Cassandane, it's too tricky for me.

Two tangential notes, though:

- This doesn't actually change the on-disk format. It's better described as normalisation: when email addresses can have several equivalent forms, it converts them to one form. But all forms already existed, and Cyrus will store them today if you `append`.

- This is a fairly common problem with RFC 6532 or perhaps 349x. A lot of code expects addresses to be comparable using `strcasecmp`, and once you say that `info@grå.org` is equal to `info@xn--gr-zia.org` that expectation breaks. In my opinion the best solution is to normalise early and conform to the expectations. (John Klensin once told me that allowing `…@xn--…` in 6532 was just an oversight.)

rsto:
>     * This doesn't actually change the on-disk format. It's better described as normalisation: [... ] In my opinion the best solution is to normalise early and conform to the expectations.

I agree on both points. It's just that our codebase currently churns a lot and we have (too) many parallel initiatives going on. I just don't feel confident right now doing that now.

arnt:
Fine. Let me know if/when you want to pick this up.

PR #6233, opened 2026-08-14
Merge cyrus-timezones into cyrus-imapd

rsto:
This merges https://github.com/cyrusimap/cyrus-timezones/ into cyrus-imapd. Cyrus now keeps a copy of the IANA Time Zone Database (tzdb). Using that tzdata, it builds the VTIMEZONE files for the zoneinfo_dir, which the TZDist service serves and which guesstz indexes. Without them, Cyrus falls back to the timezone data built into libical. Cyrus now also generates the guesstz database to allow for timezone guessing, which is a feature in JMAP for Calendars. The generated assets are installed along with the other Cyrus build artefacts.

Two things change for packagers: vzic is now a build-time dependency, where previously only the cyrus-timezones build needed it. And the compiled-in default for zoneinfo_dir is now the directory that the build installs its data into, where it previously was unset.

Installations still can choose to opt out, either by setting zoneinfo_dir, or configure --with-zoneinfo-dir=PATH, to whatever directory they want, or by not generating VTIMEZONEs at all, using configure --without-vzic.

This PR is shaped in multiple commits as follows:
- We import the tzdata from IANA unmodified
- We pull in the guesstz codebase from cyrus-timezones and add tools/update-tzdata.sh to allow updating the tzdata directory
- We generate and install the time zone data at build time, and index it with guesstz
- We fix a bug that was latent in ctl_zoneinfo but now became more prominent because we can't rely on vzic being configured as we expect it to.
- We update the build process to handle vzic generating VTIMEZONE in a different way than a cyruslibs vzic would have.

In addition, this PR removes the cyruslibs cyrus-timezones installation from the CI runner to demonstrate this all works for real. That last commit is marked DO NOT MERGE and must be dropped before this lands.


docsrc/reference/manpages/systemcommands/cyr_guesstz.rst
```diff
@@ -0,0 +1,107 @@
...
+
+.. option:: -p, --print
+
+    Print the database as JSON on standard output.
+
+.. option:: -r timerange, --timerange=timerange
+
+    The time range over which to expand observances, as two UTC date-times
+    separated by a solidus, e.g.
+    ``20000101T000000Z/20320101T000000Z``.  That is also the default.  The
```
wolfsage: I'm confused by "this is also the default" and "The range is half-open: the start is included, the end is not"

If "This is also the default" is referring to "20000101T000000Z/20320101T000000Z", that doesn't look half-open as it defines a start and an end

wolfsage: Hm, or does this mean "the range is inclusive of the start, but not the end"?

Also, 2032 is right around the corner, will this bite us in 6 years and can we push this out further or is this due to limits in time handling?

wolfsage (approved):
I realize my two comments are probably about existing code being ported from one git repo to another, so feel free to ignore unless you think they warrant more specific attention.

docsrc/reference/manpages/systemcommands/cyr_guesstz.rst
rsto: Thanks. And sorry, that was poorly worded. I also agree that we should extend the default expansion time range, I now set it to expand until January 1, 2064. On my local build, that takes 17ms longer to build and adds ~145KB size of guesstz.db, resulting in a total of ~350KB.

I rephrased the description of that flag as follow

    The time range over which to expand observances, as two UTC date-times
    separated by a solidus, each formatted as defined by "FORM #2" in
    RFC 5545, Section 3.3.5.  The start time is part of the range and the
    end time is not, so an observance onset at exactly the end time is not
    expanded.

    The default range is ``20000101T000000Z/20640101T000000Z``: from the
    zero hour of January 1, 2000 up to, but not including, the zero hour of
    January 1, 2064.

rsto:
I now also adapted the copyright headers to new-style, had imported with old-style copyright lines.

PR #6234, opened 2026-08-14
index.c: bound the thread node count and references per message

rjbs:
The thread node array is sized from the message count plus the total number of References across those messages, with no bound on either. Compute that in uint64_t and give the threaders a bool return so THREAD can fail cleanly rather than proceed with a count that does not fit, and load at most 20 references per message: the root, the parent, and the 18 most-recent predecessors of the parent, following the advice already given to clients about paring down References.

### DIFF
--- a/cassandane/tiny-tests/Thread/references_chain_long
+++ b/cassandane/tiny-tests/Thread/references_chain_long
@@ -1,0 +1,37 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_references_chain_long ($self)
+{
+    xlog $self, "test THREAD with a linear chain of more than 20 inter-message references";
+    my $talk = $self->{store}->get_client();
+
+    xlog $self, "append some messages";
+    my $sub = 'Hello World!';
+    my %exp;
+    my @refs;
+    my @expthreads;
+    my $N = 25;
+    for (1..$N)
+    {
+        my $msg;
+        if (scalar @refs)
+        {
+            $msg = $self->make_message("Re: " . $sub,
+                                       references => [ @refs ]);
+        }
+        else
+        {
+            $msg = $self->make_message($sub);
+        }
+        push(@refs, $msg->{'headers_by_name'}{'message-id'}[0]);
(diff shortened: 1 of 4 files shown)


rjbs:
clang is complaining because we're doing `UINT64_MAX / 1.5`, which casts UINT64_MAX into a double, so it rounds up.

I think we can…

```diff
-    if (nmsg >= (uint64_t) (UINT64_MAX / 1.5)) {
-        ret = false;
-        goto done;
-    }
-
-    nnode = (uint64_t) (1.5 * nmsg + 1);
+    nnode = (uint64_t) nmsg + nmsg / 2 + 1;
```

nmsg was unsigned int, so maxing out around 6e9?  The old assertion was "if nmsg is greater than 2/3 uint64_max" but it can't ever be, because the type is too small to hold such a value.

If the guard was restated like this, even gcc complained:

```
if (nmsg > (UINT64_MAX - 1) / 3 * 2) {
```

ksmurchison:
> clang is complaining because we're doing `UINT64_MAX / 1.5`, which casts UINT64_MAX into a double, so it rounds up.
> 
> I think we can…
> 
> ```diff
> -    if (nmsg >= (uint64_t) (UINT64_MAX / 1.5)) {
> -        ret = false;
> -        goto done;
> -    }
> -
> -    nnode = (uint64_t) (1.5 * nmsg + 1);
> +    nnode = (uint64_t) nmsg + nmsg / 2 + 1;
> ```
> 
> nmsg was unsigned int, so maxing out around 6e9? The old assertion was "if nmsg is greater than 2/3 uint64_max" but it can't ever be, because the type is too small to hold such a value.
> 
> If the guard was restated like this, even gcc complained:
> 
> ```
> if (nmsg > (UINT64_MAX - 1) / 3 * 2) {
> ```

Yeah, this looks sane.  We really should change the `unsigned int` to `uint32_t` since that's the actual IMAP definition, and it would protect us if `unsigned int` ever becomes 64-bit.

PR #6242, opened 2026-08-14
metadata: require ACL_LOOKUP in apply_mailbox_array

rjbs:
apply_mailbox_array() handles the parenthesized-list arm of GETMETADATA and SETMETADATA, resolving each caller-supplied mailbox name with mboxlist_lookup() and handing the result to a per-mailbox proc.  The single-string arm checks ACL_LOOKUP first; this arm should too.


ksmurchison (approved):

rjbs:
This did merge, and I can't tell why it's still open.

PR #6250, opened 2026-08-18
Enforce JMAP CalendarEvent privacy for JMAP, CalDAV and iTIP

rsto:
This enforces the semantics of the CalendarEvent "privacy" property for calendar sharees as defined in JMAP Calendars for JMAP, CalDAV and iTIP.

A secret event must be handled as non-existent for sharees. A private event must only show a minimal subset of properties (mostly the temporal span of the calendar event). If a secret event becomes public we need to report it as created in sync-collection REPORT and CalendarEvent/changes. Likewise, we need to report it as destroyed when a non-secret event becomes secret. To do so, we add a new column to caldav.db to track the "privacy_modseq". This modseq tracks the latest modseq where the event's visibility changed.

The CLASS property now also no longer is stripped from iTIP messages. It is preserved for new invites (REQUEST) and for outgoing messages where the user is the organizer, but changes to it are discarded for invite updates and for replies (REPLY).

The JMAP side of things mostly was already implemented, except for CalendarEventNotification and a few other wrinkles we needed to fix. The main changes are for CalDAV. Both both protocols, we now add extensive tests, which make up the bulk of this PR.

This PR is cut in small commits, each commit aims to take exactly one step towards the overall goal of this pull requests. This hopefully makes it more accessible for review.


ksmurchison (approved):
This all looks sane and seems to cover what was discussed in Vienna

cassandane/tiny-tests/Caldav/class_get
```diff
@@ -0,0 +1,211 @@
...
+
+my %UID = (
+    public  => 'aaaaaaaa-0000-4000-8000-000000000001',
+    private => 'bbbbbbbb-0000-4000-8000-000000000002',
+    secret  => 'cccccccc-0000-4000-8000-000000000003',
+);
+
+my sub put_event ($caldav, $uid, $dtstart, $summary, $class)
+{
+    $caldav->Request('PUT', "Default/$uid.ics", join("\r\n",
```
wolfsage, line 19: nit: This is _fine_ but might be more readable with an indented heredoc.

Something like:
```
    my $ev = <<~EOF =~ s/\r?\n/\r\n/g;
        BEGIN:VCALENDAR
        ....
        BEGIN:VEVENT
        UID:$uid
        ...
        __CLASS__
        ...
        END:VCALENDAR
        VERSION:2.0
        EOF

    if ($class) {
        $ev =~ s/__CLASS__/CLASS:$class/;
    } else {
        $ev =~ s/\r\n__CLASS__//;
    }

    $caldav->Request(...)
````

File: cassandane/tiny-tests/Caldav/class_get [hunk 2]
```diff
@@ -0,0 +1,211 @@
...
+# Privacy is a property of the whole calendar event, so a recurrence override
+# is redacted even if the recurrence exception does not set CLASS on its own.
+sub test_class_get__override
+    ($self)
+{
+    my $uid = 'dddddddd-0000-4000-8000-000000000004';
+    my $owner_caldav = $self->default_user->caldav;
+
+    xlog $self, "create a recurring private event with an override";
+    $owner_caldav->Request('PUT', "Default/$uid.ics", join("\r\n",
```
wolfsage, line 159: Nit: You _could_ refactor to use put_event if it could take an arbitrary blob of extra data to add inside the VEVENT and then the actual different data on the event here from the rest of the test would be more clear, but this is fine

cassandane/tiny-tests/Caldav/class_get_conditional
```diff
@@ -0,0 +1,83 @@
...
+
+    my $sharee_caldav = $self->{instance}->create_user('sharee')->caldav;
+
+    xlog $self, "share the default calendar with sharee";
+    my $admin = $self->{adminstore}->get_client();
+    $admin->setacl("user.cassandane.#calendars.Default",
+                   sharee => 'lrswipkxtecdn9');
+    my $cal_href = '/dav/calendars/user/cassandane/Default';
+
+    xlog $self, "a conditional $method of the public event is not modified";
```
wolfsage, line 57: Should there be tests here that the owner gets 304s across the board or is that covered elsewhere?

imap/caldav_db.c
```diff
@@ -1308,6 +1308,40 @@ EXPORTED int caldav_writeical_jmap(struct caldav_db *caldavdb,
...
+
+    icalproperty *prop =
+        icalcomponent_get_first_property(comp, ICAL_CLASS_PROPERTY);
+    if (prop) {
+        switch (icalproperty_get_class(prop)) {
+        case ICAL_CLASS_CONFIDENTIAL:
+            return CAL_PRIVACY_SECRET;
+        case ICAL_CLASS_PUBLIC:
+            return CAL_PRIVACY_PUBLIC;
+        default:
```
wolfsage, line 1329: Why default and why default to private? In case upstream adds more values to the enum?

imap/caldav_db.c
```diff
@@ -1308,6 +1308,40 @@ EXPORTED int caldav_writeical_jmap(struct caldav_db *caldavdb,
...
+}
+
+EXPORTED const char *caldav_privacy_as_string(enum caldav_privacy privacy)
+{
+    switch (privacy) {
+    case CAL_PRIVACY_PRIVATE:
+        return "private";
+    case CAL_PRIVACY_SECRET:
+        return "secret";
+    default:
```
wolfsage, line 1344: We control caldav_privacy enum, should this be more explicit and not have a default or have a default that raises an error?

rsto:
I'm going to iterate some more on this PR.

imap/http_caldav.c
```diff
@@ -1051,8 +1051,33 @@ static int caldav_check_precond(struct transaction_t *txn,
...
 
+    /* Only the owner may remove a non-public resource from their calendar,
+       or duplicate it elsewhere. */
     enum caldav_privacy privacy = caldav_privacy_for_sharee(mailbox, cdata);
 
+    if (privacy != CAL_PRIVACY_PUBLIC
+        && (txn->meth == METH_DELETE || txn->meth == METH_MOVE
+            || txn->meth == METH_COPY))
+    {
+        xsyslog_ev(LOG_NOTICE, "sharee may not remove non-public resource",
```
wolfsage, line 1096: this should probably follow the new logfmt guidelines - https://github.com/cyrusimap/cyrus-imapd/blob/master/doc/README.logfmt.md 

"sharee may not remove non-public resource" would instead be some descriptive event name, I'm just not sure what that could be. @rjbs thoughts?

imap/caldav_db.h
```diff
@@ -88,8 +88,20 @@ struct caldav_data {
     const char *dtend;
     struct comp_flags comp_flags;
     const char *sched_tag;
+    /**
+     *  When privacy last crossed CAL_PRIVACY_SECRET. Only the latest crossing
+     *  is kept, so any number of crossings since the queried modseq looks like
+     *  a single one. An event with an even number of crossings may be reported
```
wolfsage, line 101: "may be reported to a sharee as destroyed although the sharee never knew of it" 

Why might an even number of crossings cause this bug? And isn't that kind of a bad bug?

wolfsage:
Hm, I imagine there's a timing oracle because 404 not found when an item legitimately doesn't exist is going to be far faster to process than find the item, load it, parse it, determine privacy, compare privacy, ... etc

But maybe network noise is enough to hide it...

wolfsage:
It's unavoidable I reckon, but if you have write access to someone else's calendar, and they hide an event from you that you could previously see, you can determine that the event became hidden by trying to create a new event on their calendar with the same uid as you'll get a uid conflict.

cassandane/tiny-tests/Caldav/class_get
rjbs: I said I would give this PR a light reading for "could the tests be better?".  I have done so, pretty lightly.  There's a lot of code, and I think most of it is _probably_ as good as it's going to get without changing how we do tests.  But I think that Matthew's point here is the biggest thing I'd actually look to change.  You have implemented a put-event or create-event a bunch of times, each one a bit different.  I think you could put the bones into Caldav.pm with some light templating, and reduce a lot of code.  I think _maybe_ some of the assert_X methods could go there also, but my first few looks for that didn't find any wins, so I could be wrong.

"Make event from some properties and a template" feels like a big win, and in keeping with our entity code.

imap/http_caldav.c
rjbs: dunno, naming is such a pain;  `event.update-nonpublic.rejected` or something?  I do not have a great taxonomy fully formed :(

imap/caldav_db.c
rsto: > Why default

Mostly because of ICAL_CLASS_X/ICAL_CLASS_NONE enums in libical. But I changed the switch here and in `caldav_privacy_as_string` to not use a default label. That will make us notice a new enum at compilation. Annoyingly we *still* have to handle the "unmatched enum" case after the switch, because an enum value could be some integer outside the allowed enum range.

> and why default to private?

That's what both JSCalendar 2.0 and RFC 5545 state to use for unexpected privacy/CLASS values.

imap/caldav_db.c
rsto: Yes, done.

imap/caldav_db.h
rsto: > Why might an even number of crossings cause this bug?

We only track the last crossing of an event switching its privacy from or to "secret", we do not track a history of crossings. Imagine a sharee's client syncs calendar events and some calendar event has privacy "secret". The client doesn't get this event reported at all. Then the owner flips the privacy of this event to "public" and back to "secret". Only then does the client resync. In this case, we can not tell if we reported this event to that client when it was public or not, so we report it as destroyed.

Note that the other direction of crossings also exists: crossing from "public" to "secret" and back to "public" causes us to report this event as "created" even though the client might already know this id.

> And isn't that kind of a bad bug?

This depends on the threat model. How bad is it to leak an id for a secret event if the client can't then do anything with that id other than noting that it once existed? The trade-off here is that we think that it's not great to leak the id, but the alternatives make the implementation significantly more complex. That trade-off is up for discussion, of course.

imap/http_caldav.c
rsto: I have a local branch that introduces the following logfmt events:

- `dav.caldav.privacy.denied`: a sharee tried to write or remove a non-public resource, or to make a resource non-public, and the request is rejected.
- `dav.caldav.privacy.hidden`: a sharee attempted to GET or HEAD a secret resource, which is reported as 404 as though it did not exist. This only gets reported if the sharee client named that resource, a PROPFIND and REPORT hide secret resources without logging.
- `jmap.calendarevent.privacy.denied`: a sharee calls CalendarEvent/set create with a UID that already exists in the account,  but they may not read all calendar objects for that UID.

rsto:
> I imagine there's a timing oracle because 404 not found when an item legitimately doesn't exist is going to be far faster to process than find the item, load it, parse it, determine privacy, compare privacy, ... etc

> It's unavoidable I reckon, but if you have write access to someone else's calendar, and they hide an event from you that you could previously see, you can determine that the event became hidden by trying to create a new event on their calendar with the same uid as you'll get a uid conflict.

Both boil down to the question if knowing about the existence of calendar object with a given UID is a critical threat or a "soft" threat. As for the even number of crossings discussed in https://github.com/cyrusimap/cyrus-imapd/pull/6250#discussion_r3833343715 it's certainly not great what we are doing here, but the alternatives would make the implementation worse:

If we want to prevent the oracle for a CalDAV request that names a resource, we'd have to slow down every read on CalDAV for inexistent calendar objects to the latency caused by determining the privacy of existing secret events. That's going to be hard to make deterministic, especially since we also caching.

Trying to prevent an oracle due to a uid conflict is just not feasible: even if we would try to trick an "evil" sharee by accepting to create a calendar object for an existing secret UID, we'd then have to serve them a calendar store that serves them and only them their newly created calendar object for that UID.

wolfsage:
Yeah, I mostly raised these as "Here are potential problems where users can figure out things we are trying to hide" and so someone who knows more about the risks / problems with that than I can say "this is acceptable" or "we should do something"

cassandane/tiny-tests/Caldav/class_get
rsto: I refactored all event creation into a `put_event` function in Caldav.pm, so no test in this PR includes verbatim iCalendar data anymore

cassandane/tiny-tests/Caldav/class_get
rsto: done, added new `put_event` and `create_sharee` functions to Caldav.pm. Will notify you when this is pushed to Github.

cassandane/tiny-tests/Caldav/class_get_conditional
rsto: I will add tests for the owner.

PR #6251, opened 2026-08-18
Cassandane: use UID, not sequence numbers, by default

rjbs:
UIDs are the right thing.  We should focus on testing via UIDs, and sequence numbers are tested on demand.


ksmurchison (approved):
LGTM

PR #6252, opened 2026-08-18
to_crlf: add and use a network newline normalize

rjbs:


rjbs:
@rsto Sorry, it's tedious, but largely one-for-one replacements.

cassandane/Cassandane/Test/CRLF.pm
```diff
@@ -0,0 +1,47 @@
...
+    my ($self, $desc, $in, $want) = @_;
+
+    $self->assert_str_equals($want, to_crlf($in), $desc);
+}
+
+sub test_to_crlf
+{
+    my ($self) = @_;
+
+    $self->_assert_crlf('lf becomes crlf',        "a\nb",      "a\r\nb");
```
rsto, line 23: I don't see a reason to doubt the implementation, but we might add tests that line folds are preserved, just because they are so common in iCalendar and vCard. E.g.

    $self->_assert_crlf('line fold (Space) is preserved',        "a\n b",      "a\r\n b");
    $self->_assert_crlf('line fold (Tab) is preserved',          "a\n\tb",     "a\r\n\tb");

cassandane/Cassandane/Util/CRLF.pm
```diff
@@ -0,0 +1,28 @@
...
+use base qw(Exporter);
+
+our @EXPORT = qw(&to_crlf);
+
+# Return $str with every line ending normalized to CRLF.  Tests build message,
+# iCalendar and vCard payloads as LF-terminated here-docs, but IMAP, JMAP and
+# sieve all want them in wire format, so nearly every such here-doc needs this
+# on the way out.
+#
+# The \r? makes this idempotent: text that is already CRLF, or that mixes the
```
rsto: "# The \r? makes this idempotent:": I don't think this belongs here (the sentence after the `:` is fine, but the `\r?` is not IMO). If at all, it belongs in the line right before the actual regular expression that is within the function (and hence an implementation detail and not a contract of the API).

rsto (approved):
LGTM. I have two suggestions, but neither changes the logic or fixes a flaw. There is no need to ask for re-review if you update the PR with these.

I know we also have test fixtures where we concatenate multiple lines as individual strings, e.g. in JMAPCalendars.email_get_calendarevents_deduplicate:

```
       body => ""
          . "\r\n--boundary\r\n"
          . "Content-Type: text/plain\r\n"
          . "\r\n"
          . "test"
          . "\r\n--boundary\r\n"
```

I'm not asking to rewrite them in this PR to use `to_crlf`, it just seemed topical to mention this here.

cassandane/Cassandane/Test/CRLF.pm
rjbs: Great idea, I'll add that.

cassandane/Cassandane/Util/CRLF.pm
rjbs: I will review and either change this or not.  (I assume you are correct, but will verify that I agree.)

rjbs:
@rsto As to your comment about this structure:

```perl
       body => ""
          . "\r\n--boundary\r\n"
          . "Content-Type: text/plain\r\n"
          . "\r\n"
          . "test"
          . "\r\n--boundary\r\n"
```

I had it in mind, but IIRC, I swept instances of this and what I found was that _some_ were safe and some were not, ad they'd concatenate things without `\r` in it.  I didn't want to mix them up in a mechanical MR. But I will return to look at them in another pass.

cassandane/Cassandane/Test/CRLF.pm
rjbs: Done, with comments changed to keep tests aligned.

PR #6253, opened 2026-08-19
squatter.c: restart batch indexing at uid 1 after mailbox recreation

rsto:
This makes batch indexing more robust against mailbox deletions and recreation between batches.

As a follow-up to dbbea16e9bf1afab93d4d2c8e3b42edef35528a7 which had introduced `resumeuid`, let's also detect the case where the mailbox got recreated between two batches. In this case we want to restart indexing from the first message in the mailbox. To prevent getting stuck we give up after a couple of batch restarts.

We introduce a new `squatter_batch_delay` option to configure how long squatter should wait between two indexing batches of the same mailbox. This option is used for testing, but it might also be useful to prevent mailbox lock starvation between squatter and other processes.

While we are at it, we also refine the test logic in the recently added squatter_index_default test that asserts the changes in dbbea16e9bf1afab93d4d2c8e3b42edef35528a7.

### DIFF
--- a/imap/squatter.c
+++ b/imap/squatter.c
@@ -275,12 +275,21 @@ done:
  * indexing goes away, so give up rather than block this run on it. */
 #define MAX_UNPRODUCTIVE_UPDATES 3
 
+/* How many times to restart indexing the same mailbox because it got
+ * recreated between two batches. A client that keeps recreating a mailbox
+ * would otherwise block this run on that mailbox forever. */
+#define MAX_MAILBOX_RESTARTS 3
+
 /* This is called once for each mailbox we're told to index. */
 static int index_one(const char *name, int blocking)
 {
     struct mailbox *mailbox = NULL;
     int nunproductive = 0;
+    int nrestarts = 0;
     uint32_t resumeuid = 0;
+    uint32_t uidvalidity = 0;
+    char *prev_uniqueid = NULL;
+    int32_t batch_delay = config_getduration(IMAPOPT_SQUATTER_BATCH_DELAY);
     int r;
     int flags = SEARCH_UPDATE_BATCH;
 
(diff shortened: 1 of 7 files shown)


rsto:
@wolfsage rebased to master and resolved conflicts

imap/squatter.c
```diff
@@ -378,11 +385,39 @@ static int index_one(const char *name, int blocking)
...
+                r = IMAP_AGAIN;
+                goto done;
+            }
+        }
+        uidvalidity = mailbox->i.uidvalidity;
+        resumeuid = 0;
+    }
+
+    free(prev_uniqueid);
+    prev_uniqueid = xstrdupnull(mailbox_uniqueid(mailbox));
```
wolfsage: If we don't have a mailbox_uniqueid, we won't flag prev_uniqueid properly for retry logic, and nrestarts might not get incremented.

Realistically, I'd expect this cannot happen in practice, or if it did the following mailbox_open_... above will fail.

Maybe if it *does* though we should assert we got a value, or xsyslog and bail or .. something.

wolfsage (changes requested):

imap/squatter.c
rsto: You say "maybe" but you request changes on this PR. So you want this getting handled one way or the other?

wolfsage: Yeah, the code here allows for null (which would be weird / surprising?) so we should make sure we handle it sensibly. If you believe that looping around on again: will fail to open the mailbox after getting null here, then there's nothing to do, I'm just not sure we can rely on that or not.

rsto: I now made squatter report and skip that mailbox. I would have made this a fatal error, but I don't want a rolling squatter to become blocked from running for a single mailbox that lost its uniqueid.

wolfsage: yeah that seems good

wolfsage (approved):

PR #6257, opened 2026-08-19
Remove support for FUD.

ksmurchison:
As far as we know, this service was only used at CMU, and they no longer run Cyrus.


rjbs:
This has no review requested, so I didn't, but:  the change to rip out nntp left some dead links in docsrc, turned up by building the docsite.   Check on that here before review request?  I'm trying to get us warnings clean.

rjbs (approved):

PR #6258, opened 2026-08-19
httpd.c: remove the RSS and CGI modules

ksmurchison:
We don't believe anybody is using these modules,
so removing the reduces code bloat and potential
security issues.


dilyanpalauzov:
I use the CGI module.  The reason was that I had particular DDOS attach over HTTP requests, had to write a filter for the http requests and at that time (Autumn 2024) nginx /njs had not allowed such filters.  While the cgi module was pretty easy to extend. As nginx does not allow CGI-BIN, one has to use something different just for CGI: the cyrus cgi module or thttpd. For the above reasons I stopped using thttpd.

rjbs:
Nonetheless, we are going to remove this.  We want to minimize the maintenance and security surface of the project.  Cyrus is an excellent IMAP and JMAP server and we'd like to focus on improving those without excess baggage.

rjbs (approved):

PR #6259, opened 2026-08-19
jmap_calendar: process destroys before creates in Calendar/set

rjbs:
This is the update deferred by e88780b08e, which reordered the rest.  I wanted to wait until JMAP for Calendar stuff was stabler.

git diff --color-moved shows this is just a move of 29 lines, no real change beyond that.


rsto (approved):

rjbs:
The action ran.  The branch is merged.  Why is the status not merged?

PR #6262, opened 2026-08-19
caldav: do not dereference an unparsed calendar-availability

FathiBenNasr:
### The crash

`httpd` dies on a CalDAV free/busy request (`POST` to a scheduling Outbox) as soon as the query names **more than one recipient and any one of them publishes office hours** (a `CALDAV:calendar-availability` on their scheduling Inbox):

```
#0  icalcomponent_get_first_component (libical.so.3 + 0x33f16)
#1  busytime_query_local
#2  sched_busytime_query
#3  caldav_post
#4  meth_post
    signal 11 (Segmentation fault, core dumped)
```

The client sees the connection close mid-response; behind a reverse proxy it surfaces as a 502.

### Why

`add_vavailability()` stores the component and walks it in the same breath:

```c
newav->ical = ical;
vav = icalcomponent_get_first_real_component(ical);
newav->per.start = icalcomponent_get_dtstart(vav);
```

Neither pointer is checked, and both can be NULL:

* `busytime_query_local()` passes `icalparser_parse_string(buf_cstring(&attrib))` straight in. The annotation is stored verbatim and parsed only here, so a value that does not parse yields NULL.
* Even for a component that parses, `icalcomponent_get_first_real_component()` returns NULL when there is nothing it recognises inside, leaving `vav` NULL for the `icalcomponent_get_dtstart()` on the next line.

`add_vavailability()` is static and gets inlined, which is why the frame attributed to it in the backtrace is `busytime_query_local`.

### The change

Return early when there is nothing usable, before touching the array. The array owns what it is handed — entries are released through `vav->ical` at the end of the query — so a refused component is freed here rather than leaked. A client whose availability cannot be applied now simply gets a free/busy answer without it, instead of losing the connection.

### Reproducing

Observed on 3.8.3, and the same code is on master.

1. Give a user office hours:
   `PROPPATCH` `CALDAV:calendar-availability` on their scheduling Inbox with any VAVAILABILITY.
2. `POST` a `METHOD:REQUEST` VFREEBUSY to a scheduling Outbox with **two** `ATTENDEE` lines, one of them that user.
3. The child process dies with the backtrace above.

Asking about the same people **one at a time** answers correctly every time, which is what made the failure look data-dependent at first. Removing the availability also makes it go away.

I could not tell from a production core dump which of the two pointers was NULL — the guard covers both, and each is reachable from the call site.

### DIFF
--- a/cassandane/tiny-tests/Caldav/sched_busytime_query_bad_availability
+++ b/cassandane/tiny-tests/Caldav/sched_busytime_query_bad_availability
@@ -1,0 +1,50 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_sched_busytime_query_bad_availability
+    :NoVirtDomains :MagicPlus
+    ($self)
+{
+    my $CalDAV = $self->{caldav};
+
+    # Create a second calendar user
+    my $user = $self->{instance}->create_user("friend");
+    $user->caldav;
+
+    # Create a +DAV IMAP connection for the second user
+    my $plusstore = $self->{instance}->get_service('imap')->create_store(username => 'friend+dav');
+    my $imaptalk = $plusstore->get_client();
+
+    # CALDAV:calendar-availability is stored verbatim and only parsed when a
+    # free/busy query needs it, so a value that does not parse leaves
+    # icalparser_parse_string() returning NULL.  add_vavailability() used to
+    # walk that pointer, taking httpd down with it - the client sees the
+    # connection close mid-response rather than an answer.
+    $imaptalk->setmetadata("#calendars.Inbox",
+        "/shared/vendor/cmu/cyrus-httpd/<urn:ietf:params:xml:ns:caldav>calendar-availability",
+        "this is not an iCalendar object");
+    $self->assert_str_equals('ok', $imaptalk->get_last_completion_response());
+
+    my $query = <<EOF;
+BEGIN:VCALENDAR
+VERSION:2.0
+PRODID:-//Apple Inc.//Mac OS X 10.10.4//EN
+METHOD:REQUEST
+BEGIN:VFREEBUSY
+UID:5D3E1B26-8CF2-4C9C-9F4E-3D0F5C0A7A11
+DTSTAMP:20210802T131858Z
+DTEND:20210903T000000Z
+DTSTART:20210902T210000Z
+ATTENDEE:MAILTO:cassandane\@example.com
+ATTENDEE:MAILTO:friend\@example.com
+ORGANIZER:MAILTO:cassandane\@example.com
+END:VFREEBUSY
+END:VCALENDAR
+EOF
+
+    xlog $self, "freebusy query covering a user whose availability does not parse";
+    my $res = $CalDAV->Request('POST', 'Outbox',
+                               $query, 'Content-Type' => 'text/calendar');
+    my $text = Dumper($res);
+    $self->assert_matches(qr/schedule-response/, $text);
+}
--- a/changes/next/caldav-availability-null-guard
+++ b/changes/next/caldav-availability-null-guard
@@ -1,0 +1,20 @@
+Description:
+
+Fix a crash in the CalDAV free/busy handler when a scheduling query covers
+several recipients and one of them publishes a CALDAV:calendar-availability.
+
+Documentation:
+
+No documentation change: the fix restores the documented behaviour.
+
+Config changes:
+
+None.
+
+Upgrade instructions:
+
+None.
+
+GitHub issue:
+
+None filed; the crash and its reproduction are described in the pull request.
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -7470,6 +7470,19 @@ add_vavailability(struct vavailability_array *vavail, icalcomponent *ical)
     icalcomponent *vav;
     icalproperty *prop;
 
+    /* Callers hand us whatever came out of a parse: the CALDAV:calendar-availability
+       annotation is stored verbatim and only parsed here, so icalparser_parse_string()
+       may well have returned NULL, and a component that did parse may still hold
+       nothing we can take a period from.  Either way this array owns what it is
+       given, so anything refused has to be freed rather than leaked. */
+    if (!ical) return;
+
+    vav = icalcomponent_get_first_real_component(ical);
+    if (!vav) {
+        icalcomponent_free(ical);
+        return;
+    }
+
     /* Grow the array, if necessary */
     if (vavail->len == vavail->alloc) {
         vavail->alloc += 10;  /* XXX  arbitrary */
@@ -7480,8 +7493,6 @@ add_vavailability(struct vavailability_array *vavail, icalcomponent *ical)
     /* Add new vavailability */
     newav = &vavail->vav[vavail->len++];
     newav->ical = ical;
-
-    vav = icalcomponent_get_first_real_component(ical);
 
     /* Set period */
     newav->per.start = icalcomponent_get_dtstart(vav);
--- a/imap/http_caldav.c
+++ b/imap/http_caldav.c
@@ -8019,6 +8019,7 @@ icalcomponent *busytime_query_local(struct transaction_t *txn,
                                   icalparser_parse_string(buf_cstring(&attrib)));
             }
         }
+        buf_free(&attrib);
         free(mboxname);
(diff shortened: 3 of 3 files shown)


ksmurchison:
Thank you for the report and fix.  Are you able to write a Cassandane test which fails on current master, and succeeds when the fix is in place?

FathiBenNasr:
Thanks for looking at it. I have pushed a Cassandane test: `cassandane/tiny-tests/Caldav/sched_busytime_query_bad_availability`.

It stores a `CALDAV:calendar-availability` that is not iCalendar — the annotation is kept verbatim and only parsed when a query needs it — and then runs the two-recipient scheduling free/busy query from the existing `sched_busytime_query` test. Without the guard, `icalparser_parse_string()` returns NULL, `add_vavailability()` walks it and `httpd` dies, so the client gets a closed connection instead of a `schedule-response`; with the guard the availability is simply not applied and the query is answered.

Two things I should be straight about:

* I could not run the Cassandane suite myself — this is a production host with the distribution package, not a build tree — so the test is written to the idiom of its neighbour rather than verified green. If it needs adjusting to run, tell me what it reports and I will fix it.
* The crash I originally hit in production came from an availability that looked perfectly valid (a VAVAILABILITY with a single AVAILABLE block, round-tripping unchanged through PROPFIND), on a query with two or more recipients; one recipient never crashed. I could not reduce that case to something a test can set up deterministically, which is why the patch also guards the second pointer, `icalcomponent_get_first_real_component()` returning NULL. The test covers the parse-failure path only.

ksmurchison:
So, the Cass test doesn't work because users can't write annotations to non-email mailboxes by default.  I applied the following patch to fix the test.  If you want to apply it and squash merge it into your second commit, we should be good to go.
```

`diff --git a/cassandane/tiny-tests/Caldav/sched_busytime_query_bad_availability b/cassandane/tiny-tests/Caldav/sched_busytime_query_bad_availability
index 1f388c1915..63b7b7e19d 100644
--- a/cassandane/tiny-tests/Caldav/sched_busytime_query_bad_availability
+++ b/cassandane/tiny-tests/Caldav/sched_busytime_query_bad_availability
@@ -2,26 +2,28 @@
 use Cassandane::Tiny;
 
 sub test_sched_busytime_query_bad_availability
-    :min_version_3_4 :NoVirtDomains
+    :NoVirtDomains :MagicPlus
     ($self)
 {
     my $CalDAV = $self->{caldav};
 
-    my $admintalk = $self->{adminstore}->get_client();
+    # Create a second calendar user
+    my $user = $self->{instance}->create_user("friend");
+    $user->caldav;
 
-    $admintalk->create("user.friend");
-    $admintalk->setacl("user.friend", admin => 'lrswipkxtecdan');
-    $admintalk->setacl("user.friend", friend => 'lrswipkxtecdn');
+    # Create a +DAV IMAP connection for the second user
+    my $plusstore = $self->{instance}->get_service('imap')->create_store(username => 'friend+dav');
+    my $imaptalk = $plusstore->get_client();
 
     # CALDAV:calendar-availability is stored verbatim and only parsed when a
     # free/busy query needs it, so a value that does not parse leaves
     # icalparser_parse_string() returning NULL.  add_vavailability() used to
     # walk that pointer, taking httpd down with it - the client sees the
     # connection close mid-response rather than an answer.
-    $admintalk->setmetadata("user.friend.#calendars.Inbox",
+    $imaptalk->setmetadata("#calendars.Inbox",
         "/shared/vendor/cmu/cyrus-httpd/<urn:ietf:params:xml:ns:caldav>calendar-availability",
         "this is not an iCalendar object");
-    $self->assert_str_equals('ok', $admintalk->get_last_completion_response());
+    $self->assert_str_equals('ok', $imaptalk->get_last_completion_response());
 
     my $query = <<EOF;
`
```

FathiBenNasr:
Thank you — `:MagicPlus` with a `friend+dav` connection was exactly the piece I was missing, and I can confirm it works: the `SETMETADATA` is accepted where mine was refused with `NO Invalid mailbox type`.

I have since built current master from source and can now run Cassandane, so I was able to do what I could not do before — run it against both builds. I have applied your patch and squashed it into the second commit as you asked. But I should report what the runs say, because it is not the result either of us wanted:

**With your fix applied, the test passes on unpatched master as well as on the patched build.**

| build | result |
| --- | --- |
| `master` (dd81135) | `OK (1 tests)` |
| this branch | `OK (1 tests)` |

I checked that the annotation really is being read, rather than the test quietly doing nothing. Storing a *valid* VAVAILABILITY by the same `friend+dav` route and re-running the query gives:

```
FREEBUSY;FBTYPE=BUSY-UNAVAILABLE:20210902T210000Z/20210903T000000Z
```

So the setup works and `add_vavailability()` is reached. With the unparsable value, master and the patched build return byte-identical free/busy responses, and nothing crashes.

The reason appears to be libical rather than anything in the test. `icalcomponent_get_first_component()` opens with `icalerror_check_arg_rz((c != 0), "component")`, so under libical 4 the NULL is returned back rather than dereferenced, and `icalcomponent_get_dtstart(NULL)` then walks the same NULL-tolerant path to `icaltime_null_time()`. Since `configure.ac` requires `libical >= 3.99.99`, master is always on that libical. My production backtrace is `libical.so.3` — a 3.8 install against libical 3 — which is where the dereference actually faults.

So, as far as I can measure it: the guard is right and matters for the 3.8/3.10 branches, but on master this test cannot fail, and a test that cannot fail is worse than none — it will pass for ever and tell nobody anything. Three options, and I am happy with whichever you prefer:

1. Drop the test and take the two commits as hardening, with the reachability noted in the commit message.
2. Keep the fix on master and put the test where it can fail, on a branch that still builds against libical 3.
3. If you would rather the test earn its place on master, it needs an input that makes `icalcomponent_get_first_real_component()` return NULL from a *parsed* component — a VCALENDAR carrying only a VTIMEZONE does it, but `proppatch_availability()` refuses that with `CALDAV:valid-calendar-object-resource`, and the `+dav` `SETMETADATA` route is the only way I have found in. I can push that variant if you would like to see it.

For completeness, the whole `Caldav` suite on this branch is 107/108, and the one failure is `Caldav.schedule_organizer_mismatch` — it wants the syslog event `event=caldav.schedule.skipped` added in 78f6f86, which is after the base this branch sits on. It fails identically on my #6264 branch and is unrelated to either patch; I will rebase both onto master.

FathiBenNasr:
Nudge, three weeks on — and thank you again for the `:MagicPlus` /
`friend+dav` pointer, which was the piece I was missing.

Nothing has changed on my side since 21 August: the Cassandane test stores the
bad `calendar-availability` through a `friend+dav` connection as you showed,
and the crash is gone with the patch. GitHub reports the branch mergeable
against current master; `mergeable_state` is `blocked` only for want of a
review.

If there is anything left for me here — a rebase, a different test placement,
or a smaller patch — tell me which and I will do it. And if you would rather
not carry it at all, closing it is a perfectly good answer; I would rather know
than leave it sitting.

ksmurchison:
As you can see CI is failing.  The gcc+asan failure reveals a legitimate memory leak.  Can you fix that?

FathiBenNasr:
Fixed, pushed as a third commit (9b0b1b5).

The leak is in `busytime_query_local()`: the `CALDAV/CS:calendar-availability`
annotation is looked up into a local `struct buf` that is never freed.

```
Direct leak of 64 byte(s) in 1 object(s) allocated from:
    #2 _buf_ensure lib/buf.c:61
    #3 buf_cstring lib/buf.c:84
    #4 _annotate_lookup imap/annotate.c:2891
    #7 busytime_query_local imap/http_caldav.c:8009
```

It predates this branch — `_annotate_lookup()` only allocates when the
annotation actually exists, so no test had reached it until the one added
here stores an availability. In production it is one buffer per free/busy
query that consults an availability, in a long-lived `httpd`.

The fix is `buf_free(&attrib);` before the two `free()`s. I have kept it as
its own commit rather than folding it into the guard, since it is a separate
defect that the new test merely exposes — say the word if you would rather
have it squashed.

I could not run the ASAN build here to confirm the log comes back clean, so
I am relying on CI for that; if anything else falls out I will fix it.

For the record, the `clang w/cass` failure in the same run is unrelated to
this branch — `configure: error: *** A compiler with support for C++17
language features is required.`

ksmurchison (approved):

PR #6263, opened 2026-08-20
universal logfmt: phase 0

rjbs:
This is the first of what I expect to be a multi-phase process of converting everything, or roughly everything, from xsyslog to logfmt, either through xsyslog_ev or other methods.

This PR focuses on:

1. updating Cassandane to keep catching problems when they stop being DBERROR or IOERROR
2. add lf_X helpers for various popular Xes (note especially LF_FN)
3. setting up build-time checks that we're following "the rules"

Oh and it adds "the rules", some documentation in ./doc that I think needs to be significantly edited and merged into docsrc, but which felt good enough to start with at phase 0.  (We're not starting on phase 1 because this isn't Tcl.)

### DIFF
--- a/cassandane/utils/syslog.c
+++ b/cassandane/utils/syslog.c
@@ -6,6 +6,12 @@
  *
  * Set CASSANDANE_SYSLOG_FNAME in the environment to specify the file
  * to which logged lines should be appended.
+ *
+ * Each captured line is prefixed with the severity it was logged at, as
+ * "<err> ", so that log consumers can recognise problems without having
+ * to match on message text.  Real syslog discards the priority by the
+ * time it reaches the log file, which is why we record it ourselves.
+ * Cassandane::Instance::_check_syslog depends on this prefix.
  */
 
 /* need _GNU_SOURCE for RTLD_NEXT */
@@ -88,6 +94,22 @@ EXPORTED void closelog(void)
     is_opened = 0;
 }
 
+/* The severity names used by syslog.conf(5), so that they'll look familiar. */
+static const char *severity_name(int priority)
+{
+    switch (LOG_PRI(priority)) {
+    case LOG_EMERG:   return "emerg";
+    case LOG_ALERT:   return "alert";
+    case LOG_CRIT:    return "crit";
+    case LOG_ERR:     return "err";
+    case LOG_WARNING: return "warning";
+    case LOG_NOTICE:  return "notice";
+    case LOG_INFO:    return "info";
+    case LOG_DEBUG:   return "debug";
+    default:          return "unknown";
+    }
+}
+
 static void fake_vsyslog(int priority, const char *format, va_list ap)
 {
     struct timeval now = {0};
@@ -103,8 +125,11 @@ static void fake_vsyslog(int priority, const char *format, va_list ap)
 
     gettimeofday(&now, NULL);
 
+    /* The severity goes first, so that it can be matched with an anchored
+     * pattern that can never be confused by the message body. */
     strftime(timestamp, sizeof(timestamp), "%b %d %T", localtime(&now.tv_sec));
-    fprintf(out, "%s.%06" PRIdMAX " %s %s[%" PRIdMAX "]: ",
+    fprintf(out, "<%s> %s.%06" PRIdMAX " %s %s[%" PRIdMAX "]: ",
+                 severity_name(priority),
                  timestamp, (intmax_t) now.tv_usec,
                  hostname, myident, (intmax_t) pid);
     errno = saved_errno;
--- a/doc/README.logfmt.md
+++ b/doc/README.logfmt.md
@@ -7,7 +7,7 @@ logs can be parsed reliably instead of scraped with regular expressions.
 This document is the rulebook for that conversion.  It covers how to emit an
 event, how to name it, and how to name and format its fields.  The vocabulary
 itself — the list of every key you're allowed to use — lives in
-[`doc/logfmt-keys`](logfmt-keys) and is enforced by `tools/check-logfmt-keys`.
+[`doc/logfmt-keys`](logfmt-keys) and is enforced by `tools/lint-logfmt-keys`.
 
 ## Why a rulebook
 
@@ -170,7 +170,7 @@ Two structural prefixes stack on top of the above:
 Rules:
 
 - **Every key must be registered** in `doc/logfmt-keys` before use.
-  `tools/check-logfmt-keys` fails the build otherwise.  Adding a key is meant
+  `tools/lint-logfmt-keys` fails the build otherwise.  Adding a key is meant
   to be easy — a one-line patch — but deliberate.
 - **Same fact, same key, everywhere.** If you're about to invent a key for
   something that already has one, use the existing one even if you'd have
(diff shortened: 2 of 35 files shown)


cassandane/utils/syslog.c
```diff
@@ -103,8 +125,12 @@ static void fake_vsyslog(int priority, const char *format, va_list ap)
 
     gettimeofday(&now, NULL);
 
+    /* The severity goes first, so that it can be matched with an anchored
+     * pattern that can never be confused by the message body.
+     */
```
rsto: nit: stray newline. Then again, I notice that also Doxygen comments use this, so maybe I'm just not used to it.

lib/util.h
```diff
@@ -459,6 +488,48 @@ enum xsyslog_ev_arg_type {
     (xsyslog_ev_arg){ key, LF_RAW, { .s = buf_release(&value) } };          \
 })
 
+/* An argument that isn't there.  _xsyslog_ev() drops these, which is how the
+ * conditional helpers below manage to contribute nothing to the event.
+ */
+#define lf_skip()           (xsyslog_ev_arg){ "-", LF_SKIP, { .s = NULL } }
+
+/* A boolean, as 1 or 0.  Use this rather than lf_c(), which takes a char and
```
rsto, line 510: While we are it, let's make these proper Doxygen comments? E.g.

```
  /** @brief A boolean, as 1 or 0.
   *
   * Use this rather than lf_c(), which takes a char and would log
   * lf_b("k", 1)'s value as the control character 0x01.
   */
```

lib/util.h
```diff
@@ -459,6 +488,48 @@ enum xsyslog_ev_arg_type {
...
+                                              { .lld = (long long)(value) } }
+
+/* An elapsed time, in seconds, to millisecond precision.  Always seconds:
+ * don't log milliseconds under a key that doesn't say so.
+ */
+#define lf_duration(key, secs)                                              \
+    (xsyslog_ev_arg){ key, LF_DURATION, { .f = (secs) } }
+
+/* A struct buf.  Takes a pointer. */
+#define lf_buf(key, bufp)   lf_s(key, buf_cstring(bufp))
```
rsto, line 527: `buf_cstring` may cause a heap-allocation and possibly copy in case the buffer isn't zero terminated and doesn't have space for a zero byte. In practice that's irrelevant because a plain syslog call also would have required a C string, but you might want to mention here that it calls `buf_cstring` on the buffer.

imap/mboxname.h
```diff
@@ -368,4 +368,11 @@ void mboxname_zero_counters(const char *mboxname);
 void logfmt_push_mbname(struct logfmt *lf,
                         const char *key, const mbname_t *mbname);
 
+void logfmt_arg_mbname(struct logfmt *lf, const char *key, const void *value);
+
+/* Log an mbname_t under `key`, for xsyslog_ev().  Normally "mbox.name", but
+ * "old.mbox.name" and the like are why this one takes a key at all.
```
rsto: I don't understand. Is `key` supposed to be `old` or `old.mbox.name`?

doc/logfmt-keys
```diff
@@ -0,0 +1,119 @@
...
+### r. - the request or connection
+
+r.clienthost            the client end of the connection, as host [ip]
+r.sessionid             this process's session id
+r.tid                   trace id, when the client supplied one
+remote.sessionid        the session id of a backend we proxied to
+
+### u. - the user
+
+u.username              the authenticated userid the event happened for
```
rsto, line 31: why is this `username` and not `userid`? the description of the key uses "userid"

doc/logfmt-keys
```diff
@@ -0,0 +1,119 @@
...
+mbox.uniqueid           the mailbox's uniqueid
+
+### msg. - a message
+
+msg.cid                 the message's conversation id
+msg.date                the message's date, as sent
+msg.emailid             the message's JMAP id
+msg.from                the message's envelope sender
+msg.guid                the message's guid
+msg.id                  the message's RFC 5322 Message-ID
```
rsto, line 57: as long as we know what this stands for, I'm OK with it. But I would find "messageid" or the like probably less ambiguous

doc/README.logfmt.md
```diff
@@ -0,0 +1,337 @@
...
+and you want to say so with some context.
+
+```c
+xsyslog_ev(LOG_ERR, "mailbox.append.failed",
+           lf_mailbox(mailbox),
+           lf_record(record),
+           lf_err("error", r));
+```
+
+This is the direct replacement for `syslog()` and `xsyslog()`, and it's what
```
rsto, line 58: Is something off with the markup here? Github diff shows this formatted as code snippet.

imap/http_caldav_sched.c
```diff
@@ -3050,12 +3050,14 @@ void sched_reply(const char *cal_ownerid, const char *sched_userid,
...
-                    lf_s("organizer", organizer),
-                    lf_s("sched_addrs", addrs));
+            /* neither the ORGANIZER nor any ATTENDEE matched one of the
+             * user's scheduling addresses, so there's nothing to schedule
+             */
+            xsyslog_ev(LOG_WARNING, "caldav.schedule.skipped",
+                    lf_s("cal.uid", icalcomponent_get_uid(comp)),
+                    lf_s("cal.organizer", organizer),
+                    lf_s("sched.userid", sched_userid),
+                    lf_s("sched.addresses", addrs));
```
rsto: Does this deserve a `lf_strarray`? Note that we call `strarray_join` with `,` to generate that list. That's safe here, but a string list entry that contains a `,` would produce bogus output. ` lf_strarray` would allow avoiding the heap-allocated throwaway C string and also unify how we log string arrays.

rsto (approved):
This looks great. I have some nits, probably some bikeshedding, but nothing is blocking approval.

lib/util.h
rjbs: I want to do this is a subsequent branch (very soon).  I want to put in things like "how the logfmt code works" and felt like this could become a bit more of a bikeshed.  But yeah, this stuff is a great place for doxygen imho also.

imap/mboxname.h
rjbs: The second, I have clarified.

That said, more than once I thought, "possibly this should be `ln_fn(prefix_or_null, *mbname)` so it would be "old".  I planned to ask about that as a follow-up re: ergonomics.

doc/logfmt-keys
rjbs: Just following existing code.  This one in particular needs to be revisited. In Fastmail, it's confusing because our own logging wants to own username (but we use something else instead) and would call this "cyrusname" which is weird for Cyrus to call it.  I think we'll want accountId and email, but that requires finishing "only one virtualdomains setting".

doc/logfmt-keys
rjbs: This is following Fastmail internal logging, and unlikely to change.

doc/README.logfmt.md
rjbs: Screenshot 2026-08-20 at 08 31 58

Not for me!  Will leave this be.

imap/http_caldav_sched.c
rjbs: Good idea.

lib/util.h
rsto: ack

PR #6264, opened 2026-08-20
caldav: use the canonical userid when scheduling to a local recipient

FathiBenNasr:
### Symptom

On a server with `virtdomains: userid` and `defaultdomain` set, **no local user ever receives an invitation**. The organiser's copy comes back with

```
ATTENDEE;...;SCHEDULE-STATUS=5.1:mailto:someone@example.com
```

and the recipient's scheduling Inbox stays empty. The server logs

```
sched_deliver(someone@example.com)
caladdress_lookup(userid: 'someone@example.com', ...)
CalDAV scheduling delivery to someone@example.com
sched_deliver_local(organiser@example.com, someone@example.com, 0)
caldav_store_resource(user.someone.#calendars.Default) failed: 403 Forbidden ()
```

Free/busy queries for the same recipients work; only delivery fails.

### Cause

`caladdress_lookup()` trims the domain off a local calendar user address only when virtdomains is off:

```c
if (!config_virtdomains) {
    *at = '\0';  // trim off the domain
}
```

So with virtdomains on, a user in the **default** domain keeps `@defaultdomain` in `sched_param->userid`, while their mailbox and its ACL are keyed by the bare userid — which is what Cyrus itself writes when it creates their calendars (`user.someone.#calendars.Default` with an ACL entry `someone`).

`sched_deliver_local()` passes that userid to `caldav_store_resource()`, and `caldav_store_preprocess()` uses it to decide what may be written:

```c
struct auth_state *authstate = auth_newstate(userid);
rights = cyrus_acl_myrights(authstate, mailbox_acl(mailbox));
if (rights & DACL_WRITECONT) allow_propupdates = propupdate_all;
...
if (!allow_propupdates) {
    txn->error.precond = DAV_NEED_PRIVS;
```

`auth_newstate("someone@example.com")` matches no entry in an ACL keyed `someone`, so the recipient is left with whatever `anyone` grants — never `DACL_WRITECONT`. Hence 403, hence 5.1.

I confirmed the precondition under gdb (`precond = 8`, `DAV_NEED_PRIVS`), and confirmed the cause by adding a redundant ACL entry under the qualified name:

```
setacl user/someone/#calendars/Default someone@example.com lrswipkxtecdan9
```

after which the very next invitation is delivered (`SCHEDULE-STATUS=1.2`, message in the recipient's calendar). That is a workaround, not a fix — it duplicates every ACL entry.

### The change

Take `mbname_userid()` — the same canonicalisation that produced the mailbox name being written to — so the userid and the ACL agree. Sites without virtdomains, and users outside the default domain, are unaffected: for them the canonical form is the one already in use.

### Test

`cassandane/tiny-tests/Caldav/sched_deliver_virtdomains` runs a plain organiser→attendee delivery with `:VirtDomains`. The existing scheduling tests are all tagged `:NoVirtDomains`, which is presumably why this has gone unnoticed.

I could not run the Cassandane suite myself — this is a production host with the distribution package (3.8.3), not a build tree — so the test is written to the idiom of its neighbours rather than verified green. Tell me what it reports and I will fix it.

### DIFF
--- a/cassandane/Cassandane/Cyrus/TestCase.pm
+++ b/cassandane/Cassandane/Cyrus/TestCase.pm
@@ -271,6 +271,12 @@ magic(DelayedExpunge => sub {
 magic(VirtDomains => sub {
     shift->config_set(virtdomains => 'userid');
 });
+magic(DefaultDomain => sub {
+    # Pairs with :VirtDomains.  Without a default domain the recipient of a
+    # local invitation is a foreign address, and is scheduled over iMIP rather
+    # than delivered - so the local delivery path is never entered at all.
+    shift->config_set(defaultdomain => q{example.com});
+});
 magic(NoVirtDomains => sub {
     shift->config_set(virtdomains => 'off');
 });
--- a/cassandane/tiny-tests/Caldav/sched_deliver_virtdomains
+++ b/cassandane/tiny-tests/Caldav/sched_deliver_virtdomains
@@ -1,0 +1,76 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_sched_deliver_virtdomains
+    :min_version_3_4 :VirtDomains :DefaultDomain
+    ($self)
+{
+    my $CalDAV = $self->{caldav};
+
+    my $admintalk = $self->{adminstore}->get_client();
+
+    $admintalk->create("user.friend");
+    $admintalk->setacl("user.friend", admin => "lrswipkxtecdan");
+    $admintalk->setacl("user.friend", friend => "lrswipkxtecdn");
+
+    # The recipient needs a calendar home before anything can be delivered
+    # into it: caladdress_lookup() falls back to iMIP when the cal-home-set
+    # cannot be found, and the local delivery path this tests is then never
+    # entered at all.  Connecting over DAV provisions it.
+    my $service = $self->{instance}->get_service("http");
+    my $friendtalk = Net::CalDAVTalk->new(
+        user => "friend",
+        password => "pass",
+        host => $service->host(),
+        port => $service->port(),
+        scheme => "http",
+        url => "/",
+        expandurl => 1,
+    );
+
+    my $CalendarId = $CalDAV->NewCalendar({name => "invitations"});
+    $self->assert_not_null($CalendarId);
+
+    my $uid = "B0C4A0F2-3F4E-4A2E-9C1B-7E5C6B2A9D31";
+    my $ics = <<EOF;
+BEGIN:VCALENDAR
+VERSION:2.0
+PRODID:-//Cassandane//NONSGML//EN
+BEGIN:VEVENT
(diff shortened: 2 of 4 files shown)


FathiBenNasr:
Rebased onto master, and I have corrected the Cassandane test — it did not exercise the bug as submitted, and I would rather say so than have it merged that way.

I can now build and run Cassandane, and the test failed identically on master and on this branch until three things were fixed. None of them is about the patch:

1. **`:VirtDomains` alone is not this bug's condition.** It sets `virtdomains: userid`, but `defaultdomain` stays at Cassandane's default `defdomain`, so `friend@example.com` is a foreign address and is scheduled over iMIP — `SCHEDULE-STATUS=1.1`, and `caladdress_lookup()`'s local branch, where the fix lives, is never entered. There is no magic for the default domain, so the commit adds one next to `VirtDomains`.

2. **The recipient needs a calendar home before anything can be delivered into it.** `caladdress_lookup()` falls back to iMIP when `proxy_mlookup()` of the cal-home-set fails, so creating `user.friend` over IMAP is not enough. Connecting once as that user over DAV provisions it — the idiom `davsharing` already uses.

3. **A calendar collection cannot be checked over IMAP.** `SELECT` on `user.friend.#calendars.Inbox` is refused with `NO Invalid mailbox type` — which is not a missing mailbox and does not mean the delivery failed. The test now reads the scheduling Inbox with a `PROPFIND` instead.

With those corrected, the test does what it should:

| build | organiser's copy of the invitation |
| --- | --- |
| master (dd81135) | `SCHEDULE-STATUS=5.1` — refused, as the commit message describes |
| this branch | `SCHEDULE-STATUS=1.2`, and the `.ics` is in the recipient's scheduling Inbox |

The whole `Caldav` suite, master's tests against both builds after the rebase:

| build | result |
| --- | --- |
| master | 109 run, 1 failure — `Caldav.sched_deliver_virtdomains`, i.e. this bug |
| this branch | `OK (109 tests)` |

So the only test failing on master is the one this patch fixes, and nothing else moves.

One note on the added magic: I called it `:DefaultDomain` and hardcoded `example.com` to match the `calendar_user_address_set` that `Caldav.pm` already sets. If you would rather it took the domain as a parameter, or were folded into `:VirtDomains`, say which and I will change it.

FathiBenNasr:
Nudge, three weeks on.

Still mergeable against current master, and `mergeable_state` is `blocked` only
for want of a review. The Cassandane test was corrected in place after I found
it did not exercise the bug as first submitted — I would rather have said so
than let it merge looking like it proved something.

What it fixes, briefly: scheduling to a local resource resolves the attendee to
a non-canonical userid, so the reply lands in the wrong place. One code file
plus the test.

Happy to rebase or to narrow it if the shape is the problem, and equally happy
to close it if you would rather not carry it.

PR #6265, opened 2026-08-20
guesstz: fix the symlinked_alias flake, make database builds deterministic

brong:
`guesstz:symlinked_alias` intermittently fails `valgrind-make-check` on master
  (run 32261786152). Fixed root cause and two related determinism issues over 3 commits.

  ### Behaviour change

  Commits 2 and 3 change database contents. `cyr_guesstz` output differs, and some
  guesses return a different rules-equivalent TZID: canonical names instead of
  aliases, and the same answer regardless of filesystem.

### DIFF
--- a/cunit/guesstz.testc
+++ b/cunit/guesstz.testc
@@ -269,6 +269,34 @@ static void test_symlinked_alias(void)
     free(alias);
 }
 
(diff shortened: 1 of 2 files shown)


rsto (approved):
Great. Already had a fix for this on my radar, but let's better get this merged rather than waiting for my additional fixes.

PR #6266, opened 2026-08-20
Email/query: add `category` sort comparator

rsto:
This adds a new non-standard sort comparator to Email/query. It requires the "https://cyrusimap.org/ns/jmap/mail" capability. The comparator property name is "category" and it allows to sort the query results by one or more categories, where each category is expressed as a query filter. Each Email sorts by the first filter it matches, and the response reports groupByCounts with one count per filter.

The filter is independent of the Email/query 'filter' argument. A query must have at most one 'category' sort comparator and that comparator must be the first comparator. The canCalculateChanges response always is false for queries with categories.

This is an example for an Email/query that uses the comparator:

```
{
    "filter": { "inMailbox": "P7k" },
    "sort": [
        {
            "property": "category",
            "isAscending": true,
            "groupBy": [
                { "hasKeyword": "$flagged" },
                { "notKeyword": "$seen" }
            ]
        },
        {
            "property": "receivedAt",
            "isAscending": false
        }
    ]
}
```

Internally, Cyrus implements this using guidsearch: for every email that either uidsearch or guidsearch Email/query found, it evaluates the category filter by examining the G records for that email.

Not all filter conditions are supported. The `emailQueryGroupByConditions` capability lists the supported filter conditions. The `maxCategoriesPerEmailQuery` capability defines the maximum supported size of the "groupBy" property of the 'category' comparator.

This PR also aligns the semantics of keyword search between uidsearch and guidsearch Email/query and with Email/get.

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -3387,6 +3387,7 @@ enum guidsearch_expr_op {
     GSEOP_FLAGS,
     GSEOP_CONVFLAGS,
     GSEOP_ALLCONVFLAGS,
+    GSEOP_INTERNALDATE,
     GSEOP_AND,
     GSEOP_OR,
     GSEOP_NOT
@@ -3399,6 +3400,10 @@ union guidsearch_expr_value {
         uint32_t num;
         uint8_t include_trash;
     } convflag;
+    struct {
+        int64_t t;   // seconds since epoch
+        uint8_t op;  // one of SEOP_LT, SEOP_LE, SEOP_GT, SEOP_GE
+    } internaldate;
 };
 
 struct guidsearch_expr {
@@ -3529,6 +3534,23 @@ static struct guidsearch_expr *guidsearch_expr_build(struct conversations_state 
                 }
             }
             break;
+        case SEOP_LT:
+        case SEOP_LE:
+        case SEOP_GT:
+        case SEOP_GE:
+            {
+                if (e->attr == search_attr_find("internaldate")) {
+                    // before, after
+                    ge = xzmalloc(sizeof(struct guidsearch_expr));
+                    ge->op = GSEOP_INTERNALDATE;
+                    ge->v.internaldate.t = (int64_t) e->value.t;
+                    ge->v.internaldate.op = e->op;
+                }
+                else {
+                    assert(0); // guidsearch_rank_clause must reject this
+                }
+            }
+            break;
         case SEOP_MATCH:
             {
                 if (e->attr == search_attr_find("folder")) {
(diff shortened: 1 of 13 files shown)


rsto:
Tagging this as "Do not merge" for now, this might get iterated on for a while.

rsto:
@rjbs @ksmurchison as discussed, this would be great to get in an "approved for testing" state

imap/jmap_mail.c
```diff
@@ -4176,6 +4280,85 @@ static int guidsearch_rank_clause(struct conversations_state *cstate,
...
+     * its conversations state, or NULL for the conditions that hold for any
+     * account. */
+
+    /* hasKeyword and notKeyword read the system flags of the G records of a
+     * guid, and the seen state of the mailboxes holding them */
+    json_t *jkeywords = json_array();
+    json_array_append_new(jkeywords, json_string("$draft"));
+    json_array_append_new(jkeywords, json_string("$flagged"));
+    json_array_append_new(jkeywords, json_string("$answered"));
+    json_array_append_new(jkeywords, json_string("$seen"));
```
ksmurchison, line 4321: Not a huge deal, but why not just use

`json_t *jkeywords = json_pack("[s,s,s,s]", "$draft, ...`

imap/jmap_mail.c
```diff
@@ -4176,6 +4280,85 @@ static int guidsearch_rank_clause(struct conversations_state *cstate,
...
+        }
+        buf_free(&buf);
+    }
+
+    json_t *jconditions = json_object();
+    json_object_set_new(jconditions, "inMailbox", json_true());
+    json_object_set_new(jconditions, "inMailboxOtherThan", json_true());
+    json_object_set_new(jconditions, "before", json_true());
+    json_object_set_new(jconditions, "after", json_true());
+
```
ksmurchison, line 4349: Again, why not use json_pack()?

ksmurchison (approved):
This looks sane.  I guess the UI testing will be the proof

imap/jmap_mail.c
rsto: I tend to avoid json_pack, both because I find the indirection to build a value awkward but mostly because at least AFAIK it shifts validating that the format string and the arguments agree to runtime. I'll check again if my last argument stands.

rsto:
Small update to this PR: we decided to not add the new `maxCategoriesPerEmailQuery` and `emailQueryGroupByConditions` fields to the JMAP account capabilities. I removed the code that added them. Now awaiting CI and the this will get merged.

PR #6268, opened 2026-08-20
cassandane HttpH2: start of an http/2 test suite

rjbs:


wolfsage (changes requested):
1. Will this Just Work if cyrus is configured with TLS? I recall having to play games with the websocket support to work with/without TLS...
2. IMO this should log requests/responses with specific verbosity levels like JMAP/IMAP clients do to make debugging failures easier etc etc

elliefm:
It specifically requests the `http` service, which is a plaintext service.  If you do want TLS services you use the `:TLS` magic, which arranges for `imaps` and optionally `https` services to exist, which you can then use.  But this test doesn't use the magic, so it won't have TLS services, and it doesn't try to use them anyway.

You _can_ trick cassandane into enabling TLS on its plaintext services by setting various tls parameters in the `[config]` section of cassandane.ini.  But this will obviously break anything that expects the plaintext services to actually be plaintext, so don't do that expecting things to work.

PR #6270, opened 2026-08-21
imapd: decode SEARCH strings as UTF-8 for UTF8=ACCEPT and IMAP4rev2

rsto:
This is a follow-up to https://github.com/cyrusimap/cyrus-imapd/pull/6159:

Non-ASCII bytes were ignored in the search terms of all string-valued search keys, such as SUBJECT, FROM, HEADER, BODY and TEXT, if the client enabled UTF8=ACCEPT or IMAP4rev2 rather than passing the CHARSET argument. Now, these clients default to UTF-8 rather than US-ASCII.

The code comment now references the correct RFC 9755, rather than RFC 6855.

### DIFF
--- a/cassandane/tiny-tests/SearchFuzzy/search_utf8_value
+++ b/cassandane/tiny-tests/SearchFuzzy/search_utf8_value
@@ -1,0 +1,74 @@
+#!perl
+use Cassandane::Tiny;
+
+sub assert_search_utf8_value($self, %params)
+{
+    my $talk = $self->{store}->get_client();
+
+    xlog $self, "Create and index test message";
+    $self->make_message("=?utf-8?Q?Gr=C3=BC=C3=9Fe?=") || die;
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):
LGTM.  Should we have @arnt give it a look?

rsto:
> Should we have @arnt give it a look?

The former behavior was definitely wrong, so I'll merge this. I might check with him when we get back his IDNA pull request.

arnt:
Oh, this looks great. Nice scope, clear improvement: Perfect.

arnt:
... and isn't this most likely the reason that my Cassandane tests blew up?

Well, they might anyway. Many lines of cuttapasta ⇒ many lines of improperly understood code.

rsto:
> ... and isn't this most likely the reason that my Cassandane tests blew up?

Yeah, plausible. I definitely will take a look at your original PR again, it's in my to-do list. Will add a reminder to get back to it before next IETF.

PR #6272, opened 2026-08-21
search_sort_thread_maxtime: avoid CPU busywork with a better config

nwc10:
Set search_maxtime to (almost) 0 to cause even the most trivial searches to time out. Hence we can remove the extra terms previously used to hit pathological cases in the DNF normalisation code intended to burn CPU and cross a 1 second timeout.


nwc10:
Second iteration - now with this:

```
-magic(SearchMaxtime1Sec => sub {
-    shift->config_set(search_maxtime => 1);
+magic(SearchMaxtime => sub($self, $seconds) {
+    shift->config_set(search_maxtime => $seconds);
```

and hence overall now `3 files changed, 9 insertions(+), 7 deletions(-)` (previously +9 -4)

nwc10:
rebased on (moved) master

rjbs (approved):

PR #6274, opened 2026-08-21
sphinx + make: stop racing when in parallel

rjbs:
Running make in parallel sometimes hit bugs in building the docs.  The problem was that we'd run three sphinx-build processes in parallel, sharing one cache.  sphinx-build is not having it.  The cache contains pickle files that are not safely shared.  It _is_ okay to share the cache between _serialized_ runs, but not _parallel_.

So, each target gets a lock with a little "with-lock" wrapper.  Easy!

Unfortunately, by that point I was down the rabbit hold:
* docsrc/Makefile still exists, needed a fix, I took an easy way
* while we were in here, why did we still forbid sphinx from working in parallel?
* turns out it's mostly because of GitPython + gitstamp
* …which we *never published* output from, so removed!
* and there was a race in "make -j doc" to fix
* and a typo in clean-doc-doxygen
* and a DOS batch file(!!!) from ages ago

I also spent a bunch of time looking into the `%.1` rules and eventually decided I was way too deep and should file this PR and move on.

…but _then_ I noticed that the "make warnings fatal" in our docs CI was not effective, and I wanted to get rid of all of the warnings (#6275) and in doing *that* I realized that our Doxygen prereqs were wrong, and *also* that I'd broken `distcheck`, both of which are now *also* fixed on this branch.

I am hoping that this is now DONE.


rjbs:
(did one last rebase to get this atop "always maintainer mode")

elliefm (commented):
This looks okay, but I think claude has led you down a garden path into making this more complicated than it needs to be.  You got pretty close to finding the clean solution once you started cleaning up the `man-force` etc rules, but didn't follow that thread deeper.

At the point you're at now, I think everything the with-lock script solves could be solved just by rewriting the doc rule to explicitly serialise its prerequisites.  Like:

```
doc:
    $(MAKE) $(AM_MAKEFLAGS) man
    $(MAKE) $(AM_MAKEFLAGS) doc-html
    $(MAKE) $(AM_MAKEFLAGS) doc-text
    $(MAKE) $(AM_MAKEFLAGS) doc-examples
```

It really ought to have been like this all along, and the fact that it wasn't was my bad.  You can see from the way I used the `man-force` etc targets that I thought they _were_  serialised.  They aren't, but they didn't break back then, so I didn't realise I was wrong.  I now know better, but haven't revisited the docs build since.

Makefile.am
```diff
@@ -2048,9 +2047,9 @@ SPHINX_OPTS = -d $(SPHINX_CACHE) -n -q $(SPHINX_FAIL_ON_WARNINGS) -j auto
 ## The builders below share one doctree cache, which helps: later builds can
 ## reuse the cached parse from earlier ones.  You can't run these at the same
 ## time, though, because only one sphinx-build can own the cache for writing.
-## We'll lock.  Serializing the builders is still a win compared to using three
+## So, every caller (doc, dist-hook) will build them with recursive make and
+## in series. Serializing the builders is still a win compared to using three
```
elliefm: "recursive make" usually refers to a project structure where subdirectories have their own makefiles, and `make target` at the top level proceeds by invoking `make target` on each subdirectory (and if there are sub-subdirectories with their own Makefiles, it recurses).  This isn't that.

Given that the line this comment applies to is gone, I think the whole comment should go, otherwise it's left floating without context, which adds rather than resolves confusion.

Makefile.am
```diff
@@ -2159,18 +2152,20 @@ clean-doc-examples:
...
-## instance.  Depending on them would start three sub-makes, each trashing and
-## rebuilding docsrc/.sphinx-build.stamp on its own schedule.  In a VPATH
-## build, the stamp's recipe extracts a tar of docsrc into the build tree, so
-## two could extract it while a third sphinx-build reads the result.
+## "doc" is man + doc-html + doc-text + doc-examples, but each in its own
+## recursive make, one after another, rather than as four prerequisites.
+## We can't run the sphinx-builds in parallel because they fight over cache.
+## We trash the stamps here rather than letting the entry points each do it,
+## so that docsrc/.sphinx-build.stamp -- whose recipe, in a VPATH build,
+## extracts a tar of docsrc into the build tree -- is remade once per run.
```
elliefm: Again, this isn't "recursive make", and this comment is a lot of prose for something trivial.

The fact we're calling `$(MAKE) foo` instead of just listing foo as a prerequisite paints the whole picture: these need to be serialised in this order.  I don't think a comment is completely necessary here, but two details that might be worth capturing:

* the fact that we're doing it this way deliberately, and not just because we're rookies who don't know how to use normal prerequisites
* whether or not the _order_ is significant, beyond the mere fact of them being serialised

I don't think the order _is_ significant, at least not at the moment.  Generally, whichever of these we build first will build the cache and take a while, and the rest will be lightning fast because the cache was already built.  There might be room for microoptimisations in the ordering, and maybe we want to be clear whether we have or haven't done that.

So a comment along the lines of "these must be serialised, though the order is not important" would suffice, I think

Makefile.am
```diff
@@ -2159,18 +2152,20 @@ clean-doc-examples:
...
 .PHONY: doc clean-doc
 doc:
 	@$(RM) docsrc/.sphinx-build.stamp man/.sphinx-build.stamp \
 	    doc/html/.sphinx-build.stamp doc/text/.sphinx-build.stamp
-	$(MAKE) $(AM_MAKEFLAGS) $(SPHINX_MANS) doc/html/.sphinx-build.stamp \
-	    doc/text/.sphinx-build.stamp $(GENERATED_EXAMPLES)
+	$(MAKE) $(AM_MAKEFLAGS) $(SPHINX_MANS)
+	$(MAKE) $(AM_MAKEFLAGS) doc/html/.sphinx-build.stamp
+	$(MAKE) $(AM_MAKEFLAGS) doc/text/.sphinx-build.stamp
+	$(MAKE) $(AM_MAKEFLAGS) doc-examples
```
elliefm, line 2159: This would be a lot easier to just read if it directly named the targets it's building.  Don't refer to the stamp or man files here, they're implementation details; instead, build `doc-text`, `doc-html`, and `man` in the same way you've done `doc-examples`.

The pattern here is that `make doc-specific` builds one specific kind of documentation, and `make doc` builds the interesting set of specific kinds.  `man` can be thought of as a special-case spelling of `doc-man` in this context.

Makefile.am
```diff
@@ -2243,7 +2238,15 @@ distgit:
...
-dist-hook: $(DIST_DOC_STAMPS)
+## The HTML and text docs aren't in DISTFILES: they're built here and copied
+## in, after distdir has finished.  One recursive make each, because they can't
+## be built at the same time -- and as separate targets rather than the forcing
+## doc-html and doc-text, so that a tree that has already built them doesn't
+## build them again.
+dist-hook:
+	$(MAKE) $(AM_MAKEFLAGS) doc/html/.sphinx-build.stamp
+	$(MAKE) $(AM_MAKEFLAGS) doc/text/.sphinx-build.stamp
+
```
elliefm, line 2240: I'm confused about the shape of the `dist-hook:` change in this PR.  I didn't notice it earlier, but now that there's a comment on it, it's jumping out at me.  `DISTFILES` isn't ours; it's part of the Makefile.in generated by automake, we don't need to think about it.  Is claude lying to you about how this works?

On `master`, the `dist-hook:` target _removes_ all the stamp files and sphinx cache, to prevent them being included in the distribution tarball.  But in this PR, it rebuilds the docs, redundantly??

IIRC, `make dist(check)` automatically invokes the `doc` target, which is how our docs get built for distribution.  The reason the `doc` target expands to `man`, `doc-html`, `doc-text`, `doc-examples` is because these are the formats we want to ship in the tarball.  The entire `doc/` tree is included in the distribution due to being listed in EXTRA_DIST.  But our doc build process leaves cruft behind in `doc/` (stamp files, the sphinx cache), and so we have `dist-hook:` blow those bits away after the docs are built, before the tarball is assembled.  Then we get a tarball with good docs and no crap.

I expect if you `make distcheck` on this branch and then examine the tarball with like `tar tfz cyrus-whatever.tgz | sort | less`, you will find a lot of crap in it that we don't want to distribute.

elliefm (changes requested):
I failed to notice the `dist-hook:` fiasco when I looked at this earlier, so I'm glad I didn't approve it then!

Makefile.am
elliefm: The reason the `man` target is `man` and not `doc-man` is because a) that's the autotools convention, and b) the man pages wind up in `man/`, not `doc/man/`.

Makefile.am
rjbs: I actually already did that exercise.  If you compare what you get from building the dist on master versus this branch, this branch is much better.

It removes a bunch of ton of doxygen files in the new build.  That's about it.  (I see some pyc files removed, but I think that's because I needed to clean more aggressively.)

I'll have another go at it, a bit more aggressively with the cleaning, but for now I don't see the symptom you describe.  Assuming I can't produce the problem you foresaw, what is the practical issue here?

I don't know what you're wondering, re: Claude.  Did Claude say something wrong to me about how this works?  Maybe, it's been a while.  Or maybe I misunderstood something in the manual.  Or maybe I just connected the dots wrong.  This is a complex system that is foreign to me, I'm bound to stuff it up once in a while.

rjbs: Okay, I did two `dar distcheck --brief` on two cleaned checkouts, which is basically this:

```
autoreconf -i s
./configure --enable-maintainer-mode
make distcheck -j 8
```

I get:
* `cyrus-imapd-3.13.7-537-g8716c1e290.tar.gz` - this branch
* `cyrus-imapd-3.13.7-552-gcc2f4cf3ae.tar.gz` - current master

I run:
```
diff -u \
  <(tar ztf cyrus-imapd-3.13.7-552-gcc2f4cf3ae.tar.gz | cut -d / -f 2- | sort) \
  <(tar ztf cyrus-imapd-3.13.7-537-g8716c1e290.tar.gz | cut -d / -f 2- | sort)
```

(That `cut` is there to remove the top-level directory, which differs between builds.)

There are 0 lines beginning with `+` other than the diff header.

Looking for deletions -- that is, files distributed from master but not this branch, I pipe that to `grep '^-' | cut -d/ -f1-3 | sort | uniq -c` and get:

```
   1 -cassandane/Cassandane/Cyrus
   1 -cassandane/tiny-tests/Caldav
   1 -cassandane/tiny-tests/Conversations
   1 -cassandane/tiny-tests/Create
   5 -cassandane/tiny-tests/HTTP
   1 -doc/examples/imapd_conf
   1 -doc/html/_sources
   1 -doc/html/support
   1 -doc/text/support
   1 -docsrc/_doxygen/
 549 -docsrc/_doxygen/xml
   6 -docsrc/exts/sphinxlocal
   1 -docsrc/make.bat
   1 -docsrc/support/security-classification.md
```

So, a bunch of tests added to master but not yet here, then, the removal of stuff we want removed: some pyc in sphinxlocal (gitstamp), then all the doxygen.

Makefile.am
rjbs: Here was my thinking:

* if srcdir != builddir, we cpio (effectively, with tar) docsrc to the target to make the docsrc-level stamp
* doc-text and doc-html and man clear the stamp file
* so removing it at the top of the rule and then in each build we dispatch to felt redundant
* …and I think I thought we'd then rebuild again when doing the tar dance

But I have sat here and tried to carefully work out the dependency graph in my head, and I think this was bunk.  I have a nagging suspicion that there is something to this, but I can't prove it, so I'm gonna go with "rjbs is jumping at shadows".

rjbs: …but then, why bother rm-ing the stamps at all?  They will be removed by the makes inside the doc rule. But you didn't remove them in your sample change.  Intentional for reasons I don't see, or just omitted?

Makefile.am
rjbs: I _did_ fix the loss of doc/examples/imapd_conf/imapd.conf.sample

Makefile.am
rjbs: The comment applies to two things:  the `doc` and `dist-hook`, saying why they don't just depend on the other rules like they used to.  This is this "prevent obvious simplification that is actually a trap" kind of comment, and there are *two* obvious simplifications:
1. just make doc depend on other targets
2. use distinct caches per thing

Both have been tested and found wanting, but we already know that somebody will try to switch us to #2 because both Bron and I did.

It may be more effective to duplicate the comment in both places, or to put it in one and point from the other.  I will do that second one for now, as it appears to be suggested as a solution by your comment near the definition of "doc".

Makefile.am
rjbs: Done.

Makefile.am
rjbs: I went with "just omitted" and removed the RM, and I believe this is all still coherent, and few lines of Make, so great.

Makefile.am
rjbs: > "recursive make" usually refers to a project structure where subdirectories have their own makefiles, and make target at the top level proceeds by invoking make target on each subdirectory (and if there are sub-subdirectories with their own Makefiles, it recurses). This isn't that.

I was led astray by the `make` manual, I guess, which says:
> §5.7 Recursive Use of make
> 
> Recursive use of make means using make as a command in a makefile.

I'm gonna guess that "yeah but everybody uses it in a more specific way".  I have replaced all the places I introduced the term to say sub-make or subprocess.

rjbs:
I clicked "request re-review" but then did more editing, most of which should not be particularly consequential.  The main outstanding point seems to be the changes to dist-hook, which fall out of the EXTRA_DIST confusion with `doc`

I can't see any *practical* negative impact of that change, or the obvious way to minimize what's being done in the dist hook.

elliefm (approved):

PR #6278, opened 2026-08-22
automake: always be in maintainer mode

rjbs:
❓ Is this a good idea?  I'm not sure.  I *think* so, having suffered through sieve.c not rebuilding, and from reading the automake manual.  But I'm not sure.

Short version:
* unless you're in maintainer mode, edits to X.y won't cause X.c to rebuild
* the problems maintainer mode is meant to solve don't _seem_ relevant to us, given my read
* we had some prereq errors
* and a typo

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -125,6 +125,20 @@ AC_PROG_INSTALL
 AC_USE_SYSTEM_EXTENSIONS
 AC_ISC_POSIX
 AC_PROG_AWK
+
+dnl Every build needs lex and yacc, not only Sieve-enabled ones: lib/cron-lex.l
+dnl and lib/cron-parse.y are part of libcyrus_min.
+AC_PROG_YACC
+AM_PROG_LEX
+
+if test -z "$ac_cv_prog_YACC"; then
+    AC_MSG_ERROR([Cyrus requires bison or byacc, but neither is installed])
+fi
+
+if test -z "$ac_cv_prog_LEX"; then
+    AC_MSG_ERROR([Cyrus requires flex or lex, but neither is installed])
+fi
+
 AC_SYS_LONG_FILE_NAMES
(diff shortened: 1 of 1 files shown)


elliefm:
There was a reason we didn't just do this in the past, but I don't currently remember what it was.

It might have ramifications for building from release tarballs.  Please test that thoroughly.

elliefm:
If we do make this default, then `--enable-maintainer-mode` can be removed from the release process docs.

rjbs:
> It might have ramifications for building from release tarballs. Please test that thoroughly.

Will do!

rjbs:
@elliefm I think this is good to go.  I started to remove the docs about --enable-maintainer-mode but then thought, "Is somebody gonna read _today's_ docs for releasing 3.12.x?"  If so, better to leave it.  Harmless, too.

configure.ac
```diff
@@ -125,6 +125,20 @@ AC_PROG_INSTALL
 AC_USE_SYSTEM_EXTENSIONS
 AC_ISC_POSIX
 AC_PROG_AWK
+
+dnl Every build needs lex and yacc, not only Sieve-enabled ones: lib/cron-lex.l
+dnl and lib/cron-parse.y are part of libcyrus_min.
+AC_PROG_YACC
+AM_PROG_LEX
```
elliefm: This comment is only meaningful if you know that sieve uses lex/yacc, and believe that _only_ sieve uses them.  Which means it only makes sense right now, when that's the state we're moving away from.

The commit history documents the fact it changed; the fact that every build needs lex and yacc is self contained in their being requested unconditionally.  Anyone who needs a comment to explain this isn't going to be helped by this comment.

elliefm (approved):
One nit about a zero-value comment that should be removed, but I don't need to review it again after that's fixed, so approving now.

PR #6280, opened 2026-08-24
configure.ac: Remove SIZE_T_FMT - C99 provides %zu for values of type size_t

nwc10:
No need to probe for a format string for *printf for values type size_t, as C99 provides the size modifier 'z'.

The analogous probes for time_t, off_t and rlim_t are still needed, as C has not standardised these.

Remove the macro strtosizet() which was never used, as its definition is part of the stanza that calculated SIZE_T_FMT.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -274,12 +274,6 @@ AC_DEFINE_UNQUOTED([TIME_T_FMT], ["$cyr_cv_format_time_t"],
 AC_DEFINE_UNQUOTED([strtotimet(a,b,c)], [$cyr_cv_parse_time_t(a,b,c)],
                    [strtol-like parse function for time_t])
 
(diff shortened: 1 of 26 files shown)


nwc10:
@elliefm - this was the PR that I mentioned on Monday, that seems to have got lost during the security tsunami.

I'm not blocking on this, so don't fret if you're still juggling umpteen other tasks.

elliefm (approved):
This seems fine, though there's merge conflicts atm.  Approving now, because I don't need to review it again as long as CI still passes after the conflicts are resolved.

My vague recollection is that `z` is not portable, but I bet it's only "not portable" because the old MSVC libc doesn't properly support C99.  But since we require C11 nowadays (and POSIX!), that's not of any concern for cyrus-imapd.

PR #6281, opened 2026-08-24
http_caldav.c: remove support for the PATCH method

ksmurchison:
No known CalDAV client uses PATCH, and the VPATCH format was never standardized.

### DIFF
--- a/changes/next/cyr-2749-remove-caldav-patch
+++ b/changes/next/cyr-2749-remove-caldav-patch
@@ -1,0 +1,26 @@
+
+Description:
+
+Support for the HTTP PATCH method on the CalDAV endpoint has been removed.
+
+No known CalDAV client uses PATCH, and the VPATCH format was never standardized.
+
+
+Documentation:
+
+None.
+
+
+Config changes:
+
+None.
+
+
+Upgrade instructions:
+
+None.
+
+
+GitHub issue:
+
+None.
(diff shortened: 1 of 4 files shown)


rjbs (approved):

rsto:
Nothing from blocking this PR, but why not also remove PATCH for DAV (and any other HTTP endpoint in Cyrus)?

ksmurchison:
The upcoming HTTP resumable upload spec uses PATCH so we may reuse some of the existing code to support that.

rsto:
> The upcoming HTTP resumable upload spec uses PATCH so we may reuse some of the existing code to support that.

Understood. Is that going to be part of DAV, though?

ksmurchison:
I think it could apply to any resource. An CalDAV event with large embedded attachment. A CardDAV contact with a large photo. I haven't really thought about the implementation yet. But I supposed we could strip out meth_patch() now and add it back in somewhere if/when needed.

rsto:
> I think it could apply to any resource. An CalDAV event with large embedded attachment. A CardDAV contact with a large photo. I haven't really thought about the implementation yet. But I supposed we could strip out meth_patch() now and add it back in somewhere if/when needed.

I just ask because I happened to notice that `imap/http_dav.c` still includes a bunch of lines with ALLOW_PATCH and I couldn't figure out why we would want that for now. But my question does not imply that I want this PR to remove that, in fact, I look forward to the current PR landing ASAP because I can then remove PATCH handling in the "privacy" branch.

rsto (approved):
Happy to see this go

PR #6283, opened 2026-08-24
http_client: simplify whitespace trimming in auth parameter parsing

rjbs:
The trimming in http_parse_auth_params() was doing more work than it
needed to, and being less careful about it.  Skip runs of separators and
whitespace with strspn(), stop once the parameter list is exhausted, and
treat a parameter whose value is empty as absent: a token is 1*tchar, so
zero length is not allowed.

Also add tests for SCRAM Authorization, which was previously untested.


ksmurchison (approved):

elliefm:
Need to add `Authen::SCRAM::Client` to `cassandane/doc/README.deps`

PR #6287, opened 2026-08-25
enable building C++ code with clang++

nwc10:
Currently we always build the C++ code with `g++`, even if we're building the C code with `clang`

Fix the two issues that stand in the way of using `clang++`

1. The C++ compiler probe fails for `clang++` under `-Werror` - add `-Wno-error` for that probe
2. util.h contains some `__attribute__` directives to facilitate Cunit testing of `clock_gettime`. `g++` is fine with parsing these, but `clang++` chokes if it sees them

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -515,7 +515,14 @@ if test "x$enable_xapian" != xno ; then
         fi
 
         dnl Xapian requires CXX17
+        dnl this probe breaks under -Werror for clang++, so work around that
+        saved_CFLAGS="$CFLAGS"
+        saved_CPPFLAGS="$CPPFLAGS"
+        CFLAGS="$CFLAGS -Wno-error"
+        CPPFLAGS="$CPPFLAGS -Wno-error"
         AX_CXX_COMPILE_STDCXX(17)
+        CFLAGS="$saved_CFLAGS"
+        CPPFLAGS="$saved_CPPFLAGS"
 
         dnl If AC_PROG_LIBTOOL (or the deprecated older version AM_PROG_LIBTOOL)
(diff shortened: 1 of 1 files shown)


dilyanpalauzov:
https://github.com/cyrusimap/cyrus-imapd/pull/6030 takes a different approach for compiling with `clang++` in fact it goes over two distinct approaches, and as always: in case of doubt, I am always wrong.

nwc10:
> #6030 takes a different approach for compiling with `clang++` in fact it goes over two distinct approaches, and as always: in case of doubt, I am always wrong.

@elliefm - I'm tempted to take #6030 over my commit. So should we

1. complete #6030 and merge it
2. drop the `util.h` commit from this branch
3. review/iterate on just the `configure.ac` commit from this branch

?

rjbs:
ellie will do better at this conversation than I, removed self

configure.ac
```diff
@@ -515,7 +515,14 @@ if test "x$enable_xapian" != xno ; then
         fi
 
         dnl Xapian requires CXX17
+        dnl this probe breaks under -Werror for clang++, so work around that
```
elliefm, line 518: For whatever it's worth, I don't think the comment is strictly necessary.  Temporarily turning off `-Werror` for autoconf probes that are known to fail due to irrelevant warnings is routine.

elliefm (approved):
I had forgotten about Dilyan's earlier PR, my bad.  I've now approved and merged it, so you'll need to rebase.

This looks fine (once you resolve the new util.h conflicts), I don't need to look at it again unless you make some drastic other changes, so I'm approving now.

PR #6289, opened 2026-08-25
Sqldb user data ownership

brong:
This is PR one of four -- this should stop all the DBERROR syslogs for sqlite issues, which seem to happen mostly around renames and deletes, but also has some sync_reset safety that's quite essential, and might mean we don't need to reconstruct users after a rename.

### DIFF
--- a/lib/sqldb.c
+++ b/lib/sqldb.c
@@ -54,9 +54,46 @@ static int _free_open(sqldb_t *open)
 {
     int rc = sqlite3_close(open->db);
     free(open->fname);
+    free(open->owner);
     free(open);
     int r = (rc == SQLITE_OK ? SQLDB_OK : SQLDB_ERR_UNKNOWN);
     return r;
+}
+
+/* record which file we opened, for _isunmoved() to compare against.  A zero
+   ino means we couldn't tell, and we skip the check for that handle */
+static void _stampfile(sqldb_t *open)
+{
+    struct stat sbuf;
+
+    if (!*open->fname || !strcmp(open->fname, ":memory:")) return;
+
+    if (stat(open->fname, &sbuf)) {
+        xsyslog(LOG_NOTICE, "couldn't stat sqldb we just opened",
+                            "fname=<%s>", open->fname);
+        return;
+    }
+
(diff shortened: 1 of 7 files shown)


lib/sqldb.c
```diff
@@ -73,22 +110,49 @@ static int _version_cb(void *rock, int ncol, char **vals, char **names __attribu
...
+        if (!_isunmoved(open)) {
+            xsyslog(LOG_NOTICE, "sqldb file replaced while open, reopening",
+                                "fname=<%s>", open->fname);
+            open->stale = 1;
+            continue;
+        }
+
+        /* disagreement here means somebody is taking the wrong lock */
+        if (open->scope != scope || strcmpsafe(open->owner, owner)) {
+            xsyslog(LOG_ERR, "DBERROR: sqldb already open with a different scope",
```
wolfsage, line 140: This is a warning but we don't take any distinct action?

lib/sqldb.c
```diff
@@ -73,22 +110,49 @@ static int _version_cb(void *rock, int ncol, char **vals, char **names __attribu
...
+
+        /* already open! */
+        open->refcount++;
+        return open;
     }
 
     open = xzmalloc(sizeof(sqldb_t));
     open->fname = xstrdup(fname);
+    open->scope = scope;
+    open->owner = xstrdupnull(owner);
```
wolfsage, line 155: Nothing ever passes in an owner (except the cunit tests) but I guess this is split out from bigger work and the code that uses that is in https://github.com/cyrusimap/cyrus-imapd/pull/6292/files

A little weird since there's other code here that checks owner, but okay!

wolfsage (approved):

lib/sqldb.c
brong: It doesn't impact correctness or lock safety within sqlite, so this allows us to see bogus locking ordering which might have a deadlock race inherent in it, without breaking anything that's not perfect now!

lib/sqldb.c
brong: Yes, that's correct - this is setting up things for later use.

PR #6290, opened 2026-08-25
mailbox: make relocking safe, and notice when the mailbox moved

brong:
This is PR 2 of four, Claude wanted to use relock and I said "I don't trust it" so it went and looked and found that it was indeed untrustworthy!  This fixes that.

### DIFF
--- a/cunit/mailbox.testc
+++ b/cunit/mailbox.testc
@@ -5,10 +5,27 @@
 #include <stdalign.h>
 #endif
 
+#include <errno.h>
 #include <stdlib.h>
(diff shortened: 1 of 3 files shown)


ksmurchison (commented):
This looks sane, but do we have any tests that can verify the changes?

brong:
> This looks sane, but do we have any tests that can verify the changes?

Not at the Cassandane level because it's all race conditions, but now there's some cunit tests to exercise the failure cases.

ksmurchison (approved):

PR #6292, opened 2026-08-25
Cross user lock ordering

brong:
This is PR has now been split into 4!  This remaining part is the core lock ordering fixes - each place where we might lock a second or further user while still holding a user lock we do one of two changes:
1) ensure we lock all users up-front so we lock them in the right order
2) unlock the first user before locking another one

### DIFF
--- a/cassandane/tiny-tests/JMAPCalendars/calendarevent_participantreply_crossuser_concurrent
+++ b/cassandane/tiny-tests/JMAPCalendars/calendarevent_participantreply_crossuser_concurrent
@@ -1,0 +1,122 @@
+#!perl
+use Cassandane::Tiny;
+use POSIX qw(_exit);
+use Fcntl qw(:flock);
+use File::Temp qw(tempfile);
+
+# Two local users each organize an event the other attends, then RSVP on
+# the other's behalf at the same time.  Each call holds its own account's
+# lock while sched_deliver_local() writes into the other user's calendar,
+# so this is the shape in which a lock taken in the wrong order deadlocks.
+sub test_calendarevent_participantreply_crossuser_concurrent
+    :NoAltNameSpace
+    ($self)
+{
+    my $cassandane = $self->default_user;
+    my $alice = $self->{instance}->create_user('alice');
+
+    my %eventid;
+    for my $pair ([$cassandane, $alice], [$alice, $cassandane]) {
+        my ($organizer, $attendee) = @$pair;
+        my $caldav = $organizer->caldav;
+        my $calid = $caldav->NewCalendar({name => 'invites'});
+        $self->assert_not_null($calid);
+
+        my $uid = "crossuser-" . $organizer->username;
+        my $orgaddr = $organizer->username . '@example.com';
+        my $attaddr = $attendee->username . '@example.com';
+        my $card = <<EOF2;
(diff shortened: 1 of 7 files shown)


rsto (commented):
This looks to have a few bugs, I am going to paste Claude output verbatim. I did review this before you pushed the new commit about participantReply, but the points still look legit with that additional commit:

  HIGH

  1. imap/jmap_calendar.c:10236 — the new user_nslock leaks on the error return. Confirmed:
  the lock is taken at 10182, released only at 10188 and at done: (10324), but if (r) return;
  at 10236 bypasses done:. r there comes from a failed mailbox_open_irl() in the foreach
  callback. The leaked user_nslock_t keeps its mboxlock in open_mboxlocks for the process
  lifetime — blocking every other process's exclusive lock on that user, and the next
  user_nslock_lock(principalid, LOCK_EXCLUSIVE) in this httpd hits the locktype mismatch →
  IMAP_MAILBOX_LOCKED → assert(locktype == LOCK_NONBLOCKING) → abort. Same return already
  leaked db. Fix: goto done;.

  2. imap/imapd.c:8254 — releasing the rename locks there makes the respond: error fixups
  assert. Confirmed: the release sits above respond:, and the if (r) branch calls
  mboxlist_update_full(olddestmbentry) / mboxlist_delete(newdestmbentry), both of which reach
  assert_namespacelocked(name). Path: admin RENAME user.a user.b with allowusermoves, rename
  fails, fixup runs with user.b's namelock gone → abort (NDEBUG is not set in this build).
  The release belongs after the fixup block, or the fixups need their own lock.

  MEDIUM

  3. imap/imapd.c — the five goto respond statements (8016, 8049, 8066, 8074, 8084) skip the
  release, so imapd_check(NULL, 0) still runs holding both users' namelocks on those paths —
  including 8049, the successful INBOX-rename path, and the three in the admin user-rename
  branch where lockdouble() genuinely holds two users. Only the fall-through path is fixed.

  4. imap/jmap_calendar.c:8844 — the premise that sched_deliver_local() takes the users'
  locks doesn't hold. No user_nslock* anywhere in imap/itip_support.c;
  caldav_store_resource() takes none either — delivery relies on per-mailbox namelocks plus
  the index write lock. The only user lock reachable is via caldav_scheddefault(..., 1) →
  _create_mailbox(), and only when the recipient has no default calendar. Since the organizer
  is req->accountid here, the partstat write into the account's own calendar is now the only
  recipient has no default calendar. Since the organizer is req->accountid here, the partstat
  write into the account's own calendar is now the only JMAP_READ_WRITE path touching the
  account with no user nslock held, so a concurrent user_deletedata/rename/sync_reset can
  interleave. Related: LOCK_SHARED at 8576 lets two participantReply calls read the same
  oldical; sched_deliver_local re-reads under its own lock so partstats survive, but
  schedule_one_attendee can mail a REQUEST built from a stale copy.

brong:
My Claude went through and pointed out that the lock order by user is NOT the same as lock order by lock name, so the testing machinery was actually bogus. It's all been stripped, plus tons more removed.

What's left is just two clear cases of releasing locks before taking other users' locks.

rsto (approved):
Cool, this looks much cleaner.

PR #6293, opened 2026-08-26
testrunner.pl: option to make plan check failures non-fatal

elliefm:
Ordinarily, you can't run any tests unless `$plan->check_sanity()` succeeds.  But if there's a problem, sometimes you want to know what else is broken, without being forced to fix the first problem first.  This PR adds an option to let tests run even if the sanity check complained.


elliefm:
To test this, `mkdir cassandane/tiny-tests/Foo` will set up a situation that the sanity checker will complain about.  You can then run testrunner.pl with and without the `--no-fatal-plan` option to observe the difference.

wolfsage (approved):

PR #6294, opened 2026-08-26
httpd: escape calendar display-name in CalDAV GET

rsto:
This makes CalDAV GET properly escape display names in HTTP headers. It also changes httpd to not emit any header containing control characters in its name or value.

### DIFF
--- a/imap/httpd.c
+++ b/imap/httpd.c
@@ -3339,7 +3339,7 @@ EXPORTED void response_header(long code, struct transaction_t *txn)
             /* Construct Content-Disposition header */
             char *encfname = NULL;
             for (const unsigned char *p = (unsigned char *)resp_body->dispo.fname; p && *p; p++) {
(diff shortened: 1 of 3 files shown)


imap/httpd.c
```diff
@@ -3339,7 +3360,7 @@ EXPORTED void response_header(long code, struct transaction_t *txn)
             /* Construct Content-Disposition header */
             char *encfname = NULL;
             for (const unsigned char *p = (unsigned char *)resp_body->dispo.fname; p && *p; p++) {
-                if (*p < 0x20 || *p >= 0x7f) {
+                if (*p < ' ' || *p >= 127) {
                     encfname = charset_encode_mimexvalue(resp_body->dispo.fname, NULL);
```
ksmurchison, line 3364: Why are we using hex values in the first commit but using decimal and char values here?

ksmurchison (approved):

PR #6295, opened 2026-08-26
jmap_calendar.c: only support participantReply for organizer

rsto:
This makes the participantReply method reject the participant reply unless all of the following conditions are met:

1. the ORGANIZER of the event must match one of the scheduling addresses of the account to which the calendar of the event belongs
2. the authenticated user of the JMAP request either must be the owner of the calendar or must have the JMAP CalendarRight "mayWriteAll".

Attendees or non-attending sharees can not change the participation status.


ksmurchison (approved):
This looks good.

ksmurchison (approved):

PR #6300, opened 2026-08-26
cassandane: add the CalendarEvent test entity datatype

rsto:
This adds a test entity for calendar events. This originally was part of https://github.com/cyrusimap/cyrus-imapd/pull/6250, now with minor changes (e.g. uses the `default` function to determine the default calendar id, rather than using the first calendar in the list).


wolfsage (approved):
Great!

PR #6303, opened 2026-08-28
cass gdb hints: Escape '!' in path names so bash is happy

wolfsage:
When cass detects a core file, it provides a hint for how to run gdb against it. The core files usually look like this:

    .../cores/core.!usr!cyrus!libexec!httpd.12345.6.789

bash interprets '!' specially, trying to use cassandane's gdb hint usually produces an error like:

    bash: !usr!cyrus!libexec!httpd.12345.6.789: event not found

If we escape it then you can copy/paste it as is and everyone's happy.

### DIFF
--- a/cassandane/Cassandane/Instance.pm
+++ b/cassandane/Cassandane/Instance.pm
@@ -1747,6 +1747,7 @@ sub find_cores
         if (defined $prog) {
            xlog "   from program $prog";
            my ($bin) = $prog =~ m/^(\S+)/; # binary only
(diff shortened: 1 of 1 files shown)


elliefm (approved):
The change seems fine, approved.

The blanket suggestion to run gdb with sudo seems wild to me, but that wasn't your doing.  IMO you should run gdb as yourself, and if it hasn't got permissions for something, fix the permissions.  Running it under sudo should be the last resort, not the default.

PR #6305, opened 2026-08-28
universal logfmt: phase 1

rjbs:


rjbs:
@rsto I believe this is ready for review, but not merge, simply because I need to prepare log-monitoring software for new names.  Should be no big deal.

btw, I sent you this because I sent you phase 0;  probably I will send phase 2 to murch, to spread the load?

rsto (approved):
This LGTM, except that it has no changes file and I'm not sure if it needs one. Technically, renaming logfmt keys would require it, but I guess at this stage we don't bother?

rjbs:
Yeah, I think everything logfmt is 3.13, so I think the bar for needing a changesfile is higher.

PR #6308, opened 2026-08-29
Enforce valid names for mailbox components

brong:
We don't have strict enough mailbox name creation logic for JMAP (or DAV) to stop the users creating names which don't round trip to the same name.  This fixes it!

### DIFF
--- a/cassandane/tiny-tests/Create/bad_name_chars
+++ b/cassandane/tiny-tests/Create/bad_name_chars
@@ -2,7 +2,7 @@
 use Cassandane::Tiny;
 
 # With '.' as the hierarchy separator, '^' is how a '.' in a name is spelled
-# externally, and '/' has no special meaning at all.
+# externally, and '/' is not permitted.
 sub test_bad_name_chars ($self)
 {
     my $talk = $self->{store}->get_client();
--- a/cassandane/tiny-tests/Create/bad_name_chars
+++ b/cassandane/tiny-tests/Create/bad_name_chars
@@ -1,0 +1,58 @@
+#!perl
+use Cassandane::Tiny;
+
+# With '.' as the hierarchy separator, '^' is how a '.' in a name is spelled
+# externally, and '/' has no special meaning at all.
+sub test_bad_name_chars ($self)
+{
+    my $talk = $self->{store}->get_client();
+
+    my @bad = ('yes/no', '10% Happier');
+
+    $talk->create('owner');
+    $self->assert_str_equals('ok', $talk->get_last_completion_response());
+
+    foreach my $m (@bad) {
+        $talk->create($m);
+        $self->assert_str_equals('no', $talk->get_last_completion_response());
+
+        $talk->rename('owner', $m);
+        $self->assert_str_equals('no', $talk->get_last_completion_response());
+    }
+
+    $talk->create('a^b');
+    $self->assert_str_equals('ok', $talk->get_last_completion_response());
+
+    my $list = $talk->list('', '%');
+    $self->assert_deep_equals(['INBOX', 'a^b', 'owner'],
+                              [sort map { $_->[2] } @$list]);
+}
+
+# With unixhierarchysep the two swap over: '/' is the separator, and '^' is
+# the character that can't appear at all.
+sub test_bad_name_chars_unixhs
+    :UnixHierarchySep
+    ($self)
+{
+    my $talk = $self->{store}->get_client();
+
+    my @bad = ('a^b', '10% Happier');
+
+    $talk->create('owner');
+    $self->assert_str_equals('ok', $talk->get_last_completion_response());
+
+    foreach my $m (@bad) {
+        $talk->create($m);
+        $self->assert_str_equals('no', $talk->get_last_completion_response());
+
+        $talk->rename('owner', $m);
+        $self->assert_str_equals('no', $talk->get_last_completion_response());
+    }
+
+    $talk->create('a.b');
+    $self->assert_str_equals('ok', $talk->get_last_completion_response());
+
+    my $list = $talk->list('', '%');
+    $self->assert_deep_equals(['INBOX', 'a.b', 'owner'],
+                              [sort map { $_->[2] } @$list]);
+}
--- a/changes/next/cyr-3225-valid-names
+++ b/changes/next/cyr-3225-valid-names
@@ -1,0 +1,27 @@
+Description:
+
+JMAP and DAV now reject mailbox and collection names they cannot represent,
+rather than silently creating a mailbox by another name: JMAP returns
+invalidProperties for "name", and MKCOL/MKCALENDAR (and COPY/MOVE to such a
+destination) return 403.  Any mailboxes created this way previously still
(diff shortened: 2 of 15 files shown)


brong:
Pulling the review request for now. I want more tests to ensure we haven’t broken IMAP name creation in weird namespaces

brong:
> Pulling the review request for now. I want more tests to ensure we haven’t broken IMAP name creation in weird namespaces

I looked into that - this doesn't run on the IMAP creation codepath, so it's fine.

cunit/mboxname.testc
```diff
@@ -1056,6 +1057,27 @@ static void test_isnondeliverymailbox(void)
...
+    /* '.' is stored as '^', so a literal '^' would read back as '.' */
+    CU_ASSERT_EQUAL(IMAP_MAILBOX_BADNAME,
+                    mboxname_policycheck_component("a^b"));
+    CU_ASSERT_EQUAL(IMAP_MAILBOX_BADNAME,
+                    mboxname_policycheck_component("^"));
+
+    CU_ASSERT_EQUAL(IMAP_MAILBOX_BADNAME, mboxname_policycheck_component(""));
+    CU_ASSERT_EQUAL(IMAP_MAILBOX_BADNAME, mboxname_policycheck_component(NULL));
+
+    /* characters mboxname_policycheck() rejects are its business, not ours:
```
rsto: I find this comment confusing: it mentions mboxname_policycheck but doesn't call it. How does it come into play here?

imap/mboxname.c
```diff
@@ -1763,6 +1763,37 @@ EXPORTED int mboxname_same_userid(const char *name1, const char *name2)
...
+ * A component is legal exactly when it survives the round trip into an
+ * internal name and back out: an unrepresentable component would name a
+ * different mailbox than the caller asked for.  Today that means rejecting
+ * '^', which is how _append_intbuf() spells '.'; deriving it from the
+ * encoding rather than naming the character keeps the two in step.
+ */
+EXPORTED int mboxname_policycheck_component(const char *name)
+{
+    struct buf buf = BUF_INITIALIZER;
+    strarray_t *boxes = strarray_new();
```
rsto: nit: this could also save heap allocation by setting `strarray_t boxes = STRARRAY_INITIALIZER`. not that it would save much in the grand scheme of things, though.

imap/mboxname.h
```diff
@@ -288,6 +288,17 @@ char *mboxname_lockpath_suffix(const char *mboxname, const char *suffix);
  */
 int mboxname_policycheck(const char *name);
 
+/*
+ * Return 0 if 'name' is usable as a single component of a mailbox name.
+ *
+ * mboxname_policycheck() sees only the assembled internal name, by which
```
rsto: This whole paragraph mixes implementation internals with API contract - and the implementation details might change in the future but this function still exist. I suggest to only document the API contract, which looks to be "This function checks if external and internal representations of the name components are equal. It does not check if the name component is valid, use mboxname_policycheck() for that."

Also, the commit message for this change also seems mostly to describe the internals that changed, whereas that commit really seems to about adding a new function that provides this API contract.

imap/jmap_mailbox.c
```diff
@@ -1646,6 +1646,42 @@ static char *_mbox_newname(const char *name, const char *parentname, int is_topl
     return mboxname;
 }
 
+/* A Mailbox/set name is only honoured if it survives the whole trip into the
```
rsto: This whole comment looks to be a duplicate of the one at the call site of `_mbox_name_roundtrips`. I suggest to remove this one and only keep the one in `_mboxset_args_parse`

imap/http_dav.c
```diff
@@ -736,6 +736,15 @@ HIDDEN int calcarddav_parse_path(const char *path,
     if (tgt->collen) {
         collection = freeme = xstrndup(tgt->collection, tgt->collen);
 
+        /* Only creation has to be representable: on a lookup, '^' is how a
```
rsto: nit: I'd just remove this comment, there's no need to retell the "^" vs. "." story here and in `mboxname_policycheck_component` and at the other call sites.

imap/http_webdav.c
```diff
@@ -465,6 +465,17 @@ static int webdav_parse_path(const char *path, struct request_target_t *tgt,
     while (len) {
         /* Get collection(s) */
         char *val = xstrndup(p, len);
+
+        /* Only creation has to be representable: on a lookup the last
```
rsto: nit: I'd remove this comment, too.

cassandane/tiny-tests/Create/bad_name_chars
```diff
@@ -0,0 +1,63 @@
...
+# Cassandane runs without unixhierarchysep, so '.' is the separator and '/'
+# is just a character in an external name -- but no internal name may
+# contain one.
+sub test_bad_name_chars ($self)
+{
+    my $talk = $self->{store}->get_client();
+
+    my @bad = ('yes/no', '10% Happier');
+
+    $talk->create('victim');
```
rsto: nit: it's not written in stone, but we tend to not use loaded terms like "victim", "attacker", etc. They suggest a severity that almost never is warranted and neutral terms like "owner", "sharer", "sharee" typically are more to the point.

changes/next/cyr-3225-valid-names
```diff
@@ -0,0 +1,32 @@
...
+
+
+Config changes:
+
+None.
+
+
+Upgrade instructions:
+
+None.
```
rsto, line 33: The `None` here followed by a paragraph that concludes that indeed no upgrades are necessary confused me. Could this whole paragraph be condensed in a single additional sentence in the Description, e.g. "Any mailboxes created previously still are reachable by their original name."?

imap/mboxname.c
```diff
@@ -1763,6 +1763,37 @@ EXPORTED int mboxname_same_userid(const char *name1, const char *name2)
     return r;
 }
 
+/*
```
rsto, line 1804: This comment differs from the comment on the declaration of this function. I suggest not to do that, if the functions is EXPORTED and documented, that documentation should only be in the header.

rsto (approved):
I am approving this because the logic is sound. I find the prose too verbose and at points overly "agentic", though, so I'd be happy to see that revised.

cunit/mboxname.testc
brong: The argument there was "we don't need to reject things which are going to be rejected later by policy check". - I called bullshit on that and got Claude to factor out the logic and check it here too.  Each component of the name needs to be valid (including being sound mod-UTF7) to be acceptable.

imap/mboxname.c
brong: This was completely rewritten

imap/mboxname.h
brong: Yep, replaced and commit updated.

cassandane/tiny-tests/Create/bad_name_chars
brong: Changed

PR #6311, opened 2026-08-30
attachextract: don't extract text from zero-byte attachments

brong:
An attachment whose decoded content is empty has nothing to extract, but we sent it to the extractor anyway, which rejected it and left an IOERROR in the log for every such body part:

  IOERROR index: can't extract attachment
    da39a3ee5e6b4b0d3255bfef95601890afd80709 (APPLICATION/PDF):
    Item does not exist

Bail out early when the part's content guid is the SHA1 of zero bytes, before any cache lookup or HTTP request.  Matching on the guid rather than the raw data length also catches encoded parts, since the content guid hashes the decoded content.

### DIFF
--- a/cassandane/tiny-tests/SearchFuzzy/squatter_attachextract_empty_part
+++ b/cassandane/tiny-tests/SearchFuzzy/squatter_attachextract_empty_part
@@ -1,0 +1,46 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_squatter_attachextract_empty_part
+    :SearchAttachmentExtractor
+    ($self)
+{
+    my $instance = $self->{instance};
+    my $imap = $self->{store}->get_client();
+
+    my $tracedir = tempdir(DIR => $instance->{basedir} . "/tmp");
+    $self->start_echo_extractor(tracedir => $tracedir);
+
+    xlog "Append email with a zero-byte attachment";
+    $self->make_message("msg1",
+        mime_type => "multipart/related",
+        mime_boundary => "123456789abcdef",
+        body => ""
+        ."\r\n--123456789abcdef\r\n"
+        ."Content-Type: text/plain\r\n"
+        ."\r\n"
+        ."bodyterm"
+        ."\r\n--123456789abcdef\r\n"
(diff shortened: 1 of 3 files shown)


cassandane/tiny-tests/SearchFuzzy/squatter_attachextract_empty_part
```diff
@@ -0,0 +1,46 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_squatter_attachextract_empty_part
+    :min_version_3_9 :SearchAttachmentExtractor
```
rsto: no more `min_version` annotations, please

imap/attachextract.c
```diff
@@ -370,6 +380,16 @@ EXPORTED int attachextract_extract(const struct attachextract_record *axrec,
         return 0;
     }
 
+    if (is_empty_content(&axrec->guid)) {
+        /* there's no text to extract from a zero-byte attachment, and
+         * the extractor will just reject it anyway */
+        xsyslog(LOG_DEBUG, "ignoring zero-byte attachment",
```
rsto: I'm not entirely sure if we only want new code to use `xyslog_ev` or if we require it. In any case, rewriting this to use `xyslog_ev` would be great. It's documented in file "doc/README.logfmt.md".

rsto (approved):
The change is correct, but two code conventions aren't met. I'm going to approve this, but please address at least the `min_version` remark and possible also the one about logfmt.

cassandane/tiny-tests/SearchFuzzy/squatter_attachextract_empty_part
brong: Dropped, thanks.

imap/attachextract.c
brong: Converted to `xsyslog_ev(LOG_DEBUG, "search.attachextract.empty", ...)` with `msg.guid`/`part.type`/`part.subtype`, matching the existing `search.attachextract.uncached` event a few lines further down.  All three keys were already registered.

PR #6314, opened 2026-08-30
Add tests for backupcyrusd

brong:
This not only adds the tests, but also fixes all the perl module compilation and testability, there was a YAML which isn't in the dar container but wasn't being used, and a missing brace in sieveshell as well.

This also removes some dead code from the backup protocol, bumping the version to 6.

### DIFF
--- a/cassandane/Cassandane/Cyrus/FastMail.pm
+++ b/cassandane/Cassandane/Cyrus/FastMail.pm
@@ -11,6 +11,9 @@ use Net::CardDAVTalk 0.11;
 use Text::JSContact 0.01 qw(vcard_to_jscontact);
 use Data::Dumper;
 use Storable 'dclone';
+use Cyrus::Backup;
+use Cyrus::Backup::Restore;
+use Cyrus::Backup::State;
 
 use base qw(Cassandane::Cyrus::TestCase Cassandane::Mixin::QuotaHelper);
 use Cassandane::Util::Log;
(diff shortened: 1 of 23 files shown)


perl/imap/lib/Cyrus/Backup.pm
```diff
@@ -255,7 +255,7 @@ sub BackupUser {
   # remove stale data files :)
   if (opendir(DH, $DataDir)) {
     while (my $item = readdir(DH)) {
-      next if ($item eq 'backupstate.sqlite3' and $DataDir eq $Metadir);
+      next if ($item eq 'backupstate.sqlite3' and $DataDir eq $MetaDir);
       next if $item eq $NewName;
```
rjbs, line 261: briefly lost my mind wondering why you'd fix this but not add strict — but it happens later in the branch. :-D

cassandane/tiny-tests/FastMail/backupcyrusd_compress
```diff
@@ -0,0 +1,104 @@
+#!perl
+use Cassandane::Tiny;
+
+use Cyrus::Backup;
+use Cyrus::Backup::Restore;
+
+sub test_backupcyrusd_compress
+    :min_version_3_13 :want_service_backupcyrusd
```
rjbs: we've generally agreed that min_version stuff on new tests is bunk, unless they're replication related

not a blocker, but given this stuff will likely want to be purged later, could be purged now ;)

perl/sieve/scripts/sieveshell.pl
```diff
@@ -164,6 +164,7 @@ sub check_sieve_error {
 if (!defined $obj) {
     my $errstr = sieve_get_global_error() // "unknown error";
     die "unable to connect to '$acapserver': $errstr";
+}
```
rjbs, line 167: Oof.  Thanks.

rjbs (approved):
Yeah, looking good!  Is there a subtle bug somewhere?  Maybe, lots of change!  But seems about right, and tests >> no tests!

cassandane/Cassandane/Cyrus/FastMail.pm
```diff
@@ -196,6 +199,224 @@ sub _fmjmap_err
...
+    my $res = $talk->getmetadata($folder, $entry);
+    $self->assert_str_equals('ok', $talk->get_last_completion_response());
+    $self->assert_not_null($res->{$folder}{$entry});
+
+    return $res->{$folder}{$entry};
+}
+
+# Every mailbox cyrus has for a user, as { extname => uniqueid }, which is
+# what a complete backup's folder list should look like.  Only valid with
+# altnamespace off, which is how this suite configures cyrus.
```
rjbs, line 255: Not a blocker, but I think we should update this test suite to use altnamespace, to bring it more in line with Fastmail's default configuration.  I figure I'll deal with this test when we get there.  LMK if you think there's more wrinkles.

cassandane/tiny-tests/FastMail/backupcyrusd_compress
brong: Removed it

cassandane/Cassandane/Cyrus/FastMail.pm
brong: Yeah, we should at some point go clean up FastMail (maybe even lowercase it!)

PR #6315, opened 2026-08-30
Add DKIM2 Message-Instance calculation to Sieve

brong:
This means that If we use EditHeader to change anything which is covered in a Message-Instance signature, we create the recipe to undo it.


imap/dkim2_mi.c
```diff
@@ -0,0 +1,607 @@
...
+        strarray_appendm(fields, xstrndup(p, q - p));
+        p = q;
+    }
+
+    return fields;
+}
+
+/* Lowercased field name of "Name: value", or NULL if there is no colon.
+ * §6.2 deletes the WSP on both sides of the colon, so "Subject : x" has to
+ * come back as the same name as "Subject: x".
```
ksmurchison, line 63: Shouldn't this say "subject:x"

imap/dkim2_mi.c
```diff
@@ -0,0 +1,607 @@
...
+    EVP_DigestInit_ex(ctx, EVP_sha256(), NULL);
+    EVP_DigestUpdate(ctx, body, bodylen);
+    EVP_DigestUpdate(ctx, "\r\n", 2);
+    digest_b64(ctx, b64);
+    EVP_MD_CTX_free(ctx);
+}
+
+/* Compare two canonicalised fields by name alone */
+static int canon_name_cmp(const char *a, const char *b)
+{
```
ksmurchison, line 206: Why doesn't this function use field_name() to canonicalize?

imap/dkim2_mi.c
brong: It could, but the point was more that the trailing space wasn't needed.  I trimmed the comment back to the key bits though instead of putting the example in.

imap/dkim2_mi.c
brong: It doesn't need to because they're already canonicalized.  It's there to compare already canonical names.

ksmurchison (approved):
I only looked at the Sieve changes.  I expect that you have a handle on the DKIM2 stuff

PR #6318, opened 2026-08-30
sync: UNMAILBOX names the mailbox it means

brong:
A mailbox name can be reused, so by the time UNMAILBOX arrived the replica could be holding a different mailbox at that name - a recreate, or a rename onto the freed name - and it deleted that one instead, messages and all.

Carry the UNIQUEID and UIDVALIDITY of the mailbox the master meant, and refuse the delete when the replica finds anything else.  Where the name came from the replica's own folder listing (the sweep for folders which no longer exist on the master, and XFER) this also closes the window between the listing and the delete.

Gated on a new UNMAILBOX-ID capability so a new master still talks to an old replica; a replica accepts both forms, so an old master is unaffected.

### DIFF
--- a/docsrc/developer/replication-protocol.md
+++ b/docsrc/developer/replication-protocol.md
@@ -927,6 +927,7 @@ its capabilities, followed by an OK greeting:
 * STARTTLS
 * COMPRESS DEFLATE
 * SIEVE-MAILBOX
+* UNMAILBOX-ID
 * REPLICATION-ARCHIVE
 * OK servername Cyrus sync server v3.12.0-...
 ```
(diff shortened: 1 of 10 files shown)


docsrc/developer/replication-protocol.md
```diff
@@ -1492,10 +1494,28 @@ multi-tier replication setups.
...
+C: S3 APPLY UNMAILBOX %(MBOXNAME user.cassandane.OldFolder
+       UNIQUEID d4f5a6b7-... UIDVALIDITY 1711200000)
 S: S3 OK Success
 ```
 
+The replica deletes the mailbox only if the one holding that name is
+the one named by UNIQUEID and UIDVALIDITY; otherwise it returns
+`IMAP_MAILBOX_MOVED` and leaves it alone.  A name can be reused, so
+without the identity check a delete of the master's mailbox could take
+out a completely different mailbox which had since taken over the name.
```
elliefm, line 1506: For whatever it's worth, `IMAP_MAILBOX_MOVED` is logged as "Mailbox has been moved to another server", which is clearly about murder and not replication.  We should consider adding a new error code for these cases, instead of reusing the murder error code.

It's come up before on FM slack where people have seen it in the logs and been confused by it.  They don't realise the error string is unrelated to the problem being logged, and start trying to figure out what other server it thinks it was moved to, etc.

elliefm (approved):
That looks pretty complete, thanks.

I have one suggestion, but if you run with it, I think it should be its own PR, not bolted onto this one.

docsrc/developer/replication-protocol.md
brong: Thanks!  I've put that into a Linear task

PR #6320, opened 2026-08-31
Hold off shutting down during write locks

brong:
This avoids getting killed in the middle of deliveries where possible, so we get clean states.

When I wrote this it would have had problems with the sync lock, but now that the fix for that has been merged, this is safe!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -629,7 +629,6 @@ cunit_TESTS = \
     cunit/command.testc \
     cunit/conversations.testc \
     cunit/cron.testc \
-    cunit/signals.testc \
     cunit/css3_color.testc \
     cunit/cyr_qsort_r.testc \
     cunit/crc32.testc \
@@ -666,6 +665,7 @@ cunit_TESTS = \
     cunit/search_expr.testc \
     cunit/seqset.testc \
     cunit/sessionid.testc \
+    cunit/signals.testc \
     cunit/slowio.testc \
     cunit/smallarrayu64.testc \
     cunit/sqldb.testc \
--- a/Makefile.am
+++ b/Makefile.am
@@ -629,6 +629,7 @@ cunit_TESTS = \
     cunit/command.testc \
     cunit/conversations.testc \
     cunit/cron.testc \
+    cunit/signals.testc \
     cunit/css3_color.testc \
     cunit/cyr_qsort_r.testc \
     cunit/crc32.testc \
--- a/cassandane/tiny-tests/Fetch/cache_generation_mismatch
+++ b/cassandane/tiny-tests/Fetch/cache_generation_mismatch
@@ -1,0 +1,47 @@
+#!perl
+use Cassandane::Tiny;
+
+# The first four bytes of a cache file are the generation of the index it was
+# written for.  A repack under an unlocked mailbox leaves us holding offsets
+# into a file that no longer matches, which must be reported as staleness and
+# recovered from the spool file - not parsed as if it were ours and blamed on
+# the checksum.
+sub test_cache_generation_mismatch
+    :NoAltNameSpace
+    ($self)
+{
+    my $imaptalk = $self->{store}->get_client();
+
+    $self->make_message("cache test", body => "hello world\r\n");
+
+    my $mbpath = $self->{instance}->run_mbpath('user.cassandane');
+    my $cache = "$mbpath->{data}/cyrus.cache";
+
+    $imaptalk->select("INBOX");
+
+    open(my $fh, '+<', $cache) || die "open $cache: $!";
+    binmode($fh);
+    sysread($fh, my $generation, 4) == 4 || die "read $cache: $!";
+    sysseek($fh, 0, 0);
+    syswrite($fh, pack('N', 0xdeadbeef)) == 4 || die "write $cache: $!";
+    close($fh);
+
+    $self->{instance}->getsyslog();
+
+    my $res = $imaptalk->fetch('1', '(BODY.PEEK[TEXT])');
+    $self->assert_str_equals('ok', $imaptalk->get_last_completion_response());
+    $self->assert_matches(qr/hello world/, $res->{'1'}->{body});
+
+    my @log = $self->{instance}->getsyslog();
+
+    # put the spool back before anything else looks at it
+    open($fh, '+<', $cache) || die "open $cache: $!";
+    binmode($fh);
+    syswrite($fh, $generation) == 4 || die "write $cache: $!";
+    close($fh);
+
+    $self->assert(scalar(grep { m/cache generation mismatch/ } @log),
+                  "the mismatch was not reported");
+    $self->assert(!scalar(grep { m/invalid cache record/ } @log),
+                  "the mismatch was reported as corruption");
+}
--- a/cassandane/tiny-tests/Fetch/cache_not_created_by_read
+++ b/cassandane/tiny-tests/Fetch/cache_not_created_by_read
@@ -1,0 +1,47 @@
+#!perl
+use Cassandane::Tiny;
+
+# A FETCH takes the index write lock, drops it, and only then streams the
+# bodies - so the cache file is opened with no lock held at all.  It must be
+# opened read-only: without the write lock we have no right to create one, and
+# a repack unlinks any cache file it did not rewrite.
+sub test_cache_not_created_by_read
+    :NoAltNameSpace
+    ($self)
+{
+    my $imaptalk = $self->{store}->get_client();
+
+    $self->make_message("cache test", body => "hello world\r\n");
+
+    my $mbpath = $self->{instance}->run_mbpath('user.cassandane');
+    my $cache = "$mbpath->{data}/cyrus.cache";
+    $self->assert(-f $cache, "$cache should exist to start with");
+
+    $imaptalk->select("INBOX");
+
+    # Not BODY.PEEK: asking to set \Seen is what makes a FETCH take the write
+    # lock, and the lock is what decides how the cache gets opened.  Do it
+    # once first so the message is already seen - then the second one takes
+    # the write lock but never touches the cache while it holds it, and the
+    # only cache access left is the unlocked one we care about.
+    $imaptalk->fetch('1', '(BODY[TEXT])');
+    $self->assert_str_equals('ok', $imaptalk->get_last_completion_response());
+
+    rename($cache, "$cache.saved") || die "rename $cache: $!";
+
+    my $res = $imaptalk->fetch('1', '(BODY[TEXT])');
+    $self->assert_str_equals('ok', $imaptalk->get_last_completion_response());
+    $self->assert_matches(qr/hello world/, $res->{'1'}->{body});
+
+    my $created = -e $cache;
+
+    # put the spool back before anything else looks at it
+    unlink($cache) if $created;
+    rename("$cache.saved", $cache) || die "rename $cache.saved: $!";
+
+    # the missing file is logged, and that is fair enough - drain it so the
+    # instance teardown doesn't call it a failure
+    $self->{instance}->getsyslog();
+
+    $self->assert(!$created, "a read recreated $cache");
+}
--- a/changes/next/cyr-2574-lmtpd-shutdown
+++ b/changes/next/cyr-2574-lmtpd-shutdown
@@ -1,0 +1,28 @@
+Description:
+
+A graceful shutdown now waits for any write transaction in progress to finish,
+so a service signalled mid-delivery can no longer leave a mailbox committed
+without its conversations DB, or an LMTP message delivered but unacknowledged.
+Several places that held a lock while waiting on the network have been fixed so
+that wait is bounded by local work.
+
+
+Documentation:
+
+None; behaviour of existing shutdown handling.
+
+
+Config changes:
+
+None.  Note that the new behaviour relies on the global lock, so it does not
+apply when global_lock is turned off.
+
+
+Upgrade instructions:
+
+None.
+
+
+GitHub issue:
+
+None.
--- a/cunit/signals.testc
+++ b/cunit/signals.testc
@@ -31,7 +31,6 @@ static void test_deferral_lifecycle(void)
     CU_ASSERT_EQUAL(shutdowns, 0);
 
     /* two nested transactions, then the signal arrives */
-    signals_add_handlers(0);
     signals_defer_shutdown();
     signals_defer_shutdown();
     raise(SIGTERM);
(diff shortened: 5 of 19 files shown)


rjbs:
@brong As discussed, here's an agent analysis which I present unedited by mutual agreement. ;-)

----

I read all six commits against the surrounding code. The sigmask handling itself is clean; the problems are in *where* the new shutdown point lands and in the premise that justifies making the deferral unbounded.

## 1. The deferral boundary is one step too early — the headline LMTP case isn't actually fixed

The changelog and `signals.h` both promise "never … an LMTP message delivered but never acknowledged." The deferral ends when the global lock refcount hits zero, which is inside `mailbox_close()`, not when the client has been answered.

In `lmtpd.c` `deliver_mailbox()`: `append_commit()` → `duplicate_mark()` → `mailbox_close(&mailbox)`. That `mailbox_close` reaches `mailbox_unlock_index()` → `mboxname_release(&namelock)` → global lock pops → `signals_resume_shutdown()` → `shut_down()`, which does not return. The `send_lmtp_error()` calls that write the `250` are three lines further down and never run. `shut_down()` does `prot_flush(deliver_out)`, but there is nothing buffered yet.

So the outcome is: message committed, dupelim marked, connection dropped, MTA retries. The dup DB saves you *if* `duplicate` is on and the message has an id — which is exactly the pre-existing safety net, not a new guarantee. Worse, the same thing happens between two recipients of one DATA: recipient 1 delivered, recipient 2 untouched, nothing acked.

The invariant you want isn't "the last write lock closed", it's "the client has been answered." Those coincide in imapd and httpd only by luck — their responses sit in the prot buffer and `shut_down()` flushes it. lmtpd hasn't written its response yet at that point.

If you want the stated guarantee, lmtpd needs its own defer/resume bracketing the whole DATA transaction (defer after `savemsg`, resume after the `send_lmtp_error` loop), with the global-lock hook as the inner nesting.

## 2. "No transaction waits on the wire" is false in a murder

`lib/signals.c:168` deliberately has no time limit, justified in `mboxname.c:125` by "no transaction waits on the wire." `mboxlist_setacl()` disproves it: `user_nslock_lockmb_w(name)` at `imap/mboxlist.c:3394`, `mupdate_connect()`/`mupdate_activate()` at 3538–3552, `user_nslock_release()` at 3566. Same shape in `mboxlist_deletemailbox`, `mboxlist_renamemailbox` (3108–3124), `mboxlist_setquota`. That's a TCP connect plus a round trip to the mupdate master, entirely inside the deferral.

And it compounds badly with the swallowed signal. mupdate goes through prot, so the reply read lands in `signals_select()` → `signals_poll_mask()`. While deferred, that branch logs and returns 0, then `pselect()` sleeps — and no further wakeup is coming, because the signal was already consumed into `gotsignal[]` and `SA_RESETHAND` means the handler is gone. A frontend whose mupdate master is wedged is now unkillable by SIGTERM until the prot read timeout expires, where before it died promptly.

`cyr_withlock_run` (`mboxname_run_with_lock`, `cyr_withlock_run.c:98`) is the other one: global lock held exclusive across an arbitrary external command. Probably intended, but it's an unbounded non-local wait under the same rule.

## 3. `mboxevent_flush_notifications()` is a blocking round trip in the now-unprotected window

`imap/notify.c:79–101` — the `dlist:` transport is SOCK_STREAM with a synchronous `dlist_parse()` reply read over prot, i.e. `signals_select()`. In `lmtpengine.c:1144` you've placed that call after the locks are gone and before the `250`. So the widened gap in §1 now also contains a network read to notifyd, and a pending SIGTERM will fire *inside* it.

It also means a flush interrupted partway drops the rest of the queue for that call. The delayed action is still registered so `shut_down()`'s `libcyrus_run_delayed()` re-drains it — but only for the direct-call sites; worth checking you're happy with that being the recovery path.

## 4. Shutdown now fires from deep inside the mailbox layer

`signals_resume_shutdown()` runs `shutdown_cb()` from inside `remove_lockitem()`, and after this PR that's reachable from `mailbox_unlock_index()`, which is called from `mailbox_close()` and from `mailbox_relock()`. The ordering inside `remove_lockitem()` is right (unlink, free, *then* resume), and the `signals_in_shutdown` guard correctly stops re-entry when `shut_down()` re-takes the global lock. So the mechanics hold.

What doesn't hold is the assumption that a zero global-lock refcount means "between transactions":

- **`mailbox_close()` (`mailbox.c:1297`)** calls `mailbox_unlock_index()` while the mailbox is still on `open_mailboxes` and before `remove_listitem()`. `shut_down()` starts with `libcyrus_run_delayed()`, and `_delayed_cleanup()` will `mailbox_open_iwl()` the very mailbox that's half-closed. It survives on refcounting, but it's a new and unadvertised state.
- **`mailbox_relock()` (`mailbox.c:910`)** starts with `mailbox_unlock_index()`. The refcount hits zero *in the middle of a lock upgrade*. Any caller upgrading shared→exclusive partway through a write sequence gets torn apart at precisely the point the deferral exists to protect.
- **`mailbox_close()` on a deleted mailbox** relocks and then calls `mailbox_delete_cleanup()`; the relock's unlock can exit first, leaving the mailbox flagged deleted with files unreaped.

None of these are fatal today (only because every `shutdown_cb` exits), but the design contract "refcount zero ⇒ safe point" isn't true, and `mailbox_relock` is the sharp edge.

## 5. `mailbox_lock_index()` doesn't always re-take the namelock

The comment at `mailbox.c:2555` says the struct "stays usable, and `mailbox_lock_index()` re-takes both locks via `mailbox_relock()` if we come back for more." That's only true when the user nslock has *also* been dropped. At `mailbox.c:2444`, `need_relock` is set only if `!user_nslock_islocked(userid)`. If an outer caller holds the user nslock, `need_relock` is 0 and you go straight to `mailbox_lock_index_internal()` — index locked, `mailbox->namelock == NULL`.

Correctness is still fine (the outer user nslock excludes rename/delete/repack), but `open_mailboxes_namelocked()` (`mailbox.c:210`) will now answer 0 for that mailbox, and its only consumers are the `assert()`s in `user.c:736,772,773,777` that exist to catch exactly this class of lock-ordering bug. You've quietly disarmed them.

## 6. The acknowledged cost of dropping the namelock is understated

The comment says the failure mode is `mailbox_map_record()` → ENOENT → `IMAP_NO_MSGGONE`. But `mailbox_unlock_index()` also `mappedfile_close()`s every cache file before releasing the namelock, so every subsequent cache read reopens *by name*. A concurrent repack renames a new `cyrus.cache` into place; `cache_getfile()` (`mailbox.c:578`) does not validate `generation_no` on the read path, so you parse a stale offset against a fresh file. `mailbox_cacherecord_internal()` catches it via the CRC (`mailbox.c:724`) — so no garbage reaches the client — but the result is `IMAP_MAILBOX_CHECKSUM` plus `IOERROR invalid cache record` in syslog, which is not "the message is gone," it's the signature operators page on for real corruption. Either widen the comment or make that path log something distinguishable.

## 7. Services with no flush and no delayed-action drain

`mboxevent_flush_notifications()` is wired into imapd, httpd and lmtpengine. `nntpd` calls neither it nor `libcyrus_run_delayed()` anywhere, so a POST/IHAVE's `MessageAppend` events now sit in the queue until `libcyrus_done()` at connection teardown. For a long-lived NNTP feed that's an unbounded delay and unbounded memory. `timsieved` is in the same position (probably harmless, it generates few events, but worth confirming).

## Sigmasking specifically

This part I think is right. The deferred branch returning without restoring `oldmaskp` is fine — `signals_select()` restores unconditionally at the end, and the only reason the non-deferred branch restores early is so `shutdown_cb` runs unmasked. The `case SIGINT/SIGQUIT/SIGTERM: break;` in the reporting loop is genuinely unreachable when not deferred, as the comment claims. `gotsignal[]` stays `volatile sig_atomic_t` and `signals_deferred` is only touched from normal context.

Two notes:

- `signals_resume_shutdown()` calls bare `signals_poll()`, so `shutdown_cb` runs under whatever mask is current. No path reaches `remove_lockitem()` from inside `signals_select()`'s blocked window today, so it's latent, not live — but `signals_poll_mask(NULL)` is the one caller that can't restore, and it's now reachable from arbitrary depth.
- The `SA_RESETHAND` escape hatch works, but it's the *only* bound on the deferral, and it's indiscriminate: given §2, a legitimate operator sending a second TERM to a process wedged on mupdate kills it mid-transaction, which is the exact outcome the PR exists to prevent.

## Nits

- `Makefile.am`: `signals.testc` is inserted between `cron.testc` and `css3_color.testc`, breaking the sort.
- `cunit/signals.testc`: the second `signals_add_handlers(0)` is redundant — nothing has raised a signal yet, so `SA_RESETHAND` hasn't fired. The test also leaves `gotsignal[SIGINT]`/`[SIGQUIT]` unclear'd, and `signals_deferral_logged`/`signals_in_shutdown` dirty for any later test in the same binary.

rjbs:
@brong We talked very briefly.  I am okay to approve this if you discount the agent feedback, much of which seemed plausibly real to me.

brong:
@rjbs Same deal on this side — worked through it with an agent, verified each claim against the tree myself. Four commits pushed. Verdict item by item:

**§1 — right, and it's the one that mattered.** The deferral does end in `mailbox_close()`, three lines before lmtpd writes its 250, and `shut_down()`'s `prot_flush()` has nothing to flush at that point. So the headline guarantee wasn't being delivered. Fixed in `a9ab325` by bracketing the whole DATA transaction in `lmtpengine.c` — defer before `func->deliver()`, resume after the `send_lmtp_error()` loop and an explicit `prot_flush(pout)`. The per-transaction deferral in `mboxname.c` nests inside it, and it also covers the gap between two recipients of one DATA.

**§2 — right, the premise is false.** `mboxlist_setacl()` holds a namespacelock across `mupdate_connect()`/`mupdate_activate()`, same shape in delete/rename/setquota, plus `cyr_withlock_run()` across an arbitrary command. Comment fixed in `ecbeec5` to say what's actually true, including that `SA_RESETHAND` is a last resort rather than a bound. A real time limit on the deferral is worth doing but it's separate work, not this PR.

**§6 — right, and worse than the comment said.** `cache_getfile()` doesn't check `generation_no` on the read path, so a post-repack re-read dies at the CRC as `IMAP_MAILBOX_CHECKSUM` + `IOERROR invalid cache record`, which looks like corruption to whoever's reading syslog. Comment widened.

**§5 — right about the mechanism.** `need_relock` is only set when `!user_nslock_islocked()`, so with an outer nslock the namelock does stay gone and `open_mailboxes_namelocked()` answers 0. Narrower than it reads, though: those `user.c` asserts are themselves only evaluated in the branch where the user *isn't* already locked. Comment corrected either way.

**§7 — half right, and the half that's right isn't nntpd.** nntpd was removed in d8d2b3a (there's no `imap/nntpd.c` in the tree; that'll have been stale `.o` files in a build dir). timsieved generates no mboxevents at all — its only "quota" check is a script count, not `quota_db`. The service that actually had the problem is **pop3d**, which reached the queue only through the delayed backstop, putting its notifications after the response instead of before. Fixed in `667f9d9`.

**§3 — mechanism right, severity low.** It's a local unix socket, and 61260a1 already hardens notifyd against the hangup. The "interrupted flush drops the rest of the queue" part isn't so: entries are popped one at a time, so a shutdown mid-flush leaves the remainder queued and `shut_down()`'s `libcyrus_run_delayed()` drains them. With §1 fixed the whole call now sits inside the protected window anyway.

**§4 — accurate as description, but the sharp edge isn't sharp.** `mailbox_relock()` on a shared→exclusive upgrade can't drop the refcount to zero: a shared `user_nslock` never takes `$GLOBAL` in the first place (`mboxname.c` only grabs it for `LOCK_EXCLUSIVE`), so there's nothing to pop. The `mailbox_rename()` call site sits under `mboxlist_renamemailbox()`'s outer nslock, which holds the count up. That leaves `mailbox_close()`'s `OPT_MAILBOX_DELETED` relock with no outer nslock, where the worst case is a deleted mailbox left unreaped for `cyr_expire`. Leaving it.

**Nits** — both real, fixed in `0079ea4`. (The test only ever raises SIGTERM, not SIGINT/SIGQUIT, but it did leave that one set.)

Tested: `make check-signals` passes; `Delivery Pop3 Sieve MboxEvent Notify` is 165 run / 3 failed, and those three (`Sieve.badscript_sievec`, `Sieve.badscript_timsieved`, `Sieve.enotify_text_variable_bad`) fail identically with these changes stashed, so they're pre-existing on the base.

brong:
Follow-up on §6. Rather than just widening the comment about the cache, I went looking at whether we could verify the generation — and found two things.

**The check was always meant to exist.** `mailbox_repack_commit()`:

```c
/* rename index first - loader will handle un-renamed cache if
 * the generation is lower */
```

No loader ever did. So besides the unlocked-read case, a repack that dies between renaming `cyrus.index` and renaming the cache files leaves new offsets pointing into the old cache, and today that's parsed as if it were ours and caught only by the CRC.

The invariant turns out to be clean, which is what makes the check possible: `generation_no` is only bumped when `doing_cache` is set, and in that case `mailbox_repack_commit()` unlinks any cache file it didn't rewrite. So the stamp in a surviving cache file — live or archive — is exactly "the index this belongs to". No stale-archive-cache false positives.

**And a real bug underneath it.** `mailbox_cachefile()` decided read-only from `mailbox->is_readonly`, which is set once in `mailbox_open_index()` from the *requested* locktype and never revisited, plus `index_locktype == LOCK_SHARED`. After `mailbox_unlock_index()` sets `index_locktype = 0`, that whole expression is false — so every unlocked read opened `cyrus.cache` `O_RDWR | MAPPEDFILE_CREATE` with no lock at all. Combined with repack unlinking unused cache files, an unlocked read that finds one missing **creates a fresh empty `cyrus.cache` stamped with its own stale generation**.

It's reachable from a plain FETCH: `index_fetch()` unlocks at index.c:1237 and streams bodies at 1252, and a non-peek FETCH of an already-`\Seen` message takes the write lock, no-ops in `_fetch_setseen()`, drops the lock, and then opens the cache read-write with nothing held. There's a Cassandane test for exactly that.

`mailbox_get_cstate_full()` had the identical idiom and the identical exposure — `mailbox_unlock_index()` drops the cstate, so a caller after the unlock reopens conversations.db read-write with no lock.

Two commits:

- `0bb6f77` — ask `mailbox_index_islocked(mailbox, /*write*/1)` instead. `mailbox_append_cache()` already asserts the same lock, so the legitimate write path is unaffected.
- `3370614` — compare the cache's stamped generation to `mailbox->i.generation_no` before parsing, with `IMAP_MAILBOX_REPACKED` and a single `LOG_NOTICE` in place of the three IOERRORs ending in "invalid cache record".

Both have a test that is red on the parent commit and green after, verified by stashing each fix in turn.

**Correction to my last comment:** I said `Sieve.badscript_sievec`, `Sieve.badscript_timsieved` and `Sieve.enotify_text_variable_bad` were pre-existing failures on the base. They weren't failures at all — they were an artifact of a stale shared build dir. On a fresh container the whole suite is green: **2828 tests, 0 failures**.

rjbs (approved):

PR #6321, opened 2026-08-31
check for xzfree side effects at compile time

elliefm:
The `xzfree()` macro needs to evaluate its argument twice, in order to free it and then zero it.  This is dangerous when the argument has side effects (consider `xzfree(x++);`).  We have a crude form of protection that results in an assertion failure if the argument doesn't compare equal to itself, but a) this is a run-time check, which is not ideal, and b) it can't detect all kinds of side effects.

This PR updates the `xzfree()` macro to use `_Static_assert` for protection instead of `assert`.  That means misuses of `xzfree()` get detected at compile time instead of run time.

`_Static_assert` requires that its argument be a constant expression, so an interesting quirk here is that an argument with side effects causes the macro to generate invalid syntax.  That is, the `_Static_assert` fails to compile, and therefore isn't actually evaluated.  It's a compile-time error, but not the one you might have expected.  Hopefully it's not too confusing; details below the fold.

With this change, the xmalloc.h header no longer needs to drag in assert.h.  This is the change I was actually aiming for!  I had to fix one file that was using `assert()` without including assert.h.

Turns out clang doesn't properly support the syntax this relies on, even though it's valid C11, so when clang is detected, the `_Static_assert` line is left out.  More on this in a comment below, and in source comments.

---

If you misuse xzfree like this:

```
void *p;
xzfree(p++);
```

you'll get an error like this:

```
lib/xmalloc.h:29:26: error: expression in static assertion is not constant
   29 |     _Static_assert((ptr) == (ptr), "xzfree argument has side effects"); \
      |                    ~~~~~~^~~~~~~~
lib/cyrusdb_twom.c:41:5: note: in expansion of macro 'xzfree'
   41 |     xzfree(p++);
      |     ^~~~~~
```

If you misuse it like this:

```
    extern void *foo(void);
    xzfree(foo());
```

you'll get an error like this:

```
lib/xmalloc.h:29:26: error: expression in static assertion is not constant
   29 |     _Static_assert((ptr) == (ptr), "xzfree argument has side effects"); \
      |                    ~~~~~~^~~~~~~~
lib/cyrusdb_twom.c:41:5: note: in expansion of macro 'xzfree'
   41 |     xzfree(foo());
      |     ^~~~~~
```

(With apologies to cyrusdb_twom.c, which was already open in my editor when I needed a place to add bad code for testing.)

Note that in both cases the error is "expression in static assertion is not constant", and not "xzfree argument has side effects".  I think it's clear enough though, and a definite improvement.


elliefm:
Oh great it breaks under clang, what the hell.  Back to draft while I figure this out.

elliefm:
https://en.cppreference.com/c/language/constant_expression:
> An integer constant expression is an expression that consists only of
> 
> - operators other than assignment, increment, decrement, function-call, or comma, except that cast operators can only cast arithmetic types to integer types unless they are part of an operand to a sizeof , _Alignof(since C11)(until C23), alignas(since C23) or typeof/typeof_unqual(since C23) operator.
> - integer constants
> - enumeration constants
> - character constants
> - floating constants, but only if they are immediately used as operands of casts to integer type
> - sizeof operators whose operands are not VLA(since C99)
> - _Alignof(until C23) alignof(since C23) operators (since C11)
> - named and compound literal constants that are of integer type or that are of arithmetic type and are the immediate operands of casts (since C23)
> 
> Integer constant expressions are evaluated at compile time. The following contexts require expressions that are known as integer constant expressions: 
> - [...]
> - The first argument of [_Static_assert](https://en.cppreference.com/c/language/_Static_assert) (since C11)
> - [...]

clang's complaints (so far) are of the form:

```
imap/caldav_alarm.c:56:12: error: static_assert expression is not an integral constant expression
    xzfree(alarmdata->mboxname);
    ~~~~~~~^~~~~~~~~~~~~~~~~~~~
./lib/xmalloc.h:29:21: note: expanded from macro 'xzfree'
    _Static_assert((ptr) == (ptr), "xzfree argument has side effects"); \
                   ~^~~~~~~~~~~~~
```

That is, it's complaining about `(alarmdata->mboxname) == (alarmdata->mboxname)` (and similar).  But this expression uses "member access" and "comparison" operators, which are permitted, so this _is_ an "integral constant expression" and technically my implementation is fine.

Given that neither the pointer nor the struct nor the member are marked `volatile`, the left and right sides of the comparison are tautologically equal without ever needing to dereference the pointer or read the member value, so it can be evaluated at compile time.  Seems like gcc figures this out, whereas I guess clang sees the deref and spooks without analysing further.

The point of the check is to stop the programmer from doing dumb things with `xzfree()`, while permitting valid uses.  As long as one of our CI compilers does this correctly, that goal is achieved.  So I think I'll use conditional compilation to leave the `_Static_assert` out when clang is detected, and we'll rely on gcc alone to warn us when we hold xzfree wrong.

lib/xmalloc.h
```diff
@@ -25,11 +25,29 @@ extern void *xmemdup(const void *ptr, size_t size);
  *
  * CAUTION: ptr argument is evaluated multiple times, beware side effects!
  */
-#define xzfree(ptr) do {                                                \
+#ifndef __clang__
```
rsto: `_Static_assert` isn't a keyword in C++ and we already include `xmalloc.h` in `icu_wrap.cpp` and `xapian_wrap.cpp` (but don't use `xzfree` so far). `static_assert` would compile for both C++ and C11, and I think we already agreed on setting C11 as the base version for builds?

nwc10 (approved):

configure.ac
```diff
@@ -2137,6 +2137,34 @@ AS_IF(
     )]
 )
 
+AC_CACHE_CHECK(
+    [whether the C compiler supports member access operators in _Static_assert],
```
wolfsage: I'm mildly confused about this testing for "member access" in the compiler, when the problem (as reported by the compiler) seems to be about whether an expression is a integral constant expression or not.

That is, other parts of cyrus are problematic and aren't member access:
```
imap/jmap_util.c:1245:21: error: static_assert expression is not an integral constant expression
                    xzfree(groupname);
                    ^~~~~~~~~~~~~~~~~
./lib/xmalloc.h:29:20: note: expanded from macro 'xzfree'
    _Static_assert((ptr) == (ptr), "xzfree argument has side effects"); \
```

This is reproducible with clang `clang test.c`:
```
void main (void) {
  char *ptr;
  _Static_assert(ptr == ptr);
}
```
```
test.c:1:1: warning: return type of 'main' is not 'int' [-Wmain-return-type]
void main (void) {
^
test.c:1:1: note: change return type to 'int'
void main (void) {
^~~~
int
test.c:3:28: warning: '_Static_assert' with no message is a C2x extension [-Wc2x-extensions]
  _Static_assert(ptr == ptr);
                           ^
                           , ""
test.c:3:18: error: static_assert expression is not an integral constant expression
  _Static_assert(ptr == ptr);
```

And even gcc if you ask it nicely:
```
gcc -Wpedantic test.c
test.c:1:6: warning: return type of 'main' is not 'int' [-Wmain]
    1 | void main (void) {
      |      ^~~~
test.c: In function 'main':
test.c:3:3: warning: ISO C11 does not support omitting the string in '_Static_assert' [-Wpedantic]
    3 |   _Static_assert(ptr == ptr);
      |   ^~~~~~~~~~~~~~
test.c:3:22: warning: expression in static assertion is not an integer constant expression [-Wpedantic]
    3 |   _Static_assert(ptr == ptr);
      |     
```

All this is to say -- is "supports member access ... in _Static_assert" accurate when it's more that clang isn't as permissive as gcc?

lib/xmalloc.h
elliefm: `static_assert` in C before C23 is a macro defined in the system `assert.h` header, which we can't use because it conflicts with our `assert.h` header (I would like to fix that someday, but I don't have current plans).  This is why I'm using the C11 `_Static_assert` keyword directly rather than the nicer macro.

You're right about C++ not supporting the C11 spelling though.  We already use `_Static_assert` in a bunch of places and they haven't been a problem, presumably because no C++ code has tripped over them yet.

configure.ac
elliefm: Interesting.

I wrote earlier:

> clang's complaints (so far) are of the form:

And this is kinda what I was talking about when I said "so far".  Since make stops (soon) after the first error, the only errors I saw from clang in CI were from lines that were using member access operators.  Since the build stopped there, I never witnessed anything else it complained about, but I looked into that complaint and concluded that it was probably the member access that was bothering it.

But yeah you're right, good catch.  This isn't about the member access operator, it's about whether the compiler decides if it's an integer constant expression before or after resolving the expression.  If it resolves the expression first (normal gcc), then any non-volatile `(x) == (x)` is tautologically true (which gcc warns about), and resolves to an integer constant `1`, which is acceptable.  If it doesn't resolve it first (clang, pedantic gcc), then any variable access is invalid.

I'll refactor and rename the configure check and update the comments to match the behaviour we're actually relying on.

lib/xmalloc.h
elliefm: I thought about this for a bit, and I'm going to resolve the `static_assert` situation project-wide in a new PR, and then rebase this one after.

elliefm: This is #6344

elliefm:
Rebase this and switch to `static_assert` after #6344 lands

configure.ac
elliefm: @wolfsage I've rewritten the last two commits.  Is that better?

(I also rebased on master, and, while doing so, fixed another file that was newly-missing assert.h.)

rjbs:
Lots of cooks in the CR already, I'm going to step back and let them do their thing without adding more salt myself.

ksmurchison (approved):

brong (approved):
Sure.  This seems like a don't punch yourself. But also, good not to do that.

configure.ac
wolfsage: Yup! Thanks. Does this still need rebasing to use the new static_assert() def from assert.h instead or..?

elliefm: Yeah it does.  Though this places me in a bit of a conundrum, because my original objective here was to stop xmalloc.h from dragging in assert.h, and to switch to `static_assert`, I'll have to bring assert.h back.  I might be able to do some local shenanigans so that it doesn't need it, but then everything that uses xmalloc.h will inherit the shenanigans, and that's probably worse than just dragging in assert.h...

In composing this comment, I think I've convinced myself to just bring back assert.h and live with that.

elliefm:
Rebased on current master, and added one new commit which makes the same change to xmalloc.h as #6344 already made for the rest of the code base.

lib/xmalloc.h
```diff
@@ -26,12 +26,42 @@ extern void *xmemdup(const void *ptr, size_t size);
...
+ *
+ * The point of the _Static_assert check is to prevent us misusing xzfree.  As
+ * long as one of our CI compilers (gcc) does this check, it doesn't matter if
+ * other ones don't.  So it's fine to omit the check when clang is detected.
  */
-#define xzfree(ptr) do {    \
-    assert((ptr) == (ptr)); \
+#ifndef __clang__
+# define xzfree(ptr) do {                                               \
+    static_assert((ptr) == (ptr), "xzfree argument has side effects");  \
```
wolfsage, line 55: I'm curious though; how is it this doesn't warn about tautological comparisons always passing?

wolfsage: Because gcc isn't generating warnings for code in macros?? This... sounds familiar but I can't figure out what controls that or why that is

elliefm: Huh, that's a good question.  In the configure check I'm explicitly suppressing the tautological comparison warning, because it does warn there.  But when xzfree is used normally and correctly, it doesn't complain, even though the same tautology is generated.  And it must know it's a tautology, otherwise it would fail the "integer constant expression" requirement in the same way clang does.

So, yeah.  As far as I understand this _should_ warn, and I can't really explain why it doesn't...

If it _did_ warn, and we wanted to suppress the warning in this context only (with like `#pragma GCC diagnostic ignored ...`), we wouldn't be able to do so from within the macro, because we can't embed a preprocessor directive into a macro expansion.  We'd have to do it at every `xzfree()` call site, and the ergonomic benefit of the macro would be lost.  Maybe gcc came to the same conclusion and that's why it doesn't warn in this situation.  But I'm reading tea leaves here, not spitting facts; I really don't know.

The configure check isn't a macro expansion, so none of that applies, and that's why it warns normally?

wolfsage: Thanks I wanted to make sure I wasn't missing something obvious.

Seems like it's intentional behaviour because macros are often used to check for equivalence and a tautological warning doesn't make sense there (that's my understanding based on discussions in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92668 and https://gcc.gnu.org/bugzilla/show_bug.cgi?id=70477)

elliefm: Ahh right, good find.  And I guess with xzfree being a function-like macro, it gets a location assigned, and therefore it falls under the purview of the earlier fix, unlike the examples in the bug reports, which don't have locations and so are still affected.

Do you have any objection to this merging?  It has approvals already, but not from you, and since you actually raised concerns and are still engaged, I don't want to disregard that lack.  But I would like to merge it and stop thinking about it ;)

wolfsage: I swear I typed "LGTM" and clicked approve but I can't find it!

wolfsage (approved):
LGTM!

PR #6322, opened 2026-08-31
http_h2.c: don't ever advertise "h2c" in Alt-Svc

ksmurchison:
"h2c" (cleartext HTTP/2) has been deprecated and Alt-Svc is supposed to advertise "better" protocol options than what is currently being used by the client, NOT "lesser" options.


ksmurchison:
Superseded by #6329

PR #6323, opened 2026-08-31
jmap_mail.c: re-order to avoid pointer arithmetic on a void pointer

nwc10:
The value in question is actually an integer cast to a pointer so that it can be stored in a hashu64. Perform the addition as part of the integer expression prior to casting, instead of on the cast value. The latter is undefined behaviour, and clang's ubsan will fault it with optimisation enabled.

### DIFF
--- a/imap/jmap_mail.c
+++ b/imap/jmap_mail.c
@@ -6827,8 +6827,8 @@ static void _email_keywords_fini(struct email_keywords *keywords)
 static void _email_keywords_add_keyword(struct email_keywords *keywords,
                                         const char *keyword)
 {
-    uintptr_t count = (uintptr_t) hash_lookup(keyword, &keywords->counts);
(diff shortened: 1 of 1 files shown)


nwc10:
@rsto - this isn't urgent. Don't sweat it if you can't review this for a few days.

nwc10:
Also, I have no idea why clang thinks that this needs to be a **run** time check when

1. it's detectable at compile time
2. IIRC it should be a source code **error**, as void pointers don't point to anything with a size, so can't do pointer arithmetic. And whilst it's gcc-ism to assume it's OK to treat `void *` like `char *`,  I thought that we had sufficient "make everything an error" flags that this would be one too.

With this tweak, I believe we can (also) pass all the Casandane tests with clang, ubsan and `-Og`

rsto (approved):

PR #6324, opened 2026-08-31
Guesstz refactor and format warning

nwc10:
Refactor guesstz.c

* refactor encode_db() to remove some duplicate code, which makes the subsequent commits simpler
* change malloc() to xmalloc(), strdup() to xstrdup(), etc, which is now possible with this code in the main tree

Address a couple of problems

* a corrupt guesstz.db could cause bad memory reads resulting in SEGVs
* gcc ubsan refused to compile format_offset() for a timezone offset of INT_MIN

gcc is *technically* correct on this - a timezone about 1633 **years** before UTC wasn't covered by the existing overflow logic. Update the overflow logic to cover this too, but more importantly also to report errors.

### DIFF
--- a/imap/cyr_guesstz.c
+++ b/imap/cyr_guesstz.c
@@ -134,8 +134,13 @@ int main(int argc, char **argv)
 
     if (mode == ENCODE) {
         char *dump = guesstz_encode(gtz);
-        puts(dump);
-        free(dump);
+        if (dump) {
+            puts(dump);
+            free(dump);
+        } else {
+            fprintf(stderr, "%s\n", guesstz_error(gtz));
+            r = EX_DATAERR;
+        }
     }
     else {
         char *buf = slurp(stdin);
--- a/imap/guesstz.c
+++ b/imap/guesstz.c
@@ -358,7 +358,8 @@ static char *guess_timezone(struct db *db,
                             icaltimetype trend)
 {
     struct observances obs = { 0 };
-    char *tzid = NULL;
+    char etc_buf[11];
+    const char *tzid = NULL;
 
     /* Limit expansion span to database time span */
     if ((icaltime_is_null_time(trend)) ||
@@ -381,10 +382,8 @@ static char *guess_timezone(struct db *db,
 
             int32_t hh = firstob.offset / (60*60);
             if (-14 <= hh && hh <= 12) {
-                char buf[11];
-                snprintf(buf, 11, "Etc/GMT%+d", -hh);
-                buf[10] = '\0';
-                tzid = strdup(buf);
+                snprintf(etc_buf, sizeof(etc_buf), "Etc/GMT%+d", -hh);
+                tzid = etc_buf;
                 goto done;
             }
         }
@@ -425,8 +424,7 @@ static char *guess_timezone(struct db *db,
         }
 
         /* Found a match! */
-        free(tzid);
-        tzid = strdup(dbtz.tzid);
+        tzid = dbtz.tzid;
 
         /* Stop if this is a preferred timezone */
         if (is_preferred_tzidx(db, dbtz.idx)) {
@@ -436,7 +434,7 @@ static char *guess_timezone(struct db *db,
 
 done:
     free(obs.alloc);
-    return tzid;
+    return tzid ? strdup(tzid) : NULL;
 }
 
 static int compare_tzoffset(const void *va, const void *vb)
@@ -830,41 +828,37 @@ static json_t *encode_db(struct db *db)
 
     /* Encode offsets */
     json_t *jtzoffsets = json_object();
+    json_object_set_new(jdb, "tzoffsets", jtzoffsets);
 
     if (db->tzoffsets.count) {
-        json_t *jofftzs = json_array();
+        json_t *jofftzs = NULL;
         const uint8_t *data = db->tzoffsets.data;
         char offsetstr[OFFSETSTR_MAX];
+        /* gcc ubsan incorrectly assumes that this needs initialising */
         int32_t prevoff = 0;
         size_t i;
 
         for (i = 0; i < db->tzoffsets.count; i++) {
             int32_t offset = load_int32_t(data);
             data += 4;
-            if (i > 0 && prevoff != offset) {
-                format_offset(prevoff, offsetstr);
-                json_object_set_new(jtzoffsets, offsetstr, jofftzs);
+            if (!jofftzs || prevoff != offset) {
+                format_offset(offset, offsetstr);
                 jofftzs = json_array();
+                json_object_set_new(jtzoffsets, offsetstr, jofftzs);
+                prevoff = offset;
             }
-            prevoff = offset;
 
             uint64_t tzidx = load_uint64_t(data);
             data += 8;
             const char *tzid = (const char *)db->timezones + tzidx;
             json_array_append_new(jofftzs, json_string(tzid));
         }
-
-        if (json_array_size(jofftzs)) {
-            format_offset(prevoff, offsetstr);
-            json_object_set_new(jtzoffsets, offsetstr, jofftzs);
-        }
-        else json_decref(jofftzs);
     }
-
-    json_object_set_new(jdb, "tzoffsets", jtzoffsets);
 
     /* Encode preftzs */
     json_t *jpreftzs = json_array();
+    json_object_set_new(jdb, "preftzs", jpreftzs);
+
     if (db->preftzs.count) {
         uint16_t i;
         for (i = 0; i < db->preftzs.count; i++) {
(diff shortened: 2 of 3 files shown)


nwc10:
@rsto - this isn't urgent

With this change I believe we can (also) pass all the Cassandane tests with gcc ubsan -Og

imap/guesstz.c
```diff
@@ -436,7 +434,7 @@ static char *guess_timezone(struct db *db,
 
 done:
     free(obs.alloc);
-    return tzid;
+    return tzid ? strdup(tzid) : NULL;
```
rsto: This could become `xstrdupnull(tzid)`

imap/guesstz.c
```diff
@@ -550,26 +550,23 @@ static int write_timezones(icalarray *timezones, FILE *fp)
 static char *ianaversion_from_zonedir(const char *zoneinfo_dir)
 {
     char *ianaversion = NULL;
-    char *fname = malloc(strlen(zoneinfo_dir) + 9);
-    fname[0] = '\0';
-    strcat(fname, zoneinfo_dir);
-    strcat(fname, "/version");
+    char *fname = strconcat(zoneinfo_dir, "/version", (char *) NULL);
     FILE *fp = fopen(fname, "r");
+    free(fname);
```
rsto: To avoid future use-after-free, this could be freed with `xzfree`. The extra cost of assigning NULL is almost nothing and we are not on a hot path here.

imap/cyr_guesstz.c
```diff
@@ -134,8 +134,13 @@ int main(int argc, char **argv)
 
     if (mode == ENCODE) {
         char *dump = guesstz_encode(gtz);
-        puts(dump);
-        free(dump);
+        if(dump) {
```
rsto: nit: no space between `if` and `(`

imap/guesstz.c
```diff
@@ -796,12 +830,14 @@ static int read_db(struct db *db, void *data)
     data += PREFTZ_SIZE * db->preftzs.count;
 
     db->timezones = data;
+    db->timezones_max_idx = end - 1 - (uint8_t *) data;
```
rsto: now that we have `size` for the actual file size, we could check here that `db->timezones < end` holds, e.g. that we didn't overflow due to a bogus preftzs count in the file.

imap/guesstz.c
```diff
@@ -930,31 +960,35 @@ EXPORTED char *guesstz_guess(guesstz_t *gtz,
...
-    gtz->addr = mmap(NULL, gtz->sb.st_size, PROT_READ, MAP_PRIVATE, gtz->fd, 0);
+    gtz->addr = mmap(NULL, sb.st_size, PROT_READ, MAP_PRIVATE, fd, 0);
     if (gtz->addr == MAP_FAILED) {
         print_error(gtz, "mmap: %s", strerror(errno));
         goto done;
     }
+    gtz->length = sb.st_size;
 
-    read_db(&gtz->db, gtz->addr);
+    read_db(&gtz->db, gtz->addr, sb.st_size);
```
rsto: now that `read_db` can return an error (e.g. `-1`), we should handle that here with `print_error`, closing `fd` and `goto done`.

rsto (changes requested):
Thanks for this, that will definitely improve things. I just have a few minor requests.

imap/guesstz.c
nwc10: I missed that that existed. Changed - thanks.

imap/guesstz.c
nwc10: I only moved the `free`!

(yes, this is a good idea) - changed.

imap/cyr_guesstz.c
nwc10: Oops. Well spotted. Fixed.

imap/guesstz.c
nwc10: I think that there are 3 potential (further) overflow points in the file after the first size check that I added (for the magic number)

1. `ianaversion` is something bogus (ie the file ends before the `NUL` byte, or too soon after)
2. `tzoffsets.count` takes us beyond the file
3. `preftzs.count` likewise (as you comment)

so I think I ought to make it robust for all of them...

imap/guesstz.c
nwc10: I agree with your conclusion, but technically a nit - I think you're misreading the diff, as I didn't **add** the `int` return or the `return -1` - the code was already doing this and ignoring this error return.

But "my bad" for looking at all the code and not realising this myself - ignoring the error return **is** a bug, and one I should have spotted, because the calling code blunders onward despite the struct in memory being all zeroes (from the `xmalloc` in the caller). So NULL pointers and SEGVs soon...

I think that the correct solution is to

1. inline `read_db` into its only caller, `guesstz_open`
2. change the `return -1` into `print_error` and `goto done;` (closing `fd` is now moved earlier, so has already happened)
3. add checks described in the comment above for all the other "file corrupt, about to read beyond the end"

I'll start on this tomorrow, once I need a break from the current task.

rsto: > I didn't **add** the `int` return or the `return -1` - the code was already doing this and ignoring this error return.

Ah, right, sorry! Your solution sounds good to me, so if you care to implement I look forward to it. Otherwise we can fix in a later PR.

nwc10: I'm happy to implement it, as this is the fastest way to get it out of my head. Otherwise it's going to nag my subconscious as "we did a half-baked job"

nwc10: Right, I think the revised version covers your concerns. In that I *think* that it covers every possible case where a truncated or corrupted file could cause an attempted read outside of the file.

I was wrong in my comment earlier - I'd moved the `close(fd)` earlier, but not as early as I thought. Now fixed, as per your suggestion.

Tempting fate, I realise, but I hope we don't get any security reports about this now.

nwc10: @rsto - aargh, the valgrind failure report from CI is real. I'm going to have to look into that

I assume that the code in the branch is now 98% correct, but it's probably best if you hold off looking again until CI passes.

nwc10: What I missed:

```
diff --git a/cunit/guesstz.testc b/cunit/guesstz.testc
index d1727b228d..e424dc8970 100644
--- a/cunit/guesstz.testc
+++ b/cunit/guesstz.testc
@@ -640,6 +640,7 @@ static void test_truncation(void)
         const char *have = guesstz_error(gtz);
         CU_ASSERT_PTR_NOT_NULL((void *)have);
         CU_ASSERT_STRING_EQUAL(have, "database truncated");
+        guesstz_close(&gtz);
     } while (*where++);
 }

@@ -650,6 +651,7 @@ static void corrupt_and_open_fails(const vandalism what, const char *want)
     const char *have = guesstz_error(gtz);
     CU_ASSERT_PTR_NOT_NULL((void *)have);
     CU_ASSERT_STRING_EQUAL(have, want);
+    guesstz_close(&gtz);
 }

 static void test_magic(void)
```

imap/guesstz.c
```diff
@@ -211,37 +228,10 @@ static int is_preferred_tzid(const char *tzid)
...
-
-    return strlen(buffer);
-}
-
-
 static void observances_from_ical(struct observances *obs, icalarray *icalobs)
 {
     obs->count = (uint32_t) icalobs->num_elements;
-    obs->alloc = malloc(obs->count * OBSERVANCE_SIZE);
+    obs->alloc = xmalloc(obs->count * OBSERVANCE_SIZE);
```
rsto, line 240: This could degenerate to `xmalloc(0)` for a zero `obs->count` but xmalloc fails fatally if libc returns NULL for allocating zero bytes. Let's either set this to NULL explicitly if obs->count is zero, or not call `observances_from_ical` at all.

rsto (approved):
This looks good to me, except for one fix that could abort in the degenerate case of a VTIMEZONE with zero observances. No need to re-review, I'm going to approve now assuming this gets handled either in the two options I outlined, or somehow else.

imap/guesstz.c
nwc10: Oh my. Well spotted. I'd not thought of that (and will try to remember it)

I addressed it like this:

```
@@ -241,7 +241,13 @@ static int format_offset(int32_t offset, char *buffer)
 static void observances_from_ical(struct observances *obs, icalarray *icalobs)
 {
     obs->count = (uint32_t) icalobs->num_elements;
-    obs->alloc = malloc(obs->count * OBSERVANCE_SIZE);
+
+    if (!obs->count) {
+        obs->data = obs->alloc = NULL;
+        return;
+    }
+
+    obs->alloc = xmalloc(obs->count * OBSERVANCE_SIZE);
     obs->data = obs->alloc;

     const icaltimezone *utc = icaltimezone_get_utc_timezone();
```

PR #6327, opened 2026-08-31
jmap: fix pool leak in process_resultrefs() on unresolved wildcard refs

ksmurchison:
The array pool built while resolving a wildcard result reference was only drained on the per-reference success path, so a reference that failed to resolve (e.g. a later sibling branch missing) leaked every array built for branches that did resolve. Hoist the pool to function scope and drain it on the fail path too. Add a Cassandane test that reproduces the leak via repeated requests and checks worker RSS growth.

### DIFF
--- a/imap/jmap_api.c
+++ b/imap/jmap_api.c
@@ -139,6 +139,9 @@ static int process_resultrefs(json_t *args, json_t *resp, json_t **err)
     const char *arg;
     int ret = -1;
 
+    /* Pool of newly created JSON objects */
+    ptrarray_t pool = PTRARRAY_INITIALIZER;
+
     void *tmp;
     json_object_foreach_safe(args, tmp, arg, ref) {
         if (*arg != '#' || *(arg+1) == '\0') {
@@ -186,11 +189,6 @@ static int process_resultrefs(json_t *args, json_t *resp, json_t **err)
         if (!res) goto fail;
 
         /* Extract the reference argument value. */
-        /* We maintain our own pool of newly created JSON objects, since
-         * tracking reference counts across newly created JSON arrays is
-         * a pain. Rule: If you incref an existing JSON value or create
-         * an entirely new one, put it into the pool for cleanup. */
-        ptrarray_t pool = PTRARRAY_INITIALIZER;
         json_t *val = extract_value(json_array_get(res, 1), path, &pool);
         if (!val) goto fail;
 
@@ -199,16 +197,19 @@ static int process_resultrefs(json_t *args, json_t *resp, json_t **err)
         json_object_del(args, arg);
 
         /* Clean up reference counts of pooled JSON objects */
-        json_t *ref;
         while ((ref = ptrarray_pop(&pool))) {
             json_decref(ref);
         }
-        ptrarray_fini(&pool);
     }
 
+    ptrarray_fini(&pool);
     return 0;
 
   fail:
+    while ((ref = ptrarray_pop(&pool))) {
+        json_decref(ref);
+    }
+    ptrarray_fini(&pool);
     return ret;
 }
 
--- a/cassandane/tiny-tests/JMAPCore/resultref_wildcard_fail_mem_cleanup
+++ b/cassandane/tiny-tests/JMAPCore/resultref_wildcard_fail_mem_cleanup
@@ -1,0 +1,55 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_resultref_wildcard_fail_mem_cleanup
+    ($self)
+{
+    # Use compact JSON to keep us under the 4k jmap_max_request_size limit
+    my $jmap = $self->default_user->new_jmaptester({ json_pretty => 0 });
+    my $using = ['urn:ietf:params:jmap:core'];
+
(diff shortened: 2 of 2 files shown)


cassandane/tiny-tests/JMAPCore/resultref_wildcard_fail_leak
```diff
@@ -0,0 +1,156 @@
...
+    ];
+
+    xlog $self, "sanity check: the request round-trips and the wildcard";
+    xlog $self, "reference fails visibly only as a routine JMAP error";
+    my $res = $jmap->CallMethods($req, $using);
+    $self->assert_str_equals('Core/echo', $res->[0][0]);
+    $self->assert_deep_equals({ x => $tree }, $res->[0][1]);
+    $self->assert_str_equals('error', $res->[1][0]);
+    $self->assert_str_equals('invalidResultReference', $res->[1][1]{type});
+
```
wolfsage: If you stopped the test right here, that'd probably be good enough.

Our asan CI run properly flags the memory leak, and passes once your fix is applied.

I don't know that we need to play games checking rss in tests anymore since asan can do that for us

ksmurchison: Yeah, that was Claude.  Removed that stuff

wolfsage: That's one of the risks of claude - it will happily write more code than necessary to get the job done

ksmurchison: @wolfsage Are there still more changes that need to be made?

imap/jmap_api.c
```diff
@@ -139,6 +139,18 @@ static int process_resultrefs(json_t *args, json_t *resp, json_t **err)
     const char *arg;
     int ret = -1;
 
+    /* We maintain our own pool of newly created JSON objects, since
+     * tracking reference counts across newly created JSON arrays is
+     * a pain. Rule: If you incref an existing JSON value or create
+     * an entirely new one, put it into the pool for cleanup.
+     *
+     * The pool lives for the whole function, not just one iteration
```
rsto: This whole new paragraph is just an artefact of using Claude. It loves to write comments that tell it the context of a change it did, but once merged, the context almost always is irrelevant to the next reader. I suggest to remove this whole paragraph. In general, I suggest to manually rewrite *all* comments produced by Claude.

cassandane/tiny-tests/JMAPCore/resultref_wildcard_fail_leak
```diff
@@ -0,0 +1,76 @@
+#!perl
+use Cassandane::Tiny;
+
+# Regression test for a memory leak in process_resultrefs() (imap/jmap_api.c).
```
rsto: This whole test descriptions is overly specific and way too long. The test might be relevant long after the code that it validates got rewritten. At best, there should be no comment and the test name be enough to understand the test goal. If that's not possible, the next best option in my opinion would be a test description that describes the issue in context of the API contract, not the internal implementation.

cassandane/tiny-tests/JMAPCore/resultref_wildcard_fail_leak
```diff
@@ -0,0 +1,76 @@
...
+            '#x' => {
+                resultOf => 'R1',
+                name => 'Core/echo',
+                path => $path,
+            },
+        }, 'R2'],
+    ];
+
+    xlog $self, "sanity check: the request round-trips and the wildcard";
+    xlog $self, "reference fails visibly only as a routine JMAP error";
```
rsto: splitting one log sentence over two log lines only to please the 80-character Perl line length limit is a hard no, I don't know why the coding agents just love to produce this.

cassandane/tiny-tests/JMAPCore/resultref_wildcard_fail_leak
```diff
@@ -0,0 +1,76 @@
...
+    my $res = $jmap->CallMethods($req, $using);
+    $self->assert_str_equals('Core/echo', $res->[0][0]);
+    $self->assert_deep_equals({ x => $tree }, $res->[0][1]);
+    $self->assert_str_equals('error', $res->[1][0]);
+    $self->assert_str_equals('invalidResultReference', $res->[1][1]{type});
```
rsto: Agree with @wolfsage that these asserts could be condensed in fewer lines, which also would make it more clear what's the expected outcome.

rsto (changes requested):
I have nothing against the actual logic change, but this change needs more vetting IMO.

imap/jmap_api.c
rjbs: This is 1000% true.  Inside Fm we have standing instructions that _heavily_ guide comments and commits to counteract these tendencies, and I _still_ rewrite the great majority of comments and commits.  Nearly 100%.

rsto:
Also, the current commit message feels too verbose. I suggest before PR submission to either rewrite the commit messages, or at least look at them and ask "is this what I would have written"?

rsto (approved):

PR #6328, opened 2026-08-31
cass: Add test showing update -> destroy -> restore an event fails

wolfsage:
cass: Add test showing update -> destroy -> restore an event fails


ksmurchison:
Superseded by #6331

PR #6329, opened 2026-08-31
HTTP/2 fixes

ksmurchison:
Three commits that should be reviewed separately

### DIFF
--- a/imap/http_h2.c
+++ b/imap/http_h2.c
@@ -507,17 +507,12 @@ HIDDEN int http2_init(struct http_connection *conn, struct buf *serverinfo)
 
 HIDDEN void http2_altsvc(struct buf *altsvc)
 {
-    if (!https && http2_callbacks) {
+    if (http2_callbacks) {
         const char *sep = buf_len(altsvc) ? ", " : "";
         const char *config_altsvc = config_getstring(IMAPOPT_HTTP_H2_ALTSVC);
 
         if (config_altsvc) {
             buf_printf(altsvc, "%sh2=\"%s\"", sep, config_altsvc);
-            sep = ", ";
-        }
-        if (httpd_localip) {
-            const char *port = strchr(httpd_localip, ';');
-            buf_printf(altsvc, "%sh2c=\":%s\"", sep, port ? port+1 : "80");
         }
     }
 }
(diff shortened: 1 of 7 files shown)


cassandane/tiny-tests/JMAPCore/http2_upgrade_settings
```diff
@@ -0,0 +1,118 @@
+#!perl
+use Cassandane::Tiny;
+use Cassandane::Util::Socket;
+
+# Buffered raw-socket reader returning either a CRLF line or an exact
+# byte count -- needed because this connection switches from HTTP/1.1
+# text framing to HTTP/2 binary framing mid-stream, so Perl's buffered
+# readline() can't safely mix with raw byte reads on the same handle.
+# Lexical, not a named sub, since every file under tiny-tests/JMAPCore/
```
wolfsage: A lexical sub would be cleaner than assigning to a scalar
`my sub h2c_reader ($sock) { ... }`

imap/http_h2.c
```diff
@@ -664,23 +667,45 @@ static int end_resp_headers(struct transaction_t *txn, long code)
     return r;
 }
 
+/* A response body chunk handed to nghttp2 via data_source_read_cb().
+ * This is a copy the data rather than a reference into the caller's own
```
wolfsage: nit: "This is a copy the data" -> "this is a copy of the data"

imap/http_h2.c
```diff
@@ -801,69 +801,65 @@ HIDDEN int http2_start_session(struct transaction_t *txn,
...
 
+    if (txn && (txn->flags.conn & CONN_UPGRADE)) {
+        /* RFC 7540 3.2 requires exactly one HTTP2-Settings header on an
+         * h2c Upgrade request. Validate and decode it *before* creating
+         * any HTTP/2 session state, so a bad or missing header falls
+         * through to HTTP/1.1 instead of leaving a half-wired session. */
+        const char **hdr = spool_getheader(txn->req_hdrs, "HTTP2-Settings");
+        if (!hdr || hdr[1]) return 0;
+
+        if (charset_decode(&txn->buf,
```
wolfsage, line 815: We should add a test for a badly formed HTTP2-Settings header returning a 400 if there isn't one somewhere else already

imap/http_h2.c
```diff
@@ -801,69 +801,65 @@ HIDDEN int http2_start_session(struct transaction_t *txn,
...
-                                         (const uint8_t *) buf_base(buf),
-                                         outlen, txn->meth == METH_HEAD, NULL);
-            if (r) {
-                syslog(LOG_WARNING, "nghttp2_session_upgrade: %s",
-                       nghttp2_strerror(r));
-            }
+        r = nghttp2_session_upgrade2(ctx->session, settings, settings_len,
+                                     txn->meth == METH_HEAD, NULL);
+        if (r) {
+            syslog(LOG_WARNING, "nghttp2_session_upgrade: %s",
```
wolfsage, line 852: We should have a test for this too

wolfsage (commented):
I think the tests could use a little cleanup (lexical subs `my sub foo` instead of my $foo = sub { ... }`), and I think we could add tests for each change being made here.

I've called out a few, but some more that might be useful is a test for the Alt-Svc header changes, and maybe something related to the httpd_keepalive (but that might be harder so maybe not)

wolfsage (changes requested):
(Hit enter too soon, all prior comments should have been "request changes")

cassandane/tiny-tests/JMAPCore/http2_upgrade_settings
ksmurchison: Hopefully fixed.

imap/http_h2.c
ksmurchison: Not sure what the issue is.  "This" starts a new sentence

imap/http_h2.c
ksmurchison: Done

imap/http_h2.c
ksmurchison: Done

imap/http_h2.c
wolfsage: Heh, my typo on 'this' confused you. The problem is "a copy the data" vs "a copy **of** the data"

ksmurchison: Ahh. I looked right past that. Will fix

wolfsage (approved):

PR #6330, opened 2026-09-01
httpd.c: Don't read from uninitialised memory for incomplete Auth headers

nwc10:
The header "Authorization: Basic" (ie missing the base64 encoded payload) would read from the array base64, even though it had not been initialised.

Add regression tests for the error paths for Basic Authorization header parsing.

### DIFF
--- a/cassandane/tiny-tests/JMAPCore/auth_fuzz
+++ b/cassandane/tiny-tests/JMAPCore/auth_fuzz
@@ -1,0 +1,82 @@
+#!perl
+use Cassandane::Tiny;
+use MIME::Base64;
+
+sub test_auth_fuzz($self)
+{
+    my $jmap = $self->default_user->new_jmaptester;
+
+    my $auth = $jmap->get_client_session;
+    $self->assert($auth->is_success, 'default success');
+    $self->assert_str_equals('200', $auth->http_response->code, 'default 200');
+
+    my $correct_auth = $jmap->ua->get_default_header('Authorization');
+    my ($token68) = $correct_auth =~ /\ABasic\s+(\S+)/i
+        or die "Can't parse <$correct_auth>";
+    my ($user, $pass) = split ':', decode_base64($token68);
+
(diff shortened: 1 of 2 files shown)


cassandane/tiny-tests/JMAPCore/auth_fuzz
```diff
@@ -0,0 +1,83 @@
+#!perl
+use Cassandane::Tiny;
+use MIME::Base64;
+
+sub test_auth_fuzz
+    :min_version_3_5    ($self)
```
ksmurchison: We've decided to ditch the :min_version stuff.  I guess we should probably write a script to prune it out of the existing tests.

cassandane/tiny-tests/JMAPCore/auth_fuzz
```diff
@@ -0,0 +1,83 @@
...
+    my ($user, $pass) = split ':', decode_base64($token68);
+
+    $jmap->ua->set_default_header('Authorization', undef);
+
+    $auth = $jmap->get_client_session;
+    $self->assert(!$auth->is_success, "no authorization");
+    $self->assert_str_equals('401', $auth->http_response->code,
+                             "no authorization 401");
+
+    # rfc2617 states that the Authorization header has exactly 1 space, but our
```
ksmurchison: We are both dating ourselves by knowing that RFC 2617 was the original HTTP Auth spec, but the current one is https://datatracker.ietf.org/doc/html/rfc9110#name-http-authentication

ksmurchison (commented):
A couple of nits, but otherwise LGTM

cassandane/tiny-tests/JMAPCore/auth_fuzz
nwc10: I updated this. I must admit that I don't **know** this. I looked up the RFC and ended up on the one that defined the grammar very clearly (and in one place), and didn't think that "this one is now superseded".

ksmurchison (approved):

PR #6335, opened 2026-09-01
work in progress - more regression tests for libconfig.c

nwc10:
@elliefm - feedback on whether I'm on the right track here would be useful

I tried to clean up my messy branch of 'XXX' commits, and merge them into

1. a commit that adds only unit tests
2. a commit that fixes a bug and adds tests for it

My code predated your cleanup of the cunit tempdir handling, so I tried to rejig what I had to be consistent with your new (better) way of doing things.

Locally I have a messy way of refactoring libconfig.c to avoid the memory leaks in testing, but as I said on Monday, I realised just before the call that a better plan would be to throw that away and use "buf" instead, which brong seemed keen on.

The reason for all this was that (IIRC) without this we have no coverage on some of the code involving hash tables in libconfig.c, code which I'd like to tweak.

### DIFF
--- a/cunit/leaksanitizer.suppress
+++ b/cunit/leaksanitizer.suppress
@@ -3,3 +3,6 @@ leak:__cunit_test_enum_invalid
 leak:__cunit_test_duration_invalid
 leak:__cunit_test_bytesize_invalid
 leak:__cunit_test_stringlist_invalid
+leak:__cunit_test_duplicate_overflow
+leak:__cunit_test_include_loop
+leak:__cunit_test_include_self
--- a/cunit/libconfig.testc
+++ b/cunit/libconfig.testc
@@ -1220,6 +1220,15 @@ static void fatal_config_read_with_filename(const char *config,
     CU_EXPECT_CYRFATAL_END(EX_CONFIG, want);
 }
 
+static void test_empty_option_key(void)
+{
+    fatal_config_read_with_filename(
+        ": value\n",
+        "empty option name",
+        2
+    );
+}
+
 static void test_empty_option_value(void)
 {
     CU_EXPECT_CYRFATAL_BEGIN;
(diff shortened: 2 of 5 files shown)


elliefm:
So I'm really unclear on what "memory leaks in testing" you're seeing.  `make VG=1 check` runs clean on master, 3.12, 3.10, and 3.8, and I actively keep it that way.  Whether by fixing the damn leaks, or by adding suppressions where the leak is a quirk of the testing setup.  Though we only have two suppressions these days... because I've fixed everything else!  (FWIW `make VG=1 check-discrete` provides better output than `make VG=1 check`, but runs slower.)

The lifetime of libconfig's data structures is the lifetime of the process, so them still existing at the end of the process is expected, and not a leak.  Under cunit, we set up and tear down libconfig for each test, which means it may be set up and torn down multiple times per process, depending on how you invoke the tests.  This is an abuse of libconfig that's necessary for testing; it doesn't warrant a refactor of libconfig.  There was a genuine leak-under-test in here up until recently, which we had a suppression for, but I fixed it and removed the suppression in 4f4fb74d83.  Since you mentioned this was work you started months ago and set aside, it might just be that the leak I already fixed was the one you were struggling with.

Or, I wonder if you're seeing leaks with something other than valgrind, in which case, do you have the correct suppressions?  `cunit/vg.supp` contains our cunit suppressions in valgrind format; I don't know what format other tools use, or where they expect to find them.  But if you're bypassing the suppressions, that would explain seeing leaks where as far as I know there are none.

For whatever it's worth, there's a parallel `cassandane/vg.supp` that's used when running cassandane under valgrind.  You do that like `./testrunner.pl --valgrind`, but it's very slow, so better to run it on just one test or suite that you're interested in, and not the whole menagerie.

elliefm:
Oh, I see `cunit/leaksanitizer.suppress`.  I didn't know (or forgot) that was there, so I bet it's out of date... yeah the tests it suppresses (on the master branch) look like the same ones I fixed a few months ago.  The suppressions can probably be removed.

Looking at it, I don't like how it just indiscriminately suppresses whole tests.  If a test with a known, suppressed leak started leaking in a new way, you wouldn't find out, because the whole test is already suppressed.

PR #6337, opened 2026-09-02
implement draft standard JMAP for Mail Sharing

rjbs:


rjbs:
@ksmurchison I suspect this may need some nontrivial work, but it's my starting point. :)

cassandane/tiny-tests/JMAPMailbox/mailbox_set_sharewith_mailboxrights
```diff
@@ -0,0 +1,99 @@
...
+            "$label: shareWith",
+        );
+    };
+
+    $assert_share->({ mayReadItems   => JSON::true }, 'lr');
+    $assert_share->({ mayAddItems    => JSON::true }, 'i');
+    $assert_share->({ mayRemoveItems => JSON::true }, 'ted');
+    $assert_share->({ mayCreateChild => JSON::true }, 'kc');
+    $assert_share->({ mayDelete      => JSON::true }, 'xc');
+    $assert_share->({ maySubmit      => JSON::true }, 'p');
```
ksmurchison, line 73: The 'c' and 'd' rights are obsolete and Cyrus reports them out for backwards compatibility.  If we are explicitly setting the new member properties, there is no reason to set the obsolete ones.

See [https://datatracker.ietf.org/doc/html/rfc4314#section-2.1.1]

ksmurchison (commented):
This looks sane to me

PR #6342, opened 2026-09-03
sphinx: give each builder its own doctree cache

brong:
The man, html and text targets have no dependency on each other, so make -j runs all three sphinx-build invocations at once, and they all shared -d docsrc/.doctrees.  One builder would read a doctree pickle another was still writing, and distcheck failed intermittently with "UnpicklingError: pickle data was truncated".  Adding -j auto to sphinx-build widened the window.

Cache per builder under the existing SPHINX_CACHE, so clean-sphinx-cache and dist-hook still sweep the lot.


brong:
Rik has a better solution, dropping this one

PR #6343, opened 2026-09-03
NNTP: delete even more of it

rjbs:


ksmurchison (approved):
Can't believe that I left all of this cruft behind!  Approved provided CI passes

PR #6344, opened 2026-09-04
allow cyrus to use `static_assert`

elliefm:
C11 added the `_Static_assert` keyword, and added a `static_assert` macro to `assert.h`.

We can't use the system `assert.h` in Cyrus because of conflicts with our own, so we haven't been able to use the `static_assert` macro; instead, we've been using the `_Static_assert` keyword directly.

C++ provides `static_assert` as a keyword, and does not recognise `_Static_assert` at all.  This hasn't been a problem yet, but we're using `_Static_assert` more and more, so it might bite us sooner or later.  C23 deprecates `_Static_assert`, and promotes `static_assert` to a keyword.  It's clear that `static_assert` is the future and `_Static_assert` is a dead end.

This PR adds a `static_assert` macro definition to _our_ `assert.h`, so that we can use the common spelling without relying on the system header.  The macro definition is conditionalised to avoid conflict with the C++/C23 keyword of the same name.  It also switches all existing usage of `_Static_assert` to `static_assert`, bringing in our assert.h if it wasn't already present.

### DIFF
--- a/cunit/sessionid.testc
+++ b/cunit/sessionid.testc
@@ -1,6 +1,7 @@
 #include "cunit/unit.h"
 #include "lib/sessionid.h"
 
+#include "lib/assert.h"
 #include "lib/libconfig.h"
 #include "lib/xmalloc.h"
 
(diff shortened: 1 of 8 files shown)


ksmurchison (approved):
Works for me on both gcc 16.2.1 and clang 22.1.8

cunit/sessionid.testc
```diff
@@ -11,8 +12,8 @@
 #define X64     X16 X16 X16 X16
 #define X255    X64 X64 X64 X16 X16 X16 X15
 #define X256    X64 X64 X64 X64
-_Static_assert(sizeof(X255) == 255 + 1, "X255 is wrong length");
-_Static_assert(sizeof(X256) == 256 + 1, "X266 is wrong length");
+static_assert(sizeof(X255) == 255 + 1, "X255 is wrong length");
+static_assert(sizeof(X256) == 256 + 1, "X266 is wrong length");
```
rsto: nit: that message erroneously stated `X266` instead of `X256` already before, but maybe now is a good occasion to fix it

rsto (approved):

cunit/sessionid.testc
elliefm: Good eye, fixed

PR #6345, opened 2026-09-04
message: let a memo follow a split conversation

brong:
The '$memo' check set mustkeep, which also gates the newcid annotation lookup immediately below it.  The msgid index only ever records a message's basecid, so once a conversation has split, every message-id in the thread still resolves to the base cid: ordinary messages follow the newcid annotation on to the current cid, but a memo skipped that and landed in the base conversation, away from the message it annotates.

Give the no-split rule its own flag so it no longer disables the redirect.

A memo whose parent is itself in the base conversation still ends up in the current one.  The msgid index cannot say which side of a split holds the parent record.

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_set_create_memo_after_splitconv
+++ b/cassandane/tiny-tests/JMAPEmail/email_set_create_memo_after_splitconv
@@ -1,0 +1,106 @@
+#!perl
+use Cassandane::Tiny;
+
+sub test_email_set_create_memo_after_splitconv
+  : ConversationsMaxThread10
+    ($self)
+{
+  my $jmap = $self->{jmap};
+
+  xlog $self, "Create conversation with maximum thread count";
+  my $convsMaxThread
+    = $self->{instance}->{config}->get('conversations_max_thread');
+  $self->make_message('Email A', messageid => "msg1\@example.com");
+  my $lastUid = 1;
+  foreach my $i (2 .. $convsMaxThread) {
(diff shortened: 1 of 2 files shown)


ksmurchison (approved):

neilj:
> A memo whose parent is itself in the base conversation still ends up in the current one. 

Is this saying if you try to add a memo to any message in the first 100 messages of a split conversation then *that* will now fail instead? Because that's definitely not a solution, and probably worse — there are by definition a lot of messages in that part of the conversation, and most commonly only a few after the split.

brong:
> > A memo whose parent is itself in the base conversation still ends up in the current one.
> 
> Is this saying if you try to add a memo to any message in the first 100 messages of a split conversation then _that_ will now fail instead? Because that's definitely not a solution, and probably worse — there are by definition a lot of messages in that part of the conversation, and most commonly only a few after the split.

Yeah, the idea being that you're normally adding the memo to the most recent message after a split, but it's ugly.  I'm gonna mark this do-not-merge and we might need to figure the real solution.

rsto:
> I'm gonna mark this do-not-merge and we might need to figure the real solution.

The real solution might rather be to separate threading logic and memos altogether. We could still make memos pretend to be regular emails when that is better for the client implementations, but we should have separate logic in Cyrus.

brong:
Closing this in favour of https://github.com/cyrusimap/cyrus-imapd/pull/6350

PR #6351, opened 2026-09-08
jmap_calendar.c, caldav_util.c: never omit DTSTAMP/updated

ksmurchison:
JSCalendar states that "updated" mandatory, but we were only setting it in CalendarEvent/set when the acting account originated the event. An attendee's copy on a shared calendar could be created or edited with jsevent_is_origin() false, silently dropping the property from both the JSON and the stored DTSTAMP.

### DIFF
--- a/cassandane/tiny-tests/JMAPCalendars/calendarevent_set_updated_shared_sharee_creates
+++ b/cassandane/tiny-tests/JMAPCalendars/calendarevent_set_updated_shared_sharee_creates
@@ -1,0 +1,60 @@
+#!perl
+use Cassandane::Tiny;
+
+use Text::VCardFast qw(vcard2hash);
+
+sub test_calendarevent_set_updated_shared_sharee_creates
+    ($self)
+{
+    my $ownerCaldav = $self->{caldav};
+    my $default_cal = $self->default_user->calendars->default;
+    my $default_cal_id = $default_cal->id;
+
+    xlog $self, "create sharee and share calendar";
+    my $sharee_user = $self->{instance}->create_user('sharee');
+    my $shareeJmap = $sharee_user->jmap;
+    $default_cal->share_with($sharee_user => [qw(mayReadItems mayWriteAll)]);
+
+    xlog $self, "sharee creates event where owner is an invitee, not organizer";
+    my $res = $shareeJmap->CallMethods([
+        ['CalendarEvent/set', {
+            accountId => 'cassandane',
+            create => {
+                event => {
+                    version => '2.0',
+                    calendarIds => {
+                        $default_cal_id => JSON::true,
+                    },
+                    start => '2026-01-02T03:04:05',
+                    timeZone => 'Etc/UTC',
+                    duration => 'PT1H',
+                    title => 'event',
+                    organizerCalendarAddress => 'mailto:someone@example.com',
+                    participants => {
+                        cassandane => {
+                            calendarAddress => 'mailto:cassandane@example.com',
+                            expectReply => JSON::true,
+                            participationStatus => 'accepted',
+                        },
+                    },
+                },
+            },
+        }, 'R1'],
+        ['CalendarEvent/get', {
+            accountId => 'cassandane',
+            ids => [ '#event' ],
+            properties => ['isOrigin', 'updated', 'x-href'],
+        }, 'R2'],
+    ]);
+    my $event = $res->[1][1]{list}[0];
+    $self->assert_not_null($event);
+    $self->assert_cmp_deeply(superhashof({
+        isOrigin => jfalse(),
+        updated  => ignore(),
+    }), $event);
+
+    xlog $self, "assert DTSTAMP is present in the stored iCalendar";
+    $res = $ownerCaldav->Request('GET', "" . $event->{'x-href'});
+    my $vevent = vcard2hash($res->{content})->{objects}[0]{objects}[0];
+    $self->assert_not_null($vevent->{properties}{dtstamp}[0]{value});
+}
--- a/cassandane/tiny-tests/JMAPCalendars/calendarevent_set_updated_shared_sharee_updates
+++ b/cassandane/tiny-tests/JMAPCalendars/calendarevent_set_updated_shared_sharee_updates
@@ -1,0 +1,95 @@
+#!perl
+use Cassandane::Tiny;
+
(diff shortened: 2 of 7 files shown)


cassandane/tiny-tests/JMAPCalendars/calendarevent_set_updated_shared_sharee_creates
```diff
@@ -0,0 +1,58 @@
...
+            },
+        }, 'R1'],
+        ['CalendarEvent/get', {
+            accountId => 'cassandane',
+            ids => [ '#event' ],
+            properties => ['isOrigin', 'updated', 'x-href'],
+        }, 'R2'],
+    ]);
+    my $event = $res->[1][1]{list}[0];
+    $self->assert_not_null($event);
```
rsto, line 50: nit: these asserts could be checked with `assert_cmp_deeply`, see `CLAUDE.md`:
```
- When appropriate, prefer `assert_cmp_deeply` over `assert_deep_equals` or
  multiple simple assertions.
```

cassandane/tiny-tests/JMAPCalendars/calendarevent_set_updated_shared_sharee_updates
```diff
@@ -0,0 +1,93 @@
...
+    my $default_cal = $self->default_user->calendars->default;
+    my $default_cal_id = $default_cal->id;
+
+    xlog $self, "create sharee and share calendar";
+    my $sharee_user = $self->{instance}->create_user('sharee');
+    my $shareeJmap = $sharee_user->jmap;
+    $default_cal->share_with($sharee_user => [qw(mayReadItems mayWriteAll)]);
+
+    xlog $self, "owner creates event where they and sharee are invitees";
+    my $res = $ownerJmap->CallMethods([
```
rsto, line 20: nit: these might use the CalendarEvent entity to fill the event default values. But admittedly I think it's not as straight-forward to do that for sharees as it is for the cassandane user

changes/next/cyr-3113-jmap-updated
```diff
@@ -0,0 +1,26 @@
+Description:
+
+Fix JMAP CalendarEvent "updated" (and the underlying iCalendar DTSTAMP) being
+left unset when creating or updating an event on a shared calendar where the
+acting user is an attendee rather than the organizer, and add a store-time
+safety net that stamps DTSTAMP on any component that would otherwise be
+written without one.
```
rsto: This change file, and I guess the whole PR, should address how existing data is handled. It looks to me as if existing iCalendar data without DTSTAMP still gets converted to a CalendarEvent without "updated" property? I think that's a reasonable choice, as long as we can guarantee that a CalendarEvent/set{update} sets the "updated" property (which this PR now looks to ensure).

Similarly, it does not tell anything about CalDAV PUT and iTIP REQUEST, both of which seem to be covered now as well with `caldav_store_resource` setting DTSTAMP? Tests for each would assert that we can't regress on this requirement in the future.

imap/jmap_calendar.c
```diff
@@ -5918,14 +5927,12 @@ static int updateevent_apply_patch(jmap_req_t *req,
...
-        json_t *jupdated = json_string(s);
-        json_object_set(new_event, "updated", jupdated);
-        json_object_set(update->serverset, "updated", jupdated);
-        json_decref(jupdated);
-    }
+    char s[RFC3339_DATETIME_MAX+1] = { 0 };
+    time_to_rfc3339(time(NULL), s, RFC3339_DATETIME_MAX+1);
+    json_t *jupdated = json_string(s);
+    json_object_set(new_event, "updated", jupdated);
+    json_object_set(update->serverset, "updated", jupdated);
```
rsto: This looks to set "updated" in any case, but if a sharee that only writes per-user properties (say, `alerts` or `color`) then the `CalendarEvent/set` response would suggest a new "updated" property where the actual DTSTAMP in the iCalendar data stays untouched (and a subsequent CalendarEvent/get will return the old "updated" timestamp). So it could look to a client that "updated" went back in time.

imap/jmap_calendar.c
```diff
@@ -4898,15 +4898,24 @@ static int createevent_toical(jmap_req_t *req,
...
         json_t *jupdated = json_string(s);
 
         json_object_set(create->jsevent, "updated", jupdated);
         json_object_set(create->serverset, "updated", jupdated);
+        json_decref(jupdated);
+    }
+
+    if (jsevent_is_origin(create->jsevent, &create->schedule_addresses) && !create->is_copy) {
+        // Set created.
+        time_t t_now = time(NULL);
```
rsto: Creating a new timestamp here can make "created" and "updated" timestamps get out of sync. Before, they were the same timestamp for newly created events. let's reuse the same timestamp, e.g. instantiate `time_t now = ..` once

imap/caldav_util.c
rsto, line 1141: This line should get preceded with `add_missing_dtstamp(store_ical, kind);`, otherwise, we might the iCalendar data without DTSTAMP

rsto (changes requested):

imap/caldav_util.c
ksmurchison: Fixed this by adding DTSTAMP to data that we read from disk.

imap/jmap_calendar.c

cassandane/tiny-tests/JMAPCalendars/calendarevent_set_updated_shared_sharee_creates
ksmurchison: Done

cassandane/tiny-tests/JMAPCalendars/calendarevent_set_updated_shared_sharee_updates
ksmurchison: Yeah, not sure how to do this without modifying the API to take an accountId

changes/next/cyr-3113-jmap-updated
ksmurchison: Changes file was removed

imap/jmap_calendar.c
ksmurchison: Should be fixed

rsto (approved):
LGTM, but please bump JMAPCACHE_CALVERSION so that we evict the events that have no "updated" property from the cache.

PR #6352, opened 2026-09-08
imapd.c: gate non-IMAP mailboxes behind +DAV, except for admins

ksmurchison:
Prohibit non-admin users from seeing or operating on non-IMAP mailboxes (calendars, address books, etc.) unless authenticated with the magic "+dav" userid suffix.

### DIFF
--- a/imap/imapd.c
+++ b/imap/imapd.c
@@ -818,6 +818,27 @@ static int mlookup(const char *tag, const char *ext_name,
     return r;
 }
 
+/*
+ * Return true if the current session was authenticated with the magic
+ * "+dav" userid suffix.
+ */
+static int imapd_want_dav(void)
+{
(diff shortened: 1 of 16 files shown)


imap/imapd.c
```diff
@@ -11391,7 +11475,7 @@ static void cmd_xwarmup(const char *tag)
     c = prot_getc(imapd_in);
     if (!IS_EOL(c, imapd_in)) goto syntax_error;
 
-    r = index_warmup(mbentry, warmup_flags, uids);
+    if (!r) r = index_warmup(mbentry, warmup_flags, uids);
```
brong, line 11480: This looks like it belongs in its own commit but I don't mind - either way it's a legit bug fix!

brong (approved):

PR #6358, opened 2026-09-08
jmap: Email/import must send null, not {}, for empty created/notCreated

brong:
Email/import wasn't fixed to match Email/set's responses, and to be correct to the spec.  Fix that.

This will fail tests until the matching incorrect check in JMAP-TestSuite is replaced.


rjbs (approved):
Fine with me, but the spec is (imho) a bit unclear.  Is it impermissible to send a 0-entry object?  Should've been stated, and (again imho) should've permitted it.  But, I think we should do what you're doing here anyway. 👍🏽

PR #6359, opened 2026-09-08
jmap: advance the Mailbox queryState when queryChanges reports a removal

brong:
Mailbox/queryChanges returned a newQueryState identical to the one it was given whenever the only change was a mailbox being destroyed, while still listing that mailbox in "removed".  We need to bump the state.


rsto (changes requested):
This looks good for queries that don't filter by role, but the `need_role` path in the same function looks to have the same old defect. Since this PR doesn't include a test, I created one at https://gist.github.com/rsto/6751efc6b0a06de8795939484736eb72 that should validate this for both codepaths.

brong:
Fixed and your comment added as a test, thanks!

rsto (approved):

PR #6361, opened 2026-09-09
jmap: check fromAccountId before a /copy runs

brong:
The request loop validated accountId up front but never looked at fromAccountId. Each /copy checked the source account only when it went to read something from it, so a copy with an empty "create" (or empty "blobIds" for Blob/copy) from an account the caller has no access to, or from an account that does not exist, returned a successful empty response.

RFC 8620 Section 3.6.2: accountNotFound is for an accountId that "does not correspond to a valid account", and an account the caller cannot see must be indistinguishable from one that does not exist. Answering success for one and accountNotFound for the other told the caller which accounts exist.

### DIFF
--- a/cassandane/tiny-tests/JMAPEmail/email_copy_from_inaccessible_account
+++ b/cassandane/tiny-tests/JMAPEmail/email_copy_from_inaccessible_account
@@ -1,0 +1,56 @@
+#!perl
+use Cassandane::Tiny;
+
+# RFC 8620 Section 3.6.2: an accountId that does not correspond to a valid
+# account is accountNotFound. A /copy has two account ids, and fromAccountId
+# was only checked once there was something to copy -- so an empty copy from
+# an account the caller cannot see, or from no account at all, succeeded and
+# leaked that the account exists. Both ids must be checked before the method
+# runs, whatever the payload.
+
+sub test_email_copy_from_inaccessible_account
+    ($self)
+{
+    my $jmap = $self->{jmap};
+
+    xlog $self, "create user foo but do not share anything";
+    $self->{instance}->create_user("foo");
+
+    for my $from (qw(foo nosuchuser)) {
+        xlog $self, "Email/copy with nothing to copy from $from";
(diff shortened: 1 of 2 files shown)


cassandane/tiny-tests/JMAPEmail/email_copy_from_inaccessible_account
```diff
@@ -0,0 +1,51 @@
...
+
+# RFC 8620 Section 3.6.2: an accountId that does not correspond to a valid
+# account is accountNotFound. A /copy has two account ids, and fromAccountId
+# was only checked once there was something to copy -- so an empty copy from
+# an account the caller cannot see, or from no account at all, succeeded and
+# leaked that the account exists. Both ids must be checked before the method
+# runs, whatever the payload.
+
+sub test_email_copy_from_inaccessible_account
+    :min_version_3_1
```
rsto: nit: `min_version_3_1`

cassandane/tiny-tests/JMAPEmail/email_copy_from_inaccessible_account
```diff
@@ -0,0 +1,51 @@
...
+    $self->{instance}->create_user("foo");
+
+    for my $from (qw(foo nosuchuser)) {
+        xlog $self, "Email/copy with nothing to copy from $from";
+        my $res = $jmap->CallMethods([['Email/copy', {
+            fromAccountId => $from,
+            accountId     => 'cassandane',
+            create        => {},
+        }, 'R1']]);
+        $self->assert_str_equals('error', $res->[0][0]);
```
rsto: nit: these could be tested with `assert_cmp_deeply`. Interestingly `CLAUDE.md` does instruct it to do so, so did it treat these asserts as "not appropriate" to follow the instruction?

```
 51 - When appropriate, prefer `assert_cmp_deeply` over `assert_deep_equals` or
 52   multiple simple assertions.
```

rsto (approved):
LGTM, the two nits are more an expression of surprise than a request for changes

PR #6362, opened 2026-09-09
jmap: add CalendarEvent/queryChanges

brong:
Cyrus cannot calculate CalendarEvent query changes, but we can at least say "cannotCalculateChanges" rather than having an un-implemented function.

### DIFF
--- a/cassandane/tiny-tests/JMAPCalendars/calendarevent_querychanges
+++ b/cassandane/tiny-tests/JMAPCalendars/calendarevent_querychanges
@@ -1,0 +1,36 @@
+#!perl
+use Cassandane::Tiny;
+
+# RFC 8620 Section 5.6 makes /queryChanges part of every type's method set.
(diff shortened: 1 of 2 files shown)


cassandane/tiny-tests/JMAPCalendars/calendarevent_querychanges
```diff
@@ -0,0 +1,34 @@
+#!perl
+use Cassandane::Tiny;
+
+# RFC 8620 Section 5.6 makes /queryChanges part of every type's method set.
+# Cyrus cannot calculate CalendarEvent query changes, so the method must exist
+# and say so with cannotCalculateChanges -- not vanish as unknownMethod.
+
+sub test_calendarevent_querychanges
+    :min_version_3_1
```
rsto: nit: `min_version_3_1`, wonder how that slipped through, `CLAUDE.md` instructs it not do this.

rsto (approved):

PR #6365, opened 2026-09-09
Jmap testsuite require accountid

brong:
This one needs to wait on https://github.com/fastmail/JMAP-TestSuite/pull/42 (which may also need to be split up!)


rjbs:
I think this stuff needs to be sort of coordinated with Ken.  Let's all talk soon.

brong:
> I think this stuff needs to be sort of coordinated with Ken. Let's all talk soon.

This MR has suddenly become a LOT shorter now that the "require accountId" logic is merged :)

PR #6368, opened 2026-09-09
Upgrade cyruslibs

rsto:
The upcoming cyruslibs version upgrades libical, Xapian and ICU to their latest stable releases. As a result, we have to flip the logic of a Cassandane test that before asserted that we do accept invalid iCalendar that now libical rejects. Rejecting this data is the right call.


ksmurchison (approved):

rsto:
Now rerunning CI with updated Docker image.

PR #6370, opened 2026-09-09
jmap_mail.c: support Xapian-indexed sort comparators in Email/query

rsto:
In https://github.com/cyrusimap/cyrus-imapd/pull/6266 we added support for the new category Email sort criteria, which allows to sort search results by filter conditions. Originally, we only supported to sort by filter conditions that can be evaluated from conversations.db.

Now we extend them to also support most filter conditions that require the Xapian index, but keep excluding those that require to search within email body contents (`text`, `body`, `attachmentBody`). There aren't hard technical reasons to exclude them, but doing so allows us to narrow the search results to top-level messages, which reduces the search space and should help keeping this performant while we evaluate the feature.

To implement this, we refactor the Xapian search code so that the guidsearch implementation gets the complete list of all matching guids for a Xapian query. Up until now, this only was exposed to the code in `search_xapian.c`, which called guidsearch for the G record for each matching guid. A Xapian-backed sort comparator now can binary search a given guid of the Email/query result. For efficiency, we evaluate Xapian-only conditions with OR and AND operators in one Xapian query.

One caveat is that we can't rely on all Xapian indexes to have minimum version 17 or higher. The `messageId`, `references`, and `inReplyTo` filter conditions in a category sort for a user that has Xapian version 16 or older is rejected with an `unsupportedSort`. In this case, we log a syslog with LOG_WARNING so that it's clear that the account should get reindexed.

**Update**: this now also supports "category" sorts in Email/queryChanges.


imap/search_engines.h
```diff
@@ -26,8 +26,17 @@ extern int xapian_basedir(const char *tier, const char *mboxname,
 typedef int (*search_hit_cb_t)(const char *mboxname, uint32_t uidvalidity,
                                uint32_t uid, const char *partid, void *rock);
 
-typedef int (*search_hitguid_cb_t)(const conv_guidrec_t *rec, size_t nguids,
-                                   void *rock);
+/* Size of one entry in a packed array of guid representations: the hex
+ * characters of a guid, followed by a NUL byte. */
+#define SEARCH_GUIDREP_SIZE (MESSAGE_GUID_SIZE*2 + 1)
```
brong: I feel like we need a MESSAGE_GUIDREP_SIZE instead of this `*2` magic everywhere, while we're changing things from being hard-coded numbers!

imap/jmap_mail.c
```diff
@@ -3627,6 +3627,29 @@ static void guidsearch_match_fini(struct guidsearch_match *match)
...
+
+static void
+emailquery_category_guidset_free(struct emailquery_category_guidset **gsp)
+{
+    if (!gsp || !*gsp) return;
+    free((*gsp)->guidreps);
+    xzfree(*gsp);
+}
+
+static int memcmp_guidrep(const void *va, const void *vb)
```
brong: This same function is defined in search_xapian.c (with the same name even!) - why not export that instead of re-defining it?

brong (approved):

brong:
My own code review only found the couple of things I commented on above.  Claude had some comments too, none of which are significantly big issues, but the idea of keeping 20 bytes rather than 41 per message might be worth considering.

Correctness and behaviour

- Never-indexed accounts get unsupportedSort. emailquery_category_load_guidsets returns not-supported when the session's highest createdmodseq is zero, and run_guidsearch does the same when there's no Xapian database at all. So a category sort on from for a fresh account, or one squatter hasn't reached yet, errors, while the identical condition as a filter just returns results via uidsearch. Clients will see this intermittently on new accounts. It's a defensible choice, since without a bound you can't tell "unindexed" from "matches nothing", but it should be a deliberate decision and probably a line in the changes file.
- Dropped qsort now rests on four invariants. xapian_run_guidsearch_cb skips sorting under MSGDOCS_ONLY. That's correct today: slot 0 is the cyrusid, set_sort_by_value(0, false) is ascending, the length-43 check guarantees a uniform *G* prefix, and multi-tier MSets merge in key order. But both the adjacent-dedupe in the load callback and the bsearch silently return wrong answers if any of those change. I'd either keep the qsort (glibc's mergesort on sorted input is near-linear, and this buffer is transient anyway) or add a cheap sortedness assert in the callback.
- Snapshot consistency looks right. Email/query shares one session between the guidsets and guidsearch. The uidsearch fallback and queryChanges open their own session and are bounded by the smaller createdmodseq. queryChanges compiles the category before emailsearch_run, so an unsupported sort doesn't pay for a search. The generation-mismatch edge case (compaction between the two sessions) has the same exposure as existing code.
- The MSGDOCS_ONLY doctype filter buys nothing for correctness. Part documents only ever receive body-part segments (search_xapian.c routes a segment to a P doc only when search_part_is_body is true), and the category path already excludes body parts and text. So a header-only query can never hit a P doc, and the XE G OP_FILTER is an extra postlist intersection over every G document for no narrowing. Its only real effect is to justify skipping the qsort. Dropping the flag entirely, keeping the qsort, and removing the flags parameter from run_guidsearch would be simpler and no slower.

Efficiency

- Guidset memory is 41 bytes per hit per filter. A broad condition like from on the user's own domain, or a busy list's List-Id, is close to the mailbox size: roughly 41 MB per million messages per groupBy filter, on top of the transient MSet buffer Xapian hands back. Storing 20-byte binary guids would halve that and make the bsearch cheaper, for one decode pass at load time. This is the one I'd push on before "evaluate the feature" turns into "ship it".
- Locks now held across the sort. The Xapian read lock and the upload-mailbox lock stay held through guidsearch_sort and until emailquery_search closes the session, whereas the old guidsearch_run released them before returning. Sorting a large result under the index lock delays squatter for that user.
- Two B-record reads per email in queryChanges. emailquery_changes_load_convmodseq fetches each cid's conversation for convmodseq, then the thread-keyword evaluation loads the same conversation again per match. The per-match load is pre-existing, but a small per-cid cache would serve both.
- Each Xapian-backed groupBy filter is a full get_mset(0, doccount). N filters means N+1 materialisations. Fine for evaluation, but it's the cost model to keep in mind. Deduplicating identical conditions across groups via the serialised-key hash is a nice touch.
- Good win in the uidsearch and queryChanges path. emailquery_category_of_guid now only walks G records when a filter needs folders, seen or system flags. A header-only category is a pure bsearch per message.
- Nits: the index-version check sits inside the per-guidset loop but is loop-invariant; emailquery_category_is_all_xapian is re-run at every node during build, quadratic in depth but negligible.

Tests and docs

- The tests cover the right things: NOT and OR compounds, an Email indexed after the snapshot landing in no group, the uidsearch fallback via a body filter, the version-16 downgrade, and a thread-state change moving multiple emails in queryChanges.
- email_query_sort_category_legacy_index pairs assert_str_equals on the error type with a separate assert_cmp_deeply on the sort. One assert_cmp_deeply with superhashof is the pattern you've asked for elsewhere.
- xapian_query_new_has_doctype's new Doxygen claim that the child is untouched on error checks out: the delete child is after the new inside the try.
- The XE doctype term dates from index version 4 and minimum supported is 5, so the positiveery readable index.
- Conventions: the new and edited declarations in search_engines.h use plain comments where CLAUDE.md asks for Doxygen on edited headers. The PR did this right in xapian_wrap.h. A couple of comments say the index or squatter "has not seen" an email, which is the anthropomorphising CLAUDE.md asks to avoid.

imap/search_engines.h
rsto: Right, I added a new commit at the start of the branch history that introduces `MESSAGE_GUIDREP_SIZE` and rewrites all `2*MESSAGE_GUID_SIZE` expressions in the codebase to use it.

imap/jmap_mail.c
rsto: Agreed, search_engines.h now exports `search_guidrep_cmp`

rsto:
> My own code review only found the couple of things I commented on above. Claude had some comments too, none of which are significantly big issues, but the idea of keeping 20 bytes rather than 41 per message might be worth considering.

Thanks for the review. I updated the code with most of these findings, but I did not for all of them: 
- 20 vs. 41 bytes to represent guidsets: I considered using 20 bytes for guid when I wrote the PR, but chose not to. We currently use hex-encoded guids in conversations.db and Xapian. If we decode the guids that come from Xapian, it would halve the guidsets, but it moves the conversion since we still get the hex-encoded guids from conversations.db. If we think it's worth the CPU time to save memory, we could do that. If we want to consistently use one representation (most likely the more compact 20 bytes), that will require us to change this in all layers and the persisted data.
- Two B-record reads per email in queryChanges: I agree it's not great, but making it only read B records once requires an awkward rewrite and I wanted to keep the code changes small.
- Nit, the index-version check sits inside the per-guidset loop but is loop-invariant: Similar reasoning as for B records, it's an awkward rewrite and the performance gain is basically inexistent.

PR #6372, opened 2026-09-10
zeroskip.c, twom.c: pin xxhash to -O2 on gcc instead of probing at configure

brong:
The configure probe only sees configure-time CFLAGS. A build that passes -Og at make time passed the probe at -O0 and then failed in lib/zeroskip.c with gcc 14's "inlining failed in call to always_inline".

Reproduced with gcc 14 (trixie): the failure is specific to -Og. -O0, -O1, -O2, -O3 and -Os all build, and no single -f flag from -O1/-O2 rescues -Og, so there is no narrower switch. Since no preprocessor macro distinguishes -Og from -O2, compile the header itself at -O2 with a gcc pragma around the include and drop the probe. clang is excluded because it errors on the unknown pragma under -Werror and does not have the problem.

twom.c had `XXH_NO_INLINE_HINTS 1` unconditionally, paying about 3x on every hash; it gets the same treatment.

Verified: gcc 14 and clang 19 compile both files at every optimisation level with -Wall -Wextra -Werror; at -O2 the zeroskip.o disassembly is identical apart from `__LINE__` constants; full build plus the zeroskip and aaa-db cunit suites pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2136,52 +2136,6 @@ AS_IF(
     )]
 )
 
-dnl lib/zeroskip.c compiles xxhash inline (XXH_INLINE_ALL) because not doing so
-dnl costs about 3x on ALL hashing -- one-shot as well as streaming, so every span
-dnl checksum, conversion, repack and consistency check pays it.  Some compilers
-dnl cannot honour those hints at low optimisation and fail the build outright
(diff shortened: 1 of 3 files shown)


ksmurchison (approved):
CI passes but do we know if fmcyrpkg will be able to build?

PR #6373, opened 2026-09-10
imap: fix malformed COPYUID for cross-backend proxy COPY/MOVE

### LINKED ISSUE #4027: UID MOVE to remote server returns bad COPYUID response
This came up on the info list: https://cyrus.topicbox.com/groups/info/T25c7531b76d23ba0-Mb4348afa7dbb30b2bc95f978/move-mails-between-folders-on-different-backend-servers

If a copy/move is local, then the COPYUID response is calculated locally, and everything's fine.

If the target of the copy/move is remote, then the messages are APPENDed to the target, and the APPENDUID from the remote server's response is copied into the COPYUID response to the client:

https://github.com/cyrusimap/cyrus-imapd/blob/7f17801c814e22ede72368d7dbfa8e6f4bc3f4d0/imap/imapd.c#L6698-L6707

A similar pattern occurs in imap_proxy.c:

https://github.com/cyrusimap/cyrus-imapd/blob/7f17801c814e22ede72368d7dbfa8e6f4bc3f4d0/imap/imap_proxy.c#L1170-L1179

But this behaviour makes no sense:  the APPENDUID response is specified as having two fields, whereas the COPYUID response has three: https://datatracker.ietf.org/doc/html/rfc4315#section-3

jcdelepine:
cc @elliefm, since you filed #4027 and this builds on the approach from #6081

In a Cyrus murder cluster, when COPY or MOVE is proxied cross-backend (the "hard case" in proxy_copy(), imap_proxy.c, and the analogous "local mailbox -> remote mailbox" path in imapd.c's cmd_copy()), the destination server only ever returns an APPENDUID response code ("<uidvalidity> <destuid-set>", 2 fields per RFC 4315 section 3). That response was reported back to the client verbatim, mislabelled as COPYUID -- but COPYUID requires three fields: "<uidvalidity>
<srcuid-set> <destuid-set>".

Clients that strictly parse COPYUID (e.g. Horde's
Horde_Imap_Client_Socket::_responseCode(), which maps a fixed set of keys onto the parsed values with array_combine()) reject the malformed response outright, even though the COPY/MOVE itself completes successfully server-side.

This has been a known issue since #4027 (2022), but went unnoticed in practice because a cross-backend UID MOVE previously degraded silently into a COPY without expunge (see #51 / #6081); now that MOVE completes end-to-end, the malformed COPYUID surfaces for any client that validates it.

Fix: track the real source UIDs while fetching/appending messages in proxy_copy() (imap_proxy.c), and splice them into the destination's APPENDUID response to build a spec-compliant 3-field COPYUID. Also fix the equivalent path in imapd.c's cmd_copy() for the UID COPY case; the plain-sequence-number case there is left as a follow-up (documented inline), since that code path doesn't currently support MOVE either.

The imapd.c path is fixed but not covered by a test here: exercising it requires a client connected directly to a backend hosting a local mailbox, copying to a mailbox on a different backend (the "local mailbox -> remote mailbox" branch, taken when backend_current is NULL). A Cassandane test along those lines hit an unrelated CAPA_MULTIAPPEND negotiation gap between two Cassandane-managed backends, orthogonal to this fix, so it's left for a follow-up rather than block this PR on it.

Two new Cassandane tests in MurderIMAP, covering proxy_copy()'s hard case:
- copyuid_cross_backend_move: UID MOVE cross-backend reports a correct 3-field COPYUID.
- copyuid_cross_backend_copy_msn: plain COPY by sequence number cross-backend reports the *real* source UID, not the sequence number sent on the wire.

This same fix (adapted for the pre-refactor proxy_copy()) has been deployed to a production Cyrus 3.8.8 murder cluster and confirmed to resolve the exact crash reported by a Horde IMP client. Master itself is covered here by the two Cassandane tests above; the underlying logic is identical, though master's proxy_copy() has since been refactored to the copy_cb/struct copy_rock callback style.

Fixes #4027


imap/imapd.c
```diff
@@ -6772,11 +6772,46 @@ static void cmd_copy(char *tag, char *sequence, char *name, int usinguid, int is
...
+                           https://github.com/cyrusimap/cyrus-imapd/issues/4027 */
+                        destuid = strchr(appenduid, ' ');
+                        if (destuid && usinguid) {
+                            *destuid++ = '\0';
+                            prot_printf(imapd_out,
+                                        "%s OK [COPYUID %s %s %s] %s\r\n", tag,
+                                        appenduid, sequence, destuid,
+                                        error_message(IMAP_OK_COMPLETED));
+                        }
+                        else if (destuid) {
```
ksmurchison: I'd just assume drop the MSN case.

ksmurchison (commented):
Otherwise, looks good

imap/imapd.c
jcdelepine: Dropped the MSN case as suggested — updated comment above the if block accordingly. Thanks!

ksmurchison (approved):

imap/imapd.c
ksmurchison: Please rebase this on master and I will merge it.  I'll merge-bot failed to push to your branch

jcdelepine: Done

PR #6381, opened 2026-09-11
adjust http hunk sizing

rjbs:


rjbs:
I will address requests from !126

wolfsage (changes requested):
Marking as changes requested as requested by rjbs

imap/http_client.c
```diff
@@ -374,7 +379,7 @@ EXPORTED int http_read_body(struct protstream *pin, hdrcache_t hdrs,
     return 0;
 
   read_failure:
-    if (strcmpsafe(prot_error(pin), PROT_EOF_STRING)) {
+    if (!strcmpsafe(prot_error(pin), PROT_IDLE_STRING)) {
```
wolfsage, line 382: What is this (and other related) PROT_EOF_STRING vs PROT_IDLE_STRING change about? It seems unrelated (to me) and isn't mentioned in the commit message

wolfsage (changes requested):
This seems like there's unrelated changes (the IDLE vs EOF stuff?) in here, but I'm not sure.

Aside from that, while this does seem to fix smuggling via 0x100...., it looks like you can smuggle more easily by providing a chunk size of 0

nwc10:
> Aside from that, while this does seem to fix smuggling via 0x100...., it looks like you can smuggle more easily by providing a chunk size of 0

At least once of us is confused. It may well be me. I read the above as

this fixes this style of request smuggling 
```
100000000
BREW /decaf HTTP/1.1
...
```

where the proxy sees the request as having a very large body, but Cyrus (because of the overflow bug) thinks that the body is empty, and reads the input stream as a second request, smuggled past the proxy
...

and you then saying, but this

```
0
BREW /decaf HTTP/1.1
...
```

is still a bug, because the proxy and Cyrus will (again) have different ideas about where the requests split on the stream.


And then I think

But the chunk protocol **is** that 0 is "the end", so if the malicious client sends the above, the proxy is going to fault the request as "malfomed" at the start of `BREW`... because that line should be either blank, or a trailer (a header at the end)

https://datatracker.ietf.org/doc/html/rfc9112#chunked.encoding

and Cyrus and a proxy should both agree that the request is malformed. Which means that it's not chunk smuggling.

But I'm thinking "this is obvious right?" So you must be saying something more subtle, what am I missing?

But I might be biased here because I've been wrapping my head around https://datatracker.ietf.org/doc/html/rfc9112#chunked.encoding for a few days now (trying to work out where the corners and sharp edges are). That, and all the ways strtoul will trip you up.

So, in summary, your comment isn't making sense to me, and I don't know where the source of confusion is.

imap/http_client.c
rjbs: I can update the commit message.  The issue is that prot_error returns:

* if stream has an error, that
* if eof, fixed string
* otherwise, null

Then we were treating anything other than the EOF fixed string as a timeout.  So, any error at non-eof got 408  "time out" error.  Instead, now we issue a bad request.

wolfsage:
Hm, okay I suppose I was confused. I misunderstood the test and the scenario. With the test, if you use '0' as a chunk size, the test fails, but that's *expected*. The real problem was that we would never receive 0 to smuggle a request, because a higher proxy (nginx, etc) would have already handled that and done the right thing, but if it saw 2^32 it would say "sure that's valid" but cyrus would do the wrong thing.

Okay thanks!

imap/http_client.c
wolfsage: Great, yeah I think that would be good

rjbs:
@wolfsage I added an explanation of the EOF/IDLE change to the commit.  You have been convinced on the size handling.  I have re-requested approval.

wolfsage (commented):
Thanks. Commit message nit:

```
This commit also address our failure to return useful errors!
prot_error return either:
* the stream's error, if any
* a fixed "reached eof" string, if eof
* null, otherwise
```

"prot_error return either" -> "prot_error return**s** either" I think?

And maybe some vertical whitespace but that's less important

rjbs:
@wolfsage I have fixed the typo in the commit message.

wolfsage (approved):

PR #6384, opened 2026-09-12
tools: add find-case-conflicts linter

rjbs:
Some of us work on macOS, where if we have HTTP.pm and Http.pm (which happened recently), things get all wonky.


rjbs:
I'd originally used GLD, but the linter CI runs on a no-frills Ubuntu box that doesn't have it.

wolfsage (approved):

PR #6387, opened 2026-09-14
tzdata: update to IANA 2026d

rsto:
From the Release Notes:
Canada’s Northwest Territories moved to permanent -06 on 2026-08-21.
Obsolescent settings like TZ="EST5EDT" now conform better to POSIX.
Fix security, performance and porting bugs in zic and localtime.


rsto:
The `distcheck` CI pass fails with a sphinx error. This is unrelated to this change, it's a race condition in the build setup.

ksmurchison (approved):

PR #6382, opened 2026-09-11
parsehex() and %-encoding fixes

nwc10:
* parsehex() didn't handle 16 hex digits followed by garbage
* parsehex() didn't handle non-ASCII input
* hex_to_bin() masked its input to ASCII
* refactor charset_parse_mimexvalue() to use  charset_decode_percent(), removing duplicate logic
* more unit tests

PR #6388, opened 2026-09-14
parse (and ignore) HTTP/1.1 chunk size extensions

nwc10:
Previously Cyrus would reject any chunk size line that used ';' to start an extension, even if that line was valid.

This also fixes a bug, where a chunk size with a "0x" prefix would be accepted as valid - eg the line "0xFE\r\n" would be treated as size 254, not a protocol error.

### DIFF
--- a/imap/http_client.c
+++ b/imap/http_client.c
@@ -249,17 +249,22 @@ EXPORTED int http_read_body(struct protstream *pin, hdrcache_t hdrs,
 
         /* Read chunks until last-chunk (zero chunk-size) */
         do {
-            unsigned chunk;
+            unsigned long chunk;
+            char *endptr;
 
             /* Read chunk-size */
+            errno = 0;
             if (!prot_fgets(buf, PROT_BUFSIZE, pin) ||
-                sscanf(buf, "%x", &chunk) != 1) {
+                !isxdigit(*buf) ||    // no leading junk
+                ((chunk = strtoul(buf, &endptr, 16)) == ULONG_MAX &&
+                 errno == ERANGE) ||  // didn't overflow
+                *endptr != '\r') {    // no trailing junk
                 *errstr = "Unable to read chunk size";
                 goto read_failure;
 
                 /* XXX  Do we need to parse chunk-ext? */
             }
-            else if (chunk > body->max - body->len) {
+            if (chunk > body->max - body->len) {
                 return HTTP_CONTENT_TOO_LARGE;
             }
 
@@ -374,7 +379,7 @@ EXPORTED int http_read_body(struct protstream *pin, hdrcache_t hdrs,
     return 0;
 
   read_failure:
-    if (strcmpsafe(prot_error(pin), PROT_EOF_STRING)) {
+    if (!strcmpsafe(prot_error(pin), PROT_IDLE_STRING)) {
         /* client timed out */
         *errstr = prot_error(pin);
         syslog(LOG_WARNING, "%s, closing connection", *errstr);
--- a/cassandane/Cassandane/Cyrus/HTTP.pm
+++ b/cassandane/Cassandane/Cyrus/HTTP.pm
@@ -5,11 +5,30 @@ package Cassandane::Cyrus::HTTP;
 use strict;
 use warnings;
 
+use Errno;
+use IO::Select;
+use MIME::Base64 qw(encode_base64);
 use Net::HTTP;
 use Net::HTTPS;
 
 use base qw(Cassandane::Cyrus::TestCase);
 use Cassandane::Util::Log;
+
+=head1 NAME
+
+Cassandane::Cyrus::HTTP - HTTP/1.x tests
+
+=head1 OVERVIEW
+
+This suite tests the HTTP/1.x server at the wire level, in either of two
+styles.  C<http1_connect> and its companions hand back a L<Net::HTTP> socket:
+the test writes its own request bytes but gets the response parsed for it,
+which suits tests whose interest is in the reply.  C<< L</http1_request> >>
+instead assembles the request from named pieces and returns the response as
+raw bytes, which suits tests of malformed framing a real client would never
+produce, or of a connection the server answers on more than once.
+
+=cut
 
 sub new
 {
@@ -18,7 +37,10 @@ sub new
     my $config = Cassandane::Config->default()->clone();
     $config->set(tls_server_cert => '@basedir@/conf/certs/cert.pem',
                  tls_server_key => '@basedir@/conf/certs/key.pem',
-                 http_h2_altsvc => '127.0.0.1:8443');
+                 http_h2_altsvc => '127.0.0.1:8443',
+                 caldav_realm => 'Cassandane',
+                 httpmodules => 'caldav',
+                 calendar_user_address_set => 'example.com');
 
     my $self = $class->SUPER::new({
         config => $config,
@@ -27,7 +49,6 @@ sub new
     }, @_);
 
     $self->needs('component', 'httpd');
-    $self->needs('dependency', 'nghttp2');
     return $self;
 }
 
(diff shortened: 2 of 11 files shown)


nwc10:
This branch is 2 commits, built on top of

https://github.com/cyrusimap/cyrus-imapd/pull/6381
https://github.com/cyrusimap/cyrus-imapd/pull/6382

It will get rebased onto master (twice) if/when each of those is merged.

I wanted to get it "up" to reference it on https://github.com/cyrusimap/cyrus-imapd/pull/6381

imap/http_client.c
```diff
@@ -46,6 +46,32 @@ EXPORTED int is_mediatype(const char *pat, const char *type)
...
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
+};
+// clang-format: on
```
wolfsage, line 74: Nit: I'd put the comment of what chars these are *above* the items in the table they reference, as usually comments reference the code below them

(charset.c does this for example):
```
// clang-format: off
static const char QPMIMEPHRASESAFECHAR[256] = {
/* control chars are unsafe */
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
/* ASCII map... */
/*     !  "  #  $  %  &  '  (  )  *  +  ,  -  .  /  */
    0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 0, 1, 0, 1,
/*  0  1  2  3  4  5  6  7  8  9  :  ;  <  =  >  ?  */
    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0,
/*  @  A  B  C  D  E  F  G  H  I  J  K  L  M  N  O  */
    0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
/*  P  Q  R  S  T  U  V  W  X  Y  Z  [ \\  ]  ^  _  */
    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0,
/*  `  a  b  c  d  e  f  g  h  i  j  k  l  m  n  o  */
    0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
/*  p  q  r  s  t  u  v  w  x  y  z  {  |  }  ~ DEL */
    1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0,
/* all high bits are unsafe */
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
};

```

imap/http_client.c
```diff
@@ -156,6 +182,39 @@ EXPORTED int http_parse_framing(int http2, hdrcache_t hdrs,
     return 0;
 }
 
+EXPORTED bool http_parse_chunk_size(const char *buf, unsigned long *chunk)
+{
+    int len = strlen(buf);
+    const char *ext;
+    uint64_t value;
+
+    if (parsehex(buf, &ext, len, &value) == -1) return false;
```
wolfsage, line 191: Hmm. That commit message...
```
[util.c: parsehex() should handle non-ASCII input safely](https://github.com/cyrusimap/cyrus-imapd/pull/6388/commits/6ce700d0b49c81171c8d432c2eec508e9bbbeb5c)
Previously if it encountered a non-ASCII character it would read random
memory, because the lookup array unxdigit[] only had 128 entries.

Add 128 more entries, all of which are "not a hex digit". No-one should be
passing this function non-ASCII garbage, but untrusted user input doesn't
always play nice.
```

Specifically **"No-one should be passing this function non-ASCII garbage, but untrusted user input doesn't
always play nice."**

I was thinking this meant "callers should never do this", but I think we mean "please don't give us bad data, that's silly, but if you do we'll handle it" so it's fine/expected we hand untrusted user input in to this function here

imap/http_client.c
```diff
@@ -156,6 +182,39 @@ EXPORTED int http_parse_framing(int http2, hdrcache_t hdrs,
...
+    if (value > ULONG_MAX) return false;
```
wolfsage, line 192: I'm confused by this check - I believe parsehex() will return -1 if the result would overflow ULONG_MAX, and isn't the max value of a uint_64t ULONG_MAX?

tools/mk-token-table
```diff
@@ -0,0 +1,61 @@
...
+use v5.40.0;
+
+# I suspect that we can make more use of lookup tables for parsing user input
+#
+# I've structured this on the guess that we'll want more tables of this "shape".
+#
+# But it might be more performant to merge (up to) 8 true/false tables into one
+# and then use a macro to read 1 bit, in which case, I got the shape wrong.
+
+sub print_table($name,$set) {
```
wolfsage, line 12: nit:
`sub print_table ($name, $set) {`

wolfsage (changes requested):

imap/http_client.c
elliefm: (This is a drive-by comment, I haven't looked at it in context.)

The max value of a `uint64_t` is `UINT64_MAX`.

`ULONG_MAX` is the max value of `unsigned long`, whose sizes varies but is often 32 bits.  c.f. `unsigned long long`, which is often 64 bits, and whose maximum is `ULLONG_MAX`.

Reading between the lines, I _think_ this might be trying to ensure `value` will fit in 32 bits.  If that's the case, better to be explicit and compare against `UINT32_MAX`.

Another possibility is that it's specifically checking that it fits in an `unsigned long`, because some other part of the system is using that type.  In which case `ULONG_MAX` is arguably correct, but in that case I would add a comment clarifying that it's intentional, and not a misunderstanding or typo.

But!  We do have lots of old code that uses `unsigned long` when it really means `uint32_t` (that is, we have lots of code that unsafely assumes `unsigned long` is 32 bits), and if that's the case for the API we're trying to match, then it's better to do the right thing and use `UINT32_MAX`, even though the API itself doesn't.

Rule of thumb: if the number of bits is important, use the c99 fixed-width integer types, and their corresponding macros (in [stdint.h](https://en.cppreference.com/c/header/stdint)).  (If you need to printf/scanf them, the corresponding format string macros are are `inttypes.h`.)

tools/mk-token-table
nwc10: Thanks, fixed.

imap/http_client.c
nwc10: I missed that one when I was looking for things to steal. At least, I'd forgotten it by the time I came to write the code.

Consistency is good, and I think you're right, above is slightly better. Fixed.

imap/http_client.c
nwc10: Yes, 32 bit systems. But the fact that both of you comment on this makes me revisit the design. I was only using `unsigned long` because the code I was refactoring was using `unsigned long` and in turn I think that this was because it was doing the parsing with `strtoul`.

But that that variable doesn't need to be that type. We already have this slightly crazy definition (I guess we should fix this) where we (on 64 bit platforms) we notionally can think about content length >= than 2**32:

```
/* Context for reading request/response body */
struct body_t {
    unsigned char flags;                /* Disposition flags */
    unsigned char framing;              /* Message framing   */
    unsigned char te;                   /* Transfer-Encoding */
    unsigned max;                       /* Max allowed len   */
    unsigned long len;                  /* Content-Length    */
    struct buf payload;                 /* Payload           */
};
```

but actually we can't, as the maximum length we'd allow is 0xFFFFFFFF. And actually the code that initialises this caps it at half that.

So I changed the type of `chunk` in the calling code, removed that "did it fit into an unsigned long", and it now looks like this:

```
@@ -156,6 +182,36 @@ EXPORTED int http_parse_framing(int http2, hdrcache_t hdrs,
     return 0;
 }

+EXPORTED bool http_parse_chunk_size(const char *buf, uint64_t *chunk)
+{
+    int len = strlen(buf);
+    const char *ext;
+
+    if (parsehex(buf, &ext, len, chunk) == -1) return false;
+    if (*ext == '\r') return true;
+
+    /* What follows should be a syntactically valid chunk-ext
```

and

```
@@ -249,20 +305,18 @@ EXPORTED int http_read_body(struct protstream *pin, hdrcache_t hdrs,

         /* Read chunks until last-chunk (zero chunk-size) */
         do {
-            unsigned long chunk;
-            char *endptr;
+            /* The max we allow for body->max is BYTESIZE_UNLIMITED (2147483647)
+             * We're never going to accept chunks larger than this, but it makes
+             * sense to parse well formed headers as best possible and return
+             * the more accurate response "413 Content Too Large" instead of
+             * hitting read_failure, which will return "400 Bad Request" */
+            uint64_t chunk;

             /* Read chunk-size */
-            errno = 0;
             if (!prot_fgets(buf, PROT_BUFSIZE, pin) ||
-                !isxdigit(*buf) ||    // no leading junk
-                ((chunk = strtoul(buf, &endptr, 16)) == ULONG_MAX &&
-                 errno == ERANGE) ||  // didn't overflow
-                *endptr != '\r') {    // no trailing junk
+                !http_parse_chunk_size(buf, &chunk)) {
                 *errstr = "Unable to read chunk size";
                 goto read_failure;
-
-                /* XXX  Do we need to parse chunk-ext? */
             }
             if (chunk > body->max - body->len) {
                 return HTTP_CONTENT_TOO_LARGE;
```

which I think gets us closer to something good.

(This PR still waits for https://github.com/cyrusimap/cyrus-imapd/pull/6381 to be completed. https://github.com/cyrusimap/cyrus-imapd/pull/6382 was merged, so I rebased the thing as a merge of master and the branch in 6381)

PR #6415, opened 2026-09-16
JMAP /set response integrity fixes

ksmurchison:
See https://claude.ai/artifact/6DqaUxY5gjwEGWgQfTQVrJ?sk=D6SKMCrAqASr25HaVRX_gg for the list of issues.

Should be reviewed commit-by-commit


imap/jmap_calendar.c
```diff
@@ -2470,7 +2465,17 @@ static int jmap_calendar_set(struct jmap_req *req)
             buf_free(&buf);
             mbname_free(&mbname);
 
-            if (!r) {
+            if (r) {
+                /* The calendars are already written: fail just the
+                 * default change, not the whole batch. */
```
rsto, line 2470: nit: "fail just the default change" made me think unnecessarily long what's a "default change". I figure this means: "Report an error for changing the default calendar, not for the whole batch."

imap/jmap_contact.c
```diff
@@ -2867,7 +2867,17 @@ static int jmap_addressbook_set(struct jmap_req *req)
                                        req->accountid, &buf);
             buf_free(&buf);
 
-            if (!r) {
+            if (r) {
+                /* The addressbooks are already written: fail just the
+                 * default change, not the whole batch. */
```
rsto, line 2872: nit: same "default change"

imap/jmap_calendar.c
```diff
@@ -2470,7 +2465,17 @@ static int jmap_calendar_set(struct jmap_req *req)
             buf_free(&buf);
             mbname_free(&mbname);
 
-            if (!r) {
+            if (r) {
+                /* The calendars are already written: fail just the
```
rsto, line 2469: nit: the commit message title could me more specific, this change is all about only reporting an error for `onSuccessSetIsDefault`. Also, I found the commit message hard to understand, it expects the reader to understand the internal code mechanics, where it seems to mean: "Report a failure to change the default calendar or addressbook with a SetError, rather than reporting the whole /set method as failed."

imap/jmap_calendar.c
```diff
@@ -6413,6 +6406,13 @@ static void setcalendarevents_update(jmap_req_t *req,
         goto done;
     }
 
+    /* Scheduling goes last: an iTIP message can't be recalled. */
```
rsto, line 6409: nit: I'm not sure if "an iTIP message can't be recalled" is clearly enough describing the issue. The issue is that we sent iTIP message before actually changing the data.

Since I can't comment on a commit mesage: "once the store has settled" is Claude-speak and I can only guess it means that the "on-disk data" or "database" got "updated".

rsto (changes requested):
7b2f175e97 introduces a regression where failed updates are reported as both updated and notUpdated. There is also an issue where creation ids are set/looked up at one place but not the other. Tests that reproduce this: https://gist.github.com/rsto/591df55b45f8f00a110fec500872f300

General remark: the code in this PR looks good to me, but the form of this PR made review unnecessarily unpleasant.

If the PR hadn't introduced a regression, I would have approved it with the remark that I found its form subpar. But we have no established a standard for that, so it's hard to reject it if the code itself is correct:

- The PR contains a lot of commits, but the commits typically do not build on each other (e.g. they do not "tell a story"). Smaller, more focused PRs would have made it simpler to review it, and also would have got most of these smaller PRs approved in one go.
- The commit messages, both title and descriptions, and code comments are full of Claudisms. I first commented on it, then started to ignore the prose entirely and just read the code. That made things more clear, which is a sign to me that most of the prose is just noise.
- The PR defers it description to https://claude.ai/artifact/6DqaUxY5gjwEGWgQfTQVrJ?sk=D6SKMCrAqASr25HaVRX_gg and I found that document not helpful. It's full of weird LLM-speak, expects the reader to know the affected code verbatim, and it was long. Similar to the previous bullet, I chose to ignore. Also: people without a Claude account can't look at that document at all.

ksmurchison:
Superseded by #6458 - #6466

PR #6425, opened 2026-09-17
detach hidden attendees from the owning component

rjbs:
Cyrus no longer corrupts memory when restoring a calendar through the JMAP
Backup extension in which one stored version of a resource is a scheduling
poll (VPOLL) while another version carries the JMAP "hide attendees" flag.
Restoring such a resource now schedules correctly; previously it could free
the same iCalendar property twice and crash the HTTP service.


rjbs:
I am opting not to backport this.  VPOLL is an extremely unlikely configuration.

PR #6431, opened 2026-09-17
Replication failure-mode fixes

brong:
This contains fixes for two separate tasks I have in Fastmail Linear:

* CYR-1698: if cyrus is restarted while a user is being newly synced in, it can break replication
* CYR-3457: Change the sync error to something other than IMAP_MAILBOX_MOVED

The main fix is creating new mailboxes with the floor of the modseq range rather than the master's highestmodseq, meaning that a shutdown or other issue during sync doesn't leave it confused.  The rest is just new error codes.

The commits have a Claude co-author, and it wrote the tests pretty un-molested, but the code and comments have been almost entirely replaced by hand.


cassandane/tiny-tests/Replication/nocopyback_replica_ahead
```diff
@@ -0,0 +1,41 @@
...
+    eval { $self->run_replication(nosyncback => 1); };
+    my $e = $@;
+    $self->assert_not_null($e);
+    $self->assert_matches(qr/sync_client pid \d+\) exited with code/,
+                          $e->to_string());
+    $self->assert_syslog_matches($self->{instance},
+                                 qr/event=sync\.mailbox\.replica_ahead/);
+    $self->assert_syslog_does_not_match($self->{instance},
+                                        qr/failed to settle renames/);
+    $self->{instance}->getsyslog();
```
elliefm: You can't stack multiple `assert_syslog_...` checks like this: the first one will consume all the pending syslog information, and the second one will be asserting on a probably-empty log.  In this case, it only doesn't break because the second check is looking for something to not exist.

When you need to make multiple assertions against the same set of syslog lines, you should fetch the syslog lines once, and then assert against those with the usual tools.  Something like this should work:

```
my @syslog_lines = $self->{instance}->getsyslog();
$self->assert_contains(qr/.../, \@syslog_lines, $expect_count);
$self->assert_not_contains(qr/.../, \@syslog_lines);
```

`assert_contains` takes an `expect_count` parameter.  If the pattern must match >= 1 times but the exact number of matches is unpredictable, you'll need to use `assert_matches` instead, like `assert_matches(qr/.../, "@syslog_lines")`.

elliefm (changes requested):
Yeah nice.  There's a bug in one of the tests though; it's not proving what it looks like it's trying to prove.

cassandane/tiny-tests/Replication/nocopyback_replica_ahead
brong: Good catch!  This test works that way now.

elliefm (approved):

PR #6448, opened 2026-09-18
Cassandane: wire traffic logging for raw-socket http

rjbs:


rjbs:
@wolfsage As requested!

cassandane/Cassandane/Util/Wire.pm
```diff
@@ -0,0 +1,64 @@
+# SPDX-License-Identifier: BSD-3-Clause-CMU
```
wolfsage, line 1: Commit message nit:

```
Cassandane::Util::Wire adds to stupid little helpers for those to use
```

"adds to stupid" -> ? "adds two stupid" or "adds stupid" maybe

cassandane/Cassandane/Util/Wire.pm
```diff
@@ -0,0 +1,64 @@
...
+sub wire_recv ($peer, $text)
+{
+    _wire_log(">>>>>>>> from $peer", $text);
+    return;
+}
+
+sub _wire_log ($banner, $text)
+{
+    $text //= '';
+    $text =~ s/\r\n/\n/g;
```
wolfsage: I don't love that we're transforming the data we're logging but I get it I suppose. I think what I don't like is...

1. If the literal data over the wire is the exact string "\x00" (so, not a NUL byte) followed by a NUL byte, we'll log "\x00\x00". This is misleading since one of those was the verbatim string "\x00" and one was an actual NUL byte.

2. We transform both \r\n -> \n and any trailing \n\z into -> nothing, which also lies about the data we sent over the wire. That trailing \n could be important for off-by-one parsing bugs, etc etc.

I think my recommendation would be either...

1. print out the data as is (you only get it if requested or on a failure) (I think this is what Mail::IMAPTalk does?)
2. encode weird data as you are here, but encode it in such a way that \x00 and \0 are distinct, but leave \n and \r\n alone. Probably this would be any verbatim "\" becomes "\\".... so \x00\0 would be "\\x00\x00" which

wolfsage (changes requested):
Thanks for this. I think it has a few issues we could solve; what do you think..?

cassandane/Cassandane/Util/Wire.pm
rjbs: s/to/some/

I probably meant "two" but there's more than two. :)

cassandane/Cassandane/Util/Wire.pm
rjbs: Yeah, it's kinda all a mess no matter what, right?

I've dropped any munging.  I really want to replace all the telemetry stuff from tests to be less stupid, but for now "just log the bytes" is fine.

(Random thought: woah what if we just stored pcap files?)

Anyway, check a look, I amended the bottom commit to drop the munging.

wolfsage (approved):

rjbs:
GH is being weird.  This was merged.

PR #6449, opened 2026-09-18
imapd.c: keep the QRESYNC uid sequence in its own buffer

rjbs:
parse_select_params() pointed vanished_params.sequence at the same static buffer that the modifier loop reuses for the next parameter, so a parameter list like

    (QRESYNC (1299031270 420538 1:52118) ANNOTATE)

left the sequence reading "ANNOTATE".  index_vanished() then passed that to seqset_parse(), which calls fatal() on anything that isn't a sequence, and imapd aborted.  The match data already has buffers of its own; give the uid sequence one too.


brong (approved):
Approved though that buf.s syntax makes me sad every time (it should probably be `buf_cstring(&buf)` at every site, but that would be a bigger change)

PR #6452, opened 2026-09-21
cassandane: replace the wire/worker/outcome protocol

rjbs:
Okay, the main goal here is to replace the way that workers and parent communicate.

* we swap out Storable+b64 for JSON
* …which means we stop sending serialized exceptions and send a described one instead
* …during which change we make our own base class for failures, which will take over for Test::Unit::Failure when we drop it
* …and fixed some socket-handling errors while in there

* we replace the "pass half-filled struct, expect it to get filled" model with "pass args, get results"
* …and delete some of the nonsense in that struct, like "parameterized test support"

* we make forked and single-threaded tests work the same way, to eliminate weird differences between the models, with everything using the same input/outcome model
* we make "skip" work like a real test outcome, instead of "sometimes invisible, sometimes pass"

Things I *didn't* do but want to do later:

* overhaul the planner to be less weird, so we just see a queue of items to do and do them
* eliminate "reconstitute Error from a description" so the listeners and Result don't rely on the Error interface


rjbs:
@wolfsage This is item 2️⃣ (and just the first bit of 3️⃣ ) from my email about excising Test::Unit.

cassandane/Cassandane/Unit/Worker.pm
```diff
@@ -7,11 +7,17 @@ use warnings;
 
 use IO::Handle;
 use POSIX ();
-use Storable qw(freeze thaw);
-use MIME::Base64;
+use JSON::XS ();
 
 my $nextid = 1;
 
+# One message per line, so the encoding has to fit on one: JSON escapes any
```
wolfsage: I don't follow this comment fully. Why *can't* you send an object? JSON handles that fine. What will happen if you do?

Or do you mean "Don't send a blessed Perl object, but hashes/arrays are okay", not "don't send a JSON object"

wolfsage (approved):
Seems plausible!

cassandane/Cassandane/Unit/Worker.pm
rjbs: That's what i mean.

rjbs: I reworded it.

elliefm:
Nice!

Issue #15, opened 2016-08-02
sync: crash when index record has empty guid


elliefm:
This can happen in two ways:

1. The mailbox's index version is too old (<10) to contain a guid, so the guid field is initialised to zero when it is loaded
2. The mailbox's index version is recent enough (10+) to contain a guid field, but it hasn't been set for some reason (historical upgrade process?)

The crash is due to an assertion failure, but the assertion itself is good.  It's impossible to replicate messages without guids, because that's the factor used for deduplication.

For 1 at least, the sync code should check the index version when opening the mailbox, and fail immediately with a suitable error code (IMAP_MAILBOX_NOTSUPPORTED?) if it's too old to be replicated

elliefm:
We really need some Cassandane tests with built-in old-version mailboxes to test the backcompat stuff against...

elliefm:
Mailing list discussion starts here: https://lists.andrew.cmu.edu/pipermail/info-cyrus/2016-August/039051.html

elliefm:
The other question here is, how robust is our handling of null and non-null guids when the SHA1 bytes are all zeroes?

Tests (currently don't pass) here: https://github.com/cyrusimap/cyrus-imapd/compare/master...elliefm:v30/message-guid-unit-tests. If these seem like a reasonable description of how the message_guid api *should* work, then I'll fix it up so that they pass. If the API should work some other way, let's get that sorted out, make tests for it, then make it pass them.

I haven't got null/non-null-zero tests for message_guid_import/message_guid_export yet because it's not yet clear to me how they *should* work.  message_guid_import sets status to GUID_UNKNOWN -- which probably wants its own raft of tests.  message_guid_isnull treats all-zeroes as if it's GUID_NULL if it's currently GUID_UNKNOWN -- this bothers me: seems like the caller of message_guid_import should be responsible for making this determination (it knows the context in which the import occurred), not message_guid_isnull somewhere down the line.  It would be nice to get rid of GUID_UNKNOWN altogether actually, since it's only used in these two places.

brong:
I don't care about the file with sha1 of all zeros.  It doesn't happen.  If you could make it happen, you could create an arbitrary sha1 collision.  If we ever see that we need a new format anyway.

GUID_UNKNOWN can go, I'd be happy to kill it.  It's a leftover from the bad old days.

[...]

Issue #14, opened 2016-08-01
should not try to create notifications collection for non-existent users


ksmurchison:
(08:25:42 PM) brong: kmurchison: lots of this in the log - IOERROR: failed to create gmail.com!user.foobar.#notifications (Permission denied)
(08:26:28 PM) brong: caused by someone putting /dav/calendars/user/foobar@gmail.com/ in their client - that user doesn't actually exist on our system
(08:26:34 PM) brong: obviously we shouldn't even be trying to create #notifications folders like this for users who don't exist

ksmurchison:
Should be fixed with commit https://github.com/cyrusimap/cyrus-imapd/commit/cd93ca04b735afb379bee0db672904d9f3f6d649

Bron should review to see if he's satisfied with this approach.

elliefm:
@brong ^

elliefm:
@brong @ksmurchison can we close this?

ksmurchison:
If Bron isn't complaining, I'm guessing its fixed

elliefm:
that's good enough for me :)

PR #6453, opened 2026-09-22
docs.yaml: fetch the git tags from cyrusimap/cyrus-imapd.git

nwc10:
commit 6d43f86ebc2cc873 in cyrus-docker
    cyd makedocs: use "make doc" instead of docsrc's makefile

changed `cyd makedocs` to run `cyd build --with-sphinx -n`. In turn, that runs ./tools/git-version.sh to get the annotated git version. That fails with "unknown" if there are no git release tags, which is the default in the checkout of a forked repository, and the CI run aborts.

Fix this by getting the git tags from the "upstream" repository, like the other CI workflows that need to `cyd build`.


nwc10:
Failures start at run `#15` on July 8th.
Last pass was run `#14` on June 26th

https://github.com/nwc10/cyrus-imapd/actions?query=workflow%3A%22Cyrus+IMAP+CI+Doc+Gen%22

the change to `cyd builddocs` was merged on July 7th as part of https://github.com/cyrusimap/cyrus-docker/pull/91

github being github, if you follow my "actions" link, you might need to hit return on the URL bar to get it to show you more than 3 results

rjbs (approved):

PR #6455, opened 2026-09-22
Cassandane/Tiny/Loader.pm: use Package::Stash::PP

ksmurchison:
Package::Stash crashed if there were more than 323 tests 
see https://github.com/Perl/perl5/issues/24845


karenetheridge:
or you could wait for the patch

PR #6456, opened 2026-09-22
draft - CYR-3455 libconfig, but not with buf

nwc10:
As the followup from the discussion at the meeting on Monday, about adding coverage to libconfig.c that exposes paths that would leak memory

This branch does some cleanup

1. `cunit/leaksanitizer.suppress is redundant, so remove it`
2. `libconfig.c: Use errbuf for error messages, not buf`

tests

3. `cunit: more regression tests for libconfig`

a bugfix

4. `libconfig: a config line ": text" would hit an assertion failure`

refactoring to avoid leaking

5. `libconfig.c: config_add_overflowstring() must free memory on error`
6. `libconfig: config_read_file() must take and return a buffer to avoid leaks`

this is the approach without using `struct buf`.

On reflection, I realise that it's cleaner than I remembered it.


wolfsage (commented):
lgtm, though I'm amused by "remove leaksanitizer.suppress" "wait, add it back", "okay, remove it again"

wolfsage (approved):

wolfsage:
(I have approved but this is a draft?)

PR #6457, opened 2026-09-22
draft - CYR-3455 libconfig, with buf - #6456

nwc10:
This is https://github.com/cyrusimap/cyrus-imapd/pull/6456 but done with `struct buf`

First four commits are the same as the other branch:

some cleanup

1. `cunit/leaksanitizer.suppress is redundant, so remove it`
2. `libconfig.c: Use errbuf for error messages, not buf`

tests

3. `cunit: more regression tests for libconfig`

a bugfix

4. `libconfig: a config line ": text" would hit an assertion failure`

then it diverges with work on buf.c

5. `cunit: tests for buf_getline()`
6. `buf.c: Refactor buf_getline() to read line-at-a-time with fgets()`
7. `buf.c: Add buf_getline_continuation(), which folds lines ending in backslash`

then bodge it into libconfig.c

8. `XXX libconfig.c from xmalloc() to struct buf, but it CASTS AWAY CONST :-(`

then fix the actual leaks:

9. `libconfig.c: config_add_overflowstring() must free memory on error`
10. `libconfig.c: create the buf once at the top and pass it down`

on reflection, I think

1. we go with https://github.com/cyrusimap/cyrus-imapd/pull/6456 (not looked at the reported CI failure - it's getting late)
2. we keep `cunit: tests for buf_getline()` and ``buf.c: Refactor buf_getline() to read line-at-a-time with fgets()`` but put them in a different PR
3. nothing needs `buf.c: Add buf_getline_continuation(), which folds lines ending in backslash` so we, um, discard it


nwc10:
I did this (I think I fixed the valgrind fail) but github's UI doesn't show it:

```
To github.com:nwc10/cyrus-imapd.git
 + 7a2f3c21c7...a2fcf8917e CYR-3455-libconfig-buf -> CYR-3455-libconfig-buf (forced update)
```

edited, oh, it does **now**. Clearly the future isn't evenly distributed yet. Takes **several minutes**

